PR #9, opened 2021-08-11
update DG core interface (in compatible with wip ClimaCore interface) -> bickley jet 

jiahe23:
This PR is a wip update on DG backend interface so that it is compatible with the wip ClimaCore interface.

`jh/tb/bickley_cmc` is a parallel working branch on CM end as oppose to `tb/bickley_cmc` on ClimaCore. The idea is to synchronize these two branches as frequently as possible so that the development on the two backends won't diverge.


bischtob (approved):
Let' merge into the branch

Issue #17, opened 2021-08-16
Replace boilerplate macro with exports


simonbyrne:
The boilerplate macro could be replaced by simply using `export` statements inside ClimaAtmos: the [Reexport.jl](https://github.com/simonster/Reexport.jl) package might be useful for this.

This would also have the advantage of not requiring the project running the script to have all the other packages added: instead, the user would only need to add ClimaAtmos.

sandreza:
Thank you for pointing this out! I'll play around with this and open a pull request when it's done

PR #20, opened 2021-08-17
Add modules for DG backend and bring back bickley jet on sphere

jiahe23:
This PR is a first step towards a functioning DG backend interface, a follow-up effort of #13 . It adds modules for the DG core and brings back the bickley jet experiment on the sphere.

To bring back running benchmarks for the DG backend, the following will be added thru small PRs:
- To separate boilerplate for ClimaCore and ClimateMachine:  We would like to get rid of the boilerplate in the end (#17), but for the current convenience, we should keep them separate so that there is no need to import all ClimateMachine functions nor init it when running with ClimaCore;
- To add callbacks:  Abstract place holder in Interface, and ClimateMachine dependent functionality in backends;
- To optimize/simplify model physics struct;
- To add back IMEX splitting;
- To add back benchmarking tests.

### DIFF
--- a/examples/WIP_bickley_jet_dg.jl
+++ b/examples/WIP_bickley_jet_dg.jl
@@ -6,13 +6,17 @@ ClimateMachine.init()
 
 # set up parameters
 parameters = (
-    ϵ  = 0.1,  # perturbation size for initial condition
-    l  = 0.5,  # Gaussian width
-    k  = 0.5,  # Sinusoidal wavenumber
-    ρ₀ = 1.0,  # reference density
-    c  = 2,
-    g  = 10,   # gravitation constant
-    D₄ = 1e-4, # hyperdiffusion coefficient
+    g  = 0.0,
+    ρₒ = 1.0,  # reference density
+    cₛ = 1e-2, # sound speed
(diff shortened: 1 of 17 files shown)


examples/WIP_bickley_jet_dg.jl
```diff
@@ -0,0 +1,97 @@
...
+# set up model
+model = BarotropicFluidModel(
+    domain = domain,
+    physics = ModelPhysics(
+        equation_of_state = BarotropicFluid(),
+        ref_state = NoReferenceState(),
+        sources = ( 
+            coriolis = DeepShellCoriolis(), 
+        ),
+    ),
```
bischtob, line 61: Thee physics struct may ultimately be unnecessary for the barotropic fluid model, because we can just hardcode the physics. Alternatively, it might be better to simplify the physics struct.

jiahe23: I suggest we leave this for a future PR. When it comes to adding back the dry/moist atmos model, it will become clearer which part should be hard-coded and which part should be left at an option specified in the model physics struct. We will also be discussing how to simply the struct then.

PR #21, opened 2021-08-17
refactored hydostatic model to expose flux bcs

bischtob:
We refactored the hydrostatic column model to expose all boundary conditions in a way that lets us replace them by fluxes on the boundaries. The boundary condition interface for that will be part of different PR.

### DIFF
--- a/src/backends/climacore/tendencies.jl
+++ b/src/backends/climacore/tendencies.jl
@@ -41,72 +41,6 @@ function create_rhs(::ClimaCoreBackend, model::BarotropicFluidModel, function_sp
     return rhs!
 end
 
-# function create_rhs(::ClimaCoreBackend, model::HydrostaticModel, function_space::Tuple{CenterFiniteDifferenceSpace,FaceFiniteDifferenceSpace})
-#     function rhs!(dY, Y, _, t)
-#         UnPack.@unpack Cd, f, ν, ug, vg, C_p, MSLP, R_d, R_m, C_v, grav = model.parameters
-#         (Yc, Yf) = Y.x
-#         (dYc, dYf) = dY.x
-#         UnPack.@unpack ρ, u, v, ρθ = Yc
-#         UnPack.@unpack w = Yf
-#         dρ = dYc.ρ
-#         du = dYc.u
-#         dv = dYc.v
-#         dρθ = dYc.ρθ
-#         dw = dYf.w
-
-#         # density with no-flux boundary conditions S 4.4.1.
-#         gradc2f = Operators.GradientC2F()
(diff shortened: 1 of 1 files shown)


src/backends/climacore/tendencies.jl
```diff
@@ -122,45 +56,57 @@ function create_rhs(::ClimaCoreBackend, model::HydrostaticModel, function_space:
...
-        gradc2f = Operators.GradientC2F(top = Operators.SetValue(ug)) # Eq. 4.18
-        gradf2c = Operators.GradientF2C(bottom = Operators.SetValue(Cd * u_wind * u_1)) # Eq. 4.16
-        
-        A = Operators.AdvectionC2C(bottom = Operators.SetValue(0.0), top = Operators.SetValue(0.0))
-        @. du = gradf2c(ν * gradc2f(u)) + f * (v - vg) - A(w, u) # Eq. 4.8
+        # u velocity (center
+        flux_btm = Cd * u_wind * u_btm
+        flux_top = ν * (v_top - vg) / dz_top / 2
+        ∇c = Operators.GradientC2F() # Eq. 4.18
+        ∇f = Operators.GradientF2C(bottom = Operators.SetValue(flux_btm), top = Operators.SetValue(flux_top)) # Eq. 4.16
```
kmdeck, line 86: It looks like you are setting state BC on v (v_top face = vg), and on u (u_top face = ug), and then a flux BC at the bottom for the u,v equations.  (and the A operator does not add any extra constraints?)

so would the alternative be to set:
```gradc2f = Operators.GradientC2F(top  = setvalue(ug))
gradf2c = Operators.GradientF2C(bottom = setvalue(flux_btm)
@. du = gradf2c(nu*gradc2f(u))```?

and here you are explicitly making the top state BC into a flux BC as we talked about?

I dont quite understand the chain here - seeing your idea for the interface will help. Is it: (1) the user sets a state BC, (2) internally there is an intermediate function that maps that into a flux bc using something appropriate/unique for the problem and  then (3) that flux bc is passed to the create RHS function which *only* handles flux bcs?

kmdeck: is there any circumstance where approximating the derivative like this (to get a flux bc from state bc) would be inconsistent with what you would get if you explicitly set the BC like I wrote above?

bischtob: Yes, all that you wrote is what I had in mind here. The use picks some boundary condition step (1), then that boundary condition corresponds to a certain flux boundary conditions internally (2), and that flux is invoked here (3).

kmdeck (approved):
hey Toby - this looks fine to me. I'll test out something similar (mapping our various BC into flux BC).

PR #65, opened 2021-08-26
Refactor simulation to interface directly with DIffEqBase.Integrators

bischtob:
1. This replaces the simulation struct and replaces it directly with DiffEqBase.Integrators. This allows reuse of the DiffEqBase.jl interface for time stepping.
2. Remove the solve/run function in anticipation of onboarding Oceananigans.jl run function behavior.


src/Simulations/simulation.jl
```diff
@@ -1,47 +1,61 @@
...
-    Simulation(model::AbstractModel, stepper::AbstractTimestepper)
+    Simulation(model::AbstractModel, method::AbstractTimestepper; dt, tspan)
 """
-function Simulation(; model::AbstractModel, stepper::AbstractTimestepper)
-    ode_function   = make_ode_function(model)
-    state_init     = make_initial_conditions(model)
-    ode_problem    = DiffEqBase.ODEProblem(ode_function, state_init, stepper.tspan)
+function Simulation(model::AbstractModel, method; dt, tspan, kwargs...)
+    ode_function = make_ode_function(model)
+    Y = make_initial_conditions(model)
```
LenkaNovak, line 14: This may need to be split, so that you can initiate your state externally, when/if we decide to enable multiple models' state vectors to be combined and passed around as one Y vector. But we can cross that bridge then.

LenkaNovak (approved):

src/Simulations/simulation.jl
bischtob: Good point. There are so many cases :D.

PR #68, opened 2021-08-29
Tb/fieldvector inits

bischtob:
Summary: Refactoring how initial conditions are handled. For DiffEqBase integration we need to set initial conditions at `ODEProblem` level, so we need a dummy default initial condition, which I chose to be zero everywhere. This goes together with the `set!` function interface to match Oceananigans interface. This PR is building towards a cleaner and general model design useful for complicated use cases.

- [ ] Need default initial conditions (for now this is just 0.0 for all field types).
- [ ] Remove initial conditions tuple from model. Those will be handled via set functionality in PR #65.
- [ ] Refactor state vector to allow nesting (e.g, `Y.prognostic.u, Y.diagnostic.T_sfc`, etc.) in `Model`. This will be important going forward and needs good & clean design.
- [ ] Use `FieldVector` type for initial condition in `SingleColumnModel `and remove `Arraypartiation`


bischtob:
Still need to work on doing the same for the `SingleColumModel`, but that is easy.

PR #69, opened 2021-08-29
Implement general Model interface for more complex use cases

bischtob:
This is work that builds out a general flexible and lightweight model interface consistent with the current approaches at CliMA. In the future we will need to support a few use cases:

- [x] Specification of submodels (e.g., tracer, edmf, moisture, etc. - not unlike what the land team needs)
- [x] Toggling of different equations blocks (similar to the above)
- [x] Easy and transparent labeling of tendencies 
- [ ] Construction of generic rhs! and jacobian! closures that remain transparent.

PSA: This post will be updated and the list may get more items before first iteration is complete.


bischtob:
Won't be continued until #68 and #65 are merged.

src/Models/model.jl
```diff
@@ -0,0 +1,87 @@
...
+struct Coriolis{FT, S <: SteppingStyle} <: AbstractTendency 
+    rotation_rate::FT
+end
+
+# examples of a concrete constraints types
+struct Positivity{R <: Rate} <: AbstractConstraint end
+struct ThermodynamicEquilibrium{R <: Rate} <: AbstractConstraint end
+
+# returns all equations
+return_equations(model::AbstractModel)
```
charleskawczynski, line 61: It might be helpful to see an example here. For `MyModel `, would this be:
```julia
return_equations(m::MyModel) = (m.eq1, m.eq2, m.eq3)
```
?
Also, where and how will this method be used?

bischtob: Yes. I am setting up and example.

bischtob: But that you can guess the implementation is a good sign :)

bischtob:
This will be tackled when I am back.

bischtob:
Superseded by #142

PR #80, opened 2021-09-03
Fix an error in the hydrostatic ekman column example

szy21:
This PR fixes a bug in the rho_theta equation in single_colum_model.jl. The bug doesn't seem to affect the result.


bischtob (approved):
Excellent job!!!

PR #87, opened 2021-09-08
add docstrings for domains, models, and simulation modules

### LINKED ISSUE #82: Provide comprehensive docstrings for Domain, Models, and Simulation modules
**context**
Now that the documenter has been set up, we can add some more comprehensive docs.

**ideas**
I think we want to keep documentation minimal but sufficient at this stage (to not have too much maintenance). I propose that we follow [these](https://docs.julialang.org/en/v1/manual/documentation/#man-documentation) conventions. The goals are to have documentation come in through doc strings primarily so that the documentation can be easily reviewed in PRs.

**next steps**
Compare the proposed documentation guidelines with the old ones in `CM.jl` and `O.jl` and add an example or link to the contributor guide and also set up the first modules in the finalized style.

szy21:
Co-authored-by: @akshaysridhar 
Closes #82

### DIFF
--- a/src/Simulations/simulation.jl
+++ b/src/Simulations/simulation.jl
@@ -1,15 +1,9 @@
 """
     struct Simulation <: AbstractSimulation
-        # a ClimaAtmos model
-        model::AbstractModel
-        # a DiffEqBase.jl integrator used for time
-        # stepping the simulation
-        integrator::DiffEqBase.DEIntegrator
-    end
 
-    A simulation wraps an abstract ClimaAtmos model containing 
-    equation specifications and an instance of an integrator used for
-    time integration of the discretized model PDE.
+A simulation wraps an abstract ClimaAtmos `model` containing 
+equation specifications and an instance of an `integrator` used for
+time integration of the discretized model PDE.
 """
 struct Simulation <: AbstractSimulation
     model::AbstractModel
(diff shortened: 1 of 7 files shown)


src/Simulations/simulation.jl
```diff
@@ -38,7 +35,13 @@ function Simulation(model::AbstractModel, method; Y_init = nothing, dt, tspan)
...
 
 """
-    set!
+    set!(
+        simulation::AbstractSimulation,
+        submodel_name = nothing;
+        kwargs...,
+    )
+
+Set up the `simulation`.
```
bischtob, line 44: This sets the simulation state to a new state, either through an array or a function. This is signified by the "!" after the "set", aka it updates the existing simulation instance that is passed as first argument.

bischtob (approved):
There is only one clarification :), so this is great work! Please update the docstring and squash & rebase. Then feel free to r+!

src/Simulations/simulation.jl
```diff
@@ -96,8 +99,18 @@ function set!(
     nothing
 end
 
+"""
+    step!(sim::AbstractSimulation, args...; kwargs...)
+
+Step forward a simulation `sim` one time step.
+"""
 step!(sim::AbstractSimulation, args...; kwargs...) =
```
szy21, line 107: @bischtob Could I change `sim` here to `simulation` to be consistent with the `set!` function?

bischtob: Sounds good!

akshaysridhar (approved):
LGTM as a first step towards doc building. I like examples in the Oceananigans style which we can work towards in future PRs

PR #113, opened 2021-09-14
fix a bug in make_function_space for hybrid plane

szy21:
Fixes a bug (maybe?) of not using `Interval` in make_function_space for `HybridPlane`.


szy21:
I changed it because I was getting `UndefVarError: .. not defined` when using the make_function_space for `HybridPlane`. Not sure whether it is necessary (maybe I forgot to include some packages?), but I guess it might be good to be consistent with make_function_space for `Plane`.

bischtob (approved):
Good catch

PR #114, opened 2021-09-14
Updated workflow to reflect registration of ClimaCore

bischtob:
Since ClimaCore.jl is now registered, we should be able to treat it like any other package in our GH workflows.


bischtob:
bors try

PR #116, opened 2021-09-14
Fix a bug in make_function_space.jl

szy21:


szy21:
bors try

src/Domains/make_function_space.jl
```diff
@@ -30,7 +30,7 @@ function make_function_space(domain::Plane{FT}) where {FT}
         domain.nelements[2],
     )
     grid_topology = Topologies.GridTopology(mesh)
-    quad = Spaces.Quadratures.GLL{domain.npolynomial}()
+    quad = Spaces.Quadratures.GLL{domain.npolynomial + 1}()
```
bischtob, line 33: This will break some test. You could just adjust the tests by lowering the poly order by 1 in the tests.

szy21: Done!

szy21:
bors try

Issue #37, opened 2021-08-19
Implement adaptive time stepper via callbacks


bischtob:
**context**
We can leverage the callbacks infrastructure to have adaptive time stepping.

**ideas**
Reuse the calculation methods used for `OutputWriters` to calculate a new time step based on the flow. Then update the `DEIntegrator` state.

**next steps**
Finish the callbacks framework first, then prototype an example using CFL calculator.

akshaysridhar:
First iteration of `Callbacks` is now in. This is within reach in the short term.

akshaysridhar:
From Issue 41 
```
cfl monitor idea

struct CFL end
function (::CFL)(Y,Ya,integrator)
    u, v, w = Y.u.u1, Y.u.u2, Y.u.u3
    u_max, v_max, w_max = maximum.((u, v, w))
    dh, dz = minhorzdistance(axes(u)), minvertdistance(axes(w))
    dt = currenttimestep(integrator)
    cfl_x = u * dt / dh
    cfl_y = v * dt / dh
    cfl_z = w * dt / dz

    cfl_horz = maximum(cfl_x, cfl_y)
    cfl_vert = maxmium(cfl_z)    

    return (horizontal = cfl_horz, vertical = cfl_vertical)
end
```

PR #123, opened 2021-09-16
Update timestep via DiffEqBase callbacks 

akshaysridhar:
Re: Issue #37 
WIP prototype + tests.


akshaysridhar:
@jakebolewski  I think this block 
```
function get_nodal_distance(space)
    if typeof(space).name.name == :ExtrudedFiniteDifferenceSpace
        Δh_local = space.horizontal_space.local_geometry.WJ
        Δv_local = diff(space.vertical_mesh.faces)
        return (Δh = Δh_local, Δv = Δv_local)
    elseif typeof(space).name.name == :SpectralElementSpace2D
        Δh_local = space.local_geometry.WJ
        return (Δh = Δh_local)
    else
        @show ("Method for $(typeof(space).name.name) undefined")
    end
end
```  highlights the issue you pointed out (Slack) about accessing fields via direct field access - is there a more sensible way to access local coordinate spacing using existing ClimaCore functions?

**Comment by jakebolewski** [comment]
it would be more natural to write it using multiple dispatch:

Doesn't work but hopefully the idea is clear:

```
function get_nodal_difference(space::Spaces.FiniteDifferenceSpace)
   Spaces.local_geometry_data(space).WJ
end

function get_nodal _difference(space::SpectralElementSpace2D)
     Space.local_geometry_data(space).WJ
end

function get_nodal_distance(space:Spaces.ExtrudedFiniteDifferenceSpace)
    \deltah = get_nodal_difference(Spaces.horizontal_space(space))
    \deltav = get_nodal_difference(Spaces.column_space(space)) 
    return 
end
```

We might want to add the methods that you need here for all to use, I don't think we have a consistent way to get at nodal spaces and the like for abstract spaces.

**Comment by akshaysridhar** [comment]
> it would be more natural to write it using multiple dispatch:
> 
> Doesn't work but hopefully the idea is clear:
> 
> ```
> function get_nodal_difference(space::Spaces.FiniteDifferenceSpace)
>    Spaces.local_geometry_data(space).WJ
> end
> 
> function get_nodal _difference(space::SpectralElementSpace2D)
>      Space.local_geometry_data(space).WJ
> end
> 
> function get_nodal_distance(space:Spaces.ExtrudedFiniteDifferenceSpace)
>     \deltah = get_nodal_difference(Spaces.horizontal_space(space))
>     \deltav = get_nodal_difference(Spaces.column_space(space)) 
>     return 
> end
> ```
> 
> We might want to add the methods that you need here for all to use, I don't think we have a consistent way to get at nodal spaces and the like for abstract spaces.

Multiple dispatch makes most sense, I agree.

**Comment by szy21** [comment]
@akshaysridhar what's the status of this PR?

File: src/Models/Nonhydrostatic2DModels/nonhydrostatic_2d_model.jl [hunk 1]
```diff
@@ -140,3 +140,10 @@ function Models.make_ode_function(model::Nonhydrostatic2DModel{FT}) where {FT}
         return dY
     end
 end
+
+function get_velocities(u, model::Nonhydrostatic2DModel)
```
jiahe23, line 144: I suggest use `model_fields` instead of `u` here because it is easily confused with the zonal velocity.

src/Models/Nonhydrostatic2DModels/nonhydrostatic_2d_model.jl
```diff
@@ -140,3 +140,10 @@ function Models.make_ode_function(model::Nonhydrostatic2DModel{FT}) where {FT}
...
+    u_1 = getproperty(u.nhm,:ρuh) ./ getproperty(u.nhm, :ρ)
```
jiahe23, line 145: `u.nhm` is hard coded here. I think `getproperty(u, model.name)` should give `u.nhm` so that the hard coding can be avoided.

src/Models/ShallowWaterModels/shallow_water_model.jl
```diff
@@ -85,3 +85,9 @@ function Models.make_ode_function(model::ShallowWaterModel{FT}) where {FT}
         return dY
     end
 end
+
+function Models.get_velocities(Y, model::ShallowWaterModel)
+    u_1 = Y.swm.u.components.data.:1
```
jiahe23, line 90: same here

src/Models/SingleColumnModels/single_column_model.jl
```diff
@@ -114,3 +114,9 @@ function Models.make_ode_function(model::SingleColumnModel{FT}) where {FT}
 
     return rhs!
 end
+
+function get_velocities(u, model::SingleColumnModel)
+    u_1 = getproperty(u.scm, :uv)
```
jiahe23, line 119: same here

jiahe23:
A general comment in the `get_velocities` function: the `name` field in Model struct is a Symbol that is designed to be used as an identifier of the model. We could avoid hard code it in the function.

Issue #122, opened 2021-09-15
Refactor SingleColumModel to work with BC interface


bischtob:
**context**
Bring back the old BC interface into the SingleColumnModels

**steps**
1. Go [here](https://github.com/CliMA/ClimaAtmos.jl/blob/029d182e24057814d1a12e6ddddc6873764aa638/src/Models/single_column_models/single_column_model.jl) and copy the code over carefully while keeping the regression tests running as they are now.
2. Keep the code coverage great :)

PR #126, opened 2021-09-21
Re-introducing model.boundary_conditions

LenkaNovak:
This addresses issue #122, and allows boundary fluxes to be prescribed externally. 

The next iteration / PR should 
- abstract the currently hard-coded top BC for `uv`
- generalist `get_boundary_flux` for any boundary. Currently all surface fluxes are applied at the bottom boundary via `first_interior(v)` 
- add a more general interface, such as that in Land's [rhs](https://github.com/CliMA/LandHydrology.jl/blob/c2e3fc1fd2b29d2381a5edfabd292ce1de6df3b3/src/SoilModel/right_hand_side.jl#L179) with their BCs specified [here](https://github.com/CliMA/LandHydrology.jl/blob/c2e3fc1fd2b29d2381a5edfabd292ce1de6df3b3/src/SoilModel/boundary_conditions.jl#L376). However, `ClimaSimulations` will most likely introduce a new way of prescribing BCs, so this point may depend on the interim needs of developers. 
 - decide if `get_boundary_flux` is better than `set_boundary_flux!` (see discussion below)

### DIFF
--- a/src/BoundaryConditions/flux_calculations.jl
+++ b/src/BoundaryConditions/flux_calculations.jl
@@ -1,0 +1,63 @@
+# General
+"""
+    get_boundary_flux(model, ::NoFluxCondition, _...)
+
+Construct zero boundary fluxes for a scalar or 2-element vector(x,y)
+"""
+function get_boundary_flux(model, ::NoFluxCondition, var::Fields.Field, Ym, Ya)
+    FT = eltype(Ym)
+    flux = Geometry.Cartesian3Vector(FT(0))
+end
+
+# Momentum-specific
+"""
+    get_boundary_flux(model, ::DragLawCondition, _...)
+
+Construct non-zero boundary fluxes using the bulk formula and constant or Mohnin-Obukhov-based drag coefficient, Cd
+"""
+function get_boundary_flux(model, bc::DragLawCondition, uv, Ym, Ya)
+    FT = eltype(Ym)
+    coefficients = eltype(bc.coefficients) == FT ? bc.coefficients :
+        bc.coefficients(Ym, Ya)
+
+    uv_1 = first_interior(Ym.uv)
+    u_wind = LinearAlgebra.norm(uv_1)
+
+    flux = Geometry.Cartesian3Vector(coefficients.Cd * u_wind) ⊗ uv_1
+end
+
+
+# Potential temperature density-specific
+"""
+    get_boundary_flux(model, ::BulkFormulaCondition, _...)
+
+    Construct non-zero boundary fluxes using the bulk formula and constant or Mohnin-Obukhov-based heat transfer coefficient, Ch
+"""
+function get_boundary_flux(
+    model,
+    bc::BulkFormulaCondition,
+    ρθ::Fields.Field,
+    Ym,
+    Ya,
+)
+    FT = eltype(Ym)
+    coefficients = eltype(bc.coefficients) == FT ? bc.coefficients :
+        bc.coefficients(Ym, Ya)
+    θ_sfc = bc.θ_sfc
+
+    ρ_1 = first_interior(Ym.ρ)
+    ρθ_1 = first_interior(Ym.ρθ)
+    uv_1 = first_interior(Ym.uv)
+    u_wind = LinearAlgebra.norm(uv_1)
+
+    flux = Geometry.Cartesian3Vector(
+        coefficients.Ch * u_wind * ρ_1 * (ρθ_1 / ρ_1 - θ_sfc),
+    )
+end
+
+# helpers
+"""
+    first_interior(f)
+- obtains the first interior datapoint of variable `v`
+"""
+first_interior(v) = Operators.getidx(v, Operators.Interior(), 1)
--- a/src/BoundaryConditions/flux_conditions.jl
+++ b/src/BoundaryConditions/flux_conditions.jl
@@ -19,6 +19,16 @@ struct CustomFluxCondition <: AbstractBoundaryCondition
 end
 
 """
-    DragLawCondition <: AbstractFluxBoundaryCondition
+    DragLawCondition{C} <: AbstractFluxBoundaryCondition
 """
-struct DragLawCondition <: AbstractBoundaryCondition end
+struct DragLawCondition{C} <: AbstractBoundaryCondition
+    coefficients::C
+end
+
+"""
+    BulkFormulaCondition{C, T}
+"""
+struct BulkFormulaCondition{C, T} <: AbstractBoundaryCondition
+    coefficients::C
+    θ_sfc::T
+end
--- a/src/Models/SingleColumnModels/single_column_model.jl
+++ b/src/Models/SingleColumnModels/single_column_model.jl
@@ -51,28 +51,35 @@ function Models.make_ode_function(model::SingleColumnModel{FT}) where {FT}
         w = Ym.w
         ρθ = Ym.ρθ
 
+        # gather boundary conditions
+        bc_ρ = model.boundary_conditions.ρ
+        bc_uv = model.boundary_conditions.uv
+        bc_w = model.boundary_conditions.w
+        bc_ρθ = model.boundary_conditions.ρθ
+
         # density
+        flux_bottom = get_boundary_flux(model, bc_ρ.bottom, ρ, Ym, Ya)
+        flux_top = get_boundary_flux(model, bc_ρ.top, ρ, Ym, Ya)
         If = Operators.InterpolateC2F()
         ∂f = Operators.GradientC2F()
         ∂c = Operators.DivergenceF2C(
-            bottom = Operators.SetValue(Geometry.Cartesian3Vector(zero(FT))),
-            top = Operators.SetValue(Geometry.Cartesian3Vector(zero(FT))),
+            bottom = Operators.SetValue(flux_bottom),
+            top = Operators.SetValue(flux_top),
         )
         @. dρ = -∂c(w * If(ρ))
 
         # potential temperature
+        flux_bottom = get_boundary_flux(model, bc_ρθ.bottom, ρθ, Ym, Ya)
+        flux_top = get_boundary_flux(model, bc_ρθ.top, ρθ, Ym, Ya)
         If = Operators.InterpolateC2F()
         ∂f = Operators.GradientC2F()
         ∂c = Operators.DivergenceF2C(
-            bottom = Operators.SetValue(Geometry.Cartesian3Vector(zero(FT))),
-            top = Operators.SetValue(Geometry.Cartesian3Vector(zero(FT))),
+            bottom = Operators.SetValue(flux_bottom),
+            top = Operators.SetValue(flux_top),
         )
         # TODO!: Undesirable casting to vector required
         @. dρθ =
             -∂c(w * If(ρθ)) + ρ * ∂c(Geometry.CartesianVector(ν * ∂f(ρθ / ρ)))
-
-        uv_1 = Operators.getidx(uv, Operators.Interior(), 1)
-        u_wind = LinearAlgebra.norm(uv_1)
 
         A = Operators.AdvectionC2C(
             bottom = Operators.SetValue(Geometry.Cartesian12Vector(0.0, 0.0)),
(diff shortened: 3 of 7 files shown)


src/BoundaryConditions/flux_calculations.jl
```diff
@@ -0,0 +1,70 @@
...
+    uv_1 = Operators.getidx(Ym.uv, Operators.Interior(), 1)
+    u_wind = LinearAlgebra.norm(uv_1)
+
+    flux = Geometry.Cartesian3Vector(coefficients.Cd * u_wind) ⊗ uv_1
+end
+
+
+# Potential temperature density-specific
+"""
+    get_boundary_flux(model, ::DragLawCondition, _...)
```
kmdeck, line 45: ::BulkFormulaCondition

src/BoundaryConditions/flux_calculations.jl
```diff
@@ -0,0 +1,70 @@
...
+
+Construct non-zero boundary fluxes using the bulk formula and constant or Mohnin-Obukhov-based drag coefficient, Cd
+"""
+
+function get_boundary_flux(model, bc::DragLawCondition, uv, Ym, Ya)
+    FT = eltype(Ym)
+    coefficients = eltype(bc.coefficients) == FT ? bc.coefficients :
+        bc.coefficients(Ym, Ya)
+
+    uv_1 = Operators.getidx(Ym.uv, Operators.Interior(), 1)
```
kmdeck, line 36: this may not be an issue, but I think the `1` indicates that you get the top interior value of uv_1. Does that mean someone could not use the BC `DragLawCondition` at the bottom of the domain or on another face?

src/BoundaryConditions/flux_calculations.jl
```diff
@@ -0,0 +1,70 @@
...
+    ρθ::Fields.Field,
+    Ym,
+    Ya,
+)
+    FT = eltype(Ym)
+    coefficients = eltype(bc.coefficients) == FT ? bc.coefficients :
+        bc.coefficients(Ym, Ya)
+    T_sfc = bc.T_sfc
+
+    ρ_1 = Operators.getidx(Ym.ρ, Operators.Interior(), 1)
```
kmdeck, line 62: similar - this is hard coded to apply only at the bottom of atmos, but here that's OK b/c of the type of flux.

src/Models/SingleColumnModels/single_column_model.jl
```diff
@@ -51,21 +51,31 @@ function Models.make_ode_function(model::SingleColumnModel{FT}) where {FT}
...
         ρθ = Ym.ρθ
 
+        # gather boundary conditions
+        bc_ρ = model.boundary_conditions.ρ
+        bc_uv = model.boundary_conditions.uv
+        bc_w = model.boundary_conditions.w
+        bc_ρθ = model.boundary_conditions.ρθ
+
         # density
+        flux_bottom = get_boundary_flux(model, bc_ρ.bottom, ρ, Ym, Ya)
```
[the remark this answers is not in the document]
kmdeck: (where the tuple of faces is known from the domain type, e.g.)

src/BoundaryConditions/flux_calculations.jl
charleskawczynski: I'll echo what @jakebolewski has suggested: let's make a helper function for this. cc @szy21

charleskawczynski: The reason is that this code assumes that `getidx` and `Interior` live in `Operators`. We should minimize the number of places where we assume the structure of ClimaCore-- we're interfacing with it, and we should minimize this interface.

src/BoundaryConditions/flux_calculations.jl
```diff
@@ -0,0 +1,70 @@
...
+    ρθ_1 = Operators.getidx(Ym.ρθ, Operators.Interior(), 1)
+    uv_1 = Operators.getidx(Ym.uv, Operators.Interior(), 1)
```
charleskawczynski, line 64: Let's call the locally defined helper function.

src/BoundaryConditions/flux_conditions.jl
```diff
@@ -19,6 +19,16 @@ struct CustomFluxCondition <: AbstractBoundaryCondition
 end
 
 """
-    DragLawCondition <: AbstractFluxBoundaryCondition
+    DragLawCondition{C} <: AbstractFluxBoundaryCondition
 """
-struct DragLawCondition <: AbstractBoundaryCondition end
+mutable struct DragLawCondition{C} <: AbstractBoundaryCondition
```
kmdeck, line 24: someone would know better, but you can also make the struct immutable as long as the containers it has are mutable (like e.g. an array) - then you can update struct.container_contents. Not sure of the pros/cons to each approach though or if it matters!

src/BoundaryConditions/flux_conditions.jl
```diff
@@ -19,6 +19,16 @@ struct CustomFluxCondition <: AbstractBoundaryCondition
...
+    coefficients::C
+end
+
+"""
+    BulkFormulaCondition{C, T}
+"""
+mutable struct BulkFormulaCondition{C, T} <: AbstractBoundaryCondition
```
charleskawczynski, line 31: Just to note, `mutable` structs are not allowed on the GPU, so an instance of this struct cannot reside in a GPU kernel.

src/BoundaryConditions/flux_calculations.jl
```diff
@@ -0,0 +1,70 @@
+# General
+"""
+    get_boundary_flux(model, ::NoFluxCondition, _...)
+
+Construct zero boundary fluxes for a scalar or 2-element vector(x,y)
+"""
+
```
charleskawczynski, line 7:
```suggestion
```

kmdeck (approved):
looks good to me - left a couple of comments.

src/BoundaryConditions/flux_calculations.jl
charleskawczynski: Otherwise the doc string is broken

src/BoundaryConditions/flux_calculations.jl
```diff
@@ -0,0 +1,70 @@
...
+end
+
+
+# Potential temperature density-specific
+"""
+    get_boundary_flux(model, ::DragLawCondition, _...)
+
+    Construct non-zero boundary fluxes using the bulk formula and constant or Mohnin-Obukhov-based heat transfer coefficient, Ch
+"""
+
```
charleskawczynski, line 49:
```suggestion
```

src/BoundaryConditions/flux_calculations.jl
```diff
@@ -0,0 +1,70 @@
...
+        bc.coefficients(Ym, Ya)
+    T_sfc = bc.T_sfc
+
+    ρ_1 = Operators.getidx(Ym.ρ, Operators.Interior(), 1)
+    ρθ_1 = Operators.getidx(Ym.ρθ, Operators.Interior(), 1)
+    uv_1 = Operators.getidx(Ym.uv, Operators.Interior(), 1)
+    u_wind = LinearAlgebra.norm(uv_1)
+
+    flux = Geometry.Cartesian3Vector(
+        coefficients.Ch * u_wind * ρ_1 * (T_sfc - ρθ_1 / ρ_1),
```
szy21, line 68: If I understand the operators correctly, here it should be `(ρθ_1 / ρ_1 - T_sfc)`, or rather `(ρθ_1 / ρ_1 - θ_sfc)`? You can see this is consistent with how you define the momentum flux in Line 39.

src/BoundaryConditions/flux_calculations.jl
LenkaNovak: Good points. I wonder if this will be addressed by the `ClimaSimulations` interface in a more general way? In the meantime, I'm happy to add a `get_boundary_index` function, or anything else that @jakebolewski would recommend.

src/BoundaryConditions/flux_conditions.jl
LenkaNovak: 👍 on it... :)

src/BoundaryConditions/flux_calculations.jl
LenkaNovak: Ooh yes, indeed, thanks for spotting the sign. :) (p.s., θ_sfc = T_sfc in the case)

szy21: Just curious, why is this the case? Is it assuming the surface pressure to be the same as MSLP?

LenkaNovak: It is, for this simple test case. But I corrected the notation to be consistent, as you suggested. :)

src/BoundaryConditions/flux_calculations.jl
```diff
@@ -0,0 +1,63 @@
+# General
+"""
+    get_boundary_flux(model, ::NoFluxCondition, _...)
+
+Construct zero boundary fluxes for a scalar or 2-element vector(x,y)
+"""
+function get_boundary_flux(model, ::NoFluxCondition, var::Fields.Field, Ym, Ya)
```
jb-mackay, line 7: I'll echo @jakebolewski on naming conventions: 

> One note, the convention used in Julia is to neglect get_ and have mutating set_ with a bang !:
> ex: `get_interior_values` -> interior_values / set_interior_values! 

(https://github.com/CliMA/LandHydrology.jl/pull/41#issuecomment-924184894)

LenkaNovak: Good point. It depends on whether you're prescribing the flux or calculating it, so technically you could be either getting or setting!, depending on your boundary conditions. I was following the naming convention from the issue linked to this PR. I will note this in the TODO list above though (which I will later convert to an issue), and let the ClimaAtmos team decide.

szy21:
I just tagged @sandreza as a reviewer because he offered 😀

sandreza:
I'll take a look later today and leave some comments :)

src/BoundaryConditions/flux_conditions.jl
charleskawczynski: > someone would know better, but you can also make the struct immutable as long as the containers it has are mutable (like e.g. an array)

For them to reside on the GPU, they need to be stack-allocated, that's why we used `SVector`s in ClimateMachine.jl's `AtmosModel`. Mutable arrays (specifically, `MArray`s) _can_ reside on the GPU, assuming that the compiler can infer that they don't escape the kernel. But that was typically only done in the ClimateMachine.jl's backend RHS kernels.

All that said, the solution currently implemented looks fine to me.

sandreza (commented):
It seems like tests are currently failing, but this is not an issue with the current PR! I'll create a separate issue for this. Aside from the prior comments on inner constructors, this is looking good to merge! 

edit: It seems that the issue with the tests not passing is due to my local setup, rather than something inherent in ClimaAtmos

src/BoundaryConditions/flux_calculations.jl
akshaysridhar: It seems the intent of this function here is clear, given some prognostic and auxiliary variables Y, Ya, Ym, "get" the appropriate boundary flux (`set` here would imply that the `ClimaCore.Operator` is aware of the b.c. at this point which isn't the case, right? ); the "set" operation is carried out with the corresponding `ClimaCore.Operators(SetValue...` or `ClimaCore.Operators(SetGradient...` e.g. line 66 in the single column example,  and then assigned to the appropriate `face` identifier. If you flag this as an issue we can make sure this is noted as we think about how ClimaSimulations is built.

src/Models/SingleColumnModels/single_column_model.jl
```diff
@@ -51,45 +51,54 @@ function Models.make_ode_function(model::SingleColumnModel{FT}) where {FT}
...
 
         # uv
-        bcs_bottom = Operators.SetValue(
-            Geometry.Cartesian3Vector(Cd * u_wind) ⊗ uv_1,
-        )
-        bcs_top = Operators.SetValue(uvg)
+        flux_bottom = get_boundary_flux(model, bc_uv.bottom, uv, Ym, Ya)
+
+        bcs_bottom = Operators.SetValue(flux_bottom)
+        bcs_top = Operators.SetValue(uvg) # this needs abstraction
```
akshaysridhar, line 93: This abstraction is something we can address as we consider changes to the interface i.e. application of boundary conditions to faces (associated with the `Domain`, and tagged at the `Model` level) @glwagner

src/BoundaryConditions/flux_conditions.jl
```diff
@@ -19,6 +19,16 @@ struct CustomFluxCondition <: AbstractBoundaryCondition
...
+struct DragLawCondition{C} <: AbstractBoundaryCondition
+    coefficients::C
+end
+
+"""
+    BulkFormulaCondition{C, T}
+"""
+struct BulkFormulaCondition{C, T} <: AbstractBoundaryCondition
+    coefficients::C
+    θ_sfc::T
```
akshaysridhar, line 33: This generalises to representations of \theta_sfc as functions of space / time right ?

akshaysridhar (approved):

PR #130, opened 2021-09-30
First iteration of simulation restart option

akshaysridhar:
Extends JLD2 output callback to save at user-defined time intervals (previous output was at simulation final time) 

Extends `simulation` to allow users to define a restart file-path and new simulation end-time to pick-up simulation at some intermediate timestep for which model+integrator data (via jld2 files) is available. Requires model types (including domains) to be consistent across simulation model and restart model. (Currently does not include automated directory look-up/ scan for latest checkpoint files)

### DIFF
--- a/src/Simulations/simulation.jl
+++ b/src/Simulations/simulation.jl
@@ -7,6 +7,7 @@
         integrator::DiffEqBase.DEIntegrator
         # user defined callback operations 
         callbacks::Union{DiffEqBase.CallbackSet, DiffEqBase.DiscreteCallback, Nothing}
+        restart::AbstractRestart
     end
 
 A simulation wraps an abstract ClimaAtmos `model` containing 
@@ -21,6 +22,7 @@ struct Simulation <: AbstractSimulation
         DiffEqBase.DiscreteCallback,
         Nothing,
     }
+    restart::AbstractRestart
 end
 
 """
@@ -37,6 +39,7 @@ function Simulation(
     dt,
     tspan,
     callbacks = nothing,
+    restart = NoRestart(),
 )
 
     # inital state is either default or set externally 
@@ -54,7 +57,8 @@ function Simulation(
     integrator =
         DiffEqBase.init(ode_problem, method, dt = dt, callback = callbacks)
 
-    return Simulation(model, integrator, callbacks)
+    restart = restart
+    return Simulation(model, integrator, callbacks, restart)
 end
 
 """
@@ -80,7 +84,22 @@ function set!(
         # we need to use the reinit function because we don't 
         # have direct state access using DiffEqBase. For this
         # we need to copy
-        Y = copy(simulation.integrator.u)
+        if simulation.restart isa NoRestart
+            Y = copy(simulation.integrator.u)
+            time_final = simulation.integrator.sol.prob.tspan[2]
+        else
+            filename = simulation.restart.restartfile
+            restart_data = load(filename)
+            @assert "model" in keys(restart_data)
+            if restart_data["model"] == simulation.model
+                new_end_time = simulation.restart.end_time
+                restart_integrator = restart_data["integrator"]
+                Y = copy(restart_integrator.u)
+                time_final = new_end_time
+            else
+                throw(ArgumentError("Restart file needs a compatible model."))
+            end
+        end
 
         # get fields for this submodel from integrator state
         if submodel_name === nothing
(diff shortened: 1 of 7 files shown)


szy21:
Just to clarify, does this only add the restart option, or does it also add the ability for saving at user-defined time intervals (as in the description)?

src/Simulations/simulation.jl
```diff
@@ -80,7 +84,22 @@ function set!(
         # we need to use the reinit function because we don't 
         # have direct state access using DiffEqBase. For this
         # we need to copy
-        Y = copy(simulation.integrator.u)
+        if simulation.restart isa NoRestart
+            Y = copy(simulation.integrator.u)
+            tf = simulation.integrator.sol.prob.tspan[2]
```
jiahe23, line 89: I would give `tf` a more explicit varname, e.g., `time_end`

jiahe23:
I left one comment, otherwise it looks good.

test/test_callbacks.jl
```diff
@@ -44,13 +45,30 @@ simulation = Simulation(
...
+@info "Testing Simulation Restart"
+simulation = Simulation(
+    model,
+    SSPRK33(),
+    dt = 0.01,
+    tspan = (0.0, 0.03),
+    callbacks = cb_set,
+    restart = Restart(
+        restartfile = joinpath(cb_1.filedir, cb_1.filename * "_0.02.jld2"),
+        end_time = 0.05,
```
szy21, line 62: Just to make sure, if using restart, t_span[2] doesn't matter anymore and the end_time determines the end of the simulation, right?

test/test_callbacks.jl
```diff
@@ -44,13 +45,30 @@ simulation = Simulation(
...
+@test simulation.integrator.t == 0.05
+
+
+@info "Deleting Test Output Files"
+@test isfile(joinpath(cb_1.filedir, cb_1.filename * "_0.01" * ".jld2")) == true
+@test isfile(joinpath(cb_2.filedir, cb_2.filename * "_0.02" * ".jld2")) == true
 
-rm(joinpath(cb_1.filedir, cb_1.filename * ".jld2"))
-rm(joinpath(cb_2.filedir, cb_2.filename * ".jld2"))
+rm(cb_1.filedir, recursive = true)
```
szy21, line 74: Should we remove cb_2 too?

src/Simulations/simulation.jl
akshaysridhar: > Just to clarify, does this only add the restart option, or does it also add the ability for saving at user-defined time intervals (as in the description)?
> 
> Edit: I see there is a change in the JLD2Output that allows output at user-defined time intervals.

1) It allows outputs (JLD2) at user defined time intervals. 
2) It allows restarts using a user-specified (JLD2) checkpoint file (which is the same format as the "output" file), since they store solution objects (in these instances, an `integrator` that can be used to define `ClimaCore Fields)

test/test_callbacks.jl
akshaysridhar: I should generalise this to `filedir` for the test-set to avoid confusion, but the recursive delete removes all such temporary files created as part of the callback  / output test process.

test/test_callbacks.jl
akshaysridhar: Yes, if `Restart` is defined, then `Restart.end_time` determines the final solution time.

src/Simulations/simulation.jl
akshaysridhar: 1) It allows outputs (JLD2) at user defined time intervals. 
2) It allows restarts using a user-specified (JLD2) checkpoint file (which is the same format as the "output" file), since they store solution objects (in these instances, an `integrator` that can be used to define `ClimaCore Fields)

akshaysridhar: @jiahe23 resolved tf to `time_final`

szy21 (approved):
LGTM, thanks! I don't know why the codecov decreases - it seems everything is included in the tests.

akshaysridhar:
> LGTM, thanks! I don't know why the codecov decreases - it seems everything is included in the tests.

I think I see why; a "fail" case i.e. in the code it is designed to throw an error if the `model` types between restart and sim don't match.

szy21:
> > LGTM, thanks! I don't know why the codecov decreases - it seems everything is included in the tests.
> 
> I think I see why; a "fail" case i.e. in the code it is set up to throw an error if the `model` types between `restart` and `sim` don't match; I think that's the line with coverage issues

Oh, I think that's fine? Feel free to squash, rebase, and merge:)

akshaysridhar:
bors try

PR #136, opened 2021-10-22
Add the geostropic flow shallow water test case

### LINKED ISSUE #135: Add a shallow water on a 2D sphere test case
**context**
We should add one (or more) shallow water on the sphere test cases.

**next step**
Turn [this](https://github.com/CliMA/ClimaCore.jl/blob/main/examples/sphere/shallow_water.jl) test into a test case that is used for regression and validation in ClimaAtmos.

szy21:
Closes #135


bischtob:
@szy21, what is this?

szy21:
> @szy21, what is this?

I was trying to add William Test 2 before I left but got stuck on adding Coriolis. I will take a look again soon but may need some help. I may also wait for the new version of ClimaCore this week, as the shallow water equation has changed.

PR #141, opened 2021-12-03
Housekeeping 🧹🧹🧹

### LINKED ISSUE #119: Figure out why our tests don't work with Float32
**context**
Some of our integration tests don't pass with Float32.

**ideas**
no idea why, but last time I checked I thought it was a `ClimaCore.jl` issue.

**next steps**
Set up some test and track down which sub-component breaks the Float type.

dennisYatunin:
Before I add implicit solvers and radiation, I wanted to clean up a couple of things. Here is a complete list of all the changes in this PR:
- Fix type instabilities
    - Replace all abstract field types with concrete field types; e.g., `Integer` -> `Int`, `AbstractModel` -> `M <: AbstractModel`
    - Make functions specialize on type arguments; e.g., `FT::DataType` -> `::Type{FT}` (not sure if this actually improves performance, but it's done in all our other code)
    - Fix `if`-statement type instability in `initial_conditions/dry_rising_bubble_2d.jl` (not sure if this will still be necessary after Julia 1.7, but might as well)
    - Replace `1.0` with `FT(1)` and `0.0` with `FT(0)` in boundary condtions for `Nonhydrostatic2DModel` and `SingleColumnModel`
- Clean up imports and exports
    - Remove `get_boundary_flux` from the `Models` module while leaving the definition in the `BoundaryConditions` module, which gets rid of a precompilation warning
    - Move `export` statements so that they run before `include` statements, allowing submodules to access exported symbols from parent modules
    - Remove imports of unused symbols
    - Remove unnecessary qualifiers for imported symbols
    - Utilize relative paths for importing from the `ClimaAtmos` module in `/src`
    - Organize imported symbols in a manner that is consistent across files
- Clean up `Domains` module
    - Remove overloaded methods of `ndims`, `length`, and `size` for `AbstractDomain`, since they inconsistently extend the `AbstractArray` interface
    - Remove unnecessary typecasts to `FT`
    - Use consistent interval notation
    - Remove `PeriodicPlane`, which confusingly looks like a separate type from `Plane` and `HybridPlane` (both of which can be periodic)
    - Make `Plane` have a default keyword argument `periodic = (true, true)`
- Clean up `BoundaryConditions` module
    - Remove function stub for `get_boundary_flux`, since the module already provides methods for this function
- Clean up `Models` module
    - Make default methods for `default_initial_conditions` and `make_ode_function` throw errors indicating missing implementations
- Clean up `Callbacks` module
    - Remove default method for `generate_callback`, since the module already provides a method for the only possible input type (`JLD2Output`)
- Clean up `Simulations` module
    - Remove definition of `AbstractSimulation`, as all simulation code is specific to `Simulation`
    - Make `step!` a new function, rather than an overload of `DiffEqBase.step!`
- Clean up `/test`
    - Rewrite `test_domains.jl` so that it is more succinct and so that it tests `make_function_space`
    - Prevent `test_callbacks.jl` from generating an empty folder after it is done running
    - Combine `test_regression.jl` and `test_validation.jl` into `test_simulations.jl`, reducing code duplication and adding missing unit test for Ekman column
    - Ensure that running `runtests.jl` does not print anything (unit tests should be automated, so they should not rely on visual inspection)
    - Remove `Float32` type warnings for simulations, since the type instability fixes allow simulations to run with `Float32`

Note that the use of `FT` in boundary conditions will need to be rethought as I add implicit solvers, since `autodiff` solvers can pass either a `Float64` `FieldVector` or a `Dual` `FieldVector` to the tendency function, depending on the stage of the computation.

There are some other things I'd like to clean up (particularly the unit tests), but I'll leave those for later.

Closes #119


szy21:
Looks good to me in general, thanks @dennisYatunin! Should we worry about the codecov change?

dennisYatunin:
Ah, I didn't realize the coverage would go down by that much. I removed all calls to `show(<something>)` from the unit tests, since unit tests shouldn't really print anything. I'll restore the old coverage by adding back `show(IOBuffer(), <something>)`, which will just print the output of `show` to a meaningless buffer. This isn't really a unit test, though; an actual unit test for `show` would turn the buffer into a string and compare the string to its expected value. That's a tad overkill for a simple printing function, but I could do it if anyone wants.

glwagner:
> Ah, I didn't realize the coverage would go down by that much. I removed all calls to `show(<something>)` from the unit tests, since unit tests shouldn't really print anything. I'll restore the old coverage by adding back `show(IOBuffer(), <something>)`, which will just print the output of `show` to a meaningless buffer. This isn't really a unit test, though; an actual unit test for `show` would turn the buffer into a string and compare the string to its expected value. That's a tad overkill for a simple printing function, but I could do it if anyone wants.

doctests / documentation builds are a good place to test `show`, you may not need to explicitly test them elsewhere.

dennisYatunin:
I've added a documentation commit to this PR. It adds doctests for `show`, as Greg suggested. Here is a list of the additional changes:
- Allow `ClimaAtmos` to be imported in `docs/make.jl` by adding it to the `LOAD_PATH` (fixes a documentation build error)
- Add missing docstrings to docs (fixes a documentation build warning)
- Allow docs to be built locally for testing by changing `prettyurls` argument to `makedocs`
- Clean up `show` methods
    - Make everything print to `io`, as opposed to having some things print to `io` and some to `stdout` (allows doctests to work properly)
    - Improve formatting and simplify code
- Add doctests
    - Test `show` methods of structs in their docstrings
    - Add `doctest` to `test/runtests.jl` (increases code coverage)
- Add function stubs for function docstrings
- Clean up docstrings
    - Merge multiple method docstrings into function docstrings, and move struct-specific information from method docstrings to struct docstrings
    - Remove docstrings for types when preferred constructor docstrings are available
    - Remove docstrings of helper functions (i.e., `first_index`) and internal types (e.g., `AbstractHorizontalDomain`)
    - Fix improper tabs, spacing, and typos, and rewrite particularly unclear docstrings

bischtob (approved):
This looks all good to me.

bischtob:
@dennisYatunin, please squash, rebase, and bors r+

PR #150, opened 2021-12-31
Remove ShallowWaterModel

bischtob:
We won't really use the SWM and it becomes just a testing and maintenance burden.


bischtob:
bors try

bischtob:
bors try

PR #172, opened 2022-01-14
Add utils module for initial conditions and other things

bischtob:
PR moves initial conditions to new Utils module. Also reorganizes regression tests so that they can be called individually. Also condenses integration and regression tests into one section to speed up testing.


src/Utils/InitialConditions/InitialConditions.jl
```diff
@@ -0,0 +1,15 @@
...
+
+using CLIMAParameters
+using ClimaCore: Geometry, Spaces, Fields
+
+include("init_1d_ekman_column.jl")
+include("init_2d_density_current.jl")
+include("init_2d_rising_bubble.jl")
+include("init_3d_rising_bubble.jl")
+include("init_3d_solid_body_rotation.jl")
+include("init_3d_baroclinic_wave.jl")
```
jiahe23, line 13: how about 3dbox and 3dsphere in the file name to differentiate. I guess no need for 1d and 2d because we only have the box

src/Utils/InitialConditions/init_2d_rising_bubble.jl
```diff
@@ -1,7 +1,7 @@
 """
-    init_dry_rising_bubble_2d(params, thermovar = :ρθ)
+    init_2d_rising_bubble(params, thermovar = :ρθ)
 
-    Dry rising bubble initial condition for 2D box benchmarking.
+    Rising bubble initial condition for 2D box benchmarking.
```
jiahe23, line 4: looks dry to me... unless we want to add the moist option in the same file?

src/Utils/InitialConditions/init_3d_rising_bubble.jl
```diff
@@ -1,7 +1,7 @@
 """
-    init_dry_rising_bubble_3d(params)
+    init_3d_rising_bubble(params)
 
-    Dry rising bubble initial condition for 3D box benchmarking.
+    Rising bubble initial condition for 3D box benchmarking.
```
jiahe23, line 4: same here ;)

test/runtests.jl
```diff
@@ -7,17 +7,17 @@ using OrdinaryDiffEq: SSPRK33, CallbackSet, DiscreteCallback
...
-using CLIMAParameters
+# using ClimaAtmos.Domains
+# using ClimaAtmos.BoundaryConditions
+# using ClimaAtmos.Models
+# using ClimaAtmos.Models.SingleColumnModels
+# using ClimaAtmos.Models.Nonhydrostatic2DModels
+# using ClimaAtmos.Models.Nonhydrostatic3DModels
+# using ClimaAtmos.Callbacks
+# using ClimaAtmos.Simulations
+# using CLIMAParameters
```
jiahe23, line 20: why commented?

test/test_cases/run_1d_ekman_column.jl
```diff
@@ -1,7 +1,14 @@
-include("initial_conditions/ekman_column_1d.jl")
+using CLIMAParameters
+using ClimaCore: Geometry
+using ClimaAtmos.Utils.InitialConditions: init_1d_ekman_column
+using ClimaAtmos.Domains
+using ClimaAtmos.BoundaryConditions
+using ClimaAtmos.Models
+using ClimaAtmos.Models.SingleColumnModels
+using ClimaAtmos.Callbacks
+using ClimaAtmos.Simulations
```
jiahe23, line 9: These packages used to be importing in runtest.jl and can only be imported once. Is there a reason why you want to move it to every test file?

jiahe23 (changes requested):
Good job! I like how the test_names are set-up. It makes it much simpler to run only a subset of the entire test suites. I left a few minor comments.

test/test_cases/run_1d_ekman_column.jl
bischtob: Yes, for clarity of the user.

test/runtests.jl
bischtob: whoops

src/Utils/InitialConditions/init_2d_rising_bubble.jl
bischtob: Yes, we will have initial conditions that have different modes -> dry, etc.

bischtob:
bors try

jiahe23 (approved):

PR #192, opened 2022-02-01
Adds hooks for proper validation/regression pipeline

bischtob:
This PR works on adding infrastructure and some refac to build towards a regression and validation pipeline similar to TurbulenceConvection.jl.

Goal is to get a more standardized view into the model output when making changes.


charleskawczynski:
I think that a cleaner way to do this is by doing what we do in ClimaCore -- use the existing test_cases/examples and exit once we have the integrator. Then we can perform all sorts of tests on that object in one separate smaller script. I can help set that up now that I'm a bit oriented with all of the examples

PR #194, opened 2022-02-02
Yc/mwe rico

yairchn:
Moving a minimal example Rico from TC.jl works. Here we will try to restructure it in line with the SingleColumnModel already defined and with the requirements of the EDMF model


bischtob:
@yairchn, is this PR still current or can it be closed?

PR #198, opened 2022-02-02
Add buildkite CI for cluster QA tests 

jakebolewski:


jakebolewski:
bors try

bischtob (approved):
seems like this is a go :). thanks for putting this up!

jakebolewski:
It would probably be good to have project only / test only deps in the respective `doc` / `test` subfolder projects

PR #202, opened 2022-02-04
Don't capture thermodynamic states

charleskawczynski:
Thermodynamic states are not currently designed to be captured (i.e., we should not make fields of them) as that will duplicate the parameter set.

We can change this by removing parameter sets from the thermodynamic states, but for now this is a better design (will be more performant) and doesn't seem problematic.


charleskawczynski:
bors try

bischtob (approved):
Thanks for capturing this!

PR #206, opened 2022-02-05
Restart Ekman from scratch

charleskawczynski:
As of commit `fbde6a58b7b877e040606458ec8d3599914a2cb8`, running the `run_1d_ekman_column.jl` is unstable for `dt = 2.0` (the timestep used in the ClimaCore example). While the solution is stable for `dt = 0.01`, the results allocate more than 1TB and takes over 15 min to run:

```julia
julia> using Revise; include("test/test_cases/run_1d_ekman_column.jl");

julia> @time run_1d_ekman_column(Float64; dt = 0.01, test_mode=:validation)
896.577026 seconds (10.56 G allocations: 1.550 TiB, 50.28% gc time, 0.04% compilation time)
```

There are clearly performance / allocation issues with the existing driver, so I think it's perhaps easiest for us to

 - start from scratch with the `run_1d_ekman_column.jl` (and use the ClimaCore example, which takes ~8 seconds in `OrdinaryDiffEq.solve`)
 - hook in the regression tests for the solution at `t_final`
 - start iterating on an interface (and try adopting some pieces from the existing interface)

This will also make it easy to preserve using the OrdinaryDiffEq `integrator` object, so that we can continuously monitor performance.


charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

PR #207, opened 2022-02-05
Remove unnecessary deps

charleskawczynski:
This may help with our CI init time a bit. We can add things back in as needed, but we should probably only add the things that we need for now.


charleskawczynski:
bors try

charleskawczynski:
bors try

PR #211, opened 2022-02-09
fix bug

jiahe23:
dss is applied to the incorrect fields in the moist equation. Current way of fix is only to be in an computationally efficient way and stays consistent with all other vars, but definitely not the most elegant way. @dennisYatunin  has better idea for his new interface.


bischtob (approved):
LGTM

jiahe23:
bors try

PR #222, opened 2022-02-18
import ClimaCorePlots

### LINKED ISSUE #221: Plots not working properly with ClimaCore obj
With a recent updates on ClimaCore library for plotting, we need to import ClimaCorePlots so that functions in Plots are properly extended for ClimaCore obj.

jiahe23:
make sure functions in the Plots package is properly extended for ClimaCore objects.
closes #221


jiahe23:
bors try

jiahe23:
> I think the reason that there are no artifacts may be that the test_mode is regression by default.

I was told we have the `:validation` in buildkite. I am not sure how to check that but I've open an issue on it #223

PR #225, opened 2022-02-19
First attempt to incorporate rhs functions with the `BoundaryConditions` module

jiahe23:


bischtob:
This will be postponed until we use the same vector type for all boundary fluxes in all models.

Issue #203, opened 2022-02-04
Cache mechanism


charleskawczynski:
We should probably have a caching mechanism for storing temporary fields, and store them in an auxiliary field. Perhaps borrowing our design in TC.jl would be helpful here (both for the aux and prognostic fields).

bischtob:
We may want to do something simple here for now. For the SingleColumnModel, we should just copy 1-to-1 how it's done in TC.jl.

PR #252, opened 2022-03-19
Initial working example with cache

trontrytel:
This is an initial working example with the cache in CliamAtmos. Thank you @dennisYatunin for your advice on how to hijack the ODE `p`.

So far I have `q_liq` and `q_ice` (the liquid and ice specific humidities) in the cache. Would be great to get comments on the proposed way of initialising the cache, before I move on and add other things there like pressure, temperature, etc. 

Also, @dennisYatunin already has an idea on how to do it better than hijacking `p`. But it will have to come in the next interface revamp.


src/Models/Nonhydrostatic2DModels/nonhydrostatic_2d_model.jl
```diff
@@ -26,11 +26,10 @@ Models.components(model::Nonhydrostatic2DModel) = (
     vertical_diffusion = model.vertical_diffusion,
 )
 
-function Models.default_initial_conditions(model::Nonhydrostatic2DModel)
+function Models.default_initial_conditions(model::Nonhydrostatic2DModel, space_center, space_face)
     # we need to provide default initial conditions for the model, because the ode solver
     # requires inital conditions when getting instantiated, but we also want to support the `set!` function
     # interface for initialization and re-initialization.
-    space_center, space_face = Domains.make_function_space(model.domain)
```
jiahe23, line 33: Is there a reason why you remove this line and make `space_center` and `space_face` input args? I prefer this old way because it looks simpler.

src/Models/Nonhydrostatic2DModels/nonhydrostatic_2d_model.jl
```diff
@@ -65,6 +64,19 @@ function Models.default_initial_conditions(model::Nonhydrostatic2DModel)
     return Fields.FieldVector(; zero_inits...)
 end
 
+function Models.default_ode_cache(model::Nonhydrostatic2DModel, space_center, space_face)
```
jiahe23, line 67: Same here. I prefer remove `space_center` and `space_face` from the inputs and get them inside the function.

src/Models/Nonhydrostatic3DModels/nonhydrostatic_3d_model.jl
```diff
@@ -27,11 +27,10 @@ function Models.components(model::Nonhydrostatic3DModel)
     )
 end
 
-function Models.default_initial_conditions(model::Nonhydrostatic3DModel)
+function Models.default_initial_conditions(model::Nonhydrostatic3DModel, space_center, space_face)
```
jiahe23, line 30: same...

src/Models/Nonhydrostatic2DModels/nonhydrostatic_2d_model.jl
```diff
@@ -65,6 +64,19 @@ function Models.default_initial_conditions(model::Nonhydrostatic2DModel)
...
+
+    local_geometry_center = Fields.local_geometry_field(space_center)
+
+    FT = eltype(model.domain)
+    zero_instance = zero(FT) # somehow need this, otherwise eltype inference error
+
+    q_liq = map(coord -> zero_instance, local_geometry_center)
+    q_ice = map(coord -> zero_instance, local_geometry_center)
```
jiahe23, line 75: We could initialize cache names somewhere as (similar to initilize prog varnames) and loop over here as in the default init condition function.

src/Models/Nonhydrostatic2DModels/nonhydrostatic_2d_model.jl
trontrytel: Yes. They can only be initialised once per simulation. Otherwise you get an error `Broacasted spaces are the same ClimaCore.Spaces type but not the same instance` that comes from `ClimaCore` `Fields/broadcast.jl` file. 

I think it makes sense that you would want to only create one instance of space per simulation. Otherwise it would be hard to guarantee that they are the same between different variables. 

One option would be to initialise both the state and cache in one function. Then they could be hidden back inside?

src/Models/Nonhydrostatic2DModels/nonhydrostatic_2d_model.jl
trontrytel: That sounds good. I'll see if I can do this.

trontrytel:
bors try

trontrytel:
I have updated the PR with cache computations needed for the 2D nonhydrostatic model coupled to 0-moment microphysics. There are some changes to where some functions used to live. 

My main goal was to keep the allocations and the number of cached values to the minimum. At the same time I wanted to keep some intermediate computation steps written out, for easier code readability by humans. 

Not all the places are elegant. At one point I have to pass in the `$Ref(params)` as one of the arguments...

Let me know what you Folks think and whether I'm even going in the right direction...

trontrytel:
bors try

trontrytel:
bors try

trontrytel:
bors try

trontrytel:
I think this is how far as I can push it. I decided to keep the cache initialisation more spelled out for now, because I think we are still not decided what should be a default choice. So maybe this way things will be easier to change. 

Let me know if anything else needs to be changed. If not, I'm happy to move to adding the 1-moment microphysics + it's cache. 

@LenkaNovak - would this PR also work for what you need?

src/Models/style_cache.jl
```diff
@@ -0,0 +1,9 @@
+
+"""
+Supertype for all model cache styles.
+"""
+abstract type AbstractCacheStyle <: AbstractModelStyle end
+
+struct CacheEmpty <: AbstractCacheStyle end
+struct CachePressure <: AbstractCacheStyle end
+struct CacheZeroMomentMicro <: AbstractCacheStyle end
```
jiahe23, line 9: Can we do something similar as the syle_base, e.g., adding something like
```
Models.cache_variable_names(::CacheEmpty) = Nothing
Models.cache_variable_names(::CachePressure) = (:p)
Models.cache_variable_names(::CacheZeroMomentMicro) = (:p, :Φ, :K, P, :e_int, :microphysics_cache)
```
and the corresponding `variable_types` and `variable_spaces` functions.

By doing this, the `default_ode_cache` does not need to be dispatched on different cache styles, but it can automatically extracts the cache varnames and their vartype/spaces, and then loop over, just as done [here](https://github.com/CliMA/ClimaAtmos.jl/blob/main/src/Models/Nonhydrostatic3DModels/nonhydrostatic_3d_model.jl#L34-L41) in the default_init_conditions.

jiahe23: This way, when someone needs to add his/her own cache setups, he/she only needs to change the style file and does not have to touch the model file.

trontrytel: The `microphysics_cache` is a tuple of variables. I did not know how to loop over it in the way the stat vars are initialised. Can you help me with this?

trontrytel:
bors try

src/Models/Nonhydrostatic2DModels/nonhydrostatic_2d_model.jl
bischtob: I think we'd need to remove and handle the function space creation somewhere else in the end, but we could break it down like you are proposing here :)

src/Models/Nonhydrostatic2DModels/nonhydrostatic_2d_model.jl
bischtob: Would it be possible to assemble the cache using also BaseModelStyles and ThermoStyles? For example a base mode even without other equations added needs pressure in the cache.

src/Models/style_cache.jl
bischtob: Could we dispatch on ModelStyles here? Like BaseModelStyles to get the cache variable names?

src/Models/Models.jl
```diff
@@ -11,12 +11,14 @@ export AbstractModel,
...
     default_initial_conditions,
-    make_ode_function
+    make_ode_function,
+    default_ode_cache
 
 export AbstractModelStyle,
     AbstractBaseModelStyle,
     AbstractThermodynamicsStyle,
     AbstractMoistureStyle,
+    AbstractCacheStyle,
```
bischtob, line 21: Can we reuse the ModelStyles?

src/Models/Models.jl
```diff
@@ -89,13 +95,25 @@ variable_spaces(style::AbstractModelStyle) =
...
     error("default_initial_conditions not implemented for given model type")
 
+"""
+    default_ode_cache(model, cache, space_center, space_face)
+
+Construct the cache for ordinary differential equations.
+"""
+default_ode_cache(
+    model::AbstractModel,
+    cache::AbstractCacheStyle,
```
bischtob, line 112: What if we dispatched somehow on the different model styles and components?

bischtob (commented):
I am wondering if there is a possibility of reuse of the different modelstyles for moisture, base model, thermodynamics, etc. By introducing another model style for microphysics, we could handle the microphysics cache?

trontrytel:
bors try

trontrytel:
I applied the cosmetic changes we discussed. 

I'll do the `Ya` -> `cache` in a separate PR, because it will touch a lot of files. 

lmk if you see anything else that need to be fixed before merging it

jiahe23:
closes #203

jiahe23 (approved):

PR #253, opened 2022-03-22
Fix bc bug in vertical diff in 3d model and run with bw

jiahe23:


jiahe23:
bors try

LenkaNovak (approved):
Excellent! Thank you! :)

PR #254, opened 2022-03-23
Add png plotting callback

trontrytel:
This PR adds a callback that outputs the simulation results into a png file. 

I think this could be very useful for debugging. But as a downside it means adding Plots to our deps.

### DIFF
--- a/test/test_cases/run_2d_moist_bubble.jl
+++ b/test/test_cases/run_2d_moist_bubble.jl
@@ -7,7 +7,7 @@ using Test
 using OrdinaryDiffEq: SSPRK33, CallbackSet
 using ClimaCorePlots, Plots
 using UnPack
-
+using DiffEqCallbacks
 
 using CLIMAParameters
 using ClimaAtmos.Utils.InitialConditions: init_2d_moist_bubble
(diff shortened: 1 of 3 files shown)


test/test_cases/run_2d_moist_bubble.jl
```diff
@@ -125,7 +125,18 @@ function run_2d_moist_bubble(
         # mkpath(path)
         # cb_jld2 = JLD2Output(model, path, "moist_bubble", dt)
         # cb_set = CallbackSet(generate_callback(cb_jld2))
-        simulation = Simulation(model, stepper, dt = dt, tspan = (0.0, 500.0))#, callbacks = cb_set,)
+
+        path = ""
```
jiahe23, line 129: maybe something like
```
        path = joinpath(@__DIR__, first(split(basename(@__FILE__), ".jl")))
        mkpath(path)
```
so that if people fail to specify a dir, it has a default.

trontrytel:
bors try

Project.toml
```diff
@@ -14,6 +14,7 @@ JLD2 = "033835bb-8acc-5ee8-8aae-3f567f8a3819"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 OrdinaryDiffEq = "1dea7af3-3e70-54e6-95c3-0bf5283fa5ed"
 Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
+Plots = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"
```
charleskawczynski, line 17: Can we keep Plots outside of the default environment?

trontrytel: Not if we want a callback that plots things?

trontrytel: I'm ok to delete the callback and the dependency later. But right now there aren't that many options for plotting state and cache variables while simulation is running. 

I'm thinking of it as a temporary addition for debugging. Happy to delete it once we have a better output options?

jiahe23 (approved):
lgtm

Project.toml
jiahe23: @charleskawczynski  Do you want to remove Plots from any Project.toml or just the one under the home folder of this repo? For example, if we have something in Utils, is it acceptable to have Plots inside the Project there?

charleskawczynski: Just in the package environment. I'd suggest looking at ClimaCore to see how plotting is supported (i.e., in example environments) without making the package directly depend on Plots (see ClimaCore's deps [here](https://github.com/CliMA/ClimaCore.jl/blob/main/Project.toml#L25-L44)).

glwagner: Maybe worth mentioning that `ClimaCore` also has "libraries" that extent functionality, but remain external to the source code:

https://github.com/CliMA/ClimaCore.jl/tree/main/lib

A callback that makes a plot should be simple to implement though so you might not need a new external library for that. But it might be worthwhile adding a utility that makes it a little easier (ie so that users don't have to define a `struct` to implement a callback)?

trontrytel:
After the sudden but inevitable pushback against Plots in the main deps, I moved the callback definition into tests. 

Let me know if that would be ok for now @charleskawczynski @glwagner

trontrytel:
bors try

charleskawczynski (approved):

PR #257, opened 2022-03-25
bring back external boundary conditons

LenkaNovak:
This reintroduces `boundary_conditions`, which can be specified outside of the `rhs!` from the driver file, essential for coupled simulations. Fluxes are applied via the `vertical_diffusion` rhs!, so this needs to be enabled in `model`.

This PR is primarily for prototyping interfaces for coupling, and only includes:
- total energy (`BulkFormula`) and horizontal momentum (`DragLaw`), following chapter 5 of the DesignDoc
- spatially constant boundary fluxes

Follow-up PRs:
- extend to use spatially variable boundary fluxes (ClimaCore PR #325)
- extend to other thermodynamic variables in the `vertical_diffusion` rhs! (pending upcoming interface changes)
- add moisture 
- use `SurfaceFluxes.jl` and apply in the form of sensible/latent heat flux

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -22,7 +22,7 @@ UnPack = "3a884ed6-31ef-47d7-9d2a-63182c4928ed"
 
 [compat]
 CLIMAParameters = "0.4"
-ClimaCore = "0.9"
+ClimaCore = "0.9, 0.10"
 ClimaCorePlots = "0.2"
 ClimaCoreVTK = "0.6, 0.7"
 CloudMicrophysics = "0.3"
--- a/src/BoundaryConditions/BoundaryConditions.jl
+++ b/src/BoundaryConditions/BoundaryConditions.jl
@@ -5,11 +5,8 @@ using ClimaCore: Geometry, Fields, Operators
 using ClimaCore.Geometry: ⊗
 
 export get_boundary_flux,
-    AbstractBoundaryCondition,
-    CustomFluxCondition,
-    NoFluxCondition,
-    DragLawCondition,
-    BulkFormulaCondition
+    AbstractBoundary, NoFlux, NoVectorFlux, DragLaw, BulkFormula
+
 
 """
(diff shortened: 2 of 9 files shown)


LenkaNovak:
Actually hold on with reviewing for a few hours. I'll try incorporate `SurfacesFluxes.jl`, which is more general.

jiahe23:
> Actually hold on with reviewing for a few hours. I'll try incorporate `SurfacesFluxes.jl`, which is more general.

If integrating with `SurfaceFluxes.jl` is too much work, we could merge this in first and open a separate PR, so that this one does not grow too big and get stalled.

LenkaNovak:
Thanks @jiahe23 That's what I concluded as well. The `SurfaceFluxes.jl` integration is now in [this branch](https://github.com/CliMA/ClimaAtmos.jl/tree/ln/use_SurfaceFluxes) and it will most likely require Charlie's input. So is this PR good for merging?

LenkaNovak:
Good point, @jiahe23 . @akshaysridhar, as the flux vector pro, do you think we should use `Geometry.WVector` instead here?

jiahe23:
> Good point, @jiahe23 . @akshaysridhar, as the flux vector pro, do you think we should use `Geometry.WVector` instead here?

And then I have a followupo question:  Since the equation is set up using the Covariant/Contravariant system, if we use `WVector` and passed into the bc setups in operators, will it automatically convert it to the `Contravariant3Vector` with correct length?

LenkaNovak:
@jiahe23 Please can we merge at least these stubs, so we can still use this old interface to finish off some of our existing test cases in the coupler that depend on this interface? Thanks! 😊

jiahe23 (approved):
Yes, Let's have this in asap. It can also serve as a working example to refer to when we add similar functionality to current test cases under `examples/`

jiahe23:
bors try

PR #258, opened 2022-03-26
Enable passing Ya into`simulation`

LenkaNovak:
- for coupling ClimaAtmos models, it is essential to exchange auxiliary information (such as states of the neighboring models). This will be done via the existing `Ya` variable in ClimaAtmos (which is linked to `integrator.p` in the solution of `OrdinaryDiffEq`). 
- this PR allows an optional specification of `Ya` outside of `simulation`

### DIFF
--- a/src/Simulations/simulation.jl
+++ b/src/Simulations/simulation.jl
@@ -14,11 +14,13 @@ end
         tspan,
         callbacks = nothing,
         restart = NoRestart(),
+        Ya = nothing,
     )
 
 Construct a `Simulation` for a `model` with a time stepping `method`,
 initial conditions `Y_init`, and time step `dt` for a time interval of `tspan`.
 If `Y_init` is not provided, the model's default initial conditions are used.
+Auxiliary information can be additionally provided via `Ya`. 
 ```
 """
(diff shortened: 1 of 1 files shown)


trontrytel:
@LenkaNovak - I think I'm doing the same thing here: https://github.com/CliMA/ClimaAtmos.jl/pull/252

I just need to wrap up the initial condition for the cache. It should be done very soon

jiahe23:
@LenkaNovak @trontrytel  It looks like #252 made it possible already? Could you double check and close this PR if possible?

LenkaNovak:
I've just realized that #252 does not allow for specifying arbitrary `Ya` outside of `Simulation`. We will most likely need to add this functionality in the future, but we don't necessarily need it right now, so I'm happy to close this PR.

jiahe23:
> I've just realized that #252 does not allow for specifying arbitrary `Ya` outside of `Simulation`. We will most likely need to add this functionality in the future, but we don't necessarily need it right now, so I'm happy to close this PR.

Sounds good! Let's have a meeting and clarify what is needed.

PR #275, opened 2022-04-05
add hs forcing

### LINKED ISSUE #271: Add dry held suarez example

jiahe23:
Co-authored-by: Akshay Sridhar <akshaysridhar@users.noreply.github.com>
Co-authored-by: Zhaoyi Shen <szy21@users.noreply.github.com>
Closes #271

### DIFF
--- a/src/Utils/Forcings/forcing_held_suarez.jl
+++ b/src/Utils/Forcings/forcing_held_suarez.jl
@@ -3,26 +3,6 @@ struct HeldSuarezForcing{S} <: AbstractAtmosForcing
     parameters::S
 end
 
-## this parameters should goes into runscript
-# model = Nonhydrostatic3DModel(......; source = (; forcing = HeldSuarezForcing(held_suarez_parameters)) )
-# day = FT(86400)
-# held_suarez_parameters = (;
-#     k_a = FT(1 / (40 * day)),
-#     k_f = FT(1 / day),
-#     k_s = FT(1 / (4 * day)),
-#     ΔT_y = FT(60),
-#     Δθ_z = FT(10),
-#     T_equator = FT(315),
-#     T_min = FT(200),
-#     σ_b = FT(7 / 10),
-#     R_d  = parameters.R_d,
-#     day  = parameters.day,
-#     grav = parameters.g,
-#     cp_d = parameters.cp_d,
-#     cv_d = parameters.cv_d,
-#     MSLP = parameters.p0,
-# )
-
 function rhs_forcing!(dY, Y, Ya, p, hs::HeldSuarezForcing, ::TotalEnergy)
     # # unpack cache 
     # TODO: make cache capability work for 3D model
@@ -38,7 +18,7 @@ function rhs_forcing!(dY, Y, Ya, p, hs::HeldSuarezForcing, ::TotalEnergy)
 
     # obtain latitude
     c_local_geometry = Fields.local_geometry_field(Y.base.ρ)
-    φ = deg2rad.(c_local_geometry.coordinates.lat),
+    φ = deg2rad.(c_local_geometry.coordinates.lat)
 
     # forcing
     σ = p / hs.parameters.p0
@@ -73,7 +53,7 @@ function rhs_forcing!(dY, Y, Ya, p, hs::HeldSuarezForcing, ::PotentialTemperatur
 
     # obtain latitude
     c_local_geometry = Fields.local_geometry_field(Y.base.ρ)
-    φ = deg2rad.(c_local_geometry.coordinates.lat),
+    φ = deg2rad.(c_local_geometry.coordinates.lat)
 
     # forcing
     σ = p / hs.parameters.p0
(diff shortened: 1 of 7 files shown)


src/Utils/Forcings/forcing_held_suarez.jl
```diff
@@ -0,0 +1,94 @@
...
+    R_d = CLIMAParameters.Planet.R_d(hs.parameters)
+    cp_d = CLIMAParameters.Planet.cp_d(hs.parameters)  
+    cv_d = CLIMAParameters.Planet.cv_d(hs.parameters)
+    p_0 = CLIMAParameters.Planet.MSLP(hs.parameters)
+
+    # obtain latitude
+    c_local_geometry = Fields.local_geometry_field(Y.base.ρ)
+    φ = deg2rad.(c_local_geometry.coordinates.lat),
+
+    # forcing
```
tapios, line 43: The temperature relaxation that follows here should be put into a separate function, so it can be re-used for the different thermodynamic variables, rather than duplicating code.

src/Utils/Forcings/forcing_held_suarez.jl
```diff
@@ -0,0 +1,94 @@
...
+    σ = p / hs.parameters.p0
+    height_factor = max(0, (σ - σ_b) / (1 - σ_b))
+    ΔρT =
+        (k_a + (k_s - k_a) * height_factor * cos(φ)^4) *
+        Y.base.ρ *
```
tapios, line 48: I think it's clearer if this is defined as `dT_dt`, without the density factor, and then the density factor is inserted in the tendency if needed for the thermodynamic variable under consideration.

src/Utils/Forcings/forcing_held_suarez.jl
```diff
@@ -0,0 +1,94 @@
...
+        ( # ᶜT - ᶜT_equil
+            p / (Y.base.ρ * R_d) - max(
```
tapios, line 50: It would be good to define a temperature here. We'll need it in many places in any actual atmosphere model, so it should be cached in some auxiliary state and computed from thermodynamics functions. Given that we'll need that anyway, I suggest we make this change now, for better readability and generalizability.

src/Utils/Forcings/forcing_held_suarez.jl
```diff
@@ -0,0 +1,94 @@
...
+    R_d = CLIMAParameters.Planet.R_d(hs.parameters)
+    cp_d = CLIMAParameters.Planet.cp_d(hs.parameters)  
+    cv_d = CLIMAParameters.Planet.cv_d(hs.parameters)
+    p_0 = CLIMAParameters.Planet.MSLP(hs.parameters)
+
+    # obtain latitude
+    c_local_geometry = Fields.local_geometry_field(Y.base.ρ)
+    φ = deg2rad.(c_local_geometry.coordinates.lat),
+
+    # forcing
```
tapios, line 78: Make separate function, so this does not need to be duplicated.

src/Utils/Forcings/forcing_held_suarez.jl
```diff
@@ -0,0 +1,94 @@
...
+        Y.base.ρ *
+        ( # ᶜT - ᶜT_equil
+            p / (Y.base.ρ * R_d) - max(
+                T_min,
+                (T_equator - ΔT_y * sin(φ)^2 - Δθ_z * log(σ) * cos(φ)^2) *
+                σ^(R_d / cp_d),
+            )
+        )
+
+    dY.base.uh -= (k_f * height_factor) * Y.base.uh
```
tapios, line 92: Also good to put surface friction into a separate function, so code does not need to be duplicated.

test/test_cases/run_3d_held_suarez.jl
```diff
@@ -0,0 +1,90 @@
...
+using CLIMAParameters
+using ClimaAtmos.Utils.InitialConditions: init_3d_baroclinic_wave
+using ClimaAtmos.Domains
+using ClimaAtmos.BoundaryConditions
+using ClimaAtmos.Models: ConstantViscosity
+using ClimaAtmos.Models.Nonhydrostatic3DModels
+using ClimaAtmos.Simulations
+
+# Set up parameters
+const FT = Float64
```
tapios, line 14: We should run Float32 by default if we can.

test/test_cases/run_3d_held_suarez.jl
```diff
@@ -0,0 +1,90 @@
...
+const day = FT(86400)
```
tapios, line 15: Should be global parameter.

test/test_cases/run_3d_held_suarez.jl
```diff
@@ -0,0 +1,90 @@
...
+        nelements = nelements,
+        npolynomial = npolynomial,
+    )
+
+    model = Nonhydrostatic3DModel(
+        domain = domain,
+        boundary_conditions = nothing,
+        parameters = params,
+        hyperdiffusivity = FT(1e16),
+        vertical_diffusion = ConstantViscosity(ν = FT(1)),
```
tapios, line 58: Have we been running this vertical diffusion? We shouldn't. (1 m2/s is quite strong.)

szy21: No, we didn't use it.

PR #283, opened 2022-04-08
Call RRTMGP to compute radiative flux in TC.jl

yairchn:
first -- quick and dirty -- version that calls RRTMGP to compute radiative flux (needs deving to a branch in which CLIMAParmaeters are removed from `Project.toml` in the meanwhile).
 @dennisYatunin will make changes in RRTMGP


bischtob:
I think we should use better PR names :). I renamed this one.

PR #284, opened 2022-04-11
Use `SurfaceFluxes.jl`

LenkaNovak:
Demonstrates how the current implementation of `SurfaceFluxes.jl` can be used in ClimaAtmos (and thus other component models). 

Follow-up TODOs
- spatially variable BCs
- better interface for `map_fluxes.` and point-wise iteration over `Field` objects


LenkaNovak:
@akshaysridhar is assuming that that surface density is the same as that at the first interior node ok [here](https://github.com/CliMA/ClimaAtmos.jl/blob/ln/use_SurfaceFluxes/src/BoundaryConditions/flux_calculations.jl#L91)? Or is extrapolation of rho or ideal gas calculation based on `MSLP = 1e5` Pa better?

LenkaNovak:
@charleskawczynski @yairchn copying you in, in case you have a more sophisticated way of using `SurfaceFluxes.jl` within `ClimaCore`-based cases. 🤓

src/BoundaryConditions/flux_calculations.jl
```diff
@@ -1,61 +1,130 @@
...
+        model.moisture,
+        model.parameters,
+        FT,
+    )
+
+    qt_sfc = FT(0)
+    cv_d = TD.cv_m(model.parameters, FT)
+
+    # internal energies
+    e_int_in = Fields.level(TD.air_temperature_from_ideal_gas_law.(model.parameters, p, Y.base.ρ) .* cv_d, 1)
```
yairchn, line 74: is there a reason not to use TD. internal_energy(ts) here? 
https://github.com/CliMA/Thermodynamics.jl/blob/119e7271497b06cdeffaf86112c475c76ec67e5b/src/relations.jl#L377

src/BoundaryConditions/flux_calculations.jl
```diff
@@ -1,61 +1,130 @@
...
+    # collect states
+    state_sfc = SF.SurfaceValues(z_sfc, uh_sfc, ts_sfc) 
+    state_in = SF.InteriorValues(z_in, uh_in, ts_in)
+
+    # collect all surface conditions
+    kwargs = (; state_in, state_sfc)
+    sc = SF.Coefficients{FT}.(; kwargs..., Cd = bc.coefficients.Cd, Ch = bc.coefficients.Ch, z0m = FT(0), z0b = FT(0))
+
+    uf = UF.Businger()
+    bulk_flux = SF.sensible_heat_flux(model.parameters, sc.Ch, sc, uf)
```
yairchn, line 120: The idea behind SF.jl is not to use it's internal functions such as `SF.sensible_heat_flux`,  but to call the user facing function `surface_condition`, i.e.:

`result = SF.surface_conditions(param_set, sc, universal_func, scheme)`

and get the fluxes from:

`bulk_flux = results.shf`

and also:
`bulk_lhf_flux = results.lhf`


There are many reason to do that
1. the interface will not change with changing in SF
2. you get other properties (obukhuv length, similarity profiles and such at the same time)
3. the user does need to "know" the theory to used it

christophernhill (commented):
@LenkaNovak @charleskawczynski and @akshaysridhar is there a radiative flux part to SurfaceFluxes somewhere or is that elsewhere? I couldn't find anything that deals with short-wave and long-wave upward and downward and associated heating etc... ? 

e.g. Qin, Qr and Qb in https://swfsc-publications.fisheries.noaa.gov/publications/CR/1974/7406.PDF  for example.

src/BoundaryConditions/flux_calculations.jl
```diff
@@ -1,61 +1,130 @@
...
+        Geometry.Covariant12Vector(FT(0), FT(0))
+end
 
-    uv_1 = first_interior(uv)
+"""
+    get_boundary_flux(model, bc::DragLaw, uv, Y, Ya)
+
+    Vertical momentum fluxes for a bulk-formulation drag law, given the momentum exchange coefficient
+"""
+function get_boundary_flux(model, bc::DragLaw, uv, Y, Ya)
```
yairchn, line 28: I think that it would be better to use SF.jl for. compute any flux, if something is missing there we should add it there. The current DragLaw is given by SF.jl:

https://github.com/CliMA/SurfaceFluxes.jl/blob/035a833f59a90d889eadef837491da5b8fa9908e/src/SurfaceFluxes.jl#L589

LenkaNovak:
> @LenkaNovak @charleskawczynski and @akshaysridhar is there a radiative flux part to SurfaceFluxes somewhere or is that elsewhere? I couldn't find anything that deals with short-wave and long-wave upward and downward and associated heating etc... ?
> 
> e.g. Qio, Qr and Qb in https://swfsc-publications.fisheries.noaa.gov/publications/CR/1974/7406.PDF for example.

@christophernhill ,so far we have been dealing with radiative surface fluxes (of a very simple form) within the coupler. @dennisYatunin , what is the plan for when the radiation and the associated interface are fully implemented?

charleskawczynski:
> @LenkaNovak @charleskawczynski and @akshaysridhar is there a radiative flux part to SurfaceFluxes somewhere or is that elsewhere? I couldn't find anything that deals with short-wave and long-wave upward and downward and associated heating etc... ?
> 
> e.g. Qio, Qr and Qb in https://swfsc-publications.fisheries.noaa.gov/publications/CR/1974/7406.PDF for example.

Great question. Not that I know of, but I think it should be extensible as far as I understand. Perhaps @yairchn knows more about where things pop up in the equations.

christophernhill:
> > @LenkaNovak @charleskawczynski and @akshaysridhar is there a radiative flux part to SurfaceFluxes somewhere or is that elsewhere? I couldn't find anything that deals with short-wave and long-wave upward and downward and associated heating etc... ?
> > e.g. Qio, Qr and Qb in https://swfsc-publications.fisheries.noaa.gov/publications/CR/1974/7406.PDF for example.
> 
> Great question. Not that I know of, but I think it should be extensible as far as I understand. Perhaps @yairchn knows more about where things pop up in the equations.

@charleskawczynski I think we are hoping to talk with @dennisYatunin  on Monday about this stuff.

christophernhill:
oops wrong button!

LenkaNovak:
Carries over into #324 using the new interface

Issue #272, opened 2022-04-05
Add regression tests


szy21:
Depends on #269 #270 #271

charleskawczynski:
Closed by #312 and #321.

PR #285, opened 2022-04-11
Add and use cli options

charleskawczynski:
A step towards #272.


bischtob (approved):
LGTM

charleskawczynski:
I think I had one mistake, the mpi run was running on Float32 (and maybe that's why it's taking so long?). I'll fix now.

PR #286, opened 2022-04-11
Add long run job to buildkite

charleskawczynski:
This PR adds a new pipeline file: `.buildkite/longruns/pipeline.yml`, which runs long-run Held-Suarez simulations. I've added the ClimaAtmos-LongRuns pipeline which monitors this new pipeline yaml, where I've set up a scheduled builds (`@midnight` to start with), we can change to `@weekly` if it seems unnecessary.


charleskawczynski:
It may be that we need / should add other CLI options to adjust the output frequency

PR #293, opened 2022-04-13
Switch on 0-moment precip to moist bw

jiahe23:


szy21 (approved):

jiahe23:
bors try

trontrytel:
Is the theta in this example the dry potential temperature or liquid ice potential temperature? The source terms are written assuming it's the liquid ice potential temperature.  (Same goes for the previous PR with 0-moment micro)

trontrytel (approved):

jiahe23:
> Is the theta in this example the dry potential temperature or liquid ice potential temperature? The source terms are written assuming it's the liquid ice potential temperature. (Same goes for the previous PR with 0-moment micro)

It is the liquid ice potential temperature as initialized [here](https://github.com/CliMA/ClimaAtmos.jl/blob/main/examples/hybrid/sphere/baroclinic_wave_utilities.jl#L117-L118)

trontrytel:
> > Is the theta in this example the dry potential temperature or liquid ice potential temperature? The source terms are written assuming it's the liquid ice potential temperature. (Same goes for the previous PR with 0-moment micro)
> 
> It is the liquid ice potential temperature as initialized [here](https://github.com/CliMA/ClimaAtmos.jl/blob/main/examples/hybrid/sphere/baroclinic_wave_utilities.jl#L117-L118)

👍  Sounds good. In case we also want to support the dry theta it will be some more work on the source terms.

PR #294, opened 2022-04-13
Update pipeline, use output dir from cli options

charleskawczynski:
This PR rearranges the pipeline by
 - grouping the examples a bit differently and
 - changing the output folders based on the cli options (this can be retrieved by `julia .dev/pipeline_examples_generator.jl`).

This is a sub-step towards adding regression tests since we need to have a job ID for the regression tables.

We should review the buildkite results to make sure that everything is still showing up correctly.


bischtob:
LGTM

bischtob (approved):
LGTM

PR #307, opened 2022-04-15
Update examples for dss ghost buffer changes

simonbyrne:
Transferred from
https://github.com/CliMA/ClimaCore.jl/pull/631

Will tag a new release of ClimaCore soon. Note that this will break any jld2 files.


simonbyrne:
bors try

simonbyrne:
bors try

simonbyrne:
bors try

simonbyrne:
bors try

simonbyrne:
bors try

simonbyrne:
It looks like there might be some changes in the regression tables: I'm not sure what these numbers mean to know if they're important.

charleskawczynski:
> It looks like there might be some changes in the regression tables: I'm not sure what these numbers mean to know if they're important.

I'm seeing MSE changes in some other PRs (e.g., #359) that appear to be non-behavior changing. That said, here's a gif between the results of baroclinic wave (ρe) on the merged build 372 vs this PR:

PRVsMain

I'm starting to suspect that we're not fully reproduce-able. Maybe there's some randomness? Or maybe we just need to relax our tolerance a bit?

I think it's fine to merge this, and probably #359, too. For now we can just update the mse tables, but it might be worth looking into what's going on here.

charleskawczynski:
Docs for merging [here](https://github.com/CliMA/ClimaAtmos.jl/blob/main/post_processing/README.md#how-to-merge-pull-requests-pr-that-get-approved-but-fail-regression-tests)

tapios:
@charleskawczynski Can we please change the regression tests to be physically meaningful? We are dealing with a chaotic system that, in detail, is sensitive to small changes (e.g., changes in numerics affecting roundoff, or changes between single and double precision). We want meaningful regression tests that do not fail as a result of such small changes. For the baroclinic wave, it would be meaningful to check, for example, that minimum and maximum surface pressure after some time are within some bounds. Same for min and max vorticity in the upper troposphere (~8 km altitude). It's also fine to check some MSE's, but then make the tolerances generous so that the tests pass, e.g., with single and double precision without changes to numerical values. 

Also, please use physical MSE where you use them. The numbers in the table do not seem to be a *mean* squared error, but perhaps just squared error? (Values like ~400 [m/s?] do not make sense as a velocity error.)

I'm concerned that if we have tests with stringent numerical values that are not immediately interpretable, it will tempt people just to keep changing the tests when they fail, defeating the purpose of having them.

charleskawczynski:
> @charleskawczynski Can we please change the regression tests to be physically meaningful? ... For the baroclinic wave, it would be meaningful to check, for example, that minimum and maximum surface pressure after some time are within some bounds. Same for min and max vorticity in the upper troposphere (~8 km altitude).

We can certainly add physically meaningful tests, but I think it would make sense to separate these from the regression tests.

> We are dealing with a chaotic system that, in detail, is sensitive to small changes (e.g., changes in numerics affecting roundoff, or changes between single and double precision). We want meaningful regression tests that do not fail as a result of such small changes. It's also fine to check some MSE's, but then make the tolerances generous so that the tests pass, e.g., with single and double precision without changes to numerical values.

I agree. We only recently got these tests running and I think we may need to increase the tolerance, and perhaps other minor adjustments, to get things running more smoothly.

> Also, please use physical MSE where you use them. The numbers in the table do not seem to be a _mean_ squared error, but perhaps just squared error? (Values like ~400 [m/s?] do not make sense as a velocity error.)

That's a good point. I can take a look at how we do normalization to see if we can improve that for physical interpretability.

> I'm concerned that if we have tests with stringent numerical values that are not immediately interpretable, it will tempt people just to keep changing the tests when they fail, defeating the purpose of having them.

The way we've been using the regression tests in TC.jl is just to provide an indicator of when there are behavioral changes-- we update them freely so long quality of the results are retained. I don't think this has been a blocker / too much of a slowdown, maybe @yairchn, @ilopezgp, @trontrytel can chime in here?. The main advantage I see from having these tests is that we can more freely/confidently refactor with these tests in place.

Overall, I'm hoping/betting that relaxing the tolerance will make things run more smoothly.

szy21:
> Note that this will break any jld2 files.

Is this still the case? And is this only for running with mpi or for single process as well?

simonbyrne:
Yes, it changes some internals which will mean it won't be able to reconstitute the types

szy21:
If so, could we please fix this, at least for single process? Not having outputs basically makes debugging impossible.

yairchn:
@charleskawczynski If I understand correctly here in the baroclinic wave you using snapshots in time for regression tests, while in the EDMF we are using time averaged profiles. This could be the reason why we could use regression tests that compare MSE between fields in EDMF, which probably as chaotic as this global test. 
This baroclinic wave test is very transient in nature so I don't think there is much sense in looking at time averaged fields. MAYBE zonal mean fields would make sense (@tapios, @szy21 ?). If not then I think the best thing to do is what @tapios suggested (max, min of selected values) and user eye-inspection to make sure the fields look reasonable before merging.

simonbyrne:
> If so, could we please fix this, at least for single process? Not having outputs basically makes debugging impossible.

We have outputs, you just won't be able to (easily) load values saved with an earlier version.

szy21:
Oh ok, no problem then. Thanks!

charleskawczynski:
> If I understand correctly here in the baroclinic wave you are using snapshots in time for regression tests, while in the EDMF we are using time averaged profiles. This could be the reason why we could use regression tests that compare MSE between fields in EDMF, which probably as chaotic as this global test....

We compare the solution at every saved point in time, so it's even more stringent than what we do here (only at the last timestep).

tapios:
@charleskawczynski We want regression tests precisely to be able to move forward with the confidence of not breaking upstream pieces of code--without getting bogged down by pegging us to incomplete code. Rregression tests are usually employed after some piece of code is complete and functionally tested. This is not the situation we have here, or will have in a more fully developed climate model any time soon. We want tests that (a) allow us to move forward with confidence of not having broken upstream pieces but (b) do not peg us to incomplete upstream code. 

The current test pegs us to code that is not complete. This type of situation has bogged us down in the past; we want to avoid it now.

As a rule, write regression tests that 
* test what matters about the output of code
* employ physically meaningful tolerances
* are interpretable (and hence are documented, including documentation of what motivates tolerances). 

We should avoid a situation in which numbers appear in a test that *no one* understands even now (much less a year from now). For example, for the baroclinic wave, we want tests to pass even when swapping out one correct time stepper for another, despite the small differences that inevitably introduces.

So, please rewrite the regression test to address this. For the baroclinic wave, you can use some of the flow statistics (e.g., max/min surface pressure, evolution of max surface pressure over time) discussed here (https://www.cesm.ucar.edu/events/wg-meetings/2018/presentations/amwg/jablonowski.pdf). @szy21, @jiahe23 and @bischtob can help with conceptual suggestions.

charleskawczynski:
Let's move the regression discussion into an issue, since it's outside the scope of this PR. I've opened #377 to continue this discussion.

examples/hybrid/driver.jl
```diff
@@ -206,7 +192,11 @@ end
 dss_func = make_dss_func(comms_ctx)
 save_to_disk_func = make_save_to_disk_func(output_dir, is_distributed)
 
-dss_callback = FunctionCallingCallback(dss_func, func_start = true)
+dss_callback = FunctionCallingCallback(func_start = true) do Y, t, integrator
+    p = integrator.p
+    Spaces.weighted_dss!(Y.c, p.ghost_buffer.c)
+    Spaces.weighted_dss!(Y.f, p.ghost_buffer.f)
+end
```
charleskawczynski, line 199: Why are we not using `dss_func` anymore (returned from `make_dss_func`)? Should we delete that if it's no longer used?

szy21:
Oh, one more thing: Have we fixed the single process performance issue related to this, which we found in ClimaCore?

simonbyrne:
> Oh, one more thing: Have we fixed the single process performance issue related to this, which we found in ClimaCore?

Yes, that was fixed in 0.10.1

charleskawczynski (approved):

PR #309, opened 2022-04-15
Check in examples Manifest.toml

charleskawczynski:


jiahe23:
need to delete Manifest from gitignore

charleskawczynski:
I was thinking of leaving `Manifest.toml` in the .gitignore, so that we continue ignoring the one in the package directory

charleskawczynski:
We can manually add the other ones.

jiahe23:
bors try

charleskawczynski:
We'll still skip the Manifest in the test directory, since it seems that this causes issues with running `Pkg.test` (and codecov).

PR #310, opened 2022-04-15
Modify README in examples/hybrid/sphere/ 

jiahe23:


charleskawczynski (commented):
I don't think we don't need `ClimaAtmos.jl/Manifest.toml`, but it's probably good to add all the others. Also, we can change the buildkite pipelines (for all the added Manifest.toml) by removing the `Pkg.develop()` lines.

jiahe23:
> I don't think we don't need `ClimaAtmos.jl/Manifest.toml`, but it's probably good to add all the others. Also, we can change the buildkite pipelines (for all the added Manifest.toml) by removing the `Pkg.develop()` lines.

Yea... I will remove all the Manifest from this PR and rebase upon 309.

jiahe23:
bors try

charleskawczynski (approved):

charleskawczynski:
We hit the depot race condition.

Issue #301, opened 2022-04-14
Integration testing-focused AMIP roadmap


charleskawczynski:
The purpose of this issue is to outline an integration testing-focused roadmap, in order to accelerate iteration and optimally parallelize work across team members. Several tasks are needed to do this, below is an attempt to enumerate / outline these items

### Integration-testing-focused tasks

 - [x] `#1` From the AMIP roadmap document, extract a finite number of CI jobs (let's call them "milestones" with a unique JobID) which satisfy the following:
     - Using a list of composable acronyms for physics and configurations (a draft is below)
     - Write down a job unique JobID, to run on CI, for each item
 - [x] `#2` Sketch a dependency graph to help identify bottle necks and help with load balancing of distributed work -> Moved to [CA#374](https://github.com/CliMA/ClimaAtmos.jl/pull/374).
 - [x] `#3` Derive appropriate names for each job (i.e., a JobID). Depends on `#1`
 - [x] `#4` Add regression testing infrastructure. [CA#312](https://github.com/CliMA/ClimaAtmos.jl/pull/312), [CA#321](https://github.com/CliMA/ClimaAtmos.jl/pull/321), [CA#325](https://github.com/CliMA/ClimaAtmos.jl/pull/325)
 - [x] `#5` Add buildkite jobs for both regression tests and long runs. Loosely depends on `#1`, `#3`, tied to and `#4`. [CA#312](https://github.com/CliMA/ClimaAtmos.jl/pull/312), [CA#321](https://github.com/CliMA/ClimaAtmos.jl/pull/321), [CA#325](https://github.com/CliMA/ClimaAtmos.jl/pull/325)
 - [x] `#6` Add buildkite jobs for long runs [CA#298](https://github.com/CliMA/ClimaAtmos.jl/pull/298), [CA#304](https://github.com/CliMA/ClimaAtmos.jl/pull/304), [CA#317](https://github.com/CliMA/ClimaAtmos.jl/pull/317), [CA#319](https://github.com/CliMA/ClimaAtmos.jl/pull/319), [CA#323](https://github.com/CliMA/ClimaAtmos.jl/pull/323)
[...]

PR #312, opened 2022-04-15
Add regression infrastructure

charleskawczynski:
This PR adds regression tests, which involves adding the following code and logic:

New code at end of driver.jl:

1) Simulate case with `job_id`
2) load `all_best_mse` (from `mse_tables.jl`), which contains errors for each reference, and extract errors for the specific case (`best_mse` for the given `job_id`).
3) export `FieldVector` at last timestep to NC file
4) call `regression_test` which:
   - Grabs:
       - The last NC file on central that was merged into main. or
       - A specific NC file on central that was merged into main.
   - Returns a `computed_mse_dict` from computed NC (step 3) against the NC file on main
5) Export `computed_mse_dict` to a `.json`
6) Test that `computed_mse_dict` is no worse than the `best_mse` (determines if CI passes or not)

Additional jobs after all job_ids:

1) Print `computed_mse_dict` for all jobs to make updating `mse_tables.jl` easy
2) If on staging branch (all tests have passed, and the PR is effectively merging)
  - move NC files and `computed_mse_dict` for all `job_ids` into folder on central
  - folder has a commit sha prefix, and we can look this up based on when created

Docs are needed, both for functions and general workflow. I'll open an issue to document this to document the code a bit more. I'd rather delay it a bit because some of this code (e.g., `NCRegressionTests.jl` can be moved to an entirely separate repo with its own docs), and other functionality is likely to be moved into ClimaCore.

Being that things are moving quickly, and this testing infrastructure makes our CI stateful, I'm inclined to get this merged earlier than later so that we can build confidence with it.

I'm more than happy to answer questions, and I know that this is a large PR, but most of what's been added was transferred from TC.jl (which was built incrementally).

### DIFF
--- a/examples/hybrid/driver.jl
+++ b/examples/hybrid/driver.jl
@@ -266,7 +266,70 @@ if is_distributed # replace sol.u on the root processor with the global sol.u
         sol = DiffEqBase.sensitivity_solution(sol, global_sol_u, sol.t)
     end
 end
+
+import JSON
+using Test
+import OrderedCollections
 if !is_distributed || (is_distributed && ClimaComms.iamroot(Context))
     ENV["GKSwstype"] = "nul" # avoid displaying plots
     postprocessing(sol, output_dir)
+
+    include(joinpath(@__DIR__, "..", "..", "post_processing", "mse_tables.jl"))
+
+    if !occursin("Float64", job_id) # only do regression tests on Float32 jobs
+
+        Y_last = sol.u[end]
+        # This is helpful for starting up new tables
+        @info "Job-specific MSE table format:"
+        println("all_best_mse[$job_id] = OrderedCollections.OrderedDict()")
+        for prop_chain in Fields.property_chains(Y_last)
+            println("all_best_mse[$job_id][$prop_chain] = 0.0")
+        end
+        @info "Solution variables:"
+        for prop_chain in Fields.property_chains(Y_last)
+            println(prop_chain)
+        end
+
+        # Extract best mse for this job:
+        best_mse = all_best_mse[job_id]
+
+        include(joinpath(
+            @__DIR__,
+            "..",
+            "..",
+            "post_processing",
+            "compute_mse.jl",
+        ))
+
+        ds_filename_computed = joinpath(output_dir, "prog_state.nc")
+
+        function process_name(s::AbstractString)
+            # "c_ρ", "c_ρe", "c_uₕ_1", "c_uₕ_2", "f_w_1"
+            s = replace(s, "components_data_" => "")
+            s = replace(s, "ₕ" => "_h")
+            s = replace(s, "ρ" => "rho")
+            return s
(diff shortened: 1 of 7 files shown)


examples/hybrid/driver.jl
```diff
@@ -262,7 +262,58 @@ if is_distributed # replace sol.u on the root processor with the global sol.u
...
+        ))
+
+        ds_computed_filename = joinpath(output_dir, "prog_state.nc")
+
+        export_nc(Y_last, nc_filename = ds_computed_filename)
+        computed_mse = regression_test(
+            job_id,
+            best_mse,
+            ds_computed_filename,
+            ds_computed_filename;
```
charleskawczynski, line 307: We’ll need to remove this last positional argument after merging, since this temporarily used to allow merging before NC files have been saved on central.

jiahe23:
The general workflow looks good to me. I just have one comment here, do we really need the mse table?

As far as I understand, unlike `TC.jl` pycles is used as a background truth for mse calculation, here the mse will be computed between the PR that being tested and the main, that mean it is always compared against zero?

Or, do you anticipate any future scenarios where we could obtain a similar background truth that can be used to create a best mse table with main, plus a mse for PR to compare with the best mse?

charleskawczynski:
> The general workflow looks good to me. I just have one comment here, do we really need the mse table?

There are potentially other ways to do this, but the mse tables have one key advantage: any behavior change will require manually updating the tables (which we can easily do by copy-pasting the new tables from the `print_new_mse` job.

> As far as I understand, unlike `TC.jl` pycles is used as a background truth for mse calculation, here the mse will be computed between the PR that being tested and the main, that mean it is always compared against zero?

Yes, we use pycles datasets for the ground truth in TC, but I think it doesn't actually matter-- I don't think we've been really using the MSE tables to guide which behavior changing PRs should be merged. It's more of an indicator of whether results are at all changing. The numbers can float away from zero (which will happen when we approve behavior changes), but they'll be reset to zero when we update to new datasets, which will need to happen in a few failure mode scenarios:

 - Exported variable name, or where it lives, changes (since keys in `all_best_mse` change)
 - New variable added to MSE table (cannot compare against non-existent key)
 - Change spatial resolution (length along dims between computed and reference are no longer equal)

> Or, do you anticipate any future scenarios where we could obtain a similar background truth that can be used to create a best mse table with main, plus a mse for PR to compare with the best mse?

IIUC, yes, I think it's possible. There will be some limitations, though. For example, if / when we get digitized results from the Held Suarez publication, we could add mse comparisons against that. However, it will require post-processing and we'll only be able to compare in our long runs pipeline.

charleskawczynski:
I've tried my best to keep the regression infrastructure isolated in the driver, so that we can easily unhook it if/when we want to.

charleskawczynski:
Addresses item `#4` in #301.

PR #317, opened 2022-04-17
Make enable_threading a cli option, and use for long runs

charleskawczynski:
This PR:
 - Makes `enable_threading` a cli option
 - Makes the long runs use multi-threading
 - Fixes the `job_id` for HS rho-e to match the artifact path.


charleskawczynski:
Going to merge this now to see what sort of speedup we get in the long run at 5pm.

PR #322, opened 2022-04-18
Add cli args, solidify final buildkite jobs

charleskawczynski:
This PR is a step towards closing #301. In this PR:
 - `job_id`s are used for all of the long-term milestones that we need for AMIP
 - Adds CLI options so that we can compose the cases
 - Adds (and edits) the mse regression tables for these cases
 - Defaults `regression_test` to `false`, and turns on regression tests for these milestone cases.

This is just a draft bullet tracer PR for now, but I'd like feedback on names and moving in this direction in general.

### DIFF
--- a/examples/hybrid/cli_options.jl
+++ b/examples/hybrid/cli_options.jl
@@ -15,6 +15,45 @@ ArgParse.@add_arg_table s begin
     help = "Time between saving solution, 0 means do not save"
     arg_type = Float64
     default = Float64(60 * 60 * 24)
+    "--config"
+    help = "Spatial configuration [`SP`: Spherical, `SC`: Single Column, `BO`: Box]"
+    arg_type = String
+    default = "SP"
+    "--HSF"
+    help = "Held Suarez Forcing"
+    arg_type = Bool
+    default = false
+    "--topo"
+    help = "Use topology"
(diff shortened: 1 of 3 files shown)


examples/hybrid/cli_options.jl
```diff
@@ -15,6 +15,45 @@ ArgParse.@add_arg_table s begin
...
+    help = "Apply vertical diffusion"
+    arg_type = Bool
+    default = false
+    "--rad"
+    help = "Radiation model [`nothing`: no radiation, `gray`: gray radiation (RRTMGP), `all_sky`: all-sky radiation (RRTMGP)]"
+    arg_type = String
+    "--edmf"
+    help = "Use Eddy-Diffusivity Mass-Flux (EDMF) model"
+    arg_type = Bool
+    default = false
```
szy21, line 52: It might be better to have something like this, to be more consistent to other arguments (micro, rad, etc.)
```
"--turbconv"
    help = "Turbulent convection model [`nothing`: no turbulend convection, `edmf`: Eddy-Diffusivity Mass-Flux model]"
    arg_type = String
```
I don't have a strong preference though.

charleskawczynski:
I think we can close this as it's kind of evolved naturally.

PR #324, opened 2022-04-19
Add surface fluxes to diffusive boundary layer tendency

dennisYatunin:
This PR adds the use of SurfaceFluxes.jl to the `vertical_diffusion_boundary_layer` cache+tendency. The surface conditions are arbitrary (uniform temperature of 280 K, saturated with water vapor, with a drag coefficient of 0.001), and the mean of the boundary flux is used instead of the local boundary flux (since ClimaCore does not yet support spatially-varying boundary conditions), but this serves as a rough initial example.

### DIFF
--- a/examples/hybrid/sphere/baroclinic_wave_utilities.jl
+++ b/examples/hybrid/sphere/baroclinic_wave_utilities.jl
@@ -1,4 +1,7 @@
+using Statistics: mean
+using SurfaceFluxes
 using CloudMicrophysics
+const SF = SurfaceFluxes
 const CM = CloudMicrophysics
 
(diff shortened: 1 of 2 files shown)


dennisYatunin:
bors try

charleskawczynski:
Nice, it looks like the regression tests are working. Now we can tell which jobs (that use regression test infra) our PRs are changing.

I'll go over how to update PRs in the software meeting tomorrow. It's pretty easy. Just copy-paste the new MSE tables on buildkite into the `post_processing/mse_tables.jl` file. Then the PR should be ready to merge.

jiahe23 (approved):

PR #336, opened 2022-04-21
Docs: Typo fix in Equations

valeriabarra:
I think I spotted a typo. If not, never mind.


docs/src/equations.md
```diff
@@ -82,24 +82,24 @@ term treated implicitly.
...
 ```math
-\frac{\partial}{\partial t} \boldsymbol{u}  = - (2 \boldsymbol{\Omega} + \nabla \times \boldsymbol{u}) \times \boldsymbol{u} - \frac{1}{\rho} \nabla p  - \nabla(\Phi + K)
+\frac{\partial}{\partial t} \boldsymbol{u}  = - (2 \boldsymbol{\Omega} + \nabla \times \boldsymbol{u}) \times \boldsymbol{u} - \frac{1}{\rho} \nabla p  - \nabla(\Phi + K) .
 ```
 
 
 #### Horizontal velocity
 
-By breaking the curl and cross product terms into horizontal and vertical contributions, and removing zero terms (e.g. ``\nabla_v \boldsymbol{u}_v = 0``), we obtain
+By breaking the curl and cross product terms into horizontal and vertical contributions, and removing zero terms (e.g. ``\nabla_v  \times \boldsymbol{u}_v = 0``), we obtain
```
valeriabarra, line 91: @simonbyrne I think that before we had a typo. Can you please confirm?

docs/src/equations.md
```diff
@@ -111,25 +111,26 @@ The ``(2 \boldsymbol{\Omega}^v + \nabla_h \times \boldsymbol{u}_h) \times \bolds
...
-(C^f_v[\boldsymbol{u}_h] + C_h[\boldsymbol{u}_v]) \times I^f(\boldsymbol{u}_h)
+(C^f_v[\boldsymbol{u}_h] + C_h[\boldsymbol{u}_v]) \times I^f(\boldsymbol{u}_h) ,
 ```
 and the ``\frac{1}{\rho} \nabla_v p + \nabla_v(\Phi + K)`` term as
 ```math
-\frac{1}{I^f(\rho)} G^f_v[p] + G^f[K + \Phi]
+\frac{1}{I^f(\rho)} G^f_v[p] + G^f[K + \Phi] ,
 ```
-and treated implicitly.
+with the latter treated implicitly.
```
valeriabarra, line 133: @simonbyrne can you please confirm that I interpreted correctly? They way it read before was a bit confusing to understand which (only one or both terms?) are treated implicitly in this sentence. Thanks!

docs/src/equations.md
simonbyrne: Yes

docs/src/equations.md
```diff
@@ -10,8 +10,8 @@
 
 * ``\rho``: _mass_ in kg/m³. This is discretized at cell centers.
 * ``\boldsymbol{u}`` _velocity_, a vector in m/s. This is discretized via ``\boldsymbol{u} = \boldsymbol{u}_h + \boldsymbol{u}_v`` where
-  - ``\boldsymbol{u}_h = u_1 \boldsymbol{e}^1 + u_2 \boldsymbol{e}^2`` are the horizontal covariant part (with respect to the reference element), stored at cell centers.
-  - ``\boldsymbol{u}_v = u_3 \boldsymbol{e}^3`` is the vertical covariant part, stored at cell faces.
+  - ``\boldsymbol{u}_h = u_1 \boldsymbol{e}^1 + u_2 \boldsymbol{e}^2`` are the horizontal covariant vector (with respect to the reference element), stored at cell centers.
+  - ``\boldsymbol{u}_v = u_3 \boldsymbol{e}^3`` is the vertical covariant vector, stored at cell faces.
```
simonbyrne, line 14: I'm not a huge fan of the term "covariant vector": I think I would prefer "vector with horizontal covariant components" (but i also address this in #344.

valeriabarra: Okay, how do we want to proceed with these 2 somewhat overlapping PRs? Do you want me to update it here as you suggest, merge this, and then you can go on and rebase your branch?

valeriabarra: I applied your suggestion and pushed again

Issue #339, opened 2022-04-22
Do not conditionally load files based on TEST_NAME


charleskawczynski:
This kind of ruins compose-ability, and is needed for #322.

charleskawczynski:
Closed by #418.

PR #346, opened 2022-04-25
Unify driver more

charleskawczynski:
This PR:
 - Adds more cli options for moisture model and energy variable name
 - Uses the added cli options for dispatching to generalized methods in `driver.jl` (and post-processing)
 - Changes a kwarg in `center_initial_condition` to args (so that we can dispatch)
 - Adds a convenience `parse_arg` method to help trim some dynamically parsed arguments
 - Makes `params` constant

This is another step towards #339 and #301.

### DIFF
--- a/examples/hybrid/driver.jl
+++ b/examples/hybrid/driver.jl
@@ -3,6 +3,17 @@ include("cli_options.jl")
 const FT = parsed_args["FLOAT_TYPE"] == "Float64" ? Float64 : Float32
 TEST_NAME = parsed_args["TEST_NAME"]
 
+parse_arg(pa, key, default) = isnothing(pa[key]) ? default : pa[key]
+
+moisture_mode() = Symbol(parse_arg(parsed_args, "moist", "dry"))
+@assert moisture_mode() in (:dry, :equil, :nonequil)
+
+energy_variable(::Val{:rhoe}) = :ρe
+energy_variable(::Val{:rhoe_int}) = :ρe_int
+energy_variable(::Val{:rhotheta}) = :ρθ
+energy_name() =
+    energy_variable(Val(Symbol(parse_arg(parsed_args, "energy_name", "rhoe")))) # e.g., :ρθ
+
 using OrdinaryDiffEq
 using DiffEqCallbacks
 using JLD2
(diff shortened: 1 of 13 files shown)


charleskawczynski:
I'm a bit puzzled why this isn't working. I'll break it down a bit.

examples/hybrid/driver.jl
```diff
@@ -82,7 +105,7 @@ include(joinpath("sphere", "baroclinic_wave_utilities.jl"))
 const sponge = false
 
 # Variables required for driver.jl (modify as needed)
-params = BaroclinicWaveParameterSet()
+const params = BaroclinicWaveParameterSet()
```
dennisYatunin, line 108: What's the benefit of making this a `const`? It prevents us from looping over different parameter sets, and I don't see how it improves efficiency.

charleskawczynski: We don't have to, I think I added this because I was running into inference issues and I suspected this might be the problem

jiahe23:
with the current ambitious ddls, I strongly agree with @bischtob 's suggestion in the amip meeting today, on postpone driver polishing

charleskawczynski:
> with the current ambitious ddls, I strongly agree with @bischtob 's suggestion in the amip meeting today, on postpone driver polishing

I agree with @bischtob's suggestion about postponing driver polishing too. This PR is not about polishing the driver-- it's about making the driver more compose-able so that we can combine different physics components. Right now, we cannot combine the newly added radiation pieces with any other `TEST_NAME` file, because several global variables are being overwritten. Moving variables into cli arguments and allowing us to merge/load these files in a single driver is an essential integration step.

jiahe23:
Understood. Then, please modify README accordingly. Some early PRs messed up the hyperlinks there already, making it harder to set up experiment following the README.

jiahe23:
actually, since the cli is getting bigger and bigger, I am wondering if it's a good time to start thinking about namelist now?

charleskawczynski:
> actually, since the cli is getting bigger and bigger, I am wondering if it's a good time to start thinking about namelist now?

That's a good point. We use both in TC. The cli options overwrite the namelist, which is passed into the "`main`" function (which runs TC with whatever config is specified in the namelist).

I'm fine with supporting it if people want/need it now, but that would require us to clean up the global namespace (and unify the driver since the separated scripts define several global variables)

charleskawczynski:
Superseded by a bunch of other recent PRs.

PR #353, opened 2022-04-27
modify 0M microphy parameters for moist bw/hs case

jiahe23:


szy21 (approved):

examples/hybrid/sphere/baroclinic_wave_utilities.jl
```diff
@@ -13,6 +13,10 @@ Planet.grav(::BaroclinicWaveParameterSet) = 9.80616
 Planet.Omega(::BaroclinicWaveParameterSet) = 7.29212e-5
 Planet.planet_radius(::BaroclinicWaveParameterSet) = 6.371229e6
 
+# parameters for 0-Moment Microphysics
+Atmos.Microphysics_0M.τ_precip(::BaroclinicWaveParameterSet) = dt # timescale for precipitation removal
```
charleskawczynski, line 17: I just noticed this. This is reference to a global non-constant variable. This is a performance pitfall. If we really want to set `τ_precip` to `dt`, then we should probably make `dt` a property of `BaroclinicWaveParameterSet` and write this as 

```julia
Atmos.Microphysics_0M.τ_precip(param_set::BaroclinicWaveParameterSet) = param_set.dt # timescale for precipitation removal
```

charleskawczynski: It's a pretty big performance issue. Depending on how / where it's used, it could easily 4x the time to solution. I can open a PR

PR #356, opened 2022-04-27
make surface flux of density consistent

szy21:


jiahe23 (approved):

examples/hybrid/sphere/baroclinic_wave_utilities.jl
```diff
@@ -422,5 +422,6 @@ function vertical_diffusion_boundary_layer_tendency!(Yₜ, Y, p, t)
             bottom = Operators.SetValue(-mean(dif_flux_ρq_tot)),
         )
         @. Yₜ.c.ρq_tot += ᶜdivᵥ(ᶠK_E * ᶠinterp(ᶜρ) * ᶠgradᵥ(Y.c.ρq_tot / ᶜρ))
+        @. Yₜ.c.ρ += ᶜdivᵥ(FT(0) * ᶠgradᵥ(ᶜρ))
```
charleskawczynski, line 425: Isn't this just adding zero?

szy21: It has a bottom boundary condition though. Basically we want only surface flux to density, but not vertical diffusion in the interior.

PR #357, opened 2022-04-27
Add radiative equilibrium example

dennisYatunin:
This PR adds a single column radiative equilibrium example. With the way the driver is currently set up, I had to put the example in the `sphere` subfolder, and I had to add a workaround to avoid the default post-processing, but we can fix this when we next reorganize the driver.

This example starts with a column of air at 300 K, then runs RRTMGP using clear-sky optics every 3 hours to update the radiation flux until an equilibrium is reached. The concentration of water vapor in this example is 0, and the concentrations of all other gases required for clear-sky optics are read in from a dataset that comes with RRTMGP.

The resulting equilibrium temperature profile has some grid imprinting at around 80 km, but this might not be a significant issue.

### DIFF
--- a/examples/hybrid/driver.jl
+++ b/examples/hybrid/driver.jl
@@ -16,7 +16,7 @@ using ClimaCore
 # TODO: Allow some of these to be environment variables or command line arguments
 params = nothing
 horizontal_mesh = nothing # must be object of type AbstractMesh
-npoly = 0
+quad = nothing # must be object of type QuadratureStyle
 z_max = 0
 z_elem = 0
 t_end = if isnothing(parsed_args["t_end"])
@@ -34,7 +34,7 @@ dt_save_to_disk = 0 # 0 means don't save to disk
 ode_algorithm = nothing # must be object of type OrdinaryDiffEqAlgorithm
 jacobian_flags = () # only required by implicit ODE algorithms
 max_newton_iters = 10 # only required by ODE algorithms that use Newton's method
-show_progress_bar = true
+show_progress_bar = isinteractive()
 additional_callbacks = () # e.g., printing diagnostic information
 additional_solver_kwargs = () # e.g., abstol and reltol
 test_implicit_solver = false # makes solver extremely slow when set to `true`
@@ -84,7 +84,7 @@ const sponge = false
 # Variables required for driver.jl (modify as needed)
 params = BaroclinicWaveParameterSet()
 horizontal_mesh = baroclinic_wave_mesh(; params, h_elem = 4)
-npoly = 4
+quad = Spaces.Quadratures.GLL{5}()
 z_max = FT(30e3)
 z_elem = 10
 dt_save_to_disk = FT(0) # 0 means don't save to disk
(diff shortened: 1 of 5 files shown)


examples/hybrid/staggered_nonhydrostatic_model.jl
```diff
@@ -10,7 +10,7 @@ const TD = Thermodynamics
 include("schur_complement_W.jl")
 include("hyperdiffusion.jl")
 
-f_plane_coriolis_frequency(::AbstractEarthParameterSet) = 0
+f_plane_coriolis_frequency(::AbstractEarthParameterSet) = FT(0)
```
charleskawczynski, line 13: Let's leave this as it was, we need to wrap calls to parameter set functions with type casting anyway.

dennisYatunin: Ok, I moved the `FT()` call from inside this function to where this function is called.

examples/hybrid/driver.jl
```diff
@@ -235,7 +235,7 @@ integrator = OrdinaryDiffEq.init(
     dt = dt,
     adaptive = false,
     progress = show_progress_bar,
-    progress_steps = 1000,
+    progress_steps = 1,
```
charleskawczynski, line 238: This makes the output really bloated, can compromise at 100?

dennisYatunin: Since `show_progress_bar = is interactive()`, this only has an effect when running from the REPL. So having this set to 1 improves responsiveness without impacting CI.

charleskawczynski: I think we should keep progress bars in CI, just not this frequently

dennisYatunin: I'm confused as to what benefit we get from having a progress bar in CI? It just garbles the output (since the progress bar is meant for displaying in a terminal), and I don't think many people sit around watching CI logs as their code runs.

charleskawczynski: It's useful for long runs.

examples/hybrid/sphere/single_column_radiative_equilibrium.jl
```diff
@@ -0,0 +1,98 @@
...
+
+Δx = FT(1) # Note: This value shouldn't matter, since we only have 1 column.
+
+params = EarthParameterSet()
+horizontal_mesh =
+    periodic_rectangle_mesh(; x_max = Δx, y_max = Δx, x_elem = 1, y_elem = 1)
+quad = Spaces.Quadratures.GL{1}()
+z_max = FT(100e3)
+z_elem = 100
+t_end = FT(60 * 60 * 24 * 365.25 * 5)
```
szy21, line 17: so accurate...:)

szy21 (approved):

examples/hybrid/driver.jl
dennisYatunin: `progress_steps = isinteractive() ? 1 : 1000`?

charleskawczynski: That's fine by me

examples/hybrid/sphere/single_column_radiative_equilibrium.jl
```diff
@@ -0,0 +1,98 @@
+using PrettyTables
```
charleskawczynski, line 1: This kind of goes against the grain of #339, but I don't want to slow things down..

dennisYatunin: After I'm done with radiation, I'd like to do a PR that organizes all our `examples` code into modules and defines the "default driver setups" we discussed earlier. That should fix this issue.

PR #358, opened 2022-04-27
Add cli options, prep to unify driver

charleskawczynski:
This PR peels off some pieces from #346.

Hopefully this runs fine and I can return to #346.

### DIFF
--- a/.buildkite/longruns/pipeline.yml
+++ b/.buildkite/longruns/pipeline.yml
@@ -39,25 +39,25 @@ steps:
       # TODO: Add all milestone long simulation runs
 
       - label: ":computer: held suarez (ρe)"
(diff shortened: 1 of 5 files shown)


jiahe23 (changes requested):
Current changes look good to me. I do have some suggestions on the existing cli options that might be taken care of in this PR:
- default "--dt_save_to_sol" to zero;
- can we add "--dt_save_to_disk" as well?
- remove "--enable_threading" and do
```
if !is_distributed 
enable_threading() = true
end
```
in the `driver.jl`

charleskawczynski:
I spoke with @jiahe23 about this, we'll
 - Keep `dt_save_to_sol` as is, since it was the default for all the jobs
 - Add `dt_save_to_disk` and default to zero
 - Default `enable_threading` to true

jiahe23 (approved):
lgtm

Issue #326, opened 2022-04-19
Change to fixed reference datasets for regression tests


charleskawczynski:
Right now, our regression tests use a "rolling" reference-- we grab, and compare against, NC files from the last merged PR. What would be more "safe" is to fix the reference NC files from which we compare.

The rolling reference is somewhat problematic because our MSE values are not updated _after_ PRs are merged. So, what can happen is: we increase values in our MSE table and subsequent PRs could be behavior changing while remaining inside the tolerance of the new MSE values (e.g., reverting the PR would be allowed without re-updating the MSE tables).

charleskawczynski:
Closed by #366

PR #359, opened 2022-04-28
Store dt in parameter set for microphysics parameter

charleskawczynski:
This PR fixes a performance bug. See [this comment](https://github.com/CliMA/ClimaAtmos.jl/pull/353#discussion_r860327896). cc @szy21

### DIFF
--- a/examples/hybrid/sphere/baroclinic_wave_utilities.jl
+++ b/examples/hybrid/sphere/baroclinic_wave_utilities.jl
@@ -6,7 +6,9 @@ const CM = CloudMicrophysics
 
 include("../staggered_nonhydrostatic_model.jl")
 
-struct BaroclinicWaveParameterSet <: AbstractEarthParameterSet end
+struct BaroclinicWaveParameterSet{NT} <: AbstractEarthParameterSet
+    named_tuple::NT
+end
 Planet.R_d(::BaroclinicWaveParameterSet) = 287.0
 Planet.MSLP(::BaroclinicWaveParameterSet) = 1.0e5
 Planet.grav(::BaroclinicWaveParameterSet) = 9.80616
@@ -14,7 +16,8 @@ Planet.Omega(::BaroclinicWaveParameterSet) = 7.29212e-5
 Planet.planet_radius(::BaroclinicWaveParameterSet) = 6.371229e6
 
 # parameters for 0-Moment Microphysics
-Atmos.Microphysics_0M.τ_precip(::BaroclinicWaveParameterSet) = dt # timescale for precipitation removal
+Atmos.Microphysics_0M.τ_precip(param_set::BaroclinicWaveParameterSet) =
+    param_set.named_tuple.dt # timescale for precipitation removal
 Atmos.Microphysics_0M.qc_0(::BaroclinicWaveParameterSet) = 5e-6 # criterion for removal after supersaturation
 
 baroclinic_wave_mesh(; params, h_elem) =
--- a/examples/hybrid/driver.jl
+++ b/examples/hybrid/driver.jl
@@ -36,7 +36,7 @@ z_max = 0
 z_elem = 0
 z_stretch = nothing
 t_end = parse_arg(parsed_args, "t_end", FT(60 * 60 * 24 * 10))
-dt = parse_arg(parsed_args, "dt", FT(400))
+dt = FT(parse_arg(parsed_args, "dt", FT(400)))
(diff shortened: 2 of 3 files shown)


szy21 (approved):

szy21:
hmm why should this change the mse?

examples/hybrid/sphere/baroclinic_wave_utilities.jl
```diff
@@ -6,15 +6,18 @@ const CM = CloudMicrophysics
 
 include("../staggered_nonhydrostatic_model.jl")
 
-struct BaroclinicWaveParameterSet <: AbstractEarthParameterSet end
+struct BaroclinicWaveParameterSet{NT} <: AbstractEarthParameterSet
+    nt::NT
```
jiahe23, line 10: while you fix the buildkite failure, can we also think of a better property name here? I do not understand what `nt` referes to

charleskawczynski: Agreed, how about `named_tuple`?

charleskawczynski:
> hmm why should this change the mse?

@jiahe23 pointed out that we need to rebase (the mse tables have changed, so we're no longer inside the tolerance). I think all we need to do is rebase, I'll give that a shot.

szy21:
Ah, maybe I shouldn't have changed the mse table last time. The mse is within the tolerance so the ci passed, but I updated the table anyway: https://github.com/CliMA/ClimaAtmos.jl/pull/356/files. Should I change it back?

charleskawczynski:
> Ah, maybe I shouldn't have changed the mse table last time. The mse is within the tolerance so the ci passed, but I updated the table anyway: https://github.com/CliMA/ClimaAtmos.jl/pull/356/files. Should I change it back?

It's totally fine, the whole point is just to help identify / indicate when we do have behavioral changes--and in which cases.

charleskawczynski:
> How is the tolerance determined though? Why is it zero for some examples but quite large for some other examples? Maybe I misunderstood the mse. Is it compared with the current main?

The tolerance is an [averaged normalized error per variable](https://github.com/CliMA/NCRegressionTests.jl/blob/main/src/NCRegressionTests.jl#L82-L91). We currently compare with the latest main branch, but we should move to a semi-fixed reference (#326) to avoid "floating" results. It won't be too difficult to make this change, I need to tweak a few pieces, and what we have in the mean time is still effective.

szy21:
> We currently compare with the latest main branch

I see, I am not expecting any behavior change for this PR, so I should expect the mse to be all 0 after the rebase?

charleskawczynski:
> I see, I am not expecting any behavior change for this PR, so I should expect the mse to be all 0 after the rebase?

Non-behavior changing PRs should not require _updating_ the mse table (they can be, and remain non-zero).

szy21:
> > I see, I am not expecting any behavior change for this PR, so I should expect the mse to be all 0 after the rebase?
> 
> Non-behavior changing PRs should not require _updating_ the mse table (they can be, and remain non-zero).

Ok, thanks! Let's see what you get after the rebase.

charleskawczynski:
Interesting, I need to take a closer look

charleskawczynski:
I've updating the regression tests, and it should be simpler to determine if this is actually resulting in behavioral changes, since our reference dataset is now fixed (we can compare the reference here with the one in the latest merged branch).

charleskawczynski:
Ok, it seems that this PR is changing the following cases:

 - held suarez (ρe) equilmoist
 - held suarez (ρe_int) equilmoist

I'm going to try a separate branch that simply sets `const dt = `, to see if it's somehow changing. It could also be a type conversion, I suppose?

charleskawczynski:
Wow, this is definitely behavior changing. I looked at the results and compared against another branch where `dt` is set to a constant, but the changes are basically miniscule. It's possible that this somehow changes order of operations if the compiler can reason that the parameter is constant. I'm going to update the MSE tables and merge.

I think it would be helpful if we plotted figures that made comparisons a bit easier, comparing movies is difficult.

charleskawczynski:
Hmm, I want to check one more thing... We should be able to merge this either way, but understanding what's going on will be helpful down the road.

charleskawczynski:
It does seem like there are changes, it could be a reproducibility issue. See [this comment](https://github.com/CliMA/ClimaAtmos.jl/pull/307#issuecomment-1113934104).

The results still look fine, I'm updating the tables and we can merge.

PR #364, opened 2022-04-28
Add moist radiative equilibrium example

### LINKED ISSUE #349: Add single column clear sky radiative equilibrium example
- [x] Gray radiation
- [x] Clear sky radiation without water vapor
- [x] Clear sky radiation with water vapor

dennisYatunin:
This PR adds moisture to `single_column_radiative_equilibrium.jl`. Specifically, it adds an `idealized_h2o` flag to the `rrtmgp_model_cache`, which computes a volume mixing ratio at every point such that the relative humidity is at most 0.6 and the specific humidity decreases monotonically with height. Due to our unrealistic initial condition (300 K throughout the domain), the relative humidity is slowly increased from 0 to 0.6 to prevent RRTMGP from giving absurd values at the top of the domain at the start of the simulation.

Also, this PR adds grid stretching to the radiative equilibrium example (so that `dz = 100` at the bottom and `dz = 10000` at the top), and it makes grid stretching available for other examples that use `driver.jl`.

In addition, this PR allows non-idealized insolation to work for flat spaces, and it fixes a bug in non-idealized insolation that occurs when `FT = Float32`, though the radiative equilibrium example will continue to use idealized insolation by default.

Closes #349

### DIFF
--- a/examples/hybrid/sphere/single_column_radiative_equilibrium.jl
+++ b/examples/hybrid/sphere/single_column_radiative_equilibrium.jl
@@ -12,17 +12,19 @@ params = EarthParameterSet()
 horizontal_mesh =
     periodic_rectangle_mesh(; x_max = Δx, y_max = Δx, x_elem = 1, y_elem = 1)
 quad = Spaces.Quadratures.GL{1}()
-z_max = FT(100e3)
-z_elem = 100
-t_end = FT(60 * 60 * 24 * 365.25 * 5)
+z_max = FT(90e3)
+z_elem = 90
+z_stretch = Meshes.GeneralizedExponentialStretching(FT(100), FT(10000))
+t_end = FT(60 * 60 * 24 * 365.25)
 dt = FT(60 * 60 * 3)
-dt_save_to_sol = 100 * dt
+dt_save_to_sol = 10 * dt
 ode_algorithm = OrdinaryDiffEq.Rosenbrock23
 jacobian_flags = (;
     ∂ᶜ𝔼ₜ∂ᶠ𝕄_mode = 𝔼_name == :ρe ? :no_∂ᶜp∂ᶜK : :exact, ∂ᶠ𝕄ₜ∂ᶜρ_mode = :exact,
 )
 
-additional_cache(Y, params, dt) = rrtmgp_model_cache(Y, params)
+additional_cache(Y, params, dt) =
+    rrtmgp_model_cache(Y, params; idealized_h2o = true)
 additional_tendency!(Yₜ, Y, p, t) = rrtmgp_model_tendency!(Yₜ, Y, p, t)
 additional_callbacks = (PeriodicCallback(
     rrtmgp_model_callback!,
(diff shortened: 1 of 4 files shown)


szy21 (approved):

examples/hybrid/sphere/single_column_radiative_equilibrium.jl
```diff
@@ -12,17 +12,19 @@ params = EarthParameterSet()
 horizontal_mesh =
     periodic_rectangle_mesh(; x_max = Δx, y_max = Δx, x_elem = 1, y_elem = 1)
 quad = Spaces.Quadratures.GL{1}()
-z_max = FT(100e3)
-z_elem = 100
-t_end = FT(60 * 60 * 24 * 365.25 * 5)
+z_max = FT(90e3)
+z_elem = 90
+z_stretch = Meshes.GeneralizedExponentialStretching(FT(100), FT(10000))
+t_end = FT(60 * 60 * 24 * 365.25)
```
charleskawczynski, line 18: Can we use the cli option here instead?

examples/hybrid/sphere/single_column_radiative_equilibrium.jl
```diff
@@ -12,17 +12,19 @@ params = EarthParameterSet()
...
 dt = FT(60 * 60 * 3)
-dt_save_to_sol = 100 * dt
+dt_save_to_sol = 10 * dt
```
charleskawczynski, line 20: Can we use the cli option here instead?

szy21:
Waiting for Charlie's comments:)

examples/hybrid/sphere/single_column_radiative_equilibrium.jl
dennisYatunin: Hmm, passing something like `Meshes.GeneralizedExponentialStretching(FT(100), FT(10000))` through a CL argument can get rather unwieldy. I agree that we should be able to do it, but maybe we can do that in a separate PR. Also, it would be nice for examples to come with "default meshes", so that I can just run the radiative equilibrium example from the REPL without worrying about starting Julia with a particular `--stretch` argument.

examples/hybrid/sphere/single_column_radiative_equilibrium.jl
dennisYatunin: Same as with the other comment---I'd like to set a default value for this example, but have that default be different from the one for the baroclinic wave examples (1 day). Right now, this value is overriding the CL value, and that order of precedence should definitely be reversed, but I'll fix that when I modify the driver for REPL usage.

szy21:
@charleskawczynski I'll go ahead and merge this if it is ok with you:)

examples/hybrid/sphere/single_column_radiative_equilibrium.jl
charleskawczynski: I'm fine with merging for now, but this is an increasing problem since this script cannot be included with any other `TEST_NAME` script since each overwrite different global variables.

We can develop a more unified solution to making a generic script repl-friendly.

In the long term, much of this code is going to need to move into `ClimaAtmos.jl/src`, and increasing the number of global variables makes that more difficult.

dennisYatunin: Yes, I definitely agree that all of our driver-setup-related global variables need to be encapsulated in a data structure, and an order of precedence should be established for declaring these variables in different places (e.g., declaring a default `dt` for all baroclinic wave examples, declaring a default `dt` for a particular example, declaring a specific `dt` through a command-line argument, and declaring a specific `dt` through the REPL). I think I will work on this starting next Monday.

szy21:
Closes #349

PR #366, opened 2022-04-29
Update the regression test logic

charleskawczynski:
This PR closes out #326 by:
 - Upgrading to the latest NCRegressionTests, which is now more strict
 - Use the oldest reference dataset with a matching reference counter

I just recently cleared the julia depot path, so this should actually upgrade to the latest versions.


charleskawczynski:
I think I need to break this down, there's too much going on. I opened #370 to first get a handle on upgrading packages in a controlled way.

charleskawczynski:
Alright, I think this is working. I do need to update the MSE tables in order for CI to pass because we'll be comparing with an older (but fixed) dataset until we increment the reference counter.

PR #369, opened 2022-04-29
update jld2 filename

jiahe23:
Currently, the jld2 data is saved in the format of `dayX.jld2` which contains which day the data is. However, I found it useful to add seconds of the day into the filename when debugging. Usually, I save the data step-by-step on the day the simulation crashes and it overwrites if the filename contains only day info.


charleskawczynski (approved):
LGTM

PR #372, opened 2022-04-29
Hyperdiffusion (moisture) contribution to mass equation

akshaysridhar:
Update mass tendencies to reflect diffusive moisture tendencies.


jiahe23:
https://github.com/CliMA/ClimaAtmos.jl/blob/main/examples/hybrid/sphere/baroclinic_wave_utilities.jl#L425

Currently, the vertical diffusion does not apply to density (applying zero tendency), but it incorporate the surface flux of rho-qt into tho

szy21:
Let's add it to both hyperdiffusion and vertical diffusion.

jiahe23 (changes requested):
based on the discussion on slack, I think we should also do it to vertical diffusion

akshaysridhar:
> based on the discussion on slack, I think we should also do it to vertical diffusion

@jiahe23 Please see the updated PR ; if this looks OK ( + the buildkite updated mse tables ) I'll rebase with the updated tables in the postproc directory.

jiahe23 (approved):

PR #380, opened 2022-05-02
Add tracer advection to implicit tendency

dennisYatunin:
This PR moves the term for vertical advection of tracers from the explicit tendency to the implicit tendency, and it updates the Jacobian appropriately. Note that vertical advection is currently using the upwinding operator, but the Jacobian is still constructed as if there were no upwinding operator (this will be generalized in a later PR).


szy21:
Could you please also change the [docs](https://clima.github.io/ClimaAtmos.jl/dev/equations/) to make it consistent? Thanks!

Zhengyu-Huang (approved):
Great work!

PR #383, opened 2022-05-03
fix moist init cond

### LINKED ISSUE #385: Moist baroclinic wave looks wrong
There is a bug in the initial condition of q_tot in the moist baroclinic wave. After it is fixed we need a 4x decrease in dt to run the simulation, and the results look strange. See #383.

I think it is important to fix this before moving on to other physical tests. We will look into it.

jiahe23:
Fix a bug in setting up qt initial condition. radius vs degree

Closes #385


szy21 (approved):

bischtob:
Description is sometimes useful :D.

szy21:
The results look very good to me now 🎉  I don't know why the implicit vertical advection of qt makes such a difference though...

PR #388, opened 2022-05-03
switch to third-order upwinding

simonbyrne:


szy21:
The sphere cases look fine. You just need to update the mse I think. @charleskawczynski Should we turn off regression tests for moist cases for now? They are wrong anyway.

I don't know why the radiative equilibrium breaks with third-order upwinding. Maybe just the timestep is too large? I'll need to look into it.

szy21:
Could we wait for #383 if possible?

szy21 (approved):

charleskawczynski:
> @charleskawczynski Should we turn off regression tests for moist cases for now? They are wrong anyway. Edit: Let's think about the regression tests.

I think it's fine to leave them in-- they were designed to be easy to update so that we can use them while we're trying to get things working.

szy21:
Ah, I forgot to push the radiation change so it still breaks. I'll push it now.

charleskawczynski:
I'll rebase & merge

PR #389, opened 2022-05-03
clip vmr_h2o in radiation

szy21:
This PR clips the negative water vapor mixing ratio to zero in radiation.


dennisYatunin (approved):
sounds good

PR #390, opened 2022-05-03
enable stretching

simonbyrne:


szy21:
@simonbyrne Could you rebase? We have fixed several issues recently and the results with the current main are more meaningful.

simonbyrne:
Done: I would wait until the 3rd order upwinding is in before looking too closely though.

szy21:
@simonbyrne Is this ready for us to look at the results? It seems I may need to change the radiation as well. I'll look into it.

simonbyrne:
Yes? I've updated it for the 3rd order upwinding.

examples/common_spaces.jl
```diff
@@ -58,6 +58,7 @@ function make_hybrid_spaces(h_space, z_max, z_elem, z_stretch)
         boundary_tags = (:bottom, :top),
     )
     z_mesh = Meshes.IntervalMesh(z_domain, z_stretch; nelems = z_elem)
+    @info "z heights" z_mesh.faces
```
szy21, line 61: do we want this? (I think it's helpful)

simonbyrne: Yeah, no harm in keeping it.

szy21:
I changed dt to 200s and t_end to 4320000s in the baroclinic_wave_equilmoist_radiation. It should be good now.

szy21 (approved):

szy21:
We can merge it, but it would be still good to run the tracer advection over a mountain test case that Tapio suggested in ClimaCore.

szy21:
It's the post processing. Let me fix it.

.buildkite/pipeline.yml
```diff
@@ -150,7 +150,7 @@ steps:
         artifact_paths: "sphere_baroclinic_wave_rhoe_equilmoist/*"
 
       - label: ":computer: baroclinic wave (ρe) equilmoist radiation"
-        command: "julia --color=yes --project=examples examples/hybrid/driver.jl --moist equil --microphy 0M --rad clearsky --job_id sphere_baroclinic_wave_rhoe_equilmoist_radiation --dt 300"
+        command: "julia --color=yes --project=examples examples/hybrid/driver.jl --moist equil --microphy 0M --rad clearsky --job_id sphere_baroclinic_wave_rhoe_equilmoist_radiation --dt 200 --t_end 432000"
```
charleskawczynski, line 153: @szy21, it looks like this doesn't end up running to the time when we create the output plots (see [here](https://github.com/CliMA/ClimaAtmos.jl/blob/74727ebf7141c0ecced5c559489a77c95a030778/examples/hybrid/define_post_processing.jl#L141-L146)).

szy21: Yes, I'll just change it to postprocessing rather than paperplots by changing is_baro_wave. The baroclinic_wave_with_radiation test case is not a standard baroclinic wave anyway.

examples/hybrid/classify_case.jl
```diff
@@ -1,5 +1,8 @@
-is_baro_wave(parsed_args) =
-    all((parsed_args["config"] == "sphere", parsed_args["forcing"] == nothing))
+is_baro_wave(parsed_args) = all((
+    parsed_args["config"] == "sphere",
+    parsed_args["forcing"] == nothing,
+    parsed_args["rad"] == nothing,
```
szy21, line 4: @charleskawczynski looks good?

charleskawczynski: Yep, fine by me

PR #398, opened 2022-05-04
Add upwinding to Jacobian

dennisYatunin:
This PR adds a flag for upwinding, which is used in the implicit tendency and the Jacobian computation for the vertical advection of energy and tracer variables. The flag is set just like the energy variable flag (through a command-line argument and a function that returns a symbol). The possible values for this flag are

- `:none`, which uses central differencing for vertical advection; i.e., `ᶜvariableₜ = -ᶜdivᵥ(ᶠinterp(ᶜvariable) * ᶠw)`
- `:first_order`, which uses first-order upwinding; i.e., `ᶜvariableₜ = -ᶜdivᵥ(ᶠinterp(ᶜρ) * ᶠupwind_product(ᶠw, ᶜvariable / ᶜρ))`, where `ᶠupwind_product = Operators.UpwindBiasedProductC2F()`
- `:third_order`, which is the same as `:first_order`, but with `ᶠupwind_product = Operators.Upwind3rdOrderBiasedProductC2F(bottom = Operators.FirstOrderOneSided(), top = Operators.FirstOrderOneSided())`

Note that we could also use `Operators.ThirdOrderOneSided()` instead of `Operators.FirstOrderOneSided()` for `:third_order`. I don't think this makes too much of a difference, so I've just hardcoded `Operators.FirstOrderOneSided()` as the default.

### DIFF
--- a/examples/hybrid/staggered_nonhydrostatic_model.jl
+++ b/examples/hybrid/staggered_nonhydrostatic_model.jl
@@ -58,11 +58,11 @@ const ᶜFC = Operators.FluxCorrectionC2C(
     bottom = Operators.Extrapolate(),
     top = Operators.Extrapolate(),
 )
-const ᶠupwind_product = Operators.Upwind3rdOrderBiasedProductC2F(
+const ᶠupwind_product1 = Operators.UpwindBiasedProductC2F()
+const ᶠupwind_product3 = Operators.Upwind3rdOrderBiasedProductC2F(
     bottom = Operators.FirstOrderOneSided(),
     top = Operators.FirstOrderOneSided(),
 )
-
 
 const ᶜinterp_stencil = Operators.Operator2Stencil(ᶜinterp)
 const ᶠinterp_stencil = Operators.Operator2Stencil(ᶠinterp)
@@ -102,10 +102,12 @@ end
 thermo_state_ρe(ρe, Yc, K, Φ, params) =
     thermo_state_ρe_int(ρe - Yc.ρ * (K + Φ), Yc, params)
 
-get_cache(Y, params, dt) =
-    merge(default_cache(Y, params), additional_cache(Y, params, dt))
+get_cache(Y, params, upwinding_mode, dt) = merge(
+    default_cache(Y, params, upwinding_mode),
+    additional_cache(Y, params, dt),
+)
 
-function default_cache(Y, params)
+function default_cache(Y, params, upwinding_mode)
     ᶜcoord = Fields.local_geometry_field(Y.c).coordinates
     if eltype(ᶜcoord) <: Geometry.LatLongZPoint
         Ω = FT(Planet.Omega(params))
@@ -142,6 +144,9 @@ function default_cache(Y, params)
             Operators.StencilCoefs{-half, half, NTuple{2, FT}},
         ),
         params,
+        ᶠupwind_product = upwinding_mode == :first_order ? ᶠupwind_product1 :
+                          upwinding_mode == :third_order ? ᶠupwind_product3 :
+                          nothing,
         ghost_buffer = (
             c = Spaces.create_ghost_buffer(Y.c),
             f = Spaces.create_ghost_buffer(Y.f),
@@ -156,7 +161,7 @@ function implicit_tendency!(Yₜ, Y, p, t)
     ᶜρ = Y.c.ρ
     ᶜuₕ = Y.c.uₕ
     ᶠw = Y.f.w
-    (; ᶜK, ᶜΦ, ᶜts, ᶜp, params) = p
+    (; ᶜK, ᶜΦ, ᶜts, ᶜp, params, ᶠupwind_product) = p
 
     # Used for automatically computing the Jacobian ∂Yₜ/∂Y. Currently requires
     # allocation because the cache is stored separately from Y, which means that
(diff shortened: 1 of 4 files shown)


examples/hybrid/staggered_nonhydrostatic_model.jl
```diff
@@ -122,6 +128,13 @@ function default_cache(Y, params)
     else
         ts_type = TD.PhaseDry{FT}
     end
+    if upwinding_mode == :first_order
+        ᶠupwind_product = ᶠupwind_product1
+    elseif upwinding_mode == :third_order
+        ᶠupwind_product = ᶠupwind_product3
+    else
+        ᶠupwind_product = nothing
+    end
```
charleskawczynski, line 137:
```suggestion
    ᶠupwind_product = if upwinding_mode == :first_order
        ᶠupwind_product1
    elseif upwinding_mode == :third_order
        ᶠupwind_product3
    else
        nothing
    end
```

examples/hybrid/staggered_nonhydrostatic_model.jl
```diff
@@ -366,39 +407,76 @@ function Wfact!(W, Y, p, dtγ, t)
...
+        if isnothing(ᶠupwind_product)
+            # ᶜρθₜ = -ᶜdivᵥ(ᶠinterp(ᶜρθ) * ᶠw)
+            # ∂(ᶜρθₜ)/∂(ᶠw_data) = -ᶜdivᵥ_stencil(ᶠinterp(ᶜρθ) * ᶠw_unit)
+            @. ∂ᶜ𝔼ₜ∂ᶠ𝕄 = -(ᶜdivᵥ_stencil(ᶠinterp(ᶜρθ) * one(ᶠw)))
+        else
+            # ᶜρθₜ = -ᶜdivᵥ(ᶠinterp(ᶜρ) * ᶠupwind_product(ᶠw, ᶜρθ / ᶜρ))
+            # ∂(ᶜρθₜ)/∂(ᶠw_data) =
+            #     -ᶜdivᵥ_stencil(
+            #         ᶠinterp(ᶜρ) * ∂(ᶠupwind_product(ᶠw, ᶜρθ / ᶜρ))/∂(ᶠw_data),
+            #     )
```
charleskawczynski, line 419: Should we just delete this?

charleskawczynski: (same with commented out code below?)

charleskawczynski (commented):
Overall, these changes look fine. I commented on some questions I have regarding avoiding division by zero, and commented code (should we just delete)?

examples/hybrid/staggered_nonhydrostatic_model.jl
```diff
@@ -58,7 +58,11 @@ const ᶜFC = Operators.FluxCorrectionC2C(
     bottom = Operators.Extrapolate(),
     top = Operators.Extrapolate(),
 )
-const ᶠupwind_product = Operators.UpwindBiasedProductC2F()
+const ᶠupwind_product1 = Operators.UpwindBiasedProductC2F()
+const ᶠupwind_product3 = Operators.Upwind3rdOrderBiasedProductC2F(
+    bottom = Operators.ThirdOrderOneSided(),
+    top = Operators.ThirdOrderOneSided(),
+)
```
Zhengyu-Huang, line 66: to check

Zhengyu-Huang (approved):
Thanks to the great work, 2 things
1) default bc for 3rd order upwind
2) tracer, so far dTw = dT^{n} w^{n+1}

examples/hybrid/staggered_nonhydrostatic_model.jl
dennisYatunin: Is this just to improve legibility, or is there a niche Julia thing I'm unaware of?

examples/hybrid/staggered_nonhydrostatic_model.jl
dennisYatunin: The comments in this function are to help me remember how I derived the Jacobian. I'll get rid of all of them once I have automatic Jacobian constructions, but for now it's probably best to keep them.

examples/hybrid/staggered_nonhydrostatic_model.jl
```diff
@@ -335,6 +369,13 @@ function Wfact!(W, Y, p, dtγ, t)
     # ᶠw == ᶠw_data .* ᶠw_unit. The Jacobian blocks involve ᶠw_data, not ᶠw.
     ᶠw_data = ᶠw.components.data.:1
 
+    # If ∂(ᶜarg)/∂(ᶠw_data) = 0, then
+    # ∂(ᶠupwind_product(ᶠw, ᶜarg))/∂(ᶠw_data) =
+    #     ᶠupwind_product(ᶠw + εw, arg) / to_scalar(ᶠw + εw).
+    # The εw is only necessary in case w = 0.
+    εw = Ref(Geometry.Covariant3Vector(eps(FT)))
```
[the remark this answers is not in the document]
dennisYatunin: The value of `ᶠupwind_product(ᶠw, ᶜarg)` is proportional to `ᶠw_data = to_scalar.(ᶠw) = ᶠw.components.data.:1` at every point. So, its derivative is technically just `ᶠupwind_product(ᶠw, ᶜarg) ./ ᶠw_data`. Unfortunately, if `ᶠw_data` is 0, this results in `NaN`s in the Jacobian; since `0 * NaN = NaN`, this in turn results in `NaN`s in the linear solver's output. So, I need the `εw` to avoid that.

examples/hybrid/staggered_nonhydrostatic_model.jl
charleskawczynski: I think it's a julia thing 🙂

charleskawczynski (approved):

PR #399, opened 2022-05-04
Add some cli args, unify additional_cache

charleskawczynski:
I think this proposed pattern is a bit of a balance:

```
additional_cache(Y, params, dt; use_tempest_mode) = merge(
    hyperdiffusion_cache(Y; κ₄ = FT(2e17), use_tempest_mode),
    sponge ? rayleigh_sponge_cache(Y, dt) : NamedTuple(),
    isnothing(micro) ? NamedTuple() : zero_moment_microphysics_cache(Y),
    isnothing(HSF) ? NamedTuple() : held_suarez_cache(Y),
    isnothing(rad) ? NamedTuple() : rrtmgp_model_cache(Y, params; idealized_h2o),
    isnothing(turbconv) ? NamedTuple() : vertical_diffusion_boundary_layer_cache(Y),
)
```
We still use a closure, but everything that's included is kind of directly visible at the top level. Thoughts, @dennisYatunin, @jiahe23 ?

I haven't updated the cli options, yet, since they were being updated in #397.

### DIFF
--- a/examples/hybrid/cli_options.jl
+++ b/examples/hybrid/cli_options.jl
@@ -24,6 +24,19 @@ function parse_commandline()
         help = "Moisture model [`dry` (default), `equil`, `non_equil`]"
         arg_type = String
         default = "dry"
+        "--microphy"
+        help = "Microphysics model [`nothing` (default), `0M`]"
+        arg_type = String
+        "--forcing"
+        help = "Forcing [`nothing` (default), `held_suarez`]"
+        arg_type = String
+        "--turbconv"
+        help = "Turbulence Convection model [`nothing` (default), `const_diffusivity`]"
+        arg_type = String
+        "--idealized_h2o"
+        help = "Use idealized H2O in radiation model [`false` (default), `true`]"
+        arg_type = Bool
+        default = false
         "--rad"
         help = "Radiation model [`clearsky`, `gray`, `allsky`] (default: no radiation)"
         arg_type = String
--- a/.buildkite/longruns/pipeline.yml
+++ b/.buildkite/longruns/pipeline.yml
@@ -39,25 +39,25 @@ steps:
       # TODO: Add all milestone long simulation runs
 
       - label: ":computer: held suarez (ρe)"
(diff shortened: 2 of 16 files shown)


examples/hybrid/cli_options.jl
```diff
@@ -24,6 +24,20 @@ function parse_commandline()
         help = "Moisture model [`dry` (default), `equil`, `non_equil`]"
         arg_type = String
         default = "dry"
+        "--micro"
```
jiahe23, line 27: I would call it `microphy`

examples/hybrid/cli_options.jl
```diff
@@ -24,6 +24,20 @@ function parse_commandline()
...
+        help = "Microphysics model [`nothing` (default), `0M`]"
+        arg_type = String
+        "--HSF"
```
jiahe23, line 30: Can we have a `---forcing` default to none and make held-suarez forcing an option here?

examples/hybrid/cli_options.jl
```diff
@@ -24,6 +24,20 @@ function parse_commandline()
...
+        help = "Held Suarez Forcing [`false` (default), `true`]"
+        arg_type = Bool
+        default = false
+        "--turbconv"
+        help = "Turbulence Convection model [`nothing` (default), `const_diffusivity`]"
+        arg_type = String
+        "--idealized_h2o"
```
jiahe23, line 37: Can we make this an option in `--rad`

jiahe23:
I left a few comments. Also, don't forget to modify the README

examples/hybrid/cli_options.jl
charleskawczynski: Yesss, thank you!

examples/hybrid/cli_options.jl
charleskawczynski: Sounds good to me, how can we do that?

charleskawczynski: I'm not sure how to do this off the top of my head, so I'm going to save this for a separate PR.

charleskawczynski:
I forgot the readme. Fixed now.

examples/hybrid/cli_options.jl
jiahe23: might need some inputs from zhaoyi and dennis

PR #403, opened 2022-05-05
Prep for removing TEST_NAME

charleskawczynski:
This PR
 - Adds the `hyperdiff` cli arg
 - unifies `additional_tendency!`
 - Fixes the `turbconv` flag for `const_diff`
 - Deletes several files:
```
examples/hybrid/sphere/baroclinic_wave_rhoe.jl
examples/hybrid/sphere/baroclinic_wave_rhoe_equilmoist.jl
examples/hybrid/sphere/baroclinic_wave_rhoe_equilmoist_radiation.jl
examples/hybrid/sphere/baroclinic_wave_rhotheta.jl
examples/hybrid/sphere/held_suarez_rhoe.jl
examples/hybrid/sphere/held_suarez_rhoe_equilmoist.jl
examples/hybrid/sphere/held_suarez_rhoe_int.jl
examples/hybrid/sphere/held_suarez_rhoe_int_equilmoist.jl
examples/hybrid/sphere/held_suarez_rhotheta.jl
```


charleskawczynski:
Still not quite sure what's going on with the `single column radiative equilibrium`, but I think this PR is close to ready.

.buildkite/pipeline.yml
```diff
@@ -167,18 +167,18 @@ steps:
...
+        command: "julia --color=yes --project=examples examples/hybrid/driver.jl --TEST_NAME held_suarez_rhoe_equilmoist --turbconv const_diff --moist equil --forcing held_suarez --microphy 0M --job_id sphere_held_suarez_rhoe_equilmoist --regression_test true --dt 200 --t_end 432000"
         artifact_paths: "sphere_held_suarez_rhoe_equilmoist/*"
 
       - label: ":computer: held suarez (ρe_int) equilmoist"
-        command: "julia --color=yes --project=examples examples/hybrid/driver.jl --TEST_NAME held_suarez_rhoe_int_equilmoist --turbconv const_diffusivity --energy_name rhoe_int --forcing held_suarez --moist equil --microphy 0M --job_id sphere_held_suarez_rhoe_int_equilmoist --regression_test true --dt 200 --t_end 432000"
+        command: "julia --color=yes --project=examples examples/hybrid/driver.jl --TEST_NAME held_suarez_rhoe_int_equilmoist --turbconv const_diff --energy_name rhoe_int --forcing held_suarez --moist equil --microphy 0M --job_id sphere_held_suarez_rhoe_int_equilmoist --regression_test true --dt 200 --t_end 432000"
```
jiahe23, line 174: I may be careless in the previous PR review, but I think we want vertical diffusion a separate option as turbconv, meaning that we could have edmf + 3d-const-visc + vertical-diff all turned on

examples/hybrid/driver.jl
```diff
@@ -69,10 +77,27 @@ additional_cache(Y, params, dt; use_tempest_mode = false) = merge(
...
 
-additional_tendency!(Yₜ, Y, p, t) = nothing
-postprocessing(sol, output_dir) = nothing
+additional_tendency!(Yₜ, Y, p, t) = begin
+    (; rad_flux, const_diff, hs_forcing) = p.tendency_knobs
+    (; microphy_0M, hyperdiff) = p.tendency_knobs
+    hyperdiff && hyperdiffusion_tendency!(Yₜ, Y, p, t)
+    sponge && rayleigh_sponge_tendency!(Yₜ, Y, p, t)
+    hs_forcing && held_suarez_tendency!(Yₜ, Y, p, t)
+    const_diff && vertical_diffusion_boundary_layer_tendency!(Yₜ, Y, p, t)
```
jiahe23, line 97: `const_diff` is not a proper name for this particular function. `vertical_diffusion_boundary_layer_tendency!` is following the paper where diffusivity is not constant.

.buildkite/pipeline.yml
charleskawczynski: I'm certainly open to this (I think this was what I did to start with), but I think I shared the same flag for constant diff with edmf based off @yairchn's suggestion. If we split it, then what should we call the different options?

examples/hybrid/driver.jl
charleskawczynski: What do you recommend we call it?

jiahe23:
I am a little concerned on completely deleting experiment files and putting every choice into cli. I could think of a scenario where I would like to test out some self-defined forcing/tendency on one experiment (e.g., adding qt relaxation on mhs). It is most convenient if I could simply add the cache needed to the additional_cache and add the rhs function needed to the additional_tendency without worrying about how to set up a proper cli switch to enable it.

jiahe23:
@dennisYatunin might have some solution for the above situation I ran into today.

dennisYatunin:
@jiahe23 Yes, I have a concrete solution to this issue, which I think will be ready in 1~2 weeks. Essentially, there will be an `if` statement at the top of `driver.jl`, which distinguishes between REPL use and CL use. In the CL case, you can specify a base driver configuration (e.g., `--dycore_test_config` or `--radiative_equilibrium_config`), followed by a bung of options/flags that toggle between different settings for this configuration. In the REPL case, you can call a driver configuration function (e.g., `dycore_test_config(kwargs...)`), whose keyword arguments are exactly the same as the aforementioned options/flags. After the driver configuration is defined, `driver.jl` will work the same way in both cases (i.e., CL args will only ever be used to construct the driver configuration). To test out a new tendency, like qt relaxation, you will define a new `AbstractTendency` object, and pass that object to your driver configuration through the REPL. After you are done experimenting in the REPL, and you want to test out your tendency through the buildkite pipeline, you can add a flag to, e.g., `dycore_test_config` that toggles on this tendency, and add a new line to the pipeline that includes this flag.

.buildkite/pipeline.yml
yairchn: not sure I see the logic of having all `edmf + 3d-const-visc + vertical-diff` but sure why not

examples/hybrid/driver.jl
yairchn: maybe `vert_diff_BL`

charleskawczynski:
> I am a little concerned on...

Yeah, I think we have several solution options to making REPL development easier. @dennisYatunin's suggestion should work fine, I think there's a decent amount of flexibility at this level.

One issue I mentioned to @dennisYatunin earlier was that I've got to change all of these additional files if I want to change the interface for `additional_tendency`. If I followed this pattern when adding dry edmf, moist edmf, dry edmf + radiation, moist edmf + radiation, (then combinations with topography) this multi-file include approach is just not going to work. With the proposed changes in this PR, there's a single interface to change.

.buildkite/pipeline.yml
jiahe23: Yea, this is something Tapio mentioned several times. I would just call it something like `vert_diff` for vertical only diffusion and `turbulence` as a place holder if we want to add second order diffusion, and keep `turbconv` for edmf

examples/hybrid/driver.jl
charleskawczynski: Actually, `vert_diff` was proposed above, I'll stick with that for now, I think we can further improve names down the road, though.

PR #406, opened 2022-05-05
Remove non-driver examples

charleskawczynski:
The bubble MPI job is flakey due to not maintaining the example with upstream changes. This PR removes non-driver examples from CI, since they're already being maintained / used in ClimaCore for unit/integration testing.

I've also removed the job `held suarez (ρθ) Float64`, since we already have a BW (ρθ) Float64 test. This should help us in reorganizing the examples folder, maintaining code, and reduce wasting CI resources.


dennisYatunin:
@akshaysridhar You aren't developing any of these examples in ClimaAtmos at the moment, are you?

akshaysridhar:
@charleskawczynski These are being maintained in the ClimaCore repository / up-to-date with integration tests there so we can remove the 2D hybrid problems from here.

charleskawczynski:
> We currently don't support parallel runs with hybrid 2d configurations in any case?

I suppose not? The only MPI-parallel cases are the two in the MPI groups (the box bubble and sphere BW).

akshaysridhar:
Of the 6 experiments affected 
**Box** 

- 3D bubble (rho e) : Remove (Useful for MPI interface checks ; but this is part of ClimaCore buildkite) 
- 3D bubble (rho theta, flux form) : Remove (consistent with removal of held suarez rho theta option) 
- 3D bubble (rho theta, curl-advection) : Remove (consistent with removal of held suarez rho theta option)

**Plane** 

- 2D Density current (flux form) : Remove (LES demonstration, low priority for AMIP, parallel runs unsupported) 
- 2D Density current (rho e, curl-advection) : Remove (parallel runs unsupported, file is otherwise accounted for in ClimaCore integration tests) 
- 2D Density current (rho theta, curl-advection) : Remove (parallel runs unsupported, consistent with removal of held suarez rho theta option)  

Topography interface changes will first affect ClimaCore (unit + integration tests) ; then a condensed list of examples will be included in ClimaAtmos following the AMIP test plan. Removal of these files is still consistent with using the staggered non-hydrostatic model (without coriolis terms) in a box; so if it helps reduce CI burden, I'm ok with this.

PR #420, opened 2022-05-07
Remove TEST_NAME

charleskawczynski:
This PR removes the `TEST_NAME` flag, and adds two functions: `is_baro_wave`, `is_column_radiative_equilibrium` which are in some ways similar to `TEST_NAME`-- except that they're determined from the configuration, physics, etc. not the other way around.

We could effectively inline these functions, but it's not clear to me what all the right "boxes" are for each core config/model/physics we want to pivot off of.

### DIFF
--- a/examples/hybrid/classify_case.jl
+++ b/examples/hybrid/classify_case.jl
@@ -1,0 +1,8 @@
+is_baro_wave(parsed_args) =
+    all((parsed_args["config"] == "sphere", parsed_args["forcing"] == nothing))
+
+is_column_radiative_equilibrium(parsed_args) = all((
+    parsed_args["config"] == "column",
+    parsed_args["forcing"] == nothing,
+    parsed_args["rad"] != nothing,
+))
--- a/.buildkite/longruns/pipeline.yml
+++ b/.buildkite/longruns/pipeline.yml
@@ -39,13 +39,13 @@ steps:
       # TODO: Add all milestone long simulation runs
 
       - label: ":computer: held suarez (ρe)"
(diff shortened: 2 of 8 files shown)


examples/hybrid/classify_case.jl
```diff
@@ -0,0 +1,10 @@
+is_baro_wave(parsed_args) = all((
+    parsed_args["config"] == "sphere",
+    parsed_args["forcing"] == nothing,
+))
+
+is_column_radiative_equilibrium(parsed_args) = all((
+    parsed_args["config"] == "column",
+    parsed_args["forcing"] == nothing,
+    parsed_args["rad"] ≠ nothing,
```
jiahe23, line 9: I prefer `!=` than `≠`. A lot of us has experienced unresolved unicodes in certain browser/IDE setups.

jiahe23:
This is a less related comment but I never found a better time to bring it up. Can we rename `baroclinic_wave_utilities.jl` since we are also adding held-suarez forcing in this file? 

Could be done here or in a separate PR, but need some better name. @szy21 Any ideas?

szy21:
Dennis is going to change it. It will be something like dycore_test_utilities I think.

charleskawczynski:
> This is a less related comment but I never found a better time to bring it up. Can we rename `baroclinic_wave_utilities.jl` since we are also adding held-suarez forcing in this file?
> 
> Could be done here or in a separate PR, but need some better name. @szy21 Any ideas?

Yes, I've been wanting to to this also. I think we can do that and move a few other things around in `examples/hybrid/sphere`. I'm happy to open up a follow-up PR!

jiahe23 (approved):
LGTM

PR #422, opened 2022-05-09
Add time_to_string func for cl args

charleskawczynski:
@jiahe23 suggested enabling specifying `t_end` with `20days` or `20secs`. This PR adds support for this and updates the docs accordingly. For now I've added this to `t_end` and `dt`, would it be useful for other CL args?

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -127,18 +127,18 @@ steps:
         artifact_paths: "sphere_held_suarez_rhotheta/*"
 
       - label: ":computer: held suarez (ρe) equilmoist"
-        command: "julia --color=yes --project=examples examples/hybrid/driver.jl --vert_diff true --moist equil --forcing held_suarez --microphy 0M --job_id sphere_held_suarez_rhoe_equilmoist --regression_test true --dt 200 --t_end 432000"
+        command: "julia --color=yes --project=examples examples/hybrid/driver.jl --vert_diff true --moist equil --forcing held_suarez --microphy 0M --job_id sphere_held_suarez_rhoe_equilmoist --regression_test true --dt 200secs --t_end 5days"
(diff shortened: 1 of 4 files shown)


.buildkite/pipeline.yml
```diff
@@ -127,30 +127,30 @@ steps:
...
+        command: "julia --color=yes --project=examples examples/hybrid/driver.jl --vert_diff true --energy_name rhoe_int --forcing held_suarez --moist equil --microphy 0M --job_id sphere_held_suarez_rhoe_int_equilmoist --regression_test true --dt 200secs --t_end 5days"
         artifact_paths: "sphere_held_suarez_rhoe_int_equilmoist/*"
 
   - group: "Milestones"
     steps:
 
       - label: ":computer: single column radiative equilibrium"
-        command: "julia --color=yes --project=examples examples/hybrid/driver.jl --rad clearsky --idealized_h2o true --hyperdiff false --config column --t_end 31536000 --dt 10800 --dt_save_to_sol 108000 --job_id sphere_single_column_radiative_equilibrium"
+        command: "julia --color=yes --project=examples examples/hybrid/driver.jl --rad clearsky --idealized_h2o true --hyperdiff false --config column --t_end 654days --dt 3hours --dt_save_to_sol 108000 --job_id sphere_single_column_radiative_equilibrium"
```
jiahe23, line 141: can we also make it possible for `dt_save_to_sol` and `dt_save_to_disk`?

charleskawczynski: I think I didn't do that to start with since we special case `dt_save_to_sol = 0` and `dt_save_to_disk = 0` (meaning don't save). But this sounds good if we change the default to `Inf` ( which I think makes more sense anyway ).

charleskawczynski: Updated, let me know what you think!

jiahe23 (approved):

PR #428, opened 2022-05-09
Add spatially varying surface fluxes

dennisYatunin:
This PR adds a realistic surface temperature profile and applies surface fluxes pointwise (instead of using the mean of all surface fluxes).


jiahe23 (approved):
lol

szy21 (approved):

Issue #427, opened 2022-05-09
Clean Up The Driver (🧹🚗🧹)


dennisYatunin:
## Purpose
Thanks to recent work by @charleskawczynski, the driver is now unified, in that it is localized to `driver.jl` plus a small number of `include`d files, and every simulation can be run with `julia --project driver.jl <command-line args>`. Although we can start adding features like EDMF and coupling now, there are several issues with the current setup that have been becoming increasingly problematic, and we should probably sort them out before adding major new features. These issues are:

- Experimenting with the simulation through the REPL (or a Jupyter notebook) requires awkwardly circumventing the command-line argument interface. All the parameters of the driver are set through `parsed_args`, which is a dictionary that gets returned by `ArgParse.parse_args()` when `driver.jl` is run. If `include("driver.jl")` is called in the REPL, there are no command-line arguments available, so this dictionary does not contain all the information necessary to run the driver. The current workaround is to modify `driver.jl`, replacing every use of `parsed_args[<driver parameter that must be set>]` with the desired driver parameter value. This is rather tedious, and it would be better if the driver did not have to be modified unless new driver-related functionality is getting added. So, for example, in order to run a simulation that we use for CI from the REPL, and then rerun the simulation with an additional tendency term and compare the results, one should not need to modify `driver.jl`, and one certainly should not need to think about the command-line argument interface. This issue was mentioned [here](https://github.com/CliMA/ClimaAtmos.jl/pull/403#issuecomment-1119161894).
- The current chain of calls to `include` makes it difficult to trace where functions are defined, and, more broadly, makes it difficult to understand the code at a high level. In addition, since our code is a series of files that need to get `include`d at runtime in a specific order, it cannot be moved into `/src` and turned into a self-contained package.
[...]

PR #430, opened 2022-05-09
Improve interactivity, conditionally define parsed_args

charleskawczynski:
I think that this should improve life for people regarding bullet 1 in #427.


dennisYatunin (approved):
Definitely a good start on fixing that first point, but we still need something more advanced (see my other comments on that issue).

PR #433, opened 2022-05-10
Unify TC into driver

charleskawczynski:
This PR allows us to turn on TurbulenceConvection, and uses it for a new job.

I've been using this to run:

```julia
using Revise
include("examples/hybrid/cli_options.jl")
(s, parsed_args) = parse_commandline();

parsed_args["config"] = "column";
parsed_args["FLOAT_TYPE"] = "Float64";
parsed_args["moist"] = "equil";
parsed_args["turbconv"] = "edmf";
parsed_args["ode_algo"] = "Euler";
parsed_args["energy_name"] = "rhotheta";
parsed_args["split_ode"] = false;
parsed_args["job_id"] = "edmf_single_column";
parsed_args["dt"] = "1secs";
parsed_args["t_end"] = "1mins";

```

Some remaining items are (for follow-up PRs)
 - Synchronizing names
 - More refactoring / synchronizing types

I think that this is finished and ready. It's pretty clunky at the moment, due to many edmf initialization details that we should probably improve. But I think it'd be beneficial to get something working and then we can make things look better while maintaining results.


charleskawczynski:
I'm having trouble debugging through the `Cannot infer eltype` error.

@simonbyrne, can you take a look?

```
# Works
using Revise
include(joinpath("examples", "hybrid", "cli_options.jl"));
(s, parsed_args) = parse_commandline();
parsed_args["config"] = "column";
parsed_args["moist"] = "equil";
parsed_args["turbconv"] = nothing;
parsed_args["job_id"] = "edmf_single_column";
parsed_args["dt"] = "1secs ";
parsed_args["t_end"] = "1mins";
include(joinpath("examples", "hybrid", "driver.jl"))
```

```
# Does not work
using Revise
include(joinpath("examples", "hybrid", "cli_options.jl"));
(s, parsed_args) = parse_commandline();
parsed_args["config"] = "column";
parsed_args["moist"] = "equil";
parsed_args["turbconv"] = "edmf";
parsed_args["job_id"] = "edmf_single_column";
parsed_args["dt"] = "1secs ";
parsed_args["t_end"] = "1mins";
include(joinpath("examples", "hybrid", "driver.jl"))
```

I tried looking at

```
JET.@test_opt init_state(
        center_initial_condition,
        face_initial_condition,
        center_space,
        face_space,
        params,
        models,
    )
```
which doesn't error, and
```
@code_warntype init_state(
        center_initial_condition,
        face_initial_condition,
        center_space,
        face_space,
        params,
        models,
    )
```
but nothing is really sticking out as an issue.

However, types other than `turbconv_model` do seem to have a `Core.Const` wrapper around them, but it's not clear how and when that occurs. Can we print the broadcasted object in the error message so that we can more easily reproduce where inference is failing on `eltype(::Broadcasted)`?

szy21:
The overall structure looks good to me, but it would be better for someone more familiar with edmf to review it I think.

examples/hybrid/TurbulenceConvectionUtils.jl
```diff
@@ -0,0 +1,283 @@
...
+import Logging
+import TerminalLoggers
+Logging.global_logger(TerminalLoggers.TerminalLogger())
+
+import TurbulenceConvection
+const TC = TurbulenceConvection
+const tc_dir = pkgdir(TC)
+
+Base.:-(a::FT, b::CCG.Covariant3Vector{FT}) where {FT} = a .- b.u₃
+Base.:+(a::FT, b::CCG.Covariant3Vector{FT}) where {FT} = a .+ b.u₃
```
charleskawczynski, line 27: @simonbyrne, these operations weren't previously defined, is this okay in the context of a sphere? The results seem to be fine with a rectangle mesh.

dennisYatunin: It would probably be less confusing to just wrap `a` in a `Covariant3Vector` before calling `+` or `-`.

charleskawczynski:
I'll hook in the tests once we have a new release that has [CC 726](https://github.com/CliMA/ClimaCore.jl/pull/726).

szy21:
Do you mind me merging #474 then?

charleskawczynski:
> Do you mind me merging #474 then?

That's fine

examples/hybrid/TurbulenceConvectionUtils.jl
charleskawczynski: I'll open an issue for this, and remove the base overloading later. I don't think it's too confusing on the EDMF side: the covariant3 vector only has one component, so we basically treat everything like scalars.

charleskawczynski: I've migrated this to a TC issue.

PR #438, opened 2022-05-10
Improve postproc for dry held-suarez

jiahe23:
This postproc is only turned on when the simulation time is greater or equal to 400 days. The first 200 days are thrown away for the statistics computation.

It produces plots comparable to Figs 1-3 in [this paper](https://journals.ametsoc.org/view/journals/bams/75/10/1520-0477_1994_075_1825_apftio_2_0_co_2.xml?tab_body=pdf). Fig 4 needs additional capability of spectra calculation, which will be added later.

closes #421

PR #442, opened 2022-05-11
Define and use some types

charleskawczynski:
This PR is a prep PR for adding edmf. It makes more sense to use dispatch over further specializing with `Val` over more symbols that come with using edmf. So, this PR adds and uses some types, some of which will be used to create an edmf model, so that we can create the state vector. I think that the cache is nearly finished / compatible, so we're getting pretty close.

None of this is final, I think there's still many changes to come, it's just a small step towards being compatible with edmf requirements.

Once all of this is in, @dennisYatunin can apply the driver revamp and I suspect much of this will be more polished

### DIFF
--- a/examples/hybrid/define_post_processing.jl
+++ b/examples/hybrid/define_post_processing.jl
@@ -128,26 +128,26 @@ end
 # Dispatcher:
 # baroclinic wave
 paperplots_baro_wave(args...) =
-    paperplots_baro_wave(Val(energy_name()), Val(moisture_mode()), args...)
+    paperplots_baro_wave(energy_form(), moisture_model(), args...)
 
-paperplots_baro_wave(::Val{:ρθ}, ::Val{:dry}, args...) =
+paperplots_baro_wave(::PotentialTemperature, ::DryModel, args...) =
     paperplots_baro_wave_ρθ(args...)
-paperplots_baro_wave(::Val{:ρe}, ::Val{:dry}, args...) =
+paperplots_baro_wave(::TotalEnergy, ::DryModel, args...) =
     paperplots_baro_wave_ρe(args...)
-paperplots_baro_wave(::Val{:ρe}, ::Val{:equil}, args...) =
+paperplots_baro_wave(::TotalEnergy, ::EquilMoistModel, args...) =
     paperplots_moist_baro_wave_ρe(args...)
 
 # held-suarez
 paperplots_held_suarez(args...) =
-    paperplots_held_suarez(Val(energy_name()), Val(moisture_mode()), args...)
+    paperplots_held_suarez(energy_form(), moisture_model(), args...)
 
-paperplots_held_suarez(::Val{:ρθ}, ::Val{:dry}, args...) =
+paperplots_held_suarez(::PotentialTemperature, ::DryModel, args...) =
     paperplots_dry_held_suarez_ρθ(args...)
-paperplots_held_suarez(::Val{:ρe}, ::Val{:dry}, args...) =
+paperplots_held_suarez(::TotalEnergy, ::DryModel, args...) =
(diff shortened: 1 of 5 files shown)


dennisYatunin (approved):

examples/hybrid/define_post_processing.jl
```diff
@@ -126,13 +126,13 @@ end
...
+    paperplots_baro_wave(energy_form(), moisture_model(), args...)
 
-paperplots_baro_wave(::Val{:ρθ}, ::Val{:dry}, args...) =
+paperplots_baro_wave(::ρθForm, ::DryModel, args...) =
     paperplots_baro_wave_ρθ(args...)
-paperplots_baro_wave(::Val{:ρe}, ::Val{:dry}, args...) =
+paperplots_baro_wave(::ρeForm, ::DryModel, args...) =
     paperplots_baro_wave_ρe(args...)
-paperplots_baro_wave(::Val{:ρe}, ::Val{:equil}, args...) =
+paperplots_baro_wave(::ρeForm, ::EquilModel, args...) =
```
jiahe23, line 135: Can we call them `PotentialTemperature`, `TotalEnergy` instead of `ρθForm`, `ρeForm`

jiahe23: And `EquilMoistModel`

dennisYatunin: And maybe `EnergyVariable` instead of `EnergyForm`?

examples/hybrid/types.jl
```diff
@@ -0,0 +1,72 @@
+abstract type AbstractMoistureModel end
+struct DryModel <: AbstractMoistureModel end
+struct EquilModel <: AbstractMoistureModel end
+struct NonEquilModel <: AbstractMoistureModel end
```
jiahe23, line 4: How about `EquilMoistModel`, `NonEquilMoistModel`?

examples/hybrid/types.jl
```diff
@@ -0,0 +1,72 @@
...
+        EquilModel()
+    elseif moisture_name == "nonequil"
+        NonEquilModel()
+    end
+end
+
+abstract type AbstractEnergyFormulation end
+struct ρθForm <: AbstractEnergyFormulation end
+struct ρeForm <: AbstractEnergyFormulation end
+struct ρe_intForm <: AbstractEnergyFormulation end
```
jiahe23, line 21: I'd prefer `PotentialTemperature`, `TotalEnergy`, `InternalEnergy`

examples/hybrid/define_post_processing.jl
charleskawczynski: Sure, no problem. I kind of like the `ρθForm` and `ρeForm`, just because `PotentialTemperature` could cause confusion with `θ` itself. But I don't have strong opinions on names, here.

examples/hybrid/types.jl
charleskawczynski: Definitely better, thanks!

jiahe23:
I am working on have #438 merged in soon. Once done, the held_suarez paper plots will need changes accordingly.

charleskawczynski:
> I am working on have #438 merged in soon. Once done, the held_suarez paper plots will need changes accordingly.

Sounds good, I'll rebase after #438 merges, I think I'd like to merge #445 before this.

charleskawczynski:
Waiting for #447 to merge since I'm getting a formatter error..

examples/hybrid/define_post_processing.jl
charleskawczynski: I went with @jiahe23's initial suggestion+ `InternalEnergy`

PR #443, opened 2022-05-11
Disable julia depot path

charleskawczynski:
We're hitting the julia depot path-related race condition too frequently. Let's disable it temporarily.

cc @szy21, @simonbyrne


simonbyrne:
I have an idea for an alternative solution: let me open a quick PR

charleskawczynski:
Superseded by #444.

PR #444, opened 2022-05-11
Increase JULIA_MAX_NUM_PRECOMPILE_FILES

simonbyrne:
Alternative to #443 

Set it smaller for initialization, and larger for actual running of jobs
See https://docs.julialang.org/en/v1.9-dev/manual/environment-variables/#JULIA_MAX_NUM_PRECOMPILE_FILES


charleskawczynski (approved):
Certainly worth a shot!

It looks like this is only in the julia 1.9 dev docs, though, will this still work?

PR #446, opened 2022-05-12
change default model top for non baroclinic wave examples

szy21:
Note that this will increase ci time by a factor of two. We can reduce t_end if needed. Any thoughts?


jiahe23 (approved):

szy21:
Alright I decided to use a slightly weaker stretching so that we don't need to have too many levels and the ci won't take too much longer.

szy21:
I'll need @charleskawczynski 's help on the regression test:) https://buildkite.com/clima/climaatmos-ci/builds/1017#_

szy21:
The results look fine, although we need to change some `clim` to make the plots more useful. I'll go ahead and merge this and we can make better plots later.

PR #447, opened 2022-05-12
Upgrade and apply formatter

charleskawczynski:
We're using an old version of the formatter, and running into [this issue](https://github.com/domluna/JuliaFormatter.jl/issues/575). This PR
 - 1) upgrades our formatter compat to the latest version 0.22
 - 2) Applies the formatter
 - 3) Changes the margin to 120

Any opposition to 3)?

EDIT: I'll just apply the changes that are necessary to fix CI

### DIFF
--- a/.dev/Manifest.toml
+++ b/.dev/Manifest.toml
@@ -1,32 +1,104 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.7.0"
+julia_version = "1.7.2"
 manifest_format = "2.0"
+
+[[deps.ArgTools]]
+uuid = "0dad84c5-d112-42e6-8d28-ef12dabb789f"
+
+[[deps.Artifacts]]
+uuid = "56f22d72-fd6d-98f1-02f0-08ddc0907c33"
 
 [[deps.Base64]]
 uuid = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
 
 [[deps.CSTParser]]
 deps = ["Tokenize"]
-git-tree-sha1 = "60e9121d9ea044c30a04397e59b00c5d9eb826ee"
+git-tree-sha1 = "b66abc140f8b90a1d6bc7bfad5c80070f8c1ddc6"
 uuid = "00ebfdb7-1f24-5e51-bd34-a7502290713f"
-version = "2.5.0"
+version = "3.3.3"
+
+[[deps.CommonMark]]
+deps = ["Crayons", "JSON", "URIs"]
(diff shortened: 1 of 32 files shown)


haakon-e:
Slightly on the side of this PR, but would it be useful if we check in the `.dev/Manifest.toml` in the [CEDMF](https://github.com/CliMA/CalibrateEDMF.jl/tree/main/.dev) and [TC](https://github.com/CliMA/TurbulenceConvection.jl/tree/main/.dev) repos too?

.dev/Manifest.toml
```diff
@@ -1,32 +1,104 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.7.0"
+julia_version = "1.7.2"
```
haakon-e, line 3: aren't we on 1.7.0 across CliMA?

charleskawczynski: `] up`ing can result in different `julia_version`s, so this tends to switch back-and-forth, depending on who `] up`s and commits manifest changes.

charleskawczynski:
> Slightly on the side of this PR, but would it be useful if we check in the `.dev/Manifest.toml` in the [CEDMF](https://github.com/CliMA/CalibrateEDMF.jl/tree/main/.dev) and [TC](https://github.com/CliMA/TurbulenceConvection.jl/tree/main/.dev) repos too?
> 
> For context; I had a problem a few weeks back that I was using an older version of the formatter -- which resulted in different formatting for certain stuff.

Yeah, perhaps we should check in the manifest in TC. But, it's probably not a major issue since there's only one dependency and we specify the [compat entry](https://github.com/CliMA/TurbulenceConvection.jl/blob/main/.dev/Project.toml#L5). That said, we previously specified version 0.3 and the patch upgrade broke, so not a bad idea for better reproducibility.

.dev/Manifest.toml
haakon-e: so, in general, it shouldn't matter whether some people run projects on 1.7.0 or 1.7.2?

charleskawczynski: Good question, I’m not sure what bug fixes came in 1.71/1.7.2. I’d certainly recommend that everyone updates to the latest version

haakon-e: Oh interesting. I have been under the impression that it's really important we use the exact same version of Julia, so I have taken extra care to use 1.7.0 in all my environments, since that's the [current version](https://github.com/CliMA/CalibrateEDMF.jl/blob/main/experiments/SCT2_benchmark/global_parallel/ekp_par_calibration.sbatch#L21) in all the CEDMF pipelines.

Would it be worth updating those slurm scripts to 1.7.2?

PR #452, opened 2022-05-12
add high_top command line argument

szy21:
Currently, the model easily becomes unstable when the model top is high. Increasing vertical levels may help, but that makes the turnaround slow. This PR adds a command line argument `high_top` so that we can move forward while trying to fix this.

The default of high_top is false, and the model top will be 30km with 10 levels. When setting high_top to true, the model top will be 45km with 25 levels. We run one `sphere_held_suarez_rhoe_equilmoist_hightop` in ci just to make sure it runs, so no regression test is needed. We run two experimental long runs`longrun_hs_rhoe_hightop` and `longrun_hs_rhoe_equil_hightop`. It's ok if they fail, we'll keep looking at it.

Please let me know what you think!


szy21:
Hmm, the git difference for pipeline.yml under longrun looks strange. But I renamed some long runs, and that's why.

jiahe23 (approved):

charleskawczynski:
Looks fine to me, but it be better to just add `z_elem` and `z_max`? `z_elem` is getting added in #439 already.

szy21:
> Looks fine to me, but it be better to just add `z_elem` and `z_max`? `z_elem` is getting added in #439 already.

I'm ok with `z_max`, but I still need a default `z_stretch` that is dependent on `z_max`, which makes me think there is not too much difference between `z_max` and `high_top`?

charleskawczynski:
> I'm ok with `z_max`, but I still need a default `z_stretch` that is dependent on `z_max`, which makes me think there is not too much difference between `z_max` and `high_top`?

It looks like we could use

```julia
    z_max = parse_arg(parsed_args, "z_max", FT(30e3))
    z_elem = parse_arg(parsed_args, "z_elem", 10)
    z_stretch = Meshes.GeneralizedExponentialStretching(FT(500), FT(5000))
```

The only reason I'd prefer this over using `high_top` is that passing `z_max` provides more flexibility. That said, I'm fine with the proposed changes. Are we intending to change `z_max` for some of the cases?

charleskawczynski (approved):

PR #454, opened 2022-05-12
Add surface temperature as default cache variable

### LINKED ISSUE #451: Centralize sea surface temperature function
Right now, the sea surface temperature is hardcoded in surface fluxes (boundary condition) and radiation. It would be good to centralize it. This will also be helpful when we move to more realistic sea surface temperatures.

dennisYatunin:
This PR puts `T_sfc` in the default cache and uses it to compute `vertical_diffusion_boundary_layer_tendency!` and `rrtmgp_model_tendency!`. The value of `T_sfc` is `29 * exp(-lat_sfc^2 / (2 * 26^2)) + 271`, where `lat_sfc` is the latitude of points on the surface in a spherical space and 0 degrees in a flat space.

Closes #451

### DIFF
--- a/examples/hybrid/driver.jl
+++ b/examples/hybrid/driver.jl
@@ -146,7 +146,7 @@ ode_algorithm = OrdinaryDiffEq.Rosenbrock23
 
 additional_callbacks = if !isnothing(radiation_model())
     # TODO: better if-else criteria?
-    dt_rad = parsed_args["config"] == "column" ? dt : FT(6 * 60 * 60)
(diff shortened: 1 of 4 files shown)


szy21 (approved):

szy21:
Hmm why would this break the simulation? `29 * exp(-lat^2 / (2 * 26^2)) + 271` is not the same as `29 * exp(-(lat / 26)^2 / 2) + 271`?

szy21:
It seems fine when I run `sphere_aquaplanet_rhoe_equilmoist_gray` locally. I'm using the same command as the one in pipeline.yml except for adding `--threads=8`

PR #456, opened 2022-05-14
Add Modeling Interface Document

dennisYatunin:
This PR adds a design document for the ClimaAtmos modeling interface, per Tapio's request. The document is still in its early stages, but I'll be adding to it over the next few weeks whenever I have time.


bischtob:
So far so good.

PR #457, opened 2022-05-14
Remove reduntant @timev for distributed timer.

sriharshakandala:


charleskawczynski:
I might recommend grabbing elapsed time from `@timev`, since it also prints out allocations, gc calls etc., in addition to wall time. Seeing allocations and number of gc calls is occasionally helpful.

sriharshakandala:
> I might recommend grabbing elapsed time from `@timev`, since it also prints out allocations, gc calls etc., in addition to wall time. Seeing allocations and number of gc calls is occasionally helpful.

We still use @timev for serial case.
For distributed case, and @timev printout from each process generates a clunky output!

charleskawczynski:
> > I might recommend grabbing elapsed time from `@timev`, since it also prints out allocations, gc calls etc., in addition to wall time. Seeing allocations and number of gc calls is occasionally helpful.
> 
> We still use @Timev for serial case. For distributed case, a @Timev printout from each process generates a clunky output!

Ah, ok, sounds good!

Timev:
Guys, I'm not going to commit anything here, please stop tagging me

sriharshakandala:
> 

Apologies for the accidental tagging. The conversation was about a macro. All references were removed.

PR #459, opened 2022-05-14
change upwinding default to none

szy21:


charleskawczynski (approved):

charleskawczynski (commented):
Let's rebase/squash this first

szy21:
> Let's rebase/squash this first

Yes for some reason I messed up the git history. It should be good now.

charleskawczynski:
I just realized, can't we also increase the timestep for central differencing?

szy21:
> I just realized, can't we also increase the timestep for central differencing?

We have been using 400s for central for dry cases and I remember that’s close to the largest dt we can have. For moist cases we have been using 200s, but I haven’t tried to push it with central. Jia is running some simulations, we’ll see if we can increase it.
 
 I did forget to change the dt back to 400s for the dry case with high top in the longrun though. Could you change it in your PR?

PR #462, opened 2022-05-14
Fix overwriting CL arguments

charleskawczynski:
This PR fixes the CL arguments that are overwritten, and should fix our incorrectly reported scaling results.

### DIFF
--- a/examples/hybrid/cli_options.jl
+++ b/examples/hybrid/cli_options.jl
@@ -36,10 +36,6 @@ function parse_commandline()
         "--forcing"
         help = "Forcing [`nothing` (default), `held_suarez`]"
         arg_type = String
-        "--high_top"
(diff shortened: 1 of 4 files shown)


szy21:
We need to change the pipeline.yml for longrun as well:)

examples/hybrid/cli_options.jl
```diff
@@ -93,6 +89,10 @@ function parse_commandline()
         help = "number of vertical elements"
         arg_type = Int
         default = 10
+        "--z_max"
+        help = "Model top height. Default: 45km"
```
szy21, line 93:
```suggestion
        help = "Model top height. Default: 30km"
```
:)

szy21 (approved):

PR #464, opened 2022-05-14
change gray radiation parameters

### LINKED ISSUE #463: Change gray radiation parameters
Let's make the default gray radiation parameters as close as those in this [paper](https://journals.ametsoc.org/view/journals/clim/21/15/2007jcli2065.1.xml).

szy21:
This PR changes gray radiation parameters to match those in this [paper](https://journals.ametsoc.org/view/journals/clim/21/15/2007jcli2065.1.xml). There is one small difference. The paper requires different optical depths for downward (0.22) and upward (0) shortwave radiation. This is not easy to do with the current code, so I set the optical depth for shortwave to be 0 altogether. It should be fine, because some previous version of the model in the paper also assume the optical depth for shortwave to be 0.

Closes #463

### DIFF
--- a/examples/RRTMGPInterface.jl
+++ b/examples/RRTMGPInterface.jl
@@ -1178,7 +1178,7 @@ function compute_optical_props_kernel!(
         p_lev[glay, gcol],
         p_lev[glay + 1, gcol],
         p_lev[1, gcol],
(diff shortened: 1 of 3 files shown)


bischtob (approved):
LGTM@!

PR #465, opened 2022-05-16
Add Nsight profiler to scaling pipeline.

sriharshakandala:


.buildkite/scaling/pipeline.yml
```diff
@@ -35,10 +35,13 @@ steps:
     steps:
 
     - label: ":computer: MPI Held-Suarez scaling test(ρe) - (2) processes"
-      command: "mpiexec julia --color=yes --project=examples examples/hybrid/driver.jl --job_id sphere_held_suarez_rhoe --forcing held_suarez --enable_threading false --upwinding none --t_end 1hours --dt 100secs --z_elem 40 --h_elem 16 --kappa_4 1e15" 
+      command:
+        - "module load cuda/11.3 nsight-systems/2022.2.1"
+        - "nsys profile --trace=nvtx,mpi --mpi-impl=mpich --output=sphere_held_suarez_rhoe/report.%q{NPROCS} mpiexec julia --color=yes --project=examples examples/hybrid/driver.jl --job_id sphere_held_suarez_rhoe --forcing held_suarez --enable_threading false --upwinding none --t_end 1hours --dt 100secs --z_elem 40 --h_elem 16 --kappa_4 1e15"
```
charleskawczynski, line 40: Shouldn't `--output` directory match the artifact path? This needs to be unique from the other artifact paths to avoid uploading data from both runs into both jobs

charleskawczynski: Same for other jobs below

charleskawczynski:
Can we add a description of what this does, since it requires manual builds?

.buildkite/scaling/pipeline.yml
charleskawczynski: Also, the `job_id` cannot match the job ID for any other buildkite job, otherwise we'll overwrite the data..

sriharshakandala: We want to write all scaling data to the same folder and turn off any other output! This way, there would not be any overwriting. Nprocesses is appended to the filename to distinguish between the simulations

sriharshakandala: Alternately, I can append Nprocs to job_id if this is preferred!

charleskawczynski: Ah, ok, makes sense. I'm fine with that if it's somehow easier.

`--output` is NVTX specific, correct? (ClimaAtmos' output directory is specified via `--output_dir`).

charleskawczynski: I personally think it's nice to use `job_id` and set `artifacts_path` to be exactly the same string (to avoid any ambiguity), but I don't have a strong preference and this is fine, too.

sriharshakandala: Yes, --output is NVTX specific

simonbyrne:
we need to modify the driver so that it does 1 step before benchmarking: want to do it in this PR, or another?

charleskawczynski:
Should any of the config changes be applied to the other pipelines? The scaling seems to take a short time, should we just add this to our CI?

simonbyrne (approved):

PR #480, opened 2022-05-20
fix save to disk and remap to nc

jiahe23:
This PR
- fixes a bug in save to disk function for dry case;
- adds RH into moist simulation outputs;
- modifies remap_pipeline to add diagnostics into nc files;
- rename cloud_water -> cloud_liquid.

### DIFF
--- a/examples/hybrid/driver.jl
+++ b/examples/hybrid/driver.jl
@@ -307,9 +307,14 @@ mkpath(output_dir)
 
 function make_save_to_disk_func(output_dir, p, is_distributed)
     function save_to_disk_func(integrator)
-        (; ᶜts, ᶜp, ᶜS_ρq_tot, params, ᶜK, ᶜΦ) = p
-
         Y = integrator.u
+
+        if :ρq_tot in propertynames(Y.c)
+            (; ᶜts, ᶜp, ᶜS_ρq_tot, params, ᶜK, ᶜΦ) = p
+        else
+            (; ᶜts, ᶜp, params, ᶜK, ᶜΦ) = p
+        end
+
         ᶜuₕ = Y.c.uₕ
         ᶠw = Y.f.w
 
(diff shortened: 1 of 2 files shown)


examples/hybrid/driver.jl
```diff
@@ -328,24 +333,26 @@ function make_save_to_disk_func(output_dir, p, is_distributed)
...
         # vorticity 
         curl_uh = @. curlₕ(Y.c.uₕ)
         ᶜvort = Geometry.WVector.(curl_uh)
         Spaces.weighted_dss!(ᶜvort)
 
-        # cloudwater and watervapor for moist simulation
+        # cloudwater, watervapor, precipitation, and RH for moist simulation
         if :ρq_tot in propertynames(Y.c)
             ᶜq = @. TD.PhasePartition(params, ᶜts)
             ᶜcloud_water = @. ᶜq.liq
```
szy21, line 344:
```suggestion
            ᶜcloud_liquid = @. ᶜq.liq
```

szy21 (approved):

examples/hybrid/sphere/remap_pipeline.jl
```diff
@@ -17,6 +17,12 @@ else
     error("ENV[\"THERMO_VAR\"] require (\"e_tot\" or \"theta\")")
 end
 
+if haskey(ENV, "MOIST_MODE")
+    moist_mode = ENV["MOIST_MODE"]
+else
+    error("ENV[\"MOIST_MODE\"] require (\"dry\" or \"equil\")")
+end
```
charleskawczynski, line 24: Could we switch these to kwargs in `remap2latlon` and call this in our CI?

charleskawczynski: (not a blocker, it'd just be nice to see this in action on CI!)

jiahe23: > Could we switch these to kwargs in `remap2latlon` and call this in our CI?

Yes, that will be great! I am actually thinking of a way to remap in the callback setup so that we could have a proper nc outputs while running the simulation.

PR #482, opened 2022-05-23
add viscous sponge

szy21:
This PR adds a viscous sponge. There will be a following up PR that changes the sponge to a command line argument.


examples/hybrid/driver.jl
```diff
@@ -82,12 +82,14 @@ show_progress_bar = isinteractive()
 additional_solver_kwargs = () # e.g., abstol and reltol
 test_implicit_solver = false # makes solver extremely slow when set to `true`
 
-const sponge = false
+const rayleigh_sponge = false
+const viscous_sponge = true
```
szy21, line 86: change it to false before merging

szy21:
@akshaysridhar This is ready I think. The results looks fine to me: https://buildkite.com/clima/climaatmos-ci/builds/1258#b38991b1-14a4-4b23-89d4-5afb80053413

akshaysridhar (approved):
If the changes to the buildkite results are acceptable let's merge this.

szy21:
> If the changes to the buildkite results are acceptable let's merge this.

Thanks. The buildkite results won't change in the end because the default is false for sponge. I don't know what would be the best way to test it in ci. Maybe I will add a run in ci with sponge after #483, or a long run with sponge.

PR #483, opened 2022-05-24
Sponges via command-line arguments 

akshaysridhar:
Command line arguments for `rayleigh` and `viscous` sponges. Follows PR482.


examples/hybrid/driver.jl
```diff
@@ -82,12 +89,11 @@ show_progress_bar = isinteractive()
 additional_solver_kwargs = () # e.g., abstol and reltol
 test_implicit_solver = false # makes solver extremely slow when set to `true`
 
-const sponge = false
-
 # TODO: flip order so that NamedTuple() is fallback.
 additional_cache(Y, params, dt; use_tempest_mode = false) = merge(
     hyperdiffusion_cache(Y; κ₄ = FT(κ₄), use_tempest_mode),
-    sponge ? rayleigh_sponge_cache(Y, dt) : NamedTuple(),
+    rayleigh_sponge ? rayleigh_sponge_cache(Y, dt; zd_rayleigh) : NamedTuple(),
```
szy21, line 95:
```suggestion
    rayleigh_sponge ? rayleigh_sponge_cache(Y, dt; zd_rayleigh = FT(zd_rayleigh)) : NamedTuple(),
```

examples/hybrid/driver.jl
```diff
@@ -82,12 +89,11 @@ show_progress_bar = isinteractive()
...
+    viscous_sponge ? viscous_sponge_cache(Y; zd_viscous, κ₂=κ₂_sponge) : NamedTuple(),
```
szy21, line 96:
```suggestion
    viscous_sponge ? viscous_sponge_cache(Y; zd_viscous = FT(zd_viscous), κ₂ = FT(κ₂_sponge)) : NamedTuple(),
```

szy21 (approved):

charleskawczynski:
Do we need to be able to run them all together?

charleskawczynski:
I think that the other way we could do this, to reduce the number of CL args, would be to specify the sponge name, and the sponge coefficient/intensity? This is fine too, though

szy21:
> Do we need to be able to run them all together?

I'm not sure if we need both sponges for the production run in the end, but for now I think it would be good to keep the option to add both.

PR #490, opened 2022-05-25
switch to ARS343

simonbyrne:
This switches the timestepper to using ClimaTimeSteppers ARS343 stepper. 

Note the change in interface: the remaining tendency is now computed as a "step" function. 

Unfortunately it appears to be crashing, I might need some help with this one.


examples/hybrid/sphere/baroclinic_wave_utilities.jl
```diff
@@ -430,7 +430,7 @@ function vertical_diffusion_boundary_layer_tendency!(Yₜ, Y, p, t)
             top = Operators.SetValue(Geometry.WVector(FT(0))),
             bottom = Operators.SetValue(dif_flux_energy),
         )
-        @. Yₜ.c.ρe += ᶜdivᵥ(ᶠK_E * ᶠinterp(ᶜρ) * ᶠgradᵥ((Y.c.ρe + ᶜp) / ᶜρ))
+        @. Yx.c.ρe += dt*ᶜdivᵥ(ᶠK_E * ᶠinterp(ᶜρ) * ᶠgradᵥ((Y.c.ρe + ᶜp) / ᶜρ))
```
valeriabarra, line 433: the argument of this function is `Yt` here but you're using `Yx`?

simonbyrne: good catch!

examples/hybrid/sphere/baroclinic_wave_utilities.jl
```diff
@@ -396,7 +396,7 @@ function sensible_heat_flux_ρe_int(param_set, Ch, sc, scheme)
     return -ρ_sfc * Ch * SF.windspeed(sc) * (cp_m * ΔT) - (hd_sfc) * E
 end
 
-function vertical_diffusion_boundary_layer_tendency!(Yₜ, Y, p, t)
+function vertical_diffusion_boundary_layer_step!(Yt, Y, p, t, dt)
```
simonbyrne, line 399:
```suggestion
function vertical_diffusion_boundary_layer_step!(Yx, Y, p, t, dt)
```

PR #495, opened 2022-05-27
change initial condition for sphere examples

szy21:
This PR changes the initial condition for sphere examples to isothermal except for baroclinic wave, as suggested by Tapio.


jiahe23 (approved):
Would you like to add one example using the new initial condition? Otherwise looks good to me.

PR #504, opened 2022-05-30
change initial condition to DecayingTemperatureProfile

### LINKED ISSUE #502: Use a more realistic initial condition
Currently, the initial condition for sphere configuration (except for baroclinic wave) is isothermal. It would be better to use something like `DecayingTemperatureProfile` in `Thermodynamics`.

szy21:
Closes #502.

@charleskawczynski Is this the right way to use Thermodynamics?


charleskawczynski (approved):
Yep, so long the numbers are reasonable (`H_t` is the one I'm least familiar with off hand).

szy21:
> Yep, so long the numbers are reasonable (`H_t` is the one I'm least familiar with off hand).

Thanks! `H_t` is the scale height of atmosphere, which is around 8 km.

PR #506, opened 2022-05-31
add option for IMEXEuler timestepper

simonbyrne:
cc: @dennisYatunin 

Should add an integration test: any suggestions?

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -133,6 +133,10 @@ steps:
   - group: "Sphere Examples"
     steps:
 
+      - label: ":computer: baroclinic wave (ρe_tot) IMEXEuler"
+        command: "julia --color=yes --project=examples examples/hybrid/driver.jl --t_end 2days --dt 60secs --ode_algo IMEXEuler --job_id sphere_baroclinic_wave_rhoe_IMEXEuler"
+        artifact_paths: "sphere_baroclinic_wave_rhoe_IMEXEuler/*"
+
       - label: ":computer: baroclinic wave (ρe_tot) Float64"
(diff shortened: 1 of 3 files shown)


dennisYatunin:
For the integration test, maybe just the default baroclinic wave with total energy example, using the IMEXEuler method with a slightly smaller tilmestep? It'll be good to test more than a single ODE solver method in the pipeline.

simonbyrne:
@dennisYatunin can you please take this on, and check it is reasonable?

It turns out the IMEXEuler is actually equivalent to an ARS scheme, so once we have this working we can compare them directly.

dennisYatunin:
Ok, I've added an IMEXEuler example to the pipeline. Unfortunately, I can't get it to run with a single Newton iteration because OrdinaryDiffEq's convergence criterion for Newton's method looks like
```
if (iter == 1 && ndz < 1e-5) || (iter > 1 && (η >= zero(η) && η * ndz < κ))
      nlsolver.status = Convergence
      nlsolver.nfails = 0
      break
end
```
This means that, unless `ndz < 1e-5` on the first Newton iteration (which is unlikely to happen for us because the values of `ndz` I observed were greater than 1), we need to take at least 2 Newton iterations. I ensured that we never need to take more than 2 iterations by setting `κ = Inf`.

It does appear that IMEXEuler with 2 Newton iterations is not very well-suited for our problem, as I needed to reduce the timestep from 400 seconds to 60 seconds (80 seconds did not work, but 60 did). Nevertheless, there is now a working IMEXEuler example in our pipeline, though it takes quite a while to run. I guess the next step would be to implement IMEXEuler with 2 Newton iterations in ClimaTimeSteppers and compare the results.

PR #508, opened 2022-06-01
add surface properties to cache

### LINKED ISSUE #493: Add some surface properties to cache
To allow access to / modification of surface properties, such as albedo and emissivity. This is needed by the coupler.

szy21:
Closes #493 
co-authored-by: @dennisYatunin

### DIFF
--- a/examples/hybrid/radiation_utilities.jl
+++ b/examples/hybrid/radiation_utilities.jl
@@ -100,7 +100,7 @@ function rrtmgp_model_cache(
             @. 1360 * (1 + FT(1.2) / 4 * (1 - 3 * sind(latitude)^2)) /
                (4 * cos(solar_zenith_angle))
     else
-        solar_zenith_angle = weighted_irradiance = NaN # initialized in tendency
+        solar_zenith_angle = weighted_irradiance = NaN # initialized in callback
     end
 
     if idealized_h2o && radiation_mode isa RRTMGPI.GrayRadiation
@@ -119,23 +119,19 @@ function rrtmgp_model_cache(
         center_pressure = NaN, # initialized in callback
         center_temperature = NaN, # initialized in callback
         surface_temperature = NaN, # initialized in callback
-        surface_emissivity = FT(1),
-        direct_sw_surface_albedo = FT(0.38),
-        diffuse_sw_surface_albedo = FT(0.38),
+        surface_emissivity = 1,
+        direct_sw_surface_albedo = 0.38,
+        diffuse_sw_surface_albedo = 0.38,
         solar_zenith_angle,
         weighted_irradiance,
         kwargs...,
     )
     close(input_data)
     return (;
-        ᶜT = similar(Y.c, FT),
-        ᶜvmr_h2o = similar(Y.c, FT),
-        insolation_tuple = similar(Spaces.level(Y.c, 1), Tuple{FT, FT, FT}),
-        zenith_angle = similar(Spaces.level(Y.c, 1), FT),
-        weighted_irradiance = similar(Spaces.level(Y.c, 1), FT),
-        ᶠradiation_flux = similar(Y.f, Geometry.WVector{FT}),
         idealized_insolation,
         idealized_h2o,
+        insolation_tuple = similar(Spaces.level(Y.c, 1), Tuple{FT, FT, FT}),
+        ᶠradiation_flux = similar(Y.f, Geometry.WVector{FT}),
         rrtmgp_model,
     )
 end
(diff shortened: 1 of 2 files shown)


szy21:
@LenkaNovak @kmdeck surface albedo and emissivity will be in cache after this PR is merged

dennisYatunin (approved):

kmdeck:
> @LenkaNovak @kmdeck surface albedo and emissivity will be in cache after this PR is merged

Precip next? 😬 

On our end, I think we should merge the current PR without radiation (passing conservation) and then have land cache albedo as well and pass that to the coupler.

szy21:
> Precip next? 😬

Let me take a look :)

> On our end, I think we should merge the current PR without radiation (passing conservation) and then have land cache albedo as well and pass that to the coupler.

Sounds good!

szy21:
@dennisYatunin This actually doesn't work (see the error [here](https://buildkite.com/clima/climaatmos-ci/builds/1380#01811cc0-8414-4458-9b1b-af19ac562f82)) because surface emissivity and albedo have the shape (nband, ncol). Is there a way to fix this (we can simply assume the values for different bands are the same for now)?

szy21:
actually @kmdeck will the land have different emissivity and albedo for different rrtmg bands (I would guess so)?

kmdeck:
> actually @kmdeck will the land have different emissivity and albedo for different rrtmg bands (I would guess so)?

eventually yes the land albedo will be a function of wavelength, though I dont know which bands yet, it depends on our model. The simplest would be a single value (Manabe), then next most complex would be 2 bandxdiffuse/direct, eventually we will have a multi-band model but the bands may not align with rrtmg/we may need to map between.

szy21:
> > actually @kmdeck will the land have different emissivity and albedo for different rrtmg bands (I would guess so)?
> 
> eventually yes the land albedo will be a function of wavelength, though I dont know which bands yet, it depends on our model. The simplest would be a single value (Manabe), then next most complex would be 2 bandxdiffuse/direct, eventually we will have a multi-band model but the bands may not align with rrtmg/we may need to map between.

So if we make it a constant for different wavelengths, it would be good enough for now?

dennisYatunin:
I modified this PR to bring it more in line with my original vision for the RRTMGP interface. Since we are only going to use the surface properties for RRTMGP, we should only store them in the `rrtmgp_model` object, rather than also storing them separately in the cache. This PR now adds an `array2field` function that makes it easier to set values in the `rrtmgp_model`. Specifically, if you have a `Field` of surface emissivities, you can do either `rrtmgp_model.surface_emissivity .= field2array(surface_emissivity_field)` or `array2field(rrtmgp_model.surface_emissivity, surface_space) .= surface_emissivity_field`. Moreover, if you do not want to separately allocate a `Field` for the surface emissivity, you can now do `array2field(rrtmgp_model.surface_emissivity, surface_space) .= @. <broadcast expression for surface emissivity field>`.

Note that the surface temperature must still be stored separately in the cache because it is used by more than just RRTMGP (it is also used in the vertical diffusion tendency).

kmdeck (commented):
not seeing any unit tests of array2field. Would be good to add some tests to verify this is doing what is expected.

examples/hybrid/radiation_utilities.jl
```diff
@@ -119,23 +119,19 @@ function rrtmgp_model_cache(
         center_pressure = NaN, # initialized in callback
         center_temperature = NaN, # initialized in callback
         surface_temperature = NaN, # initialized in callback
-        surface_emissivity = FT(1),
-        direct_sw_surface_albedo = FT(0.38),
-        diffuse_sw_surface_albedo = FT(0.38),
+        surface_emissivity = 1,
```
kmdeck, line 122: How are these defined as floats here, but later as fields?

also, not seeing these defaults being used - later on, the values of 1 and 0.38 are used, but hardcoded.

examples/hybrid/staggered_nonhydrostatic_model.jl
```diff
@@ -121,6 +121,9 @@ function default_cache(Y, params, upwinding_mode)
     end
     ᶜf = @. Geometry.Contravariant3Vector(Geometry.WVector(ᶜf))
     T_sfc = @. 29 * exp(-lat_sfc^2 / (2 * 26^2)) + 271
+    direct_albedo_sfc = @. FT(0.38) + 0 * lat_sfc
```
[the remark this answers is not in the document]
kmdeck: should 0.38 be read from the rrtmg model parameter set?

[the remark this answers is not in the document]
szy21: Oh, we should remove these three lines. @dennisYatunin ?

examples/hybrid/radiation_utilities.jl
szy21: Yes we should remove these properties from the cache in the staggered_nonhydrostatic_model. Those are not being used. Here is where these properties are defined (hardcoded). I think we chose not to use parameter set for now, @dennisYatunin what is the reason?

kmdeck: And I guess in the coupled runs rrtmgp knows to look for albedo and emissivity in the cache?

szy21: Yes.

dennisYatunin: Since ClimaParameters is going to change significantly very soon, we should probably avoid adding things there unless absolutely necessary. I'm not sure if `surface_emissivity = 1` is a parameter that is urgently needed across CliMA, so it's probably fine to leave this hardcoded for now.

The reason these values are constants here is that I wrote the `RRTMGPModel` interface to be as user-friendly as possible. If the user specifies a constant, then the entire field is set to that constant. If the user specifies a 1D array for a 3D field, that array is used to set each column. The full set of possible inputs is described in the `RRTMGPModel` docstring.

In order to modify, the surface emissivity of each band from the coupler, you will probably want to do something like
```
number_of_bands = size(rrtmgp_model.surface_emissivity, 1)
for band in 1:number_of_bands
    surface_emissivity = RRTMGPI.array2field(rrtmgp_model.surface_emissivity[band, :], bottom_level_space)
    @. surface_emissivity = <broadcast expression for surface emissivity for current wavelength band>
end
```

PR #513, opened 2022-06-04
Add simple cloudy radiation example

dennisYatunin:
This PR adds support for all-sky radiation to `rrtmgp_model_callback`. Specifically, it allows for cloud properties to be computed based on the thermodynamic state, or, if `idealized_clouds == true`, it puts an icy cloud between 4km and 5km, and it puts a wet cloud between 1km and 1.5km.

### DIFF
--- a/examples/hybrid/radiation_utilities.jl
+++ b/examples/hybrid/radiation_utilities.jl
@@ -11,7 +11,18 @@ function rrtmgp_model_cache(
     bottom_extrapolation = RRTMGPI.SameAsInterpolation(),
     idealized_insolation = true,
     idealized_h2o = false,
+    idealized_clouds = false,
 )
+    if idealized_h2o && radiation_mode isa RRTMGPI.GrayRadiation
+        error("idealized_h2o can't be used with $radiation_mode")
+    end
+    if idealized_clouds && (
+        radiation_mode isa RRTMGPI.GrayRadiation ||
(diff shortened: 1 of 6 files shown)


examples/hybrid/radiation_utilities.jl
```diff
@@ -242,6 +284,31 @@ function rrtmgp_model_callback!(integrator)
...
+            axes(Y.c),
+        )
+        ᶜmask = RRTMGPI.array2field(
+            rrtmgp_model.center_cloud_boolean_mask,
+            axes(Y.c),
+        )
+        # multiply by 1000 to convert from kg/m^2 to g/m^2
+        @. ᶜlwp = 1000 * Y.c.ρ * TD.liquid_specific_humidity(params, ᶜts) * ᶜΔz
+        @. ᶜiwp = 1000 * Y.c.ρ * TD.ice_specific_humidity(params, ᶜts) * ᶜΔz
+        @. ᶜmask =
```
charleskawczynski, line 307: What is this condition? We should probably add this as a thermo function. We have `has_condensate`, but this looks like a different condition?

szy21: Oh! I made this up. We can use `has_condensate`, thanks! We may need a threshold at some point. I will let you know.

szy21 (approved):

PR #518, opened 2022-06-06
add clear sky radiation diagnostics

### LINKED ISSUE #515: add clear sky radiative flux diagnostics
After #513, we can add both clear and all sky radiative flux to diagnostics.

szy21:
Closes #515


dennisYatunin (approved):
Sounds good!

PR #521, opened 2022-06-08
Add NVTX ranges for dss calls and tendency functions.

sriharshakandala:
Add NVTX ranges for dss calls and tendency functions.


charleskawczynski (commented):
@simonbyrne, is this the only way to apply NVTX monitoring?

sriharshakandala:
> @sriharshakandala Would you be able to keep the implementation of performance monitoring machinery confined to ClimaCore? For example, in this case, could you allow `weighted_dss!` and related functions to accept a `monitor_nvtx` boolean flag, and move the code you wrote here to the definitions of those functions?

It looks like ClimaAtmos is defining its own tendency functions, etc.,  separate from ClimaCore!

charleskawczynski:
> It looks like ClimaAtmos is defining its own tendency functions, etc., separate from ClimaCore. We would want to monitor and tune all performance critical components of ClimaAtmos!

@dennisYatunin's suggestion would allow us to limit the boiler plate in this PR. I don't think anyone working on physics should have to jump around several performance monitoring code blocks.

Perhaps more importantly, why can't we just wrap `OrdinaryDiffEq.step!(integrator)` with an NVTX call like we do with the other performance monitoring? Wouldn't this still give us the breakdown we need?

simonbyrne:
We want to be able to monitor all parts, I asked @sriharshakandala to do DSS first, as there might be some opportunities there. If we want to make the code faster, we need to know where we're spending our time.

If the concern is the boiler-plate, we could add a macro? e.g.
```
@nvtx "name" color=red begin
   # do stuff
end
```

simonbyrne (approved):

simonbyrne:
> Perhaps more importantly, why can't we just wrap `OrdinaryDiffEq.step!(integrator)` with an NVTX call like we do with the other performance monitoring? Wouldn't this still give us the breakdown we need?

To answer this specifically, the aim is to figure out how much each component of `step!` takes.

PR #531, opened 2022-06-14
Upgrade deps, including latest TC version

charleskawczynski:
Upgrade deps, including latest TC version.


charleskawczynski:
@dennisYatunin, did SciMLBase update when you `] up`ed earlier? This PR is running into a bunch of ODE solver issues, could you take a look?

dennisYatunin:
I think it did, yeah

PR #548, opened 2022-06-20
Use new way of handling clima parameters

charleskawczynski:
This PR applies the new way of handling clima parameters:
 - Adds a package module, `Parameters`, where we define a struct specifically for this packages parameters.
 - The struct from this module is then used in all places where we originally called CLIMAParameter methods (e.g., `Planet.grav(param_set)` becomes `Parameters.grav(param_set)`, and I've added some aliases so that this is a bit more terse
 - Removes the use of CLIMAParameters in `src/` completely
 - Adds a folder, `parameters` to the top-level package, where a parameter-creating function was added, which constructs the struct instance that houses all of the necessary parameters are need (from this package, and its dependencies).


charleskawczynski:
I think I need to put this on pause to fix [ClimaCore#720](https://github.com/CliMA/ClimaCore.jl/issues/720).

PR #558, opened 2022-06-24
add coupled command line argument

szy21:
# PULL REQUEST

## Purpose and Content
Adds a coupled command line argument so that we use surface fluxes directly from the cache when running coupled simulations.

## Benefits and Risks
It is a quick but not elegant way to allow us to run coupled simulations. A more elegant way would be to have a dispatch on the vertical diffusion tendency, but we think it would be better to change it when we change the atmos interface.

## PR Checklist

### DIFF
--- a/examples/hybrid/driver.jl
+++ b/examples/hybrid/driver.jl
@@ -12,6 +12,7 @@ idealized_h2o = parsed_args["idealized_h2o"]
 idealized_insolation = parsed_args["idealized_insolation"]
 idealized_clouds = parsed_args["idealized_clouds"]
 vert_diff = parsed_args["vert_diff"]
+coupled = parsed_args["coupled"]
 hyperdiff = parsed_args["hyperdiff"]
 turbconv = parsed_args["turbconv"]
 h_elem = parsed_args["h_elem"]
@@ -124,7 +125,7 @@ additional_cache(Y, params, dt; use_tempest_mode = false) = merge(
         idealized_clouds,
     ),
     vert_diff ?
-    vertical_diffusion_boundary_layer_cache(Y; diffuse_momentum) :
+    vertical_diffusion_boundary_layer_cache(Y; diffuse_momentum, coupled) :
     NamedTuple(),
     (;
         tendency_knobs = (;
--- a/examples/hybrid/cli_options.jl
+++ b/examples/hybrid/cli_options.jl
@@ -40,6 +40,10 @@ function parse_commandline()
         help = "Vertical diffusion [`false` (default), `true`]"
         arg_type = Bool
         default = false
+        "--coupled"
+        help = "Coupled simulation [`false` (default), `true`]"
+        arg_type = Bool
+        default = false
         "--turbconv"
         help = "Turbulence convection scheme [`nothing` (default), `edmf`]"
         arg_type = String
--- a/examples/hybrid/sphere/baroclinic_wave_utilities.jl
+++ b/examples/hybrid/sphere/baroclinic_wave_utilities.jl
@@ -514,6 +514,7 @@ function vertical_diffusion_boundary_layer_cache(
     Cd = FT(0.0044),
     Ch = FT(0.0044),
     diffuse_momentum = true,
+    coupled = false,
 )
     ᶠz_a = similar(Y.f, FT)
     z_bottom = Spaces.level(Fields.coordinate_field(Y.c).z, 1)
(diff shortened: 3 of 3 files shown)


jiahe23 (approved):
lgtm

charleskawczynski:
I'd like to merge #548 first, if you don't mind.

szy21:
> I'd like to merge #548 first, if you don't mind.

Yes, that's fine. I don't think I need to change anything here?

charleskawczynski:
> Yes, that's fine. I don't think I need to change anything here after 548?

You'll need to rebase because some of the lines changed in this PR are indenting a bunch of lines, those lines will be changed in 548, so it'll result in git conflicts. It should be easy to fix here though, because all you'll need to do is accept main and re-indent with the `if` statements.

szy21:
> You'll need to rebase because some of the lines changed in this PR are indenting a bunch of lines, those lines will be changed in 548, so it'll result in git conflicts. It should be easy to fix here though, because all you'll need to do is accept main and re-indent with the `if` statements.

ok thanks! Do you plan to merge it today?

charleskawczynski:
> ok thanks! Do you plan to merge it today?

Yep, I think the last push should give me the new mse tables, I'll update and r+ (hopefully in 30 ish minutes?)

examples/hybrid/driver.jl
```diff
@@ -124,7 +125,7 @@ additional_cache(Y, params, dt; use_tempest_mode = false) = merge(
         idealized_clouds,
     ),
     vert_diff ?
-    vertical_diffusion_boundary_layer_cache(Y; diffuse_momentum) :
+    vertical_diffusion_boundary_layer_cache(Y; diffuse_momentum, coupled) :
```
kmdeck, line 128: Is this an initialization step? Does it need "coupled"?

kmdeck (commented):
The code in `tendency` function looks fine - but we dont need to compute surface fluxes unless we are not in coupled mode. As it is now, in coupled mode we would compute it 2x per step, once here (not used) and once in the coupler.

kmdeck (commented):
In general, when are cache variables updated by ClimaAtmos? Are we going to run into an issue where the coupler updates the cache with the correct boundary fluxes, but they are changed somewhere else by ClimaAtmos?

kmdeck (commented):
looks OK but I have some questions before we merge.

examples/hybrid/driver.jl
szy21: This function is only called once, at the beginning of the simulation.

szy21:
@kmdeck Right now everything we need in the tendency function (e.g. `coupled` in this PR) is stored in cache. All `cache` functions are called once at the beginning of the simulation. The cached variables may be changed (only) in the `tendency` functions. I think this PR will fix the issue of boundary fluxes.

I added another `if !coupled` before `flux_coefficients .= ...` which should solve the problem of calculating surface fluxes twice. I checked it multiple times because we have some other checks in the `vertical_diffusion_boundary_layer_tendency`. Actually now looking at it I think the coupled simulation wouldn't be correct if we use internal energy as a prognostic variable?

kmdeck:
> @kmdeck Right now everything we need in the tendency function (e.g. `coupled` in this PR) is stored in cache. All `cache` functions are called once at the beginning of the simulation. The cached variables may be changed (only) in the `tendency` functions. I think this PR will fix the issue of boundary fluxes.
> 
> I added another `if !coupled` before `flux_coefficients .= ...` which should solve the problem of calculating surface fluxes twice. I checked it multiple times because we have some other checks in the `vertical_diffusion_boundary_layer_tendency`. Actually now looking at it I think the coupled simulation wouldn't be correct if we use internal energy as a prognostic variable?

thanks! Looks good now. And thanks for explaining about when cached variables are changed.

Yes, youre right that the coupled simulation is only setting dif_flux_energy. Would it work if we also checked which variable is being used for energy, and compute the one expression vs the other depending on the result (SF.SHF vs sensible_heat_flux_ρe_int)? 

Does that make sense in terms of energy flux between land and atmos?

szy21:
> Yes, youre right that the coupled simulation is only setting dif_flux_energy. Would it work if we also checked which variable is being used for energy, and compute the one expression vs the other depending on the result (SF.SHF vs sensible_heat_flux_ρe_int)?
> 
> Does that make sense in terms of energy flux between land and atmos?

Do you mean checking the variable in the coupler and computing the surface fluxes based on which variable it is? Yes, I think that would work. But it seems we define `sensible_heat_flux_ρe_int` ourselves - it's probably not in SurfaceFluxes.jl yet?

PR #570, opened 2022-06-28
Try filter on moisture case

charleskawczynski:
This PR tries out a filter for the moist Held Suarez, I'm not sure if we've tried this, and I'm curious what happens here.

### DIFF
--- a/.buildkite/longruns/pipeline.yml
+++ b/.buildkite/longruns/pipeline.yml
@@ -54,6 +54,12 @@ steps:
         agents:
           slurm_cpus_per_task: 8
 
+      - label: ":computer: held suarez (ρe_tot) equilmoist filter"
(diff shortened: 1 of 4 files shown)


szy21 (approved):
My only major concern is that changing ρq_tot but not ρ is not physically consistent. But I think it's worth trying. Would you like to add it to the standard pipeline as well? If I remember correctly it breaks within 10 days.

charleskawczynski:
> My only major concern is that changing ρq_tot but not ρ is not physically consistent. But I think it's worth trying. Would you like to add it to the standard pipeline as well? If I remember correctly it breaks within 10 days.

Yeah, that makes sense. Sure, I'll add a test to the regular CI.

charleskawczynski:
It looks like applying this to the regular runs does have behavioral changes, which is interesting. I'm going to skip the regular CI for now since we're only testing.

PR #571, opened 2022-06-28
Fix some variables that should be constant

charleskawczynski:
JET found 654 possible errors. This PR fixes a few: some variables should be declared `const` to avoid runtime checks.

cc @simonbyrne


simonbyrne:
Good catch!

PR #583, opened 2022-07-06
Port TurbulenceConvection to ClimaAtmos

charleskawczynski:
There was a decision made in the recent EDMF meeting to port TurbulenceConvection into ClimaAtmos in order to smooth development of several pieces that need to be updated together:

 - Using consistent energy between grid mean and updraft variables
 - Using a fully compressible model
 - Using an implicit solver

These pieces cannot be (easily) done in TC alone because TC has a built-in anelastic assumption, which is incompatible with the energy formulation.

If what I've wrote is somehow inaccurate, please chime in, @yairchn, @tapios.

Once we merge this, we can add a note/warning in the TC readme indicating that development has been moved to this repo.

### DIFF
--- a/src/TurbulenceConvection/EDMF_Environment.jl
+++ b/src/TurbulenceConvection/EDMF_Environment.jl
@@ -1,0 +1,390 @@
+function microphysics(
+    ::SGSMean,
+    grid::Grid,
+    state::State,
+    edmf::EDMFModel,
+    precip_model::AbstractPrecipitationModel,
+    Δt::Real,
+    param_set::APS,
+)
+    FT = float_type(state)
+    thermo_params = TCP.thermodynamics_params(param_set)
+    tendencies_pr = center_tendencies_precipitation(state)
+    aux_en = center_aux_environment(state)
+    prog_pr = center_prog_precipitation(state)
+    prog_gm = center_prog_grid_mean(state)
+    aux_gm = center_aux_grid_mean(state)
+    ts_env = center_aux_environment(state).ts
+    p_c = aux_gm.p
+    ρ_c = prog_gm.ρ
+    aux_en_sat = aux_en.sat
+    aux_en_unsat = aux_en.unsat
+    precip_fraction = compute_precip_fraction(edmf, state)
+
+    @inbounds for k in real_center_indices(grid)
+        # condensation
+        ts = ts_env[k]
+        # autoconversion and accretion
+        mph = precipitation_formation(
+            param_set,
+            precip_model,
+            prog_pr.q_rai[k],
+            prog_pr.q_sno[k],
+            aux_en.area[k],
+            ρ_c[k],
+            FT(grid.zc[k].z),
+            Δt,
+            ts,
+            precip_fraction,
+        )
+
+        # update_sat_unsat
+        if TD.has_condensate(thermo_params, ts)
+            aux_en.cloud_fraction[k] = 1
+            aux_en_sat.θ_dry[k] = TD.dry_pottemp(thermo_params, ts)
+            aux_en_sat.θ_liq_ice[k] = TD.liquid_ice_pottemp(thermo_params, ts)
+            aux_en_sat.T[k] = TD.air_temperature(thermo_params, ts)
+            aux_en_sat.q_tot[k] = TD.total_specific_humidity(thermo_params, ts)
+            aux_en_sat.q_vap[k] = TD.vapor_specific_humidity(thermo_params, ts)
+        else
+            aux_en.cloud_fraction[k] = 0
+            aux_en_unsat.θ_dry[k] = TD.dry_pottemp(thermo_params, ts)
+            aux_en_unsat.θ_virt[k] = TD.virtual_pottemp(thermo_params, ts)
+            aux_en_unsat.q_tot[k] = TD.total_specific_humidity(thermo_params, ts)
+        end
+
+        # update_env_precip_tendencies
+        # TODO: move ..._tendency_precip_formation to diagnostics
+        aux_en.qt_tendency_precip_formation[k] = mph.qt_tendency * aux_en.area[k]
+        aux_en.e_tot_tendency_precip_formation[k] = mph.e_tot_tendency * aux_en.area[k]
+        if edmf.moisture_model isa NonEquilibriumMoisture
+            aux_en.ql_tendency_precip_formation[k] = mph.ql_tendency * aux_en.area[k]
+            aux_en.qi_tendency_precip_formation[k] = mph.qi_tendency * aux_en.area[k]
+        end
+        tendencies_pr.q_rai[k] += mph.qr_tendency * aux_en.area[k]
+        tendencies_pr.q_sno[k] += mph.qs_tendency * aux_en.area[k]
+    end
+    return nothing
+end
+
+function quad_loop(en_thermo::SGSQuadrature, precip_model, vars, param_set, Δt::Real)
+
+    env_len = 8
+    src_len = 9
+    i_ql, i_qi, i_T, i_cf, i_qt_sat, i_qt_unsat, i_T_sat, i_T_unsat = 1:env_len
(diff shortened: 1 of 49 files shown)


tapios:
Thanks, @charleskawczynski. This is all correct. However, the order of priorities and workflow is
 
1. Fully compressible (e.g., BOMEX).
2. Implicit solver.
3. Use energy consistently.

If we can finish 3 quickly, we can build the implicit solver directly with energy; otherwise, implicit solver can start with liquid-ice potential temperature.

src/TurbulenceConvection/Parameters.jl
```diff
@@ -0,0 +1,68 @@
...
+
+abstract type AbstractTurbulenceConvectionParameters end
+const ATCP = AbstractTurbulenceConvectionParameters
+
+#####
+##### TurbulenceConvection parameters
+#####
+
+Base.@kwdef struct TurbulenceConvectionParameters{FT, MP, SFP} <: ATCP
+    Omega::FT
```
tapios, line 19: Omega should be inherited from ClimaAtmos now.

src/TurbulenceConvection/Parameters.jl
```diff
@@ -0,0 +1,68 @@
...
+    planet_radius::FT
```
tapios, line 20: Ditto for planet_radius: inherited from ClimaAtmos

src/TurbulenceConvection/EDMF_Environment.jl
```diff
@@ -0,0 +1,447 @@
...
+            prog_pr.q_sno[k],
+            aux_en.area[k],
+            ρ_c[k],
+            FT(grid.zc[k].z),
+            Δt,
+            ts,
+            precip_fraction,
+        )
+
+        # update_sat_unsat
```
tapios, line 41: Would be good to leave TODO comment here: We should clean the following up and remove the variables we do not need in a subsequent iteration (e.g., θ_liq_ice). It's also confusing that θ_virt is only defined for unsaturated condensations but not otherwise.

src/TurbulenceConvection/EDMF_Environment.jl
```diff
@@ -0,0 +1,447 @@
...
+    p_c = aux_gm.p
+    ρ_c = prog_gm.ρ
+
+    #TODO - if we start using eos_smpl for the updrafts calculations
+    #       we can get rid of the two categories for outer and inner quad. points
+
+    # arrays for storing quadarature points and ints for labeling items in the arrays
+    # a python dict would be nicer, but its 30% slower than this (for python 2.7. It might not be the case for python 3)
+
+    epsilon = 10e-14 # eps(float)
```
tapios, line 293: Why is this hard-coded.

src/TurbulenceConvection/EDMF_Updrafts.jl
```diff
@@ -0,0 +1,147 @@
+"""
+Computes tendencies to q_liq and q_ice due to
+condensation, evaporation, deposition and sublimation
```
tapios, line 3: It's confusing that this is called `EDMF_Updrafts.jl`, suggesting it deals with updrafts broadly; however, the scope here is much narrower.

src/TurbulenceConvection/EDMF_functions.jl
```diff
@@ -0,0 +1,1352 @@
...
+function area_surface_bc(
+    surf::SurfaceBase{FT},
+    edmf::EDMFModel,
+    i::Int,
+)::FT where {FT}
+    N_up = n_updrafts(edmf)
+    return surf.bflux > 0 ? edmf.surface_area / N_up : FT(0)
+end
+
+function w_surface_bc(::SurfaceBase{FT})::FT where {FT}
```
tapios, line 455: Need to revisit this: If we are solving for covariant vertical velocity, the BC is not zero at topography. And generally, we should be careful to distinguish between physical vertical velocity (w) and covariant vertical velocity (u_3) now.

src/TurbulenceConvection/EDMF_functions.jl
```diff
@@ -0,0 +1,1352 @@
...
+        end
+
+        tends_ρarea[kc_surf] = 0
+        tends_ρaθ_liq_ice[kc_surf] = 0
+        tends_ρaq_tot[kc_surf] = 0
+    end
+
+    # Solve for updraft velocity
+
+    # We know that, since W = 0 at z = 0, BCs for entr, detr,
```
tapios, line 779: Here too: physical vertical velocity and covariant vertical velocities are nonzero at surfaces with topography. Revisit.

src/TurbulenceConvection/EDMF_functions.jl
```diff
@@ -0,0 +1,1352 @@
...
+            (ρaw * (I0f(entr_w) * w_en - I0f(detr_w) * w_up)) +
+            (ρ_f * Iaf(a_up) * I0f(buoy)) +
+            nh_pressure
+        tends_ρaw[kf_surf] = 0
+    end
+
+    return nothing
+end
+
+function filter_updraft_vars(
```
tapios, line 812: Can we call this `crop_updraft_vars` because it is clipping/cropping, rather than filtering (that term suggests conservation properties that we do not have here).

src/TurbulenceConvection/Grid.jl
```diff
@@ -0,0 +1,243 @@
+
```
tapios, line 1: We shouldn't need this TurbulenceConvection. Mesh will be inherited from ClimaAtmos. If we want smaller domains in a SCM configuration of ClimaAtmos running the EDMF scheme, this should be configured in a ClimaAtmos driver.

src/TurbulenceConvection/Parameters.jl
```diff
@@ -0,0 +1,68 @@
...
+    microph_scaling::FT # TODO: move to microphysics parameter set? or Clima1M?
+    microph_scaling_dep_sub::FT # TODO: move to microphysics parameter set? or Clima1M?
+    microph_scaling_melt::FT # TODO: move to microphysics parameter set? or Clima1M?
+    microphys_params::MP
+    surf_flux_params::SFP
+end
+
+thermodynamics_params(ps::ATCP) =
```
tapios, line 28: Should be inherited from ClimaAtmos too.

src/TurbulenceConvection/Parameters.jl
```diff
@@ -0,0 +1,68 @@
...
+#####
+
+##### Forwarding Thermodynamics.jl
+
+const TDPS = TD.Parameters.ThermodynamicsParameters
+for var in fieldnames(TDPS)
+    @eval $var(ps::ATCP) = TD.Parameters.$var(thermodynamics_params(ps))
+end
+
+# derived parameters
```
tapios, line 51: Ditto: Inherit from ClimaAtmos.

tc_driver/NetCDFIO.jl
```diff
@@ -0,0 +1,212 @@
+import NCDatasets
```
tapios, line 1: We should also integrate this with ClimaAtmos I/O

tc_driver/TimeStepping.jl
```diff
@@ -0,0 +1,54 @@
+mutable struct TimeStepping{FT}
```
tapios, line 1: TimeStepping will be taken care of by ClimaAtmos driver.

tapios (commented):
I left a few comments, most concerning things that now will need to be inherited from ClimaAtmos (parameters, mesh, I/O), rather than being defined standalone in the parameterization. We do not have to address all of them right away, but it would be good to start as cleanly as we can, and leave TODOs about the rest.

src/TurbulenceConvection/Parameters.jl
charleskawczynski: See this [comment](https://github.com/CliMA/ClimaAtmos.jl/pull/583#discussion_r915939028)

src/TurbulenceConvection/EDMF_Environment.jl
charleskawczynski: Legacy artifact.

charleskawczynski:
In the interest of time, let's merge this and I'll open issues for the review points.

Issue #392, opened 2022-05-03
Remove global closures


charleskawczynski:
We have a lot of closures over global variables in the ClimaAtmos [driver](https://github.com/CliMA/ClimaAtmos.jl/blob/ac910ccc9db304de9ceb3335707e8310b7457cb5/examples/hybrid/driver.jl) and all _scripts_ (non-modules), that it includes. Here is a very simple example of a closure:

```julia
a = 1

function foo(x, y)
   z = a + x*y     # closure over variable `a`
   return R
end

x = 2
y = 3
foo(x, y)
```

These closures have several issues:

 - Such code is not portable (i.e., we cannot move the code into `src/` since it depends)
 - Analyzing performance is clouded by many (many) non-inferred references to variables outside functions. This is not necessarily disastrous in terms of performance (it can be, however, if the closure variable is not declared `const` and appears inside a broadcast expression)
 - Code cannot be "siloed" away. It's helpful that well designed and encapsulated code becomes mature, it can be moved into a dedicated location (e.g., sometimes a package). This is an important principle to apply to very large codebases so that we can narrow the scope of more experimental code.

One challenge with removing closures is that there is a design decision to be made. One could remove closures by making variables arguments:

```julia
function foo(x, y, a_1, a_2, a_3, a_4, a_5, a_6, a_7)
...
end
```

However, this is not a good design because it can be very easy to accidentally swap arguments when calling, and this can be a difficult bug to find.

This kind of forces us to think about data structures, the above example could be written more cleanly if we pack some of these variables in a `NamedTuple` (or `struct`):

```julia
a_vals = (; a_1, a_2, a_3, a_4, a_5, a_6, a_7) # or we could define / use a struct
function foo(x, y, a_vals)
...
end
```

This is especially helpful if these parameters are common among many function calls.

Looking at the driver, it seems like we could group various variables into a few core groups:

 - `spaces`: container around topology, mesh, topography etc.
 - `simulation`: container around start/end time?, callbacks / diagnostics
 - `numerics`: container around discretization specs (e.g., upwinding/central differencing)
 - `model` container around what defines the continuous equations that are solved, turbconv, thermal variable, moisture variables etc.

Some abstractions / ideas around this already exist in ClimaAtmos, so it may be good to think about how we can integrate these pieces, too.

charleskawczynski:
Closed after #1620

PR #592, opened 2022-07-07
Remove some global closures

charleskawczynski:
This PR is a drop in the bucket, but it removes a few global closures, and I've moved `additional_tendency!` into ClimaAtmos src.


simonbyrne (approved):

charleskawczynski:
A small step towards #392.

PR #585, opened 2022-07-06
evaluate Wfact bycolumn

simonbyrne:
# PULL REQUEST

## Purpose and Content
This uses the new `bycolumn` evaluation for calling `Wfact!`: it fuses operations by columns, and enables better use of threading.

## Benefits and Risks
It gives about 2x speedup in single cores, and much better threading scaling.

## Linked Issues
(Provide references to any link issues. Use closes #issuenum to automatically close an open issue)
- Fixes #
- Closes #

## PR Checklist
- [ ] This PR has a corresponding issue OR is linked to an SDI.
- [ ] I have followed CliMA's codebase [contribution](https://clima.github.io/ClimateMachine.jl/latest/Contributing/) and [style](https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/) guidelines OR N/A.
- [ ] I have followed CliMA's [documentation policy](https://github.com/CliMA/policies/wiki/Documentation-Policy).
- [ ] I have checked all issues and PRs and I certify that this PR does not duplicate an open PR.
- [ ] I linted my code on my local machine prior to submission OR N/A.
- [ ] Unit tests are included OR N/A.
- [ ] Code used in an integration test OR N/A.
- [ ] All tests ran successfully on my local machine OR N/A.
- [ ] All classes, modules, and function contain docstrings OR N/A.
- [ ] Documentation has been added/updated OR N/A.

PR #595, opened 2022-07-08
Remove regression tests for trivially zero solutions

charleskawczynski:
This PR removes regression tests for trivially zero solutions in the GABLS test.


charleskawczynski:
Confirmed with @yairchn that this is the right thing to do. It should fix the remaining issue related to why CI is failing in #585.

yairchn (approved):

PR #599, opened 2022-07-08
add a compressibility model option which updates pressure in assign_thermo_aux

yairchn:
Add the update of pressure from density if the model is compressible


charleskawczynski (approved):

yairchn:
Bors r+

PR #611, opened 2022-07-10
Only move results to main in case of self reference

charleskawczynski:
This PR improves our regression workflow by only moving results onto central if we're doing a self-reference comparison (since these are the only datasets that are used for comparison).


bischtob:
@charleskawczynski , is this current or should be closed?

charleskawczynski:
I guess we'll merge!

PR #612, opened 2022-07-10
Rename post processing folder to `regression_tests`

charleskawczynski:
This PR renames the `post_processing` folder to `regression_tests`, since it's more accurate of what's being done in there.

I think, to keep this working, we'll either need to either

 - increment the reference counter (or reset to 1), and let this be the new folder we push things into (leaning towards this)
 - "look for" the `ref_counter.jl` / `mse_tables.jl` files when doing comparisons (which seems a bit more confusing)

We could change the history of the older reference datasets on the cluster, but I'm not sure that's ideal-- we should probably clean that out soon since it's accumulating fast. #611 will help with this, though.

I'm going to wait to merge this until we figure out what's going on with the long runs.


charleskawczynski:
I've got a sketch of updating the readme, but I think it'd make more sense to apply that after this is merged.

PR #614, opened 2022-07-11
Revert 585

charleskawczynski:
This fixes most of the long runs (see this [build](https://buildkite.com/clima/climaatmos-longruns/builds/235)). I'm not sure that we want to merge this or try to figure out why these changes are breaking.


jiahe23:
I believe the longrun problem is resolved. Can we close this?

PR #618, opened 2022-07-11
revert linsolve changes

### LINKED ISSUE #610: The `bycolumn` PR broke the long runs
The long runs broke on #585. See builds

 - [231](https://buildkite.com/clima/climaatmos-longruns/builds/231): long runs work (for PR #589)
 - [232](https://buildkite.com/clima/climaatmos-longruns/builds/232): long runs break (for PR #585)

This also kind of makes sense since #585 adjusted the mse tables which, now that I'm thinking of it, doesn't make sense to me.  @simonbyrne do you know why these changes would result in behavioral differences?

Also, if this is another instance of machine precision differences, we really need to make the long runs more robust by not running so close to the unstable limit. @szy21, @jiahe23, can we use slightly smaller timesteps for these runs so that they're not so unstable? Configurations like this are really difficult to maintain / update, our model (and configuration) shouldn't be sensitive to machine precision. #585 was, was aimed at making the model faster, so we should be able to afford a slightly smaller timestep.

Before changing the timestep, however, let's make sure that there wasn't a bug in the bycolumn implementation. (cc @simonbyrne )

simonbyrne:
## Purpose and Content
Reverts just the linsolve changes from #585.

Long run here:
https://buildkite.com/clima/climaatmos-longruns/builds/238

## Benefits and Risks
(State concisely the benefits to be derived from and the risks associated with this PR)

## Linked Issues
- Potentially fixes #610 (hopefully)

## PR Checklist

### DIFF
--- a/examples/hybrid/schur_complement_W.jl
+++ b/examples/hybrid/schur_complement_W.jl
@@ -164,78 +164,67 @@ function linsolve!(::Type{Val{:init}}, f, u0; kwargs...)
         @nvtx "linsolve" color = colorant"lime" begin
 
             # Compute Schur complement
-            Fields.bycolumn(axes(x.c)) do colidx
 
-                # TODO: Extend LinearAlgebra.I to work with stencil fields. Allow more
-                # than 2 diagonals per Jacobian block.
-                FT = eltype(eltype(S))
-                I = Ref(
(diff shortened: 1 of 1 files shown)


simonbyrne:
This seems to fix it: @charleskawczynski

charleskawczynski (approved):

PR #622, opened 2022-07-11
Update regression readme

charleskawczynski:
This PR updates the regression test readme.

It would be great to get feedback on this since I'm really not the best reviewer for this PR.

Here is the new [readme](https://github.com/CliMA/ClimaAtmos.jl/tree/ck/regression_readme/regression_tests#regression-tests)


charleskawczynski:
Hmm, I think that this is objectively better than what's in main, so let's merge and iterate on it.

PR #626, opened 2022-07-12
Expand test coverage

charleskawczynski:
@jiahe23 pointed out that this will break, so we should fix whatever code is broken that's not exercised.


jiahe23:
It looks like something in #602 introduce weird behaviour of the mpi

charleskawczynski:
> It looks like something in #602 introduce weird behaviour of the mpi:
> 
> * It breaks output for mpi;
> * It outputs log twice https://buildkite.com/clima/climaatmos-ci/builds/1944#0181e0ff-8275-490f-8c56-8b91182ffd41

Looking back at the logs, it seems that #602 did break the output for the mpi config, but the duplicate log output seems to have existed for quite a while. Let's open an issue about that and merge this in the meantime.

PR #629, opened 2022-07-13
Fix distributed logger

### LINKED ISSUE #627: MPI jobs have duplicate output
@jiahe23 pointed out that our MPI jobs have duplicate output. @simonbyrne, is this an issue?

charleskawczynski:
This seems to fix #627, I'm not exactly sure why, but I'm not super familiar with how `atexit` works in a distributed context (assuming that's the reason).

It'd be nice if we could add some sort of assertion, but maybe not necessary as global scope is decluttered.

### DIFF
--- a/examples/hybrid/driver.jl
+++ b/examples/hybrid/driver.jl
@@ -178,18 +178,17 @@ if simulation.is_distributed
         error("ENV[\"CLIMACORE_DISTRIBUTED\"] only supports the \"MPI\" option")
     end
     const pid, nprocs = ClimaComms.init(comms_ctx)
-    atexit() do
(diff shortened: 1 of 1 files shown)


charleskawczynski:
It would be good to get confirmation that this is not just hiding a bug. cc @simonbyrne

jiahe23:
> It would be good to get confirmation that this is not just hiding a bug. cc @simonbyrne

Yes... Better to rule out potential bugs. 

A while ago I set up a distributed simulation which had duplicate log info, it turned out that is due to an incorrect MPI setup and I was in fact running two jobs from the two processors I requested (instead of a distributed job on the two processor)... 

I hope this time it is just a logger problem...

simonbyrne (approved):

PR #634, opened 2022-07-13
remove covar dissipation

yairchn:
covar dissipation is used only for diagnostics, as the equation is solved implicitly


charleskawczynski (approved):

yairchn:
Bors r+

PR #641, opened 2022-07-13
backout env w from grid mean w and updrafts

yairchn:
Adding the grid mean w to the update aux of the EDMF


src/TurbulenceConvection/update_aux.jl
```diff
@@ -53,9 +54,6 @@ function update_aux!(
     ##### center variables
     #####
     C123 = CCG.Covariant123Vector
-    @. aux_en.e_kin =
```
yairchn, line 56: this is wrong, w of env is not updated here yet

src/TurbulenceConvection/update_aux.jl
```diff
@@ -160,6 +160,7 @@ function update_aux!(
         aux_en.q_tot[k] =
             max(val1 * aux_gm.q_tot[k] - val2 * aux_bulk.q_tot[k], 0) #Yair - this is here to prevent negative QT
         aux_en.h_tot[k] = val1 * aux_gm.h_tot[k] - val2 * aux_bulk.h_tot[k]
+        aux_en.e_kin[k] = val1 * aux_gm.e_kin[k] - val2 * aux_bulk.e_kin[k]
```
yairchn, line 163: instead of computing e_kin in the env from the various velocities I am backing out its value form the updraft and grid mean , it is not the same but should be ok, its small anyways

src/TurbulenceConvection/update_aux.jl
charleskawczynski: Oh no! Is this an issue in TC, too?

yairchn: I undid this change as it has some weird effects I need to investigate separately. lets merge just adding gm_w without behavior changes first ?

charleskawczynski (approved):
Code changes look fine to me.

yairchn:
Bors r+

PR #653, opened 2022-07-14
Add perf target bk job, update benchmarks

charleskawczynski:
This PR

 - Addresses @cmbengue's request to polish / automate a performance analysis report for single processor (unthreaded + multithreaded) for our nearest neighbor to amip. At the moment, the performance targets are missing edmf-- which I'll try to fix in a subsequent PR, after which the performance report will auto-update.

 - Adds/uses some benchmark utilities developed in Thermodynamics
 - Adds two benchmarks: unthreaded (adjusted existing benchmark script) + multithreaded (new script)

For now, we're explicitly avoiding IO, however, we should revisit this soon. Also, our benchmark for `step!` is not "robust" to callbacks-- we may call some of the callbacks depending on the interval / frequency specified. This should also be revisited.

Here's an example output (for the unthreaded):

```
Benchmarking Wfact...
Benchmarking linsolve...
Benchmarking implicit_tendency!...
Benchmarking remaining_tendency!...
Benchmarking step!...
┌─────────────────────┬────────────┬──────────┬────────────┬────────────┬────────────┬────────────┬───────────┐
│ Function            │     Memory │   allocs │       Time │       Time │       Time │       Time │ N-samples │
│                     │   estimate │ estimate │        min │        max │       mean │     median │           │
├─────────────────────┼────────────┼──────────┼────────────┼────────────┼────────────┼────────────┼───────────┤
│ Wfact               │ 335.53 MiB │   259207 │ 528.069 ms │ 731.836 ms │ 605.385 ms │ 592.255 ms │         8 │
│ linsolve            │ 205.67 MiB │   280806 │ 277.099 ms │ 405.817 ms │ 316.950 ms │ 287.764 ms │        10 │
│ implicit_tendency!  │  13.67 KiB │       51 │ 649.578 ms │ 674.674 ms │ 658.891 ms │ 657.224 ms │         8 │
│ remaining_tendency! │  13.82 MiB │    43442 │    2.690 s │    2.697 s │    2.693 s │    2.693 s │         2 │
│ step!               │ 838.18 MiB │   951620 │   13.180 s │   13.180 s │   13.180 s │   13.180 s │         1 │
└─────────────────────┴────────────┴──────────┴────────────┴────────────┴────────────┴────────────┴───────────┘
```

Note that some of these functions call other functions, so there's a scope understanding that's required for interpreting results. `step!` is the outer-most call and can be considered the wall time per timestep (note that a single `step!` performs multiple `wfact`, `linsolve`, `implicit_tendency!` and `remaining_tendency!` calls).

These results will be available on every merged PR in the "performance targets" jobs (threaded / unthreaded).

I'm using BenchmarkTools to perform the benchmarks, we could call multiple times to force a larger sample size, but the existing config will automatically increase the sample size as we improve performance.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -225,6 +225,21 @@ steps:
         artifact_paths: "sphere_aquaplanet_rhoe_equilmoist_clearsky/*"
 
 
+  - group: "Performance targets"
+    steps:
+
+      # TODO: add/exercise edmf code
+      - label: ":computer: Unthreaded performance target"
+        command: "julia --color=yes --project=perf perf/benchmark.jl --job_id perf_target_unthreaded --enable_threading false --vert_diff true --moist equil --rad allskywithclear --microphy 0M --dt 1secs --t_end 10secs --dt_save_to_sol Inf --z_elem 25 --h_elem 12"
+        artifact_paths: "perf_target_unthreaded/*"
+
+      - label: ":computer: Threaded performance target"
+        command: "julia --color=yes --threads 8 --project=perf perf/benchmark.jl --job_id perf_target_threaded --enable_threading true --vert_diff true --moist equil --rad allskywithclear --microphy 0M --dt 1secs --t_end 10secs --dt_save_to_sol Inf --z_elem 25 --h_elem 12"
+        artifact_paths: "perf_target_threaded/*"
+        agents:
+          slurm_nodes: 1
+          slurm_tasks_per_node: 8
+
   - group: "TurbulenceConvection"
     steps:
 
--- a/examples/hybrid/cli_options.jl
+++ b/examples/hybrid/cli_options.jl
@@ -295,21 +295,22 @@ parsed_args = dict["sphere_aquaplanet_rhoe_equilmoist_allsky"];
 include("examples/hybrid/driver.jl")
 ```
 """
-function parsed_args_per_job_id()
+function parsed_args_per_job_id(; trigger = "driver.jl")
     ca_dir = joinpath(@__DIR__, "..", "..")
     buildkite_yaml = joinpath(ca_dir, ".buildkite", "pipeline.yml")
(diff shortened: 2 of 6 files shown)


cmbengue-ec-intl (commented):
This looks good to me Charlie. Thanks for your expeditious implementation of it. A few minor comments:

- The functions and scripts could benefit from some more detailed in-file documentation.
- The threaded and unthreaded scripts seem to have a lot of shared code, I'd try to merge them if I were you.

Otherwise, good job.

.buildkite/pipeline.yml
```diff
@@ -225,6 +225,19 @@ steps:
...
+    steps:
+
+      # TODO: add/exercise edmf code
+      - label: ":computer: Unthreaded performance target"
+        command: "julia --color=yes --project=perf perf/benchmark.jl --job_id perf_target_unthreaded --enable_threading false --vert_diff true --moist equil --rad allskywithclear --microphy 0M --dt 1secs --t_end 10secs --dt_save_to_sol Inf --z_elem 25 --h_elem 12"
+        artifact_paths: "perf_target_unthreaded/*"
+
+      - label: ":computer: Threaded performance target"
+        command: "julia --color=yes --threads 8 --project=perf perf/benchmark.jl --job_id perf_target_threaded --enable_threading true --vert_diff true --moist equil --rad allskywithclear --microphy 0M --dt 1secs --t_end 10secs --dt_save_to_sol Inf --z_elem 25 --h_elem 12"
+        artifact_paths: "perf_target_threaded/*"
```
simonbyrne, line 238:
```suggestion
        artifact_paths: "perf_target_threaded/*"
        agents:
          slurm_cpus_per_task: 8

```

charleskawczynski:
Looks like we need some sort of config adjustment to get accurate numbers for the threading case. We can follow up on that and get the report in now.

PR #655, opened 2022-07-14
rm ρ_ae_KM and ρ_ae_KH

yairchn:


src/TurbulenceConvection/EDMF_functions.jl
```diff
@@ -255,13 +255,10 @@ function compute_diffusive_fluxes(
...
-    @. aux_tc_f.ρ_ae_KM = IfKM(aeKM) * ρ_f
-
-    aeKHq_tot_bc = -surf.ρq_tot_flux / a_en[kc_surf] / aux_tc_f.ρ_ae_KH[kf_surf]
-    aeKHh_tot_bc = -surf.ρe_tot_flux / a_en[kc_surf] / aux_tc_f.ρ_ae_KH[kf_surf]
-    aeKMu_bc = -surf.ρu_flux / a_en[kc_surf] / aux_tc_f.ρ_ae_KM[kf_surf]
-    aeKMv_bc = -surf.ρv_flux / a_en[kc_surf] / aux_tc_f.ρ_ae_KM[kf_surf]
+    aeKHq_tot_bc = FT(0)
+    aeKHh_tot_bc = FT(0)
+    aeKMu_bc = FT(0)
+    aeKMv_bc = FT(0)
```
charleskawczynski, line 261: How is this possible?

yairchn: these are only here for diagnostics as we apply a lower bc on the SGS flux given by the flux from SF.jl. what ever is set here is overwritten later

charleskawczynski (commented):
I get that these values are only used for diagnostics, I'm just wondering if this will need to be added back in for any reason (the values we're replacing are not exactly correct, even though they're not used). I guess my concern is that someone looking at these intermediate fields may be confused. Also, doesn't calibration depend on these intermediate fields, cc @ilopezgp

yairchn:
I am hoping to unify them as a total sgs flux in CA in a future PR with SF fluxes as power level bc. I think it's just too many variable to hold on a gcm

charleskawczynski:
> I am hoping to unify them as a total sgs flux in CA in a future PR with SF fluxes as power level bc. I think it's just too many variable to hold on a gcm

Fair point, are we able to actually remove `ρ_ae_KM` and `ρ_ae_KH` from the aux field? Or is it still being used somewhere?

trontrytel:
Would it be possible to remove those variables entirely? It is a but confusing to see the bc set this way

PR #656, opened 2022-07-14
remove intermidiaries where not needed

yairchn:


trontrytel (commented):
I thought some of those intermiediate instances were supposed to be faster than referring to the actual variable?

charleskawczynski:
> I thought some of those intermiediate instances were supposed to be faster than referring to the actual variable?

https://github.com/CliMA/ClimaCore.jl/issues/341

charleskawczynski (approved):

PR #657, opened 2022-07-15
add hs with AMIP targeted resolution to longrun

jiahe23:
# PULL REQUEST

## Purpose and Content
This PR adds high resolution hs simulations to longrun buildkite.
It sets up mpi with 128 tasks.

## Benefits and Risks
(State concisely the benefits to be derived from and the risks associated with this PR)

## Linked Issues
(Provide references to any link issues. Use closes #issuenum to automatically close an open issue)
- Fixes #
- Closes #

## PR Checklist
- [ ] This PR has a corresponding issue OR is linked to an SDI.
- [ ] I have followed CliMA's codebase [contribution](https://clima.github.io/ClimateMachine.jl/latest/Contributing/) and [style](https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/) guidelines OR N/A.
- [ ] I have followed CliMA's [documentation policy](https://github.com/CliMA/policies/wiki/Documentation-Policy).
- [ ] I have checked all issues and PRs and I certify that this PR does not duplicate an open PR.
- [ ] I linted my code on my local machine prior to submission OR N/A.
- [ ] Unit tests are included OR N/A.
- [ ] Code used in an integration test OR N/A.
- [ ] All tests ran successfully on my local machine OR N/A.
- [ ] All classes, modules, and function contain docstrings OR N/A.
- [ ] Documentation has been added/updated OR N/A.

PR #660, opened 2022-07-15
Fix threading performance target

charleskawczynski:
This PR corrects the threading flags for the performance target job. There's several options that can / should be used together, and from [these docs](https://slurm.schedmd.com/cpu_management.html#Overview):

> Some experimentation may be required to discover the exact combination of options needed to produce a desired outcome.

So, I've experimented a bit to understand what needs to be defined. Here's what I found:

```yaml
  agents: # works
    slurm_nodes: 1
    slurm_ntasks: 8
    slurm_ntasks_per_node: 8

  agents: # does not work
    slurm_nodes: 1
    slurm_ntasks_per_node: 8

  agents: # works
    slurm_ntasks: 8
    slurm_ntasks_per_node: 8

  agents: # works
    slurm_nodes: 1
    slurm_ntasks: 8
```

The takeaway is: `slurm_ntasks` _must_ be specified, _along with_ either `slurm_nodes` or `slurm_ntasks_per_node`.

We need to apply corrections to the threading specification in #657. I think that PR also needs fixes with how we specify the distributed configuration (maybe a similar job like this would be useful for breaking down the problem with a distributed config). Also, @sriharshakandala, it looks like there's a typo in ClimaCore.jl's scaling pipeline:
 - [`slurm_tasks_per_node`](https://github.com/CliMA/ClimaCore.jl/blob/d1d73c92869047b7d44e2ab3c68fa3685ab7e1ed/.buildkite/scaling/pipeline.yml#L41) should be [`slurm_ntasks_per_node`](https://www.hpc.caltech.edu/documentation/slurm-commands)

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -237,8 +237,9 @@ steps:
         command: "julia --color=yes --threads 8 --project=perf perf/benchmark.jl --job_id perf_target_threaded --enable_threading true --vert_diff true --moist equil --rad allskywithclear --microphy 0M --dt 1secs --t_end 10secs --dt_save_to_sol Inf --z_elem 25 --h_elem 12"
         artifact_paths: "perf_target_threaded/*"
         agents:
-          slurm_nodes: 1
-          slurm_tasks_per_node: 8
+          slurm_ntasks: 8
+          slurm_ntasks_per_node: 8
+
 
   - group: "TurbulenceConvection"
     steps:


sriharshakandala:
`SLURM_TASKS_PER_NODE` is a valid option https://slurm.schedmd.com/sbatch.html#lbAG

charleskawczynski:
> `SLURM_TASKS_PER_NODE` is a valid option https://slurm.schedmd.com/sbatch.html#lbAG

It may be a valid slurm option, but from [this ref]():

> CPU Management through user commands is constrained by the configuration parameters chosen by the Slurm administrator.

and I don't see `TASKS_PER_NODE` on [Caltech's site](https://www.hpc.caltech.edu/documentation/slurm-commands), only `ntasks-per-node`.

Also, from the experiments I ran, `slurm_ntasks` was necessary to specify.

Have we tested that increasing threads alone (while keeping processors constant) improves the performance in the scaling pipeline? It looks like they're increased together, so it's not clear to me that both are actually working.

PR #671, opened 2022-07-18
Refactor surface flux calculation

jiahe23:
# PULL REQUEST

## Purpose and Content
Refactors the surface flux calculation to use surface_conditions

## Benefits and Risks
This makes the surface flux calculation cleaner, and is a step towards using Monin Obukhov

## PR Checklist

co-authored-by: @szy21

### DIFF
--- a/examples/hybrid/sphere/baroclinic_wave_utilities.jl
+++ b/examples/hybrid/sphere/baroclinic_wave_utilities.jl
@@ -566,17 +566,14 @@ function vertical_diffusion_boundary_layer_cache(
     else
         ts_type = TD.PhaseDry{FT}
     end
-    coef_type = SF.Coefficients{
-        FT,
-        SF.InteriorValues{FT, Tuple{FT, FT}, ts_type},
-        SF.SurfaceValues{FT, Tuple{FT, FT}, TD.PhaseEquil{FT}},
-    }
+
+    cond_type = NamedTuple{(:shf, :lhf, :E, :ρτxz, :ρτyz), NTuple{5, FT}}
 
     return (;
         ᶠv_a = similar(Y.f, eltype(Y.c.uₕ)),
         ᶠz_a,
         ᶠK_E = similar(Y.f, FT),
-        flux_coefficients = similar(z_bottom, coef_type),
+        surface_conditions = similar(z_bottom, cond_type),
         dif_flux_uₕ,
         dif_flux_energy = similar(z_bottom, Geometry.WVector{FT}),
         dif_flux_ρq_tot,
@@ -596,7 +593,7 @@ function eddy_diffusivity_coefficient(norm_v_a, z_a, p)
     return p > p_pbl ? K_E : K_E * exp(-((p_pbl - p) / p_strato)^2)
 end
 
-function constant_T_saturated_surface_coefs(
+function constant_T_saturated_surface_conditions(
     T_sfc,
     ts_int,
     uₕ_int,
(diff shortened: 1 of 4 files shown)


szy21:
bors try

examples/hybrid/sphere/baroclinic_wave_utilities.jl
```diff
@@ -606,22 +603,42 @@ function constant_T_saturated_surface_coefs(
...
+
+    E = SF.evaporation(sc, sf_params, tsf.Ch)
+
+    return (;
+        shf = tsf.shf,
+        lhf = tsf.lhf,
+        E = E,
+        ρτxz = tsf.ρτxz,
+        ρτyz = tsf.ρτyz,
     )
```
charleskawczynski, line 642: I think that this is all going to be allocating

szy21: Should I put them in the cache then? Maybe it would be better to have a pair programming session if you have time?

examples/hybrid/sphere/baroclinic_wave_utilities.jl
```diff
@@ -566,17 +566,14 @@ function vertical_diffusion_boundary_layer_cache(
     else
         ts_type = TD.PhaseDry{FT}
     end
-    coef_type = SF.Coefficients{
-        FT,
-        SF.InteriorValues{FT, Tuple{FT, FT}, ts_type},
-        SF.SurfaceValues{FT, Tuple{FT, FT}, TD.PhaseEquil{FT}},
-    }
+
+    cond_type = NamedTuple{(:shf, :lhf, :E, :ρτxz, :ρτyz), NTuple{5, FT}}
```
charleskawczynski, line 570:
```suggestion
    cond_type = SF.SurfaceFluxConditions{FT <: Real}
```
We could alternatively cache the surface conditions, like this

examples/hybrid/sphere/baroclinic_wave_utilities.jl
charleskawczynski: Let's see how the perf summary looks. It shouldn't be that complicated, though

examples/hybrid/sphere/baroclinic_wave_utilities.jl
```diff
@@ -606,22 +603,42 @@ function constant_T_saturated_surface_coefs(
...
-        Cd,
-        Ch,
         z0m = FT(0),
         z0b = FT(0),
+        Cd = Cd, #FT(0.001),
+        Ch = Ch, #FT(0.0001),
+    )
+
+    # calculate all fluxes
+    tsf = SF.surface_conditions(sf_params, sc, SF.FVScheme())
```
charleskawczynski, line 632: We would just return this if caching the surface conditions.

charleskawczynski: The reason I'm suggesting this is that basically all the elements of the returned NamedTuple are grabbed from the surface conditions, which makes me think that we should just return that instead

szy21: We need evaporation E (for rho and rhoq_tot tendency) too, I think that's why we did this.

charleskawczynski (approved):
Overall, these changes look good and I think that the removed call to `get_momentum_fluxes` will actually improve things w.r.t. allocations. We can make changes w.r.t. the cache structure later.

szy21:
I will also remove `sphere_held_suarez_rhoe_int_equilmoist` from the buildkite as well unless we have to keep it, because after this PR we won't support surface fluxes for internal energy anymore.

charleskawczynski:
> I will also remove `sphere_held_suarez_rhoe_int_equilmoist` from the buildkite as well unless we want to keep it, because after this PR we won't support surface fluxes for internal energy anymore.

Fine by me.

examples/hybrid/sphere/baroclinic_wave_utilities.jl
```diff
@@ -743,35 +743,11 @@ function vertical_diffusion_boundary_layer_tendency!(Yₜ, Y, p, t)
         )
         @. Yₜ.c.ρe_tot +=
             ᶜdivᵥ(ᶠK_E * ᶠinterp(ᶜρ) * ᶠgradᵥ((Y.c.ρe_tot + ᶜp) / ᶜρ))
-    elseif :ρe_int in propertynames(Y.c)
```
szy21, line 746: @charleskawczynski Maybe instead of removing it, we can throw an error here and stop the simulation if we are using `ρe_int`. Would this be better? If so how should I do this?

charleskawczynski: We can try throwing an error. Maybe add in one place and make sure that JET doesn't complain before going and adding it everywhere. Alternatively, we could just add the assertion/error when creating `Y` (that's actually probably much simpler)

szy21: Maybe we can do it in `get_model_spec` [here](https://github.com/CliMA/ClimaAtmos.jl/blob/227c9039ff341a081492c7d4a28f2ea8a5cf9e0f/examples/hybrid/types.jl#L102)? Or do you think it's better to do it [here](https://github.com/CliMA/ClimaAtmos.jl/blob/227c9039ff341a081492c7d4a28f2ea8a5cf9e0f/examples/hybrid/sphere/baroclinic_wave_utilities.jl#L27) when creating Y?

charleskawczynski: Ah, true, we could error in model spec. Either seems fine

charleskawczynski (approved):

PR #694, opened 2022-07-25
zero tendencies

yairchn:
We found that we need to zero out the tendencies in the beginning of `implicit_tendency_special` as the EDMF components are coded assuming that it is done before updating tendencies.

### DIFF
--- a/examples/hybrid/staggered_nonhydrostatic_model.jl
+++ b/examples/hybrid/staggered_nonhydrostatic_model.jl
@@ -173,6 +173,11 @@ function implicit_tendency_special!(Yₜ, Y, p, t)
     @nvtx "implicit tendency special" color = colorant"yellow" begin
         Fields.bycolumn(axes(Y.c)) do colidx
 
+            if p.tendency_knobs.has_turbconv
+                parent(Yₜ.c.turbconv[colidx]) .= FT(0)
+                parent(Yₜ.f.turbconv[colidx]) .= FT(0)
+            end
+
             @. ᶜK[colidx] =
                 norm_sqr(C123(ᶜuₕ[colidx]) + C123(ᶜinterp(ᶠw[colidx]))) / 2
 


examples/hybrid/staggered_nonhydrostatic_model.jl
```diff
@@ -166,6 +166,9 @@ function implicit_tendency_special!(Yₜ, Y, p, t)
 
     ref_thermo_params = Ref(thermo_params)
     ref_zuₕ = Ref(zero(eltype(Yₜ.c.uₕ)))
+    FT = eltype(Yₜ)
+    parent(Yₜ.c.turbconv) .= FT(0)
+    parent(Yₜ.f.turbconv) .= FT(0)
```
charleskawczynski, line 171: `.turbconv` doesn't exist by default, so you'll need a branch for when it doesn't exist.

examples/hybrid/staggered_nonhydrostatic_model.jl
```diff
@@ -166,6 +166,9 @@ function implicit_tendency_special!(Yₜ, Y, p, t)
 
     ref_thermo_params = Ref(thermo_params)
     ref_zuₕ = Ref(zero(eltype(Yₜ.c.uₕ)))
+    FT = eltype(Yₜ)
```
charleskawczynski, line 169: I don't think that this is how you get the float type

examples/hybrid/staggered_nonhydrostatic_model.jl
```diff
@@ -166,6 +166,9 @@ function implicit_tendency_special!(Yₜ, Y, p, t)
...
+    parent(Yₜ.c) .= FT(0)
+    parent(Yₜ.f) .= FT(0)
```
charleskawczynski, line 171: We should probably add this to the generic one, too. Also, can we move this into the threaded loop in this function so that we leverage threading?

yairchn: >  Also, can we move this into the threaded loop in this function so that we leverage threading?
can you suggest a code?

examples/hybrid/staggered_nonhydrostatic_model.jl
yairchn: is this better `FT = eltype(Yₜ.c.uₕ)`

trontrytel: I think `FT` is already defined. Seems like all the other functions are just magically using it in this file

examples/hybrid/staggered_nonhydrostatic_model.jl
trontrytel: Do you mean to do it in line 175 like this:

` @. Yₜ.c[colidx] = FT(0)
`

examples/hybrid/staggered_nonhydrostatic_model.jl
```diff
@@ -173,6 +173,17 @@ function implicit_tendency_special!(Yₜ, Y, p, t)
     @nvtx "implicit tendency special" color = colorant"yellow" begin
         Fields.bycolumn(axes(Y.c)) do colidx
 
+            # this assumes that thermodynamic covariances are diagnostic
+            if p.tendency_knobs.has_turbconv
+                @. Yₜ.c.turbconv.up[1].ρarea[colidx] = FT(0)
+                @. Yₜ.c.turbconv.up[1].ρaθ_liq_ice[colidx] = FT(0)
```
trontrytel, line 179: Does the `[1]` assume we only have one updraft?

charleskawczynski: Yes

charleskawczynski: Does
```julia
@. Yₜ.c.turbconv[colidx] = FT(0)
@. Yₜ.f.turbconv[colidx] = FT(0)
```
work? I imagine that would be ideal

yairchn: yes that would work

charleskawczynski (approved):

yairchn:
Bors r+

PR #696, opened 2022-07-26
Updated if statement to work with the coupler driver

kmdeck:
# PULL REQUEST

## Purpose and Content
The coupler driver file is also run with command line arguments. Currently, the coupler cli arguments are specified in `cli_options.jl`. They are a set which include the exact same arguments of the atmosphere `cli_options.jl`, but include additional coupler specific args. 

In the current setup, the coupler driver reads in cli_options.jl (coupler version), and reads in the parsed args correctly. Later on, we include the clima atmos examples/driver.jl file. This file includes the atmos cli_options.jl, which redefines the parse_command_line function. This function then errors on the parsed_args which include the coupler args as well - when called within the atmos script.

## Benefits and Risks
Benefits: this lets the coupler run with command line arguments
Risks: This is a hacky solution. A better solution would be some sort of simulations function which takes in parsed_args and sets up the simulation (and integrator). in a clima atmos driver file, this function could be called with clima atmos parsed_args, and the simulation ran. In a clima coupler driver file, this function would be called with coupler+clima atmos parsed args, and the returned integrator could be used for the coupled simulation. Is this the right time to implement something like this? What other fixes are there?

Note that it also means we have essentially two copies of clima atmos cli_options.jl - one in the coupler, and one in clima atmos, that need to be manually kept consistent, so this is part of a broader issue.

## PR Checklist
- [X] I have checked all issues and PRs and I certify that this PR does not duplicate an open PR.
- [X] Unit tests are included OR N/A.
- [X] Code used in an integration test OR N/A.
- [X] All classes, modules, and function contain docstrings OR N/A.
- [X] Documentation has been added/updated OR N/A.

### DIFF
--- a/examples/hybrid/driver.jl
+++ b/examples/hybrid/driver.jl
@@ -1,5 +1,5 @@
-include("cli_options.jl")
 if !(@isdefined parsed_args)
+    include("cli_options.jl")
     (s, parsed_args) = parse_commandline()
 end
 


kmdeck:
@LenkaNovak correctme if the description of the problem is incorrect :)

charleskawczynski (commented):
I would prefer that we decompose `driver.jl` into `driver.jl` and `runner.jl` or something, where `parsed_args` is assumed to be defined in `runner.jl`. This way, the coupler can just use the `runner.jl` script and do whatever custom things that may need in `cli_options.jl`.

Alternatively, the coupler could define methods that don't clash with those we define in `cli_options.jl` and use the script as is.

In my experience, this sort of conditional code loading tends to confuse people.

LenkaNovak (approved):

LenkaNovak:
> I would prefer that we decompose `driver.jl` into `driver.jl` and `runner.jl` or something, where `parsed_args` is assumed to be defined in `runner.jl`. This way, the coupler can just use the `runner.jl` script and do whatever custom things that may need in `cli_options.jl`.
> 
> Alternatively, the coupler could define methods that don't clash with those we define in `cli_options.jl` and use the script as is.

I see you point, @Charlie. I wonder though, whether we could do minimal changes that work for us now, and address the interface issues thereafter, also consulting with @dennisYatunin @jiahe23 and @bischtob ? I know they also had some great ideas / preferences on this.

kmdeck:
> Alternatively, the coupler could define methods that don't clash with those we define in `cli_options.jl` and use the script as is.

If this is possible, it seems like it would be an easy change. But I dont know that it will work? The atmos script includes the atmos cli_options.jl, then runs functions from it. It is these functions that fail with our larger parsed_args set (I think `jobid_from_parsed_args` in `get_simulation`). I don't know that redefining coupler methods fixes this, but we can double check.

jiahe23:
This looks good to me. If the alternative takes less than a day's work, it is worth trying. Otherwise it is not worth putting too much effort only to meet the current Atmos interface, since @dennisYatunin will start working on the refined interface very soon [right after we have first amip]...

kmdeck:
@LenkaNovak what is the call on this? Merge this or just use this commit of CA in the coupler (only a short term solution...)?

LenkaNovak:
I would say (unless someone has the bandwidth to do/test this right now) merge as is and optimize later. @charleskawczynski  @jia ?

charleskawczynski:
Alright, let's go with the patch for now.

PR #698, opened 2022-07-26
Add assertion that job finishes

charleskawczynski:
This PR adds an assertion that the job finishes to completion.


yairchn (approved):

yairchn:
Bors r+

PR #700, opened 2022-07-26
update RRTMGP version

szy21:


szy21:
bors try

examples/Manifest.toml
```diff
@@ -1,6 +1,6 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.7.3"
+julia_version = "1.7.0"
```
szy21, line 3: @charleskawczynski I used julia 1.7.0 which results in this change. Could I just change it back to 1.7.3, or do I need to change something else as well?

charleskawczynski: I think it's probably fine, but you should upgrade!

charleskawczynski (approved):

examples/Manifest.toml
szy21: ok I will! :)

szy21:
@charleskawczynski I think I probably did something wrong…Could you teach me how to update dependencies?

charleskawczynski:
> @charleskawczynski I think I probably did something wrong…Could you teach me how to update dependencies?

Updating the compat entries and using `julia .dev/up_deps.jl` should be all that's needed. Based on the log alone 🧐, the (very consistent) error you're getting, `ERROR: LoadError: importing ODE into Main conflicts with an existing identifier`, seems unrelated to package updates. Based on the warning, `┌ Warning: Replacing module OrdinaryDiffEq`, I'd say that there's maybe something wrong with the depot path. Have you tried clearing the depot path?

szy21:
> Updating the compat entries and using `julia .dev/up_deps.jl` should be all that's needed. Based on the log alone 🧐, the (very consistent) error you're getting, `ERROR: LoadError: importing ODE into Main conflicts with an existing identifier`, seems unrelated to package updates. Based on the warning, `┌ Warning: Replacing module OrdinaryDiffEq`, I'd say that there's maybe something wrong with the depot path. Have you tried clearing the depot path?

Ah, I thought it was related to the package updates because other PRs seem fine. Let me try clearing the depot path, thanks!

charleskawczynski:
This didn't actually update RRTMGP (the manifest files didn't change)

szy21:
> This didn't actually update RRTMGP (the manifest files didn't change)

Hmm what did I change then 🤦‍♀️ Instead of doing what you suggested, for this PR I did ]up RRTMGP, commit the Manifest.toml, and change the Project.toml. Maybe this is wrong?

szy21:
I should definitely go to the workshop for onboarding new schmidt scholars! :)

charleskawczynski:
> Hmm what did I change then

Other dependencies

> Instead of doing what you suggested, for this PR I did ]up RRTMGP, commit the Manifest.toml, and change the Project.toml. Maybe this is wrong?

Yes, that's incorrect. Updating the manifests takes the project toml into account. So the order resulted in the manifest not getting updated (and now the compat entry is incorrect).

No worries though, fixing #703 will fix this in that what you did will break when trying with CI because the test env will auto-update to the latest versions (which I discovered is incompatible when trying to upgrade deps elsewhere).

PR #711, opened 2022-07-28
Upgrade deps

charleskawczynski:
This PR upgrades our deps, which includes the recent ClimaCore perf improvements.


charleskawczynski:
Results seem fine from what I can see.

PR #724, opened 2022-08-04
Aj/hydrostatic profiles

trontrytel:
# PULL REQUEST

## Purpose and Content

This PR makes sure that the initial condition for pressure and density for the single column TC test cases in CA is 
-  consistent with the temperature and humidity profiles prescribed in the test cases
-  hydrostatic. 

It is done during the TC initialisation by solving via ODE an initial value problem for a given temperature and humidity profile and surface pressure. Then the dycore pressure and density are overwritten.

The good news is that with this update I can run a compressible Bomex simulation with TC in CA. Here is the command line I used to run the model:

`julia --color=yes --project=examples examples/hybrid/driver.jl --config column --FLOAT_TYPE Float64 --hyperdiff false --moist equil --turbconv edmf --turbconv_case Bomex --dt_save_to_sol 1mins --z_elem 60 --z_stretch false --z_max 3e3 --dt 1secs --t_end 1hours --regression_test true --job_id edmf_bomex`

This is how the vertical velocity looks like without this PR (on the left) and with it (on the right). Looks like I need the discrete version of it?

Screenshot 2022-08-04 at 12 19 04

Also, what I'm doing here is relevant for the single column test cases such as Bomex. The next question is how we want to initialize the simulations on a sphere so that we don't run into this

## TODO
- [x] fix documentation
- [x] fix TRMM
- [x] wait for SurfaceFluxes.jl release

## Moved to follow up PRs
- [ ] extend to LES_driven_SCM cases
- [ ] add discrete hydrostatic / fix stretched cases

## PR Checklist


tc_driver/dycore.jl
```diff
@@ -201,7 +201,11 @@ function set_grid_mean_from_thermo_state!(param_set, state, grid)
     prog_gm_f = TC.face_prog_grid_mean(state)
     aux_gm = TC.center_aux_grid_mean(state)
     prog_gm_uₕ = TC.grid_mean_uₕ(state)
+
+    @. prog_gm.ρ = TD.air_density(thermo_params, ts_gm)
```
charleskawczynski, line 205:
```suggestion
```
This doesn't seem correct. We solve for `prog_gm.ρ`.

trontrytel: I thought thats what you suggested - to overwrite the dycore state from TC.jl 
I want to do it in the initialization only. The initialisation done in the dycore right now is not compatible with the prescribed profile of Bomex and other single column test cases

charleskawczynski: Ah, yes. Maybe we can put this directly inside the init function though? So that it's not accidentally called inside the RHS?

tapios (commented):
This looks good. But can we make it a bit more modular, so that it works more easily for any profile of potential temperature and specific humidity? That is, e.g., write a generic dp_dz function that takes theta and q as input, and let theta an q be anything.

trontrytel:
> This looks good. But can we make it a bit more modular, so that it works more easily for any profile of potential temperature and specific humidity? That is, e.g., write a generic dp_dz function that takes theta and q as input, and let theta an q be anything.

Sure! I was rushing to test if this solves our compressible Bomex problems. And looks like even with the kinks in the velocity profile we can have a 6 hours simulation:

Screenshot 2022-08-04 at 13 50 38

Will be working on making it more general for other cases we have next

tapios:
> > This looks good. But can we make it a bit more modular, so that it works more easily for any profile of potential temperature and specific humidity? That is, e.g., write a generic dp_dz function that takes theta and q as input, and let theta an q be anything.
> 
> Sure! I was rushing to test if this solves our compressible Bomex problems. And looks like even with the kinks in the velocity profile we can have a 6 hours simulation:
> 
> Screenshot 2022-08-04 at 13 50 38
> 
> Will be working on making it more general for other cases we have next

I'm glad it's working out!

tc_driver/dycore.jl
trontrytel: It's part of the list of functions that are being called from `init_tc`: 
https://github.com/CliMA/ClimaAtmos.jl/blob/7a922bcd3c69d959f642c233113e660edf79c9e8/examples/hybrid/TurbulenceConvectionUtils.jl#L116-L150
Plus, we are already overwriting the `prog_gm.rhoe_tot` and prog_gm.rhoq_tot` from that function. 

Are you suggesting the name of the function should be different. Or that it should be explicitly coded inside `init_tc`, without it being a separate function?

trontrytel:
I'm working on extending the hydrostatic profile calculation from Bomex to all our test cases. 

I have two questions and was wondering if I could get some hints on them @charleskawczynski:

- I need to know the `z_0` and `z_max` as well as the pressure at the surface when computing the pressure profile. The surface pressure I have available in the `Cases` file, but the domain extend is a namelist or command line argument. So I'm not sure how to handle it here.

- `GATE_III` and our `LES driven cases` don't specify the `theta_liq_ice` profile. I was wondering if for the `LES driven cases` we could just read in the pressure profile from the LES dataset. But I'm not sure what to do about `GATE_III`

yairchn:
> 

The PyCLES pressure should be equally problematic as its computed from hydrostatic balance with uniform moist entropy from the surface.

tc_driver/Cases.jl
```diff
@@ -211,6 +211,46 @@ update_forcing(::AbstractCaseType, grid, state, t::Real, param_set) = nothing
...
+    - given θ_liq_ice and q_tot initial profiles
+"""
+function dp_dz!(p, params, z)
+    (; param_set, prof_θ_liq_ice, prof_q_tot) = params
+    thermo_params = TCP.thermodynamics_params(param_set)
+
+    θ_liq_ice = prof_θ_liq_ice(z)
+    q_tot = prof_q_tot(z)
+
+    q = TD.PhasePartition(q_tot)
```
yairchn, line 230: could we form a thermodynamic state given this assumption and use it below?

tc_driver/Cases.jl
```diff
@@ -226,21 +266,26 @@ end
...
     prog_gm = TC.center_prog_grid_mean(state)
-    ρ_c = prog_gm.ρ
     FT = TC.float_type(state)
     prof_q_tot = APL.Soares_q_tot(FT)
     prof_θ_liq_ice = APL.Soares_θ_liq_ice(FT)
     prof_u = APL.Soares_u(FT)
     prof_tke = APL.Soares_tke(FT)
+    # Solving the initial value problem
+    p_0::FT = FT(1000 * 100) # TODO - duplicated from surface_ref_state
+    z_0::FT = FT(0)          # TODO - where to get it from? Those are also our CLI and namelist args...
```
yairchn, line 276: maybe 
```
kf_surf = TC.kf_surface(grid)
kf_toa = TC.kf_top_of_atmos(grid)
z_0::FT = grid.z[kf_surf]
z_max::FT = grid.z[kf_toa]
```
would work here

tc_driver/Cases.jl
```diff
@@ -226,21 +266,26 @@ end
...
     aux_gm = TC.center_aux_grid_mean(state)
     prog_gm = TC.center_prog_grid_mean(state)
-    ρ_c = prog_gm.ρ
     FT = TC.float_type(state)
     prof_q_tot = APL.Soares_q_tot(FT)
     prof_θ_liq_ice = APL.Soares_θ_liq_ice(FT)
     prof_u = APL.Soares_u(FT)
     prof_tke = APL.Soares_tke(FT)
+    # Solving the initial value problem
+    p_0::FT = FT(1000 * 100) # TODO - duplicated from surface_ref_state
```
yairchn, line 275: I honestly think that copying `p_0` is fine, the entire `Cases` structure will change at the end

tc_driver/Cases.jl
```diff
@@ -862,8 +933,14 @@ function initialize_profiles(
     prog_gm_uₕ = TC.grid_mean_uₕ(state)
     TC.set_z!(prog_gm_uₕ, prof_u, prof_v)
 
+    # TODO - no theta profile given
```
yairchn, line 936: you could use `T` given from the GATE paper directly in `dp_dz` right? 
what if we add `dp_dz` as a function in thermodynamics and have several versions of it these?

tc_driver/Cases.jl
```diff
@@ -1193,6 +1296,8 @@ end
 ##### LES_driven_SCM
 #####
 
+# TODO - what to do about those? Could we read in pressure from LES?
```
yairchn, line 1299: line 1407 you have the thetali profile from LES that can use in a similar way. 
@costachris can you take a look as well?

tc_driver/dycore.jl
yairchn: `set_grid_mean_from_thermo_state` is used only in initialization anyways.

tc_driver/Cases.jl
trontrytel: How about I put all the surface values in the `AtmosphericProfilesLibrary.jl`? This way I could load them in both places, without duplicating. Would that be ok @charleskawczynski?

trontrytel:
> > 
> 
> The PyCLES pressure should be equally problematic as its computed from hydrostatic balance with uniform moist entropy from the surface.

As long as it is hydrostatically balanced it should not be giving us waves. Even if the uniform moist entropy assumption is not ideal

tc_driver/Cases.jl
trontrytel: Thanks @yairchn . This works, though it feels like running in circles ;)
```
    z_0::FT = grid.zf[TC.kf_surface(grid)].z                                    
    z_max::FT = grid.zf[TC.kf_top_of_atmos(grid)].z 
```

tc_driver/Cases.jl
trontrytel: Never mind. Will just hard code them twice for now.

tc_driver/Cases.jl
trontrytel: I don't think so (?). We need to compute the pressure or density first. And then pass it to our thermo state constructor.

tc_driver/Cases.jl
trontrytel: done

tc_driver/Cases.jl
trontrytel: I think it would be better to just take the pressure profile from LES. It should be hydrostatic. 

@costachris do you know if it's available?

trontrytel:
Update from today:
- The compressible Bomex and DYCOMS_RF01 simulations are comparable to the anelastic ones. 
- TRMM doesn't work right now. I suspect that the issue is the pressure loaded from APL, so I'll see if I can compute it differently. 
- GABL fails right now because of SurfaceFluxes. This issue should be fixed soon. 

Aside from that, I still need to update the LES driven cases.

trontrytel:
GABLS is fixed by the updates from the main SurfaceFluxes branch. 

Making TRMM profiles hydrostatic stops the convection for both the compressible and anelastic configuration. I'll continue working on this, but I would also vote to merge this PR as soon as the SF.jl updates are released. And work on the remaining issues as a follow up PRs.

tapios:
> GABLS is fixed by the updates from the main SurfaceFluxes branch.
> 
> Making TRMM profiles hydrostatic stops the convection for both the compressible and anelastic configuration. I'll continue working on this, but I would also vote to merge this PR as soon as the SF.jl updates are released. And work on the remaining issues as a follow up PRs.

Awesome, @trontrytel! Can you do TRMM with a hydrostatic profile to which you add a buoyancy (e.g., warm air) perturbation at the bottom?

trontrytel:
I fixed the `q_tot` profile in `TRMM` and now the anelasitc simulation is working fine without any tweaks. The compressible simulation crashes, but not immediately. I'm testing if reducing the time step would help.

trontrytel:
TRMM and the LES driven cases are the remaining things to fix. But regardless, PR is ready for review.

regression_tests/ref_counter.jl
```diff
@@ -1 +1 @@
-9
+12
```
charleskawczynski, line 1: We'll only want to increase the counter once

yairchn (approved):
Great work

PR #728, opened 2022-08-05
Add a simulation without dt_save_to_sol to the longrun

szy21:
Adds a simulation without saving to sol to check the performance. Also updates the MSE table (This PR shouldn't affect the MSE, so probably the MSE table was not updated correctly before).


szy21:
bors try

szy21:
hmm why would this break edmf mse? @charleskawczynski

charleskawczynski:
> hmm why would this break edmf mse? @charleskawczynski

Hmm, interesting. This could happen if someone incremented the reference counter without updating the mse tables. I'm not totally sure. Feel free to update the mse tables and put a note in the description.

charleskawczynski (approved):

PR #733, opened 2022-08-09
Update perf target and cleanup

charleskawczynski:
This PR:
 - Avoids threaded functions when threading is disabled
 - Adds held suarez forcing to the perf target (I noticed allocations in Held Suarez forcing, so it could be costing us)
 - Adds a call to `Profile.clear()` in the flame graph script (I'm hoping that this may clear out global scope calls)
 - Cleans up some unused functions


charleskawczynski:
Wow, this actually changes the picture quite a bit. Alleviating `^` seems definitely worth the effort now.

PR #738, opened 2022-08-10
Make sure overloading, hack `calc_tderivative!`

charleskawczynski:
This PR:
 - Moves some code loading, to make sure that we're overloading some OrdinaryDiffEq methods (`SplitFunction`). It turns out that it was, but moving this code up is also good because it confirms that there's no closures in these functions.
 - Hacks `OrdinaryDiffEq.calc_tderivative!` to skip calling `f` (this presumably is not needed when no explicit time derivatives exist).

It looks like:
 - All of our regression tests are passing, which is great
 - One additional RHS eval is eliminated (so we get a ~30% speedup)

### DIFF
--- a/examples/hybrid/driver.jl
+++ b/examples/hybrid/driver.jl
@@ -3,6 +3,9 @@ if !(@isdefined parsed_args)
     (s, parsed_args) = parse_commandline()
 end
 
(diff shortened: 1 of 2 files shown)


dennisYatunin (approved):
Looks good. Great catch, Charlie!

PR #744, opened 2022-08-14
try removing TC.jl from CA.jl deps

trontrytel:
I noticed we still had `TurbulenceConvection.jl` in `test` and `perf` `toml` files. Could those dependencies be deleted?


charleskawczynski (commented):
Probably better not to remove it.

I think we’ve effectively converged on this env configuration. There’s various trade offs and issues with removing TC as a dep in these environments, and I think I’d prefer to leave as is.

If we were to change it, additional changes may be needed

trontrytel:
Sounds good! I just noticed it because I was trying to update CloudMicrophysics.jl and was getting conflicts from TurbulenceConvection.jl. Which gave me a pause, since this is ClimaAtmos.jl

But this is not urgent at all. Was just thinking we should clean up

PR #751, opened 2022-08-17
Split out ICs to separate file

charleskawczynski:
I can never remember where the initial conditions are defined, baro wave utilities, or staggered nh model. Now they'll live in `initial_conditions.jl`

@jiahe23, feel free to merge while I'm out!


jiahe23 (approved):
Yep, much cleaner!

Issue #721, opened 2022-08-01
Refactor the TurbulenceConvection argument handling in ClimaAtmos


trontrytel:
## Purpose
The code related goal is to refactor the way we handle simulation arguments and parameters in TC in CA. Right now some of them are passed via command line, while others are passed by `namelist` files, while others are passed via `toml` files.

The second goal of this project is to onboard a new team member and expose him to the TC and CA structure.

## Cost/Benefits/Risks
Benefits:
- Ensure that parameters passed for a given simulation configuration are the relevant parameters for it
- Ensure that the defaults are defined in one place and not overwritten by defaults in another place
- Embrace more the new `toml` file way of handling parameter values that can be calibrated
Risks:
- The calibration pipeline for EDMF still depends on the `namelist` files. When doing this refactor it would be good to know the specifications for switching the calibration to rely more on the `toml` files and less on the `namelist` files. 

## People and Personnel
(@ the personnel required/requested to implement the software solution)
- Lead: @nefrathenrici 
- Collaborators: @trontrytel 
- Reviewers: @charleskawczynski @dennisYatunin @odunbar @costachris 

## Components

1) Flattening the `namelist` structure in TC (As suggested here: https://github.com/CliMA/TurbulenceConvection.jl/issues/908)
2) Changing the flattened `namelist` file into `toml` file(s). (We could potentially split it between files with parameter values and files with simulation flags).
3) <del> Adding a way to automatically generate a `namelist` file from the `toml` file. This way we would move towards using the new `toml` files from the model interface standpoint, while still keeping the `namelist` file for calibration pipeline. <\del>
4) Identifying which of the `namelist/toml` parameters should also be allowed to be changed via the command line and adding them to cli. 
5) Using ArgParse to group commad line args(?)

## Inputs

We would appreciate some feedback on this from both the calibration side (@costachris, @odunbar) and model development side (@charleskawczynski @dennisYatunin @yairchn). Do points 1-3 make sense for everyone?

From the calibration point of view - do we envision making a new calibration pipeline for CA + TC? Or are we hoping to use CEDMF for the whole model as well? I think that knowing this would help decide if we should start working on parameters from TC or CA+TC

odunbar:
My only concern - likely as I don't know enough about the user experience of running the code - is the movement towards more flags. It seems like we start with a 3 input system
1. command line arguments and flags + Argparsing
2. a structured namelist
3. a flat TOML

[...]

PR #756, opened 2022-08-22
Add initial API page for Namelist parameters

nefrathenrici:
This will allow developers (especially myself) to see the type and description of the parameters passed in via NameList files.
Will be expanded on as I work on https://github.com/CliMA/ClimaAtmos.jl/issues/721 

Please let me know if I need to be more descriptive for future PRs, this feels appropriate for such a small change.


charleskawczynski (commented):
My preference would be to collapse the namelist into toml entries and CL entries, and then automatically generate this documentation (like what we've done with the [CL args](https://clima.github.io/ClimaAtmos.jl/dev/cl_args/))

PR #758, opened 2022-08-23
Add job with check bounds

charleskawczynski:
This PR adds a job that runs our perf target with `--check-bounds=yes` to make sure that we're not accidentally accessing invalid memory.


bischtob:
@charleskawczynski,  is this current or should be closed?

charleskawczynski:
> @charleskawczynski, is this current or should be closed?

It's current.

charleskawczynski:
Merging with soft-fail

PR #762, opened 2022-08-23
Use openmpi for parallel hdf5 output support in long runs

### LINKED ISSUE #760: high resolution held suarez in longrun buildkite fails after changing to hdf5
See [here](https://buildkite.com/clima/climaatmos-longruns/builds/325#0182c987-2297-4586-8a78-97fe4c7ad1e7) for example. Maybe it just misses some mpi hdf5 packages. @jiahe23 Could you take a look and fix it?

charleskawczynski:
I think that this should close #760.


szy21:
Thanks! I think 761 is a different issue though?

charleskawczynski:
> Thanks! I think 761 is a different issue though?

Oops! Yep

simonbyrne (approved):

PR #770, opened 2022-08-25
upgrade SurfaceFluxes

szy21:


szy21:
bors try

szy21:
bors try

szy21:
@charleskawczynski Two EDMF GABLS jobs fail in this PR. They will be fixed by #724. But Anna and I think it doesn't make sense to combine these two PRs. So we are thinking about disabling these two tests in this PR, and adding them back in 724. What do you think?

charleskawczynski:
Sounds good to me

szy21:
Oh @charleskawczynski, do I need to Increment the reference counter as well if I _remove_ some regression tests?

charleskawczynski:
> Oh @charleskawczynski, do I need to Increment the reference counter as well if I _remove_ some regression tests?

Yeah, or, I think, you could remove the entries from the mse table

charleskawczynski (approved):

PR #773, opened 2022-08-26
Enable ClimaTimeSteppers

### LINKED ISSUE #510: Integrate ClimaTimeSteppers
The rough plan:

- [x] Run some test cases with `IMEXEuler` (#506)
- [ ] Modify the driver to work with `explicit_step!(yx, y, p, t, dt)` instead of `explicit_tendency!(dy, y, p, t)`: I made a start on this in #490, but that is now out-of-date. For compatibility, we can then write `explicit_tendency! as:
```julia
function explicit_tendency!(dy, y, p, t)
   # assert no limiter
   dy .= 0
   explicit_step!(dy, y, p, t, one(t))
end
```
- [ ] Compare the results of `IMEXEuler` vs `ARS111` (these should be the same)
- [ ] Check results of `ARS232`/`ARS343`.

dennisYatunin:
# PULL REQUEST

## Purpose and Content
Enables the use of algorithms from `ClimaTimeSteppers.jl`. From now on, algorithms from `OrdinaryDiffEq.jl` must be prefaced with `ODE.` when specified through command-line arguments (i.e., `--ode_algo ODE.Rosenbrock23`). Makes `ARS343` the new default algorithm. Adds `max_newton_iters` as a command-line argument.

Since `DiffEqCallbacks.jl` assumes that `integrator` has certain fields that are not present in a `ClimaTimeSteppers.DistributedODEIntegrator`, callbacks from that package are replaced with simple custom implementations.

## Benefits and Risks
We will now be able to play around with all of the IMEX ARK algorithms in `ClimaTimeSteppers.jl`.

## Linked Issues
- Closes #510 

## PR Checklist


dennisYatunin:
bors try

examples/hybrid/staggered_nonhydrostatic_model.jl
```diff
@@ -453,84 +532,41 @@ function default_remaining_tendency_special!(Yₜ, Y, p, t)
         end
     end
 end
-function default_remaining_tendency_generic!(Yₜ, Y, p, t)
+
+function explicit_vertical_advection_tendency_generic!(Yₜ, Y, p, t)
```
Zhengyu-Huang, line 536: what is the different between generic function and special function?

Zhengyu-Huang (approved):
Thank you for the great work! I reviewed modifications in `examples/hybrid/driver.jl`, they look correct for me! I am not familiar with callback functions, I think if the test can pass, that should be correct.

examples/hybrid/staggered_nonhydrostatic_model.jl
charleskawczynski: Generic is unthreaded, special is threaded (we should probably change the name)

valeriabarra: Yeah it should be documented

examples/hybrid/callbacks.jl
```diff
@@ -83,23 +63,58 @@ function get_callbacks(parsed_args, simulation, model_spec, params)
...
+        affect!;
+        initialize = (cb, u, t, integrator) -> begin
+            skip_first || f!(integrator)
+            t_end = integrator.sol.prob.tspan[2]
+            next_t[] =
+                (call_at_end && t < t_end) ? min(t_end, t + dt) : t + dt
+        end,
+        save_positions = (false, false),
+    )
+end
```
charleskawczynski, line 101: What is this doing? Why do we need the Ref here?

dennisYatunin: Well we need some way of accessing the next stop time from the various closures. Is it better to have next_t be a Ref (so that it doesn't change but the value in it does change), or to have it be a regular Float (so that it actually changes)? I was under the impression that closures which use variables that change might run into performance issues.

dennisYatunin:
bors cancel

dennisYatunin:
This PR no longer makes ARS343 the default time stepper due to some performance issues. Instead, it adds several new jobs to CI that test ARS343.

charleskawczynski:
Looks like the restart is hanging, I'll cancel and push a commit to retry, like what we do in ClimaCore for the [HDF5 tests](https://github.com/CliMA/ClimaCore.jl/blob/6151815963ff9def9d8692452378b8a8df88ae68/.buildkite/pipeline.yml#L156-L160).

valeriabarra:
@charleskawczynski do you know why we see these IOErrors?

charleskawczynski:
> @charleskawczynski do you know why we see these IOErrors?

They're upstream race conditions

PR #775, opened 2022-08-27
Upgrade deps

charleskawczynski:
This PR upgrades the deps, including the new version of ClimaCore which has some performance improvements.


charleskawczynski:
Clear wins across the board. Some jobs were slightly slower overall due to compilation time, but their runtime improved and allocations improve significantly.

charleskawczynski:
Ah, need to increase alloc request for dycoms

PR #781, opened 2022-08-29
Add unthreaded version of linsolve

### LINKED ISSUE #737: Linsolve is hidden from flame graph
Threading, which doesn't play nice with flame graphs, was unconditionally added to `linsolve`. We should somehow make this optional so that we can include all computations in the flame graph for understanding unthreaded single processor performance.

charleskawczynski:
This PR:
 - Adds an unthreaded version of `linsolve` (to close #737)
 - Renames `ᶠ𝕄_named_tuple` to `∂ᶜ𝕋ₜ∂ᶠ𝕄_field`, since that's now more appropriate (should have been incorporated in #779).

### DIFF
--- a/examples/hybrid/linsolve_test.jl
+++ b/examples/hybrid/linsolve_test.jl
@@ -1,7 +1,7 @@
 
 function verify_matrix(x, A, b, update_matrix = false; kwargs...)
     (; dtγ_ref, S, S_column_arrays) = A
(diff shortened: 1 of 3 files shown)


charleskawczynski:
Superseded by #788 and https://github.com/CliMA/ClimaCore.jl/pull/929

PR #782, opened 2022-08-30
Move allocations outside threading loop

charleskawczynski:
`filter` is allocating, and it seems like these dynamic function calls were triggering inference to fail. This PR moves some of these function calls outside of the threading loop.


charleskawczynski:
Superseded by #809

PR #790, opened 2022-08-31
Add FCT and separate energy/tracer upwinding

dennisYatunin:
# PULL REQUEST

## Purpose and Content
This PR splits the `upwinding_mode` flag into `energy_upwinding` and `tracer_upwinding` flags, each of which can be set to `none`, `first_order`, `third_order` (which uses third-order BCs), `boris_book` (which uses first-order BCs), or `zalesak` (which uses third-order BCs). Per Toby's suggestion, this PR also makes a change in notation in `staggered_nonhydrostatic_model.jl`, replacing `𝕋` with `ρc` (which stands for density times concentration).

Since the derivatives of the outputs of the FCT operators with respect to `w` are quite challenging to evaluate analytically, they are approximated by the derivatives for the third-order upwinding operator when the Jacobian is computed.

## Benefits and Risks
This will hopefully allow us to stabilize our model using Zalesak flux-corrected transport. I have tested the new Jacobian terms by setting `test_implicit_solver = true` for several dry simulations (moist simulations do not currently support the implicit solver test) in order to reduce the risk of incorrect implementation.

## PR Checklist

### DIFF
--- a/.buildkite/longruns/pipeline.yml
+++ b/.buildkite/longruns/pipeline.yml
@@ -119,32 +119,32 @@ steps:
     steps:
 
       - label: ":computer: held suarez (ρe_tot) third order"
-        command: "julia --color=yes --project=examples --threads=8 examples/hybrid/driver.jl --forcing held_suarez --upwinding third_order --dt 350secs --t_end 1200days --fps 30 --job_id longrun_hs_rhoe_third --dt_save_to_sol 1days --dt_save_to_disk 10days"
+        command: "julia --color=yes --project=examples --threads=8 examples/hybrid/driver.jl --forcing held_suarez --tracer_upwinding third_order --dt 350secs --t_end 1200days --fps 30 --job_id longrun_hs_rhoe_third --dt_save_to_sol 1days --dt_save_to_disk 10days"
         artifact_paths: "longrun_hs_rhoe_third/*"
         agents:
           slurm_cpus_per_task: 8
 
       - label: ":computer: held suarez (ρe_tot) third order topography"
-        command: "julia --color=yes --project=examples --threads=8 examples/hybrid/driver.jl --forcing held_suarez --upwinding third_order --dt 300secs --t_end 1200days --fps 30 --job_id longrun_hs_rhoe_third --dt_save_to_sol 1days --dt_save_to_disk 10days --topography DCMIP200"
+        command: "julia --color=yes --project=examples --threads=8 examples/hybrid/driver.jl --forcing held_suarez --tracer_upwinding third_order --dt 300secs --t_end 1200days --fps 30 --job_id longrun_hs_rhoe_third --dt_save_to_sol 1days --dt_save_to_disk 10days --topography DCMIP200"
         artifact_paths: "longrun_hs_rhoe_third/*"
         agents:
           slurm_cpus_per_task: 8
 
 
       - label: ":computer: held suarez (ρe_tot) equilmoist third order"
-        command: "julia --color=yes --project=examples --threads=8 examples/hybrid/driver.jl --upwinding third_order --forcing held_suarez --moist equil --vert_diff true --microphy 0M --dt 150secs --t_end 600days --fps 30 --job_id longrun_hs_rhoe_equil_third --dt_save_to_sol 1days --dt_save_to_disk 10days"
+        command: "julia --color=yes --project=examples --threads=8 examples/hybrid/driver.jl --tracer_upwinding third_order --forcing held_suarez --moist equil --vert_diff true --microphy 0M --dt 150secs --t_end 600days --fps 30 --job_id longrun_hs_rhoe_equil_third --dt_save_to_sol 1days --dt_save_to_disk 10days"
         artifact_paths: "longrun_hs_rhoe_equil_third/*"
         agents:
           slurm_cpus_per_task: 8
 
       - label: ":computer: aquaplanet (ρe_tot) equilmoist gray radiation third order"
-        command: "julia --color=yes --project=examples --threads=8 examples/hybrid/driver.jl --upwinding third_order --vert_diff true --moist equil --rad gray --microphy 0M --dt 200secs --t_end 600days --fps 30 --job_id longrun_aquaplanet_rhoe_equil_gray_third --dt_save_to_sol 1days --dt_save_to_disk 10days"
+        command: "julia --color=yes --project=examples --threads=8 examples/hybrid/driver.jl --tracer_upwinding third_order --vert_diff true --moist equil --rad gray --microphy 0M --dt 200secs --t_end 600days --fps 30 --job_id longrun_aquaplanet_rhoe_equil_gray_third --dt_save_to_sol 1days --dt_save_to_disk 10days"
         artifact_paths: "longrun_aquaplanet_rhoe_equil_gray_third/*"
         agents:
           slurm_cpus_per_task: 8
 
       - label: ":computer: aquaplanet (ρe_tot) equilmoist clearsky radiation third order"
-        command: "julia --color=yes --project=examples --threads=8 examples/hybrid/driver.jl --upwinding third_order --vert_diff true --moist equil --rad clearsky --microphy 0M --viscous_sponge true --kappa_2_sponge 1e7 --dt 150secs --t_end 600days --fps 30 --job_id longrun_aquaplanet_rhoe_equil_clearsky_third --dt_save_to_sol 1days --dt_save_to_disk 10days"
+        command: "julia --color=yes --project=examples --threads=8 examples/hybrid/driver.jl --tracer_upwinding third_order --vert_diff true --moist equil --rad clearsky --microphy 0M --viscous_sponge true --kappa_2_sponge 1e7 --dt 150secs --t_end 600days --fps 30 --job_id longrun_aquaplanet_rhoe_equil_clearsky_third --dt_save_to_sol 1days --dt_save_to_disk 10days"
         artifact_paths: "longrun_aquaplanet_rhoe_equil_clearsky_third/*"
         agents:
           slurm_cpus_per_task: 8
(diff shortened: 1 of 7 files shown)


jiahe23:
Thanks, @dennisYatunin !

Could you also add a simulation in the longrun buildkite with [this](https://github.com/CliMA/ClimaAtmos.jl/blob/main/.buildkite/longruns/pipeline.yml#L200) setup but using 3rd order tracer upwinding? We were not able to reach a stable simulation with the current 3rd upwinding but I am curious if the tracer only setup will make any differences.

examples/hybrid/cli_options.jl
```diff
@@ -89,10 +89,14 @@ function parse_commandline()
         help = "Energy variable name [`rhoe` (default), `rhoe_int` , `rhotheta`]"
         arg_type = String
         default = "rhoe"
-        "--upwinding"
-        help = "Upwinding mode [`none` (default), `first_order` , `third_order`]"
-        arg_type = String
-        default = "none"
+        "--energy_upwinding"
+        help = "Energy upwinding mode [`none` (default), `first_order` , `third_order`, `boris_book`, `zalesak`]"
```
valeriabarra, line 93: I think Tapio mentioned that we don't need BB at all in CA. It was an intermediate operator that was needed for us to understand how to set up these FCT as CC operators and to have a baseline comparison for QA

examples/hybrid/staggered_nonhydrostatic_model.jl
```diff
@@ -54,12 +54,16 @@ const ᶠcurlᵥ = Operators.CurlC2F(
     bottom = Operators.SetCurl(Geometry.Contravariant12Vector(FT(0), FT(0))),
     top = Operators.SetCurl(Geometry.Contravariant12Vector(FT(0), FT(0))),
 )
-const ᶜFC = Operators.FluxCorrectionC2C(
-    bottom = Operators.Extrapolate(),
-    top = Operators.Extrapolate(),
+const ᶠupwind1 = Operators.UpwindBiasedProductC2F()
+const ᶠupwind3 = Operators.Upwind3rdOrderBiasedProductC2F(
+    bottom = Operators.FirstOrderOneSided(),
+    top = Operators.FirstOrderOneSided(),
```
valeriabarra, line 60: Tapio mentioned a few times that we want to use the `ThirdOrderOneSided` BCs for the `Upwind3rdOrderBiasedProductC2F` operator, for accuracy and convergence (you can see the unit tests for this operator where it's clear that the `ThirdOrderOneSided` BC outperforms the `FirstOrderOneSided` one). In the FCT operator though, we decided to use the lower-order BCs as they reduced some wiggles

examples/hybrid/staggered_nonhydrostatic_model.jl
```diff
@@ -110,8 +114,8 @@ function default_cache(Y, params, spaces, numerics, simulation)
     if apply_limiter
         tracers = filter(is_tracer_var, propertynames(Y.c))
         make_limiter =
-            ᶜ𝕋_name ->
-                Limiters.QuasiMonotoneLimiter(getproperty(Y.c, ᶜ𝕋_name), Y.c.ρ)
+            ᶜρc_name ->
```
valeriabarra, line 117: what does `ρc` stand for here? Should we just stick to `ρq` ?

examples/hybrid/staggered_nonhydrostatic_model.jl
```diff
@@ -142,13 +146,36 @@ function default_cache(Y, params, spaces, numerics, simulation)
...
-                          upwinding_mode == :third_order ? ᶠupwind_product3 :
-                          nothing,
+        energy_upwinding,
+        tracer_upwinding,
         ghost_buffer = ghost_buffer,
     )
 end
 
+# TODO: All of these should use dtγ instead of dt, but dtγ is not available in
+# the implicit tendency function. Since dt >= dtγ, we can safely use dt for now.
```
valeriabarra, line 156: I'm not sure I follow this comment here. What does it mean that since "dt >= dtγ, we can safely use dt for now"? Wouldn't be the opposite more safe? I.e., using the smaller of the two, to make sure that we do not incur in CFL violations?

examples/hybrid/staggered_nonhydrostatic_model.jl
```diff
@@ -173,29 +200,18 @@ function implicit_tendency_special!(Yₜ, Y, p, t)
...
+    thermo_params = CAP.thermodynamics_params(params)
+    dt = simulation.dt
 
     ref_thermo_params = Ref(thermo_params)
     ref_zuₕ = Ref(zero(eltype(Yₜ.c.uₕ)))
 
     @nvtx "implicit tendency special" color = colorant"yellow" begin
         Fields.bycolumn(axes(Y.c)) do colidx
-
-            if p.tendency_knobs.has_turbconv
```
valeriabarra, line 189: Is this change unrelated to this PR? Otherwise I am not sure I follow entirely why this would change with vertical upwinding mode

examples/hybrid/staggered_nonhydrostatic_model.jl
```diff
@@ -766,12 +736,49 @@ function Wfact_generic!(W, Y, p, dtγ, t)
...
+        # If ᶜρcₜ = -ᶜdivᵥ(ᶠinterp(ᶜρc) * ᶠw), then
+        # ∂(ᶜρcₜ)/∂(ᶠw_data) =
+        #     -ᶜdivᵥ_stencil(ᶠinterp(ᶜρc) * ᶠw_unit) -
+        #     ᶜdivᵥ_stencil(ᶠw) * ᶠinterp_stencil(1) * ∂(ᶜρc)/∂(ᶠw_data)
+        # If ᶜρcₜ = -ᶜdivᵥ(ᶠinterp(ᶜρ) * ᶠupwind(ᶠw, ᶜρc / ᶜρ)), then
+        # ∂(ᶜρcₜ)/∂(ᶠw_data) =
+        #     -ᶜdivᵥ_stencil(ᶠinterp(ᶜρc) *
+        #     ᶠupwind(ᶠw + εw, ᶜρc) / to_scalar(ᶠw + εw)) -
+        #     ᶜdivᵥ_stencil(ᶠinterp(ᶜρ)) * ᶠupwind_stencil(ᶠw, 1 / ᶜρ) *
+        #     ∂(ᶜρc)/∂(ᶠw_data)
```
valeriabarra, line 757: All of these (and other mathy comments) should really be part of our docs. Not all browsers or IDEs render math/unicode symbols the same, but we have a way of rendering them in a cross-platform fashion via Documenter.jl :)

examples/hybrid/staggered_nonhydrostatic_model.jl
```diff
@@ -791,155 +798,60 @@ function Wfact_generic!(W, Y, p, dtγ, t)
...
             ᶜinterp_stencil(one(ᶠw_data))
 
-        # ᶜρₜ = -ᶜdivᵥ(ᶠinterp(ᶜρ) * ᶠw)
-        # ∂(ᶜρₜ)/∂(ᶠw_data) = -ᶜdivᵥ_stencil(ᶠinterp(ᶜρ) * ᶠw_unit)
-        @. ∂ᶜρₜ∂ᶠ𝕄 = -(ᶜdivᵥ_stencil(ᶠinterp(ᶜρ) * one(ᶠw)))
-
-        @. ᶜK = norm_sqr(C123(ᶜuₕ) + C123(ᶜinterp(ᶠw))) / 2
-        thermo_state!(ᶜts, Y, params, ᶜinterp, ᶜK)
-        @. ᶜp = TD.air_pressure(thermo_params, ᶜts)
+        # vertical_transport!(Yₜ.c.ρ, ᶠw, ᶜρ, ᶜρ, dt, Val(:none))
```
valeriabarra, line 801: stale comment / dead code?

.buildkite/longruns/pipeline.yml
```diff
@@ -119,32 +119,32 @@ steps:
     steps:
 
       - label: ":computer: held suarez (ρe_tot) third order"
-        command: "julia --color=yes --project=examples --threads=8 examples/hybrid/driver.jl --forcing held_suarez --upwinding third_order --dt 350secs --t_end 1200days --fps 30 --job_id longrun_hs_rhoe_third --dt_save_to_sol 1days --dt_save_to_disk 10days"
+        command: "julia --color=yes --project=examples --threads=8 examples/hybrid/driver.jl --forcing held_suarez --tracer_upwinding third_order --dt 350secs --t_end 1200days --fps 30 --job_id longrun_hs_rhoe_third --dt_save_to_sol 1days --dt_save_to_disk 10days"
```
valeriabarra, line 122: It's good to test some of these with `third_order` just to see if it improved relative to when we were upwinding both energy and tracers with the same upwinding mode, but ultimately, we want to use FCT since it is supposed to reduce oscillations. This PR should definitely add some `--tracer_upwinding zalesak` mode jobs

examples/hybrid/staggered_nonhydrostatic_model.jl
valeriabarra: Ah, I saw the PR description mentions this

valeriabarra:
@dennisYatunin I pushed a fix to the rho theta runs. Now we need to investigate the data layout issue in the runs with topography

examples/hybrid/cli_options.jl
dennisYatunin: Might as well leave it in if it's just a few extra lines of code? If it's removed from ClimaCore, I'll get rid of it here.

examples/hybrid/staggered_nonhydrostatic_model.jl
dennisYatunin: Gotcha, I'll switch to ThirdOrderOneSided for that one.

examples/hybrid/staggered_nonhydrostatic_model.jl
dennisYatunin: If I understand correctly, using a larger dt means that the effect of FCT is larger to compensate for the higher Courant number. Also, @Zhengyu-Huang confirmed that we can use dt here instead of dtγ.

examples/hybrid/staggered_nonhydrostatic_model.jl
dennisYatunin: Yeah this was just me doing some housekeeping to make the special and generic tendencies look identical, since I was already modifying them.

examples/hybrid/staggered_nonhydrostatic_model.jl
dennisYatunin: These comments are purely for my own derivations and sanity-checking. They are not intended to be documentation for ClimaAtmos. I will eventually write some documentation for how to extend the implicit solver, but this will only come after the overall interface update, which I expect will drastically change how the implicit solver works.

examples/hybrid/staggered_nonhydrostatic_model.jl
dennisYatunin: This is a comment that indicates where the following line comes from.

examples/hybrid/staggered_nonhydrostatic_model.jl
valeriabarra: So please change this

.buildkite/longruns/pipeline.yml
dennisYatunin: I just realized that this test is meaningless because dry Held-Suarez has no tracers. I will replace it with a moist Held-Suarez test that uses Zalesak FCT.

examples/hybrid/staggered_nonhydrostatic_model.jl
dennisYatunin: https://github.com/CliMA/ClimaAtmos.jl/pull/790/commits/e01104c761ba67edab26e1eb00d332cafd5d47c3

PR #819, opened 2022-09-07
initialize and update ρ_f

yairchn:
We found that `ρ_f` is not initialize or updated by the EDMF, which means that it was using dycore values (that should be overwritten by the EDMF column by column).


charleskawczynski (approved):

trontrytel (approved):

yairchn:
Bors r+

yairchn:
Bors r+

yairchn:
Bors r+

PR #836, opened 2022-09-12
Aim allocation monitoring at perf targets

charleskawczynski:
This PR aims our allocation monitoring, at the performance targets, so that we get finer granularity of allocations across more functions.


charleskawczynski:
I think we need to fix code loading before we can do this--using `--track-allocation` tracks code loading and this may be much more expensive with the perf target since a lot of physics are turned on, which also results in more code loading.

PR #853, opened 2022-09-19
add sphere and column hydrostatic balance tests

szy21:
This PR adds hydrostatic balance tests in sphere and column configurations. We should make sure these tests work when updating numerics.

### DIFF
--- a/examples/hybrid/define_post_processing.jl
+++ b/examples/hybrid/define_post_processing.jl
@@ -111,6 +111,16 @@ function postprocessing(sol, output_dir, fps)
     end
     Plots.mp4(anim, joinpath(output_dir, "v.mp4"), fps = fps)
 
+    anim = Plots.@animate for Y in sol.u
+        ᶠw = Geometry.WVector.(Y.f.w).components.data.:1
+        Plots.plot(
+            ᶠw,
+            level = ClimaCore.Utilities.PlusHalf(3),
+            clim = (-0.02, 0.02),
+        )
+    end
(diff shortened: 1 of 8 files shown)


szy21:
bors try

examples/hybrid/define_post_processing.jl
```diff
@@ -981,4 +991,17 @@ function custom_postprocessing(sol, output_dir)
...
+        plot(
+            vec(w_phy),
+            vec(Fields.coordinate_field(Y.f).z ./ 1000);
+            xlabel = "w [m/s]",
+            ylabel = "z [km]",
+            xlims = (-0.02, 0.02),
+            legend = false,
+        )
+    end
+    Plots.mp4(anim, joinpath(output_dir, "w.mp4"), fps = 10)
```
charleskawczynski, line 1006: Could we load the solution from the exported HDF5 files? I think post-processing with `sol` is destined to be deprecated

Zhengyu-Huang (approved):
Thanks for the great work! This PR looks good for me!

examples/hybrid/define_post_processing.jl
szy21: I didn't find a current post-processing example that uses the hdf5 files. Could you point me to an example? Maybe we can also work on it together if we want to change the post-processing now.

szy21: If we do that I think it would be better to separate the change into a separate PR?

charleskawczynski (approved):

PR #860, opened 2022-09-22
CLI option: nh_poly help string

valeriabarra:
# PULL REQUEST

## Purpose and Content
I don't know if this was discussed with the recent changes when we lowered the default value. This can be considered a small nuance by some, but it can be really confusing for other people (like me). 

See the polynomial [degree](https://en.wikipedia.org/wiki/Degree_of_a_polynomial#:~:text=For%20a%20univariate%20polynomial%2C%20the,a%20polynomial%20(disambiguation)).) definition, which we represent by the current `nh_poly` option. In some contexts, especially finite and spectral elements (and I would say in general, in approximation theory) the word "order" is usually referred to either the degree+1 of the polynomials defining the basis function or the number of knot points used to determine it (see this [ref](https://en.wikipedia.org/wiki/Order_of_a_polynomial)). 

## Benefits and Risks
Improve the help string of the `--nh_poly` CLI option

## PR Checklist

### DIFF
--- a/examples/hybrid/cli_options.jl
+++ b/examples/hybrid/cli_options.jl
@@ -151,7 +151,7 @@ function parse_commandline()
         arg_type = Int
         default = 10
         "--nh_poly"
-        help = "Horizontal polynomial order"
+        help = "Horizontal polynomial degree. Note: The number of quadrature points in 1D within each horizontal element is then Nq = <--nh_poly> + 1"
         arg_type = Int
         default = 3
         "--z_max"


valeriabarra:
Along the same lines, we could consider renaming the CLI option `nh_poly` -> `ph_poly`, since we use `Nq` for "number of quadrature points" and it is not clear to me what the "n" in "nh_poly" was meant to represent (but this is really a minor preference)

bischtob:
Works for me. I added Sriharsha as reviewer to make sure he's on the same page.

bischtob (approved):

sriharshakandala:
> Along the same lines, we could consider renaming the CLI option `nh_poly` -> `ph_poly`, since we use `Nq` for "number of quadrature points" and it is not clear to me what the "n" in "nh_poly" was meant to represent (but this is really a minor preference)

`nh_poly` seems reasonable to me. May be `Np_horz` / `Np_horizontal` if we would like to be more informative!

examples/hybrid/cli_options.jl
```diff
@@ -151,7 +151,7 @@ function parse_commandline()
         arg_type = Int
         default = 10
         "--nh_poly"
-        help = "Horizontal polynomial order"
+        help = "Horizontal polynomial degree"
```
akshaysridhar, line 154:
```suggestion
        help = "Horizontal polynomial degree. Note: The number of quadrature points (or knots) within each element is then Nq = <--nh_poly> + 1."
```

valeriabarra: Thanks @akshaysridhar ! How about the proposed change in [[7794af]](https://github.com/CliMA/ClimaAtmos.jl/pull/860/commits/7794af2084180ad7d45bff85fca58c2f3790d4d6)? I added "by default" b/c we can in principle specify different Nq's to allow for under/over-integration

examples/hybrid/cli_options.jl
```diff
@@ -151,7 +151,7 @@ function parse_commandline()
         arg_type = Int
         default = 10
         "--nh_poly"
-        help = "Horizontal polynomial order"
+        help = "Horizontal polynomial degree. Note: The number of quadrature points in 1D within each horizontal element is then Nq = <--nh_poly> + 1 by defualt"
```
akshaysridhar, line 154:
```suggestion
        help = "Horizontal polynomial degree. Note: The number of quadrature points in 1D within each horizontal element is then Nq = <--nh_poly> + 1."
```
Suggest removing the `by default` comment since this is true following our definitions of those terms (Putting in the default statement would be fine if we present an example with our default setting of  --nh_poly = 3). (Since this is the outward facing interface and has no information about over-integration for instance)

valeriabarra: okay, done.

PR #862, opened 2022-09-22
set dycore pressure to EDMF gm_aux pressure

yairchn:
This PR is fixing a bug in the current EDMF implementation that the grid mean pressure used in the EDMF is not the same as the dycore pressure.

### DIFF
--- a/examples/hybrid/TurbulenceConvectionUtils.jl
+++ b/examples/hybrid/TurbulenceConvectionUtils.jl
@@ -133,11 +133,13 @@ function init_tc!(Y, p, param_set, namelist)
     CC.Fields.bycolumn(axes(Y.c)) do colidx
         # `nothing` goes into State because OrdinaryDiffEq.jl owns tendencies.
         state = tc_column_state(Y, aux, nothing, colidx)
+        aux_gm = TC.center_aux_grid_mean(state)
 
         grid = TC.Grid(state)
         FT = eltype(grid)
         t = FT(0)
         compute_ref_state!(state, grid, tc_params; ts_g = surf_ref_state)
+        @. p.ᶜp[colidx] = aux_gm.p
 
         Cases.initialize_profiles(case, grid, tc_params, state)
         set_thermo_state_pθq!(state, grid, edmf.moisture_model, tc_params)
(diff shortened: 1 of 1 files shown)


examples/hybrid/TurbulenceConvectionUtils.jl
```diff
@@ -133,6 +133,8 @@ function init_tc!(Y, p, param_set, namelist)
     CC.Fields.bycolumn(axes(Y.c)) do colidx
         # `nothing` goes into State because OrdinaryDiffEq.jl owns tendencies.
         state = tc_column_state(Y, aux, nothing, colidx)
+        aux_gm = TC.center_aux_grid_mean(state)
+        @. p.ᶜp[colidx] = aux_gm.p
```
charleskawczynski, line 137: This will (eventually) be reversed, correct? Also, I think this should be moved to after `compute_ref_state!`, where `aux_gm.p` is populated

examples/hybrid/TurbulenceConvectionUtils.jl
```diff
@@ -167,6 +169,8 @@ function sgs_flux_tendency!(Yₜ, Y, p, t)
     # TODO: write iterator for this
     CC.Fields.bycolumn(axes(Y.c)) do colidx
         state = tc_column_state(Y, aux, Yₜ, colidx)
+        aux_gm = TC.center_aux_grid_mean(state)
+        @. p.ᶜp[colidx] = aux_gm.p
```
charleskawczynski, line 173: I think this needs to be moved to after `aux_gm.p` is populated

examples/hybrid/TurbulenceConvectionUtils.jl
yairchn: `in cases like Bomex`the logic will stay this way (the dycore will match the pressure given by the balanced reference state of the case). We would at some point have to make up our minds about the by-column vs. global code structure and unify these by populating `p.ᶜp[colidx]` based on the case

charleskawczynski: Ok, makes sense

trontrytel:
Are we sure the pressure is initialized at that point?

charleskawczynski:
I think what we should do first / instead is continue with unifying the dycore's cached variables with the edmf cache, so that we can share more functions without dealing with assigning data between copies of variables.

yairchn:
agreed

PR #864, opened 2022-09-22
Change default max_newton_iters=1 and dt for longruns

### LINKED ISSUE #863: Change default time-stepper to ARS343 and increase dt
From @szy21 :

- change the default `ode_algo`  to `ARS343`  and max_newton_iters  to 1

- We should also increase dt by a factor of 2 in all the ci and longrun buildkite jobs that uses the default timestepper (we should get a factor of 2 increase in dt as a result of the change from Rosenbrock23 to ARS343.)

valeriabarra:
# PULL REQUEST

## Purpose and Content
This PR updates some longrun CI jobs to use a higher `dt` and fewer Newton iterations for the `ARS343` scheme.

## Benefits and Risks
~~This allows us to use the newly fixed ARS343 algorithm as default time-stepping scheme and we should see a factor of 2 increase in the `dt`.~~ EDIT: For now, until perf analysis is cleared, it was deemed safer to not change the default algorithm but only increase the `dt` and decrease the number of Newtown iterations for the longruns that use `ARS343`

## Linked Issues
- This is an incremental step towards #863 

## PR Checklist

### DIFF
--- a/examples/hybrid/cli_options.jl
+++ b/examples/hybrid/cli_options.jl
@@ -111,7 +111,7 @@ function parse_commandline()
         "--max_newton_iters"
         help = "Maximum number of Newton's method iterations (only for ODE algorithms that use Newton's method)"
         arg_type = Int
-        default = 3
+        default = 1
         "--split_ode"
         help = "Use split of ODE problem. Examples: [`true` (implicit, default), `false` (explicit)]"
         arg_type = Bool
--- a/.buildkite/longruns/pipeline.yml
+++ b/.buildkite/longruns/pipeline.yml
@@ -59,7 +59,7 @@ steps:
           slurm_ntasks: 8
 
       - label: ":computer: baroclinic wave (ρe_tot) equilmoist high resolution ars343"
-        command: "mpiexec julia --project=examples examples/hybrid/driver.jl --ode_algo ARS343 --moist equil --microphy 0M --z_elem 45 --dz_bottom 30 --h_elem 16 --kappa_4 1e16 --dt 200secs --t_end 100days --dt_save_to_disk 1days --job_id longrun_bw_rhoe_equil_highres_ars343"
+        command: "mpiexec julia --project=examples examples/hybrid/driver.jl --ode_algo ARS343 --moist equil --microphy 0M --z_elem 45 --dz_bottom 30 --h_elem 16 --kappa_4 1e16 --dt 400secs --t_end 100days --dt_save_to_disk 1days --job_id longrun_bw_rhoe_equil_highres_ars343"
         artifact_paths: "longrun_bw_rhoe_equil_highres_ars343/*"
         env:
           CLIMACORE_DISTRIBUTED: "MPI"
@@ -67,7 +67,7 @@ steps:
           slurm_ntasks: 8
 
       - label: ":computer: baroclinic wave (ρe_tot) equilmoist high resolution zalesak ars343"
-        command: "mpiexec julia --project=examples examples/hybrid/driver.jl --ode_algo ARS343 --moist equil --microphy 0M --tracer_upwinding zalesak --z_elem 45 --dz_bottom 30 --h_elem 16 --kappa_4 1e16 --dt 200secs --t_end 100days --dt_save_to_disk 1days --job_id longrun_bw_rhoe_equil_highres_zalesak_ars343"
+        command: "mpiexec julia --project=examples examples/hybrid/driver.jl --ode_algo ARS343 --moist equil --microphy 0M --tracer_upwinding zalesak --z_elem 45 --dz_bottom 30 --h_elem 16 --kappa_4 1e16 --dt 400secs --t_end 100days --dt_save_to_disk 1days --job_id longrun_bw_rhoe_equil_highres_zalesak_ars343"
         artifact_paths: "longrun_bw_rhoe_equil_highres_zalesak_ars343/*"
         env:
           CLIMACORE_DISTRIBUTED: "MPI"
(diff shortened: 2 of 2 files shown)


examples/hybrid/cli_options.jl
```diff
@@ -105,13 +105,13 @@ function parse_commandline()
         arg_type = Symbol
         default = :none # TODO: change to :zalesak
         "--ode_algo"
-        help = "ODE algorithm [`ARS343`, `IMKG343a`, `ODE.Euler`, `ODE.IMEXEuler`, `ODE.Rosenbrock23` (default), etc.]"
+        help = "ODE algorithm [`ARS343` (default), `IMKG343a`, `ODE.Euler`, `ODE.IMEXEuler`, `ODE.Rosenbrock23`, etc.]"
```
charleskawczynski, line 108: Can we do this incrementally and first just change `dt` and `max_newton_iters`, and change the default tomorrow once we see that the first set of changes work? It'll also give us a heads up in terms of how to compare performance.

valeriabarra: Sure! I have done it in [c36272](https://github.com/CliMA/ClimaAtmos.jl/pull/864/commits/c36272b1bb281551f58849b3fc9ea4e32c36fa76). 
I can trigger a longrun build

szy21 (approved):
I don't fully understand the advantage of separating the change in max_newton_iter and default timestepper into two PRs, but if we want to do that this PR looks good to me.

valeriabarra:
I think Charlie meant to change the default algo tomorrow after perf is evaluated. He didn't mean in another PR. I only changed the PR title to reflect the current changes but can change it back when we put the default on ARS back :)

szy21:
I see, thanks for the explanation. I will be mostly away from the computer during the day tomorrow, so if you need someone to take a look at this PR again it would be faster to ask someone else. @jiahe23 @bischtob @Zhengyu-Huang should all be able to review it.

charleskawczynski:
Ok, it looks like there's one performance hit we take:

 - 💻 held suarez (ρe_tot) high resolution Ran in 4h 16m - ODE.Rosenbrock
 - 💻 held suarez (ρe_tot) high resolution Ran in 11h 17m - ClimaTimeSteppers.ARS343

(just a heads up, this is likely to have an impact on all of the scaling results). And, strangely,

-  💻 baroclinic wave (ρe_tot) equilmoist high resolution ars343 Ran in 18h 16m

breaks. Actually, this may just be a memory request issue. So, if we're okay with the HS perf/scaling hit, I'm fine with merging. If we merge the default algo change, then please add one job (perhaps the coarse HS ρe_tot) into CI that uses the old ode algo, for the perf team as a running reference.

Alternatively, we could do things more carefully and incrementally by switching over for all of the cases that are converging and performing on par with the Rosenbrock method (the good news is that it looks like there are some). This would be my preference.

bischtob:
This is Zhaoyi's comment from Slack: 

> For baroclinic waves, the time to solution is similar with ARS343 dt=400s and Rosenbrock23 dt=200s. This suggests that ARS343 is 2 times slower for time per timestep. This is likely from the climatimestepper implementation. It’s unlikely from the limiters because the same is true for the dry case.

charleskawczynski:
> How can that be consistent?

Hm, it looks like this may be variation in our distributed jobs (i.e., scaling bugs)

valeriabarra:
Okay, I kept only the incremental changes without changing the default for now, as discussed w/ @charleskawczynski . 

I have an appointment this afternoon, so please feel free to merge as is or add more to this one.
Thanks all

szy21:
@charleskawczynski which build is this? Just want to double check the setup.

> Ok, it looks like there's one performance hit we take:
> 
> * 💻 held suarez (ρe_tot) high resolution Ran in 4h 16m - ODE.Rosenbrock
> * 💻 held suarez (ρe_tot) high resolution Ran in 11h 17m - ClimaTimeSteppers.ARS343
> 
> (just a heads up, this is likely to have an impact on all of the scaling results). And, strangely,
> 
> * 💻 baroclinic wave (ρe_tot) equilmoist high resolution ars343 Ran in 18h 16m
> 
> breaks. Actually, this may just be a memory request issue. So, if we're okay with the HS perf/scaling hit, I'm fine with merging. If we merge the default algo change, then please add one job (perhaps the coarse HS ρe_tot) into CI that uses the old ode algo, for the perf team as a running reference.
> 
> Alternatively, we could do things more carefully and incrementally by switching over for all of the cases that are converging and performing on par with the Rosenbrock method (the good news is that it looks like there are some). This would be my preference.

charleskawczynski:
Being that our distributed performance has such a large variance, perhaps we should switch to 32 mpi procs. So, and being that the performance difference is related to distributed performance, it looks like we can switch the defaults afterall.

I'll try following up on this, but I'd like to also add at least one case with the previous timestepper to help with verifying/debugging ARS performance

valeriabarra:
I agree! I've been mentioning in multiple meetings that it's better to focus on on-node performance first, hence, advocating for using 32 processes for now. And yes, I also agree that we should keep/duplicate at least a couple of tests to use Rosenbrock, just in case.

szy21:
> > which build is this? Just want to double check the setup.
> 
> It looks like @valeriabarra's builds are no longer on buildkite, but this was the build from the main branch I was comparing against, [here](https://buildkite.com/clima/climaatmos-longruns/builds/454)

Are you refering to this [build](https://buildkite.com/clima/climaatmos-longruns/builds/456#0183679f-30b2-4915-9f59-b9d7ffbd12c4) from Valeria? It seems `longrun_hs_rhoe_highres_ars343` only takes 3h 37m. It is the job `longrun_hs_rhoe_highres` which uses Rosenbrock that takes 11h17m, which I don't know why.

I agree that we should keep one (or some) case with Rosenbrock while we are debugging ARS performance.

charleskawczynski:
Yeah, I think that was the build. Yeah, I think you're right, and I think it's because the distributed runs are not giving accurate numbers.

The non-`mpiexec` jobs are the ones we should be comparing, and all of those have increased in runtime

charleskawczynski:
That said, I'm not sure if that first commit used 1 newton iteration, let's open a new PR that uses ARS as default to see how the perf target looks

valeriabarra:
Yes, I did use 1 Newton iteration is both commits. Only changed the default back in the second one. But yeah, let's do it

PR #868, opened 2022-09-24
Change default ode solver to ARS343

charleskawczynski:
This PR changes the default ODE solver to ARS343, a continuation of #864.


charleskawczynski:
The performance difference doesn't seem egregious, so I'm fine with merging. I'm assuming that the quality of the results were reviewed in #864. So I'm fine with merging and see how things go.

valeriabarra:
I think we were also supposed to double the `dt`? (as per #863 )

PR #875, opened 2022-09-29
Unify edmf and grid mean pressure

charleskawczynski:
This PR unifies the edmf and grid-mean pressure


tc_driver/dycore.jl
```diff
@@ -397,8 +399,7 @@ function compute_gm_tendencies!(
     aux_en_f = TC.face_aux_environment(state)
     aux_up = TC.center_aux_updrafts(state)
     aux_bulk = TC.center_aux_bulk(state)
-    ρ_f = aux_gm_f.ρ
```
yairchn, line 400: is it not used ?

yairchn (approved):
left one comment - otherwise looks great

trontrytel (approved):

tc_driver/dycore.jl
charleskawczynski: Just not inside the scope of this function, it seems

PR #877, opened 2022-09-30
Yc/restructure upd tend

yairchn:
I am moving the clipping of diagnostic variables to the filter of the prognostic variables (this requires defining a small non zero value). The hope is that this reduce the number of diagonals in the Jacobian matrix


src/TurbulenceConvection/types.jl
```diff
@@ -457,6 +457,7 @@ struct EDMFModel{
 }
     surface_area::FT
     max_area::FT
+    small_area::FT
```
charleskawczynski, line 460: How does this differ from `minimum_area`?

yairchn: only if we have `>=` in the conditions. I changed it to `>` and now we can use only `a_min`

charleskawczynski (approved):

yairchn:
Bors r+

yairchn:
Bors r+

Issue #686, opened 2022-07-21
Memory allocations


simonbyrne:
Memory allocations are currently cause performance issues as they trigger the garbage collector. This is worse when using MPI, as it can be triggered at different times in different processes, causing the communication to get out of sync. For example, this profile:
<img width="1111" alt="Screen Shot 2022-07-21 at 10 23 29 AM" src="https://user-images.githubusercontent.com/1692009/180320096-183db4b5-ee76-475c-b1dc-ba881acab1f0.png">
You can see that the GC pause (in red) on the top process means that the MPI_Waitall on the bottom process takes longer (as it is waiting on data to be sent from the top process). This effect will get worse at higher core counts, hurting scaling.

# Potential solutions
## Minimize memory allocations
The memory allocations seem to be caused by a few things:
- [x] LU factorization in the linear solver (#648)
- [x] Use of `Ref`s inside `bycolumn` closures. For some reason these aren't being elided. The easiest fix seems to be to use an immutable object as a scalar container for broadcasting (will require some changes to ClimaCore).

## Synchronize calls to the garbage collector
- [ ] This gets a little bit tricky, but we can disable Julia's GC (`GC.enable(false)`), and then periodically call `GC.gc()` manually. We would need to experiment with calling full vs incremental GC (see https://bkamins.github.io/julialang/2021/06/11/vecvec.html)

tapios:
I suggest you focus on a fully explicit model first, before dealing with the implicit part (linear solver).

charleskawczynski:
Our allocation table in #836 for each `step!` looks like:
```
┌───────────────────────────────────────────────────────────────────────┬─────────────┬───────────────┐
│ <file>:<line number>                                                  │ Allocations │ Allocations % │
│                                                                       │   (bytes)   │    (xᵢ/∑x)    │
├───────────────────────────────────────────────────────────────────────┼─────────────┼───────────────┤
│ ClimaCore/jUYzD/src/Fields/broadcast.jl:82                            │   1853472   │      84       │
│ ClimaAtmos.jl/examples/hybrid/schur_complement_W.jl:248               │   221184    │      10       │
│ ClimaAtmos.jl/examples/hybrid/sphere/baroclinic_wave_utilities.jl:171 │    56000    │       3       │
│ ClimaAtmos.jl/examples/hybrid/staggered_nonhydrostatic_model.jl:780   │    55296    │       3       │
[...]

PR #880, opened 2022-10-01
Cache flux BCs

charleskawczynski:
This PR caches flux boundary conditions, for 
 - `dif_flux_uₕ`
 - `dif_flux_energy`
 - `dif_flux_ρq_tot`
to address [this comment](https://github.com/CliMA/ClimaAtmos.jl/issues/686#issuecomment-1260282007) (`Operators.SetValue(.-dif_flux_uₕ)` is allocating).

Note that it seems that the coupler may be impacted by this, due to the `coupled` flag. cc @LenkaNovak

A step towards #686.


szy21 (approved):

szy21:
Yes, this will affect the coupler I think. But it should be straightforward to change - just updating e.g. p.dif_flux_uₕ_bc instead of p.dif_flux_uₕ in the coupler.

PR #884, opened 2022-10-03
Refactor vert diff BL tendency

charleskawczynski:
This PR refactors the `vertical_diffusion_boundary_layer_tendency!` to:

 - Not use `parent`
 - Use `bycolumn` to leverage threading
 - Remove allocations


LenkaNovak:
These changes look good to me, and we'll do something similar in the coupler where we calculate fluxes! Presumably the MPI examples crash because of some other issue?

examples/hybrid/sphere/baroclinic_wave_utilities.jl
```diff
@@ -258,14 +252,16 @@ function vertical_diffusion_boundary_layer_cache(
 
     cond_type = NamedTuple{(:shf, :lhf, :E, :ρτxz, :ρτyz), NTuple{5, FT}}
 
+    normal = similar(Y.c, Geometry.WVector{FT})
```
LenkaNovak, line 255: Actually, shouldn't this be `Geometry.WVector(ones(axes(Y.c)))`, because it's a unit vector?

charleskawczynski: Ahh! That's probably why it was crashing, thank you!

szy21 (approved):

charleskawczynski:
Actually, I think they're pretty small, based on the mse tables

PR #886, opened 2022-10-04
specify cloud properites in the boundary layer for radiation

szy21:
Fixes the problem that all-sky radiation breaks in a non-deterministic way. Also adds a regression test with all-sky radiation in the buildkite.


szy21:
bors try

szy21:
bors try

PR #888, opened 2022-10-04
Increase memory for longrun

szy21:


szy21:
This might not be needed after #891

Issue #789, opened 2022-08-31
Implicit solver for the EDMF 


yairchn:
## Purpose
The goal is to run the EDMF with a timestep of ~100sec and make sure the RHS evaluation is not significantly changed. This would allow us to use EDMF in a global model.

**Jan 18th update:** The implicit EDMF simulations we have tried running were unstable. We decided to refactor the EDMF filters to improve stability before debugging the implicit implementation further. See #1104 for updates.

## People and Personnel
- Lead: @trontrytel 
- Collaborators: @dennisYatunin , @dennisYatunin , @trontrytel, @Zhengyu-Huang
- Reviewers: @charleskawczynski,  @trontrytel, @haakon-e 

## Inputs
The equations relating to this code change will be added to the following document:
https://www.overleaf.com/6233444242pddjjhxqjpyw

## Results and deliverables
- [ ] Plots showing last hour mean profiles of Bomex with explicit (dt~1sec) and implicit (dt = 100sec) EDMF. 
- [ ] Similar plots for other compressible single column cases: DYCOMS RF01, TRMM (with 0 moment microphysics).
 
## Task Breakdown And Tentative Due Date
(A preliminary list of PRs and a preliminary timeline of PRs, milestones, and key results)
- [x] Use `Covariant3Vector` for vertical velocity type (@trontrytel , @simonbyrne ) https://github.com/CliMA/ClimaAtmos.jl/pull/992 ~~Oct 19th~~, Oct 26
- [x] Fix https://github.com/CliMA/ClimaAtmos.jl/issues/1017 Oct 25
- [x] Add JFNK test to the CI pipeline: https://github.com/CliMA/ClimaAtmos.jl/pull/1080
- [x] Cleanup the way filtering is done for edmf: https://github.com/CliMA/ClimaAtmos.jl/pull/1092. Nov 9th
- [x] Solve the EDMF advection + diffusion implicitly using JFNK methods https://github.com/CliMA/ClimaAtmos.jl/pull/1021  (@dennisYatunin, @trontrytel ) ~~Oct 19th~~, ~~Oct 28~~, Nov 14th.
- [x] Create a sandbox for testing different Jacobians for edmf. We need a way to specify Jacobian blocks and a simple method to use them (preconditioner for JFNK) https://github.com/CliMA/ClimaAtmos.jl/pull/1130 (@dennisYatunin ) Nov 19th
[...]

PR #895, opened 2022-10-05
Prep unifying dycore and edmf types

charleskawczynski:
This PR moves some code into `src/`, so that we can start unifying dycore and edmf types (a step towards #789).


trontrytel:
Whats the status on this one? Should I try to revive it, or where there any issues that popped up when the code was moved?

Would be great to unify those model choices

charleskawczynski:
> Whats the status on this one? Should I try to revive it, or where there any issues that popped up when the code was moved?
> 
> Would be great to unify those model choices

I need to take a second look, I was trying to avoid adding NCDatasets as a dependency

PR #901, opened 2022-10-06
Wrap surface flux calculation into a function

szy21:
## Purpose 

Separates surface flux calculation into a function that can be used by the coupler

## Content

- Move surface flux calculation to `get_surface_fluxes!` and call this function in `vertical_diffusion_boundary_layer_tendency`

### DIFF
--- a/examples/hybrid/sphere/baroclinic_wave_utilities.jl
+++ b/examples/hybrid/sphere/baroclinic_wave_utilities.jl
@@ -381,15 +381,15 @@ function saturated_surface_conditions(
 end
 
 function vertical_diffusion_boundary_layer_tendency!(Yₜ, Y, p, t)
+    (; coupled) = p
     Fields.bycolumn(axes(Y.c.uₕ)) do colidx
+        !coupled && get_surface_fluxes!(Y, p, colidx)
         vertical_diffusion_boundary_layer_tendency!(Yₜ, Y, p, t, colidx)
     end
 end
 
-function vertical_diffusion_boundary_layer_tendency!(Yₜ, Y, p, t, colidx)
-    ᶜρ = Y.c.ρ
-    FT = Spaces.undertype(axes(ᶜρ))
-    (; z_sfc, ᶜts, ᶜp, T_sfc, ᶠK_E, C_E) = p # assume ᶜts and ᶜp have been updated
+function get_surface_fluxes!(Y, p, colidx)
+    (; z_sfc, ᶜts, T_sfc) = p
     (;
         surface_conditions,
         dif_flux_uₕ,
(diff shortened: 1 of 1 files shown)


examples/hybrid/sphere/baroclinic_wave_utilities.jl
```diff
@@ -382,11 +382,12 @@ end
 
 function vertical_diffusion_boundary_layer_tendency!(Yₜ, Y, p, t)
     Fields.bycolumn(axes(Y.c.uₕ)) do colidx
+        get_surface_fluxes(Y, p, t, colidx)
```
charleskawczynski, line 385:
```suggestion
        get_surface_fluxes!(Y, p, t, colidx)
```

szy21: Fixed, thanks!

examples/hybrid/sphere/baroclinic_wave_utilities.jl
```diff
@@ -400,23 +399,13 @@ function vertical_diffusion_boundary_layer_tendency!(Yₜ, Y, p, t, colidx)
...
     @. uₕ_int_local[colidx] = Geometry.UVVector(uₕ_int)
-    @. ᶠp[colidx] = ᶠinterp(ᶜp[colidx])
-    @. ᶠK_E[colidx] = eddy_diffusivity_coefficient(
-        C_E,
-        norm(uₕ_int),
-        z_bottom[colidx],
-        ᶠp[colidx],
-    )
 
     if !coupled
```
szy21, line 410: I think we don't need the `coupled` argument anymore. Will check with the coupler.

szy21:
I think it works now, but @charleskawczynski any idea if or why this would lead to behavior changes?

charleskawczynski (approved):

examples/hybrid/sphere/baroclinic_wave_utilities.jl
LenkaNovak: Actually I think we do.... would it be possible to make the `get_surface_fluxes!(Y, p, colidx)` call conditional based on whether it's coupled? Otherwise I think it will overwrite the fluxes the fluxes in the cache that were saved by the coupler before the atmmos step... Happy to connect to discuss! :)

LenkaNovak:
Hello, this looks great! The only thing I think we should add is the `coupled` argument in `vertical_diffusion_boundary_layer_tendency` (see comment above). Thanks very much! 🚀

examples/hybrid/sphere/baroclinic_wave_utilities.jl
szy21: Changed, thanks!

examples/hybrid/sphere/baroclinic_wave_utilities.jl
```diff
@@ -381,8 +381,9 @@ function saturated_surface_conditions(
 end
 
 function vertical_diffusion_boundary_layer_tendency!(Yₜ, Y, p, t)
+    (; coupled) = p
```
charleskawczynski, line 384: I think we may need to move this inside the `bycolumn` loop to reduce allocations when running with threading

LenkaNovak (approved):

PR #907, opened 2022-10-06
cleanup log output

simonbyrne:
## Purpose 
Cleans up the log output: moves logged values into keyword arguments (instead of the message), and combines some messages.


simonbyrne:
See here for the output:
https://buildkite.com/clima/climaatmos-ci/builds/4106#0183af85-a3d1-45b6-b8ef-e64256a313f5

charleskawczynski (approved):

PR #908, opened 2022-10-07
Add profiles plots to all TC tests, make them appear first in artifacts

trontrytel:
This PR adds the plot with final profiles of TC single column simulations to the buildkite artifacts of all TC single column test cases. (Before it was only done for Bomex). It also changes the final plot title so that it should appear first in the list of artifacts (easier to find among all the generated mp4 animations).

I think we need this for the explicit/implicit edmf work


charleskawczynski (approved):

charleskawczynski:
Hmm, maybe we forgot to reset the mse tables in another PR?

examples/hybrid/define_tc_quicklook_profiles.jl
```diff
@@ -4,7 +4,7 @@ using Plots
 function plot_tc_profiles(folder, hdf5_filename)
 
     input_filename = joinpath(folder, hdf5_filename)
-    output_filename = joinpath(folder, "final_profiles.png")
+    output_filename = joinpath(folder, "____________________________________final_profiles.png")
```
charleskawczynski, line 7: 🙀

trontrytel: lol, I have no idea how else to make the profiles plots stick out from all the animations

trontrytel:
Added some case twice to buildkite. Will rebase one more time

examples/hybrid/define_tc_quicklook_profiles.jl
charleskawczynski: Fair point

trontrytel: The formatter also does not like me 😹

trontrytel:
Hmm... @charleskawczynski I got another mse not regressed error here and I just rebased. Seems like there is something not deterministic going on? I restarted the simulation in question, but I'm pretty sure that this PR should not change anything

PR #911, opened 2022-10-07
use dycore w in w_up filter

yairchn:
I am fixing a bug in the filter that was introduced in #877 for the filtering of updraft w

### DIFF
--- a/src/TurbulenceConvection/EDMF_functions.jl
+++ b/src/TurbulenceConvection/EDMF_functions.jl
@@ -799,6 +799,7 @@ function filter_updraft_vars(
 
     prog_up = center_prog_updrafts(state)
     prog_gm = center_prog_grid_mean(state)
+    prog_gm_f = face_prog_grid_mean(state)
     aux_gm_f = face_aux_grid_mean(state)
     aux_gm = center_aux_grid_mean(state)
     aux_up = center_aux_updrafts(state)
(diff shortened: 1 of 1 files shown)


charleskawczynski (approved):

trontrytel:
Looks like DYCOMS buoyancy, temperature and vertical velocity above the cloud top are very unhappy with the proposed changes. Bomex and TRMM look fine, so it's a little bit of a mystery right now for me. 

Before:

Screenshot 2022-10-09 at 18 46 39


After:

Screenshot 2022-10-09 at 18 44 08


I'll try setting it to zero to see what happens.

trontrytel:
oh wow! Looks like no changes if we set it to zero 🎉 
Should we just o with that one @yairchn  (though I still don't understand why it even matters....)

yairchn:
Thinking again about the problem. It might be that the value we set here has to be used in this type of filtering below 
`@. prog_up[i].ρarea =
            ifelse(Ic(prog_up_f[i].ρaw) <= 0, FT(0), prog_up[i].ρarea)`

trontrytel:
Do you want me to test anything else or should we just merge it as is

yairchn:
> Do you want me to test anything else or should we just merge it as is

we can try using the grid mean velocity, but also use it instead of the `0` I n these type of filtering 

`@. prog_up[i].ρarea = ifelse(Ic(prog_up_f[i].ρaw) <= 0, FT(0), prog_up[i].ρarea)` -->
`@. prog_up[i].ρarea = ifelse(Ic(prog_up_f[i].ρaw) <= w_gm, FT(0), prog_up[i].ρarea)`

trontrytel:

PR #913, opened 2022-10-07
add surface density and humidity to the cache

szy21:
## Purpose 

Adds surface density and humidity to the cache, which enables the coupler to use `get_surface_flux!`.

## To-do

It might be better to calculate surface density and humidity in `get_surface_flux!` rather than `saturated_surface_conditions`, but either way is fine I think.

## Content

- Adds surface density and humidity to the cache and updates them when not coupled
- Wrap the calculation of surface density and humidity into a function


LenkaNovak:
Thank you! Just testing it in the coupler...

charleskawczynski (approved):

trontrytel:
@charleskawczynski - would that be a good time to cache the thermo state at the surface. EDMF will also need some surface things. I think we were still debating how to go about accessing them

szy21:
I think as Tapio suggested before thermo state of the surface is just not very physical, and that's why I was a bit reluctant to cache it. But if it makes coding easier, we can easily cache it. I don't have a strong preference.

trontrytel:
> I think as Tapio suggested before thermo state of the surface is just not very physical, and that's why I was a bit reluctant to cache it. But if it makes coding easier, we can easily cache it. I don't have a strong preference.

Sounds good. I think we are just used to interfacing with thermodynamics via the thermo state constructor. But I see the point of it being hard to have a solid (har har) thermo state at the ground. Lets not do it then and just be careful about caching the surface variables consistently

charleskawczynski:
I don't have a strong preference either way. It seems that we at least compute the thermo state at the surface. I suppose that the criteria of whether it should be cached or not is: will we use/need it more than once? Also, maybe we can leave it for another PR if it's not relevant to what's needed in the PR which, from what I gathered from @szy21 is mostly about making things work easier in the coupler. So I'm fine waiting on caching it if that's easier

szy21:
Yes, Lenka is already testing this change in the coupler and it seems to work, so I will go ahead and merge this PR. We can have another PR to cache surface thermo state whenever we think it is necessary.

PR #917, opened 2022-10-08
Cache some edmf variables

charleskawczynski:
This PR caches some edmf variables to fix some allocations.


trontrytel (approved):
I'll force push, rebase and try to merge it after the kinetic energy PR lands. Hopefully this time MPI test wont run out of memory and this will be herded in

PR #918, opened 2022-10-08
Change updraft momentum equation to affective form

yairchn:
After #914 I am starting a new PR for advective updraft vertical velocity. I have removed `w` from the auxiliary variables in the updrafts and made the corresponding changes. Most notable changes in the PR are the change of the updraft `w` equation in `compute_updraft_tendencies` and the corresponding changes to the non hydrostatic pressure term, which entails some change is the non hydrostatic pressure tendency in the TKE equation (in `update_aux.jl`).

PR #923, opened 2022-10-10
exploding head emoji

yairchn:
@charleskawczynski help me out here!

### DIFF
--- a/src/TurbulenceConvection/EDMF_functions.jl
+++ b/src/TurbulenceConvection/EDMF_functions.jl
@@ -778,7 +778,7 @@ function compute_up_tendencies!(
         @. tends_w +=
             w * I0f(entr_w) * (w_en - w_up) +
             I0f(buoy) +
-            nh_pressure / (ρ_f * Iaf(max(a_up, a_min)))
+            nh_pressure * Iaf(a_up) / (ρ_f * Iaf(max(a_up, a_min)))
         tends_w[kf_surf] = 0
     end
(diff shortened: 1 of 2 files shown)


src/TurbulenceConvection/EDMF_functions.jl
```diff
@@ -778,7 +778,7 @@ function compute_up_tendencies!(
         @. tends_w +=
             w * I0f(entr_w) * (w_en - w_up) +
             I0f(buoy) +
-            nh_pressure / (ρ_f * Iaf(max(a_up, a_min)))
+            nh_pressure * Iaf(a_up) / (ρ_f * Iaf(max(a_up, a_min)))
```
trontrytel, line 781: Did you try just getting rid of `a_up` here?

yairchn: yes, that related to the comment in #918

trontrytel: Then how about having the max(a_up, a_min) in both nominator and denominator? I'm just guessing wildly here

yairchn: the main issue is not that I multiply and divide by `a_up` is that multiplying inside `compute_nh_pressure` is different than multiplying in the equation ....

yairchn:
I forgot that `nh_press` is also used in the TKE equation.

PR #927, opened 2022-10-10
plots vs main

haakon-e:
## Purpose 
Plot EDMF profiles from main branch in buildkite runs in a PR. 

## To-do
- [x] find out how buildkite saves output.
- [x] Find out how to consistently get the most recent (latest) ref counter data-folder
- [x] Plot edmf static profiles

## Outstanding issues for a future PR:
- [ ] Plot profiles from TRMM longrun (requires saving longrun hfd5 output)
- [ ] Plot also LES lines on profiles (requires storing LES data somewhere)
- [ ] Plot edmf movies (requires saving timeseries as hdf5 data)

### DIFF
--- a/examples/hybrid/define_tc_quicklook_profiles.jl
+++ b/examples/hybrid/define_tc_quicklook_profiles.jl
@@ -1,102 +1,91 @@
 import ClimaCore: Fields, InputOutput
 using Plots
 
-function plot_tc_profiles(folder, hdf5_filename)
+function plot_tc_profiles(folder; hdf5_filename, main_branch_data_path)
 
-    input_filename = joinpath(folder, hdf5_filename)
-    output_filename = joinpath(
-        folder,
-        "____________________________________final_profiles.png",
-    )
-
-    reader = InputOutput.HDF5Reader(input_filename)
-    Y = InputOutput.read_field(reader, "Y")
-    D = InputOutput.read_field(reader, "diagnostics")
-
-    zc = Fields.coordinate_field(Y.c).z
-    zf = Fields.coordinate_field(Y.f).z
+    PR_filename = joinpath(folder, hdf5_filename)
+    main_filename = joinpath(main_branch_data_path, hdf5_filename)
 
     args =
         (; tickfontsize = 13, guidefontsize = 16, legendfontsize = 10, lw = 3)
 
-    p1 = plot(
-        parent(D.bulk_up_area)[:],
-        parent(zc)[:],
-        label = "up area";
-        args...,
-    )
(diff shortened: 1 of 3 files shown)


trontrytel (approved):
Thank you!

examples/hybrid/define_tc_quicklook_profiles.jl
```diff
@@ -1,102 +1,133 @@
...
 
-    input_filename = joinpath(folder, hdf5_filename)
-    output_filename = joinpath(
-        folder,
-        "____________________________________final_profiles.png",
-    )
+Return the folder of the most recent buildkite run that contains a `ref_counter.jl` file,
+and therefore also contains hdf5 files with output data.
+"""
+function get_main_branch_buildkite_path()
```
charleskawczynski, line 12: This looks similar to `self_reference_or_path`. I'm wondering if we should move it into the same file? (and maybe later rename / move it)

haakon-e: yes, it does basically the same thing, except that it fetches the _most recent_ data on the highest ref_counter, as opposed to the oldest data on the highest ref_counter.
I can move it in there as long as it's okay to do `include(../../regression_tests/self_reference_or_path.jl` in the `examples/hybrid/driver.jl` file.

haakon-e:
> Overall looks good, the only thing I'm thinking is if we should put `get_main_branch_buildkite_path` into a different location, since it's not really TC-specific.

any suggestions where to move it? Your suggestion in the comment seems weird to me as I'd have to include code from `regression_tests` into `examples/hybrid`, but if you're fine with that I'll do it.

examples/hybrid/define_tc_quicklook_profiles.jl
charleskawczynski: > I can move it...

I think that's fine, we use

```julia
include(
        joinpath(
            @__DIR__,
            "..",
            "..",
            "regression_tests",
            "regression_tests.jl",
        ),
    )
```
above in the driver.

haakon-e: done.

charleskawczynski: One thing I just realized is that the data on the most recent ref counter is not guaranteed to exist, because we only save data to central if the ref counter has changed. Can we cycle back on how this works, @haakon-e?

charleskawczynski: Ah, I see that `get_main_branch_buildkite_path` is no longer referenced

PR #932, opened 2022-10-11
move all clipping to affect_filter

yairchn:
I am retrying to move the filtering operators to `update_aux` - lets be extra careful with merging this time. 
I think I found a way to handle some of the strange spikes that came up above the clouds when the filters are changed. A factor 2 is needed for the minimum area value in the filtering of updraft w [here](https://github.com/CliMA/ClimaAtmos.jl/blob/1d5d6277f08ac6dc637db13c5c874a21330c0e71/src/TurbulenceConvection/EDMF_functions.jl#L837) as that area is coming from an interpolation of the area from cell to face, which at the top of the updraft is like dividing  the area fraction by half.


regression_tests/mse_tables.jl
```diff
@@ -127,6 +127,7 @@ all_best_mse["compressible_edmf_trmm"][(:c, :turbconv, :en, :ρatke)] = 0.0
 all_best_mse["compressible_edmf_trmm"][(:c, :turbconv, :up, 1, :ρarea)] = 0.0
 all_best_mse["compressible_edmf_trmm"][(:c, :turbconv, :up, 1, :ρaθ_liq_ice)] = 0.0
 all_best_mse["compressible_edmf_trmm"][(:c, :turbconv, :up, 1, :ρaq_tot)] = 0.0
+all_best_mse["compressible_edmf_trmm"][(:f, :turbconv, :up, 1, :w)] = 0.0
```
charleskawczynski, line 130:
```suggestion
```
This is exactly zero and, unfortunately, breaks the regression infrastructure.

trontrytel:

PR #933, opened 2022-10-11
Compute velocities before they are used

charleskawczynski:
This PR is an attempt to fix a bug: we currently use these velocities before they are computed. The result is that, on main, velocities from the previous timestep are used, rather than the current timestep.


yairchn (approved):
LGTM

PR #934, opened 2022-10-11
Add edmf consistency test

charleskawczynski:
This PR adds an edmf consistency test. It works by running Bomex and first setting all the edmf auxiliary variables to NaNs, which should work so long we compute everything in the correct order.

Since it's not yet working, I've labeled the job as a `soft_fail`, so that we can chip away until it's working and change it to a test proper.


trontrytel:
This is great. Thank you!

PR #947, opened 2022-10-13
Fix conversions

charleskawczynski:
Supersedes #946.


trontrytel:
> Ok, it looks like the environment velocities are noticeably different in TRMM. I think this should be fine, but would be nice to get confirmation, cc @trontrytel

They are different in all our compressible cases aside from Bomex... This is very strange. Possibly a bug somewhere

haakon-e:
* Compressible single column EDMF Bomex: results are virtually identical to main branch.
* Compressible single column EDMF DYCOMS_RF01:  results are virtually identical to main branch.
* Compressible single column EDMF TRMM_LBA: `main en w` is max 0.08;  `PR en w` is max 0.0004. But profile is otherwise identical.
    * could this be due to plotting of physical vs covariant velocities? We should double-check.
    * I think labeling of `up w` and `env w` might be switched. CHECK.
* Compressible single column EDMF GABLS: `main en w` is min -0.001;  `pr en w` is min -0.00002.
    * Check same plotting issue?

haakon-e:
Q: Is `profile_f_w_1.mp4` the grid-mean vertical velocity?

On main, the max of that is O(100) smaller than `en w`. In the PR, they are exactly the same.

trontrytel:
After some more discussions with @haakon-e we think that the PR is correct. Meaning that the O(100) change is correct: 
 - It's the same order of magnitude change we saw when fixing environmental kinetic energy and velocity in the sphere case.
 - It's only affecting the cached (wrongly diagnosed) variables and not the prognostic state variables.
 - The animations based on state look consistent with the new plots.
 - We don't see it in Bomex and Dycoms plots because there the axis scale is set by the correct prognostic updraft velocity and not the wrongly diagnosed envirionmental velocity
 
 
 Bottom line is. We want to add a PR that plots updraft and environmental velocity separately before merging this PR. This way we will see if its the same fix for all 4 cases and consistent with animations based on the state

trontrytel:
Looking at the plots one more time with @haakon-e . The Bomex and Dycoms ones are not affected by the fix, because of the additional subsidence forcing. The TRMM and GABLS ones are now fix and match the animation with the grid mean velocity. 

Hence we are ok with the changes in the PR 🎉

trontrytel (approved):

PR #821, opened 2022-09-07
make GC deterministic in distributed

simonbyrne:
# PULL REQUEST

## Purpose and Content
This should reduce MPI Waitall time by manually triggering the GC across all processes at the same time.

## Benefits and Risks
The number of steps will require some tuning to avoid out-of-memory errors

## Linked Issues
- Item 3 of #635 
- Mentioned in #686
- Supersedes #687

## PR Checklist
- [x] This PR has a corresponding issue OR is linked to an SDI.
- [x] I have followed CliMA's codebase [contribution](https://clima.github.io/ClimateMachine.jl/latest/Contributing/) and [style](https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/) guidelines OR N/A.
- [x] I have followed CliMA's [documentation policy](https://github.com/CliMA/policies/wiki/Documentation-Policy).
- [x] I have checked all issues and PRs and I certify that this PR does not duplicate an open PR.
- [x] I linted my code on my local machine prior to submission OR N/A.
- [x] Unit tests are included OR N/A.
- [x] Code used in an integration test OR N/A.
- [x] All tests ran successfully on my local machine OR N/A.
- [x] All classes, modules, and function contain docstrings OR N/A.
- [x] Documentation has been added/updated OR N/A.

PR #950, opened 2022-10-13
Re-enable GC callbacks

simonbyrne:
## Purpose 
Adds GC callbacks. Behavior can be controlled by an environment variable `CLIMAATMOS_GC_NSTEPS` (default = every 1000 steps).

Unlike #821, this doesn't disable the GC, but by manually calling it we should be able to avoid the runtime needing to call it.

## To-do
- waiting on https://buildkite.com/clima/climaatmos-scaling/builds/332

### DIFF
--- a/examples/hybrid/driver.jl
+++ b/examples/hybrid/driver.jl
@@ -262,6 +262,8 @@ end
 
 if simulation.is_distributed
     OrdinaryDiffEq.step!(integrator)
+    #GC.enable(false)
+    GC.gc()
     ClimaComms.barrier(comms_ctx)
     if ClimaComms.iamroot(comms_ctx)
         @timev begin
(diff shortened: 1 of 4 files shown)


examples/hybrid/driver.jl
```diff
@@ -262,6 +262,8 @@ end
 
 if simulation.is_distributed
     OrdinaryDiffEq.step!(integrator)
+    #GC.enable(false)
```
charleskawczynski, line 265: Do we want to uncomment this?

simonbyrne: no, it's causing a memory leak

simonbyrne:
```suggestion
    # GC.enable(false) # disabling GC causes a memory leak
```

charleskawczynski (approved):

PR #952, opened 2022-10-14
change default post processing

szy21:
## Purpose 

Changes default post processing to held suarez paper plots because those are more helpful

## Content

- Use `paperplots_held_suarez` for most of the experiments
- Add a case `is_solid_body` which uses `postprocessing`, so we exercise the code in the ci


szy21:
bors try

PR #956, opened 2022-10-14
add drag coefficient and roughness to cache

szy21:
## Purpose 

Cache drag coefficient and roughness as 2D fields. This is needed by the coupler.

### DIFF
--- a/examples/hybrid/sphere/baroclinic_wave_utilities.jl
+++ b/examples/hybrid/sphere/baroclinic_wave_utilities.jl
@@ -241,6 +241,7 @@ function vertical_diffusion_boundary_layer_cache(
             zeros(axes(z_bottom)),
             zeros(axes(z_bottom)),
         )
+    dif_flux_energy = similar(z_bottom, Geometry.WVector{FT})
     dif_flux_ρq_tot = if :ρq_tot in propertynames(Y.c)
         similar(z_bottom, Geometry.WVector{FT})
     else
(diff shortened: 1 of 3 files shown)


examples/hybrid/sphere/baroclinic_wave_utilities.jl
```diff
@@ -303,9 +319,10 @@ function saturated_surface_conditions(
     params,
     coupled,
 )
-    Cd = surface_scheme.Cd
-    Ch = surface_scheme.Ch
-    # parameters
+    Cd = parent(p.Cd[colidx])[1]
```
szy21, line 322: If I don't do this I got `Cannot convert an object of type Field{} (trunc disp) to an object of type Float32`

LenkaNovak: Hmm would it be better to pass in Cd and Ch as en explicit argument to `saturated_surface_conditions`? Maybe in the long term, we can separate the cache into fields and other objects, so we can just pass in `p.fields_vector[colidx]`, but that change may break many things, so probably not worth it right now.

szy21:
Oh this is allocating a lot...https://buildkite.com/clima/climaatmos-ci/builds/4408#0183d822-6472-4081-9183-e01830927310/140-186 @charleskawczynski is there anything I did wrong in the cache?

examples/hybrid/sphere/baroclinic_wave_utilities.jl
charleskawczynski, line 322:
```suggestion
    Cd = Spaces.level(p.Cd[colidx], 1)
```

examples/hybrid/sphere/baroclinic_wave_utilities.jl
```diff
@@ -303,9 +319,10 @@ function saturated_surface_conditions(
...
+    Ch = parent(p.Ch[colidx])[1]
```
charleskawczynski, line 323:
```suggestion
    Ch = Spaces.level(p.Ch[colidx], 1)
```

examples/hybrid/sphere/baroclinic_wave_utilities.jl
```diff
@@ -353,8 +372,9 @@ function saturated_surface_conditions(
     params,
     coupled,
 )
-    z0m = surface_scheme.z0m
-    z0b = surface_scheme.z0b
+    z0m = parent(p.z0m[colidx])[1]
+    z0b = parent(p.z0b[colidx])[1]
```
charleskawczynski, line 376:
```suggestion
    z0m = Spaces.level(p.z0m[colidx], 1)
    z0b = Spaces.level(p.z0b[colidx], 1)
```

examples/hybrid/sphere/baroclinic_wave_utilities.jl
charleskawczynski: I think we shouldn't ever need parent anymore

charleskawczynski (approved):

PR #957, opened 2022-10-14
fix call_every_dt logic

simonbyrne:
if the dt in the callback is less than the timestep, it will be called multiple times per timestep. This changes it so that it will be called at most once per timestep.

## Purpose 
Fixes https://github.com/CliMA/ClimaTimeSteppers.jl/issues/75

### DIFF
--- a/examples/hybrid/callbacks.jl
+++ b/examples/hybrid/callbacks.jl
@@ -79,13 +79,13 @@ end
 function call_every_dt(f!, dt; skip_first = false, call_at_end = false)
     next_t = Ref{typeof(dt)}()
     affect! = function (integrator)
+        f!(integrator)
+
         t = integrator.t
         t_end = integrator.sol.prob.tspan[2]
-        while t >= next_t[]
-            f!(integrator)
-            next_t[] =
-                (call_at_end && t < t_end) ? min(t_end, next_t[] + dt) :
-                next_t[] + dt
+        next_t[] = max(t, next_t[] + dt)
+        if call_at_end
+            next_t[] = min(next_t[], t_end)
         end
     end
     return ODE.DiscreteCallback(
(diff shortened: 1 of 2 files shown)


charleskawczynski:
> if the dt in the callback is less than the timestep

What's the difference between dt and timestep?

> it will be called multiple times per timestep

The callback will be called multiple times per timestep?

simonbyrne:
The problem is that in the callbacks profiling job, you set the callback dt to be 1 second, whereas the model dt is much larger (say 150 seconds). So doing one step will call the callback 150 times.

This changes the logic so that a callback will be called at most once per timestep, no matter how small the callback dt is.

charleskawczynski (approved):

charleskawczynski:
>  whereas the model dt is much larger (say 150 seconds).

Where is the model dt defined?

simonbyrne:
The model `dt` default is here:
https://github.com/CliMA/ClimaAtmos.jl/blob/8d005397a4456dcefcb34a87cff9e403041c13f3/examples/hybrid/cli_options.jl#L13-L16
(it is 600 seconds)

In the callbacks flame graph job, you set the following:
https://github.com/CliMA/ClimaAtmos.jl/blob/8d005397a4456dcefcb34a87cff9e403041c13f3/.buildkite/pipeline.yml#L338

charleskawczynski:
> The model `dt` default is here:

Isn't that the timestep? I guess I'm just still confused about timestep vs model dt.

simonbyrne:
yes, timestep is the model dt

charleskawczynski:
I can confirm that this PR fixes things (ran in [this build](https://buildkite.com/clima/climaatmos-ci/builds/4412)).

charleskawczynski:
Upgrading CTS will give us some nice allocation improvements:
```
[ Info: `allocs (flame_perf_target_rhoe)`: 143632
[ Info: TODO: lower `allocs_limit[flame_perf_target_rhoe]` to: 143632
[ Info: Allocation change (allocs/allocs_limit): 0.26246235710317806              # 26% allocations compared to main 🎉
```

Issue #949, opened 2022-10-13
Explicit EDMF on a sphere


charleskawczynski:
## Purpose
The goal is to add a spherical baroclinic wave test case using rhoe, equilibrium moisture, 0-moment microphysics and edmf turbulence closure. The simulation should be added to the CI and should be able to run for ~20minutes total CI time. 

## Cost/Benefits/Risks
This is essential for the final model. We could also expand on this setup and later have CI cases in a multi-column box configuration for testing edmf advection terms and topography.

## People and Personnel
- Lead: @trontrytel, @dennisYatunin  
- Collaborators: @haakon-e  
- Reviewers: @tapios , @charleskawczynski, @szy21 

## Results and deliverables
 - [x] A buildkite job with edmf on a sphere running for 20 minutes computational time.
 - [x] Add edmf to the perf target (flame graph).

## Task Breakdown And Tentative Due Date
 - [x] Fix AxisTensor (metric terms) conversions (#947)
 - [x] Refactor case-specific radiation for SCM vs GCM runs (@charleskawczynski , @trontrytel) #960 #959, https://github.com/CliMA/ClimaAtmos.jl/pull/974, https://github.com/CliMA/ClimaAtmos.jl/pull/960, https://github.com/CliMA/ClimaAtmos.jl/pull/971, https://github.com/CliMA/ClimaAtmos.jl/pull/968, https://github.com/CliMA/ClimaAtmos.jl/pull/966, https://github.com/CliMA/ClimaAtmos.jl/pull/964, https://github.com/CliMA/ClimaAtmos.jl/pull/959
 - [x] Unify thermo state (@charleskawczynski )
 - [x] Refactor case-specific subsidence and LS advection for SCM vs GCM runs, (@charleskawczynski , @trontrytel) #1022, #1018
 - [x] Refactor case-specific surface BC  for SCm vs GCM runs https://github.com/CliMA/ClimaAtmos.jl/pull/1147 (@charleskawczynski , @trontrytel)
 - [x] Refactor precipitation treatment between the TC and CA, https://github.com/CliMA/ClimaAtmos.jl/pull/1149,  https://github.com/CliMA/ClimaAtmos.jl/pull/1167, https://github.com/CliMA/ClimaAtmos.jl/pull/1108 (@trontrytel , @charleskawczynski)
 - [x] Add a buildkite job with edmf on a sphere https://github.com/CliMA/ClimaAtmos.jl/pull/916 ~~Oct 28~~
[...]

PR #960, opened 2022-10-15
Move tc radiation into dycore

charleskawczynski:
This PR moves the edmf radiation into the dycore, so that TC does less overwriting of dycore components. A step towards #949.

TODO:
 - [x] Add cache for TC radiation structs
 - [x] Extend cache appropriately based on radiation model
 - [x] Don't use TC internals (grid, state) for dycoms radiation
 - [x] Don't use TC internals (grid, state) for TRMM radiation


trontrytel:
Looks like we are still changing the DYCOMS clouds quite a bit. Should this all be non-behavior changing?

charleskawczynski:
Hmm, that's still results in noticeable changes. 🤔

charleskawczynski:
> Should this all be non-behavior changing?

Yes, basically within machine precision

trontrytel:
Can't spot it in the PR - is the DYCOMS radiation applied at every timestep?

charleskawczynski:
Ahhh, we haven't moved the grid-mean computations out yet, let me try a patch

charleskawczynski:
Hmm, I'm not sure what's going on, I'm going to try breaking this down further

charleskawczynski:
Turned out that the issue was a typo I made `Y` vs `Yₜ`..

PR #962, opened 2022-10-15
Update manifests

charleskawczynski:
This PR updates our dependencies


simonbyrne:
This includes an upgrade to MPI.jl, which requires a different configuration, see https://juliaparallel.org/MPI.jl/stable/configuration/

PR #963, opened 2022-10-17
refactor remapping pipeline to share the weights

szy21:
## Purpose 

Refactor the remapping pipleline so the weightfile is only created once.

## To-do

I don't think this makes it much faster. Will still need to do parallel remapping in another PR.

### DIFF
--- a/examples/hybrid/sphere/remap_pipeline.jl
+++ b/examples/hybrid/sphere/remap_pipeline.jl
@@ -4,14 +4,10 @@ using ClimaCore:
 using NCDatasets
 using ClimaCoreTempestRemap
 
-using JLD2
-
-if haskey(ENV, "JLD2_DIR")
-    data_dir = ENV["JLD2_DIR"]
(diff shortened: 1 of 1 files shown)


charleskawczynski (commented):
I definitely approve the overall idea of writing this in a way that we can share the weights, but I think there are a few changes that we can make to make this a bit more flexible. Alright if I push a commit?

szy21:
> I definitely approve the overall idea of writing this in a way that we can share the weights, but I think there are a few changes that we can make to make this a bit more flexible. Alright if I push a commit?

Yes please feel free! I will be in a symposium today so will only get to this later tonight I think.

charleskawczynski (approved):
Actually, I don't think that this is really tested in our CI, so if you've tested this locally and it works, then I'm fine with merging as is.

Ultimately I'd like to add a convenience wrapper in ClimaCoreTempestRemap that does this sort of thing

PR #968, opened 2022-10-17
Remove TC internals in thermo state computation

charleskawczynski:
Peel off from #966


charleskawczynski:
Results in [this build](https://buildkite.com/clima/climaatmos-ci/builds/4488) look the same within machine precision. Merging.

Issue #967, opened 2022-10-17
Possible bug in advective w equation 


trontrytel:
Our flux form equation looked like this:

`@. tends_ρaw = -(div_f(wvec(LBC(ρaw * w_up))))`

After changing to advective form we now have:

`@. tends_w = -(div_f(wvec(LBC(w_up * w_up))))`

But shouldn't it be something more like `w_up dot grad (w_up)`?

And a more general question. Aren't we missing here horizontal transport terms? For example looking at DesignDocs for ClimaNumerics or ClimaAtmos docs for momentum: https://clima.github.io/ClimaAtmos.jl/dev/equations/#Momentum
we could write this with kinetic energy and worticity. But that would use full velocity vector, not just the vertical component...

@tapios , @costachris @ilopezgp @szy21

trontrytel:
Trying: https://github.com/CliMA/ClimaAtmos.jl/pull/969

tapios:
> Our flux form equation looked like this:
> 
> `@. tends_ρaw = -(div_f(wvec(LBC(ρaw * w_up))))`
> 
> After changing to advective form we now have:
> 
> `@. tends_w = -(div_f(wvec(LBC(w_up * w_up))))`
>

You can do this (but factor 1/2 is missing).

> But shouldn't it be something more like `w_up dot grad (w_up)`?

> And a more general question. Aren't we missing here horizontal transport terms? For example looking at DesignDocs for ClimaNumerics or ClimaAtmos docs for momentum: https://clima.github.io/ClimaAtmos.jl/dev/equations/#Momentum we could write this with kinetic energy and worticity. But that would use full velocity vector, not just the vertical component...

Yes! We should write the vertical momentum equation exactly like the vertical (u_3) component of the momentum equation in the dycore, which has kinetic energy and vorticity. It will have horizontal advection terms (appearing in vorticity and kinetic energy) too, which are just zero in the SCM setting. But we should include them to be able to include the parameterization in the GCM, and simply set them to 0 for SCM testing. 

I didn't get around to writing the equations down yet, but this would be the essence: write them like the dycore momentum equations, which also split out the vertical parts for implicit timestepping. 

> @tapios , @costachris @ilopezgp @szy21

trontrytel:
Closed via https://github.com/CliMA/ClimaAtmos.jl/pull/969

PR #969, opened 2022-10-17
try updraft kinetic energy based on w_up

trontrytel:
See https://github.com/CliMA/ClimaAtmos.jl/issues/967


trontrytel:
The results for single column tests in the CI are not bad. I'm testing now how it affects the long run TRMM

charleskawczynski (approved):
Code changes look fine to me. Should we delete ∇f if it's no longer being used?

PR #970, opened 2022-10-18
Narrow scope of allocation table

charleskawczynski:
This PR narrows the scope of the allocation table in hopes to speed up CI times, while still keeping an eye on the remaining top allocation sites


charleskawczynski:
Hmm, this doesn't seem to be speeding up the allocation table job at all.

PR #975, opened 2022-10-18
Shorten t_end for compressible TRMM longruns

costachris:
## Purpose 
Shorten t_end to allow TRMM long runs to finish.


trontrytel:
Temporary change so that we have some plots from the long-runs. 

We still need to fix it to be able to run full 6 hours.

PR #984, opened 2022-10-19
Enable dispatching thermo state on compressibility model

charleskawczynski:
This PR moves the anelastic thermo state computations into `thermo_state!`, where the dycore computes the grid-mean thermo state. This will make moving other edmf pieces out into the dycore, where they belong.


charleskawczynski:
Note to self: move precompute outside of compressible if-statement

charleskawczynski:
Results look good!

PR #987, opened 2022-10-20
add surface temperature and qt to diagnostics

szy21:
## Purpose 

Adds surface temperature and qt to diagnostics. This will be helpful for debugging.


szy21:
bors try

PR #989, opened 2022-10-20
add noniterative option for surface flux

szy21:
## Purpose 

Adds noniterative option for surface flux, so it can work with Float32


szy21:
@akshaysridhar could you change if the code change here to use noniterative solution is correct? Thanks!

PR #972, opened 2022-10-18
Aj/covariant3vec

trontrytel:
## Purpose 
The goal is to change the vertical velocity type in CATs (Clima Atmos Turbulence Convection :) to be `Covariance3Vector` instead of a scalar. 

This is linked to SDI: https://github.com/CliMA/ClimaAtmos.jl/issues/789

## To-do
- [x] Make the code compile
- [ ] Make the simulation results unchanged

PR #992, opened 2022-10-20
covariant w for edmf

simonbyrne:
## Purpose 
Updates EDMF variables to use `Covariant3Vector` for vector quantities

Trying to do this incrementally to avoid breaking stuff.

See also #972

## To-do

Still need to do the closures themselves.


simonbyrne:
@trontrytel @charleskawczynski do we have an EDMF test case with a non-uniform grid?

charleskawczynski:
> @trontrytel @charleskawczynski do we have an EDMF test case with a non-uniform grid?

We do in TurbulenceConvection.jl ([this job](https://github.com/CliMA/TurbulenceConvection.jl/blob/2bfeb1b543acf6271a00348ba50e0f21456d9306/.buildkite/pipeline.yml#L111)), but I think we need [CC#821](https://github.com/CliMA/ClimaCore.jl/pull/821) to add it back in cleanly (I think I removed the logic because unifying it with the ClimaAtmos mesh init was getting too messy).

simonbyrne:
Okay, in that case perhaps it would be worth merging this in as-is. I haven't tackled the closures themselves, but perhaps that can be another PR.

trontrytel:
The anelastic results look great (aka unchanged). 

For some reason none of the compressible simulations were able to generate plots. Looking at the CI logs, each threw a different error. I'm just going to update the branch and check again. 

I'm wondering if there is something to be fixed in how we output the velocity for plotting the animations between the scalar and the covariant version?

simonbyrne:
It could be because I deleted a diagnostic variable? I can remove that change if it's easier

trontrytel:
> It could be because I deleted a diagnostic variable? I can remove that change if it's easier

Probably. I think some of the plots are done based on what is in cache.

trontrytel:
btw, what is the difference between this PR and what we tried in PR https://github.com/CliMA/ClimaAtmos.jl/pull/972 ?

simonbyrne:
I'm honestly not sure: I basically tried making the same changes, but only 1 variable at a time and made sure BOMEX was unchanged at each step.

trontrytel:
The plots are working now for both compressible and anelastic CI 🎉 

We should increase the ref counter to update the reference profiles after this PR. I'll push to this branch.

simonbyrne:
Thanks!

examples/hybrid/callbacks.jl
```diff
@@ -282,7 +282,6 @@ function save_to_disk_func(integrator)
             bulk_up_qt_tendency_precip_formation = tc_cent(
                 p,
             ).bulk.qt_tendency_precip_formation,
-            env_w = tc_cent(p).en.w,
```
trontrytel, line 285: Could we keep it? I think we use it in debugging plots.

src/TurbulenceConvection/EDMF_functions.jl
```diff
@@ -271,29 +265,27 @@ function compute_diffusive_fluxes(
...
         aeKHq_liq_bc = FT(0)
         aeKHq_ice_bc = FT(0)
 
-        ∇q_liq_en = CCO.DivergenceC2F(;
-            bottom = CCO.SetDivergence(aeKHq_liq_bc),
-            top = CCO.SetDivergence(FT(0)),
+        ∇q_liq_en = CCO.GradientC2F(;
+            bottom = CCO.SetGradient(CCG.WVector(aeKHq_liq_bc)),
+            top = CCO.SetGradient(CCG.Covariant3Vector(FT(0))),
         )
```
trontrytel, line 279: Why do we want our bottom condition to be based on `CCG.WVector` but the top to be based on `CCG.Covariant3Vector` ?

trontrytel (approved):

charleskawczynski:
This looks good, I'll go ahead and rebase and merge!

PR #1002, opened 2022-10-21
Try dss2

sriharshakandala:
## Purpose 
Move to new DSS implementation with optimized halo exchange.

### DIFF
--- a/examples/hybrid/callbacks.jl
+++ b/examples/hybrid/callbacks.jl
@@ -123,12 +123,12 @@ function dss_callback(integrator)
     Y = integrator.u
     ghost_buffer = integrator.p.ghost_buffer
     @nvtx "dss callback" color = colorant"yellow" begin
-        Spaces.weighted_dss_start!(Y.c, ghost_buffer.c)
-        Spaces.weighted_dss_start!(Y.f, ghost_buffer.f)
-        Spaces.weighted_dss_internal!(Y.c, ghost_buffer.c)
-        Spaces.weighted_dss_internal!(Y.f, ghost_buffer.f)
-        Spaces.weighted_dss_ghost!(Y.c, ghost_buffer.c)
-        Spaces.weighted_dss_ghost!(Y.f, ghost_buffer.f)
+        Spaces.weighted_dss_start2!(Y.c, ghost_buffer.c)
+        Spaces.weighted_dss_start2!(Y.f, ghost_buffer.f)
+        Spaces.weighted_dss_internal2!(Y.c, ghost_buffer.c)
+        Spaces.weighted_dss_internal2!(Y.f, ghost_buffer.f)
+        Spaces.weighted_dss_ghost2!(Y.c, ghost_buffer.c)
+        Spaces.weighted_dss_ghost2!(Y.f, ghost_buffer.f)
     end
     # ODE.u_modified!(integrator, false) # TODO: try this
 end
@@ -205,7 +205,7 @@ function save_to_disk_func(integrator)
     # vorticity
     curl_uh = @. curlₕ(Y.c.uₕ)
     ᶜvort = Geometry.WVector.(curl_uh)
-    Spaces.weighted_dss!(ᶜvort)
+    Spaces.weighted_dss2!(ᶜvort)
 
     dry_diagnostic = (;
         pressure = ᶜp,
--- a/examples/hybrid/staggered_nonhydrostatic_model.jl
+++ b/examples/hybrid/staggered_nonhydrostatic_model.jl
@@ -92,15 +92,15 @@ function default_cache(Y, params, model_spec, spaces, numerics, simulation)
     T_sfc = @. 29 * exp(-lat_sfc^2 / (2 * 26^2)) + 271
     ts_type = CA.thermo_state_type(Y.c, FT)
     ghost_buffer = (
-        c = Spaces.create_ghost_buffer(Y.c),
-        f = Spaces.create_ghost_buffer(Y.f),
-        χ = Spaces.create_ghost_buffer(Y.c.ρ), # for hyperdiffusion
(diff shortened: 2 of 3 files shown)


charleskawczynski (commented):
I'm not sure where the segfault is coming from, but it looks like the new weighted dss doesn't work with single columns. For reference, the single column configuration uses:

```julia
        FT = eltype(params)
        Δx = FT(1) # Note: This value shouldn't matter, since we only have 1 column.
        quad = Spaces.Quadratures.GL{1}()
        horizontal_mesh = periodic_rectangle_mesh(;
            x_max = Δx,
            y_max = Δx,
            x_elem = 1,
            y_elem = 1,
        )
        h_space = make_horizontal_space(horizontal_mesh, quad, comms_ctx)
        z_stretch = if parsed_args["z_stretch"]
            Meshes.GeneralizedExponentialStretching(dz_bottom, dz_top)
        else
            Meshes.Uniform()
        end
        make_hybrid_spaces(h_space, z_max, z_elem, z_stretch)
```
for constructing the `center_space, face_space`. We should probably add a test in ClimaCore to make sure that the new dss operations are compatible with these spaces

sriharshakandala:
> I'm not sure where the segfault is coming from, but it looks like the new weighted dss doesn't work with single columns. For reference, the single column configuration uses:
> 
> ```julia
>         FT = eltype(params)
>         Δx = FT(1) # Note: This value shouldn't matter, since we only have 1 column.
>         quad = Spaces.Quadratures.GL{1}()
>         horizontal_mesh = periodic_rectangle_mesh(;
>             x_max = Δx,
>             y_max = Δx,
>             x_elem = 1,
>             y_elem = 1,
>         )
>         h_space = make_horizontal_space(horizontal_mesh, quad, comms_ctx)
>         z_stretch = if parsed_args["z_stretch"]
>             Meshes.GeneralizedExponentialStretching(dz_bottom, dz_top)
>         else
>             Meshes.Uniform()
>         end
>         make_hybrid_spaces(h_space, z_max, z_elem, z_stretch)
> ```
> 
> for constructing the `center_space, face_space`. We should probably add a test in ClimaCore to make sure that the new dss operations are compatible with these spaces

The column model seems to be only using the Finite Difference discretization. DSS is therefore not relevant case. We need to revise the column model not to call any DSS functions.

charleskawczynski:
> The column model seems to be only using the Finite Difference discretization. DSS is therefore not relevant to this case. We need to revise the column model to not call any DSS functions.

The column model does not call any DSS functions, it's just that we always call DSS functions regardless of the type of the input spaces.

We can either
 - wrap our calls to DSS in a `if axes(Y.c) isa ColumnField` block (or something like that), or
 - Make the DSS perform a no-op on column fields / when there's only one

Can't we make the DSS a no-op when there's a single column? That seems more elegant

simonbyrne:
I think the best solution would be:
1. trying to DSS a GL field should throw a concrete error (not a segfault)
2. in ClimaAtmos, disable the DSS call if using a single column setup

simonbyrne:
I'd prefer it not to be a no-op, since one could still technically define a DSS op in a GL{N} field, but it would be much more complicated (which is why we use a GLL field)

charleskawczynski:
> I'd prefer it not to be a no-op, since one could still technically define a DSS op in a GL{N} field, but it would be much more complicated (which is why we use a GLL field)

Do we support a method for querying if a field is a single column or not? Is this the correct condition?

```julia
function is_single_column(space)
    hspace = Spaces.horizontal_space(space)
    Nh = Topologies.nlocalelems(hspace)
    return Nh == Fields.ncolumns(space)
end
```
?

simonbyrne:
check if the quadrature style is `GL{1}()`?

charleskawczynski:
> check if the quadrature style is `GL{1}()`?

Sounds good, so

```julia
is_single_helem(space) =
    Spaces.horizontal_space(space).quadrature_style isa Spaces.GL{1}
```
? cc @sriharshakandala

sriharshakandala:
> > check if the quadrature style is `GL{1}()`?
> 
> Sounds good, so
> 
> ```julia
> is_single_helem(space) =
>     Spaces.horizontal_space(space).quadrature_style isa Spaces.GL{1}
> ```
> 
> ? cc @sriharshakandala

So, a quick couple of comments. We need to incorporate the following:

1. DSS is not defined on GL grid points (I will add a PR in ClimaCore to throw an error)
2. The single column model can still need DSS (i.e. DSS1d) if spectral element discretization is used in the vertical direction, instead of finite difference discretization, so we need to be a bit more specific!
3. Can we do without definition of horizontal spaces for single column models, which is just a vertical discretization, and just define a vertical space?
4. Do we intend to support a "single column" model that contains one spectral element, i.e. (N+1)^2 columns in the future?

charleskawczynski:
> trying to DSS a GL field should throw a concrete error (not a segfault)

Can't we specialize DSS to be a no-op for `GL{1}`? cc @simonbyrne 

> The single column model can still need DSS (i.e. DSS1d) if spectral element discretization is used in the vertical direction, instead of finite difference discretization, so we need to be a bit more specific!

If we support spectral element in the vertical, then it'd only be for spaces for which we do not support `bycolumn` (i.e., very idealized examples), and since much of the ClimaAtmos infrastructure uses `bycolumn`, I'm not sure that this is something we should be concerned with.

> Can we do without definition of horizontal spaces for single column models, which is just a vertical discretization, and just define a vertical space?

We can try that, but I suspect it won't work. For example, can we call `bycolumn` on a vertical space? One requirement is that we are able to run the same physics kernel in a column and on a sphere.

sriharshakandala:
This is replaced by PR https://github.com/CliMA/ClimaAtmos.jl/pull/1215

PR #1011, opened 2022-10-23
non-breaking clipping and filtering

haakon-e:


trontrytel:

PR #1018, opened 2022-10-24
Merge subsidence into dycore tendencies

charleskawczynski:
This PR:
 - moves subsidence computations into the dycore tendencies
 - Removes TC-internals from subsidence calculations
 - Moves the subsidence cache into the dycore cache (which is now optional)
 - Removes `total_enthalpy` and uses the Thermodynamics `total_specific_enthalpy`

Subsidence is not always a function of time, however this PR changes things so that it's always computed. This is a bit simpler in that there's no need for an initialization _and_ the update. Also, the cache is now tied to the tendency (which is now optional) and in closer proximity to the tendency.

This PR is a step towards #949.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -278,25 +278,25 @@ steps:
     steps:
 
       - label: ":balloon: Single column EDMF Bomex"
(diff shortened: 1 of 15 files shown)


trontrytel (commented):
This is great!

PR #1020, opened 2022-10-24
Increase mem request for longrun

charleskawczynski:
This PR increases the memory request for a longrun job which seems to have failed in [this build](https://buildkite.com/clima/climaatmos-longruns/builds/565#01840b64-a0cc-435e-970c-f225f8f14025/141-188)


szy21:
Oh, this experiment failed because zalesak wasn't working I think. I don't know whether we need to increase the memory, but we can try.

charleskawczynski:
> Oh, this experiment failed because zalesak wasn't working I think. I don't know whether we need to increase the memory, but we can try.

I think the 137 error code indicates OOM, let’s give it a try and see what happens

szy21 (approved):

PR #1022, opened 2022-10-25
Move large-scale advection forcing to src

charleskawczynski:
This PR:
 - Moves large-scale advection to the dycore
 - Removes some unused cached variables

This PR is a step towards #949.


charleskawczynski:
Results in [this build](https://buildkite.com/clima/climaatmos-ci/builds/4824) were not significantly changed. Changes very well may be due to recomputing the LS advection at every timestep rather than once.

trontrytel (commented):
Thank you!

PR #1025, opened 2022-10-25
Refactor precipitation

charleskawczynski:
This PR refactors the precipitation:
 - Removes unused variables
 - Makes some precipitation fraction model functions not depend on the edmf model. The result is that `EDMF_Precipitation.jl` no longer depends on the edmf struct


trontrytel (approved):
LGTM!

charleskawczynski:
Not sure what this failure was about, maybe some race condition?

trontrytel:
This is the error: ERROR: LoadError: No key f_turbconv_up_1_w for mse computation.

I think this might have been caused by the covariant3vector PR. Not sure how yet

PR #1031, opened 2022-10-26
Remove module loading from global scope

charleskawczynski:
Via slack:

> [re-running an example is] pretty slow going on account of the >10 minute compilation time every time I want to run a new integrator

So, we can try to alleviate this by, for now, removing the module loading from global scope so that fewer functions get recompiled on re-including.

cc @dennisYatunin


trontrytel:
Looks like it doesn't work. Should we close it @charleskawczynski ?

charleskawczynski:
Yeah, I think this won't actually fix the issue, and we've mave some incremental progress on improving compilation time

PR #1032, opened 2022-10-26
Use newer limiter interface

charleskawczynski:
The long-run TRMM [build](https://buildkite.com/clima/climaatmos-longruns/builds/567#0184108a-60be-4b22-8e62-0d5fe747e92c) is flooded with limiter warnings: `Warning: Limiter failed to converge with rtol`

@valeriabarra, does updating to the new interface (without prescribing `rtol`) change how the limiter works?

### DIFF
--- a/examples/hybrid/staggered_nonhydrostatic_model.jl
+++ b/examples/hybrid/staggered_nonhydrostatic_model.jl
@@ -114,7 +114,7 @@ function default_cache(Y, params, model_spec, spaces, numerics, simulation)
         tracers = filter(CA.is_tracer_var, propertynames(Y.c))
         make_limiter =
             ᶜρc_name ->
-                Limiters.QuasiMonotoneLimiter(getproperty(Y.c, ᶜρc_name), Y.c.ρ)
+                Limiters.QuasiMonotoneLimiter(getproperty(Y.c, ᶜρc_name))
         limiters = NamedTuple{tracers}(map(make_limiter, tracers))
     else
         limiters = nothing


valeriabarra:
> The long-run TRMM [build](https://buildkite.com/clima/climaatmos-longruns/builds/567#0184108a-60be-4b22-8e62-0d5fe747e92c) is flooded with limiter warnings: `Warning: Limiter failed to converge with rtol`
> 
> @valeriabarra, does updating to the new interface (without prescribing `rtol`) change how the limiter works?

A stricter tolerance assures mass conservation. A larger tolerance may lead to mass loss

valeriabarra:
The current default now is `rtol = eps(FT)`, if you don't explicitly prescribe it

charleskawczynski:
Hi @valeriabarra, thanks for explaining how the tolerance work, but does updating to the new interface (without prescribing rtol) change how the limiter works?

examples/hybrid/staggered_nonhydrostatic_model.jl
```diff
@@ -113,8 +113,10 @@ function default_cache(Y, params, model_spec, spaces, numerics, simulation)
     if apply_limiter
         tracers = filter(CA.is_tracer_var, propertynames(Y.c))
         make_limiter =
-            ᶜρc_name ->
-                Limiters.QuasiMonotoneLimiter(getproperty(Y.c, ᶜρc_name), Y.c.ρ)
+            ᶜρc_name -> Limiters.QuasiMonotoneLimiter(
+                getproperty(Y.c, ᶜρc_name);
+                rtol = FT(10eps(FT)),
```
valeriabarra, line 118: I am not sure I follow your question when you say "updating to the new interface". Here I see that you are prescribing the tolerance in these changes. 
Which new interface are you referring to? Last time the limiter interface was changed (to give back to the user the possibility to specify a user-defined tolerance, as it was in the original implementation) was done in this PR by Dennis in august https://github.com/CliMA/ClimaCore.jl/pull/919 
Is this the "new interface" you are referring to? 
That PR only gave the user back the possibility of specifying a tolerance. But did not change the internals of how the limiter algo works

charleskawczynski: Yes, it looks like the method where density is provided is now deprecated.

> But did not change the internals of how the limiter algo works

Sounds good

@szy21, @trontrytel, being that this is a relative tolerance, does `10eps(FT)` seem reasonable? I'm actually kind of surprised that we're not using `1e-3` or something, `eps` seems very small

valeriabarra: Keep in mind that when Dennis found out that the changes in [CC # 719](https://github.com/CliMA/ClimaCore.jl/pull/719) had disrupted the previous mass-conservation mechanism he found out by only using a tolerance of `sqrt(eps(FT))`. So even with that small of a tolerance it was already large enough to loose a lot of mass (see this Slack [message](https://climate-machine.slack.com/archives/C03RXGTU2H3/p1660934574419029) for reference).

charleskawczynski: Ah, good catch, I guess we'll need to fix the log by figuring out what's going wrong. Maybe we can somehow add an option to write to the log once? I'll update this first by reverting the tolerance change

charleskawczynski:
I've rolled back the tolerance, and now we'll only update to the non-deprecated `QuasiMonotoneLimiter`.

PR #1037, opened 2022-10-27
Try to fix edmf inconsistency job

charleskawczynski:
This PR breaks up `compute_ref_state` into `compute_ref_pressure!` and `compute_ref_density` so that we can call `compute_ref_pressure` in `sgs_flux_tendency!` when anelastic and `test_consistency`.


charleskawczynski:
This doesn't fix the problem, yet, but it's a clear correction.

PR #1038, opened 2022-10-27
Add `debug_state` function

charleskawczynski:
This PR adds a `debug_state` function into ClimaAtmos src/ code, which can be called with, for example: `debug_state(t; Y)`.

I've tested this out locally, we should probably add a test for this, but no calls to it in src/ should ever really be merged

### DIFF
--- a/src/utilities.jl
+++ b/src/utilities.jl
@@ -15,3 +15,137 @@ is_tracer_var(symbol) = !(
 # we may be hitting a slow path:
 # https://stackoverflow.com/questions/14687665/very-slow-stdpow-for-bases-very-close-to-1
 fast_pow(x::FT, y::FT) where {FT <: AbstractFloat} = exp(y * log(x))
+
+
+#####
+##### State debugging tools
+#####
+
+import ClimaCore.Fields as Fields
+
+function debug_state_generic!(state, var::Union{Fields.FieldVector, NamedTuple})
+    for pn in propertynames(var)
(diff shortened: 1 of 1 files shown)


szy21 (approved):

trontrytel:
I think this is useful. Could we actually merge it, but also add a use example somewhere in the docs?

charleskawczynski:
> I think this is useful. Could we actually merge it, but also add a use example somewhere in the docs?

I think I did add an example at first but then deleted because the use is usually just, e.g., `CA.debug_state(t; Y, p)`

charleskawczynski:
I'll add it in

trontrytel:
> > I think this is useful. Could we actually merge it, but also add a use example somewhere in the docs?
> 
> I think I did add an example at first but then deleted because the use is usually just, e.g., `CA.debug_state(t; Y, p)`

Right. Would be nice to have some "debugging hints" page in the docs. But since we don't have it, then maybe just a test that uses it would be enough?

jiahe23 (approved):

charleskawczynski:
Ok, also, I realized that there's a bug 😅, fixing/improving!

charleskawczynski:
I pushed some fixes, I think it's working to some degree now

trontrytel:
> Maybe we can open an issue about adding unit test infrastructure?

Sounds good. I don't want to delay merging it in

Issue #1017, opened 2022-10-24
EDMF is failing the consistency test


trontrytel:
This means that our `update_aux` is done inconsistently, which means we are using some cached values before updating them.

This needs to be fixed. It could potentially be affecting implicit edmf testing.

PR #1043, opened 2022-10-27
Fix 0M/1M microphysics inconsistency

haakon-e:
Finally fixes NaN-inconsistencies that has plagued us for a while.

Since we don't run 1-moment microphysics, the "abstract" version of `compute_precipitation_sink_tendencies` is used to compute precipitation sink tendencies. The 1M version computed tendencies for precipitation sinks of `qt` and `e_tot`, which is used in `compute_gm_tendencies!` to compute grid-mean tendencies for `ρq_tot` and `ρe_tot`. 

In the non-1M-case, these precipitation sink tendencies are not set at all.

This PR sets these tendencies to zero.

Ref: #1039, #1037.

Closes: #1017

### DIFF
--- a/src/TurbulenceConvection/EDMF_Precipitation.jl
+++ b/src/TurbulenceConvection/EDMF_Precipitation.jl
@@ -74,14 +74,21 @@ end
 Computes the tendencies to θ_liq_ice, q_tot, q_rain and q_snow
 due to rain evaporation, snow deposition and sublimation and snow melt
 """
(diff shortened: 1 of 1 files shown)


charleskawczynski (approved):
I'm hoping that we can avoid these computations in the long run by making precipitation more modular, but this is a fine solution for now

charleskawczynski:
cc @dennisYatunin

PR #1046, opened 2022-10-28
Remove unused noneq

trontrytel:
Tmp deleting the non-eq code to speed up the integration with ClimaAtmos and debugging

### DIFF
--- a/src/TurbulenceConvection/EDMF_Environment.jl
+++ b/src/TurbulenceConvection/EDMF_Environment.jl
@@ -24,17 +24,6 @@ function microphysics(
     @inbounds for k in real_center_indices(grid)
         # condensation
         ts = ts_env[k]
-        if edmf.moisture_model isa NonEquilMoistModel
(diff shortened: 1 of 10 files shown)


charleskawczynski:
Is this different from the precipitation we want to modularity? If so, LGTM, if not, then why don’t we leave it in so that we can incrementally move it to the correct place?

trontrytel:
Different from precipitation

> Is this different from the precipitation we want to modularity? If so, LGTM, if not, then why don’t we leave it in so that we can incrementally move it to the correct place?

Different than precipitation

haakon-e:
does this stuff also depend on any external packages that can be removed as well?

trontrytel:
> does this stuff also depend on any external packages that can be removed as well?

Nope, no additional packages.

charleskawczynski (approved):

haakon-e (approved):

PR #1055, opened 2022-10-29
Add parallel remap script

charleskawczynski:
This is an alternative to #1050. I played around in the repl a bit and I think this should work, but it would be good to get confirmation

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -153,6 +153,14 @@ steps:
         command: "julia --color=yes --project=examples examples/hybrid/sphere/remap_pipeline.jl --data_dir prep_remap/ --out_dir remap_pipeline_output"
         artifact_paths: "remap_pipeline_output/*"
 
+      - label: ":computer: Exercise parallel remap pipeline"
+        key: "parallel_remap_pipeline"
+        depends_on: "mpi_baro_wave_make_restart"
+        command: "julia -p 4 --color=yes --project=examples examples/hybrid/sphere/remap_pipeline_parallel.jl --data_dir prep_remap/ --out_dir parallel_remap_pipeline_output"
+        artifact_paths: "parallel_remap_pipeline_output/*"
+        agents:
+          slurm_ntasks: 5
+
   - group: "Sphere Examples"
     steps:
 
--- a/examples/hybrid/sphere/remap_pipeline.jl
+++ b/examples/hybrid/sphere/remap_pipeline.jl
@@ -1,4 +1,28 @@
 #=
+Sbatch script example:
+# Serial remapping
+```
+#!/bin/bash
+#SBATCH --mem=32G
+#SBATCH --cpus-per-task=8
+#SBATCH --job-name=remap
+#SBATCH --time=01:00:00
+#SBATCH --output=remap.log
+module purge
+module load julia/1.8.2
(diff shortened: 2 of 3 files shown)


charleskawczynski:
> I got the following error message when running it on the cluster:

Ah, whoops! I forgot to create weights on the main processor only. I've updated the script, hoping this works!

charleskawczynski:
Ah, darn, I just re-tried my MWE and it doesn't seem to be working, let me re-read the Distributed docs

charleskawczynski:
The MWE I'm testing this on is:

```julia
#= Start with, for example `julia -p 4` =#
using Distributed
@everywhere begin
    foo() = 1
    datas = 1:10
end
bar₀ = 2
@everywhere begin
    bar = @fetch bar₀
end
pmap(datas) do data
    @show data
end
```

charleskawczynski:
cc @szy21

szy21:
I got
```
ERROR: LoadError: On worker 2:
LoadError: cannot declare ca_dir constant; it already has a value
Stacktrace:
  [1] top-level scope
    @ ~/ClimaAtmos.jl/examples/hybrid/sphere/remap_pipeline.jl:3
  [2] include
    @ ./client.jl:476
  [3] top-level scope
    @ ~/ClimaAtmos.jl/examples/hybrid/sphere/remap_pipeline_parallel.jl:12
  [4] eval
    @ ./boot.jl:368
  [5] #invokelatest#2
    @ ./essentials.jl:729
  [6] invokelatest
    @ ./essentials.jl:726
  [7] #114
    @ /central/software/julia/1.8.2/share/julia/stdlib/v1.8/Distributed/src/process_messages.jl:301
  [8] run_work_thunk
    @ /central/software/julia/1.8.2/share/julia/stdlib/v1.8/Distributed/src/process_messages.jl:70
  [9] run_work_thunk
    @ /central/software/julia/1.8.2/share/julia/stdlib/v1.8/Distributed/src/process_messages.jl:79
 [10] #100
    @ ./task.jl:484
in expression starting at /home/zhaoyi/ClimaAtmos.jl/examples/hybrid/sphere/remap_pipeline.jl:3

...and 8 more exceptions.
```
Maybe just remove `const` in remap_pipeline.jl:3?

charleskawczynski:
Ah, interesting, that might work, if not we can try just inlining the `ca_dir`

charleskawczynski:
> BTW, I usually use the following script to run the remap with single processor.
> 
> Could you also provide a similar script but for mpiruns? I am mostly concerned about the correct way to setup the ENV `HDF5_DIR` and `THERMO_VAR` so that all processors can access them with no problem.
> 
> ...

Ah, yes, seeing this is definitely helpful. I'm thinking we can:
 - Eliminate `THERMO_VAR` env entirely and grab that information from `Y`
 - Use the similar `weightfile = fetch(@spawnat :any weightfile_0)` for computed environment variables so that all processors have this information
 - Add this example script as a comment in the pipeline (and make any additional adjustments needed for MPI). I think we may just need to see how things are done in our CI pipeline to find needed adjustments

I'll include the first two in this PR since they're pretty simple changes. @szy21, how have you been trying out the parallel remap? Interactively?

charleskawczynski:
Alright, this is working, merging! Perhaps next we can add a comment in the script or in the docs showing an example of how to run this? cc @jiahe23, @szy21

PR #1067, opened 2022-11-01
Use log2 y-axis for SYPD scaling plot

sriharshakandala:
## Purpose 
Use log10 y axis for SYPD scaling plot

----

### DIFF
--- a/examples/hybrid/plot_scaling_results.jl
+++ b/examples/hybrid/plot_scaling_results.jl
@@ -77,7 +77,7 @@ ENV["GKSwstype"] = "100"
 Plots.GRBackend()
 plt1 = plot(
     log2.(nprocs_clima_atmos),
-    sypd_clima_atmos,
+    log2.(sypd_clima_atmos),
     markershape = :circle,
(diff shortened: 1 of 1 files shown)


examples/hybrid/plot_scaling_results.jl
```diff
@@ -77,22 +77,31 @@ ENV["GKSwstype"] = "100"
 Plots.GRBackend()
 plt1 = plot(
     log2.(nprocs_clima_atmos),
-    sypd_clima_atmos,
+    log.(sypd_clima_atmos),
```
valeriabarra, line 80: Why we don't want to use the same bases?
I wouldn't mix different logarithmic bases on the different axes. It is common to either use a semilog (or lin-log) plot as you had before or a log-log plot, but not a log-log2 plot where one axis has base 2 and the other one natural base.

charleskawczynski (approved):
Code changes seem fine to me, maybe you can post here or on slack what the new figures look like? (if it's not too much trouble)

PR #1073, opened 2022-11-02
Move implicit tendencies to src

charleskawczynski:
This PR
 - Moves the implicit tendencies into src/
 - Moves implicit tendency verification tools into test/
 - Removes some closures over global constants


charleskawczynski:
Since these changes were a bit complicated, I'm going to not squash to preserve the history (sometimes git's diffs are poor when file changes are combined with file renames).

PR #1080, opened 2022-11-03
Add JFNK pipeline example and other minor changes

dennisYatunin:
## Purpose 
In order to identify the bug causing non-determinism in the IMEX EDMF experiments, I am splitting that PR into several smaller PRs. This first PR is a collection of minor changes that (hopefully) have no effect on the behavior of the simulation.

## Content
- Change the `compressible_edmf_bomex` pipeline example `dt` from 1s to 40s
- Add an identical example that uses JFNK instead of Newton's method (this can actually run with a 45s `dt`, but the resulting profiles are extremely noisy)
- Change `ds_filename_reference` from `joinpath(path, ds_filename_computed)` to `joinpath(path, job_id, "prog_state.nc")` in `compute_mse.jl`, so that both examples can use the same reference data
- Modify `define_tc_quicklook_profiles.jl`, `self_reference_or_path.jl`, and `driver.jl` so that the quicklook plot can be generated locally, without looking up the reference data from `main` (this is very helpful for debugging)
- Rename `CA.implicit_tendency!` to `CA.implicit_vertical_advection_tendency!`, and move the zeroing out of the implicit EDMF tendency from `src` to `examples` (this is so that, when `TCU.sgs_flux_tendency!` is split into `TCU.explicit_sgs_flux_tendency!` and `TCU.implicit_sgs_flux_tendency!`, `TCU.implicit_sgs_flux_tendency!` can be called from `examples`, right after the implicit EDMF tendency is zeroed out)


post_processing/define_tc_quicklook_profiles.jl
```diff
@@ -89,8 +85,12 @@ function plot_tc_profiles(folder; hdf5_filename, main_branch_data_path)
         plot!(p16, parent(D.env_HQTcov)[:], zc; label = "$data_source")
     end
 
+    PR_filename = joinpath(folder, hdf5_filename)
     add_to_plots!(PR_filename; data_source = "PR")
-    add_to_plots!(main_filename; data_source = "main")
+    if ispath(main_branch_data_path)
+        main_filename = joinpath(main_branch_data_path, hdf5_filename)
+        add_to_plots!(main_filename; data_source = "main")
+    end
```
charleskawczynski, line 93:
```suggestion
    else
        @warn "No path found for comparison against main"
    end
```

examples/hybrid/driver.jl
```diff
@@ -364,7 +364,6 @@ if parsed_args["debugging_tc"]
     )
 
     main_branch_root = get_main_branch_buildkite_path()
-    @info "Comparing PR profiles against main with commit id: $(basename(main_branch_root))"
```
charleskawczynski, line 367: Was this deletion intentional? I didn't write this, but these can be useful breadcrumbs when looking back through logs

regression_tests/compute_mse.jl
```diff
@@ -29,7 +29,7 @@ function regression_test(; job_id, reference_mse, ds_filename_computed, varname)
     if haskey(ENV, "BUILDKITE_COMMIT")
         path = self_reference_or_path()
         path == :self_reference && return reference_mse
-        ds_filename_reference = joinpath(path, ds_filename_computed)
+        ds_filename_reference = joinpath(path, job_id, "prog_state.nc")
```
charleskawczynski, line 32: Is this necessary? How is this change used?

examples/hybrid/driver.jl
dennisYatunin: This isn't a deletion; it's a [move](https://github.com/CliMA/ClimaAtmos.jl/blob/fe3b9d47c74b15304723c00a32561c534ca66f0d/regression_tests/self_reference_or_path.jl#L120-L121).

regression_tests/compute_mse.jl
dennisYatunin: This is necessary in order to run multiple pipeline examples that compare against the same reference data, like [here](https://github.com/CliMA/ClimaAtmos.jl/blob/fe3b9d47c74b15304723c00a32561c534ca66f0d/.buildkite/pipeline.yml#L313-L323).

post_processing/define_tc_quicklook_profiles.jl
dennisYatunin: I added the warning [here](https://github.com/CliMA/ClimaAtmos.jl/blob/ba696500f562eb49ff60f61719c10cec4b6ad5b7/regression_tests/self_reference_or_path.jl#L89), in the same function where I moved the comparison `@info`.

regression_tests/self_reference_or_path.jl
```diff
@@ -85,6 +85,10 @@ This is very similar to `self_reference_or_path`, except that it always returns
 function get_main_branch_buildkite_path()
     # TODO: Combine this method with `self_reference_or_path`, above.
     cluster_data_prefix = "/central/scratch/esm/slurm-buildkite/climaatmos-main"
+    if !ispath(cluster_data_prefix)
```
charleskawczynski, line 88:
```suggestion
    on_buildkite = haskey(ENV, "BUILDKITE_COMMIT") || haskey(ENV, "BUILDKITE_BRANCH")
    if !on_buildkite && !ispath(cluster_data_prefix)
```

Let's ensure that this errors somewhere rather than falling back on a local-run mode and silently passing if we're on buildkite, I think this should do it

regression_tests/compute_mse.jl
charleskawczynski: Ok, I think this is fine to try out

charleskawczynski (approved):
We'll need to increment the ref counter and add an MSE entry for the new job that's opting into the regression tests, but LGTM aside from that

examples/hybrid/staggered_nonhydrostatic_model.jl
```diff
@@ -169,7 +169,12 @@ end
 function implicit_tendency!(Yₜ, Y, p, t)
     @nvtx "implicit tendency" color = colorant"yellow" begin
         Fields.bycolumn(axes(Y.c)) do colidx
-            CA.implicit_tendency!(Yₜ, Y, p, t, colidx)
+            CA.implicit_vertical_advection_tendency!(Yₜ, Y, p, t, colidx)
+
```
charleskawczynski, line 173:
```suggestion
            FT = Spaces.undertype(axes(Y.c))
```

dennisYatunin:
> We'll need to increment the ref counter and add an MSE entry for the new job that's opting into the regression tests, but LGTM aside from that

I don't think we need to add a new MSE entry, since the new job uses the same reference data as one of the old jobs. I think the entry in the MSE table will just be the maximum of the MSEs of the two jobs.

regression_tests/ref_counter.jl
```diff
@@ -1 +1 @@
-31
+32
```
trontrytel, line 1: If we are bumping the ref counter we don't need to also change the mase tables?

examples/hybrid/staggered_nonhydrostatic_model.jl
```diff
@@ -169,7 +169,12 @@ end
...
+            if p.tendency_knobs.has_turbconv
+                parent(Yₜ.c.turbconv[colidx]) .= zero(eltype(Yₜ))
+                parent(Yₜ.f.turbconv[colidx]) .= zero(eltype(Yₜ))
```
trontrytel, line 176: Where are the non-turbconv tendencies zeroed out? Shouldn't we add this bit of code there?

dennisYatunin: They aren't zeroed out anywhere. They are just set in `implicit_vertical_advection_tendency`. At some point, we should probably go through the tendencies and make them all "incrementing" rather than "setting", since that will allow us to change the order of tendency evaluations or to temporarily disable certain tendencies.

regression_tests/ref_counter.jl
charleskawczynski: There's a new job which opts into the regression tests, so we do need to increment the ref counter

dennisYatunin:
@charleskawczynski Can we drop the "Ensure mse tables are reset when necessary" test? This would let us have multiple jobs use the same reference data, where one of the jobs has a nonzero MSE compared to the reference.

dennisYatunin:
Per @charleskawczynski's suggestion, I have added a new command-line argument called `quicklook_reference_job_id` (whose default value is `job_id`), which is used in place of `job_id` in the "quicklook" plots. The regression tests will continue to be specific to each job, instead of being shared across jobs.

PR #1086, opened 2022-11-04
Add timings to init functions

charleskawczynski:
This PR adds some timings to the init functions, so that we can hopefully track down the long init times.


charleskawczynski:
Let's merge this so that we can track timings over time. Hoping this will help us improve init time a bit.

PR #1090, opened 2022-11-07
Do not cache log_pressure_fun

charleskawczynski:
This PR removes `log_pressure_fun` from the cache, and replaces it with two `ColumnField`s. This has two advantages:
 - It seems to reduce compile time for `OrdinaryDiffEq.init()` by ~30%, which is dominating the cost of some of the compressible examples (observed in a previous build, will keep an eye on this one.)
 - It alleviates the need for pressure BCs, since we can just evaluate the pressure solution at the surface
 - Storing the column field is cheaper than storing the entire field, and it's very possibly cheaper than saving the ODE solution.

cc @dennisYatunin

### DIFF
--- a/examples/hybrid/TurbulenceConvectionUtils.jl
+++ b/examples/hybrid/TurbulenceConvectionUtils.jl
@@ -58,12 +58,18 @@ function get_edmf_cache(
     )
     edmf = turbconv_model
     ᶠspace_1 = axes(Y.f[CC.Fields.ColumnIndex((1, 1), 1)])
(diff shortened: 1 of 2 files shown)


charleskawczynski:
@ChrisRackauckas, any idea why this might be a performance (latency) issue? It's a bit difficult to confirm exactly what's going on, but it seems like having an OrdinaryDiffEq solution stored in the cache (`p`) of an `OrdinaryDiffEq.integrator` makes compiling `OrdinaryDiffEq.init` take a very long time.

I'll see if I can make a reproducer, but I thought I'd reach out in case @ChrisRackauckas may know something off hand

PR #1092, opened 2022-11-08
Modify EDMF filtering

dennisYatunin:
## Purpose 
This PR changes how EDMF filtering works by making it so that we filter a copy of the state vector instead of the state vector itself. In order to prevent the state and its filtered version from diverging during each timestep, this PR also adds a relaxation tendency.

## Content
After a great deal of trial and error, it has been determined that, if we split the EDMF tendency into implicit and explicit components, the state vector (`Y`) cannot be filtered from within the implicit component, as that somehow breaks the GMRES algorithm. Currently, the only known solution to this problem is to filter a copy of the state vector instead of the state vector itself, and to use this copy to compute the implicit component of the EDMF tendency. However, only doing this in the implicit component of the tendency will cause filtering to be inconsistent between the implicit and explicit components, which seems like a bad idea. So, before we actually split the EDMF tendency, we should modify it to use the form of filtering that will be required by the implicit component.

However, just using a filtered copy of the state vector is not sufficient, as it causes several of the examples in the pipeline (particularly anelastic BOMEx and compressible TRMM) to develop spurious updrafts near the top of the atmosphere. This is because, if the state itself is not filtered on every Runge-Kutta stage, these spurious updrafts can quickly develop over the course of a single timestep; then, even though we filter the state after every timestep through a callback, the spurious updrafts are sometimes sufficiently large to survive this filtering. The solution that seems to work best is to add a "filter relaxation tendency" of the form `(filter(Y) - Y) / dt`, which makes the results look nearly identical to those on the main branch. Interestingly, scaling this tendency down by a factor of a 100 also seems to work, which indicates that only a very small relaxation tendency is needed to prevent `filter(Y)` and `Y` from diverging too much before the callback at the end of each timestep realigns them. However, using a large relaxation tendency works well enough, so a scaling factor does not need to be included in the tendency.

----


trontrytel:
Basted on what we discussed today:
- We want to merge this PR as soon as possible. So that we can start Jacobian testing.
- We want to refactor the filters: (i) make sure we only have the filters we need, (ii) use regularisation instead of area fraction clipping, (iii) move filtering to be part of tendency computation. We will move that work to a separate issue.

haakon-e (approved):

PR #1097, opened 2022-11-09
Add contour diagnostic plots for single column config

### LINKED ISSUE #654: Split out distributed output from figures
The distributed output (i.e., JLD2 files) make finding any figures on buildkite very difficult. We should split out the output folder to make navigation a bit easier (if possible).

charleskawczynski:
This PR
 - Adds contour diagnostic plots for single column configurations, when using `debugging_tc`.
 - Increases the frequency at which diagnostics are exported when `debugging_tc == true`
 - Zips the `.hdf5` files so that the artifacts path isn't too cluttered

Closes #654.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -310,62 +310,62 @@ steps:
     steps:
 
       - label: ":balloon: Single column EDMF Bomex"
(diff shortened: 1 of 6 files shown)


trontrytel:
Thank you!

PR #1100, opened 2022-11-09
update NVTX version, re-enable hooks

simonbyrne:
## Purpose 
I made some updates to NVTX.jl that makes it easier to use inside a package.

## Content

----


charleskawczynski (approved):

charleskawczynski:
I think we just need to add the NVTX dep to the perf/ and ode_compat_examples/ environments?

PR #1114, opened 2022-11-12
Refactor surface

charleskawczynski:
This PR refactors the surface handling in several ways:
 - Caches the surface fluxes object, so it can later be used by the edmf model
 - Caches the surface fluxes inputs, so that we can more simply dispatch
 - Changes (maybe fixes) `get_surface_density_and_humidity` to be mutating (now `set_surface_density_and_humidity!`) cc @szy21, @LenkaNovak, maybe we can discuss this change?
 - I've kept `surface_conditions` for now, just to minimize the blast radius, but I'd like to remove that since it's basically directly dependent on `SurfaceFluxConditions` and the surface flux input types (depending on which one is used)

Ultimately, this should allow us to delete nearly all of edmf's `get_surface` function, and just accept the already created `SurfaceFluxConditions`.

EDIT:

 - I also replaced the `coupled` Bool with a `coupling` type because I was trying to hunt down allocations in one of the broadcast expressions. I think it turned out to be that the nested thermo state and surface fluxes input expression was a bit too complex for the compiler to infer, and things started working smoothly when that was broken up. I didn't revert the coupling type because I figure that it's not a bad idea to do that anyway, since it will yield more precise generated code than the runtime Bool.

### DIFF
--- a/src/tendencies/vertical_diffusion_boundary_layer.jl
+++ b/src/tendencies/vertical_diffusion_boundary_layer.jl
@@ -2,6 +2,7 @@
 ##### Vertical diffusion boundary layer parameterization
 #####
 
+import StaticArrays
 import ClimaCore.Geometry: ⊗
 import ClimaCore.Utilities: half
 import LinearAlgebra: norm
@@ -22,7 +23,7 @@ function vertical_diffusion_boundary_layer_cache(
     surface_scheme = nothing,
     C_E::FT = FT(0),
     diffuse_momentum = true,
-    coupled = false,
+    coupling,
 ) where {FT}
     z_bottom = Spaces.level(Fields.coordinate_field(Y.c).z, 1)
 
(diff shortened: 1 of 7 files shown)


src/tendencies/vertical_diffusion_boundary_layer.jl
```diff
@@ -42,22 +42,42 @@ function vertical_diffusion_boundary_layer_cache(
...
+        sfc_inputs_type = typeof(SF.Coefficients{FT}(;
+            state_in = SF.InteriorValues(FT(0), (FT(0), FT(0)), ts_type),
+            state_sfc = SF.SurfaceValues(FT(0), (FT(0), FT(0)), ts_type),
+            z0m = FT(0),
+            z0b = FT(0),
+            Cd = FT(0),
+            Ch = FT(0),
+        ))
+        sfc_inputs = similar(Fields.level(Y.f, half), sfc_inputs_type)
+        fill!(sfc_inputs.Cd, FT(0.0044))
```
LenkaNovak, line 60: This is not urgent, but it would be better to be able to specify this value in the command line arguments.

src/tendencies/vertical_diffusion_boundary_layer.jl
```diff
@@ -42,22 +42,42 @@ function vertical_diffusion_boundary_layer_cache(
...
-        )
+
+        sfc_inputs_type = typeof(SF.ValuesOnly{FT}(;
+                state_in = SF.InteriorValues(FT(0), (FT(0), FT(0)), ts_type),
+                state_sfc = SF.SurfaceValues(FT(0), (FT(0), FT(0)), ts_type),
+                z0m = FT(0),
+                z0b = FT(0),
+        ))
+        sfc_inputs = similar(Fields.level(Y.f, half), sfc_inputs_type)
+        fill!(sfc_inputs.z0m, FT(1e-5))
```
LenkaNovak, line 71: same here (better if specifiable externally)

src/tendencies/vertical_diffusion_boundary_layer.jl
```diff
@@ -226,22 +214,34 @@ function get_surface_fluxes!(Y, p, colidx)
...
+        params,
+        coupled,
+    )
+
+    # calculate all fluxes (saturated surface conditions)
+    @. sfc_conditions[colidx] = SF.surface_conditions(sf_params, sfc_inputs[colidx])
+
+    @. E[colidx] = SF.evaporation(sf_params, sfc_inputs[colidx], sfc_conditions[colidx].Ch)
+
+    # TODO: remove this!
```
LenkaNovak, line 236: I presume this is a TODO for later?

LenkaNovak (approved):
Thanks, @charleskawczynski ! Looks good to me -I only had a few minor comments. Still happy to meet if it's helpful though!

trontrytel (approved):
LGTM!

src/tendencies/vertical_diffusion_boundary_layer.jl
charleskawczynski: Yep!

szy21 (approved):
Looks good, thanks!

charleskawczynski:
I had to make a few more changes, going to merge but I'm happy to change names later if people prefer different names. I think the structure is an overall improvement though.

src/tendencies/vertical_diffusion_boundary_layer.jl
```diff
@@ -218,33 +269,49 @@ function get_surface_fluxes!(Y, p, colidx)
...
-            q_sfc[colidx],
-            Spaces.level(ᶜts[colidx], 1),
-            uₕ_int_local[colidx],
-            z_bottom[colidx],
-            z_sfc[colidx],
-            params,
-            coupled,
-        )
+
+    # TODO: Remove use of parent
```
charleskawczynski, line 282: Note to self: ` # TODO: Remove use of parent` can be deleted.

PR #1108, opened 2022-11-10
Move precipitation tendencies to dycore

charleskawczynski:
This PR:

 - Moves the precipitation tendencies to the dycore and
 - Adds a 0-moment precip model job, to exercise this code

PR #1116, opened 2022-11-14
compute precipitation tendencies after edmf

trontrytel:
This is a spinoff from https://github.com/CliMA/ClimaAtmos.jl/pull/1108. Testing to see how much difference we get just by changing the order of cimputations


charleskawczynski (approved):

trontrytel:
@szy21 - can you look at the plots for the breaking case? I don't see any differences. So maybe we can merge it in?

szy21:
Yes, they look good to me as well.

PR #1130, opened 2022-11-17
Add diagonal Jacobian blocks for TC

dennisYatunin:
## Purpose 

This PR adds the ability to set tridiagonal blocks along the diagonal of the implicit tendency's Jacobian, where each block corresponds to $\partial\mathcal{V}_t/\partial\mathcal{V}$, where $\mathcal{V}$ is any variable in `Y.c.turbconv` or `Y.f.turbconv`.

## Content

This PR modifies `SchurComplementW` by adding the new fields `∂ᶜTCₜ∂ᶜTC` and `∂ᶠTCₜ∂ᶠTC`, which store a tridiagonal matrix for each TC variable, and modifying `ldiv!` to use these fields appropriately. By default, these fields are filled with 0s, but they could be modified from `Wfact!` or from a callback to contain better approximations of the corresponding blocks from the exact Jacobian. This could allow us to decrease the number of iterations required by the Krylov solver on each implicit Runge-Kutta stage.

----

### DIFF
--- a/src/tendencies/implicit/schur_complement_W.jl
+++ b/src/tendencies/implicit/schur_complement_W.jl
@@ -9,7 +9,7 @@ import ClimaCore.Fields as Fields
 import ClimaCore.Operators as Operators
 using ClimaCore.Utilities: half
 
-struct SchurComplementW{F, FT, J1, J2, J3, J4, J5, S, A, T}
+struct SchurComplementW{F, FT, J1, J2, J3, J4, J5, J6, J7, S, A, T}
     # whether this struct is used to compute Wfact_t or Wfact
     transform::Bool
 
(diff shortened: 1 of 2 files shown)


src/tendencies/implicit/schur_complement_W.jl
```diff
@@ -302,15 +352,45 @@ function _ldiv_serial!(
...
+            xᶜvar = Fields.single_field(xᶜTC, var_prop_chain, identity)
+            bᶜvar = Fields.single_field(bᶜTC, var_prop_chain, identity)
+            xᶜvar .= bᶜvar
+            xᶜvar_view = parent(xᶜvar)
+            ∂ᶜvarₜ∂ᶜvar =
+                Fields.single_field(∂ᶜTCₜ∂ᶜTC, var_prop_chain, identity)
+            @views ∂ᶜvarₜ∂ᶜvar_array = LinearAlgebra.Tridiagonal(
+                S_column_array.dl[1:(end - 1)],
+                S_column_array.d[1:(end - 1)],
+                S_column_array.du[1:(end - 1)],
```
charleskawczynski, line 370: Is this properly viewed into? Above we're using `@views S_column_array.dl .= parent(S_column.coefs.:1)[2:end]`

charleskawczynski (commented):
Overall I think this is a fine way to do this, I think we just need to double check indexing.

I should probably point out / remind you that `single_field` does transform some covariant fields velocity. There's an optional argument if we want to get the identity back: `single_field(f::Union{Field, FieldVector}, prop_chain, transform = transform_field)`. I'm now starting to wonder if we should make the transformation the identity by default. This was originally added for plotting purposes, and we never really wanted to plot the raw data.

Also, I kind of like the idea of the modular use of `property_chains` here, where we index into a submodel and we can then iterate over all of the subfields of the submodel.

dennisYatunin:
@charleskawczynski Yeah I think I used `identity` as the third argument everywhere I called `single_field`. I assume that means I'm just getting views of the underlying arrays instead of new arrays.

src/tendencies/implicit/schur_complement_W.jl
charleskawczynski: Ah, ok, right, `thomas_algorithm!` is populating the matrix

charleskawczynski (approved):

PR #1134, opened 2022-11-18
Refactor surface functions

charleskawczynski:
This PR:
 - Refactors the surface functions by changing `set_surface_density_and_humidity!` to `set_surface_thermo_state!`, which we can use to dispatch between the GCM case vs a prescribed surface thermo state case (of which there are many for the TC SCM examples)
 - Removes `q_sfc` and `ρ_sfc` from the cache, since we can just store the surface thermo state instead.

This does add allocations to the diagnostics callback, but I've been thinking that it may be nice to split out the cache into solver vs diagnostics cache so that we can somehow ensure that computing diagnostics doesn't impact the solve.

Maybe these changes will slightly impact how we interact with the coupler? cc-ing @LenkaNovak

### DIFF
--- a/examples/hybrid/callbacks.jl
+++ b/examples/hybrid/callbacks.jl
@@ -298,10 +298,10 @@ function save_to_disk_func(integrator)
             col_integrated_rain,
             col_integrated_snow,
             T_sfc,
-            q_sfc,
+            ts_sfc,
         ) = p
     else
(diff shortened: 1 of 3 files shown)


szy21 (approved):
Looks good, thanks! I think the coupler can just modify ts_sfc instead of rho_sfc and q_sfc, but Lenka would know more. Btw she is taking today off and may not see this.

Issue #1079, opened 2022-11-02
Check energy conservation when there are sources or sinks


szy21:
Check that the atmos model conserves energy when there are sources or sinks, such as radiation, surface fluxes, and precipitation.

cc @LenkaNovak

szy21:
superseded by #2568

PR #1136, opened 2022-11-18
add mass and energy conservation check

szy21:
## Purpose 

Partially addresses #1079. Right now it only works when radiation is the only external tendency, as all other external tendencies change within one timestep. @simonbyrne suggests that a better way is to add some state variables to accumulate fluxes. I still would like to merge this in as a temporary solution though, as I think it would be good to make sure we are conserving mass and energy in ci.

## Content

----

### DIFF
--- a/examples/hybrid/callbacks.jl
+++ b/examples/hybrid/callbacks.jl
@@ -20,6 +20,11 @@ function get_callbacks(parsed_args, simulation, atmos, params)
 
     tc_callbacks =
         call_every_n_steps(turb_conv_affect_filter!; skip_first = true)
+    flux_accumulation_callback = call_every_n_steps(
+        flux_accumulation!;
+        skip_first = true,
+        call_at_end = true,
+    )
 
     additional_callbacks =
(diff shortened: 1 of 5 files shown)


charleskawczynski (approved):
Code changes look fine to me

examples/hybrid/callbacks.jl
```diff
@@ -150,6 +160,23 @@ function turb_conv_affect_filter!(integrator)
     ODE.u_modified!(integrator, false)
 end
 
+horizontal_integral_at_boundary(f, lev) = sum(
+    Spaces.level(f, lev) ./ Fields.dz_field(axes(Spaces.level(f, lev))) .* 2,
```
charleskawczynski, line 164: Can you remind me why the `*2` is needed here?

szy21: Because at the boundary faces, `sum` calculates the integral over dz_field/2.

charleskawczynski: Ah, right, thanks!

PR #1138, opened 2022-11-18
Remove zeroing out of momentum fluxes

charleskawczynski:
This PR attempts to remove the zero-ing out of momentum fluxes, as this is inconsistent with how momentum fluxes are determined in SurfaceFluxes.

The only reason why I think this could have been done intentionally is because these fluxes are used as BCs in the momentum SGS fluxes, and maybe this is done to somehow avoid double counting?

Either way it'll be interesting to see what happens if we don't zero them out


szy21:
It seems TRMM is fine but ARM changes a lot

charleskawczynski:
Superseded by #1144.

PR #1146, opened 2022-11-20
Remove SurfaceBase

charleskawczynski:
This PR removes `SurfaceBase`, and instead works with the `SurfaceFluxes.SurfaceFluxConditions` type, which the dycore uses.


charleskawczynski:
bors ping

PR #916, opened 2022-10-08
Add edmf sphere example

charleskawczynski:
This PR adds an edmf sphere example:

 - Fixes `surface_thermo_state` to return the correct thermo state type (dry or moist)
 - Fixes `surface_params` to use the correct thermo state type (dry or moist)

PR #1149, opened 2022-11-22
Add precipitation prognostic variables

charleskawczynski:
This PR adds precipitation prognostic variables

This is a peel off from #916.

### DIFF
--- a/src/InitialConditions/baro_wave.jl
+++ b/src/InitialConditions/baro_wave.jl
@@ -120,6 +120,7 @@ function center_initial_condition_baroclinic_wave(
         ᶜ𝔼_kwarg...,
         uₕ,
         moisture_vars(thermo_params, ts, atmos)...,
+        precipitation_vars(FT, atmos)...,
         turbconv_vars(FT, atmos)...,
     )
 end
--- a/src/InitialConditions/box.jl
+++ b/src/InitialConditions/box.jl
@@ -58,6 +58,7 @@ function center_initial_condition_box(
         ᶜ𝔼_kwarg...,
         uₕ,
         moisture_vars(thermo_params, ts, atmos)...,
+        precipitation_vars(FT, atmos)...,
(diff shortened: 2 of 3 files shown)


src/InitialConditions/baro_wave.jl
```diff
@@ -129,10 +129,30 @@ function center_initial_condition_baroclinic_wave(
     end
     # TODO: Include ability to handle nonzero initial cloud condensate
 
+    precip_kwargs =
+        if precip_model isa NoPrecipitation &&
```
trontrytel, line 133: If we use 0-moment microphysics we also should not add q_rai and q_sno to variable list

src/InitialConditions/baro_wave.jl
```diff
@@ -129,10 +129,30 @@ function center_initial_condition_baroclinic_wave(
...
+           !(turbconv_model isa TC.EDMFModel)
+            NamedTuple()
+        else
+            (; q_rai = FT(0), q_sno = FT(0))
```
trontrytel, line 137: Should our tracers for rain be the same as other tracers in the dycore. I.e. be rho * q_something ?

charleskawczynski: Probably. This is just a peel off from #916, but it's introducing a pretty major performance penalty due to inference failure

trontrytel: :/ Does that mean we don't want to merge it? Do you have any ideas how to help the compiler there?

charleskawczynski: Not as is, it's slowing some jobs down _significantly_. I think the best course of action is to somehow make this optional, and merge it while only adding a single job that exercises it (maybe the perf job?). This way, it doesn't impact everyone, but we can keep an eye on / experiment with it.

charleskawczynski:
Superseded by #1164.

PR #1164, opened 2022-11-28
Add perf mode flag and experimental perf job

charleskawczynski:
This PR is a peel off from #916, and adds a performance flag / type in order to monitor the performance of adding tracers, which seems to hurt performance (found in #1149)


charleskawczynski:
Supersedes #1149.

charleskawczynski:
Only the JFNK jobs are changing, and it's changing by a small amount. This may be due to feeding additional zeros into Krylov.jl

PR #1168, opened 2022-11-29
fix scaling/comparison pipelines

simonbyrne:
## Purpose 
Scaling and comparison pipelines are broken after rearranging files. This simplifies them a bit

## Content

----


simonbyrne:
Running here: https://buildkite.com/clima/climaatmos-scaling/builds/400

PR #1174, opened 2022-12-01
Refactor ode config

charleskawczynski:
This PR refactors the ode config by:
 - Breaking `ode_configuration` down into several functions, which allows us to only return the ode algorithm (so `ode_configuration` is now renamed to `get_ode_algorithm`)
 - Several helper functions were added so that users can more easily test, using the `ode_algo_type`, intermediate parts of `get_ode_algorithm`

I think this should be slightly easier to extend for the new CTS changes, and it may have better inference due to all the function barriers

### DIFF
--- a/examples/hybrid/driver.jl
+++ b/examples/hybrid/driver.jl
@@ -254,15 +254,14 @@ end
 # @info "Model composition" p.atmos...
 @info "Tendencies" p.tendency_knobs...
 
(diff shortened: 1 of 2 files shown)


charleskawczynski:
Superseded by https://github.com/CliMA/ClimaAtmos.jl/pull/1193

PR #1178, opened 2022-12-02
Fix stack traces

charleskawczynski:
This PR should fix the stack traces when simulations crash.


charleskawczynski:
cref: #1176, #1175, #1170.

PR #1179, opened 2022-12-02
Update to new ClimaTimeSteppers

charleskawczynski:
This PR upgrades to use the new ClimaTimeSteppers ([CTS#PR85](https://github.com/CliMA/ClimaTimeSteppers.jl/pull/85)).


examples/hybrid/types.jl
```diff
@@ -364,25 +372,32 @@ function args_integrator(parsed_args, Y, p, tspan, ode_config, callback)
...
+            ODE.SplitODEFunction(
+                implicit_func,
+                remaining_tendency!, # should be remaining_tendency_plus_dss!
+            )
+        else
+            CTS.ClimaODEFunction(;
+                T_exp! = remaining_tendency!,
+                T_imp! = implicit_func,
+                # Can we just pass implicit_tendency! and jac_prototype etc.?
+                dss!,
```
charleskawczynski, line 391: Reminder: comment out the `dss!` in the `remaining_tendency!` and `implicit_func`s

charleskawczynski:
@szy21, it seems that updating to the new time steppers requires slightly increasing the conservation limit, is this okay?

charleskawczynski:
I suspect that this is not a major issue since the moisture cases are very similar. I’ll merge and we can further investigate if need be

szy21:
> @szy21, it seems that updating to the new time steppers requires slightly increasing the conservation limit, is this okay?
> 
> From `@test sum(sol.u[1].c.ρ) ≈ sum(sol.u[end].c.ρ) rtol = 10 * eps(FT)` to `@test sum(sol.u[1].c.ρ) ≈ sum(sol.u[end].c.ρ) rtol = 20 * eps(FT)`

It’s probably fine. Thanks for getting it in!

PR #1187, opened 2022-12-07
add compressible Soares test case

trontrytel:


trontrytel:
Will be done here: https://github.com/CliMA/ClimaAtmos.jl/pull/1258

PR #1194, opened 2022-12-09
Fix thermo_state for pressure interpolation in radiation

szy21:
## Purpose 

Fix the bug of setting kinetic energy to zero when computing pressure for interpolation in radiation.

## Content

----


charleskawczynski:
Superseded by #1195

PR #1209, opened 2022-12-16
Increase allocation limit for longrun zalesak job

charleskawczynski:
This PR increases the allocation limit for the longrun zalesak job, which seems to be needed based on the 137 error in the [last build](https://buildkite.com/clima/climaatmos-longruns/builds/692#018519c2-d89b-4bd2-9837-7d2b4ea44217)


szy21:
I think this run is just unstable. I wouldn't worry too much about it.

PR #1215, opened 2022-12-20
Update DSS to DSS2

sriharshakandala:
## Purpose 
Move to new DSS implementation with optimized halo exchange.
This will replace PR https://github.com/CliMA/ClimaAtmos.jl/pull/1002

----


sriharshakandala:
bors try

simonbyrne:
maybe the simpler option is to check if the quadrature rule == `Spaces.Quadratures.GL{1}()`

sriharshakandala:
> maybe the simpler option is to check if the quadrature rule == `Spaces.Quadratures.GL{1}()`

Yes. I tried this one and it works. I am moving the `skip_dss` entry into the `ghost_buffer` so that it only need to be checked once!

charleskawczynski (approved):
Code changes look fine to me

PR #1224, opened 2022-12-22
Add beta to sfc_inputs cache

### LINKED ISSUE #1214: Add `beta` factor to AtmosSimulation cache
SurfaceFluxes and ClimaLSM both now include a beta factor to more realistically represent evaporation (see https://github.com/CliMA/SurfaceFluxes.jl/pull/75 and https://github.com/CliMA/ClimaLSM.jl/pull/124). We want to update the coupler to perform the calculation of the beta factor and store it in the atmos cache (https://github.com/CliMA/ClimaCoupler.jl/issues/199).

To do this, we need an additional field in the atmos cache to store the beta factor (i.e. `atmos_sim.integrator.p.beta`).

szy21:
## Purpose 

Closes #1214 

## Content

----

### DIFF
--- a/src/tendencies/vertical_diffusion_boundary_layer.jl
+++ b/src/tendencies/vertical_diffusion_boundary_layer.jl
@@ -65,11 +65,13 @@ function vertical_diffusion_boundary_layer_cache(
                 z0b = FT(0),
                 Cd = FT(0),
                 Ch = FT(0),
+                beta = FT(0),
             ),
         )
         sfc_inputs = similar(Fields.level(Y.f, half), sfc_inputs_type)
         fill!(sfc_inputs.Cd, FT(0.0044)) #FT(0.001)
         fill!(sfc_inputs.Ch, FT(0.0044)) #FT(0.0001)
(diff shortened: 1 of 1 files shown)


kmdeck (approved):
LGTM!

I noticed that this file only uses ValuesOnly and Coefficients, from SurfaceFluxes - both need beta. But there were other types in SF that did not use beta. 

maybe there could be some confusion in the future if CA uses SF.Fluxes or SF.FluxesandFrictionVelocity, but beta is always in the cache. but it's probably ok!

szy21:
> LGTM!
> 
> I noticed that this file only uses ValuesOnly and Coefficients, from SurfaceFluxes - both need beta. But there were other types in SF that did not use beta.
> 
> maybe there could be some confusion in the future if CA uses SF.Fluxes or SF.FluxesandFrictionVelocity, but beta is always in the cache. but it's probably ok!

Yeah, if we ever use other SF types, there will be another dispatch of `set_surface_inputs!`, where we could get the cache we need. For example, right now we have `set_surface_inputs!(sfc_inputs, ::Nothing, args...) = nothing`.

PR #1231, opened 2022-12-27
Add and use NonOrographicGravityWave type

charleskawczynski:
This PR adds and uses a `NonOrographicGravityWave` type, and deletes `tendency_knobs`, since it's no longer needed.

### DIFF
--- a/examples/hybrid/cli_options.jl
+++ b/examples/hybrid/cli_options.jl
@@ -311,7 +311,7 @@ function parse_commandline()
         arg_type = Bool
         default = false
         "--non_orographic_gravity_wave"
-        help = "Apply parameterization for convective gravity wave forcing on horizontal mean flow"
+        help = "Apply parameterization for convective gravity wave forcing on horizontal mean flow [`false` (default), `true`]"
         arg_type = Bool
         default = false
         "--apply_remap"
--- a/examples/hybrid/driver.jl
+++ b/examples/hybrid/driver.jl
@@ -107,12 +107,10 @@ function additional_cache(Y, parsed_args, params, atmos, dt;)
         CA.forcing_cache(Y, forcing_type),
         radiation_cache,
         CA.vertical_diffusion_boundary_layer_cache(Y, atmos),
-        atmos.non_orographic_gravity_wave ?
-        CA.gravity_wave_cache(atmos.model_config, Y, FT) : NamedTuple(),
-        (;
-            tendency_knobs = (;
-                non_orographic_gravity_wave = atmos.non_orographic_gravity_wave
-            )
+        CA.non_orographic_gravity_wave_cache(
+            atmos.non_orographic_gravity_wave,
+            atmos.model_config,
+            Y,
         ),
         (; thermo_dispatcher),
         (; Δt = dt),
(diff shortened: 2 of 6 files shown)


examples/hybrid/driver.jl
```diff
@@ -165,8 +159,7 @@ function additional_tendency!(Yₜ, Y, p, t)
         CA.precipitation_tendency!(Yₜ, Y, p, t, colidx, p.precip_model)
     end
     # TODO: make bycolumn-able
-    (; non_orographic_gravity_wave) = p.tendency_knobs
-    non_orographic_gravity_wave && CA.gravity_wave_tendency!(Yₜ, Y, p, t)
+    CA.gravity_wave_tendency!(Yₜ, Y, p, t, p.atmos.gravity_wave)
```
szy21, line 162: Should we change the `gravity_wave_tendency!` function to dispatch on `atmos.gravity_wave` then?

szy21:
The change looks good. Just keep in mind that in the end we will have both nonorographic and orographic gravity waves, with different tendency functions. Maybe we could just rename the current `gravity_wave` to `nonorographic_gravity_wave` when we have the orographic one ready.

charleskawczynski:
> The change looks good. Just keep in mind that in the end we will have both nonorographic and orographic gravity waves, with different tendency functions. Maybe we could just rename the current `gravity_wave` to `nonorographic_gravity_wave` when we have the orographic one ready.

Ah, ok, they will both be used at the same time?

szy21:
> > The change looks good. Just keep in mind that in the end we will have both nonorographic and orographic gravity waves, with different tendency functions. Maybe we could just rename the current `gravity_wave` to `nonorographic_gravity_wave` when we have the orographic one ready.
> 
> Ah, ok, they will both be used at the same time?

Yes, we will use both of them most of the time. I think it's better to separate them though, e.g. have both `nonorographic_gravity_wave` and `orographic_gravity_wave` arguments, in case we use different schemes in the future. So the current change is fine, we just need better names.

examples/hybrid/cli_options.jl
```diff
@@ -310,9 +310,9 @@ function parse_commandline()
         help = "Check conservation of mass and energy [`false` (default), `true`]"
         arg_type = Bool
         default = false
-        "--non_orographic_gravity_wave"
-        help = "Apply parameterization for convective gravity wave forcing on horizontal mean flow"
-        arg_type = Bool
+        "--gravity_wave"
+        help = "Apply parameterization for convective gravity wave forcing on horizontal mean flow [`false` (default), `NonOrographic`]"
+        arg_type = String
```
jiahe23, line 315: I prefer to keep the `--non_orographic_gravity_wave` option as Boolean and add a `--orographic_drag` option as Bool later.

jiahe23:
> The change looks good. Just keep in mind that in the end we will have both nonorographic and orographic gravity waves, with different tendency functions. Maybe we could just rename the current `gravity_wave` to `nonorographic_gravity_wave` when we have the orographic one ready.

Maybe we could directly rename `gravity_wave` to `nonorographic_gravity_wave` when making the current change. And later add the `orographic_drag` related types/functions.

szy21:
> > The change looks good. Just keep in mind that in the end we will have both nonorographic and orographic gravity waves, with different tendency functions. Maybe we could just rename the current `gravity_wave` to `nonorographic_gravity_wave` when we have the orographic one ready.
> 
> Maybe we could directly rename `gravity_wave` to `nonorographic_gravity_wave` when making the current change. And later add the `orographic_drag` related types/functions.

Yes, that sounds good to me as well.

examples/hybrid/driver.jl
charleskawczynski: Good catch! I've updated that

examples/hybrid/cli_options.jl
charleskawczynski: I've reverted it back to `non_orographic_gravity_wave`, which is still a Bool, and that is later converted to a `NonOrographyGravityWave` or `Nothing` type, depending on the Bool. This will help improve inference in our tendency function.

PR #1234, opened 2023-01-05
kinetic energy conservation changes

simonbyrne:
Following discussion with @tapios and @Zhengyu-Huang, there are some changes that are necessary to ensure kinetic energy conservation.

## Purpose 
Changes the reconstruction in the divergence term to be weighted by the Jacobian, this is necessary for kinetic energy conservation.

## To-do
There are 3 changes we have identified
- [x] use Jacobian-weighted reconstructions of the mass in the advection operator
- [x] in the kinetic energy term, compute $I(w^2)$ instead of $I(w)^2$.
- [x] use a mass-weighted reconstruction of the vertical curl term in the horizontal tendency

## Content

----

### DIFF
--- a/src/TurbulenceConvection/update_aux.jl
+++ b/src/TurbulenceConvection/update_aux.jl
@@ -1,3 +1,5 @@
+import LinearAlgebra
+
 function update_aux!(
     edmf::EDMFModel,
     grid::Grid,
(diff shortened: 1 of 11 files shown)


src/TurbulenceConvection/update_aux.jl
```diff
@@ -141,7 +140,8 @@ function update_aux!(
     end
 
     @. aux_en.e_kin =
-        LA.norm_sqr(C123(prog_gm_uₕ) + C123(Ic(wvec(aux_en_f.w)))) / 2
+        LinearAlgebra.norm_sqr(prog_gm_uₕ) +
+        ᶜinterp(LinearAlgebra.norm_sqr(aux_en_f.w)) / 2
```
tapios, line 144: This KE calculation seems to be repeated many times. It would be good to isolate it into a function.

akshaysridhar:
@simonbyrne following this morning's discussion there are now #1240 #1241 #1242 that split the changes from this PR.

akshaysridhar:
Superseded by 
#1286 
#1287 
#1288

PR #1242, opened 2023-01-17
Add utility function for kinetic energy computation

### LINKED ISSUE #1272: Isolate the function to compute kinetic energy
- [ ] #1242

akshaysridhar:
### Purpose
Introduces a utility function for kinetic energy computation. 
Closes #1272

### DIFF
--- a/examples/hybrid/TurbulenceConvectionUtils.jl
+++ b/examples/hybrid/TurbulenceConvectionUtils.jl
@@ -1,7 +1,6 @@
 module TurbulenceConvectionUtils
 
 using LinearAlgebra
-using LinearAlgebra: norm_sqr
 import ClimaAtmos
 import ClimaAtmos as CA
 import ClimaAtmos.Parameters as CAP
--- a/examples/hybrid/callbacks.jl
+++ b/examples/hybrid/callbacks.jl
@@ -203,10 +203,9 @@ function rrtmgp_model_callback!(integrator)
 
     radiation_model.surface_temperature .= RRTMGPI.field2array(T_sfc)
 
-    C123 = Geometry.Covariant123Vector
     ᶜp = RRTMGPI.array2field(radiation_model.center_pressure, axes(Y.c))
     ᶜT = RRTMGPI.array2field(radiation_model.center_temperature, axes(Y.c))
-    @. ᶜK = LinearAlgebra.norm_sqr(C123(Y.c.uₕ) + C123(ᶜinterp(Y.f.w))) / 2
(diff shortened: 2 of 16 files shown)


akshaysridhar:
Changes affect MSE tables : Could someone please confirm this is acceptable ?

szy21:
My understanding is that we don't need to merge this now? @Zhengyu-Huang ?

akshaysridhar:
> My understanding is that we don't need to merge this now? @Zhengyu-Huang ?

That's my understanding as well - #1240 is the first in the sequence of PRs we need to merge (the weighted interpolation component changes are separated in that one)

tapios:
Could we isolate the KE computation to a function, so we do not have to repeat this code so many times and can apply the changes necessary once?

simonbyrne:
@akshaysridhar rather than try to merge the commit, use `git cherry-pick` to pull out the single commit from #1234

docs/Manifest.toml
```diff
@@ -134,9 +134,15 @@ version = "0.2.1"
 
 [[deps.CUDA]]
 deps = ["AbstractFFTs", "Adapt", "BFloat16s", "CEnum", "CompilerSupportLibraries_jll", "ExprTools", "GPUArrays", "GPUCompiler", "LLVM", "LazyArtifacts", "Libdl", "LinearAlgebra", "Logging", "Printf", "Random", "Random123", "RandomNumbers", "Reexport", "Requires", "SparseArrays", "SpecialFunctions", "TimerOutputs"]
+<<<<<<< HEAD
+git-tree-sha1 = "6717cb9a3425ebb7b31ca4f832823615d175f64a"
+uuid = "052768ef-5323-5732-b1bb-66c8b64840ba"
+version = "3.13.1"
+=======
```
simonbyrne, line 141: Cherry picking the commit will avoid this problem

simonbyrne (changes requested):

tapios:
> Could we isolate the KE computation to a function, so we do not have to repeat this code so many times and can apply the changes necessary once?

We need two things here (or in whatever becomes of this):
1.  Isolate the KE changes to one place (rather than repeating the same changes many times)
2. Document how KE is being computed. (It looks like there is some assumption of orthogonality between horizontal and vertical components, which wouldn't be true in generalized coordinates. I would like to be able to understand what is happening here.)

akshaysridhar:
@simonbyrne Here are the changes related to the `compute_kinetic!` utility; buildkite checks are passing so I'm ready to merge this with your approval.

PR #1254, opened 2023-01-20
Testing filters

trontrytel:
I will be commenting out different filters that I think/hope are not needed.


trontrytel:
Not needed. Replaced by EDMFX regularizations

PR #1258, opened 2023-01-21
Switch the remaining EDMF SCM cases from anelastic to compressible

trontrytel:
Lets see if we can move the rest of EDMF CI to the compressible configuration!


szy21 (approved):
I saw that we keep anelastic TRMM - is that what we want?

trontrytel:
> I saw that we keep anelastic TRMM - is that what we want?

Yes, for now. Our Compressible TRMM case is (a) too long to run in the normal CI, (b) is breaking...

We use the anelastic TRMM case to still have something that runs a deep convective case in the CI

PR #1269, opened 2023-01-25
Remove duplicated code

trontrytel:
I don't see the difference between the two options?

### DIFF
--- a/examples/hybrid/TurbulenceConvectionUtils.jl
+++ b/examples/hybrid/TurbulenceConvectionUtils.jl
@@ -130,25 +130,15 @@ function init_tc!(Y, p, params, colidx)
     C123 = CCG.Covariant123Vector
     t = FT(0)
 
-    if CA.is_anelastic_column(p.atmos)
-        @. p.ᶜp[colidx] = p.edmf_cache.ᶜp₀
+    @. p.ᶜp[colidx] = p.edmf_cache.ᶜp₀
 
-        CA.compute_ref_density!(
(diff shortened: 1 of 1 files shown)


szy21 (approved):
I don't see a difference too 🤔

charleskawczynski:
See https://github.com/CliMA/ClimaAtmos.jl/pull/916/files#diff-5b3935d3b7019d92b43d4746458a9507727e323cf0c3768daa98126392ab4fd2L143-R149 !

trontrytel:
> See https://github.com/CliMA/ClimaAtmos.jl/pull/916/files#diff-5b3935d3b7019d92b43d4746458a9507727e323cf0c3768daa98126392ab4fd2L143-R149 !
> 
> I left it as duplicate for the time to reduce the diff 🙂

Hi @charleskawczynski  :) 

Sounds good. But then maybe we can change the name from `is_anelastic_column` to `is_single_column`? Since we will be doing the same for the compressible single column setups?

charleskawczynski:
> Sounds good. But then maybe we can change the name from is_anelastic_column to is_single_column? Since we will be doing the same for the compressible single column setups?

Yes, I think that's fine, so long it makes sense. I think I wasn't sure if we'd have compressible single column configurations

trontrytel:
Closing. This will be done via https://github.com/CliMA/ClimaAtmos.jl/pull/1327 or https://github.com/CliMA/ClimaAtmos.jl/pull/916

PR #1288, opened 2023-01-26
Weighted interpolation in w tendency equation

akshaysridhar:
Follows design-docs (kinetic energy notes) 
Section 5.2, Equation (39)
Weighted interpolation in w tendency (explicit)


simonbyrne:
I don't think we want this (cf design docs, eqn 19)

akshaysridhar:
> I don't think we want this (cf design docs, eqn 19)

cc @Zhengyu-Huang

PR #1289, opened 2023-01-26
Bugfix in explicit tendency term

akshaysridhar:
Fixes bug introduced in 1240 
Explicit tendency w term.


Zhengyu-Huang (approved):
This needs to be merged ASAP.

PR #1306, opened 2023-01-31
Delete entrainment_massflux_div_factor

trontrytel:
This parameter was added for the dry bubble case, which we don't have anymore.


costachris (approved):
LGTMN

Issue #1305, opened 2023-01-31
Replace NameList with ClimaParameters


nefrathenrici:
Replace generate_namelist and create a parameter struct for the EDMF parameters. Remove all calls to ``parse_namelist``
Depends on: https://github.com/CliMA/CLIMAParameters.jl/pull/103

trontrytel:
Draft PR: https://github.com/CliMA/ClimaAtmos.jl/pull/1311

trontrytel:
Draft PR: https://github.com/CliMA/ClimaAtmos.jl/pull/1377

PR #1311, opened 2023-02-03
Remove hardcoded NameList/EDMF parameters

nefrathenrici:
## Purpose 
This implements #1305 

This PR removes all namelist functionality, replacing it with calls to ClimaParameters. This includes the whole `generate_namelist ` file and all usage of `parse_namelist`. 

The existing `TurbulenceConvectionParameters` struct has been expanded to include all former namelist parameters. This includes a few microphysics parameters as well. In most cases where `namelist` is passed in, I have just replaced it with the expanded struct.

I also changed the EDMF case parsing, passing the case name into `tc_parameter_set`.

----


src/TurbulenceConvection/Parameters.jl
```diff
@@ -23,6 +23,44 @@ Base.@kwdef struct TurbulenceConvectionParameters{FT, MP, SFP} <: ATCP
     microph_scaling_melt::FT # TODO: move to microphysics parameter set? or Microphysics1Moment?
     microphys_params::MP
     surf_flux_params::SFP
+    case_name::String
```
charleskawczynski, line 26: Let's remove the strings since they're GPU-incompatible. We can use types instead, and that way we can also leverage dispatch.

Also, we should use `::Int` over `::Integer` since `Integer` is an abstract type and we should avoid structs with abstract types (see https://docs.julialang.org/en/v1/manual/performance-tips/#Avoid-fields-with-abstract-type)

examples/Manifest.toml
```diff
@@ -157,9 +157,9 @@ version = "0.1.2"
 
 [[deps.CLIMAParameters]]
 deps = ["DocStringExtensions", "TOML", "Test"]
-git-tree-sha1 = "8cb1d69b6b5fc41be48bea969a68763a17843f1e"
+git-tree-sha1 = "a769996f7ed7a2f1fb070c4837370603387e5eb6"
```
charleskawczynski, line 160: We can make these changes in a separate `julia .dev/up_deps.jl` PR, since changes miss the other environments (e.g., docs, test, perf)

charleskawczynski: I'll open one now

charleskawczynski: Opened #1335

src/TurbulenceConvection/Parameters.jl
charleskawczynski: Also, `case` is not needed, nor is it currently used, past initialization, and we don't want simulations to be case-dependent (this is why _all_ case information in SCAMPy was in a `Cases` file). We should avoid making it available in our RHS functions to ensure it doesn't happen.

nefrathenrici: I can turn the String fields into types, and fix the Integer fields.
Should I get rid of the `case_name `and `case ` occurrences entirely then?

examples/Manifest.toml
trontrytel: The above PR got merged, So we should be able to rebase this one

src/TurbulenceConvection/Parameters.jl
charleskawczynski: It looks like it's being deleted in #1327, maybe the simplest thing for now is to grab `case_name` from `parsed_args` (which is available in `turbconv_cache`). Do we need to do anything with `case`?

nefrathenrici: I don't think we do, I couldn't find any examples of it actually being used.

src/TurbulenceConvection/Parameters.jl
```diff
@@ -8,22 +8,21 @@ import Thermodynamics as TD
 import SurfaceFluxes as SF
 import CloudMicrophysics as CM
 
+include("../model_getters.jl")
```
charleskawczynski, line 11: We probably shouldn't re-include this, why is this necessary?

charleskawczynski: (same as below)

src/TurbulenceConvection/types.jl
```diff
@@ -1,3 +1,5 @@
+include("../model_getters.jl")
+import ClimaAtmos.TurbulenceConvection.Parameters as TCP
```
charleskawczynski, line 2:
```suggestion
import .Parameters as TCP
```
I think this should work, we shouldn't ever need / use `import ClimaAtmos.anything` inside `src/` of `ClimaAtmos`. It technically opens the door to use multiple versions of ClimaAtmos.

charleskawczynski: (same elsewhere)

src/model_getters.jl
```diff
@@ -359,3 +359,59 @@ function ThermoDispatcher(atmos)
...
+    elseif precip_fraction_model_name == "cloud_cover"
+        return DiagnosticPrecipFraction(paramset.precip_fraction_limiter)
+    else
+        error(
+            "Something went wrong. Invalid `precip_fraction` model: `$precip_fraction_model_name`",
+        )
+    end
+end
+
+function AbstractQuadratureType(s::String)
```
charleskawczynski, line 390: This is a confusing name, because Julia prohibits creating instances of abstract types, can we just call it `QuadratureType`?

charleskawczynski: Also, why not just pass in a Dict and read it from the Dict? Can we also error if none of the cases are satisfied?

nefrathenrici: Maybe it makes sense to just move it all into the TurbulenceConvectionParameters? It shouldn't get used anywhere else.

PR #1327, opened 2023-02-07
Unify initial conditions of dycore and EDMF

### LINKED ISSUE #1295: Unify initial condition handling between EDMF-SCM, EDMF on a sphere and ClimaAtmos.
EDMF in ClimaAtmos is initialized via tc_init and followed by a bunch of overwrites to the model state and aux. We should get rid of this and be able to set the ClimaAtmos and EDMF state based on the SCM cases setup or the spherical simulation setup.

We should:
- get rid of the convoluted overwrites
- unify the pressure calculations between the initial value problem solved in reference state and cases, and move it to ClimaAtmos.

dennisYatunin:
## Purpose 
Make EDMF and the dycore use a single interface for defining and setting initial conditions. Close #1295.

Fix two general issues with our current implementation of initial conditions:
 - The code for initializing EDMF involves many computations and re-computations of variables, a lot of which ultimately go unused, and the sequence of these computations is hard to follow. This makes it difficult to debug the code.
 - Every block of code that specifies an initial condition makes numerous assumptions about the structure of the atmosphere model. This makes it difficult to extend the model without also modifying all of the code for initial conditions.

## Content
 - Add a struct, `LocalState`, for initial conditions to use in order to specify the state at a particular point, but without needing to know about the model configuration.
 - Add a function, `atmos_state`, that converts the generic representation of a state into a model-specific one. This is essentially a combination of the code that is currently [here](https://github.com/CliMA/ClimaAtmos.jl/blob/6daa7c3d9344e78b4ca56cfd1f4ff5173835812a/src/InitialConditions/InitialConditions.jl#L24-L84) and [here](https://github.com/CliMA/ClimaAtmos.jl/blob/6daa7c3d9344e78b4ca56cfd1f4ff5173835812a/src/TurbulenceConvection/variables.jl#L164-L200), except that it now initializes all prognostic variables and is written in terms of `LocalState`.
 - Make all initial conditions currently used in ClimaAtmos into objects of the same type (`InitialCondition`), and rewrite them in terms of `LocalState`. This involves combining what is currently `src/InitialConditions/baro_wave.jl`, `src/InitialConditions/sphere.jl`, `src/InitialConditions/box.jl`, `src/InitialConditions/single_column.jl`, and `tc_driver/Cases.jl` into a single file, `src/InitialConditions/initial_conditions.jl`. In the process, a lot of redundant and unnecessary code is removed. Since TurbulenceConvection uses its own surface parametrization, which is distinct from the rest of ClimaAtmos, the `surface_params` methods from `tc_driver/Cases.jl` are also moved into `src/InitialConditions/initial_conditions.jl`. Unifying our surface parametrizations could be a good follow-up task.
 - Modify `examples/hybrid/driver.jl`, `examples/hybrid/TurbulenceConvectionUtils.jl`, `examples/hybrid/types.jl`, and `src/discrete_hydrostatic_balance.jl` to use the new initial conditions interface. Remove all other code related to the old interface; e.g., the `get_state_fresh_start` function and the `TurbulenceConvectionUtils.init_tc!` function (along with everything it calls, like `initialize_edmf`, `initialize_covariance`, and `initialize_updrafts`). Keep some of the logic from `get_state_fresh_start` in `get_initial_condition`.
 - Remove all code related to the surface reference state that was removed in #1330: `src/ref_state.jl`, as well as the `set_thermo_state_pθq!` and `set_grid_mean_from_thermo_state!` functions in `tc_driver/dycore.jl`.
 - Run the `tc_callback` before the first step in order to ensure that the boundary conditions are satisfied.
 - Fix a bug in the old implementation that was allocating `Y.c.ρq_tot` whenever the EDMF model was in use, even when the moisture model was a `DryModel`. This required making it so that the `dif_flux_ρq_tot` and `dif_flux_ρq_tot_bc` cached fields and the `sfc_evaporation` diagnostic field are only allocated when `:ρq_tot in propertynames(Y.c)`, making it so that the EDMF model only sets the tendency of `Y.c.ρq_tot` when it is available, and making it so that the remap pipeline only uses `sfc_evaporation` when it is available.
 - Modify the "explicit EDMF on a sphere" CI example by renaming it from `sphere_baroclinic_wave_rhoe_equilmoist_mo_edmf` to `sphere_baroclinic_wave_rhoe_equilmoist_edmf` and removing the `--turbconv_case Soares` flag (this PR makes the flag unnecessary; this example will now use the baroclinic wave initial condition, without overwriting it with the Soares initial condition).

----


trontrytel:
Lets remember to delete https://github.com/CliMA/ClimaAtmos.jl/blob/main/src/ref_state.jl when we are done. It should not be needed

src/InitialConditions/initial_conditions.jl
```diff
@@ -0,0 +1,763 @@
...
+    grav = TD.Parameters.grav(thermo_params)
+    θ_virt(z) = TD.virtual_pottemp(
+        thermo_params,
+        TD.PhaseEquil_pθq(thermo_params, p(z), θ(z), q_tot(z)),
+    )
+    θ_virt_0 = θ_virt(FT(0))
+    Ri_bulk(z) =
+        grav * (θ_virt(z) - θ_virt_0) * z / θ_virt_0 / (u(z)^2 + v(z)^2)
+    Ri_bulk_crit = FT(0.2)
+    z_values = FT(0):FT(100):FT(30e3)
```
charleskawczynski, line 486: Let's not hard-code numbers for zmax here

src/InitialConditions/initial_conditions.jl
```diff
@@ -0,0 +1,763 @@
...
+struct GABLS <: InitialCondition end
+
+for IC_Type in (:Dycoms_RF01, :Dycoms_RF02, :GABLS)
+    IC_Type_caps = Symbol(uppercase(string(IC_Type)))
+    θ_func_name = Symbol(IC_Type, :_θ_liq_ice)
+    q_tot_func_name = Symbol(IC_Type, :_q_tot)
+    u_func_name = Symbol(IC_Type, IC_Type == :Dycoms_RF01 ? :_u0 : :_u)
+    v_func_name = Symbol(IC_Type, IC_Type == :Dycoms_RF01 ? :_v0 : :_v)
+    tke_func_name = Symbol(IC_Type, :_tke)
+    @eval function (initial_condition::$IC_Type_caps)(params)
```
charleskawczynski, line 426: Please somehow remove the `@eval`.

src/InitialConditions/initial_conditions.jl
```diff
@@ -0,0 +1,763 @@
...
+hydrostatically balanced pressure profile.
+"""
+struct ARM_SGP <: InitialCondition end
+
+for IC_Type in (:Soares, :Bomex, :LifeCycleTan2018, :ARM_SGP)
+    θ_func_name = Symbol(IC_Type, :_θ_liq_ice)
+    q_tot_func_name = Symbol(IC_Type, :_q_tot)
+    u_func_name = Symbol(IC_Type, :_u)
+    tke_func_name = Symbol(IC_Type, :_tke)
+    @eval function (initial_condition::$IC_Type)(params)
```
charleskawczynski, line 367: Please somehow remove the `@eval`.

src/InitialConditions/initial_conditions.jl
```diff
@@ -0,0 +1,763 @@
...
+    ts(p, z, T, ::Nothing, ::Nothing) = TD.PhaseDry_pT(thermo_params, p, T(z))
+    ts(p, z, ::Nothing, θ, ::Nothing) = TD.PhaseDry_pθ(thermo_params, p, θ(z))
+    ts(p, z, T, ::Nothing, q_tot) =
+        TD.PhaseEquil_pTq(thermo_params, p, T(z), q_tot(z))
+    ts(p, z, ::Nothing, θ, q_tot) =
+        TD.PhaseEquil_pθq(thermo_params, p, θ(z), q_tot(z))
+    dp_dz(p, _, z) =
+        -grav * TD.air_density(thermo_params, ts(p, z, T, θ, q_tot))
+
+    z_span = (FT(0), FT(30e3)) # assume no one will ever need to go above 30km
```
charleskawczynski, line 261: The grid information should probably be passed in, we have "high-top" simulations, so people may run into strange results with this assumption.

dennisYatunin: Making the initial conditions depend on the grid information results in more complicated code, and it also throws off convergence studies. The profile to which we are interpolating should probably not depend on the model configuration.

src/InitialConditions/initial_conditions.jl
dennisYatunin: We know that the target z will be in the first few kilometers above the surface, so it's pretty safe to add an upper bound of tens of kilometers. Also, we should probably not make the initial conditions depend on the model configuration.

src/InitialConditions/initial_conditions.jl
dennisYatunin: Also, the input profiles from `Cases.jl` don't work above 30 km anyway, which is why I used that value.

src/InitialConditions/initial_conditions.jl
charleskawczynski: Maybe this is fine actually

src/InitialConditions/initial_conditions.jl
charleskawczynski: Maybe this is fine actually

src/InitialConditions/initial_conditions.jl
charleskawczynski: @trontrytel's voted to keep this as is. Let's roll with it.

src/InitialConditions/initial_conditions.jl
charleskawczynski: @trontrytel's voted to keep this as is. Let's roll with it.

dennisYatunin:
bors cancel

PR #1329, opened 2023-02-08
Aj/filter flags

trontrytel:
Add flags for the first four edmf filters to help debug implicit simulations.

### DIFF
--- a/src/TurbulenceConvection/EDMF_functions.jl
+++ b/src/TurbulenceConvection/EDMF_functions.jl
@@ -723,6 +723,7 @@ function filter_updraft_vars(
     prog_gm = center_prog_grid_mean(state)
     aux_gm_f = face_aux_grid_mean(state)
     aux_gm = center_aux_grid_mean(state)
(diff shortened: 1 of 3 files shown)


src/TurbulenceConvection/EDMF_functions.jl
```diff
@@ -735,10 +736,32 @@ function filter_updraft_vars(
...
+                aux_bulk.filter_flag_3[k] = 1
+            else
+                aux_bulk.filter_flag_3[k] = 0
+            end
+            if prog_up[i].ρarea[k] > ρ_c[k] * a_max
+                aux_bulk.filter_flag_4[k] = 1
+            else
+                aux_bulk.filter_flag_4[k] = 0
+            end
+        end
```
charleskawczynski, line 760: Can we write this as broadcast expressions instead? This relies on `getindex`, and will need to be changed.

trontrytel: Will do!

PR #1330, opened 2023-02-08
Remove surface_reference_thermo_state

dennisYatunin:
## Purpose 

This PR removes `surface_reference_thermo_state` from the EDMF code, along with all related values. Prelude to #1327.

@trontrytel has checked the papers for various test cases to resolve inconsistencies between `surface_params` and `surface_reference_thermo_state`. The results have slightly changed for several test cases, but the changes are small enough to be attributable to roundoff error.

----
- [x] I have read and checked the items on the review checklist.

PR #1331, opened 2023-02-08
Remove `p_ivp`

charleskawczynski:
This PR removes `p_ivp`, and instead relies on the pressure defined in `log_pressure_profile` (we're currently re-defining pressure).

I think if we were to merge this, that #1330 would not work.

### DIFF
--- a/tc_driver/Cases.jl
+++ b/tc_driver/Cases.jl
@@ -95,50 +95,6 @@ get_case_name(case_type::AbstractCaseType) = string(case_type)
 ##### Pressure helper functions for making initial profiles hydrostatic.
 #####
 
-"""
-    Pressure derivative with height assuming:
-    - hydrostatic
-    - given θ_liq_ice and q_tot initial profiles
-"""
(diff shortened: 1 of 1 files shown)


trontrytel:
I don't think thats correct? We need to solve the p_ivp as defined per each paper based test case

trontrytel:
I think that what is done in the log_pressure_profile assumes T=constant going from the surface up. This is not a good profile from the paper cases. We should be using the p_ivp solution that uses the temperature and humidity profiles prescribed in each case.

charleskawczynski:
Right, this is basically reverting [724](https://github.com/CliMA/ClimaAtmos.jl/pull/724), which had extensive review. I think the `p_ivp` is different from the pressure log function because the moisture profile in the case is used as the reference moisture profile. Closing.

charleskawczynski:
We probably should have removed the log pressure fun in #724, though, since it introduced redefining the pressure. Removing it could be incorporated into #1330.

PR #1339, opened 2023-02-10
entrainment_detrainment cleanup

trontrytel:
This PR simplifies the entrainment detrainment logic, so I can work easier with the code. 

Previously the computations were split over many different functions that were only used once. There is no reason to define them separately. Also the input model variables were packed into a separate object with some name changes along the way. That was also confusing.

I'm about to be adding different detrainment options and it will help me to have all detrainment physics happening in one place.

Also, this is just refactoring. The CI cases should be the same


charleskawczynski (commented):
Maybe @costachris has a better idea. The changes make the code simpler to follow, but it also seems less flexible to extend. My only concern is: will we just revert this PR in the future when adding new entr-detr models back in?

trontrytel:
> Maybe @costachris has a better idea. The changes make the code simpler to follow, but it also seems less flexible to extend. My only concern is: will we just revert this PR in the future when adding new entr-detr models back in?

I'll be adding different detrainment options to the code now and the changes make it easier for me to do so. 

Once we have implicit edmf working we can think of adding back some more options. But I think it will be a while before it happens.

costachris:
The functionally to easily switch between all the entr/detr models originally in standalone TC was removed in the big ClimaAtmos TC code delete some time ago (there were a bunch of objects in types.jl, for example). So all the general entr/detr functions and intermediate objects aren't really serving a purpose anymore in ClimaAtmos TC. I think this is a good solution for now, since realistically we won't need to test every permutation of detr/entr dimensional scales and non-dimensional functions. We can and should add some of the flexibility/generalizability back after everything is working, but probably won't need the level of complexity we had in standalone TC.

src/TurbulenceConvection/closures/entr_detr.jl
```diff
@@ -155,48 +46,92 @@ function compute_entr_detr!(
...
+                else
+                    εδ_closure.params.c_δ
+                end
+
+                # Define helper variables
+                w_up = w_up_c[k] # updraft vertical velocity
+                w_en = w_en_c[k] # environment vertical velocity
+                b_up = aux_up[i].buoy[k] # updraft buoyancy
+                b_en = aux_en.buoy[k] # environment buoyancy
+                dMdz = ∇m_entr_detr[k] # updraft momentum divergence
```
costachris, line 88: I don't think we're using `dMdz` anymore

src/TurbulenceConvection/closures/entr_detr.jl
```diff
@@ -155,48 +46,92 @@ function compute_entr_detr!(
...
+                M = m_entr_detr[k] # updraft momentum
```
costachris, line 89: Not using `M`

charleskawczynski (approved):

costachris (approved):

src/TurbulenceConvection/closures/entr_detr.jl
trontrytel: Nice! Two more lines to delete!

PR #1350, opened 2023-02-13
Add job for testing n-inference failures

charleskawczynski:
This PR adds a JET test to ensure that we don't increase the number of JET failures in `step!`

### DIFF
--- a/perf/jet_nfailures.jl
+++ b/perf/jet_nfailures.jl
@@ -1,0 +1,46 @@
+import Profile
+
+ENV["CI_PERF_SKIP_RUN"] = true # we only need haskey(ENV, "CI_PERF_SKIP_RUN") == true
+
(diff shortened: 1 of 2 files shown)


perf/jet_nfailures.jl
```diff
@@ -0,0 +1,46 @@
...
+# dict = parsed_args_per_job_id();                           # if job_id uses driver.jl
+# parsed_args = dict["sphere_aquaplanet_rhoe_equilmoist_allsky"];
+
+try
+    include(filename)
+catch err
+    if err.error !== :exit_profile
+        rethrow(err.error)
+    end
+end
```
simonbyrne, line 18: I'd like a better way to do this....

simonbyrne (approved):

perf/jet_nfailures.jl
charleskawczynski: Yeah, we can probably do something cleaner once we further simplify the driver

PR #1356, opened 2023-02-14
Fix environment area initialization

dennisYatunin:
## Purpose 

The process of debugging the TRMM case for #1327 has revealed that there are several bugs in the EDMF initial conditions. Here is the first bugfix: the environment area fraction was previously being set to 1 (`aux_bulk.area` was initialized to 0 instead of the correct value), and now it will be set to `1 - edmf.minimum_area * TC.n_updrafts(edmf)`.

----


trontrytel (approved):

tc_driver/initial_conditions.jl
```diff
@@ -35,15 +35,12 @@ function initialize_covariance(
...
-
-    kc_surf = TC.kc_surface(grid)
     aux_gm = TC.center_aux_grid_mean(state)
     prog_en = TC.center_prog_environment(state)
     aux_en = TC.center_aux_environment(state)
     prog_gm = TC.center_prog_grid_mean(state)
     ρ_c = prog_gm.ρ
-    aux_bulk = TC.center_aux_bulk(state)
-    ae = 1 .- aux_bulk.area # area of environment
+    ae = 1 - edmf.minimum_area * TC.n_updrafts(edmf)
```
charleskawczynski, line 43: This doesn’t look correct for a generic number of updrafts, where is this formula from?

charleskawczynski: I agree this should be fixed, though!

charleskawczynski: Oh, wait, it is correct. This is based on the domain decomposition equation, correct?

charleskawczynski (approved):

tc_driver/initial_conditions.jl
dennisYatunin: [This line](https://github.com/CliMA/ClimaAtmos.jl/blob/5c6e776c041e3718c85f25c87744e406206d174e/src/TurbulenceConvection/update_aux.jl#L116) and [this line](https://github.com/CliMA/ClimaAtmos.jl/blob/5c6e776c041e3718c85f25c87744e406206d174e/tc_driver/initial_conditions.jl#L83)

PR #1365, opened 2023-02-15
move remap to separate command

### LINKED ISSUE #1317: Move remapping to separate command
The current remapping step in the longrun job is [called as part of the driver](https://github.com/CliMA/ClimaAtmos.jl/blob/ce47d12c88f4455c41f5852317a4094a3dbb8c92/examples/hybrid/driver.jl#L441). Unfortunately this triggers an error:
https://buildkite.com/clima/climaatmos-longruns/builds/759#0185aee3-e22c-4c23-9840-82a749974ca7/140-471
```

HDF5-DIAG: Error detected in HDF5 (1.12.1) thread 0:
  #000: H5F.c line 711 in H5Fclose(): decrementing file ID failed
    major: File accessibility
    minor: Unable to close file
  #001: H5Iint.c line 1018 in H5I_dec_app_ref(): can't decrement ID ref count
    major: Object atom
    minor: Unable to decrement reference count
  #002: H5Iint.c line 949 in H5I_dec_ref(): can't locate ID
    major: Object atom
    minor: Unable to find atom information (already closed?)
```

I'm not quite sure why this is failing: my main hypothesis is that written file isn't completely closed on some MPI processes, so when it is read for remapping it triggers an error.

The easiest solution is to make the remapping a completely separate script, which should be run after the driver (either as part of the current buildkite step, or as a separate step which depends on the model run step (the latter is probably preferable in cases where you are using large CPU resources).

simonbyrne:
## Purpose 
Move remapping to separate command in longruns script. Fixes #1317.

## Content

----

### DIFF
--- a/.buildkite/longruns/pipeline.yml
+++ b/.buildkite/longruns/pipeline.yml
@@ -60,7 +60,9 @@ steps:
           slurm_ntasks: 32
 
       - label: ":computer: baroclinic wave (ρe_tot) equilmoist high resolution"
-        command: "mpiexec julia --project=examples examples/hybrid/driver.jl --moist equil --precip_model 0M --z_elem 45 --dz_bottom 30 --h_elem 16 --kappa_4 1e16 --dt 400secs --t_end 100days --dt_save_to_disk 10days --job_id longrun_bw_rhoe_equil_highres --apply_remap true"
+        command:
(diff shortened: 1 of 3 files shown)


simonbyrne:
Longrun build: https://buildkite.com/clima/climaatmos-longruns/builds/805

simonbyrne:
This seems to work:
https://buildkite.com/clima/climaatmos-longruns/builds/805#01865741-8cd6-48f8-85b2-cf31c5c6b3b1

@szy21 are these files ok?

simonbyrne:
Also, should we delete `apply_remap` option from the driver? After this change, it won't be used anywhere.

szy21 (approved):
The files look great, thanks! We should remove `apply_remap`. You can do it in this PR, but don't worry if you don't have time for that. I can also remove it in another PR.

PR #1376, opened 2023-02-17
Simplify init_tc! to facilitate debugging

dennisYatunin:
## Purpose 

This PR simplifies `init_tc!` in order to facilitate the debugging of the TRMM case in #1327. Following the bug fixes in #1356 and #1327, this should not change any of the CI cases. Since `init_tc!` is going to be removed in #1327, the specifics of this simplified implementation don't really matter.

## Content

- Move all of the code from `set_thermo_state_pθq!` and `set_grid_mean_from_thermo_state!` to `init_tc!`, and remove those functions from `dycore.jl`.
- Move all of the code from `initialize_edmf` (and, by extension, `initialize_updrafts` and `initialize_covariance`) to `init_tc!`, and delete `initial_conditions.jl`.
- Remove all unnecessary manipulations of auxiliary variables from `init_tc!`, and remove the call to `assign_thermo_aux!`.
- Add a sanity check to the beginning of `init_tc!` that sets all uninitialized auxiliary variables to `NaN`.
- Replace the uninitialized auxiliary variable `aux_gm.q_liq` with `0` in `ARM_SGP`'s `initialize_profiles` method.
- Clearly label every prognostic variable being manipulated in `init_tc!`, and ensure that it is only set once.

----


examples/hybrid/TurbulenceConvectionUtils.jl
```diff
@@ -102,31 +101,81 @@ function init_tc!(Y, p, params, colidx)
...
+        @. Y.c.turbconv.up[i].ρaθ_liq_ice[colidx] =
+            Y.c.turbconv.up[i].ρarea[colidx] * ᶜθ_liq_ice[colidx]
+        @. Y.f.turbconv.up[i].w[colidx] =
+            CC.Geometry.Covariant3Vector(CC.Geometry.WVector(FT(0)), lg[colidx])
+    end
+    a_en = (1 - N_up * a_min)
+    @. Y.c.turbconv.en.ρatke[colidx] = Y.c.ρ[colidx] * a_en * ᶜtke[colidx]
+    if p.atmos.turbconv_model.thermo_covariance_model isa
+        TC.PrognosticThermoCovariances
+        @. Y.c.turbconv.en.ρaHvar[colidx] = Y.c.ρ[colidx] * a_en * ᶜHvar[colidx]
```
trontrytel, line 166: Would you also need the `QTvar` and `HQTcov`?

That's what is deleted from initial conditions below:
```
    if edmf.thermo_covariance_model isa TC.PrognosticThermoCovariances
        prog_en.ρaHvar .= aux_gm.Hvar .* ρ_c .* ae
        prog_en.ρaQTvar .= aux_gm.QTvar .* ρ_c .* ae
        prog_en.ρaHQTcov .= aux_gm.HQTcov .* ρ_c .* ae
```

examples/hybrid/TurbulenceConvectionUtils.jl
```diff
@@ -102,31 +101,81 @@ function init_tc!(Y, p, params, colidx)
...
+    # parent(state.aux.cent) .= NaN
 
-    # TODO: convert initialize_profiles to set prognostic state, not aux state
+    # Initialization of auxiliary variables:
+    (; ᶜp, ᶜts) = p
+    ᶜθ_liq_ice = p.edmf_cache.aux.cent.θ_liq_ice
+    ᶜq_tot = p.edmf_cache.aux.cent.q_tot # only initialized for moist profiles
+    ᶜtke = p.edmf_cache.aux.cent.tke
+    ᶜHvar = p.edmf_cache.aux.cent.Hvar # only initialized for GABLS
+    ᶜHvar .= 0
```
trontrytel, line 120: You are overwriting it with 0 anyway?

examples/hybrid/TurbulenceConvectionUtils.jl
```diff
@@ -102,31 +101,81 @@ function init_tc!(Y, p, params, colidx)
...
+    a_min = edmf.minimum_area
+    lg = CC.Fields.local_geometry_field(axes(Y.f))
+    @inbounds for i in 1:N_up
+        @. Y.c.turbconv.up[i].ρarea[colidx] = Y.c.ρ[colidx] * a_min
+        @. Y.c.turbconv.up[i].ρaq_tot[colidx] =
+            Y.c.turbconv.up[i].ρarea[colidx] * ᶜq_tot[colidx]
+        @. Y.c.turbconv.up[i].ρaθ_liq_ice[colidx] =
+            Y.c.turbconv.up[i].ρarea[colidx] * ᶜθ_liq_ice[colidx]
+        @. Y.f.turbconv.up[i].w[colidx] =
+            CC.Geometry.Covariant3Vector(CC.Geometry.WVector(FT(0)), lg[colidx])
```
trontrytel, line 160: IN the deleted code we also had this hack:
```
        a_surf = TC.area_surface_bc(surf, edmf, i)
        aux_up[i].area[kc_surf] = a_surf
        prog_up[i].ρarea[kc_surf] = ρ_c[kc_surf] * a_surf
```

Do you think this could be the reason why the results are changed here?

trontrytel:
In theory this PR should not be changing any results in the CI, right?

examples/hybrid/TurbulenceConvectionUtils.jl
dennisYatunin: Yes, but these other values weren't being initialized in `initialize_profiles`. I've added them in anyway, but set them to 0.

examples/hybrid/TurbulenceConvectionUtils.jl
dennisYatunin: Yeah, because it is only initialized for GABLS, but the value could still be used for other cases, as long as the model requires prognostic covariances.

examples/hybrid/TurbulenceConvectionUtils.jl
dennisYatunin: This operation is already being performed in `set_edmf_surface_bc`.

dennisYatunin:
In theory, yes. However, it looks like the only changes are in the last digits of a couple of values in the initial state, so we can just attribute them to roundoff error. This should be good to merge in.

trontrytel:
I just looked at the profiles from the CI, and everything looks fine to me. There is just some plotting issue with the Nieuwstadt case?

dennisYatunin:
The plotting issue was that `ρaq_tot` was not being initialized for the dry case. It's fixed now.

PR #1377, opened 2023-02-17
Replace namelist usage, expand TCParams struct

### LINKED ISSUE #1305: Replace NameList with ClimaParameters

nefrathenrici:
## Purpose 
This implements #1305 

This PR removes all namelist functionality, replacing it with calls to ClimaParameters. This includes the whole `generate_namelist ` file and all usage of `parse_namelist`. 

The existing `TurbulenceConvectionParameters` struct has been expanded to include all former namelist parameters. This includes a few microphysics parameters as well. In most cases where `namelist` is passed in, I have just replaced it with the expanded struct.

This replaces PR #1311.
Closes #1305 

----


parameters/create_parameters.jl
```diff
@@ -161,5 +167,5 @@ function create_climaatmos_parameter_set(
     )
     # logfilepath = joinpath(@__DIR__, "logfilepath_$FT.toml")
     # CP.log_parameter_information(toml_dict, logfilepath)
-    return param_set
+    return param_set, config_params
```
trontrytel, line 170: I like the split between config params and parameters that need to be passed around in the model. Maybe it would be good to add a docstring somewhere that explains what that function is doing? 

We still have the `examples/hybrid/parameter_set.jl` file with `create_parameter_set`, `baro_wave_parameter_set` and `tc_parameter_set`. Could you also bring some sanity there? Maybe just overriding things as done for the `is_column_without_edmf(parsed_args)` case without specifying the functions for it?

parameters/create_parameters.jl
```diff
@@ -109,22 +109,28 @@ function create_climaatmos_parameter_set(
...
 
+    config_param_names = [
+        "sgs",
+        "quadrature_type",
+        "quadrature_order",
+        "precip_fraction_model",
+        "prescribed_precip_frac",
+        "precip_fraction_limiter",
+        "thermo_covariance_model",
+        "diagnostic_covar_limiter"
```
trontrytel, line 129: The `quadrature_order`, `prescribed_precip_frac`, `precip_fraction_limiter` and `diagnostic_covar_limiter` are all numbers. So they could still be part of the `param_set` and not `config_params`?

examples/hybrid/driver.jl
```diff
@@ -128,7 +120,6 @@ function additional_cache(Y, parsed_args, params, atmos, dt;)
             Y,
             turbconv_model,
             atmos,
-            namelist,
```
trontrytel, line 131: Since now we have the `params` for numbers and `config` for flags, would it make sense (at some point) to move the data from `parsed_args` into either of the two and pass them into `turbcov_cache`, etc?

trontrytel (approved):
LGTM! I left some small comments/questions. Would be great to get a second pair of eyes on it from the software engineers.

parameters/create_parameters.jl
nefrathenrici: They seemed to only get used for constructing the data structs, and I thought that we should use the `config_params `for model construction, but I can move them back into the `TurbulenceConvectionParameters` if that makes more sense.

trontrytel: Either way works for me. I'm fine with them being config params too if there is a good software engineering reason for it.

PR #1381, opened 2023-02-21
Explicit assignment for b.c. in precomputed quantities 

### LINKED ISSUE #1367: Correctly set vertical velocity at cell boundaries
Find a way to set $u_v$ at cell boundaries so that $\tilde{u}^v = 0$. Can fake it for now, but should add a boundary condition to ClimaCore that handles this.

akshaysridhar:
Closes #1367. Requires preliminary review to see if this is appropriate way to do this in ClimaAtmos - I think we should be able to this in ClimaCore. 
cc @simonbyrne 
cc @Zhengyu-Huang

### DIFF
--- a/src/precomputed_quantities.jl
+++ b/src/precomputed_quantities.jl
@@ -12,13 +12,29 @@ function precomputed_quantities!(Y, p, t)
     end
 end
 
+function set_boundary_velocity!(uₕ, uᵥ)
+    uₕ_surface_data = Fields.level(Fields.field_values(uₕ), 1)
+    uₕ_surface_geom = Fields.level(Spaces.local_geometry_data(axes(uₕ)), 1)
+    uᵥ_surface_data = Fields.level(Fields.field_values(uᵥ), 1)
+    uᵥ_surface_geom = Fields.level(Spaces.local_geometry_data(axes(uᵥ)), 1)
+
+    @. uᵥ_surface_data = Geometry.Covariant3Vector(
+        -Geometry.contravariant3(uₕ_surface_data, uₕ_surface_geom) /
+        Geometry.contravariant3(one(uᵥ_surface_data), uᵥ_surface_geom),
+    )
+end
+
 function precomputed_quantities!(Y, p, t, colidx)
     ᶜuₕ = Y.c.uₕ
     ᶠw = Y.f.w
+
     (; ᶜK, ᶠu_tilde, ᶜu_bar, ᶠu³, ᶜts, ᶜp, params) = p
     (; ᶠwinterp, ᶜinterp) = p.operators
     ᶜJ = Fields.local_geometry_field(axes(Y.c.ρ)).J
     C123 = Geometry.Covariant123Vector
+
+    set_boundary_velocity!(ᶜuₕ[colidx], ᶠw[colidx])
+
     @. ᶜu_bar[colidx] = C123(ᶜuₕ[colidx]) + C123(ᶜinterp(ᶠw[colidx]))
     @. ᶠu_tilde[colidx] =
         ᶠwinterp(
--- a/docs/src/equations.md
+++ b/docs/src/equations.md
@@ -76,6 +76,11 @@ We make use of the following operators
 * ``p`` is air pressure, derived from the thermodynamic state, reconstructed at cell centers.
 * ``\boldsymbol{F}_R`` are the radiative fluxes: these are assumed to align vertically (i.e. the horizontal contravariant components are zero), and are constructed at cell faces from [RRTMGP.jl](https://github.com/CliMA/RRTMGP.jl).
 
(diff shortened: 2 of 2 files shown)


src/precomputed_quantities.jl
```diff
@@ -15,10 +16,24 @@ end
...
+    u₃_bc[colidx] .= @. Geometry.Covariant3Vector(
+        -1 * (
+            g¹³[colidx] * ᶠuₕ_bc.components.data.:1[colidx] +
+            g²³[colidx] * ᶠuₕ_bc.components.data.:2[colidx]
+        ) / g³³[colidx],
+    )
+    enforce_boundary = Operators.SetBoundaryOperator(
+        bottom = Operators.SetValue(u₃_bc[colidx]),
+    )
+    @. ᶠw[colidx] = enforce_boundary(ᶠw[colidx])
```
simonbyrne, line 36: I don't think this is right: to be consistent, we want to define it so that $\bar{u}^3 = 0$. 

1) interpolate the 3rd contravariant component of $u_h$ to he boundary: since $I(\rho J)$  at the boundary is equal to $\rho J$ at the first lecel, we can just use regular (unweighted) interpolation:
```julia
@. Fields.level(ᶠinterp(Geometry.contravariant3(ᶜuₕ[colidx])), half)
```
2) divide this by the 3rd contravariant component of a unit `Covariant3Vector` at the boundary:
```julia
@. Fields.level(Geometry.contravariant3(one(ᶠw[colidx])), half)
```

3) Set the boundary to velocity to be `- (1) / (2)`. Putting this altogether:
```julia
ᶠw_surface = level(ᶠw, half)
@. ᶠw_surface = Geometry.Covariant3Vector(Fields.level(ᶠinterp(Geometry.contravariant3(ᶜuₕ)), half) / Geometry.contravariant3(one(ᶠw_surface)))
```

akshaysridhar: > 

Agreed that this changes now that we have u_bar and u_tilde defined following the 1364 merge - I think the above is true with the u_tilde^3 b.c. rather than the u_bar^3 bc

simonbyrne: yes, I meant u_tilde, not u_bar!

akshaysridhar:
@simonbyrne current buildkite allocation issues are here : https://buildkite.com/clima/climaatmos-ci/builds/7502#01868203-3f53-4aa8-a33a-745035f8d884

akshaysridhar:
Is it useful to do this outside our `column` loops ? https://buildkite.com/clima/climaatmos-ci/builds/7510#01868452-b6c0-40a3-9cdd-87bbbd08d565 ?

simonbyrne:
Do you get allocations if you use:
```julia
ᶠw_surface = level(ᶠw, half)
@. ᶠw_surface = Geometry.Covariant3Vector(Fields.level(ᶠinterp(Geometry.contravariant3(ᶜuₕ)), half) / Geometry.contravariant3(one(ᶠw_surface)))
```
?

akshaysridhar:
> Do you get allocations if you use:
> 
> ```julia
> ᶠw_surface = level(ᶠw, half)
> @. ᶠw_surface = Geometry.Covariant3Vector(Fields.level(ᶠinterp(Geometry.contravariant3(ᶜuₕ)), half) / Geometry.contravariant3(one(ᶠw_surface)))
> ```
> 
> ?

I did try this as the first step yesterday and this created similar issues; I'll attempt this as well.

akshaysridhar:
> Do you get allocations if you use:
> 
> ```julia
> ᶠw_surface = level(ᶠw, half)
> @. ᶠw_surface = Geometry.Covariant3Vector(Fields.level(ᶠinterp(Geometry.contravariant3(ᶜuₕ)), half) / Geometry.contravariant3(one(ᶠw_surface)))
> ```
> 
> ?

There were issues with the `Geometry.contravariant3` function - type inconsistencies which prevent a division operation

src/precomputed_quantities.jl
```diff
@@ -16,9 +16,11 @@ function precomputed_quantities!(Y, p, t, colidx)
     ᶜuₕ = Y.c.uₕ
     ᶠw = Y.f.w
     (; ᶜK, ᶠu_tilde, ᶜu_bar, ᶠu³, ᶜts, ᶜp, params) = p
-    (; ᶠwinterp, ᶜinterp) = p.operators
+    (; ᶠinterp, ᶠwinterp, ᶜinterp) = p.operators
     ᶜJ = Fields.local_geometry_field(axes(Y.c.ρ)).J
     C123 = Geometry.Covariant123Vector
+    enforce_boundary = Operators.SetBoundaryOperator(bottom = Operators.SetValue(Geometry.Covariant3Vector.(Fields.level(ᶠinterp.(Geometry.Contravariant3Vector.(-1 .* ᶜuₕ[colidx])), half).components.data.:1 ./ Geometry.Contravariant3Vector.(one.(Fields.level(ᶠw[colidx], half))).components.data.:1)))
```
simonbyrne, line 22: Hmm, so the problem is that we're copying data unnecessarily.

PR #1384, opened 2023-02-22
Rico initial condition

trontrytel:
This PR computes the q_tot at surface for Rico assuming saturated surface conditions for the given surface temperature and pressure.

It doesn't solve our time evolution plots for Rico. We are still developing the convection too late. The reason is that a couple PRs ago we fixed the surface temperature from the wrong T=301K to T=299.8. See [this PR](https://github.com/CliMA/ClimaAtmos.jl/pull/1330).

I think the difference could be calibrated away. So I'm voting to keep the case as it is for now.


tc_driver/Cases.jl
```diff
@@ -453,17 +453,17 @@ function surface_params(case::Rico, thermo_params, moisture_model)
...
     cq = zc_surf -> cq0 * grid_adjust(zc_surf) # TODO: not yet used..
     psurface::FT = 1.0154e5
-    Tsurface::FT = 299.8
-
-    # For Rico we provide values of transfer coefficients
+    Tsurface::FT = 299.8 # 301.1
+    # Saturated surface condtions for a given surface temperature and pressure
+    p_sat_surface::FT = TD.saturation_vapor_pressure(thermo_params, Tsurface, TD.Liquid())
+    ϵ_v::FT = TD.Parameters.R_d(thermo_params) / TD.Parameters.R_v(thermo_params)
+    qsurface::FT = ϵ_v * p_sat_surface / (psurface - p_sat_surface * (1 - ϵ_v))
```
szy21, line 460: I thought we would have a function for this in Thermodynamics, but apparently we don't ;)

szy21 (approved):

tc_driver/Cases.jl
trontrytel: Haha - Yes, Thermodynamics.jl is mostly build around the `thermo_state`. Here we don't have it yet, since we don't have `q_tot`. At least I'm using saturation vapour pressure and the constants from TD.jl.

Issue #1389, opened 2023-02-22
Separate cache from diagnostics


trontrytel:
(i.e. don't rely on diagnostic variables being defined and/or updated in `update_aux`).

This is a spinoff from https://github.com/CliMA/ClimaAtmos.jl/pull/1376. We are trying to see if https://github.com/CliMA/ClimaAtmos.jl/issues/1388 would be enough of a temporary fix there. IF yes, we might come back to it later, while refactoring `update_aux`.

trontrytel:
No longer relevant. We moved to EDMFX

PR #1391, opened 2023-02-23
Call update_aux at the end of init_tc!

### LINKED ISSUE #1388: Don't allocate precipitation tendecies in 0-moment EDMF
This is a spinoff from https://github.com/CliMA/ClimaAtmos.jl/pull/1376 and is preventing us from correctly outputing diagnostics there.

dennisYatunin:
## Purpose 

This PR adds a call to `update_aux` at the end of `init_tc!`, in order to populate `Field`s that are required for diagnostics. This is required for #1376, since that PR sets the `aux` state to `NaN` in order to ensure consistency. It is a step toward closing #1389.

## Content
- Removes the modification of the rain and snow tendencies from `update_aux!`. Now, the rain and snow tendencies will be stored in the `aux` state, along with the tendencies of energy and total moisture due to precipitation.
- Changes the initial conditions so that the rain and snow prognostic variables are only allocated for 1-moment microphysics, which closes #1388.
- *Note*: The rain and snow tendencies are still being computed for models with no precipitation or with 0-moment microphysics; it's just that now they will be stored in the `aux` state instead of being used to modify the net tendency of the prognostic state. In the future, we should eliminate these unused computations altogether.

----


trontrytel:
The changes in the code look good to me. The EDMF SCM cases also look acceptable. 

However it looks like out spherical case is now breaking. I'm looking at the command line arguments for it. It seems like it is trying to run a dry model with 0-moment precipitation. That doesn't look right. Maybe we could switch to using no precipitation in that case for now?

dennisYatunin:
The spherical case was breaking because it was being initialized with infinite updraft moisture at the surface, which was not previously detected because the auxiliary state was not being properly initialized. Since it's called "aquaplanet (ρe_tot) equilmoist with EDMF", I've replaced the incorrect `--moist dry` with `--moist equil`. This gets rid of the infinite moisture problem.

trontrytel (approved):

PR #1403, opened 2023-02-27
Don't gather data in export_scaling_file

### LINKED ISSUE #1400: Long run failures
e.g. https://buildkite.com/clima/climaatmos-longruns/builds/833#0186872d-610a-49a6-aa25-ecf006dd1b64

Cause appears to be
https://github.com/CliMA/ClimaAtmos.jl/blob/2379b55ff93d146e4a11848c4f510b0df3029bb9/examples/hybrid/driver.jl#L374-L380

which is defined here:
https://github.com/CliMA/ClimaAtmos.jl/blob/5cf6ec260056da95f420a1568a9d3f71f523322e/examples/hybrid/utilities.jl#L28

We appear to reduce everything to the root node to get `sol`: I don't think we need that any more?

simonbyrne:
Fix #1400

## Purpose 
This code is causing the long runs to fail, and doesn't appear to be used.

## Content

----

### DIFF
--- a/examples/hybrid/utilities.jl
+++ b/examples/hybrid/utilities.jl
@@ -30,7 +30,6 @@ function export_scaling_file(sol, output_dir, walltime, comms_ctx, nprocs)
     if ClimaComms.iamroot(comms_ctx)
         Y = sol.u[1]
         center_space = axes(Y.c)
-        face_space = axes(Y.f)
         horz_space = ClimaCore.Spaces.horizontal_space(center_space)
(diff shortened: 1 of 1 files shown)


szy21 (approved):

simonbyrne:
https://buildkite.com/clima/climaatmos-ci/builds/7587#0186940c-3903-4fcf-8735-cbea6b37f4bb/140-257

szy21:
fixing it now: #1404

PR #1407, opened 2023-02-28
Remove incorrect anelastic option

dennisYatunin:
## Purpose 

In the process of debugging #1327, we have found the the anelastic option is not implemented correctly. It has two issues:
- the pressure `p.ᶜp` is initially set in `init_tc!` (which #1327 removes), and it is used to set the thermodynamic state `p.ᶜts` at the start of each tendency evaluation, but the thermodynamic state is then in turn used to set the pressure
- more importantly, the thermodynamic state computed from the pressure implies a certain value for density, but this is not the same as the density `Y.c.ρ` that is used throughout ClimaAtmos (i.e., the prognostic variable that gets evolved by the time stepper)

The first issue can be easily fixed by adding some additional if-statements to the code, but the fixing the second issue would require updating all of ClimaAtmos to switch between using density as a prognostic variable or as a diagnostic variable. Instead of doing this, we will simply delete the two remaining anelastic TRMM cases in CI, along with all of the code for running the model in anelastic mode.

----


trontrytel (approved):

trontrytel:
LGTM! I'm sad to see the anelastic option go, but we can always turn to the standalone TC.jl code. And it's better to simplify the CA.jl code than to try to keep it as an option.

I think our EDMF SCM cases are not fully compressible right now. I'm reading through the code and will have more updates today/tomorrow.

dennisYatunin:
Yup, this should be the last blocker 🤞

PR #1409, opened 2023-03-01
Add arg parsing for a TOML file, unify with parsed_args

### LINKED ISSUE #1340: Add argument parsing for override TOML file
This will be passed in and used to override the default TOML in `create_parameters.jl`. 
Should be a simple change, since there is already full functionality for this within CLIMAParameters.

nefrathenrici:
## Purpose 

This PR unifies `parsed_args` and the EDMF model config, and adds support for an additional TOML file via CLI option.
 closes #1340.

## To-do

Clean up code, add testing for parsed_args.
Get feedback.

## Content

1. Adds CLI options for a TOML file.
2. Parses the optional TOML file and merges `parsed_args` into the resulting `toml_dict`. `parsed_args` overrides in case of clashes.
3. Unifies the EDMF `config_params` and `parsed_args` into a `config` Dict used to store model configurations. This is created from the `toml_dict`. `config` then overwrites the `parsed_args` Dict in driver. This is done for two reasons: to keep the current `parsed_args` functionality without introducing bugs, and to allow the override TOML file to set model configurations, not just parameters.
4. Adds a TOML file for running the SCM EDMF Bomex cases, modifies the pipeline to use this file.

----


.buildkite/pipeline.yml
```diff
@@ -325,189 +325,146 @@ steps:
...
-
       - label: ":balloon: SCM EDMF Bomex (Default: compressible, Newton's method)"
         command: >
           julia --color=yes --project=examples examples/hybrid/driver.jl
-          --config column --turbconv edmf --debugging_tc true --regression_test true
-          --FLOAT_TYPE Float64 --hyperdiff false --apply_limiter false
-          --turbconv_case Bomex --subsidence Bomex --edmf_coriolis Bomex --ls_adv Bomex
-          --z_elem 60 --z_stretch false --z_max 3e3
-          --t_end 6hours --dt_save_to_sol 5mins --dt_save_to_disk 5mins
-          --moist equil
```
trontrytel, line 457: Should the options that are moved to the `toml` files be deleted from the supported `cli_args`? Otherwise we are running into issues who overwrites whom.

.buildkite/toml/bomex.toml
```diff
@@ -0,0 +1,79 @@
+[config] 
+value = "column"
```
trontrytel, line 2: Could we add the toml files for the other SCM EDMF test cases as well? I would like to keep them all in the same format

trontrytel: Or alternatively, we could just  do it for one case (Bomex) and keep the others as they are for now. And even add a CI job that tests if the `toml` driven and `cli_args` driven simulations are the same

.buildkite/pipeline.yml
nefrathenrici: I think they should still be supported for now, since people probably rely on them for their workflow. The specified CLI arguments always take priority over the TOML file. I can add a note in the documentation to clarify this?

.buildkite/toml/bomex.toml
nefrathenrici: Will do.

.buildkite/pipeline.yml
trontrytel: Sounds good. Adding it to the docs would be great. Not sure where it would be most useful. Maybe in the `cli_args` docstring for the `--toml` option?

nefrathenrici: I'll at least add to the docstring in `cli_options.jl`. I can look around if there is another place in the actual docs as well.

examples/hybrid/parameter_set.jl
```diff
@@ -1,28 +0,0 @@
-import ClimaAtmos
-import ClimaAtmos as CA
-
-include(joinpath(pkgdir(ClimaAtmos), "parameters", "create_parameters.jl"))
-
-function create_parameter_set(::Type{FT}, parsed_args) where {FT}
```
trontrytel, line 6: Not sure about moving this function. I thought it should be in the `examples` folder because it is related to specific cases and not the model itself. Is it ok to put it in parameters folder instead? But maybe it is a better place for it, since it's still not in `src`

trontrytel (approved):
LGTM! I left some small comments and also poked @dennisYatunin to have a quick SE look at it.

examples/hybrid/parameter_set.jl
nefrathenrici: I'm not sure either - I can leave it as-is but I know that @szy21 wanted to move it as part of the overall refactor.

trontrytel: Sounds good to me. Lets do whatever works best for the refactor

szy21: Yes, I plan to move this file to `src`, if it still exists after this PR.

trontrytel: @szy21 are we supposed to be removing things from `examples/hybrid/`? Maybe it is better to put case-specific overrides in `parameters` instead?

.buildkite/toml/bomex.toml
trontrytel: Sounds good to me. We can have Bomex and toml driven Bomex

examples/hybrid/parameter_set.jl
trontrytel: This function overwrites the defaults for different cases. Not sure if it should live in `src`?

szy21: Yep, the plan is to move everything we can out of examples/hybrid/, but we can think about where to put this file. For this PR I don't have a preference - up to you.

.buildkite/pipeline.yml
```diff
@@ -497,6 +497,18 @@ steps:
         agents:
           slurm_mem: 20GB
 
+      - label: ":balloon: SCM EDMF Bomex (Default: compressible, Newton's method, toml-driven)"
+        command: >
+          julia --color=yes --project=examples examples/hybrid/driver.jl
+          --regression_test true
+          --FLOAT_TYPE Float64
+          --toml ./.buildkite/toml/bomex.toml
+          --job_id compressible_edmf_bomex
```
trontrytel, line 506:
```suggestion
          --job_id toml_compressible_edmf_bomex
```

trontrytel: And also: `--quicklook_reference_job_id compressible_edmf_bomex`

PR #1412, opened 2023-03-02
Removing quadratures, (co)variances and third moment diagnostics

trontrytel:
Following our meeting yesterday, here is my first go at slimming down `update_aux`. I'm deleting here the 2nd order terms other than TKE. I will delete TKE equations in the next PR. 

This is a temporary delete to allow us to faster disentangle and delete `update_aux` and change the code to the new equations.

I think it also simplifies some of the command line arg parsing.


trontrytel:
Closes https://github.com/CliMA/ClimaAtmos.jl/issues/1418

PR #1447, opened 2023-03-10
Unify surface parametrization of dycore and EDMF

dennisYatunin:
## Summary
This PR creates a new submodule, SurfaceSchemes, which has three components
1. `surface_state.jl` contains the new surface state structs used to store the surface parameterization, which match the SurfaceFluxes interface better than the previous TC surface representation.
2. `surface_schemes.jl` contains the initial surface conditions for all of the EDMF cases, defined using the new surface state structs. 
3. `surface_conditions.jl` contains the methods used to update these surface conditions at each timestep. Surface values are now updated in `set_precomputed_quantities!` instead of  in `additional_tendency`
`p` now contains `sfc_conditions`, a field of the values at the surface, rather than individual variable fields
The types of these have also been changed to proper C/CTVectors
#### New flow:
1. `sfc_state` is created in `default_cache` from one of the initial surface conditions in `surface_schemes.jl` `sfc_conditions` is created in `default_cache > precomputed_quantities` to store the values at the surface
2. `update_surface_conditions!` is called from `set_precomputed_quantities!`, broadcasting `surface_conditions` across the `sfc_conditions` field (NOT `SF.surface_conditions`)
3. `surface_conditions` handles the logic to construct the surface thermo state, then computes and stores the updated surface conditions

#### Large file changes:
`EDMF_functions.jl`: Replaced some calls to helper functions with code. These helper functions were only called once and didn't make sense to keep around. Changed to use the new sfc_conditions and CT3Vectors
`update_aux.jl`: Removed helper functions, changed to use the new sfc_conditions
`TC/types.jl`: Removed all surface structs and old helper functions. Removed some column getter functions. Removed surf_params from State
`sgs_flux_tendencies.jl`: Update function calls to match new APIs
`vector_diffusion_boundary_layer.jl`: Removed the vdbl cache. Changed calculations to use the new sfc_conditions
`utils/surface.jl`: Deleted
`src/utills/types.jl`: Removed the Coupling and SurfaceScheme types, changed AtmosModel to have a surface_updater field instead of a coupling field.

## Content

----
- [ ] I have read and checked the items on the review checklist.

PR #1453, opened 2023-03-11
Align vertical_diffusion_boundary_layer_tendency with design docs

dennisYatunin:
## Purpose 

The design docs specify that diffusion of momentum should have the form `@. -ᶜdivᵥ(ᶠρτ)/ᶜρ`. However, the `vertical_diffusion_boundary_layer_tendency!` currently has the form `@. ᶜdivᵥ(-ᶠτ)`, where `ᶠτ` at the bottom face boundary is given by `ᶠρτ[half]` (taken from SurfaceFluxes) divided by `ᶜρ[1]` (taken from the bottom cell center). This PR changes the momentum tendency so that it has the correct form. It also adds negations to the other diffusive tendencies, so that we do not need to negate the boundary conditions given by SurfaceFluxes. This is a prequel to #1447. 

## Content

- Change the tendency of momentum due to diffusion from `@. Yₜ.c.uₕ += ᶜdivᵥ(ᶠK_E * ᶠgradᵥ(Y.c.uₕ))` to `@. Yₜ.c.uₕ -= ᶜdivᵥ(-ᶠK_E * ᶠinterp(Y.c.ρ) * ᶠgradᵥ(Y.c.uₕ))) / Y.c.ρ`. Change `ᶜdivᵥ` so that it sets the value at the bottom boundary to `WVector(1) ⊗ UVVector(ρτxz, ρτyz)`, instead of `-WVector(1) ⊗ UVVector(ρτxz / ρ_bottom_center, ρτyz / ρ_bottom_center)`.
- Change the tendencies of energy and moisture due to diffusion from `@. tendency += ᶜdivᵥ(-ᶠflux)` to `@. tendency -= ᶜdivᵥ(ᶠflux)`, and negate their boundary conditions (i.e., use the unnegated boundary values from SurfaceFluxes).
- Remove the "negated boundary condition" cached fields `dif_flux_uₕ_bc`, `dif_flux_energy_bc`, and `dif_flux_ρq_tot_bc`.
- Rename `dif_flux_uₕ` to `dif_flux_ρuₕ`, and rename `dif_flux_energy` to `dif_flux_ρe_tot`.

----


szy21:
I think the change in momentum diffusion is ok, but it would be good if @tapios can make sure.

You also need to change the corresponding surface flux variable names in `post_processing/remap/remap_helpers.jl` (and make sure the sign is correct).

src/tendencies/vertical_diffusion_boundary_layer.jl
```diff
@@ -323,25 +310,25 @@ function get_surface_fluxes!(Y, p, t, colidx, ::VerticalDiffusion)
...
-            Geometry.Contravariant3Vector(surface_normal[colidx]) ⊗
-            Geometry.Covariant12Vector(
-                Geometry.UVVector(ρτxz / ρ_1, ρτyz / ρ_1),
-            )
-        @. dif_flux_uₕ_bc[colidx] = -dif_flux_uₕ[colidx]
+        if isnothing(p.surface_scheme)
+            @. dif_flux_ρuₕ[colidx] *= 0
+        else
+            @. dif_flux_ρuₕ[colidx] =
+                Geometry.WVector(FT(1)) ⊗ Geometry.UVVector(
```
szy21, line 317: Why could we remove the `Contravarian3Vector` and `Covariant12Vector`?

tapios (commented):
The change in momentum flux looks good to me. However, I did not try to understand all details (e.g., the signs) here, so I cannot comment on the overall correctness. 

I suggest renaming `dif_flux_ρe_tot` to `dif_flux_ρh_tot` consistently (and similarly for other diffusive fluxes of total enthalpy, e.g., hyperdiffusion), to make clear that enthalpy, not energy, is being diffused and appears in fluxes. 

I also think the names `dif_flux_ρuₕ` etc., with `ρ` in it, are potentially misleading, since what is being diffused is `uₕ`, `q_tot` etc., not the quantity per unit volume with density in it. I think what you mean is something like `ρ_dif_flux_uₕ`

src/tendencies/vertical_diffusion_boundary_layer.jl
dennisYatunin: For the same reason we don't need any conversion [here](https://github.com/CliMA/ClimaAtmos.jl/blob/01ae1f01f9435636b0fb42aa6d8841683585fa86/src/tendencies/vertical_diffusion_boundary_layer.jl#L410); the finite difference operators automatically convert axis tensors to the required types.

dennisYatunin:
Updates: I've renamed variables as Tapio suggested (`ρ_dif_flux_uₕ`, `ρ_dif_flux_h_tot`, and `ρ_dif_flux_q_tot`), and I've moved the division of `ρτ` by `ρ` to the diagnostics callback, so that the value of `sfc_flux_momentum` remains unchanged. I also had to increment the ref counter because the `sphere_aquaplanet_rhoe_equilmoist_allsky_gw` example's results slightly changed (this is the only example that actually computes a surface flux of momentum, so it was the only example affected by this PR). I also removed the change of `Covariant3Vector` to `WVector` to make it clear that the changed results are only due to the tendency alteration; I'll make the storage type change in #1447.

PR #1454, opened 2023-03-13
Move callbacks to source code

szy21:
## Purpose 

Note: `get_callbacks` and `turb_conv_affect_filter!` are still in examples (in the new file `get_callbacks.jl`), because they use `TurbulenceConvectionUtils` in examples.

## Content

----


src/callbacks.jl
```diff
@@ -131,7 +130,7 @@ function dss_callback(integrator)
     Y = integrator.u
     ghost_buffer = integrator.p.ghost_buffer
     if !ghost_buffer.skip_dss
-        @nvtx "dss callback" color = colorant"yellow" begin
+        NVTX.@range "dss callback" color = colorant"yellow" begin
```
jiahe23, line 133: Why is this change needed?

jiahe23:
LGTM

jiahe23 (approved):

src/callbacks.jl
szy21: The @nvtx macro is defined in example (https://github.com/CliMA/ClimaAtmos.jl/blob/main/examples/hybrid/nvtx.jl) and cannot be used in src. But it is essentially the same as NVTX.@range.

szy21:
bors try

PR #1461, opened 2023-03-16
move types to src

szy21:
## Content

----


src/model_getters.jl
```diff
@@ -1,4 +1,4 @@
-function moisture_model(parsed_args)
+function get_moisture_model(parsed_args)
```
jiahe23, line 1: I like this new name!

jiahe23:
can we give `example_types` a better name? Or can we maybe move functions in this file to other places based on their functionalities, e.g., `get_atmos` to `model_getters.jl` and `get_space` to `common_spaces.jl`

szy21:
> can we give `example_types` a better name? Or can we maybe move functions in this file to other places based on their functionalities, e.g., `get_atmos` to `model_getters.jl` and `get_space` to `common_spaces.jl`

As discussed offline, this will be done as a separate PR.

PR #1470, opened 2023-03-17
Add Bomex box experiment

trontrytel:
This PR adds a bomex box experiment setup (without the sub-grid scale model). 

I think the initial condition and forcings are applied correctly. Not sure about the surface parameters...

### DIFF
--- a/post_processing/post_processing_funcs.jl
+++ b/post_processing/post_processing_funcs.jl
@@ -130,6 +130,16 @@ function postprocessing_box(sol, output_dir)
         clim = (-0.1, 0.1),
     )
     Plots.png(p, joinpath(output_dir, "w.png"))
+
+    if :ρq_tot in propertynames(Y.c)
+        qt = Y.c.ρq_tot ./ Y.c.ρ
+        pq = Plots.plot(
+            qt,
+            slice = (:, FT(parsed_args["y_max"] / 2), :),
+            clim = (-0.02, 0.020),
+        )
+        Plots.png(pq, joinpath(output_dir, "qt.png"))
+    end
 end
 
 function postprocessing(sol, output_dir, fps)
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -97,31 +97,49 @@ steps:
     steps:
 
       - label: ":computer: single column hydrostatic balance float64"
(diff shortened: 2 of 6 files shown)


post_processing/post_processing_funcs.jl
```diff
@@ -130,6 +130,13 @@ function postprocessing_box(sol, output_dir)
         clim = (-0.1, 0.1),
     )
     Plots.png(p, joinpath(output_dir, "w.png"))
+    qt = Y.c.ρq_tot ./ Y.c.ρ
```
szy21, line 133: This will cause a problem for the dry case. You can add a check for whether `ρq_tot` is in `Y.c`.

szy21 (approved):

post_processing/post_processing_funcs.jl
trontrytel: Right! thanks

trontrytel:
Seems like the initial condition for `rhoq_tot` is not used - we have `q_tot = 0` at the end. I'll look into this.

szy21:
Oh, I think it may be because [here](https://github.com/CliMA/ClimaAtmos.jl/blob/e8e5570cf4ddad394f55681024400ef42470a06b/src/examples_types.jl#L229) if the config is box then the dry initial condition will be used. However, even if we start dry the large-scale tendency in q_tot should result in non-zero q_tot in the end?

trontrytel:
Right. Something else is suspicious. I'll see if I can disentangle the initial condition from the case configuration. And then take a look at the forcings again

src/examples_types.jl
```diff
@@ -220,22 +220,19 @@ function get_state_restart(comms_ctx)
 end
 
 function get_initial_condition(parsed_args)
-    if is_baro_wave(parsed_args)
-        if parsed_args["moist"] == "dry"
-            return ICs.DryBaroclinicWave(parsed_args["perturb_initstate"])
+    if isnothing(parsed_args["turbconv_case"])
+        if parsed_args["initial_condition"] in ["DryBaroWave", "MoistBaroWave", "DecayingProfile"]
```
szy21, line 224:
```suggestion
        if parsed_args["initial_condition"] in ["DryBaroclinicWave", "MoistBaroclinicWave", "DecayingProfile"]
```

trontrytel: Must have forgotten to push the fix to this! Thank you.

PR #1475, opened 2023-03-20
remove ARM updrafts from regression

trontrytel:
Right now we can't have zero reference state in regression tests because we normalize the error norm by the reference state. 

I'm deleting here the ARM updraft properties from regression tests because after the recent changes they are now zero.


trontrytel:
@szy21 it's running! 🤞

PR #1479, opened 2023-03-21
Delete entr/detr from cache

trontrytel:
This PR:
- removes `ε_nondim`, `δ_nondim`, `frac_turb_entr`, `entr_turb_dyn`, `detr_turb_dyn`, `detr_loss`, `interdomain` from cached variables
- replaces the k-loop based ` compute_entr_detr!` with point wise, dispatch friendly `pi_groups_detrainment!`
- moves the `compute_covariance_entr` from `update aux` inside the `compute_explicit_en_tendencies!` in `EDMF_functions`
- deletes the unused `update_diagnostic_covariances!` and `compute_covariance_interdomain_src`

The entrainment tendencies for the updrafts are written now in a way that makes them easy to port over to the EDMFX code when needed. The only reason why we still need to cache the detrainment coefficient is because it is used in the buoyancy exchange term that is then used in the mixing length computations. I'll think some more on how to disentangle this. But I think this is good enough to be merged as is.


src/TurbulenceConvection/update_aux.jl
```diff
@@ -301,34 +301,38 @@ function update_aux!(
...
         @. b_exch +=
-            a_up * Ic(wcomponent(CCG.WVector(w_up))) * δ_dyn / a_en *
-            (1 / 2 * (Ic(wcomponent(CCG.WVector(w_up - w_en))))^2 - tke_en) -
-            a_up *
+            aux_up[i].area *
             Ic(wcomponent(CCG.WVector(w_up))) *
-            (Ic(wcomponent(CCG.WVector(w_up - w_en)))) *
-            ε_turb *
-            Ic(wcomponent(CCG.WVector(w_en))) / a_en
+            aux_up[i].detr_sc / aux_en.area *
```
szy21, line 334:
```suggestion
            aux_up[i].detr / aux_en.area *
```
Shoud this be detr?

szy21 (approved):

szy21:
The threaded allocation job is a bit random. Usually we just try again or increase the limit a bit. Maybe we should just add a buffer?

src/TurbulenceConvection/update_aux.jl
szy21, line 334:
```suggestion
            aux_up[i].detr / aux_en.area *
```
😛

trontrytel: 🥲 I have to figure out what I do wrong with the recent rebases

PR #1482, opened 2023-03-22
Add EDMFX model without tendencies and add new post-processing

nefrathenrici:
## Purpose 

This PR adds the "EDMFX" model to ClimaAtmos, but without any tendencies. This includes adding prognostic variables, precomputed quantities, and diagnostics, as described in  ClimaAtmos's new [documentation for EDMF](https://clima.github.io/ClimaAtmos.jl/dev/edmf_equations/), along with adding the post-processing that we will need in order to debug this new model's tendencies.

## Content

- Add the total draft area to the initial condition's `EDMFState`, and set it to 0 by default. Clean up `getproperty` for `EDMFState` and `NoTurbconvState`, since it was not updated when the second moments were removed.
- Define the `EDMFX` struct and the `n_mass_flux_subdomains` function (which was called `n_updrafts` in the original EDMF model) in `types.jl`. Also define the `times_a` function in `utilities.jl`, which takes in a `NamedTuple` of variables and returns a new `NamedTuple` with every variable of the form `ρ` or `ρχ` transformed to `ρa` or `ρaχ`, respectively. Construct the EDMFX model in `model_getters.jl`.
- Separate the initial condition's `atmos_center_variables` into `grid_scale_center_variables` and `turbconv_center_variables`. Leave the original EDMF model's `turbconv_center_variables` unchanged, but have the EDMFX model's `turbconv_center_variables` return a copy of `grid_scale_center_variables` with every variable multiplied by `a` (the total draft area divided by the number of mass flux subdomains). Also add a new method for `turbconv_face_variables` for the EDMFX model that sets all of the draft vertical velocities to 0. Modify `is_turbconv_var` to recognize the new groups of EDMFX variables.
- Add `ᶜρa⁰`, `ᶠw⁰`, `ᶜu⁰`, `ᶠu³⁰`, `ᶜK⁰`, `ᶜts⁰`, `ᶜρa`, `ᶜuʲs`, `ᶠu³ʲs`, `ᶜKʲs`, `ᶜtsʲs`, and `ᶜρʲs` to `precomputed_quantities` for the EDMFX model, and set these values in `set_precomputed_quantities!`. Make `set_precomputed_quantities` add the SGS kinetic energy to the grid-scale `ᶜK`, and make `set_boundary_velocity!` set the draft vertical velocities equal to the grid-scale vertical velocity at the surface. Compute all specific SGS quantities from the prognostic state using the new `divide_by_ρa` function, which regularizes divisions by returning quantities closer to the grid-scale values when `a` is small.
- Define a `diagnostic_edmfx_quantities` function, which allocates and sets `ᶜsgs⁰`, `ᶜsgsᴰ`, `ᶜuᴰ`, and `ᶜtsᴰ` in a manner that is consistent with `set_precomputed_quantities!`. Call this function from `save_to_disk_func`, and use it to define the actual EDMFX diagnostics that are saved to hdf5 files.
- Make the implicit solver set the Jacobian blocks for the EDMFX model to 0 by modifying the `ldiv!` method for `SchurComplementW`. More generally, this modification causes the implicit solver to automatically zero out all Jacobian blocks that are not set in `_ldiv_serial!`.
- Ensure that none of the original EDMF code is run for the EDMFX model. This includes removing the `::Nothing` type restrictions from `turbconv_cache`, `implicit_sgs_flux_tendency!`, `explicit_sgs_flux_tendency!`, and `compute_precipitation_cache!`. It also includes replacing `!isnothing(atmos.turbconv_model)` with `p.atmos.turbconv_model isa TC.EDMFModel` in `get_callbacks`, replacing `any(is_turbconv_var, propertynames(Y.c))` with `:turbconv in propertynames(Y.c)` in `SchurComplementW`, and replacing `any(is_turbconv_var, propertynames(xc))` with `:turbconv in propertynames(xc)` in `_ldiv_serial!`.
- Rename the CLI option `quicklook_reference_job_id` to `reference_job_id`, since it will now be used by the new post-processing as well as the original "quicklook profiles".
- Define the new post processing in `contours_and_profiles.jl`, which is the bulk of this PR. This post-processing generates contour and profile plots using Makie, with the column data extracted from 0°N 0°E, 30°N 0°E, 60°N 0°E, and 90°N 0°E on a sphere or from the first column in a box (it does not currently support extracting data from a plane but that can be easily changed). In the future, we may also want to add averaging over multiple columns. When running on Buildkite, the reference data is read in from the latest commit on the main branch that contains hdf5 files (either zipped or unzipped) for the specified `reference_job_id` (which is just the `job_id` by default). The reference files do not need to have the same timestamps or contain data on the same spaces as the simulation's files. The reference files also do not need to contain all of the same diagnostic variables as the simulation; any reference diagnostics that are missing do not get plotted. Values that should never be negative (e.g., specific quantities or temperatures) but are still negative get highlighted in red on the contour plots.
- Call `contours_and_profiles` from the driver when the `post_process` CLI option is `true` and the EDMFX model is being used. Ensure that the other post-processing is not run for the EDMFX model by adding checks for `!is_edmfx` to the relevant `if`-blocks. After calling `contours_and_profiles`, also call `zip_and_cleanup_output` in order to make the various plot files easier to find among the Buildkite artifacts. Move the `include` call for `define_tc_quicklook_profiles.jl` to an earlier line so that `zip_and_cleanup_output` can be used without needing to be redefined.
- Add `env_q_vap` and `draft_q_vap` to the old EDMF model's diagnostics, so that they can be compared to the values obtained with the new model.
- Add a `MoistBaroclinicWaveWithEDMF` initial condition, which is exactly like the `MoistBaroclinicWave` initial condition, except that it also has an initial draft area of 0.5 and an initial TKE of 1.
- Add three new jobs to CI that test the EDMFX model. The first is a copy of `sphere_baroclinic_wave_rhoe_equilmoist` with the initial condition changed to `MoistBaroclinicWaveWithEDMF` and `t_end` reduced to 4 days. The second is a copy of `edmf_bomex` that uses the default `sphere` configuration, `Float32` type, and active hyperdiffusion, with `dt` increased to 60 seconds and `t_end` reduced to 1 hour. Both jobs use the original versions as references for plotting. The third job is a performance test for the first job, which checks that the amount of memory allocated per timestep is reasonable (it is currently around 4 kB, whereas the amount allocated per timestep for the old EDMF model is around 8 GB).

----

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -355,9 +355,26 @@ steps:
           slurm_mem: 20GB
 
       - label: ":dart: Dycore Consistency test"
-        command: "julia --color=yes --project=examples examples/hybrid/driver.jl --config column --FLOAT_TYPE Float64 --hyperdiff false --moist equil --subsidence Bomex --edmf_coriolis Bomex --ls_adv Bomex --turbconv edmf --turbconv_case Bomex --dt_save_to_sol 5mins --z_elem 60 --z_stretch false --z_max 3e3 --use_krylov_method true --job_id dycore_consistency --quicklook_reference_job_id edmf_bomex --dt 40secs --t_end 6hours --apply_limiter false --test_dycore_consistency true --debugging_tc true --dt_save_to_disk 5mins"
+        command: "julia --color=yes --project=examples examples/hybrid/driver.jl --config column --FLOAT_TYPE Float64 --hyperdiff false --moist equil --subsidence Bomex --edmf_coriolis Bomex --ls_adv Bomex --turbconv edmf --turbconv_case Bomex --dt_save_to_sol 5mins --z_elem 60 --z_stretch false --z_max 3e3 --use_krylov_method true --job_id dycore_consistency --reference_job_id edmf_bomex --dt 40secs --t_end 6hours --apply_limiter false --test_dycore_consistency true --debugging_tc true --dt_save_to_disk 5mins"
         artifact_paths: "dycore_consistency/*"
         soft_fail: true
+        agents:
+          slurm_mem: 20GB
+
+  - group: "EDMFX"
+    steps:
+
+      - label: ":genie: no lim ARS baroclinic wave (ρe) equilmoist with EDMFX"
+        command:
+          - julia --color=yes --project=examples examples/hybrid/driver.jl --job_id sphere_baroclinic_wave_rhoe_equilmoist_edmfx --initial_condition MoistBaroclinicWaveWithEDMF --moist equil --precip_model 0M --turbconv edmfx --apply_limiter false --dt 450secs --t_end 4days --dt_save_to_disk 2days --reference_job_id sphere_baroclinic_wave_rhoe_equilmoist
+        artifact_paths: "sphere_baroclinic_wave_rhoe_equilmoist_edmfx/*"
+        agents:
+          slurm_mem: 20GB
+
+      - label: ":genie: EDMFX Bomex on a sphere"
+        command:
+          - julia --color=yes --project=examples examples/hybrid/driver.jl --job_id edmfx_bomex_sphere --initial_condition Bomex --moist equil --turbconv edmfx --subsidence Bomex --edmf_coriolis Bomex --ls_adv Bomex --apply_limiter false --z_elem 60 --z_stretch false --z_max 3e3 --dt 60secs --t_end 1hours --dt_save_to_disk 5mins --reference_job_id edmf_bomex
+        artifact_paths: "edmfx_bomex_sphere/*"
         agents:
           slurm_mem: 20GB
 
@@ -409,7 +426,7 @@ steps:
           --imex_edmf_turbconv true
           --imex_edmf_gm true
           --use_krylov_method true
-          --quicklook_reference_job_id edmf_nieuwstadt
+          --reference_job_id edmf_nieuwstadt
           --job_id edmf_nieuwstadt_jfnk_imex
           --dt 50secs
         artifact_paths: "edmf_nieuwstadt_jfnk_imex/*"
@@ -444,7 +461,7 @@ steps:
           --imex_edmf_turbconv true
           --imex_edmf_gm true
           --use_krylov_method true
-          --quicklook_reference_job_id edmf_soares
+          --reference_job_id edmf_soares
           --job_id edmf_soares_jfnk_imex
           --dt 50secs
         artifact_paths: "edmf_soares_jfnk_imex/*"
@@ -477,7 +494,7 @@ steps:
           --t_end 6hours --dt_save_to_sol 5mins --dt_save_to_disk 5mins
           --moist equil
           --use_krylov_method true
-          --quicklook_reference_job_id edmf_bomex
+          --reference_job_id edmf_bomex
           --job_id edmf_bomex_jfnk
           --dt 15secs
         artifact_paths: "edmf_bomex_jfnk/*"
@@ -496,7 +513,7 @@ steps:
           --imex_edmf_turbconv true
           --imex_edmf_gm true
           --use_krylov_method true
-          --quicklook_reference_job_id edmf_bomex
+          --reference_job_id edmf_bomex
           --job_id edmf_bomex_jfnk_imex
           --dt 30secs
         artifact_paths: "edmf_bomex_jfnk_imex/*"
@@ -510,7 +527,7 @@ steps:
           --FLOAT_TYPE Float64
           --toml ./.buildkite/toml/bomex.toml
           --job_id toml_edmf_bomex
-          --quicklook_reference_job_id edmf_bomex
+          --reference_job_id edmf_bomex
           --dt 20secs
         artifact_paths: "toml_edmf_bomex/*"
         agents:
@@ -623,6 +640,12 @@ steps:
       - label: ":fire: Flame graph: perf target (with tracers)"
         command: "julia --color=yes --project=perf perf/flame.jl --job_id flame_perf_target_tracers --perf_mode PerfExperimental"
         artifact_paths: "flame_perf_target_tracers/*"
+        agents:
+          slurm_mem: 20GB
+
+      - label: ":fire: Flame graph: perf target (edmfx)"
+        command: "julia --color=yes --project=perf perf/flame.jl --job_id flame_perf_target_edmfx --target_job sphere_baroclinic_wave_rhoe_equilmoist_edmfx"
+        artifact_paths: "flame_perf_target_edmfx/*"
         agents:
           slurm_mem: 20GB
 
--- a/src/InitialConditions/atmos_state.jl
+++ b/src/InitialConditions/atmos_state.jl
@@ -33,25 +33,32 @@ In order to extend this function to new sub-models, add more functions of the
 form `variables(ls, model, args...)`, where `args` can be used for dispatch or
 for avoiding duplicate computations.
 """
-atmos_center_variables(ls, atmos_model::AtmosModel) = (;
-    ρ = ls.ρ,
-    uₕ = Geometry.project(Geometry.Covariant12Axis(), ls.velocity, ls.geometry),
-    energy_variables(ls, atmos_model.energy_form)...,
-    moisture_variables(ls, atmos_model.moisture_model)...,
-    precip_variables(ls, atmos_model.precip_model, atmos_model.perf_mode)...,
-    turbconv_center_variables(ls, atmos_model.turbconv_model)...,
-)
+function atmos_center_variables(ls, atmos_model)
+    gs_vars = grid_scale_center_variables(ls, atmos_model)
+    sgs_vars =
+        turbconv_center_variables(ls, atmos_model.turbconv_model, gs_vars)
+    return (; gs_vars..., sgs_vars...)
+end
 
 """
     atmos_face_variables(ls, atmos_model)
 
 Like `atmos_center_variables`, but for cell faces.
 """
-atmos_face_variables(ls, atmos_model::AtmosModel) = (;
+atmos_face_variables(ls, atmos_model) = (;
     w = Geometry.project(Geometry.Covariant3Axis(), ls.velocity, ls.geometry),
     turbconv_face_variables(ls, atmos_model.turbconv_model)...,
 )
 
+grid_scale_center_variables(ls, atmos_model) = (;
+    ρ = ls.ρ,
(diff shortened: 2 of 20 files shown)


src/InitialConditions/atmos_state.jl
```diff
@@ -115,3 +128,25 @@ turbconv_face_variables(ls, turbconv_model::TC.EDMFModel) = (;
...
+
+"""
+    times_a(variables, a)
+
+Given a NamedTuple that specifies variables of the form `ρ` or `ρχ`, generates a
+NamedTuple containing variables of the form `ρa` or `ρaχ`, respectively,
+multiplying the value of each variable by `a`. Any variables whose names do not
+begin with `ρ` (i.e., velocities) are omitted from the result.
+"""
+@generated function times_a(variables::NamedTuple{names}, a) where {names}
```
tapios, line 146: Is this GPU friendly?

src/precomputed_quantities.jl
```diff
@@ -3,49 +3,340 @@
...
+                Φ,
+                a_min,
+                thermo_params,
+                energy_form,
+                moisture_model,
+            )
+        end
+
+    w₀(ρaⱼs, wⱼs, ρ, w) =
+        divide_by_ρa(ρ * w - sum(ρaⱼs .* wⱼs), ρ - sum(ρaⱼs), ρ * w, ρ, a_min)
```
tapios: We'll probably need `ρ0 = ρ - sum(ρaⱼs)` in other places too (e.g., in diffusion terms in GS equations), so it may make sense to cache it before this computation.

src/precomputed_quantities.jl
```diff
@@ -3,49 +3,340 @@
...
+        @. ᶜtsⱼs = tsⱼs(Y.c, ᶜKⱼs, ᶜp, ᶜΦ)
+
+        Fields.bycolumn(axes(Y.c)) do colidx
+            @. ᶠw₀[colidx] = w₀(
+                ᶠinterp(map(sgsⱼ -> sgsⱼ.ρa, Y.c.sgsⱼs[colidx])),
+                map(sgsⱼ -> sgsⱼ.w, Y.f.sgsⱼs[colidx]),
+                ᶠinterp(Y.c.ρ[colidx]),
+                Y.f.w[colidx],
+            )
+        end # TODO: Use an upwind-biased product to compute ᶠρa₀ and ᶠρaⱼs.
```
tapios:
```suggestion
        end 
```

.buildkite/pipeline.yml
```diff
@@ -239,6 +239,20 @@ steps:
         agents:
           slurm_mem: 20GB
 
+      - label: ":computer: no lim ARS baroclinic wave (ρe) equilmoist with EDMFX"
```
trontrytel, line 242: Could we have a different symbol for EDMFX jobs so that they are easier to find in the CI list?

.buildkite/pipeline.yml
```diff
@@ -239,6 +239,20 @@ steps:
...
+        command:
+          - julia --color=yes --project=examples examples/hybrid/driver.jl --job_id sphere_baroclinic_wave_rhoe_equilmoist_edmfx --initial_condition MoistBaroclinicWaveWithEDMF --moist equil --precip_model 0M --turbconv edmfx --apply_limiter false --dt 450secs --t_end 6days --dt_save_to_disk 2days --reference_job_id sphere_baroclinic_wave_rhoe_equilmoist
+        artifact_paths: "sphere_baroclinic_wave_rhoe_equilmoist_edmfx/*"
+        agents:
+          slurm_mem: 20GB
+
+      - label: ":computer: EDMFX Bomex on a sphere"
+        command:
+          - julia --color=yes --project=examples examples/hybrid/driver.jl --job_id edmfx_bomex_sphere --initial_condition Bomex --moist equil --turbconv edmfx --subsidence Bomex --edmf_coriolis Bomex --ls_adv Bomex --apply_limiter false --z_elem 60 --z_stretch false --z_max 3e3 --dt 60secs --t_end 1hours --dt_save_to_disk 5mins --reference_job_id edmf_bomex
```
trontrytel, line 251: Should we also add a "consistency test" fir EDMFX?

src/staggered_nonhydrostatic_model.jl
```diff
@@ -28,12 +28,30 @@ import ClimaCore.Fields: ColumnField
 function precomputed_quantities(Y, atmos)
     FT = eltype(Y)
     TST = thermo_state_type(atmos.moisture_model, FT)
+    precomputed_sgs_quantities = if atmos.turbconv_model isa EDMFX
+        n = n_mass_flux_subdomains(atmos.turbconv_model)
+        (;
+            ᶜuʲs = similar(Y.c, NTuple{n, Geometry.Covariant123Vector{FT}}),
```
trontrytel, line 34: Would it be useful to add a short comment here explaining what each of those variables is?

src/callbacks/callbacks.jl
```diff
@@ -283,7 +283,54 @@ function save_to_disk_func(integrator)
...
+            cloud_fraction = ᶜcloud_fractionᴰ,
+        )
+        env_diagnostics = (;
+            common_diagnostics(ᶜu⁰, ᶜts⁰)...,
+            area = ᶜarea⁰,
+            cloud_fraction = ᶜcloud_fraction⁰,
+            tke = divide_by_ρa.(ᶜsgs⁰.ρatke, ᶜsgs⁰.ρa, 0, Y.c.ρ, a_min),
+        )
+        if p.atmos.precip_model isa Microphysics1Moment
+            ᶜq_raiᴰ =
```
trontrytel, line 314: We currently only treat precipitation as a grid mean quantity. So we don't have the draft and environment rain and snow. 

Sorry, I guess that breaks the symmetry here. Would it be possible to opt out rain and snow from the SGS variables?

src/precomputed_quantities.jl
```diff
@@ -48,6 +56,21 @@ function set_velocity_quantities!(ᶜu, ᶠu³, ᶜK, ᶠw, ᶜYc)
...
+"""
+    divide_by_ρa(ρaχ, ρa, ρχ, ρ, a_min)
+
+Computes `ρaχ / ρa`, regularizing the result to avoid issues when `a` is small.
+This is done by performing a linear interpolation from `ρaχ / ρa` to `ρχ / ρ`,
+interpolating closer to `ρχ / ρ` when `a` is small. When `a == 0`, the result is
+exactly `ρχ / ρ` (a tiny value of `ε` is added to the denominator of `ρaχ / ρa`
+in order to avoid returning `NaN` when `a == 0`, since `0 * Inf` is `NaN`).
+"""
+function divide_by_ρa(ρaχ, ρa, ρχ, ρ, a_min)
```
trontrytel, line 68: This is very neat!

szy21 (approved):

src/precomputed_quantities.jl
```diff
@@ -111,16 +134,218 @@ function ts(Yc, K, Φ, thermo_params, energy_form, moisture_model)
...
+                    p,
+                    Φ,
+                    a_min,
+                    thermo_params,
+                    energy_form,
+                    moisture_model,
+                )
+            end
+
+        w⁰(ρa⁰, ρaʲs, wʲs, ρ, w) =
```
trontrytel, line 234: The \^0 looks more like a B on my screen. Not sure how does it render for others?

src/precomputed_quantities.jl
```diff
@@ -111,16 +134,218 @@ function ts(Yc, K, Φ, thermo_params, energy_form, moisture_model)
...
+        ᶜKᴰ,
+        ᶜp,
+        ᶜΦ,
+        a_min,
+        thermo_params,
+        energy_form,
+        moisture_model,
+    )
+
+    return (; ᶜsgs⁰, ᶜsgsᴰ, ᶜuᴰ, ᶜtsᴰ)
```
trontrytel, line 350: `ᶜsgs⁰`,  `ᶜsgsᴰ` are very hard to tell apart. I think we should do something else unfortunately

szy21: Oh, I agree with this. subscript 0 always looks like \theta to me.

src/precomputed_quantities.jl
```diff
@@ -111,16 +134,218 @@ function ts(Yc, K, Φ, thermo_params, energy_form, moisture_model)
...
+
+    @assert turbconv_model isa EDMFX
+
+    (; a_min) = turbconv_model
+    (ᶠwᴰ, ᶜuᴰ, ᶠu³ᴰ, ᶜKᴰ) = similar.((p.ᶠw⁰, p.ᶜu⁰, p.ᶠu³⁰, p.ᶜK⁰))
+
+    wᴰ(ρaʲs, wʲs, ρ, w) =
+        divide_by_ρa(sum(ρaʲs .* wʲs), sum(ρaʲs), ρ * w, ρ, a_min)
+    sgsᴰ(Yc) = map(+, Yc.sgsʲs...)
+    sgs⁰(Yc) = merge(map(-, times_a(Yc, 1), sgsᴰ(Yc)), Yc.sgs⁰)
```
trontrytel, line 324: That replaces a bunch of lines from EDMF!

trontrytel (approved):
Left some minor comments. The biggest thing to remember to ask is if we should delete the precipitation from the SGS variables. 

Lets make shorter PRs in the future!

.buildkite/pipeline.yml
dennisYatunin: How does :genie: look?

.buildkite/pipeline.yml
dennisYatunin: I don't think we'll need a consistency test if we design our code to avoid dependency issues. Moving values from the "aux state" into local (broadcastable) functions will be a big part of that.

src/staggered_nonhydrostatic_model.jl
dennisYatunin: Sure, will do.

src/callbacks/callbacks.jl
dennisYatunin: Yes, it's definitely possible; I can just add an `if`-block to the definition of `times_a` that makes it ignore `ρq_rai` and `ρq_sno`, and then remove the corresponding draft and environment quantities computed in `save_to_disk_func`. Do we want to check with Tapio first, though?

src/precomputed_quantities.jl
dennisYatunin: Per an offline discussion, I'll change the ᴰ to a ⁺. That should be easier to distinguish, and it will match the ⟨⋅⟩₊ notation from Tapio's Overleaf.

dennisYatunin:
bors cancel

PR #1488, opened 2023-03-23
Add initial condition argument in longrun

szy21:
## Purpose 

cc @jiahe23 

## Content

----


jiahe23:
https://github.com/CliMA/ClimaAtmos.jl/blob/main/.buildkite/pipeline.yml#L234

szy21:
cc @akshaysridhar, we forgot to change the argument for one topography case in ci in a previous PR

PR #1494, opened 2023-03-27
Optional reference enthalpy in hyperdiffusion tendency. 

akshaysridhar:
Adds a command line option to define hyperdiffusion tendencies as a function of the perturbation of total enthalpy from a reference state. The reference specific enthalpy is defined based on a state defined by `T_{ref}`, `rho_{ref}` and `p_{ref}`, with zero velocity.

modified:   src/staggered_nonhydrostatic_model.jl
modified:   src/tendencies/hyperdiffusion.jl
modified:   examples/hybrid/cli_options.jl
modified:   docs/src/equations.md


src/staggered_nonhydrostatic_model.jl
```diff
@@ -84,10 +84,16 @@ function default_cache(
     R_d = FT(CAP.R_d(params))
     MSLP = FT(CAP.MSLP(params))
     grav = FT(CAP.grav(params))
+    cv_d = FT(CAP.cv_d(params))
+    T_0 = FT(CAP.T_0(params))
     T_ref = FT(255)
     ᶜΦ = CAP.grav(params) .* ᶜcoord.z
     ᶜρ_ref = @. MSLP * exp(-grav * ᶜcoord.z / (R_d * T_ref)) / (R_d * T_ref)
     ᶜp_ref = @. ᶜρ_ref * R_d * T_ref
+    ᶜh_ref = @. ᶜp_ref / ᶜρ_ref + (cv_d * (T_ref - T_0) + grav * ᶜcoord.z)
```
trontrytel, line 93: I think there are some enthalpy functions in Thermodynamics.jl. Maybe we could use them?

akshaysridhar: Good idea, I see we have functions for `total_specific_enthalpy` and `specific_enthalpy` in `relations.jl` - I'll switch these out.

akshaysridhar: Modified to use `total_specific_enthalpy`.

szy21:
This seems outdated. Should we close it?

PR #1495, opened 2023-03-27
Delete (tmp) env tke related cached variables

trontrytel:
## Purpose 
Tmp delete environmental cahced variables relevant to tke. This is to help with moving edmf code to edmfx.


szy21 (approved):
Great, let's delete more code 🚀 (I think Tapio asked to make sure TurbulenceConvection has everything that is deleted here, but I don't know what we have there.)

trontrytel:
I was looking at old TC.jl code when deleting and seems to me like everything is there. Plus, we can always come back to this PR if needed.

PR #1496, opened 2023-03-27
Move microphysics to closures, combine precipitation files.

trontrytel:
This PR:

- moves the saturated/unsaturated `aux` update from precipitation files into `update_aux`
- combines `EDMF_updrafts`, `EDMF_Environment` with `EDMF_Precipitation` into one file that now only deals with precipitation tendencies (to the grid mean and updrafts). 
- renames the `microphysics_coupling.jl` file and moves it into `closures` folder in EDMF
- deletes the unused `noneq_moisture_sources`.

I think those changes make it easier to separate the precipitation from the rest of EDMF and make it easier to move things to the new code.


src/TurbulenceConvection/EDMF_Precipitation.jl
```diff
@@ -16,8 +134,7 @@ function compute_precipitation_sink_tendencies(
 
     nothing
 end
-
-function compute_precipitation_sink_tendencies(
+function grid_mean_precipitation_sink_tendencies(
```
trontrytel, line 137: @szy21 - this is a grid-mean tendency from grid-mean precipitation sinks. We could move it somewhere into ClimaAtmos?

szy21: Maybe move it to src/parameterizations/microphysics/precipitation.jl?

trontrytel:
Moved to https://github.com/CliMA/ClimaAtmos.jl/pull/1575

PR #1499, opened 2023-03-29
Unify precomputed quantities, clean up diagnostics

dennisYatunin:
## Purpose 

This PR is split off from #1482. Its purpose is to ensure that all computations of velocity, kinetic energy, and thermodynamic state are done consistently, and to clean up some of the diagnostics code. Since the kinetic energy was previously computed incorrectly in `thermo_state!` (using `norm_sqr` instead of `compute_kinetic!`), this PR is expected to slightly change simulation results.

## Content

- Rename `precomputed_quanitites!` to `set_precomputed_quantities!`, and rewrite `set_precomputed_quantities!` so that merging in #1482 requires as few changes as possible.
- Rename the precomputed quantity `ᶜu_bar` to `ᶜu`, per Tapio's request.
- Delete `thermo_state.jl`. Remove `thermo_dispatcher` from `p` and delete the definition of `ThermoDispatcher`. Replace all computations of `ᶜu`, `ᶜK`, and `ᶜts` outside of `precomputed_quantities.jl` with calls to `set_precomputed_quantities!`. In particular, make `horizontal_limiter_tendency!`, `remaining_tendency!`, and `implicit_tendency!` all begin by zeroing out the tendency and calling `set_precomputed_quantities!`. Simplify the definition of `thermo_state_type` and move it from the deleted `thermo_state.jl` to `type_getters.jl`. Define the (broadcastable) functions `ts` and `thermo_state` in `precomputed_quantities.jl`; these functions replace all of the other code that was in `thermo_state.jl` (the `thermo_state` function internally does what `ThermoDispatcher` used to do).
- Isolate `precomputed_quantities` from the rest of the `default_cache`, initialize the `precomputed_quantities` using `set_precomputed_quantities!` when the `default_cache` is constructed, and pass the `default_cache` wherever these quantities are needed in `additional_cache` (i.e., `radiation_model_cache`).
- Rename `water_vapor`, `cloud_liquid`, and `cloud_ice` in the `hdf5` diagnostics files to `q_vap`, `q_liq`, and `q_ice`.
- Replace `dry_diagnostic` and `moist_diagnostic` in `save_to_disk_func` with `dycore_diagnostic` and `precip_diagnostic`. Define `common_diagnostics` (dry and moist) that should be generated for both GS and SGS quantities, and include them in `dycore_diagnostic`. Replace the duplicated computation of the zero-moment microphysics diagnostics with a call to the function that computes those diagnostics in `precipitation.jl` (`precipitation_tendency!`). Generate the `q_rai` and `q_sno` diagnostics for one-moment microphysics, and generate the grid-mean `cloud_fraction` diagnostic for EDMF. Change the condition for generating EDMF diagnostics to `p.atmos.turbconv_model isa TC.EDMFModel`.
- Move the `include` call for `callbacks.jl` after the `include` calls for the tendency files, so that the duplicate computation of the zero-moment microphysics diagnostics can be removed.
- Remove all unused variables from `horizontal_limiter_tendency!`, `horizontal_advection_tendency!`, `explicit_vertical_advection_tendency!`, `implicit_vertical_advection_tendency`, `Wfact!`, `default_cache`, `radiation_model_cache`, `save_to_disk_func`, `rrtmgp_model_callback`, `turb_conv_affect_filter!`, and all of the plotting functions in `post_processing_funcs.jl`; some of these variables are no longer used because the calls to `compute_kinetic!` and `thermo_state!` were removed, but many of them were not being used to begin with.
- Increment the `ref_counter` and reset the MSE tables to 0.

----


szy21 (approved):
The code change looks good to me. It seems something is breaking in the post processing. I'll look at the plots once you fix that.

src/precomputed_quantities.jl
```diff
@@ -3,49 +3,124 @@
...
+        compute_kinetic!(ᶜK[colidx], ᶜYc.uₕ[colidx], ᶠw[colidx])
+    end
+end
+
+"""
+    thermo_state(thermo_params; ρ, p, θ, e_int, q_tot, q_pt)
+
+Constructs a thermodynamic state based on the supplied keyword arguments.
+"""
+function thermo_state(
```
trontrytel, line 56: I like this way of dispatching over different variables!

src/precomputed_quantities.jl
```diff
@@ -3,49 +3,124 @@
...
+        elseif moisture_model isa NonEquilMoistModel
+            q_liq = Yc.ρq_liq / Yc.ρ
+            q_ice = Yc.ρq_ice / Yc.ρ
+            (; q_pt = TD.PhasePartition(q_tot, q_liq, q_ice))
+        end
+    end
+    return thermo_state(thermo_params; ρ = Yc.ρ, energy_var..., moisture_var...)
+end
+
+function set_precomputed_quantities!(Y, p, t)
```
trontrytel, line 114: Was wondering if we could go for a more descriptive name here? For example `set_thermo_state_and_pressure`. Though I guess we rick it becoming outdated if we add more things there

src/staggered_nonhydrostatic_model.jl
```diff
@@ -228,22 +237,17 @@ function dss!(Y, p, t)
...
-    (; ᶜinterp) = p.operators
-    ᶜuₕ = Y.c.uₕ
-    ᶠw = Y.f.w
-    @. ᶜu_bar = C123(ᶜuₕ) + C123(ᶜinterp(ᶠw))
-
     Yₜ .= zero(eltype(Yₜ))
+    set_precomputed_quantities!(Y, p, t)
+
+    (; ᶜu) = p
+    divₕ = Operators.Divergence()
```
trontrytel, line 244: Why do we store `interp` in `p.operators`, but have to define `div_h` here? - Just asking to understand the logic on "what goes where"

trontrytel (approved):
LGTM! I looked at the EDMF cases that were flagged by the regression tests and everything seems fine there.

src/precomputed_quantities.jl
```diff
@@ -3,49 +3,124 @@
...
-    @. ᶜp[colidx] = TD.air_pressure(thermo_params, ᶜts[colidx])
-    return nothing
+"""
+    set_velocity_quantities!(ᶜu, ᶠu³, ᶜK, ᶠw, ᶜYc)
+
+Uses `ᶠw` and `ᶜYc` to set the values of `ᶜu`, `ᶠu³`, and `ᶜK`, assuming that
+the horizontal velocity is `ᶜYc.uₕ` and that it should be interpolated to cell
+faces using `ᶜYc.ρ * ᶜJ` in the weighted interpolation.
+"""
+function set_velocity_quantities!(ᶜu, ᶠu³, ᶜK, ᶠw, ᶜYc)
```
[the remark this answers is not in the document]
dennisYatunin: So that the SGS velocity quantities can be computed using the same function, but with `ᶠsgsʲs[j].w` instead of `ᶠw` (still using the same `ᶜYc`, though).

dennisYatunin:
Longrun build is running [here](https://buildkite.com/clima/climaatmos-longruns/builds/864). I'll merge this in now.

dennisYatunin:
bors cancel

src/staggered_nonhydrostatic_model.jl
trontrytel: I agree, I think it would be cleaner to define them once for all ClimaAtmos.

Issue #1383, opened 2023-02-22
ClimaAtmos.jl folder restructuring plan


bischtob:
### Purpose
We need a clean and simple code structure for ClimaAtmos.jl that is not to dissimilar from conventional already existing climate models. File and folder naming conventions need to reflect this. In addition, we want that the code functions like a collection of libraries that can easily be extended. This issue complements the testing issue #1328. The goal is to have the equations of motion be localized and extended by various right-hand side and left-hand-side terms that can be toggled on and off and that live in easy to find and change places in the code.

As guiding examples we will use the NOAA-GFDL models' source code as well as the Oceananigans.jl source code (see references).

### Cost/Benefits/Risks
**Costs**: 
Several developer/scientist days/weeks

**Benefits**:
Clean and recognizable code structure for internal and external contributors. Streamlined documentation that reflects current state of the code.

**Risks**:
Disruption of concurrent development efforts and hence delays of other ClimaAtmos.jl deliverables. The aim of this plan is to minimize this risk as much as possible to not incur **unnecessary** delays. However, some delays due to restructuring may be inevitable.

In order to mitigate these risks, we plan on first restructuring the folder structure on a series of PRs without renaming functions. This way rebasing existing work streams remains easy. After that, we will restructure some functions in a series of small PRs to keep the rebasing burden low.

### People and Personnel
- Lead: @bischtob 
- Collaborators: @akshaysridhar 
- Reviewers: @simonbyrne, @szy21 

### Components
**Part 1 (new file & folder structure)**
We aim to reorganize the files and folder structure to reflect a common climate model and to be similar to Oceananigans.jl's code structure.The following sketch shows the plan
```
src
     - prognostic_equations
         - staggered_nonhydrostatic_model.jl
         - edmf.jl
         - implicit.jl
         - ...  
     - tendencies
          - gravity_wave
              - file1.jl
              - file2.jl
              - ....
          - radiation
              - file1.jl
              - file2.jl
              - ....
          - ...
     - utils
     - shared
     - ...
test 
     - prognostic_equations
         - test_cool_thing1.jl
         - ...
     - tendencies
          - test_cool_thing2.jl
         - ...
     - utils
         - test_a_util.jl 
     - shared
     - ...
validation 
     - driver1.jl
     - driver2.jl
     - ...
     - utils_val.jl
examples 
     - driver99.jl
     - driver98.jl
     - ...
     - utils_examples.jl
performance
     - driver1.jl
     - ...
[...]

PR #1516, opened 2023-03-31
Move TurbulenceConvectionUtils to src

charleskawczynski:
This PR moves TurbulenceConvectionUtils into src. To help push along #1383

Perhaps we want to distribute these functions into different places in src? Suggestions?

Co-authored by @jiahe23


jiahe23:
In the end, we would like to have `dycore_equations`, `physical_parameterizations`, `shared_utils` in the src. EDMF will sit in the dycore folder. We have moved as many parameterizations as we can to the parameterization folder for now, but haven't touch the dycore part yet (to avoid to many conflicts with the ongoing developments/debugging). 

I think for now it is ok to move the TCUtils to the TC folder and later on move the entire TC into dycore folder once it's ready.

charleskawczynski:
> In the end, we would like to have `dycore_equations`, `physical_parameterizations`, `shared_utils` in the src. EDMF will sit in the dycore folder. We have moved as many parameterizations as we can to the parameterization folder for now, but haven't touch the dycore part yet (to avoid to many conflicts with the ongoing developments/debugging).
> 
> I think for now it is ok to move the TCUtils to the TC folder and later on move the entire TC into dycore folder once it's ready.

That all sounds good, I don't mind waiting if that's smoother

PR #1526, opened 2023-04-01
Fix perf bug in edmf update aux

charleskawczynski:
Broadcast computations for kinetic energy were accidentally added inside a `k`-loop (`@inbounds for k in real_center_indices(grid)`) [here](https://github.com/CliMA/ClimaAtmos.jl/pull/1439/files#diff-4609bee2e6e4c6cb0baf3fa9e708cc2cc2d7b8a4f5e13a27acfd795034d3a955R273-R284) in #1439.

This shouldn't be behavior changing, but it should definitely improve performance of edmf-related jobs.

### DIFF
--- a/src/TurbulenceConvection/update_aux.jl
+++ b/src/TurbulenceConvection/update_aux.jl
@@ -245,18 +245,19 @@ function update_aux!(
             0
         end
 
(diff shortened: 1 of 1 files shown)


charleskawczynski:
Looks like this is non-behavior changing

PR #1540, opened 2023-04-04
Move regression test to the end

szy21:
## Content

----

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -113,7 +113,9 @@ steps:
         artifact_paths: "single_column_radiative_equilibrium_allsky_idealized_clouds/*"
 
       - label: ":computer: single column non-orographic gravity wave parameterization"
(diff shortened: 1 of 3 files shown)


szy21:
@charleskawczynski I think the multiple command doesn’t work well with multi-line command for TC?

.buildkite/pipeline.yml
```diff
@@ -501,6 +523,7 @@ steps:
           --moist equil
           --job_id edmf_soares
           --dt 0.25secs
+          - julia --color=yes --project=examples regression_tests/test_mse.jl --job_id edmf_soares --out_dir edmf_soares
```
charleskawczynski, line 526:
```suggestion
        command: julia --color=yes --project=examples regression_tests/test_mse.jl --job_id edmf_soares --out_dir edmf_soares
```
What if we did this instead?

PR #1569, opened 2023-04-11
compute T tensor and hmax from elevation

### LINKED ISSUE #1430: OGW: compute orographic tensor directly from elevation
Currently, the orographic tensor (which contains all topography related information including amplitude, variance, orientation, and anisotropy) is obtained from a restart file of GFDL model. We need to directly compute it from the known elevation.

jiahe23:
closes #1430


jiahe23:
hmax

jiahe23:
t11

jiahe23:
t12

jiahe23:
t21

jiahe23:
t22

jiahe23:
Those plots compare the hmax and tensor T patterns between gfdl restart file and the calculation from elevation. In general, they demonstrate reasonably well resemblance in spatial structure; while the magnitude are off, which is most likely due to the differences in the input data resolution (1/30 degrees fro gfdl and 0.83 degrees for us).

Therefore, I added the option to choose between using gfdl restart and direct calculation from elevation, and set the gfdl restart as default.

test/orographic_gravity_wave/ogwd_3d.jl
```diff
@@ -201,8 +201,8 @@ epsilon = 0.622
 @. Y.c.ρ = gfdl_ca_p / Y.c.T / R_d / (1 - Y.c.qt + Y.c.qt / epsilon)
 
 # Initialize cache vars for orographic gravity wave
-ogw = CA.OrographicGravityWave{FT}()
-p = CA.orographic_gravity_wave_cache(ogw, Y)
+ogw = CA.OrographicGravityWave{FT, String}()
+p = CA.orographic_gravity_wave_cache(ogw, Y, FT(6.4e6))
```
szy21, line 205:
```suggestion
p = CA.orographic_gravity_wave_cache(ogw, Y, FT(radius))
```

test/orographic_gravity_wave/ogwd_baseflux.jl
```diff
@@ -46,8 +46,8 @@ end
 Y = Fields.FieldVector(c = Yc, f = Yf)
 
 # Initialize cache vars for orographic gravity wave
-ogw = CA.OrographicGravityWave{FT}()
-p = CA.orographic_gravity_wave_cache(ogw, Y)
+ogw = CA.OrographicGravityWave{FT, String}()
+p = CA.orographic_gravity_wave_cache(ogw, Y, FT(6.4e6))
```
szy21, line 50:
```suggestion
p = CA.orographic_gravity_wave_cache(ogw, Y, FT(radius))
```

charleskawczynski:
It looks like there several components added here that are GPU-incompatible. Are we okay with this? cc @tapios, @cmbengue. This is adding to the list of GPU work needed for an atmosphere model (assuming that we want to use these added components).

src/utils/types.jl
```diff
@@ -78,7 +78,7 @@ Base.@kwdef struct NonOrographyGravityWave{FT} <: AbstractGravityWave
     dϕ_s::FT = -5
 end
 
-Base.@kwdef struct OrographicGravityWave{FT} <: AbstractGravityWave
+Base.@kwdef struct OrographicGravityWave{FT, ST} <: AbstractGravityWave
```
charleskawczynski, line 81: Can we define a type rather than using strings here?

src/parameterizations/gravity_wave_drag/orographic_gravity_wave_helper.jl
```diff
@@ -1,7 +1,212 @@
 using NCDatasets
 using Interpolations
 
-function get_OGW_info(Y, orographic_info_rll)
+include(joinpath(pkgdir(ClimaAtmos), "artifacts", "artifact_funcs.jl"))
```
charleskawczynski, line 4: Can we fix the code loading here? This file is being repeatedly included in `src/` and I've been seeing a lot of gravity wave-related errors in CI lately (possibly related to this)

src/parameterizations/gravity_wave_drag/orographic_gravity_wave_helper.jl
```diff
@@ -1,7 +1,212 @@
...
+    dAdy =
+        hcat(
+            A[:, 2] .- A[:, 1],
+            (A[:, 3:end] .- A[:, 1:(end - 2)]) / FT(2),
+            A[:, end] .- A[:, end - 1],
+        ) ./ (deg2rad(maximum(lat) - minimum(lat)) * earth_radius)
+    return (dAdx, dAdy)
+end
+
+function calc_hmax_latlon(elev, lon, lat)
```
charleskawczynski, line 48: Can we add a doc string to this?

src/parameterizations/gravity_wave_drag/orographic_gravity_wave_helper.jl
```diff
@@ -1,7 +1,212 @@
...
+            elev_highpass = elev[irange, jrange] .* hp_wts
+            hmax[i, j] =
+                (mean((elev_highpass .- mean(elev_highpass)) .^ FT(4)))^FT(0.25)
+        end
+    end
+    hmax[:, abs.(lat) .> FT(89)] .= FT(0)
+    return hmax
+end
+
+function calc_velocity_potential(elev, lon, lat, earth_radius)
```
charleskawczynski, line 115: Can we add a doc string to this?

src/parameterizations/gravity_wave_drag/orographic_gravity_wave_helper.jl
```diff
@@ -35,6 +240,59 @@ function get_OGW_info(Y, orographic_info_rll)
     return topo_cg
 end
 
+function compute_OGW_info(Y, elev_data, earth_radius)
```
charleskawczynski, line 243: Can we add a doc string to this?

jiahe23:
> It looks like there several components added here that are GPU-incompatible. Are we okay with this? cc @tapios, @cmbengue. This is adding to the list of GPU work needed for an atmosphere model (assuming that we want to use these added components).

Can you make it clear which components are GPU-incompatible and maybe software engineers can help out?

Also, this PR is to provide the option that we compute the orographic tensor from elevation in addition to the default setting of using the GFDL restart file. We don't need to run this option for every run. In fact, we would in the end also archive a restart type of file that saves the orographic tensor and run every simulation from it as oppose to do this calculation every single time. But it is still good to save the functions somewhere so that if someone wants to run a mountain of analytical shape, there is a way to compute those tensors.

tapios:
> > It looks like there several components added here that are GPU-incompatible. Are we okay with this? cc @tapios, @cmbengue. This is adding to the list of GPU work needed for an atmosphere model (assuming that we want to use these added components).
> 
> Can you make it clear which components are GPU-incompatible and maybe software engineers can help out?
> 
> Also, this PR is to provide the option that we compute the orographic tensor from elevation in addition to the default setting of using the GFDL restart file. We don't need to run this option for every run. In fact, we would in the end also archive a restart type of file that saves the orographic tensor and run every simulation from it as oppose to do this calculation every single time. But it is still good to save the functions somewhere so that if someone wants to run a mountain of analytical shape, there is a way to compute those tensors.

If we can achieve GPU compatibility with low effort, we should do it, even if this is a one-time offline calculation. 

Also, can you generate these tensors from higher-resolution topography, to make them directly comparable with GFDL?

charleskawczynski:
As we discussed in the atmos meeting, let's merge even if we can't resolve everything. @jiahe23, can you confirm that we cannot remove the call to `parent` in the `bycolumn` loop? That's one part, the other piece would be to have some sort of wrapper around the NC files that are being read in. We'll need to convert these to and from CuArrays to work when we start doing gpu runs, this is more minor and I think we can fix this up later when we start running gpu atmos simulations.

jiahe23:
> As we discussed in the atmos meeting, let's merge even if we can't resolve everything. @jiahe23, can you confirm that we cannot remove the call to `parent` in the `bycolumn` loop? That's one part, the other piece would be to have some sort of wrapper around the NC files that are being read in. We'll need to convert these to and from CuArrays to work when we start doing gpu runs, this is more minor and I think we can fix this up later when we start running gpu atmos simulations.

Yes, I will try if we could work things out without parent; and update what I found out here.

jiahe23:
#1633 for hmax calculation
#1662 for T tensor calculation

PR #1574, opened 2023-04-17
use total energy as EDMF prognostic variable

### LINKED ISSUE #1471: Test using total energy in the updraft equations (not in second order terms)

szy21:
## Purpose 

Co-authored-by: @trontrytel. Closes #1471  

## To-do

- [x] Update documentation

## Content

Two major changes:

- Adds the volume term (-p (\partial a/\partial t)) to the \rhoae_tot tendency. This is approximated by (-p/\rho_gm) (\partial \rhoa / \partial t) 
- Adds a filter to rhoa and rhoae_tot consistently with rhoaq_tot

----


szy21:
All the JFNK IMEX cases break, @trontrytel Any idea why and what we should expect? The EDMF consistency test also breaks, so I guess I did something wrong with the initialization. @charleskawczynski Any idea?

szy21:
I think there might be some issue with [this](url) function. Something is not correctly initialized so θ_surf can be a random value, e.g. 0. It was not a problem for thetali because this will be updated at the next time step, but is a problem for total energy because we are constructing a thermo state for the surface form θ_surf during the initialization.

charleskawczynski:
> All the JFNK IMEX cases break, @trontrytel Any idea why and what we should expect? The EDMF consistency test also breaks, so I guess I did something wrong with the initialization. @charleskawczynski Any idea?

Hm, it generally suggests that a variable is used before it's defined. Although, IRRC, the test is a vanilla Bomex test. If the original one has changed, it may need tweaking / synchronization to remain stable.

trontrytel:
> > All the JFNK IMEX cases break, @trontrytel Any idea why and what we should expect? The EDMF consistency test also breaks, so I guess I did something wrong with the initialization. @charleskawczynski Any idea?
> 
> Hm, it generally suggests that a variable is used before it's defined. Although, IRRC, the test is a vanilla Bomex test. If the original one has changed, it may need tweaking / synchronization to remain stable.

We started printing things and see some NaN at the surface. Should get to the bottom of this soon. But I vote to get rig of `assign_thermo_aux` as a next refactoring step or even during the surface refactor PR

src/TurbulenceConvection/EDMF_functions.jl
```diff
@@ -258,25 +258,40 @@ function set_edmf_surface_bc(
...
     prog_gm = center_prog_grid_mean(state)
+    prog_gm_uₕ = grid_mean_uₕ(state)
     prog_up = center_prog_updrafts(state)
     prog_en = center_prog_environment(state)
     prog_up_f = face_prog_updrafts(state)
     ts_gm = center_aux_grid_mean_ts(state)
     cp = TD.cp_m(thermo_params, ts_gm[kc_surf])
     ρ_c = prog_gm.ρ
-    ae_surf::FT = 1
+    aux_up = center_aux_updrafts(state)
```
trontrytel, line 268: We are not using `aux_up` here

trontrytel (approved):
Awesome. Thank you!

PR #1585, opened 2023-04-19
 Fix ρ_q_tot and ρ_θ for surface conditions

nefrathenrici:
Small MSE-changing PR. Changes how `ρ_q_tot` and `ρ_θ` are calculated for the surface conditions to match the new EDMF. This requires passing the thermo state in via the `State` struct.

I still need to format, but the MSE changes are only small so it should be good to go.


src/TurbulenceConvection/EDMF_functions.jl
```diff
@@ -336,8 +333,8 @@ function θ_surface_bc(
     prog_gm = center_prog_grid_mean(state)
     ρ_c = prog_gm.ρ
     kc_surf = kc_surface(grid)
-    ts_gm = center_aux_grid_mean_ts(state)
-    c_p = TD.cp_m(thermo_params, ts_gm[kc_surf])
+    ts_sfc = state.ts
+    c_p = TD.cp_m(thermo_params, ts_sfc)
```
szy21, line 337: Why we don't need the kc_surf here anymore?

src/InitialConditions/initial_conditions.jl
```diff
@@ -673,7 +673,7 @@ end
 ##
 
 # By default, use the dycore's surface parametrization.
-surface_params(::InitialCondition, thermo_params) = nothing
+surface_params(::InitialCondition, thermo_params) = surface_params(Bomex(), thermo_params)
```
trontrytel, line 676: Why are we prescribing Bomex here?

src/TurbulenceConvection/EDMF_functions.jl
nefrathenrici: `c_p` should be calculated from the thermo state at the surface instead of at `kc_surf = Cent(1)`, so instead of using `kc_surf` we just pass in the thermo state at the surface directly using `state.ts`

src/InitialConditions/initial_conditions.jl
nefrathenrici: This is for CI cases where `turbconv_case` doesn't get set but EDMF is still run and needs some pseudo-default `InitialCondition` to get the `surface_params`. This will be removed again in the full surface PR.
For example this [case](https://buildkite.com/clima/climaatmos-ci/builds/9069#01879ad7-955c-4d04-bdb0-4ad8933da133):
```
julia --color=yes --project=examples examples/hybrid/driver.jl 
--job_id sphere_baroclinic_wave_rhoe_equilmoist_edmf --vert_diff true 
--surface_scheme bulk --moist dry --dt 1secs --t_end 20secs 
--FLOAT_TYPE Float64 --turbconv edmf --config sphere --z_stretch false 
--post_process false --apply_limiter false --dt_save_to_sol 10secs --dt_save_to_disk 10secs
```

src/TurbulenceConvection/EDMF_functions.jl
szy21: So `state.ts` is the thermo state at the surface, not the thermo state of the entire column?

nefrathenrici: Yes, it's the thermo state at the surface

charleskawczynski: > Yes, it's the thermo state at the surface

Can we rename it to `ts_sfc` or something more clear?

charleskawczynski (commented):
As a general comment, we should push to _eliminate_ `surf_params` because it is a column-wise-only construct (it may not vary across columns). So, I'm a bit concerned that it's use seems to be spreading in this PR.

nefrathenrici:
@charleskawczynski `surf_params` will be totally removed in the [surface unification PR](https://github.com/CliMA/ClimaAtmos.jl/pull/1447/files), this is meant to be a small bug-fix/MSE changing PR so that the next one doesn't have any regression.

dennisYatunin (approved):

PR #1512, opened 2023-03-31
Add advection, hyperdiffusion, and limiters to EDMFX model

dennisYatunin:
## Purpose 

Adds advection, hyperdiffusion, and limiters to the EDFMX model.

## Content

- Modify the `MoistBaroclinicWaveWithEDMF` initial condition so that it sets the initial draft area to 1 and the initial TKE to 0. Add an additional `DryBaroclinicWaveWithEDMF` initial condition, which sets the same values of area and TKE.
- Add `variable_manipulations.jl`, which will contain all useful manipulations of grid-scale and sub-grid-scale variables. Move `divide_by_ρa` to this file, along with the new regularization function `sgs_weight_function`. Move `times_a` to this file, and rename it to `gs_to_sgs`. Add the additional generated functions `specific_gs`, `specific_sgs`, and `matching_subfields`, all of which need to be generated in order to avoid triggering allocations due to type inference failures (ideally, these will not need to be generated in future versions of Julia). Add other helpful functions for computing environment and draft quantities. Ensure that all functions in `variable_manipulations.jl` are thoroughly documented.
- Rename the EDFMX model's `a_min` to `a_half`, in order to reflect how `sgs_weight_function` uses this value.
- Add the specific grid-scale and sub-grid-scale states (`ᶜspecific`, `ᶜspecific⁰`, and `ᶜspecificʲs`) to the precomputed quantities. Use the specific states to greatly simplify the computation of the thermodynamic states. Comment out the modifications to `ᶜK` due to the sub-grid-scale quantities, at least until the modifications are fully detailed in the Overleaf document. Add a hacky roundoff error correction to `ᶜρʲ` in order to ensure that the EDMFX CI examples with a draft area of 1 exhibit the correct behavior (the sub-grid-scale quantities should match the grid-scale quantities when there is no precipitation). Change the diagnostic sub-grid-scale quantities to `ᶜspecific⁺`, `ᶠw⁺`, `ᶜu⁺`, `ᶠu³⁺`, `ᶜK⁺`, `ᶜts⁺`, `ᶜa⁺`, and `ᶜa⁰`. Remove all uses of `getindex` for column fields (e.g., `Y.c.sgsʲs[j]`) in `precomputed_quantities.jl`, since using this syntax involves type piracy through the `TurbulenceConvection` module, and replace them the default `getproperty` syntax instead (e.g., `Y.c.sgsʲs.:($j)`, or `@. Y.c.sgsʲs.:($$j)`).
[...]

PR #1595, opened 2023-04-22
Simplify cache and remove TempestHyperdiffusion

dennisYatunin:
## Purpose 

This PR is also split off from #1512. It removes `TempestHyperdiffusion`, which is no longer needed, and slightly simplifies the cache `p`.

## Content

- Remove the `TempestHyperdiffusion` struct and the `enable_qt_hyperdiffusion` function. Remove the `B` parameter from the `ClimaHyperdiffusion` struct.
- Allow the `hyperdiff` CLI option to be `ClimaHyperdiffusion`, `true`, `none` or `false`, and remove the `enable_qt_hyperdiffusion` CLI option.
- Rename `horizontal_limiter_tendency!` to `limited_tendency!`, and rename `hyperdiffusion_tracers_tendency!` to `tracer_hyperdiffusion_tendency!`.
- Move the `hyperdiffusion_cache` to `default_cache`, and move the `hyperdiffusion_tendency!` to `remaining_tendency!`.
- Move the `NVTX.@range` outside of `hyperdiffusion_tendency!` and `tracer_hyperdiffusion_tendency!`.
- Separate the `hyperdiffusion_ghost_buffer` from the main `ghost_buffer`, and allocate it in `hyperdiffusion_cache`.
- Replace `p.ghost_buffer.skip_dss` with `p.do_dss`.
- Only apply DSS during hyperdiffusion if `do_dss` is `true`.
- Replace `ᶜρh` with `p.ᶜtemp_scalar`, `ᶜω³` with `p.ᶜtemp_CT3`, and `ᶠω¹²` with `p.ᶠtemp_CT12`.
- Replace the n identical copies of the limiter in `p` with a single limiter.

----


perf/jet_nfailures.jl
```diff
@@ -35,7 +35,7 @@ using Test
     # inference. By increasing this counter, we acknowledge that
     # we have introduced an inference failure. We hope to drive
     # this number down to 0.
-    n_allowed_failures = 30
+    n_allowed_failures = 31
```
charleskawczynski, line 38: What’s the inference failure being introduced? Can you try to fix this if it’s not too much trouble?

charleskawczynski (commented):
Can we fix the inference failure?

and what is behavior changing in this PR?

dennisYatunin:
@charleskawczynski The behavior change is not really a behavior change--I moved the hyperdiffusion tendency from the additional tendency to right after the horizontal advection tendency, and this change in the order of floating point additions modified the last bit or two in the remaining tendency. I'll see if I can fix the inference failure.

dennisYatunin:
@charleskawczynski The inference failure is from moving the `NVTX.@range` that was in `hyperdiffusion_tendency!` to `remaining_tendency!` and changing it to an `@nvtx`. I think we can let this slide for now, right?

charleskawczynski:
Sounds good, thank you @dennisYatunin!

PR #1598, opened 2023-04-25
Add volume term to EDMFX

szy21:
## Purpose 

Add volume term to the EDMF equations
co-authored-by: @jiahe23 @trontrytel 

## Content

- Add volume term to the subgrid-scale total energy equation
- Change the initial condition of draft area to 0.2 

----

### DIFF
--- a/src/tendencies/implicit/implicit_tendency.jl
+++ b/src/tendencies/implicit/implicit_tendency.jl
@@ -122,6 +122,9 @@ function implicit_vertical_advection_tendency!(Yₜ, Y, p, t, colidx)
                 dt,
                 edmfx_upwinding,
             )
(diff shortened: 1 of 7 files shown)


src/tendencies/implicit/implicit_tendency.jl
```diff
@@ -122,6 +122,8 @@ function implicit_vertical_advection_tendency!(Yₜ, Y, p, t, colidx)
                 dt,
                 edmfx_upwinding,
             )
+            @. Yₜ.c.sgsʲs.:($$j).ρae_tot[colidx] -=
```
trontrytel, line 125: Lets think on how to ensure that the `Y_t` has all the tendencies that we need it to have. Should we cache it? Maybe at least we could leave a comment about what order of updates we are assuming.

trontrytel (approved):
LGTM, just one question about the tendency

Issue #1417, opened 2023-03-03
Implement entrainment/detrainment boundary conditions in EDMFX


trontrytel:
Following @costachris 

```
entr = (1/z)*(-4.013288 - 0.000968*Π₁ + 0.356974*Π₃ - 0.403124*Π₄ + 1.503261*Π₆
detr = (3.535208 + 0.598496*Π₁ + 1.583348*Π₃ + 0.046275*Π₄ - 0.344836*Π₆) + max_area_limiter
```
where `max_area_limiter` forces more detrainment when updraft area approaches max area. Without it, an overwhelming majority of TC simulations failed in the first couple of EKI iterations when updraft area maxed out. 

The max area limiter is
`max_area_limiter = 0.1 * exp(-10 * (maximum_area - updraft_area))`

The Pi-groups are defined in [here](https://github.com/CliMA/TurbulenceConvection.jl/blob/9274195055be873f19e9540c6a7a0e777842311c/src/closures/nondimensional_exchange_functions.jl)

Calibrations are unstable for deep convection (TRMM) and raining cumulus (Rico). Regardless, these should be enough to get started with EDMFX

trontrytel:
Also, the above formulation is `1/length`. We want it to be `1/time`, for example to treat parts of it implicitly or to be able to limit the tendency my the model time step. 

We should absorb the physical velocity into entrainment detrainment. This will give us the form we want

trontrytel:
@costachris - looking at the units here. Shouldn't `detr` also be divided by `z`?

PR #1613, opened 2023-04-28
EDMFX entrainment/detrainment

trontrytel:
For now I'm adding entrainment ~ `1/z` and constant detrainment to simplify the testing. I'll keep the full formulation in the code, so that we don't get used to simple things too much.

Entrainment needs to know the buoyancy flux at the surface. Therefore the entrainment closure assumes that `get_surface_fluxes!` from `vertical_diffusion_boundary_layer_tendency` was called. Therefore we need to run EDMFX with:
`--vert_diff true` to switch it on, 
`--C_E 0` to zero out the coefficients
`--surface_scheme bulk` chose some surface scheme (for example bulk).

Closing #1417


src/tendencies/edmfx_closures.jl
```diff
@@ -0,0 +1,175 @@
...
+                    thermo_params,
+                    ᶜts⁰[colidx],
+                    ᶜspecific⁰.e_tot[colidx],
+                ) -
+                ᶜentr_detrʲs.:($$j).detr[colidx] * TD.total_specific_enthalpy(
+                    thermo_params,
+                    ᶜtsʲs.:($$j)[colidx],
+                    ᶜspecificʲs.:($$j).e_tot[colidx],
+                ) -
+                ᶜp[colidx] / Y.c.ρ[colidx] * (
```
szy21, line 155: You shouldn't need this term here anymore:)

src/tendencies/edmfx_closures.jl
```diff
@@ -0,0 +1,175 @@
...
+            Y.c.sgsʲs.:($$j).ρa[colidx] * (
+                ᶜspecific⁰.q_tot[colidx] * ᶜentr_detrʲs.:($$j).entr[colidx] -
+                ᶜspecificʲs.:($$j).q_tot[colidx] *
+                ᶜentr_detrʲs.:($$j).detr[colidx]
+            )
+
+        # TODO - is this C3 correct?
+        @. Yₜ.f.sgsʲs.:($$j).w[colidx] += ᶠinterp(
+            ᶜentr_detrʲs.:($$j).entr[colidx] *
+            (C3(ᶜu⁰[colidx] - ᶜuʲs.:($$j)[colidx])),
```
szy21, line 171: Is `Y.f.sgsʲs.w` better than `ᶠinterp(C3(ᶜuʲs))` here?

szy21 (approved):
I didn't check the actual pi groups but I assume it's the same as the old code:) Otherwise it looks good to me, thanks!

src/tendencies/edmfx_closures.jl
trontrytel: Great!

src/tendencies/vertical_diffusion_boundary_layer.jl
```diff
@@ -263,6 +265,7 @@ function get_surface_fluxes!(Y, p, t, colidx, ::VerticalDiffusion)
         uₕ_int_phys,
         params,
         coupling,
+        buoy_flux_surface,
```
charleskawczynski, line 268: Shouldn't we just use `sfc_conditions[colidx].buoy_flux` in place of where this is used instead?

charleskawczynski: Or are they allowed to have different values for some reason?

trontrytel: I think you are right. For some reason I thought we are not caching the full `sfc_conditions`. Thanks. I'll open a PR with a fix.

trontrytel: And really thank you for reviewing!

PR #1614, opened 2023-04-28
advection test for edmfx in a box

### LINKED ISSUE #1600: Test EDMFX in kinematic advection case
The kinematic test case: 
- use box setup
- set momentum tendency to be zero for the grid-mean and EDMFX (We want to keep this test case for a while in ClimaAtmos, so lets add it properly)
- use constant velocity to test our advection especially at the updraft edge (top and bottom)

Different things to test:
- [ ] different updraft area fraction profiles going from very smooth to more steep
- [x] upwinding 1st and 3rd, center, others?
- [x] positive and negative velocities
- [x] without/with moisture
- [ ] stretched vs non-stretched
- [ ] box vs sphere

Notes from meeting on May 3rd:
 - [x] Test for smoother initial conditions. We want to check if oscillations in updraft area fraction are reduced with smoother initial condition.
 - [ ] Try upwinding everywhere in SGS (also summoning @Zhengyu-Huang to weigh in on this)
 - [ ] Change the weighs in regularization of precomputed quantities such that we tend to grid-mean properties when area goes to zero faster.
 - [ ] Test also for not constant initial potential temperature profile
 - [ ] Test on the sphere with horizontal advection

Tests to be done later:
 - [ ] explicit vs implicit time stepping (good test case for the volume term and implicit handling of updraft edges) -> Once we have implicit solver for EDMFX
 - [ ] with/without limiters -> If we ever decide we need limiters in the vertical for SGS scheme

jiahe23:
closes #1600


src/InitialConditions/initial_conditions.jl
```diff
@@ -378,6 +378,54 @@ function (initial_condition::MoistBaroclinicWaveWithEDMF)(params)
...
+        T, p = temp_profile(thermo_params, z)
+        if perturb
+            T += rand(FT) * FT(0.1) * (z < 5000)
+        end
+
+        return LocalState(;
+            params,
+            geometry = local_geometry,
+            thermo_state = TD.PhaseDry_pT(thermo_params, p, T),
+            velocity = Geometry.UVVector(FT(0), FT(0)),
```
szy21, line 418:
```suggestion
```
(The default is zero I think)

szy21 (approved):
Looks good, thanks!

trontrytel (approved):
Looks great to me!

PR #1616, opened 2023-04-28
fix a bug in Earth Warp setup

### LINKED ISSUE #1615: A bug in earth warp setup
THe smoothed earth topography is not used on the interpolation object https://github.com/CliMA/ClimaAtmos.jl/blob/main/src/utils/type_getters.jl#L105

jiahe23:
This PR fixes a bug in the Earth Warp setup.

closes #1615


jiahe23:
@akshaysridhar  Do we want to keep the smooth_degree to be 15 here?

akshaysridhar:
> @akshaysridhar Do we want to keep the smooth_degree to be 15 here?

Yes, for now excessive smoothing is fine as we iron out any remaining issues within the test-configurations.

PR #1620, opened 2023-04-30
Move code into src

charleskawczynski:
This PR
 - moves more code into src.
 - Removes the `ordinary_diff_eq_bug_fixes` since no jobs use this feature anymore, and our diffeq compat environment was borked due to upstream incompatibilities for unmaintained compat ranges.


charleskawczynski:
A step towards #1383.

PR #1605, opened 2023-04-26
Add hyperdiffusion to vertical velocity

szy21:
## Purpose 

Adds hyperdiffusion to grid-scale and subgrid-scale vertical velocity. Closes #1604 

## To-do

- [x] Update doc
- [x] Update ref counter and mse table

## Content

----
- [ ] I have read and checked the items on the review checklist.

PR #1623, opened 2023-05-01
Reduce dt in some SSP longrun jobs

szy21:
## Purpose 

Two SSP moist baroclinic wave jobs (central and first-order upwinding) failed after #1605. With a slightly smaller dt they run stably (see build [949](https://buildkite.com/clima/climaatmos-longruns/builds/949#0187d600-35e6-4351-881c-7fafaeae042a)), so I reduced the dt. cc @valeriabarra 

## Content

----


valeriabarra:
Interesting. Thanks!

PR #1626, opened 2023-05-01
remove hyperdiffusion of grid-scale w

szy21:
## Purpose 

Remove hyperdiffusion of grid-scale vertical velocity as we are not sure if it is correct, especially when there is topography. Fow now we keep only the hyperdiffusion of subgrid-scale vertical velocity.

This partially reverts #1605 (reverts the change to grid-scale w but keeps the change to subgrid-scale w).

## Content

- Remove hyperdiffusion of grid-scale vertical velocity @akshaysridhar 
- Update docs
- Change back the dt in some SSP longrun jobs @valeriabarra  

----


valeriabarra:
Thanks! Is this PR exactly the reverse of #1605 ? If so, it might be good to mention that in the PR description, for future reference.

szy21:
> Thanks! Is this PR exactly the reverse of #1605 ? If so, it might be good to mention that in the PR description, for future reference.

Not exactly, it reverses part of #1605. I will add it to the description.

jiahe23:
;)

jiahe23 (approved):

regression_tests/ref_counter.jl
```diff
@@ -1 +1 @@
-99
+100
```
trontrytel, line 1: 🥇

glwagner: 💯

szy21: 😀

PR #1630, opened 2023-05-02
Fix default_args, separate argparsesettings

### LINKED ISSUE #1627: Fix default_args issue for the coupler
The toml_dict/parsed_args merge currently assumes that all of the keys in the parsed_args are found in the default_args. This is not the case when the atmosphere model is run via the coupler, which has its own additional arguments.

nefrathenrici:
Closes #1627


valeriabarra:
Thanks Nat!

valeriabarra (approved):

valeriabarra:
Just need to propagate the change in a few instances to make CI happy :)

valeriabarra:
I think we only need to add the same fix to the `perf/` files. We're close!

PR #1462, opened 2023-03-16
Add 3d test for orographic gravity wave

jiahe23:
This PR adds a test for orographic gravity wave, using GFDL outputs.

closes #1429

PR #1632, opened 2023-05-02
Fix race condition in ogwd tests

charleskawczynski:
This PR fixes a race condition, introduced in #1462, where the same filename is written to the same directory by different buildkite jobs. Here is an example failure [here](https://buildkite.com/clima/climaatmos-ci/builds/9401#0187de22-2d33-49df-913e-dafe25ef84e7).

### DIFF
--- a/test/orographic_gravity_wave/ogwd_3d.jl
+++ b/test/orographic_gravity_wave/ogwd_3d.jl
@@ -74,7 +74,6 @@ end
 p_center = 0.5 * (p_half[:, :, 1:(end - 1), :] .+ p_half[:, :, 2:end, :])
 
 # earth warp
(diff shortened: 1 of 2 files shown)


test/orographic_gravity_wave/ogwd_3d.jl
```diff
@@ -74,7 +74,6 @@ end
 p_center = 0.5 * (p_half[:, :, 1:(end - 1), :] .+ p_half[:, :, 2:end, :])
 
 # earth warp
-include(joinpath(pkgdir(ClimaAtmos), "artifacts", "artifact_funcs.jl"))
```
charleskawczynski, line 77: Note: This file was already included above

szy21 (approved):
Thanks Charlie!

PR #1633, opened 2023-05-03
OGW: compute hmax from elevations

jiahe23:
We've decided to split #1569 into two separate task: 1) to compute hmax, 2) to compute T tensor.

This PR adds the computation for hmax (used in orographic gravity wave) from elevation (right now it is read from gfdl restart file).

Please be noted that this PR only won't enable the feature of directly computing OGW related topo info from elevation, but I included the code structure changes and will test it once the second PR (T tensor calculation) is ready. I have left a note in the ogw parameterization setup as well.

The hmax calculation has been tested thoroughly.


jiahe23:
hr_hmax_4min_100e3
Comparison of hmax from the gfdl restart (upper) and our calculation from a 4min resolution elevation (bottom).

jiahe23:
coarse_hmax_100e3
Comparison of hmax from the gfdl restart (upper) and our calculation from the elevation used in current Earth wrap setup(bottom).

regression_tests/mse_tables.jl
szy21, line 1: Why do we need to change this file?

src/parameterizations/gravity_wave_drag/orographic_gravity_wave_helper.jl
```diff
@@ -1,5 +1,57 @@
...
+    ilon_range =
+        min.(
+            Int.(round.(scale ./ (deg2rad(dlon) .* cosd.(lat)))),
+            Int(round(length(lon) / 16)),
+        )
+
+    hmax = zeros(size(elev))
+    for (i, ilon) in enumerate(lon)
+        if i % 10 == 1
+            println("i = $i, and ilon = $ilon")
```
szy21, line 36: Is this print intentional?

src/utils/cli_options.jl
```diff
@@ -332,9 +332,8 @@ function argparse_settings()
         arg_type = Bool
         default = false
         "--orographic_gravity_wave"
-        help = "Apply parameterization for orographic drag on horizontal mean flow"
-        arg_type = Bool
-        default = false
+        help = "Orographic drag on horizontal mean flow [`nothing` (default), `gfdl_restart`, `elevation`]"
```
szy21, line 335: `raw_topo` may be a better name but up to you:)

szy21 (approved):
Thanks for wrapping this up!

src/parameterizations/gravity_wave_drag/orographic_gravity_wave_helper.jl
jiahe23: oops.. for debugging purposes only. good catch!

regression_tests/mse_tables.jl
jiahe23: A buildkite failed due to mse... but I don't know why when I copy it doesn't change anything... Let me double check

PR #1636, opened 2023-05-03
Rename w to u_3

szy21:
## Purpose 

Rename w to u_3 except for the old EDMF

## To-do

- [ ] update mse table

## Content

----


simonbyrne:
Could we call it `u_v` so it matches the horizontal (`u_h`)?

simonbyrne:
in the docs, i distinguish between them: $u_3$ is the 3rd covariant component, $\boldsymbol{u}_v = \boldsymbol{e}^3 u_3$ is a vector

szy21:
> Could we call it `u_v` so it matches the horizontal (`u_h`)?

We talked about this. Some people don't like `u_v` because v is usually for meridional velocity. We could consider renaming `u_h` to `u_12` (in a separate PR) though.

simonbyrne:
that's fair. We could be more descriptive (`u_vert` / `u_horz`)?

simonbyrne:
but I'm fine with this change: it is an improvement on `w`.

charleskawczynski (approved):
The diff looks good to me

trontrytel (approved):

PR #1640, opened 2023-05-04
fix mse table

szy21:
## Purpose 

Fix a bug in the mse table introduced in #1636.

## Content

----


szy21:
@charleskawczynski The previous PR (1636) got merged in with the bug because the ref counter is increased, so the mse table is not checked? Would there be a way to prevent this from happening?

charleskawczynski:
> @charleskawczynski The previous PR (1636) got merged in with the bug because the ref counter is increased, so the mse table is not checked? Would there be a way to prevent this from happening?

Yeah, I think we can do that

PR #1642, opened 2023-05-04
Ensure mse key entries are correct

charleskawczynski:
This PR adds a test to ensure that the mse table entries are correct. Suggested by @szy21. This should help prevent CI from breaking after ref counters are incremented.

### DIFF
--- a/examples/hybrid/driver.jl
+++ b/examples/hybrid/driver.jl
@@ -216,6 +216,7 @@ if !CA.is_distributed(comms_ctx) &&
     end
 end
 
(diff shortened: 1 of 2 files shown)


examples/hybrid/driver.jl
```diff
@@ -235,6 +236,16 @@ if parsed_args["regression_test"]
     )
 end
 
+@testset "Test regression table entries" begin
```
szy21, line 239: Should this be in `if parsed_args["regression_test"]`

szy21 (approved):
Thanks Charlie!

PR #1643, opened 2023-05-04
Upgrade to latest ClimaCore, ClimaComms and ClimaCoreTempestRemap

charleskawczynski:
This PR upgrades to the latest ClimaCore, ClimaComms and ClimaCoreTempestRemap.

There are some slowdowns with performance regressions in ClimaCore (see https://github.com/CliMA/ClimaCore.jl/issues/1223 for example), but those should be fixable with patch releases.


akshaysridhar:
I think this helps with some of the `helpdesk` issues ; but we have a working solution in the meantime through manual updates of the compat entries and calls to `ClimaComms` functions as opposed to `ClimaCommsMPI`. Thanks!

szy21 (approved):

src/utils/type_getters.jl
```diff
@@ -627,3 +627,31 @@ function args_integrator(parsed_args, Y, p, tspan, ode_algo, callback)
...
+        Logging.global_logger(Logging.NullLogger())
+    end
+    if comms_ctx isa ClimaComms.SingletonCommsContext
+        @info "Setting up single-process ClimaAtmos run"
+    else
+        @info "Setting up distributed ClimaAtmos run" nprocs =
+            ClimaComms.nprocs(comms_ctx)
+    end
+    if NVTX.isactive()
+        NVTX.enable_gc_hooks()
```
charleskawczynski, line 647: Delete

charleskawczynski: (suggested by @simonbyrne)

PR #1645, opened 2023-05-04
use cached sfc_conditions

trontrytel:
Fix to [ review comment](https://github.com/CliMA/ClimaAtmos.jl/pull/1613#discussion_r1185540107)

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -415,14 +415,14 @@ steps:
 
       - label: ":genie: explicit baroclinic wave (ρe) equilmoist with EDMFX"
         command:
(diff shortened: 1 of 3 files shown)


charleskawczynski (approved):
Did you intend to make the pipeline changes? Changes in src LGTM

trontrytel:
> Did you intend to make the pipeline changes? Changes in src LGTM

Yeah. In the previous iteration I had to ensure that the `vertical_diffusion_boundary_layer` cache was populated. So I was switching it on, but at the same time setting the coefficient to zero. 

But `sfc_conditions` live happily in the `default_cache` (thats why I did not notice we cache them in the first place) so I don't need to rely on vertical diffusion to do anything for me

PR #1208, opened 2022-12-16
Use Thermodynamics.jl's T_guess

charleskawczynski:
This PR adds the use of Thermodynamics.jl's T_guess for specifying the initial guess for saturation adjustment.

PR #1648, opened 2023-05-05
Set thermo state maximum iteration

szy21:
## Purpose 

Set thermo state iteration to 1 and use previous temperature as initial guess
Supersedes #1208 #785 

## To-do

- [x] Fix allocation
- [x] Add T_guess
- [x] See if we can change iteration to 1 (currently it is set to 2) and decrease tolerance to default 1e-4 (currently it is set to 1e-3)
- [ ] Check targeted resolution longruns

## Content

----

### DIFF
--- a/src/cache/precomputed_quantities.jl
+++ b/src/cache/precomputed_quantities.jl
@@ -1,6 +1,8 @@
 #####
 ##### Precomputed quantities
 #####
+
+import RootSolvers as RS
 import Thermodynamics as TD
 import ClimaCore: Spaces, Fields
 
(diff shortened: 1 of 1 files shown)


szy21:
It seems maxiter=2 and tol=1e-3 work except for moist baroclinic wave without precipitation (`sphere_baroclinic_wave_rhoe_equilmoist_conservation_ft64`), which is not a very physical test case. See this [build](https://buildkite.com/clima/climaatmos-ci/builds/9536).

src/precomputed_quantities.jl
```diff
@@ -147,6 +147,7 @@ function thermo_state(
     q_tot = nothing,
     q_pt = nothing,
 )
+    FT = eltype(thermo_params)
```
szy21, line 150: I guess the allocation is from the FT here? Maybe I can just inline it? @charleskawczynski

charleskawczynski: Yes I think so

charleskawczynski: Hopefully that'll work, otherwise we may just need to revert this and remove the closures?

szy21: it works:)

szy21:
maxiter=2 works for low resolution simulations but not for high resolution ones, see this [build](https://buildkite.com/clima/climaatmos-longruns/builds/952#_) for example. We probably need the T_guess.

bischtob:
bors try

bischtob:
Taking this on for now. I have refactored our dispatches for getting the thermo state. @szy21 and I added another option so that T_guess is used now. We'll check the max_iter 1 question.

bischtob:
bors try

bischtob:
CI crashes with max_iter = 1. Trying max_iter = 2.

bischtob:
bors try

szy21:
@charleskawczynski Any idea which part of this PR increases allocation?

src/cache/precomputed_quantities.jl
```diff
@@ -288,7 +413,16 @@ function set_precomputed_quantities!(Y, p, t)
...
+    @. ᶜts =
+        t > 0 ?
+        ts_gs(
+            thermo_args...,
+            ᶜspecific,
+            ᶜK,
+            ᶜΦ,
+            Y.c.ρ,
+            nothing,
+        ) : ts_gs(thermo_args..., ᶜspecific, ᶜK, ᶜΦ, Y.c.ρ)
```
charleskawczynski, line 425: Maybe it's this? I would change this to:
```julia
    if t > 0
    @. ᶜts =
        ts_gs(
            thermo_args...,
            ᶜspecific,
            ᶜK,
            ᶜΦ,
            Y.c.ρ,
            nothing,
        )
else
    @. ᶜts = ts_gs(thermo_args..., ᶜspecific, ᶜK, ᶜΦ, Y.c.ρ)
end
```
so that the runtime if-else block is not inside the broadcast expression.

Sbozzolo:
Is this PR still relevant, or can it be closed?

szy21:
The problem hasn’t been solved, but this PR is probably outdated. I’ll close it.

PR #1653, opened 2023-05-08
Pass comms_ctx through to Topology

nefrathenrici:
Since I have a CUDA GPU, I got an error when running some cases locally. The comms context isn't properly passed through, so it gets reset to the default value in the topology constructor. This causes incompatible allocations on the GPU and CPU.
As @charleskawczynski has already suggested elsewhere, this could be fixed each time by adding the following snippet when running locally, so I don't know if this is necessary to add, but it would make life easier for local testing.
```
import RRTMGP
RRTMGP.Device.array_device(::Union{Array}) = RRTMGP.Device.CPU()
RRTMGP.Device.array_type() = Array
```

### DIFF
--- a/src/utils/common_spaces.jl
+++ b/src/utils/common_spaces.jl
@@ -34,15 +34,18 @@ end
 function make_horizontal_space(
     mesh,
     quad,
-    ::ClimaComms.SingletonCommsContext,
+    comms_ctx::ClimaComms.SingletonCommsContext,
     bubble,
 )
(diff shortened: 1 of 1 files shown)


charleskawczynski (commented):
Ah, I meant to add this in #1643, I must not have applied everything from my dev branch. Thanks! Can you also propagate `comms_ctx` through in `make_hybrid_spaces` below, too?

charleskawczynski (approved):

charleskawczynski (approved):

PR #1660, opened 2023-05-11
Fix topography bugs in implicit solver

### LINKED ISSUE #1670: Fix bug in the implicit solver Jacobian

dennisYatunin:
## Purpose 

This PR fixes bugs in the `wfact!` function, both of which only become relevant when using topography. It also removes some unused code from `wfact!`, updates the code comments, and makes some other minor cosmetic changes.

Closes #1670 
## Content

- Take the correct derivative of `vertical_transport!` with respect to `u₃_data`, accounting for the fact that `∂(u³_data)/∂(u₃_data) = g³³`.
- Take the correct derivative of kinetic energy with respect to `u₃_data`, accounting for the derivative of the `dot(CT123(ᶜuₕ), ᶜinterp(C123(ᶠu₃)))` cross term.
- Remove the `∂ᶠ𝕄ₜ∂ᶜρ_mode` Jacobian flag, since have never used Oswald's `gradΦ_shenanigans` approximation.
- Format the `Wfact!` function so that it looks more similar to the `implicit_tendency!` function, and make sure that it uses the type-stable `matching_subfields` function.
- Update all of the Jacobian block derivations in the code comments to reflect the current implicit tendency. Replace `w` with `u₃` in all code comments. Rename the variable `∂ᶜK∂ᶠw_data` to `∂ᶜK∂ᶠu₃_data`, and rename the function `vertical_transport_jac!` to `set_∂ᶜρχₜ∂ᶠu₃!`.
- Decrease the `dt`s for two of the SSP longruns because they appear to be crashing due to (hopefully) unrelated numerical issues.

----


szy21:
I wonder if we should separate the bugfix from adding h_tot to precomputed_quantities? I understand that we want it, but this makes the PR longer, while these two items are not really related. I think it would be more clear if this PR is only about the bugfix and removing unused Jacobians.

src/tendencies/implicit/wfact.jl
```diff
@@ -7,73 +7,52 @@ import ClimaCore.Spaces as Spaces
...
-    @. ∂ᶜρcₜ∂ᶠu₃ = -(ᶜadvdivᵥ_stencil(
-        ᶠwinterp(ᶜJ, ᶜρ) * ᶠupwind1(C3(magnitude_plus_eps(ᶠu₃)), ᶜρc / ᶜρ) /
-        magnitude_plus_eps(ᶠu₃),
+# For simplicity, we approximate the Jacobian for FCT using the Jacobian for
+# third-order upwinding.
+# In the vertical_transport_jac! function, we assume that ∂(ᶜχ)/∂(ᶠu₃_data) = 0;
+# if this is not the case, the additional term should be added to the result of
+# this function.
+get_data_plus_ε(vector) = vector.u³ + eps(vector.u³)
+vertical_transport_jac!(∂ᶜ𝕋ₜ∂ᶠ𝕄, ᶜJ, ᶜρ, ᶠu³, ᶜχ, ᶠg³³, ::Val{:none}) =
```
szy21, line 33: Why is `∂ᶜρcₜ∂ᶠu₃` renamed to `∂ᶜ𝕋ₜ∂ᶠ𝕄`?

.buildkite/longruns/pipeline.yml
```diff
@@ -67,7 +67,7 @@ steps:
 
       - label: ":computer: SSP baroclinic wave (ρe_tot) equilmoist high resolution centered diff"
         command:
-          - "mpiexec julia --project=examples examples/hybrid/driver.jl --moist equil --precip_model 0M --z_elem 45 --dz_bottom 30 --h_elem 16 --kappa_4 1e16 --dt 300secs --max_newton_iters 3 --initial_condition MoistBaroclinicWave --t_end 100days --dt_save_to_disk 10days --job_id longrun_ssp_bw_rhoe_equil_highres --ode_algo SSP333"
+          - "mpiexec julia --project=examples examples/hybrid/driver.jl --moist equil --precip_model 0M --z_elem 45 --dz_bottom 30 --h_elem 16 --kappa_4 1e16 --dt 250secs --max_newton_iters 3 --initial_condition MoistBaroclinicWave --t_end 100days --dt_save_to_disk 10days --job_id longrun_ssp_bw_rhoe_equil_highres --ode_algo SSP333"
```
trontrytel, line 70: So the result of fixing the bug is that we have to run with smaller dt? Do we understand why?

src/tendencies/implicit/wfact.jl
```diff
@@ -7,73 +7,52 @@ import ClimaCore.Spaces as Spaces
...
-#     -ᶜdivᵥ_stencil(ᶠwinterp(ᶜJ, ᶜρ) * ᶠw_unit * ᶠinterp(ᶜc)) -
-#     ᶜdivᵥ_stencil(ᶠu₃) * ᶠinterp_stencil(1) * ∂(ᶜρc)/∂(ᶠw_data)
-# If ᶜρcₜ = -ᶜdivᵥ(ᶠwinterp(ᶜJ, ᶜρ) * ᶠupwind(ᶠu₃, ᶜρc / ᶜρ)), then
-# ∂(ᶜρcₜ)/∂(ᶠw_data) =
-#     -ᶜdivᵥ_stencil(ᶠwinterp(ᶜJ, ᶜρ) *
-#     ᶠupwind(ᶠu₃ + εw, ᶜρc / ᶜρ) / to_scalar(ᶠu₃ + εw)) -
-#     ᶜdivᵥ_stencil(ᶠinterp(ᶜρ)) * ᶠupwind_stencil(ᶠu₃, 1 / ᶜρ) *
-#     ∂(ᶜρc)/∂(ᶠw_data)
-# The εw is only necessary in case w = 0.
+# From the chain rule, we know that
```
trontrytel, line 10: Is there any plan to move those comments into proper documentation? Or do people prefer to have them here?

dennisYatunin: Yes, these will all turn into a proper documentation page during the implicit solver refactoring.

akshaysridhar:
CI Checks (additional checks include those with pressure reference states in Schar mountain test cases): 
- https://buildkite.com/clima/climaatmos-ci/builds/9729 

Long-run Build (Dry Baroclinic Wave day100, dt400s) 
- Result off main: [Summary_Level1_DBW_dt400_d100_Pre1660.pdf](https://github.com/CliMA/ClimaAtmos.jl/files/11500337/Summary_Level1_DBW_dt400_d100_Pre1660.pdf)
- Result off dy/wfact_fixes[Summary_Level1_DBW_dt400_d100_Post1660.pdf](https://github.com/CliMA/ClimaAtmos.jl/files/11500343/Summary_Level1_DBW_dt400_d100_Post1660.pdf)

Based on this I think this PR is ready to be merged cc @Zhengyu-Huang

akshaysridhar (approved):

PR #1675, opened 2023-05-16
Update to CloudMicrophysics v0.11

trontrytel:


nefrathenrici (approved):

charleskawczynski:
We need to apply the `.dev/up_deps.jl` script for this to have any effect.

charleskawczynski:
Also, that won't work because the latest version of Krylov.jl is breaking due to an issue I'm now digging into. This PR will have to wait

trontrytel:
I think I'm on the slightly older version of julia than the one here. I'll have to update mine before I run the script

trontrytel:
ok - I will wait. Thanks!

trontrytel:
Done via https://github.com/CliMA/ClimaAtmos.jl/pull/2047

PR #1689, opened 2023-05-19
Add vertical integral for diagnostic EDMFX

### LINKED ISSUE #1655: Add diagnostic EDMFX type and vertical integral

szy21:
## Purpose 

Adds vertical integral for diagnostic EDMF updraft variables.
co-authored-by: @dennisYatunin 
Supersedes #1656 
Closes #1655 

## Content

----

### DIFF
--- a/src/callbacks/callbacks.jl
+++ b/src/callbacks/callbacks.jl
@@ -320,7 +320,7 @@ function save_to_disk_func(integrator)
 
     if p.atmos.turbconv_model isa EDMFX
         (; ᶜspecific⁰, ᶜu⁰, ᶜts⁰) = p
-        (; ᶜu⁺, ᶜts⁺, ᶜa⁺, ᶜa⁰) = diagnostic_sgs_quantities(Y, p, t)
+        (; ᶜu⁺, ᶜts⁺, ᶜa⁺, ᶜa⁰) = output_sgs_quantities(Y, p, t)
         env_diagnostics = (;
             common_diagnostics(ᶜu⁰, ᶜts⁰)...,
             area = ᶜa⁰,
@@ -337,6 +337,17 @@ function save_to_disk_func(integrator)
             add_prefix(draft_diagnostics, :draft_)...,
             cloud_fraction = ᶜa⁰ .* cloud_fraction.(ᶜts⁰, ᶜa⁰) .+
                              ᶜa⁺ .* cloud_fraction.(ᶜts⁺, ᶜa⁺),
+        )
+    elseif p.atmos.turbconv_model isa DiagnosticEDMFX
+        (; ᶜu⁺, ᶜts⁺, ᶜa⁺) = output_diagnostic_sgs_quantities(Y, p, t)
+        draft_diagnostics = (;
+            common_diagnostics(ᶜu⁺, ᶜts⁺)...,
+            area = ᶜa⁺,
+            cloud_fraction = cloud_fraction.(ᶜts⁺, ᶜa⁺),
+        )
+        turbulence_convection_diagnostic = (;
+            add_prefix(draft_diagnostics, :draft_)...,
+            cloud_fraction = ᶜa⁺ .* cloud_fraction.(ᶜts⁺, ᶜa⁺),
         )
     elseif p.atmos.turbconv_model isa TC.EDMFModel
         tc_cent(p) = p.edmf_cache.aux.cent.turbconv
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -500,6 +500,19 @@ steps:
         artifact_paths: "edmfx_adv_test_sphere/*"
         agents:
           slurm_mem: 20GB
+      
+      - label: ":genie: Diagnostic EDMFX test in a box"
+        command: >
+          julia --color=yes --project=examples examples/hybrid/driver.jl --job_id diagnostic_edmfx_test_box
+          --moist equil --turbconv diagnostic_edmfx --edmfx_entr_detr false
+          --config box --hyperdiff true --kappa_4 1e8
+          --x_max 1e4 --y_max 1e4 --x_elem 2 --y_elem 2 --z_elem 45 --z_max 30e3 --dz_bottom 30
+          --dt 1secs --t_end 100secs
+          --dt_save_to_disk 10secs
+
+        artifact_paths: "diagnostic_edmfx_test_box/*"
+        agents:
+          slurm_mem: 20GB
 
 
   - group: "TurbulenceConvection"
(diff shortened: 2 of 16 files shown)


trontrytel:
I think there is something wrong with the initial condition in the Bomex box setup when we switch on the EMDFX. I'm seeing a similar issue with the prognostic edmf. - I did not have the time to find it yet... soon...

src/callbacks/callbacks.jl
```diff
@@ -320,7 +320,7 @@ function save_to_disk_func(integrator)
 
     if p.atmos.turbconv_model isa EDMFX
         (; ᶜspecific⁰, ᶜu⁰, ᶜts⁰) = p
-        (; ᶜu⁺, ᶜts⁺, ᶜa⁺, ᶜa⁰) = diagnostic_sgs_quantities(Y, p, t)
+        (; ᶜu⁺, ᶜts⁺, ᶜa⁺, ᶜa⁰) = output_sgs_quantities(Y, p, t)
```
trontrytel, line 323: I like `output_sgs_quantities`!

trontrytel: Also, could you remind me what the \^+ is symbolizing here?

szy21: It's for bulk draft properties (which I am setting to the first and only updraft for diagnostic edmf:)

src/precomputed_quantities.jl
```diff
@@ -319,18 +370,236 @@ function set_precomputed_quantities!(Y, p, t)
...
+
+                @. ρaʲu³ʲ_datau³ʲ_data =
+                    (1 / local_geometry_halflevel.J^2) * (
+                        local_geometry_prev_halflevel.J^2 *
+                        ρaʲ_prev_level *
+                        u³ʲ_data_prev_halflevel *
+                        u³ʲ_data_prev_halflevel
+                    )
+
+                # @. ρaʲu³ʲ_datau³ʲ_data -=
```
szy21, line 508: (This is for testing. I will add it back.)

src/precomputed_quantities.jl
```diff
@@ -319,18 +370,239 @@ function set_precomputed_quantities!(Y, p, t)
...
+                    Fields.level(Fields.local_geometry_field(Y.c), i - 1),
+                )
+                local_geometry_prev_halflevel = Fields.field_values(
+                    Fields.level(
+                        Fields.local_geometry_field(Y.f),
+                        i - 1 - half,
+                    ),
+                )
+
+                ρaʲu³ʲ_data = p.temp_data_level
```
szy21, line 493: Try moving this outside the for loop.

src/precomputed_quantities.jl
```diff
@@ -319,18 +370,238 @@ function set_precomputed_quantities!(Y, p, t)
...
+
+            local_geometry_level = Fields.field_values(
+                Fields.level(Fields.local_geometry_field(Y.c), 1),
+            )
+            local_geometry_halflevel = Fields.field_values(
+                Fields.level(Fields.local_geometry_field(Y.f), half),
+            )
+
+            @. u³ʲ_halflevel = CT3(FT(0))
+            @. ρaʲ_level = ρ_level * FT(0.1)
```
trontrytel, line 423: Would it be at all useful to have `0.1` as parameter and not hard coded?

src/precomputed_quantities.jl
```diff
@@ -319,18 +370,238 @@ function set_precomputed_quantities!(Y, p, t)
...
+            ᶜρaʲ = ᶜρaʲs.:($j)
+            ᶠu³ʲ = ᶠu³ʲs.:($j)
+            ᶜKʲ = ᶜKʲs.:($j)
+            ᶜh_totʲ = ᶜh_totʲs.:($j)
+            ᶜtsʲ = ᶜtsʲs.:($j)
+            ᶜρʲ = ᶜρʲs.:($j)
+            ᶜq_totʲ = ᶜq_totʲs.:($j)
+
+            # boundary condition
+            ρaʲ_level = Fields.field_values(Fields.level(ᶜρaʲ, 1))
```
trontrytel, line 400: Would it be useful to keep the `\^c` notation for the `level` variables that are defined here and below?

src/precomputed_quantities.jl
```diff
@@ -319,18 +370,238 @@ function set_precomputed_quantities!(Y, p, t)
...
+                    q_totʲ_level,
+                    p_level,
+                    Φ_level,
+                )
+            end
+
+            u³ʲ_halflevel = Fields.field_values(
+                Fields.level(ᶠu³ʲ, Spaces.nlevels(axes(Y.c)) + half),
+            )
+            @. u³ʲ_halflevel = CT3(FT(0))
```
trontrytel, line 589: Why does it have to be set to zero here?

src/precomputed_quantities.jl
```diff
@@ -319,18 +370,238 @@ function set_precomputed_quantities!(Y, p, t)
...
+                @. ρaʲu³ʲ_datau³ʲ_data =
+                    (1 / local_geometry_halflevel.J^2) * (
+                        local_geometry_prev_halflevel.J^2 *
+                        ρaʲ_prev_level *
+                        u³ʲ_data_prev_halflevel *
+                        u³ʲ_data_prev_halflevel
+                    )
+
+                @. ρaʲu³ʲ_datau³ʲ_data -=
+                    (1 / local_geometry_halflevel.J^2) * (
```
trontrytel, line 513: I was trying to follow along with the overleaf equations. It would be great to add some documentation to ClimaAtmos that explains the operators. For example, I got confused about where the `g^3l` term is?

trontrytel (approved):
Amazing work!

src/precomputed_quantities.jl
szy21: For the integral, we solve for rhoa[i] and u^3[i-half] at each i. But u^3 has one more level than rhoa, and u^3 at the last level (top cell face) is not defined in the integral, so we set it to zero.

src/precomputed_quantities.jl
szy21: Yes! I will add some documentation next week. The `g^3l` term is in the project e.g. [here](https://github.com/CliMA/ClimaAtmos.jl/blob/a6ed6d10ea53beb91ae56482ca3790d0a60b12e0/src/cache/precomputed_quantities.jl#L381).

src/precomputed_quantities.jl
szy21: I think so. I put it [here](https://github.com/CliMA/ClimaAtmos.jl/blob/a6ed6d10ea53beb91ae56482ca3790d0a60b12e0/src/solver/types.jl#L106) for now. I think the idea is we want to set this to zero when we have entrainment? But actually, I'm confused about how we can get updraft when looking at the equations. If we need this number in the end, we can add it to ClimaParameters.

src/precomputed_quantities.jl
szy21: I thought about it too. I decided not to keep it in the end, as the `level` variables are `DataLayouts` instead of `Field`( because we take `field_values`), and `\^c` notation is for `Field`.

PR #1692, opened 2023-05-19
Fix exchange coefficient expression in vertical diffusion tendency. 

akshaysridhar:
Remove specialised use of z_interior (Reasonable for "flat" configurations, unreasonable otherwise).
z_{interior} in the expression for exchange coefficients is appropriate if the surface elevation is zero. 
The general expression for bulk formula coefficients should be evaluated with the local `dz` taken into account.


szy21 (approved):
Thanks Akshay!

PR #1698, opened 2023-05-22
Add CUDA with compat, update deps

charleskawczynski:
This PR adds CUDA with a compat and updates the dependencies. This is a peel off from #1693.


szy21 (approved):

Project.toml
```diff
@@ -54,6 +55,7 @@ ArgParse = "1"
 ArtifactWrappers = "0.2"
 AtmosphericProfilesLibrary = "0.1"
 CLIMAParameters = "0.7"
+CUDA = "4"
```
valeriabarra, line 58: If this PR adds/updates the `compat` entry, should we also bump another ClimaAtmos release? (I'm thinking for the downstream update, e.g., the Coupler)

charleskawczynski: This PR's scope is related to enabling GPU jobs. We can bump the patch release in a separate PR.

PR #1702, opened 2023-05-22
Fix conditional_assign! allocations

nefrathenrici:
Reduce allocations by specializing the `conditional_assign!` function, used in the old EDMF.
See [here](https://docs.julialang.org/en/v1/manual/performance-tips/#Be-aware-of-when-Julia-avoids-specializing) - 
`g_func(g::Function, num) = ntuple(g, div(num, 2))` will not specialize, but 
`h_func(h::H, num) where {H} = ntuple(h, div(num, 2))` does.

Thanks @simonbyrne


dennisYatunin (approved):

charleskawczynski (approved):

charleskawczynski:
Just to note, this used to be inferred on older versions of ClimaCore (it might be a good to add a ClimaCore test at some point)

PR #1713, opened 2023-05-25
Refactor driver for performance benchmark

charleskawczynski:
This PR refactors the driver a bit for setting up simple performance benchmarks.

Here's the docs for the workflow we're trying to achieve:

    benchmark_step!(integrator)

#### Example

To instantiate the environment:
```
julia --project=examples
```

Then, to run interactively:
```julia
import ClimaAtmos as CA
config = CA.AtmosPerfConfig();
integrator = CA.get_integrator(config);
Y₀ = deepcopy(integrator.u);
CA.benchmark_step!(integrator, Y₀);
```

Alternatively, this can be run non-interactively,
with adjusted flags as:
```
julia --project=examples perf/benchmark_step.jl --h_elem 6
```

See `argparse_settings` for the
method defining the flags, and

https://clima.github.io/ClimaAtmos.jl/dev/cl_args/

for the flags outlined in a table.


simonbyrne (approved):

charleskawczynski:
I reviewed all of the failing jobs, and they all appear to be due to numerical roundoff error. RRTMGP is not reproducible w.r.t. its random number generation (https://github.com/CliMA/RRTMGP.jl/issues/316#issuecomment-1565144755). So, it's possible that these roundoff errors are due to moving `import Random; Random.seed!(1234)`.

szy21:
> CliMA/RRTMGP.jl#316 (comment)

Just want to note that this line should only have an effect when we run `allsky` radiation, so I don't think all the roundoff errors are due to moving the random number generation.

charleskawczynski:
> Just want to note that this line should only have an effect when we run `allsky` radiation, so I don't think all the roundoff errors are due to moving the random number generation.

Yeah, that’s a good point. Not all of the non-determinism is coming from RRTMGP, but we can expect to see non-determinism from it. I think https://github.com/CliMA/ClimaAtmos.jl/pull/1720 shows that there may be other sources of non-determinism, or at least use of random numbers— it’s not apparent to me where though.

PR #1714, opened 2023-05-25
Skip dss with empty fields

charleskawczynski:
Fixes issue we're running into with https://github.com/CliMA/ClimaCore.jl/issues/1265

### DIFF
--- a/src/prognostic_equations/hyperdiffusion.jl
+++ b/src/prognostic_equations/hyperdiffusion.jl
@@ -38,6 +38,13 @@ function hyperdiffusion_cache(Y, atmos, do_dss)
     return quantities
 end
 
+maybe_weighted_dss_start!(x, y) =
+    isempty(propertynames(x)) ? nothing : Spaces.weighted_dss_start!(x, y)
+maybe_weighted_dss_internal!(x, y) =
+    isempty(propertynames(x)) ? nothing : Spaces.weighted_dss_internal!(x, y)
+maybe_weighted_dss_ghost!(x, y) =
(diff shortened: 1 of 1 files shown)


sriharshakandala:
I think its best to have the exception handling code higher up rather at the location of the `weighted_dss` call. 
As discussed in https://github.com/CliMA/ClimaCore.jl/issues/1265 , there is no reason to call `tracer_hyperdiffusion_tendency!` when there are no tracers!

charleskawczynski:
@dennisYatunin, any idea why this is resulting in behavioral changes for some jobs? cc @szy21, @simonbyrne

szy21:
Hmm, the results actually look very strange.

szy21:
Ah, I think I might know. For some quantities like `ᶜ∇²specific_tracers` (which is a NamedTuple), it should be `isempty(propertynames(x))`, but for some other quantities like `ᶜ∇²uₕ`, it should be just `isempty(x)`?

charleskawczynski:
> Ah, I think I might know. For some quantities like `ᶜ∇²specific_tracers` (which is a NamedTuple), it should be `isempty(propertynames(x))`, but for some other quantities like `ᶜ∇²uₕ`, it should be just `isempty(x)`?

Ok, that makes sense.

I've been thinking about this a bit more, and I'm also wondering what broadcast expressions are doing with empty fields-- I think this is exercising an untested edge case in ClimaCore fields, cc @simonbyrne. Should we add some unit tests to ClimaCore so that we're sure this short circuits to no-ops?

charleskawczynski:
This is fixing an issue that was fixed in the latest ClimaCore update.

PR #1726, opened 2023-06-01
add diagnostic edmfx environment variables

### LINKED ISSUE #1724: Add diagnostic EDMFX environment variables and entrainment

szy21:
## Purpose 

Adds diagnostic edmfx environment variables and entrainment and detrainment. It's a bit long, please see the content for details.
Closes #1724.

## Content

- Most of the changes in `precomputed_quantities` is related to changing the order of the loop. Before we are looping over updrafts first, then levels, but we need to loop over levels first to calculate environment variables from all of the updrafts. Additionally, environment \rhoa and u^3 have been added.
- Entrainment and detrainment tendencies have been added to `precomputed_quantities.jl`. This file is too long now. After this PR I will remove the diagnostic edmfx to a separate file.
- Updates the diagnostic edmfx bomex box test. It works, but I keep it soft fail for now as it is sensitive to entrainment, reference state, etc. I think it will be more robust once we have an idea about how to regularize area fraction and other updraft variables in diagnostic edmfx.

----

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -478,11 +478,11 @@ steps:
         command: >
           julia --color=yes --project=examples examples/hybrid/driver.jl --job_id diagnostic_edmfx_bomex_box
           --initial_condition Bomex --subsidence Bomex --edmf_coriolis Bomex --ls_adv Bomex --surface_setup Bomex
-          --turbconv diagnostic_edmfx
+          --turbconv diagnostic_edmfx --edmfx_entr_detr true
           --moist equil
           --config box --hyperdiff true --kappa_4 1e8
           --x_max 1e5 --y_max 1e5 --x_elem 2 --y_elem 2 --z_elem 60 --z_max 3e3 --z_stretch false
-          --dt 10secs --t_end 1hours --dt_save_to_disk 5mins
+          --dt 10secs --t_end 6hours --dt_save_to_disk 10mins
         artifact_paths: "diagnostic_edmfx_bomex_box/*"
         soft_fail: true
         agents:
--- a/src/cache/precomputed_quantities.jl
+++ b/src/cache/precomputed_quantities.jl
@@ -89,6 +89,12 @@ function precomputed_quantities(Y, atmos)
             ᶜρʲs = similar(Y.c, NTuple{n, FT}),
             ᶜh_totʲs = similar(Y.c, NTuple{n, FT}),
             ᶜq_totʲs = similar(Y.c, NTuple{n, FT}),
+            ᶜentr_detrʲs = similar(
+                Y.c,
+                NTuple{n, NamedTuple{(:entr, :detr), NTuple{2, FT}}},
+            ),
+            ᶜρa⁰ = similar(Y.c, FT),
+            ᶠu³⁰ = similar(Y.f, CT3{FT}),
         ) : (;)
     return (; gs_quantities..., sgs_quantities..., diagnostic_sgs_quantities...)
 end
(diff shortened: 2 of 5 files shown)


.buildkite/pipeline.yml
```diff
@@ -478,15 +478,30 @@ steps:
...
           --config box --hyperdiff true --kappa_4 1e8
           --x_max 1e5 --y_max 1e5 --x_elem 2 --y_elem 2 --z_elem 60 --z_max 3e3 --z_stretch false
-          --dt 10secs --t_end 1hours --dt_save_to_disk 5mins
+          --dt 10secs --t_end 6hours --dt_save_to_disk 10mins
         artifact_paths: "diagnostic_edmfx_bomex_box/*"
         soft_fail: true
         agents:
           slurm_mem: 20GB
+      
+      - label: ":genie: Diagnostic EDMFX Bomex in a box no ref state"
```
szy21, line 491: I will either remove this or replace the current bomex test with this.

src/cache/precomputed_quantities.jl
```diff
@@ -603,25 +749,45 @@ function set_precomputed_quantities!(Y, p, t)
...
+                        local_geometry_prev_level.J^2 *
+                        ρaʲ_prev_level *
+                        ((
                             ∇p_perturb³_prev_level_data / ρʲ_prev_level +
                             ∇Φ³_prev_level_data *
                             (ρʲ_prev_level - ρ_ref_prev_level) / ρʲ_prev_level
-                        )
+                        ))
+                    )
+                @. ρaʲu³ʲ_datau³ʲ_data +=
```
szy21, line 760: This is a hack I have to make updraft in bomex. I will try to remove it (because we don't want it in the end), but there will probably be no updraft in bomex, or it will be sensitive at least.

src/cache/precomputed_quantities.jl
```diff
@@ -488,33 +557,38 @@ function set_precomputed_quantities!(Y, p, t)
...
-            h_tot_level = Fields.field_values(Fields.level(ᶜh_tot, 1))
-            q_tot_level = Fields.field_values(Fields.level(q_tot, 1))
-
-            p_level = Fields.field_values(Fields.level(ᶜp, 1))
-            Φ_level = Fields.field_values(Fields.level(ᶜΦ, 1))
 
-            local_geometry_level = Fields.field_values(
-                Fields.level(Fields.local_geometry_field(Y.c), 1),
+            @. u³ʲ_halflevel = CT3(
+                Geometry.WVector(FT(0.01), local_geometry_halflevel),
```
szy21, line 569: We need something non zero here, but the value doesn't seem to be very important, so I'm hard coding it without using a parameter for now.

src/prognostic_equations/edmfx_closures.jl
```diff
@@ -184,7 +184,7 @@ function pi_groups_entr_detr(
         # TODO - Temporary: Switch to Π groups after simple tests are done
         # (kinematic, bubble, Bomex)
         # and/or we can calibrate things in ClimaAtmos
-        entr = max(0, 1e-1 * ᶜwʲ / ᶜz)
+        entr = max(0, 1 * ᶜwʲ / ᶜz)
```
szy21, line 187: Another hack for bomex. We don't have to change it though, maybe we can just get this PR in and work on making bomex look better after adding other tendencies?

.buildkite/pipeline.yml
szy21: removed

src/cache/precomputed_quantities.jl
szy21: removed.

trontrytel (approved):

PR #1737, opened 2023-06-05
Update dependencies

charleskawczynski:

Unfortunately https://github.com/CliMA/ClimaCore.jl/pull/1198 1) fixed a bug and 2) fixed inference for several cases when calling dss, but https://github.com/CliMA/ClimaCore.jl/pull/1295 demonstrates that inference fails pretty badly for our use case in ClimaAtmos. The result is that we need to increment the allocation limit. I'm going to try and fix the broken test in https://github.com/CliMA/ClimaCore.jl/pull/1295, at which point we should be able to reduce the limit back down. All in all, the allocations are a ClimaCore issue, and not ClimaAtmos related.


examples/Manifest.toml
```diff
@@ -248,10 +248,10 @@ uuid = "3a4d1b5c-c61d-41fd-a00a-5873ba7a1b0d"
 version = "0.4.2"
 
 [[deps.ClimaCore]]
-deps = ["Adapt", "BlockArrays", "CUDA", "ClimaComms", "CubedSphere", "DataStructures", "DiffEqBase", "DocStringExtensions", "ForwardDiff", "GaussQuadrature", "GilbertCurves", "HDF5", "InteractiveUtils", "IntervalSets", "LinearAlgebra", "PkgVersion", "RecursiveArrayTools", "RootSolvers", "Rotations", "SparseArrays", "Static", "StaticArrays", "Statistics", "UnPack"]
-git-tree-sha1 = "9ef8838ca6cab5c79c090854c7159f3aee09d433"
+deps = ["Adapt", "BlockArrays", "CUDA", "ClimaComms", "CubedSphere", "DataStructures", "DiffEqBase", "DocStringExtensions", "ForwardDiff", "GaussQuadrature", "GilbertCurves", "HDF5", "InteractiveUtils", "IntervalSets", "LinearAlgebra", "PkgVersion", "RecursiveArrayTools", "Requires", "RootSolvers", "Rotations", "SparseArrays", "Static", "StaticArrays", "Statistics", "UnPack"]
+git-tree-sha1 = "f253815e33a8437e62cadbfcb24e3e713a6f8dd4"
 uuid = "d414da3d-4745-48bb-8d80-42e94e092884"
-version = "0.10.36"
```
akshaysridhar, line 254: can we bump this to v0.10.38 (if we want to do this in 2 steps that is also OK)

akshaysridhar (approved):

examples/Manifest.toml
charleskawczynski: It was bumped to 0.10.38. I just force pushed, but your comment isn't stale, maybe due to funny timing? Anyway, I think this is what you want!

akshaysridhar:
Yup LGTM! Indeed the v0.38 comment was perhaps a few seconds before your force push update.

PR #1748, opened 2023-06-09
Use addition assignment in vertical transport

szy21:
## Purpose 

Uses addition assignment (+=) in vertical transport. We sometimes move around the vertical advection functions so I think this would be safer. We already zero out the tendencies at the beginning of the implicit and remaining tendencies, so this shouldn't cause any change.

## Content

----


charleskawczynski (approved):
I’m fine with these changes

PR #1754, opened 2023-06-11
Update manifest files

charleskawczynski:
This PR updates the dependencies, including updates with improved inference (https://github.com/CliMA/ClimaCore.jl/pull/1306)


charleskawczynski:
😞 I just can't reproduce the ClimaAtmos JET failure in ClimaCore. I guess I can try and develop ClimaCore from ClimaAtmos to fix this. When I made the reproducer, I actually tested that the types of the two fields were exactly the same, so there are some mysterious compiler heuristics, or contexts that matter, that must make the real-world situation more difficult to infer.

PR #1757, opened 2023-06-12
Use report_opt in nfailures

charleskawczynski:
This PR changes the n-failure macros to use `report_opt`, since this seems to catch more jet failures.

### DIFF
--- a/perf/jet_test_nfailures.jl
+++ b/perf/jet_test_nfailures.jl
@@ -12,7 +12,7 @@ OrdinaryDiffEq.step!(integrator) # Make sure no errors
 # Suggested in: https://github.com/aviatesk/JET.jl/issues/455
 macro n_failures(ex)
     return :(
-        let result = JET.@report_call $(ex)
+        let result = JET.@report_opt $(ex)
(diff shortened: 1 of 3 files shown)


szy21 (approved):
What do these failures mean actually?

perf/jet_test_nfailures.jl
```diff
@@ -26,7 +26,7 @@ using Test
     # inference. By increasing this counter, we acknowledge that
     # we have introduced an inference failure. We hope to drive
     # this number down to 0.
-    n_allowed_failures = 84
+    n_allowed_failures = 844
```
akshaysridhar, line 29: The increase in `n_allowed_failures` -> is this because `call / opt` interpret the reported "failures" differently ?

charleskawczynski: Yes

charleskawczynski:
`perf/jet_test_nfailures.jl` may impose quite a bit of friction with changing to `report_opt`. We can remove it if it's too much and just leave the indicator (in `perf/jet_report_nfailures.jl`).

PR #1758, opened 2023-06-12
Fix aqua tests-- remove unused deps

charleskawczynski:
This PR fixes some aqua tests by removing unused dependencies. We should be able to use the examples env when we want to use `ClimaCorePlots`/`ClimaCoreVTK`.


akshaysridhar (approved):
LGTM

Issue #1761, opened 2023-06-12
Explicit ClimaAtmos without bycolumn for GPU support


simonbyrne:
While we're figuring out the ClimaCore changes, it might be worth seeing if we can get a fully explicit GPU case working here by adding an alternative codepath without the `bycolumn` pieces.

charleskawczynski:
Being that we've had a hand full of regressions with #720, I'm wondering if we should just remove some of the bycolumn calls-- the small ones may not be buying us much

simonbyrne:
Fixed by #1811

PR #1765, opened 2023-06-12
Remove some bycolumn calls

charleskawczynski:
This PR removes some bycolumn calls for explicit tendencies. A step towards #1761. I think the big remaining one will be in `additional_tendency!`.

### DIFF
--- a/perf/flame.jl
+++ b/perf/flame.jl
@@ -57,7 +57,7 @@ allocs_limit["flame_perf_target"] = 4320
 allocs_limit["flame_perf_target_tracers"] = 185904
 allocs_limit["flame_perf_target_edmfx"] = 33117648
 allocs_limit["flame_perf_target_edmf"] = 8636906992
-allocs_limit["flame_perf_target_threaded"] = 3850064
+allocs_limit["flame_perf_target_threaded"] = 6175664
 allocs_limit["flame_perf_target_callbacks"] = 11159912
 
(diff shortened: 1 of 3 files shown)


simonbyrne (approved):

juliasloan25:
What do `bycolumn` calls do? How is it different from just broadcasting over a field?

charleskawczynski:
> bycolumn

They loop over columns and call the passed function with `colidx::Fields.ColumnIndex` as an argument so that users can index into fields to extract column views.

juliasloan25:
> > What do bycolumn calls do? How is it different from just broadcasting over a field?
> 
> They loop over columns and call the passed function with `colidx::Fields.ColumnIndex` as an argument so that users can index into fields to extract column views. For example:
> 
> ```julia
> Fields.bycolumn(axes(f)) do colidx
> @. f[colidx] += 1
> end
> ```
> 
> Will loop over each column in `f` and add 1 across the vertical stack. It's a bit easier to write certain algorithms for `ClimaCore.ColumnField`s, and the loop over columns can be parallelized (assuming each column does not write to other columns).

That makes sense, thank you! Why do you need to remove `bycolumn` calls to run on GPUs?

charleskawczynski:
> That makes sense, thank you! Why do you need to remove `bycolumn` calls to run on GPUs?

https://github.com/CliMA/ClimaCore.jl/issues/1287

PR #1778, opened 2023-06-16
Use radians in radiation

### LINKED ISSUE #1716: Low resolution gray radiation aquaplanet has wrong result
Found by @valeriabarra @LenkaNovak. See this [build](https://buildkite.com/clima/climaatmos-longruns/builds/994#01885968-6016-42cc-a163-a67b66feb24a) for example.

Here is the qt (g/kg) at level 6 at day 30
image

szy21:
## Purpose 

RRTMGP uses radians for latitude and longitude. This PR fixes the bug that we pass in latitude in degrees.
Closes #1716.

## Content

----


szy21:
We decided to change to use degrees in RRTMGP to be more consistent. Ignore this PR for now.

PR #1781, opened 2023-06-16
Improve GPU test label, use baro wave init

charleskawczynski:
This PR improves the GPU test label. As far as I can tell, the default config is basically the baroclinic wave (assuming we use the baro wave init profile). So, this PR updates the label and changes initialization since this is a well known configuration.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -64,13 +64,12 @@ steps:
 
   - group: "GPU"
     steps:
-      - label: "Simple GPU test"
-        key: "gpu_example"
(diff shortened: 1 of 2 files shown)


.buildkite/pipeline.yml
```diff
@@ -64,13 +64,12 @@ steps:
 
   - group: "GPU"
     steps:
-      - label: "Simple GPU test"
-        key: "gpu_example"
+      - label: "GPU baroclinic wave"
```
charleskawczynski, line 67:
```suggestion
      - label: "GPU explicit baroclinic wave"
```

PR #1785, opened 2023-06-16
use maxiter 3 for saturation adjustment

szy21:
## Purpose 

Splitted from #1648. Changes the maximum iteration to 3 and relative tolerance to 0.003 for saturation adjustment. It works for most of the longruns: https://buildkite.com/clima/climaatmos-longruns/builds/1037#_. Most failures are from timeout. There are two jobs that actually fail, `longrun_bw_rhoe_equil_highres` and `longrun_ssp_bw_rhoe_equil_highres`, but they both fail because of unphysical input to SA. One of them is failing in the current main.

## Content

----


charleskawczynski (approved):

szy21:
I removed some SSP cases (Basically I only kept SSP + first order upwinding or zalesak for moist baroclinic wave, and SSP + first order upwinding for moist held suarez. We can do some further cleanup of ci at some point, if we decide to move on with ARS + limiter), and reduced dt and t_end for some SSP cases for the ci to pass. cc @valeriabarra

Issue #1788, opened 2023-06-16
Make definition of `comms_ctx` optional


valeriabarra:
In the latest ClimaAtmos version, `comms_ctx` is not optionally definable anymore. This construct conflicts with the Coupler, which has to define `comms_ctx` at the very beginning of the simulation. 

We should make this optional again.

PR #1789, opened 2023-06-16
Fix 1788

charleskawczynski:
Closes https://github.com/CliMA/ClimaAtmos.jl/issues/1788

### DIFF
--- a/src/solver/types.jl
+++ b/src/solver/types.jl
@@ -325,6 +325,7 @@ end
 function AtmosConfig(
     s = argparse_settings();
     parsed_args = parse_commandline(s),
+    comms_ctx = get_comms_context(parsed_args),
 )
     FT = parsed_args["FLOAT_TYPE"] == "Float64" ? Float64 : Float32
     toml_dict = CP.create_toml_dict(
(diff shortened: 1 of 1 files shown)


LenkaNovak:
In fact, @valeriabarra, this works too! :) Thanks, @charleskawczynski !

src/solver/types.jl
```diff
@@ -325,6 +325,7 @@ end
 function AtmosConfig(
     s = argparse_settings();
     parsed_args = parse_commandline(s),
+    comms_ctx = get_comms_context(parsed_args),
```
valeriabarra, line 328: Hey @charleskawczynski ! Thanks for the prompt fix! Did you see #1790 ? We probably both opened a PR in the same exact moment! :laughing: 

@LenkaNovak and I pushed that fix in #1790 together. If you guys think that this one is better and supersedes #1790, then please go ahead and close that one. I am going to be off starting shortly. Thank you.

valeriabarra (approved):

LenkaNovak (approved):

src/solver/types.jl
valeriabarra: Okay, I closed #1789  as superseded. Thank you both.

charleskawczynski: Oh, I didn't see it, sorry! Either one is fine!

PR #1793, opened 2023-06-19
Log simulated years per day

charleskawczynski:
This PR prints the simulated years per day to the log. Can someone verify my sypd formula / computations is correct?


szy21 (approved):
The computation looks correct to me.

Issue #1795, opened 2023-06-20
Solution should be independent from `save_to_disc` frequency


charleskawczynski:
One issue with `save_to_disc` is that that’s not all that it does— it also computes diagnostics and variables in the solver cache.

In addition, these computations should be isolated to a diagnostics cache, and not our solver cache. This is a problem because right now, it seems that our solution depends on the frequency that we export data, which doesn’t seem good.

We should really
- separate out diagnostic computations from exporting.
- Have a dedicated diagnostics cache, independent from the solver cache. I think it’s fine if it’s duplicate, for now. Now that I think of it— why would we need to recompute the solver cache in the diagnostics function anyway? Wouldn’t it already be up to date? I suppose it may not be if it’s radiative fluxes, which get computed in other callbacks.. 

cc @trontrytel, @bischtob, @akshaysridhar, @nefrathenrici, @juliasloan25

szy21:
I remember when we first do it, @dennisYatunin mentioned the cache may not be up-to-date when saving diagnostics. I don't know whether it's still the case.

I agree with this in general. I think it would be good to separate the diagnostics from the solver. I don't think the solution depends on the frequency of output right now though?

charleskawczynski:
>  @dennisYatunin mentioned the cache may not be up-to-date when saving diagnostics.

This makes sense for purely diagnostic quantities, but I don't think it makes sense for the solver cache. @dennisYatunin, can you elaborate? Is it a matter of when the callback is called? Calling it at the end of the step, before the solution is incremented, makes the most sense (for the very reason that variables don't need to be re-computed).

> I don't think the solution depends on the frequency of output right now though?

It's easy enough to test!

trontrytel:
Agreed! Once we separate the diagnostics it would be super nice to add things like averages over certain time or space to the output. Not sure how much of that can be re-used from old ClimateMachine. We had a big diagnostics module there

dennisYatunin:
> Calling it at the end of the step, before the solution is incremented, makes the most sense (for the very reason that variables don't need to be re-computed).

@charleskawczynski This is only true when using an explicit timestepping scheme. If we are using an implicit scheme, then every Newton iteration of the implicit stages will look as follows:

- Compute the cache and tendency for the current estimate of the implicit stage.
- Use the tendency to set up a linear system of equations, and solve the linear system.
- Modify the current estimate of the implicit stage using the solution of the linear system.

[...]

PR #1797, opened 2023-06-20
Adjust save_to_disc freq to test 1795

charleskawczynski:
Testing #1795.


charleskawczynski:
I think that this shouldn't be an issue with the new `set_precomputed_quantities!` post callback hooks, since they're always called after the end of each timestep (before callbacks), so the state is mutated by `set_precomputed_quantities!` at every timestep.

PR #1798, opened 2023-06-20
Update gpu performance benchmark script

charleskawczynski:
This PR updates the performance benchmark script to run the gpu example for 10 steps. Note that it runs for 20 steps total (the first call is to compile). cc @bloops.


simonbyrne (approved):

bloops:
Can we have it print the walltime it took when running the final benchmark step? Thanks!

charleskawczynski:
Done, @bloops!

PR #1803, opened 2023-06-21
Assert`SurfaceState` type fix

### LINKED ISSUE #1802: Assert`SurfaceState` type fix
For the coupled runs we require the query for `eltype(coupler_sfc_setp)` to be true if type of `SurfaceSetup`, i.e.: 

Screen Shot 2023-06-20 at 5 57 16 PM

Required by this ClimaCoupler PR that updates for the latest ClimaAtmos: https://github.com/CliMA/ClimaCoupler.jl/issues/308

LenkaNovak:
Fix when asserting type of `SurfaceState` for coupled runs. 

## Purpose 

closes #1802 

(see issue for details)

----

### DIFF
--- a/src/surface_conditions/surface_conditions.jl
+++ b/src/surface_conditions/surface_conditions.jl
@@ -61,7 +61,7 @@ function update_surface_conditions!(Y, p, t)
             atmos,
         )
     elseif sfc_setup isa Fields.Field # This case is needed for the Coupler
-        @assert eltype(sfc_setup) isa SurfaceState
+        @assert eltype(sfc_setup) <: SurfaceState
         sfc_setup_values = Fields.field_values(sfc_setup)
         @. sfc_conditions_values = surface_state_to_conditions(
             sfc_setup_values,


charleskawczynski:
This change looks good, can we add a coupler compatibility test to our CI?

LenkaNovak:
Thanks, @charleskawczynski ! And totally, that would be awesome! What in particular were you thinking? So far we've been trying to avoid Atmos depending on the Coupler. We were hoping that eventually all component models would host their own extensions of coupling methods dispatched on their own simulation types. In practice, this means moving the functionality in our [init files](https://github.com/CliMA/ClimaCoupler.jl/blob/valeria/atmos-v0.12/experiments/AMIP/modular/components/atmosphere/climaatmos_init.jl) to Atmos and Land (and later Ocean), which would make local testing easier. This is what @glwagner and I discussed for ClimaEarth (we'll return to developing it probably towards the end of the summer), but we can think about it for this version of the coupler too.

charleskawczynski:
I think this PR is fixing a bug-- I don't think we never drop into `elseif sfc_setup isa Fields.Field`

> What in particular were you thinking?

So, I think we can add a test that simply exercises when we do exercise this particular `elseif` block. For example:

```julia
import ClimaAtmos as CA
using Test

@testset "Coupler compatibility" begin
    config = CA.AtmosConfig(#=arguments the coupler uses=#)

    (; p, t) = integrator
    Y = integrator.u;
    du = integrator.u;
    Yₜ = similar(Y);

    # Need to be able to call update_surface_conditions!
    # with this config.
    CA.update_surface_conditions!(Y, p, t)
end
```

LenkaNovak:
Ahh you mean implementing this in the coupler?

charleskawczynski:
> Ahh, are you suggesting to implement this in the coupler?

Ah, sorry, no I meant to implement in ClimaAtmos-- this PR is fixing a ClimaAtmos bug, because we're not exercising these code paths

charleskawczynski:
I don't think that code needs to be replicated. We don't necessarily have to test anything at the moment, I'm fine with settling with exercising the code. Right now, it looks broken, and that's because it's not exercised anywhere. We can think more about how we test things later

LenkaNovak:
Sounds good! It is ok if I merge this as is?

charleskawczynski:
> It is ok if I merge this as is?

Can we add the test I proposed above ([here](https://github.com/CliMA/ClimaAtmos.jl/pull/1803#issuecomment-1601093917))?

dennisYatunin:
It sounds like it might be a bit of a hassle to exercise this code path in ClimaAtmos, as it would require replicating a bunch of Coupler setup code. Also, I don't think this branch of the if-statement will last very long, as the Coupler is just using it to test things while they set up their own infrastructure for surface conditions. Once the API is clearer on their end, Nat and I will have to rethink how we want to pass spatially-dependent surface information to ClimaAtmos.

LenkaNovak:
@charleskawczynski and I met and decided to follow up on a unit test in a different PR.

charleskawczynski:
Yes, opened https://github.com/CliMA/ClimaAtmos.jl/pull/1809

charleskawczynski (approved):

PR #1805, opened 2023-06-21
Remove more bycolumn loops

charleskawczynski:
This PR removes more `bycolumn` loops. These calls are inside the implicit tendency, which are still called with explicit solvers (except that the tendencies are added explicitly). This was found in this [build](https://buildkite.com/clima/climaatmos-ci/builds/10597#0188d9a4-752f-4a25-8841-b4d7aebff0bc).

This should fix some rather large performance issues for the gpu example (cc @bloops)

The benchmark is looking more promising:

Main branch:
`[ Info: Ran step! 10 times in 17.764 s (45437150 allocations: 9.56 GiB), (1.780 s per step)`
This PR:
`[ Info: Ran step! 10 times in 127.031 ms (244420 allocations: 35.30 MiB), (15.710 ms per step)`

I will say, one thing is very odd: why is the benchmark, which runs 20 steps, so much faster than the target job, which runs 60 steps? (at `15.710 ms` per step, the difference should be < 1 second)

### DIFF
--- a/src/prognostic_equations/implicit/implicit_tendency.jl
+++ b/src/prognostic_equations/implicit/implicit_tendency.jl
@@ -11,9 +11,9 @@ function implicit_tendency!(Yₜ, Y, p, t)
         @nvtx "precomputed quantities" color = colorant"orange" begin
             set_precomputed_quantities!(Y, p, t)
         end
-        Fields.bycolumn(axes(Y.c)) do colidx
-            implicit_vertical_advection_tendency!(Yₜ, Y, p, t, colidx)
-            if p.turbconv_model isa TurbulenceConvection.EDMFModel
+        implicit_vertical_advection_tendency!(Yₜ, Y, p, t)
+        if p.turbconv_model isa TurbulenceConvection.EDMFModel
+            Fields.bycolumn(axes(Y.c)) do colidx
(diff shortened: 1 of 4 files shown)


simonbyrne:
was this on the CPU or GPU?

charleskawczynski:
> was this on the CPU or GPU?

This was for the gpu example

simonbyrne:
It could just be compilation time? GPU kernels are much more expensive to compile.

bloops:
> I will say, one thing is very odd: why is the benchmark, which runs 20 steps, so much faster than the target job, which runs 60 steps? (at 15.710 ms per step, the difference should be < 1 second)

I saw that the first 10 steps (which includes compilation time) in the benchmark takes 40x more walltime than the next 10 steps (even without this PR). So perhaps the target job running 60 steps is also counting the compilation time?

charleskawczynski:
Ah, my mistake. We modify `h_elem` in the buildkite performance job (I did it originally kind of as a demo). Let's just remove that argument. That should make it more similar.

charleskawczynski:
@bloops, we're going to close this in favor of https://github.com/CliMA/ClimaCore.jl/pull/1341 and https://github.com/CliMA/ClimaCore.jl/pull/1332.

PR #1808, opened 2023-06-21
Modify dt for failing long-run builds.

akshaysridhar:
Following longrun build (test) no. 1045, decrease timestep for 2 experiments in the target AMIP resolution list. 
- no lim ARS baroclinic wave
- SSP equilmoist baroclinic wave high resolution

### DIFF
--- a/.buildkite/longruns/pipeline.yml
+++ b/.buildkite/longruns/pipeline.yml
@@ -56,7 +56,7 @@ steps:
 
       - label: ":computer: no lim ARS baroclinic wave (ρe_tot) equilmoist high resolution"
         command:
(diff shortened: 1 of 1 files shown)


szy21 (approved):
Does it fix the problem?

akshaysridhar:
> 

Yes - see here for instance : https://buildkite.com/clima/climaatmos-longruns/builds/1045. The build was triggered to test larger dt in topo (this part was unsucessful) ; I had included reduced dt in the failing experiments in the two experiments mentioned above.

akshaysridhar:
Runtime ends up being 6h15m for the no-lim ARS case, and 14h20m for the SSP equilmoist BW case.

szy21:
Great, thanks!

PR #1809, opened 2023-06-21
Add coupler compatibility tests

charleskawczynski:
This PR adds coupler compatibility coverage tests.

### DIFF
--- a/src/solver/cli_options.jl
+++ b/src/solver/cli_options.jl
@@ -382,6 +382,7 @@ function argparse_settings()
 end
 
 parse_commandline(s) = ArgParse.parse_args(ARGS, s)
+parse_commandline(args, s) = ArgParse.parse_args(args, s)
 
 function cli_defaults(s::ArgParse.ArgParseSettings)
     defaults = Dict()
--- a/test/coupler_compatibility.jl
+++ b/test/coupler_compatibility.jl
@@ -1,0 +1,171 @@
+using Test
+import Thermodynamics as TD
+import SurfaceFluxes as SF
+import StaticArrays as SA
+import ClimaAtmos as CA
+import ClimaAtmos.Parameters as CAP
+import ClimaAtmos.SurfaceConditions: projected_vector_data, CT1, CT2
+import ClimaCore: Spaces, Fields
+import ClimaCore.Utilities: half
+
+const z0m = 1e-3
+const z0b = 1e-5
+const gustiness = 1
+const beta = 1
+const T1 = 300
+const T2 = 290
+
+# This file contains two tests that illustrate how the coupler can modify the
+# surface temperature in ClimaAtmos.
+
+# In the first test, the ClimaAtmos "cache" p is overwritten so that it contains
+# a surface field specified by the coupler, and then the internal function
+# set_precomputed_quantities! is called to verify that this surface field is
+# used correctly.
+
+# In the second test, the cache is not overwritten. Instead, the coupler defines
+# its own surface fields, and then it copies the data from those fields into the
+# cache by calling set_surface_conditions!.
+
+# Since overwriting the internal data structures of ClimaAtmos makes it harder
+# to develop a modular API, the pattern demonstrated in the first test should
+# only be used to quickly write tests for the coupler. For more long-term code,
+# the pattern demonstrated in the second test is preferable.
+
+@testset "Coupler Compatibility (Hacky Version)" begin
+    # Initialize a model. The value of --surface_setup is irrelevant, since it
+    # will get overwritten.
+    config = CA.AtmosConfigArgs(;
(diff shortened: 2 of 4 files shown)


dennisYatunin:
I've fixed the test that was originally set up in this PR, and I also added a second test that demonstrates the use of `set_surface_conditions!`, which is how Nat and I envisioned the coupler would link to ClimaAtmos. @charleskawczynski @juliasloan25 Let me know if there are any other changes you'd like to see.

test/coupler_compatibility.jl
```diff
@@ -0,0 +1,171 @@
...
+# surface temperature in ClimaAtmos.
+
+# In the first test, the ClimaAtmos "cache" p is overwritten so that it contains
+# a surface field specified by the coupler, and then the internal function
+# set_precomputed_quantities! is called to verify that this surface field is
+# used correctly.
+
+# In the second test, the cache is not overwritten. Instead, the coupler defines
+# its own surface fields, and then it copies the data from those fields into the
+# cache by calling set_surface_conditions!.
```
juliasloan25, line 28: Could you add an inline comment showing where the coupler defines the surface fields?

juliasloan25 (commented):
Looks good, I just had a couple of comments. I also asked Lenka to take a look since she's more familiar with the intended atmos/coupler interactions

src/solver/cli_options.jl
```diff
@@ -382,6 +382,7 @@ function argparse_settings()
 end
 
 parse_commandline(s) = ArgParse.parse_args(ARGS, s)
+parse_commandline(ARGS, s) = ArgParse.parse_args(ARGS, s)
```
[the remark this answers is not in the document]
charleskawczynski: It's so that we can use this syntax: `config = CA.AtmosConfigArgs(; args = ["--initial_condition", "DryBaroclinicWave"],)`

test/coupler_compatibility.jl
charleskawczynski: I feel like that will go stale very quickly.. Would it be helpful to just provide a link to the coupler repo?

juliasloan25: I guess I'm just confused about how we can be using a function from the coupler here since we aren't dependent on it. When we say "the coupler defines its own surface fields", are we using a function defined here that simulates the coupler's behavior (maybe `update_surface_fields!`)?

LenkaNovak (approved):
Looks good to me! We will implement this in the next (and last) coupler SF PR. Thanks for doing this!

src/solver/cli_options.jl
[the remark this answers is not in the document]
charleskawczynski: maybe we should instead call this `args` to avoid confusion with the `ARGS` environment variable, though

test/coupler_compatibility.jl
valeriabarra: I think Julia has a good point. We should clarify

charleskawczynski: Maybe @dennisYatunin can clarify?

test/coupler_compatibility.jl
```diff
@@ -0,0 +1,171 @@
...
+    thermo_params = CAP.thermodynamics_params(p.params)
+
+    # Allocate fields for storing the thermodynamic state and fluxes at the
+    # surface.
+    sfc_ts = similar(Spaces.level(Y.f, half), TD.PhaseDry{FT})
+    sfc_conditions = similar(sfc_ts, SF.SurfaceFluxConditions{FT})
+
+    # Define a function for updating these fields, given the temperature at the
+    # surface and the current cache p.
+    function update_surface_fields!(sfc_ts, sfc_conditions, surface_T, p)
```
charleskawczynski, line 98: This is more boilerplate than I expected. Does the coupler define these functions? If so, maybe we should move them into ClimaAtmos src and have the coupler call them and change this test accordingly.

test/coupler_compatibility.jl
dennisYatunin: Oh, sorry if this was unclear. I meant that, if the coupler were to use the design pattern shown in this section, it would need to define its own surface fields, and then it would need to copy the data from those fields into the cache. I do not believe the coupler is currently doing this; instead, it is using the design pattern shown in the preceding section, which involves overwriting the cache.

test/coupler_compatibility.jl
dennisYatunin: The coupler is not currently defining these functions. I am not sure what the full set of requirements will be for these functions (I know that they will need to support computing linear combinations of land and ocean surface fluxes, which we probably don't want to do on the ClimaAtmos side), so I just added a simple example to sketch out the pattern we should aim for. Maybe after a few design iterations in the coupler, we can move them into some centralized location?

PR #1810, opened 2023-06-21
Revert "Remove some bycolumn calls"

charleskawczynski:
In light of https://github.com/CliMA/ClimaCore.jl/pull/1341, this PR reverts https://github.com/CliMA/ClimaAtmos.jl/pull/1765.

It's unclear to me that threading over these loops is effective. There's not as much compute in them compared to the one in `additional_tendency!`.


charleskawczynski:
This is pretty stale, and doesn't impact gpu performance.

PR #1819, opened 2023-06-23
Allocations fix for diagnostic edmf

nefrathenrici:
This PR:
- fixes allocations for `set_diagnostic_edmf_precomputed_quantities!`
- Adds a performance test for job diagnostic_edmfx_aquaplanet

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -850,6 +850,12 @@ steps:
         agents:
           slurm_mem: 20GB
 
+      - label: ":fire: Flame graph: perf target (diagnostic edmfx)"
+        command: "julia --color=yes --project=perf perf/flame.jl --job_id flame_perf_target_diagnostic_edmfx --target_job diagnostic_edmfx_aquaplanet"
+        artifact_paths: "flame_perf_target_diagnostic_edmfx/*"
+        agents:
+          slurm_mem: 20GB
+
       - label: ":fire: Flame graph: perf target (edmf)"
         command: "julia --color=yes --project=perf perf/flame.jl --job_id flame_perf_target_edmf --target_job sphere_baroclinic_wave_rhoe_equilmoist_edmf --moist dry --precip_model nothing --rad nothing"
         artifact_paths: "flame_perf_target_edmf/*"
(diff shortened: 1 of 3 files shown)


charleskawczynski:
Hi @nefrathenrici, this allocation very well may be due to `^2`, see https://github.com/CliMA/ClimaCore.jl/issues/1126

charleskawczynski (approved):

src/cache/diagnostic_edmf_precomputed_quantities.jl
```diff
@@ -314,31 +325,52 @@ function set_diagnostic_edmf_precomputed_quantities!(Y, p, turbconv_model)
...
                         entr_detrʲ_prev_level.entr * u³ʲ_data_prev_halflevel
                     )
                 )
-
-            @. u³ʲ_halflevel = ifelse(
-                (
-                    u³ʲ_datau³ʲ_data < (FT(1e-6) / ∂x³∂ξ³_level^2) ||
+            # Helper function for broadcasting ternary operator across fields.
+            # This is necessary to prevent allocations using ifelse().
+            function pointwise_ifelse(
```
charleskawczynski, line 342: Is this needed? I haven't seen `ifelse` allocate in the past

charleskawczynski: Sorry, I don't remember seeing these changes when I approved the PR, this looks highly specialized and it's not clear to me that it's needed

.buildkite/pipeline.yml
```diff
@@ -843,6 +843,12 @@ steps:
         agents:
           slurm_mem: 20GB
 
+      - label: ":fire: Flame graph: perf target (diagnostic edmfx)"
+        command: "julia --color=yes --project=perf perf/flame.jl --job_id flame_perf_target_diagnostic_edmfx --target_job diagnostic_edmfx_aquaplanet"
+        artifact_paths: "flame_perf_target_diagnostic_edmfx/*"
+        agents:
+          slurm_mem: 100GB
```
charleskawczynski, line 850: I don't think this is a good idea

nefrathenrici: I agree, it was the only way I could get the job to complete. Maybe we should hold off on adding this perf test until I figure out what's going on with the ifelse allocations?

PR #1826, opened 2023-06-26
Add to nfailures list, track allocations too

charleskawczynski:
This PR adds some methods to the `nfailures` list, and tracks allocations per method call, too.

### DIFF
--- a/perf/jet_report_nfailures.jl
+++ b/perf/jet_report_nfailures.jl
@@ -10,6 +10,14 @@ macro n_failures(ex)
     )
 end
 
+macro jet_and_allocs(ex)
+    return quote
+        local n = @n_failures $(ex)
+        local _allocs = @allocated $(ex)
+        (; nfailures = n, allocs = _allocs)
+    end
+end
+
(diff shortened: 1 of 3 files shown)


charleskawczynski:
I just realized some issues with attempting to call granular RHS kernels after a simulation:
 - Each kernel is conditionally called, based on the model
 - The order of kernel calls is important (e.g., needing to call precomputed quantities before others)
 - Calling a kernel mutates the cache, which could affect post processing
We could create a deepcopy, but that only fixes the 3rd point.

I'm thinking the better approach here is to time / record allocations in-situ. This basically fixes all of these issues, and has the benefit of being able to accumulate statistics (e.g., with TimerOutputs.jl).

The only issue/decision there is deciding between timing methods:
 - NVTX - CPU or GPU, accumulates statistics, requires using Nsight compute etc.
 - TimerOutputs.jl - maybe CPU only, accumulates statistics, nice output, no post processing required
 - others?

I'll still peel off the rename / return stuff from this PR.

PR #1829, opened 2023-06-27
Break up `save_to_disk_func`

charleskawczynski:
This PR breaks up `save_to_disk_func` into two parts.


charleskawczynski:
This doesn't seem to be improving inference, however, I still kind of like the change.

szy21 (approved):

PR #1832, opened 2023-06-27
Add flame graph for gravity wave kernels

charleskawczynski:
Turning on the gravity wave results in lots of allocations-- the `aquaplanet (ρe_tot) equilmoist allsky radiation monin_obukhov varying insolation gravity wave (raw_topo) high top` job allocates 185.00 GiB, which is a lot considering it only performs 162 steps.

This PR adds a flame graph, and allocation limit for a job that includes gravity wave.

This means that further developments in these kernels will be throttled by allocation tests.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -862,6 +862,12 @@ steps:
         agents:
           slurm_mem: 20GB
 
+      - label: ":fire: Flame graph: gravity wave"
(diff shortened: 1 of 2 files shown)


nefrathenrici (approved):

szy21:
Thanks Charlie! We definitely didn't pay any attention to the performance when coding the gravity wave. It would be great to improve the performance here. I would focus on other parts (if there is any) though, as gravity wave is not essential to the first AMIP simulation.

PR #1834, opened 2023-06-28
Cache intermediate fields

charleskawczynski:
This PR caches some intermediate fields in the GW kernel, since jobs that use the GWP are hitting memory issues (see build https://buildkite.com/clima/climaatmos-ci/builds/10849). Maybe this will help with inference, too.


szy21 (approved):

charleskawczynski:
This is stale and needs to be reworked.

PR #1841, opened 2023-06-29
add diffusive flux for edmfx

### LINKED ISSUE #1840: Add diffusive flux for edmfx

szy21:
## Purpose 

I decided to add the diffusive flux first as I'm a bit stuck on the mixing length. We need this to test the mixing length anyway so I think it would be good to have. We can also have surface flux with edmfx after this change, without turning on vertical diffusion. The bomex result doesn't look good because I am using a constant diffusivity, but it should look better once we switch to the more physical diffusivity.

Closes #1840 

## Content

----


szy21:
I have two questions (maybe for @tapios?)
1. Do we need vertical diffusion for u_3? (I think we do, we just don't have it for the simple vertical diffusion, but we should probably add it?)
2. The diffusivity K_h will be on cell center. To calculate the flux on the cell face, should we do ᶠinterp(ρ * K_h), or ᶠinterp(ρ) *  ᶠinterp(K_h), or something else?

tapios:
> I have two questions (maybe for @tapios?)
> 
> 1. Do we need vertical diffusion for u_3? (I think we do, we just don't have it for the simple vertical diffusion, but we should probably add it?)

Vertical diffusion for grid-scale u_3 should come from the ED part of the EDMF scheme. It should not matter for the coarse resolutions we currently use, though, as they are approximately hydrostatic. 

> 2. The diffusivity K_h will be on cell center. To calculate the flux on the cell face, should we do ᶠinterp(ρ * K_h), or ᶠinterp(ρ) *  ᶠinterp(K_h), or something else?

I'm not sure, but ᶠinterp(ρ) *  ᶠinterp(K_h) looks close to what we would do for other fluxes (so -K_h grad(phi) plays the role of what otherwise would be an advective flux, and we should treat advective and diffusive fluxes similarly).

szy21:
@charleskawczynski This increases the edmfx allocation by 1%: https://buildkite.com/clima/climaatmos-ci/builds/10917#01890d08-a0f9-494f-b5e4-13143f89dca7. I think it's from adding two more fields to the cache? Does this look ok to you?

szy21:
I will leave the vertical diffusion of u_3 to another PR. I need to check the boundary condition with Dennis and/or Tapio.

src/prognostic_equations/edmfx_sgs_flux.jl
```diff
@@ -48,7 +61,28 @@ function edmfx_sgs_flux_tendency!(Yₜ, Y, p, t, colidx, turbconv_model::EDMFX)
...
+                ᶠρaK_h[colidx] * ᶠgradᵥ(
+                    TD.total_specific_enthalpy(
+                        thermo_params,
+                        ᶜts⁰[colidx],
+                        ᶜspecific⁰.e_tot[colidx],
+                    ),
+                )
+            ),
+        )
+
```
akshaysridhar, line 83: Energy eq. change LGTM.

src/prognostic_equations/edmfx_sgs_flux.jl
```diff
@@ -75,10 +109,31 @@ function edmfx_sgs_flux_tendency!(Yₜ, Y, p, t, colidx, turbconv_model::EDMFX)
...
+        # diffusive flux
+        ᶠρaK_u = p.ᶠtemp_scalar
+        @. ᶠρaK_u[colidx] = ᶠinterp(ᶜρa⁰[colidx]) * ᶠinterp(ᶜK_u[colidx])
+        ᶜdivᵥ_uₕ = Operators.DivergenceF2C(
+            top = Operators.SetValue(C3(FT(0)) ⊗ C12(FT(0), FT(0))),
+            bottom = Operators.SetValue(sfc_conditions.ρ_flux_uₕ[colidx]),
+        )
+        @. Yₜ.c.uₕ[colidx] -=
+            ᶜdivᵥ_uₕ(-(ᶠρaK_u[colidx] * ᶠgradᵥ(Y.c.uₕ[colidx]))) / Y.c.ρ[colidx]
```
akshaysridhar, line 133: `moisture` and `uh` eq LGTM. bctype for uh bottom is a  
```
1×2 ClimaCore.Geometry.AxisTensor
```
with value 
```
 0  0
```

src/prognostic_equations/edmfx_sgs_flux.jl
```diff
@@ -118,7 +176,19 @@ function edmfx_sgs_flux_tendency!(
...
+        ᶠρaK_h = p.ᶠtemp_scalar
+        @. ᶠρaK_h[colidx] = ᶠinterp(Y.c.ρ[colidx]) * ᶠinterp(ᶜK_h[colidx])
+
+        ᶜdivᵥ_ρe_tot = Operators.DivergenceF2C(
+            top = Operators.SetValue(C3(FT(0))),
+            bottom = Operators.SetValue(sfc_conditions.ρ_flux_h_tot[colidx]),
+        )
+        @. Yₜ.c.ρe_tot[colidx] -=
+            ᶜdivᵥ_ρe_tot(-(ᶠρaK_h[colidx] * ᶠgradᵥ(ᶜh_tot[colidx])))
+
```
akshaysridhar, line 189: This has the same form as the equation for the `EDMFX` type. LGTM overall -  `ᶜh_tot[colidx]` here is the value computed from cached dycore variables, and the corresponding enthalpy above in the equation for `EDMFX` 
```
                    TD.total_specific_enthalpy(
                        thermo_params,
                        ᶜts⁰[colidx],
                        ᶜspecific⁰.e_tot[colidx],
                    ),
```
accounts for the subgrid updraft variables (?) - is my understanding right here?

src/prognostic_equations/edmfx_sgs_flux.jl
```diff
@@ -133,10 +203,31 @@ function edmfx_sgs_flux_tendency!(
...
+        # diffusive flux
+        ᶠρaK_u = p.ᶠtemp_scalar
+        @. ᶠρaK_u[colidx] = ᶠinterp(Y.c.ρ[colidx]) * ᶠinterp(ᶜK_u[colidx])
+        ᶜdivᵥ_uₕ = Operators.DivergenceF2C(
+            top = Operators.SetValue(C3(FT(0)) ⊗ C12(FT(0), FT(0))),
+            bottom = Operators.SetValue(sfc_conditions.ρ_flux_uₕ[colidx]),
+        )
+        @. Yₜ.c.uₕ[colidx] -=
+            ᶜdivᵥ_uₕ(-(ᶠρaK_u[colidx] * ᶠgradᵥ(Y.c.uₕ[colidx]))) / Y.c.ρ[colidx]
```
akshaysridhar, line 227: Identical to eqn for `EDMFX` type, and to the `vertical diffusion` tendency in the dycore prognostic equations.

src/prognostic_equations/edmfx_sgs_flux.jl
szy21: It's the environment. I will probably add h_tot^0 to cache in a separate PR.

szy21:
Akshay and I talked offline and all the conversations are resolved.

src/prognostic_equations/edmfx_sgs_flux.jl
```diff
@@ -118,7 +176,19 @@ function edmfx_sgs_flux_tendency!(
             )
         end
 
+        # diffusive flux
+        ᶠρaK_h = p.ᶠtemp_scalar
+        @. ᶠρaK_h[colidx] = ᶠinterp(Y.c.ρ[colidx]) * ᶠinterp(ᶜK_h[colidx])
+
+        ᶜdivᵥ_ρe_tot = Operators.DivergenceF2C(
```
szy21, line 183: Actually maybe the allocation is from this. Is this ok @charleskawczynski ?

charleskawczynski: Maybe? Can you check to see if the number of jet failures is changed or unchanged? (It’s printed in a dictionary in the log)

szy21: Ah, it is changed. It changes from 
```
"limited_tendency!" => 8
"remaining_tendency!" => 8
"set_precomputed_quantities!" => 8
"step!" => 235
"implicit_tendency!" => 8
```
to
```
"additional_tendency!" => 2
"limited_tendency!" => 8
"remaining_tendency!" => 10
"set_precomputed_quantities!" => 8
"step!" => 236
"implicit_tendency!" => 8
```
Note the command line for the target job also changed (it changed from --vert_diff true to false(default), see the pipeline change).

szy21: I guess it means there are type instabilities in my code?

charleskawczynski: Yes, we can modify https://github.com/CliMA/ClimaAtmos.jl/blob/main/perf/jet.jl#L5 to point to the breaking job in order to identify where we're adding type-unstable code.

Issue #897, opened 2022-10-05
Improve design of nonorographic gravity wave parameterization


jiahe23:
## Files
 - `src/parameterized_tendencies/gravity_wave_drag/non_orographic_gravity_wave.jl`
 - All tests in `test/parameterized_tendencies/gravity_wave/non_orographic_gravity_wave/`

## Notes
 - Remove use of `parent`
 - Remove use of `copy` and `vec` in `copy(vec(parent(u_phy[colidx])))`
  - if `non_orographic_gravity_wave_forcing` is purely functional, we don't need to make
    a copy of the inputs, we can pass pointers to the inputs (i.e., `u_phy`)
 - Reduce allocations. And can check/test allocations with `@allocated` (see docs about usage)
   - `ᶜbuoyancy_frequency = @. (grav / ᶜT) * (ᶜdTdz + grav / TD.cp_m(thermo_params, ᶜts))` should become
   - `@. ᶜbuoyancy_frequency = (grav / ᶜT) * (ᶜdTdz + grav / TD.cp_m(thermo_params, ᶜts))`
   - Move `source_level = similar(Fields.level(Y.c.ρ, 1))` to cache
   - Move `u_phy` to cache in `u_phy = Geometry.UVVector.(Y.c.uₕ).components.data.:1`
   - Move `uforcing` in `uforcing = ones(axes(u_phy))` to cache
   - preallocate `Bw_exp` in `Bw_exp = @. exp(-log(2.0) * ((c * flag + c_hat0 * (1 - flag) - c0) / cw)^2)`
   - Apply the same to other variables that use similar patterns
   - Remove use of `append!`

 - Remove use of `Fields.bycolumn` (this may require developing custom CPU/GPU kernels in ClimaCore). Ideally we can just use `column_mapreduce!`
 - Remove use of `findlast` and `findfirst`. Hopefully we can just use `column_mapreduce!`? If not, then replace with `ᶜupdraft_top` in https://github.com/CliMA/ClimaAtmos.jl/pull/2819/files.
 - Any use of `getindex` (e.g., `[1]` in `parent(source_level[colidx]) .= findlast(parent(ᶜp[colidx]) .> gw_source_pressure)[1]`)
 - Refactor `non_orographic_gravity_wave_forcing` to be purely functional (no mutation of input arguments). See the comments and cross-referenced issues / pull requests

charleskawczynski:
Concretely:
[...]

PR #1844, opened 2023-06-30
Simplify GWP

charleskawczynski:
This PR simplifies the GWP a bit by not unpacking and packing a bunch of variables. A small step towards #897.

@szy21, I'm going to refactor this a bit, I'll add you on review if I start running into modifications that require behavioral changes.


charleskawczynski:
Ah, the test suite doesn’t seem to use that same type.. need to make fixes there

charleskawczynski:
This is stale and needs to be reworked.

PR #1845, opened 2023-06-30
Use WVector for temp gradient in GWP

charleskawczynski:
This PR uses the `WVector` for temperature gradient in the GWP in order to remove a call to `parent`.


charleskawczynski:
This needs to be reworked.

PR #1867, opened 2023-07-10
Dont use AtmosTargetParsedArgs in gpu perf scripts

charleskawczynski:
This PR:
 - Removes the gpu perf jobs that use the default driver, since these were really just duplicates of the benchmark step jobs, and using `target_job_id` was not really composeable.
 - Removes the use of `AtmosTargetParsedArgs`, so that all arguments are passed the the CL
 - Adds a moisture job

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -790,15 +790,6 @@ steps:
   - group: "GPU Performance"
     steps:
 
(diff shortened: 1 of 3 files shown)


charleskawczynski:
Next, I'll add a moist case, which I'll just leave as soft fail to start with

PR #1856, opened 2023-07-06
fix surface flux diagnostics in remapped files

szy21:
## Purpose 

Closes #1854 

## Content

----
- [ ] I have read and checked the items on the review checklist.

PR #1870, opened 2023-07-11
Fix surface fluxes

simonbyrne:
Surface normal is from normalized `Covariant3Vector(1)` (as that is the contravariant basis vector)

I missed this in the review of #1856. It should only affect runs with topography.

## Content

----


szy21:
It failed I think: https://buildkite.com/clima/climaatmos-ci/builds/11097#0189464a-2848-40e7-aa6e-5e123b19dcc8/137-204

juliasloan25:
out of date

PR #1879, opened 2023-07-13
Use nsight-systems/2023.1.1

sriharshakandala:
Use nsight-systems/2023.1.1

----


simonbyrne:
For context, the reason is that we're seeing occasional `The timeout expired.` errors on certain nodes with `nsight-systems/2023.2.1`.

simonbyrne (approved):

PR #1871, opened 2023-07-12
surface temp using sfc setup as a field

etmaca5:
## Purpose 

Temporary surface temperature implementation using surface setup as a field

----
- [ ] I have read and checked the items on the review checklist.

PR #1889, opened 2023-07-17
prognostic surface temp implementation

etmaca5:
## Purpose 

Surface Temperature implementation using a prognostic surface tendency
Supersedes #1871 

----


src/surface_conditions/surface_conditions.jl
```diff
@@ -28,6 +28,7 @@ function update_surface_conditions!(Y, p, t)
     int_z_values =
         Fields.field_values(Fields.level(Fields.coordinate_field(Y.c).z, 1))
     sfc_conditions_values = Fields.field_values(sfc_conditions)
+    sfc_field = hasproperty(Y, :sfc) ? Fields.field_values(Y.sfc) : nothing
```
dennisYatunin, line 31:
```suggestion
    sfc_prognostic_values = hasproperty(Y, :sfc) ? Fields.field_values(Y.sfc) : nothing
```

src/surface_conditions/surface_conditions.jl
```diff
@@ -122,6 +126,7 @@ function surface_state_to_conditions(
     interior_z,
     params,
     atmos,
+    sfc_field
```
dennisYatunin, line 129:
```suggestion
    surface_prognostic_values
```

src/surface_conditions/surface_conditions.jl
```diff
@@ -131,8 +136,8 @@ function surface_state_to_conditions(
 
     (!isnothing(q_vap) && atmos.moisture_model isa DryModel) &&
         error("surface q_vap cannot be specified when using a DryModel")
-
-    if isnothing(T)
+    if isnothing(sfc_field)
+        #TODO change with more accurate latitude and longitude model
```
dennisYatunin, line 140:
```suggestion
    if isnothing(T)
        if isnothing(surface_prognostic_values)
```

src/surface_conditions/surface_conditions.jl
```diff
@@ -141,6 +146,8 @@ function surface_state_to_conditions(
             # Assume that the latitude is 0.
             T = FT(300)
         end
+    else
+        T = FT(sfc_field.T)
```
dennisYatunin, line 150:
```suggestion
    else
        T = surface_prognostic_values.T
    end
```

src/surface_conditions/surface_conditions.jl
dennisYatunin, line 140:
```suggestion
    if isnothing(surface_prognostic_values)
        if isnothing(T)
```

src/surface_conditions/surface_conditions.jl
dennisYatunin, line 150:
```suggestion
    else
        T = surface_prognostic_values.T
    end
```

src/solver/types.jl
```diff
@@ -18,6 +18,10 @@ struct SphericalModel <: AbstractModelConfig end
 struct BoxModel <: AbstractModelConfig end
 struct PlaneModel <: AbstractModelConfig end
 
+abstract type AbstractSurfaceTemperature end
+Base.@kwdef struct NoSurfaceTemperature <: AbstractSurfaceTemperature end
```
charleskawczynski, line 22:
```suggestion
struct NoSurfaceTemperature <: AbstractSurfaceTemperature end
```

src/solver/types.jl
```diff
@@ -18,6 +18,10 @@ struct SphericalModel <: AbstractModelConfig end
...
+Base.@kwdef struct PrognosticSurfaceTemperature <: AbstractSurfaceTemperature end
```
charleskawczynski, line 23:
```suggestion
struct PrognosticSurfaceTemperature <: AbstractSurfaceTemperature end
```
```suggestion
Base.@kwdef struct PrognosticSurfaceTemperature <: AbstractSurfaceTemperature end
```

charleskawczynski: (No need for kwdef if it’s a singleton)

src/solver/types.jl
```diff
@@ -19,8 +19,10 @@ struct BoxModel <: AbstractModelConfig end
 struct PlaneModel <: AbstractModelConfig end
 
 abstract type AbstractSurfaceTemperature end
-Base.@kwdef struct NoSurfaceTemperature <: AbstractSurfaceTemperature end
-Base.@kwdef struct PrognosticSurfaceTemperature <: AbstractSurfaceTemperature end
+# Base.@kwdef struct NoSurfaceTemperature <: AbstractSurfaceTemperature end
+struct NoSurfaceTemperature <: AbstractSurfaceTemperature end
```
szy21, line 23:
```suggestion
struct PrescribedSurfaceTemperature <: AbstractSurfaceTemperature end
```
(I think it is more clear than NoSurfaceTemperature)

src/initial_conditions/atmos_state.jl
```diff
@@ -69,6 +70,14 @@ energy_variables(ls, ::TotalEnergy) = (;
     )
 )
 
+atmos_surface_field(surface_space, config, ::NoSurfaceTemperature) = 
+    (;)
+function atmos_surface_field(surface_space, config, ::PrognosticSurfaceTemperature)
+    surface_coords = Fields.coordinate_field(surface_space)
+    #default prognostic surface temperature set to 300, could change later 
+    return (; sfc = map(_ -> (; T = eltype(config)(300)), surface_coords))
```
szy21, line 78: Is `config` here only for getting FT? @dennisYatunin Could we get FT from `surface_space`, so we don't need to pass in `config`?

src/solver/types.jl
```diff
@@ -241,6 +248,7 @@ Base.@kwdef struct AtmosModel{
     vert_diff::VD = nothing
     viscous_sponge::VS = nothing
     rayleigh_sponge::RS = nothing
+    prognostic_surface_model::PS = nothing
```
szy21, line 251:
```suggestion
    surface_model::SM = nothing
```
I think we can just call it `surface_model`, and it is either `PrescribedSurfaceTemperature` or `PrognosticSurfaceTemperature`. @dennisYatunin What do you think?

src/surface_conditions/surface_conditions.jl
```diff
@@ -131,16 +136,20 @@ function surface_state_to_conditions(
 
     (!isnothing(q_vap) && atmos.moisture_model isa DryModel) &&
         error("surface q_vap cannot be specified when using a DryModel")
-
-    if isnothing(T)
+    if isnothing(sfc_prognostic_values)
```
szy21, line 139: @dennisYatunin Will this affect the coupler behavior (I think not but want to double check)?

src/surface_conditions/surface_conditions.jl
```diff
@@ -122,6 +128,7 @@ function surface_state_to_conditions(
     interior_z,
     params,
     atmos,
+    sfc_prognostic_values,
```
etmaca5, line 131: @charleskawczynski I am trying to decrease the number of allocations in this PR to meet the threshold in the flame tests (it is 2% over the threshold for the default case, and lower for others). The problem seems to be that passing sfc_prognostic_values into surface_state_to_conditions causes unnecessary allocations (and I don't see a way to update surface temperature without passing sfc_prognostic_values in). Can you please let me know if you have any ideas for how to prevent extra allocations here or if it's even preventable?

charleskawczynski: It's difficult and not very fruitful to diagnose those issues until this branch is rebased, since that may impact allocations. Can you rebase and fix the git conflicts first? cc me after and I can take a look

charleskawczynski: I believe these functions need to be refactored

etmaca5: @charleskawczynski I've now rebased, and the number of allocations has not changed

etmaca5: @charleskawczynski what exactly do you think should be refactored with these functions apart from the if statements?

charleskawczynski: It's not a matter of this PR, the function is already allocating and allocations seem to easily change when adjusting this function, which makes me think that it's not being inferred well. @dennisYatunin, can you please help @etmaca5 with the allocations?

One question that I have is: why is this function computing `T` when it's being passed in via `surface_state`? I've seen situations in the past where the compiler

src/solver/cli_options.jl
```diff
@@ -86,6 +86,10 @@ function argparse_settings()
         help = "Vertical diffusion [`false` (default), `VerticalDiffusion`, `true` (defaults to `VerticalDiffusion`)]"
         arg_type = String
         default = "false"
+        "--prognostic_surface"
+        help = "Determines if surface temperature is prognostic [`false` (default), , `true`, `PrognosticSurfaceTemperature`, `NoSurfaceTemperature`, `PrescribedSurfaceTemperature`]"
```
szy21, line 90:
```suggestion
        help = "Determines if surface temperature is prognostic [`false` (default), , `true`, `PrognosticSurfaceTemperature`, `PrescribedSurfaceTemperature`]"
```

src/solver/types.jl
```diff
@@ -285,6 +290,7 @@ Base.@kwdef struct AtmosModel{
     viscous_sponge::VS = nothing
     rayleigh_sponge::RS = nothing
     sfc_temperature::ST = nothing
+    surface_model::PS = nothing
```
szy21, line 293:
```suggestion
    surface_model::SM = nothing
```

src/surface_conditions/surface_conditions.jl
```diff
@@ -21,50 +21,151 @@ function update_surface_conditions!(Y, p, t)
...
-
     (; ᶜts, ᶜu, sfc_conditions, sfc_setup, params, atmos) = p
     int_ts_values = Fields.field_values(Fields.level(ᶜts, 1))
     int_u_values = Fields.field_values(Fields.level(ᶜu, 1))
     int_z_values =
         Fields.field_values(Fields.level(Fields.coordinate_field(Y.c).z, 1))
     sfc_conditions_values = Fields.field_values(sfc_conditions)
 
+    is_prognostic_sfc = atmos.surface_model isa PrognosticSurfaceTemperature
+    possible_sfc_prognostic_temp =
```
szy21, line 32: Any reason why you have a `possible` in the name?

src/surface_conditions/surface_conditions.jl
```diff
@@ -21,50 +21,151 @@ function update_surface_conditions!(Y, p, t)
...
-            sfc_setup_values,
-            sfc_local_geometry_values,
-            int_ts_values,
-            projected_vector_data(CT1, int_u_values, int_local_geometry_values),
-            projected_vector_data(CT2, int_u_values, int_local_geometry_values),
-            int_z_values,
-            params,
-            atmos,
-        )
+        if (is_prognostic_sfc)
```
szy21, line 87: @dennisYatunin Would `sfc_setup` only be a `Field` when running from the coupler? If so I suggest that we ignore `is_prognostic_sfc` in this case, as the coupler should always overwrite surface temperature.

szy21 (approved):
Thanks Etienne! This looks good overall, I left some minor comments (for you and for Dennis).

src/surface_conditions/surface_conditions.jl
etmaca5: It's meant to explain sfc_prognostic_temp is nothing in many cases, although I'm not sure it's necessary. What do you think?

szy21: I think just something like `prognostic_sfc_temp` would be good.

src/solver/model_getters.jl
```diff
@@ -63,6 +63,19 @@ function get_vertical_diffusion_model(
     end
 end
 
+function get_surface_model(parsed_args, ::Type)
```
charleskawczynski, line 66:
```suggestion
function get_surface_model(parsed_args)
```

src/solver/type_getters.jl
```diff
@@ -73,6 +73,7 @@ function get_atmos(config::AtmosConfig, turbconv_params)
         viscous_sponge = get_viscous_sponge_model(parsed_args, FT),
         rayleigh_sponge = get_rayleigh_sponge_model(parsed_args, FT),
         sfc_temperature = get_sfc_temperature_form(parsed_args),
+        surface_model = get_surface_model(parsed_args, FT),
```
charleskawczynski, line 76:
```suggestion
        surface_model = get_surface_model(parsed_args),
```

src/surface_conditions/surface_conditions.jl
```diff
@@ -21,25 +21,66 @@ function update_surface_conditions!(Y, p, t)
...
+                ),
+                projected_vector_data(
+                    CT2,
+                    int_u_values,
+                    int_local_geometry_values,
+                ),
+                int_z_values,
+                params,
+                atmos,
+                sfc_prognostic_temp,
```
charleskawczynski, line 58: This function call appears to have too many arguments

src/surface_conditions/surface_conditions.jl
```diff
@@ -55,16 +96,46 @@ function update_surface_conditions!(Y, p, t)
...
```
charleskawczynski, line 117: Same here

src/prognostic_equations/surface_temp.jl
```diff
@@ -0,0 +1,13 @@
+surface_temp_tendency!(Yₜ, Y, p, t, ::PrescribedSurfaceTemperature) = nothing
+
+function surface_temp_tendency!(Yₜ, Y, p, t, ::PrognosticSurfaceTemperature)
+    depth = 10 # ocean mixed layer depth [m]
+    ρ_ocean = 1020 # ocean density [kg / m³]
+    cp_ocean = 4184 # ocean heat capacity [J/(kg * K)]
```
charleskawczynski, line 6: Should we add these as input parameters? cc @tapios

src/surface_conditions/surface_conditions.jl
etmaca5: I added a default argument to the surface_state_to_conditions() function, which is why the function calls have a different number of arguments. Is this okay with you?

charleskawczynski: Adding the argument is fine, can we reduce the code duplication in this if-else block?

(
```julia
                sfc_setup(
                    sfc_local_geometry_values.coordinates,
                    int_z_values,
                    t,
                ),
                sfc_local_geometry_values,
                int_ts_values,
                projected_vector_data(
                    CT1,
                    int_u_values,
                    int_local_geometry_values,
                ),
                projected_vector_data(
                    CT2,
                    int_u_values,
                    int_local_geometry_values,
                ),
                int_z_values,
                params,
                atmos,
                sfc_prognostic_temp,
```
)

etmaca5: sounds good, thanks Charles!

szy21: I think we had this before, but it resulted in allocations?

etmaca5: I think it did, but I can try to reduce code duplication in other ways. Could we talk about it tomorrow?

src/prognostic_equations/surface_temp.jl
szy21: They should stay in a parameter file in the end. But for now I think it's fine, these are not very important parameters (and not those we need to calibrate).

src/surface_conditions/surface_conditions.jl
charleskawczynski: Sorry, I'm not sure I'll have time today. Shouldn't we be able to just use dispatch here?

etmaca5: @charleskawczynski Since this PR is so long, do you think I could merge it, then open another PR with multiple dispatch?

szy21:
It seems the aqua test failed. You can try following here for debugging: https://github.com/CliMA/ClimaAtmos.jl/blob/7dbff41636b3d42fe032ac0878545191d58e8cc9/test/aqua.jl#L19

szy21:
Actually, the errors probably have nothing to do with this PR. I see the same errors with some other PRs as well.

PR #1891, opened 2023-07-18
add mixing length to diagnostic edmf

szy21:
## Purpose 

Add mixing length with prescribed shear production for diagnostic edmf.

## Content

----


charleskawczynski (approved):
Code changes look fine to me

trontrytel (approved):

src/cache/diagnostic_edmf_precomputed_quantities.jl
```diff
@@ -458,5 +458,60 @@ function set_diagnostic_edmf_precomputed_quantities!(Y, p, t)
 
     @. ᶜu⁰ = C123(Y.c.uₕ) + ᶜinterp(C123(ᶠu³⁰))
 
+    @. ᶜlinear_buoygrad = buoyancy_gradients(
```
szy21, line 461: This function increases the allocation a bit. The allocation for `diagnostic_edmfx_aquaplanet` goes from 10704 to 11376 with this function. This is the same with prognostic edmfx I think, I just didn't try to fix it last time as the allocation only increases slightly, and the jet failure doesn't change. @charleskawczynski @nefrathenrici Anything that looks obvious to you?

src/cache/diagnostic_edmf_precomputed_quantities.jl
```diff
@@ -458,5 +458,60 @@ function set_diagnostic_edmf_precomputed_quantities!(Y, p, t)
...
+            ᶜρaʲs.:($$j) * ᶜentr_detrʲs.:($$j).detr / Y.c.ρ * (
+                1 / 2 *
+                (
+                    get_physical_w(ᶜuʲs.:($$j), ᶜlg) - get_physical_w(ᶜu⁰, ᶜlg)
+                )^2 - ᶜtke⁰
+            )
+    end
+
+    sfc_tke = Fields.level(ᶜtke⁰, 1)
+    @. ᶜmixing_length = mixing_length(
```
szy21, line 503: This function also increases the allocation a bit (again, same for progrnostic edmf, and the jet failure doesn't change). With this function the allocation increases from 10480 to 10704. The function is here: https://github.com/CliMA/ClimaAtmos.jl/blob/657e87d1fc0220d1678489129a2ec556cc7bf3a7/src/prognostic_equations/edmfx_closures.jl#L265

szy21: Actually maybe not. Let me check again.

szy21: No, this function does not allocate.

src/cache/diagnostic_edmf_precomputed_quantities.jl
```diff
@@ -458,5 +458,60 @@ function set_diagnostic_edmf_precomputed_quantities!(Y, p, t)
...
+    @. ᶜprandtl_nvec = 1
+    ᶜtke_exch = p.ᶜtemp_scalar_2
+    @. ᶜtke_exch = 0
+    for j in 1:n
+        @. ᶜtke_exch +=
+            ᶜρaʲs.:($$j) * ᶜentr_detrʲs.:($$j).detr / Y.c.ρ * (
+                1 / 2 *
+                (
+                    get_physical_w(ᶜuʲs.:($$j), ᶜlg) - get_physical_w(ᶜu⁰, ᶜlg)
+                )^2 - ᶜtke⁰
```
szy21, line 498: It seems some of the allocation is from the ^2 here, but I think this issue has been solved? @charleskawczynski

szy21:
I'm going to increase the allocation limit for now as we have a tight deadline. The allocation for diagnostic edmf is not very large, so it should be fine.

src/cache/diagnostic_edmf_precomputed_quantities.jl
charleskawczynski: Yeah, we can increase the allocation limit for now-- these should go away after the next ClimaCore update.

Issue #1259, opened 2023-01-21
Refactoring Parameter Usage in CA


nefrathenrici:
This SDI continues the discussion from [SDI#721](https://github.com/CliMA/ClimaAtmos.jl/issues/721).
## Purpose

Currently, ClimaParameters is in a decent initial state as a single source for parameters, but there a few issues. 
1. In ClimaAtmos, there are currently three sources of parameters: ClimaParameters, NameList, and the CLI. This will ideally be reduced to just ClimaParameters. The CLI should just handle configuration for running the model, and should eventually also support a TOML file.
2. A model running a simple simulation should not have to gather all default parameters used in all components.
3. The growing numbers of parameters should not become unwieldy.
4. Parameters and configuration should be clearly logged for reproducibility.

This SDI is meant to further discussion around these issues and propose changes.

## Summary of Proposed Steps

1. Remove [NameList](https://github.com/CliMA/ClimaAtmos.jl/blob/main/tc_driver/generate_namelist.jl) usage, add the parameters to ClimaParameters.
2. Use ArgParse to group common configurations in order to simplify the CI.
3. Add support to store a model configuration in a TOML file, separate from the parameter TOML file.
4. Use a tagging system to ‘tag’ each parameter with the component(s) that it is used in, allowing for only sufficient parameters to be grabbed from ClimaParameters.
5. Continue using parameter structures to read in parameters from the flat ClimaParameters TOML and define derived parameters.
6. Log config and override parameters to the same place.

## NameList

This is the easy part. The NameList dictionary will be phased out and replaced with TOML. [Basic functionality](https://github.com/CliMA/ClimaAtmos.jl/pull/812) to convert NameList dicts to TOML already exists. With this in place, we can add all of the default parameters from NameList into the main `parameters.toml` in ClimaParameters and add a parameter structure to replace them. Supported TOML file types may need to be expanded.

This can act as a pilot/example for any changes we want to make to the ClimaParameters interface, such as the underlying TOML structure.

## Parameter Categorization
Models should not need to load in the whole TOML dictionary for running simple simulations or test cases. Two possible solutions to this:

1. Split the big TOML file into multiple files, at the very least one for each high-level component. This is simple, but it may be difficult to split parameters neatly.

[...]

PR #1900, opened 2023-07-20
Remove CLI, replace with file-based config

nefrathenrici:
### PR Summary:
Removed the command line argparsing, and replaced it with YAML files using the same key/value format. Changed the utilities and tests that depend on the current cli.

### Contents:
Added default configuration files which store model argument values, types (implicitly), and descriptions.  There is one overall default file and one edmf default file: `.buildkite/default_config.yml` and `.buildkite/default_edmf_config.yml`.

Added config files for all buildkite jobs: normal model runs, longruns, performance, and mpi jobs.

Updated the CI pipeline YAML and the longrun YAML to use the config files and improved formatting.

Updated documentation with guides for running the model with custom configurations and parameters. Added OrderedCollections to docs dependencies to maintain ordering and improve readability when generating the default configuration table.

Removed the majority of the `cli_options.jl` interface and ported functionality to work with the new config system. The new functions can now be found in `src/solver/yaml_helper.jl`
- `job_id_from_parsed_args` is now `job_id_from_config`, otherwise it should produce the same result
- `print_repl_script` is now in `docs/src/repl_scripts.jl`, and is slightly tweaked.
- `parsed_args_per_job_id` is now `configs_per_job_id`, and it is tweaked to read in all job-specific YAML files from the `.buildkite` folder. I've added an optional argument to include performance configs. The `filter_string` kwarg is now called `filter` and takes a key-value pair instead of a string.
- `non_default_command_line_flags_parsed_args` is now `non_default_config_entries`
Added several other config-related utilities to `yaml_helper.jl`:
- `default_config_dict` returns the default config, read in from `.buildkite/default_config.yml` and `.buildkite/default_edmf_config.yml`
- `config_from_target_job` gets the config of the passed in `job_id`
- `override_default_config()` takes a vector of `Dict`s or filepaths and returns a `Dict` with the default config overriden by the given dicts or parsed YAML files. This is meant to generate the actual config used by the model. All config values are typecast to the type of the default config value. The function will error if a key that is not in the default config is passed in, which may pose issues for the coupler if they pass in extra arguments in the config.
- Removed `commands_from_yaml`. The closest function is now `configs_per_job_id`.
For `src/solver/types.jl`:
- `AtmosCoveragePerfConfig` now longer takes `parsed_args` and `AtmosCoveragePerfParsedArgs` has been removed. Otherwise, the precedence for options in `AtmosCoverPerfConfig` is the same.
- `AtmosConfig()` has been changed. The priority for setting the config is as follows: if an optional Dict `config_dict` is passed in, that will be used to override the defaults. If `parsed_args` contains a `config_file`, that file will be read in and override the defaults. Otherwise, the defaults will be used. TOML files can no longer provide model configuration.
- `merge_parsed_args_with_toml()` has been removed.

Removed old parameter tests, since we no longer need to test `merge_parsed_args_with_toml`

Updated coupler compatibility tests

Removed `toml_edmf_bomex` job, since this is no longer functional.

Very slightly increased allocs on two jobs (max 1.0006x increase)

Updated `perf` scripts for the new config scheme.

There is still a bit of work to do, but this PR is already large:
- Clean up the configuration. Some arguments should be parameters. Some arguments can be merged with others. We could also experiment with a more hierarchical config, since the YAML structure makes this easy. Some arguments are currently strings and should be booleans. The typechecking could also be improved.
- Change `parsed_args` to `config_dict` or something similar, since the name is no longer accurate.

### Original Proposal

This is a proposal for file-based model configuration and a request for feedback. The immediate idea is to parse a flat file instead of the command line. Some basic examples have already been completed for EDMF cases.
### Motivation
The CLI has grown large and unwieldy. Buildkite jobs are hard to manually parse through. As we continue development, the CLI will only grow and get more complex. We also need to be able to drive the model from a file for calibration, but the logic of this is unnecessarily complicated because we need to support both the CLI and a file config.
For more background on this, see [here](https://github.com/CliMA/ClimaAtmos.jl/issues/721#issuecomment-1226559829) and [here](https://github.com/CliMA/ClimaAtmos.jl/issues/1259).
### Primary Goals
- The new model config must be easy to change.
- Configuration settings must be easily found in the documentation.
- Existing internals (`yaml_helper`, perf config) that rely on the current argument parsing must not be broken. 
- Model parameters must be separate from the model configuration. Model parameters must only be changed through TOML files. 

### Summary
I want to remove the bulk of the command line arguments and replace it with YAML files. YAML files are easily read by both humans and parsers, and YAML.jl is already a dependency.
There is a default config in `.buildkite/model_configs/default_config.yml`. This contains a list of configuration entries with default values and help messages taken directly from the existing CLI. This allows us to continue autogenerating the configuration documentation. It also implicitly stores the type information of each config argument. It would also be simple to introduce nested configurations. This could be immediately useful for something like the rayleigh sponge arguments.
Here is the top of the default config:
```
test_edmf_consistency:
  help: "Test edmf equation consistency [`false` (default), `true`]"
  value: false
surface_thermo_state_type:
  help: "Surface thermo state type [`GCMSurfaceThermoState` (default), `PrescribedThermoState`]"
  value: "GCMSurfaceThermoState"
topo_smoothing:
  help: "Choose whether to order-2 smoothing on the LGL mesh"
  value: false
hyperdiff:
  help: "Hyperdiffusion [`ClimaHyperdiffusion` (or `true`; default), `none` (or `false`)]"
  value: "ClimaHyperdiffusion"
```
This config is automatically read in by the current `AtmosConfig()`, and is overridden by "override config file(s)" passed in via the `--config_file` argument. 
The override config files have simple YAML syntax so that they are not cumbersome to create. Here is the file `.buildkite/model_configs/edmf/edmf_common.yml`:
```
apply_limiter: false
config: "column"
hyperdiff: "false"
debugging_tc: true
turbconv: "edmf"
FLOAT_TYPE: "Float64"
```
The values of these key-value pairs are typecast to the types found in the default config to prevent silent typing issues.
Multiple files can be passed in, allowing for common parameters to be used in multiple buildkite jobs with case-specific overrides.
I've fleshed this out for most of the EDMF cases already.
The main code changes for this are pretty minimal - since the CLI produced a dictionary, this just parses and overrides the default config file in `AtmosConfig()`, slotting it into the `parsed_args`. Some parts will need to be changed, but I tried to keep this as a minimal example.

There is still a good amount of work to do, but I would like feedback before I spend time on the following:
- [x] Finish the rest of the config files for the rest of the buildkite tests
- [x] Change certain parts of the `parsed_args` flow, like the `comms_ctx`
- [x] Change the various tools that rely on parsing the current buildkite files and command-line arguments. 
- [x] Add documentation on using YAML files for configuration and TOML files for parameter overrides.


.buildkite/model_configs/default_config.yml
```diff
@@ -0,0 +1,295 @@
+test_edmf_consistency:
```
charleskawczynski, line 1: So, this is replacing ArgParse?

charleskawczynski: Can we put this one in a separate folder for easier findability?

nefrathenrici: Yes, the goal to (almost) totally get rid of ArgParse, the only remaining entry being the `--config_file` argument.

szy21:
What would be the relationship between these YAML files and the TOML files (e.g. bomex.toml in the current codebase)?

nefrathenrici:
> What would be the relationship between these YAML files and the TOML files (e.g. bomex.toml in the current codebase)?

@szy21 I will get rid of the bomex.toml file, since the TOML files will just be for adding parameters (not model configuration). Hopefully there will be a YAML config file for each testcase in the CI.

tapios (commented):
This looks like a great improvement over the current state. 

We should just organize the various configuration parameteres in a logical way in the file, e.g., collect all resolution parameters in a block, all ODE parameters in a block etc. 

A few of the parameters (e.g., for EDMF) should not live in the config file but where other adjustable parameters live. 

I suggested some changes in this direction. Even more structure, which helps users (who should not be forced to search a file, e.g., for vertical resolution parameters), would be possible.

perf/benchmark_dump.jl
```diff
@@ -15,21 +15,18 @@ using CUDA
...
-s = CA.argparse_settings()
-parsed_args = CA.parse_commandline(s);
-output_dir = joinpath(parsed_args["job_id"])
-
-# Set non-varying arguments
-parsed_args["z_elem"] = 50
-parsed_args["dt"] = "50secs"
+parsed_args = CA.parse_commandline(CA.argparse_settings())
+config_dict = merge(YAML.load_file.(parsed_args["config_file"])...)
+output_dir = joinpath(config_dict["job_id"])
```
charleskawczynski, line 22: Where is the toml for this job?

perf/flame.jl
```diff
@@ -58,9 +58,9 @@ allocs_limit["flame_perf_target"] = 4384
 allocs_limit["flame_perf_target_tracers"] = 185968
 allocs_limit["flame_perf_target_edmfx"] = 298688
 allocs_limit["flame_perf_target_diagnostic_edmfx"] = 631408
-allocs_limit["flame_perf_target_edmf"] = 7467397968
```
charleskawczynski, line 61: Why are these changing?

charleskawczynski: This suggests that you're not reproducing the same conditions that are in CI

post_processing/remap/remap_helpers.jl
```diff
@@ -357,6 +357,12 @@ function remap2latlon(filein, data_dir, remap_tmpdir, weightfile, nlat, nlon)
         rad_flux_clear_variables,
         draft_variables,
     )
-    apply_remap(datafile_latlon, datafile_cc, weightfile, netcdf_variables)
+    apply_remap(
+        datafile_latlon,
+        datafile_cc,
+        weightfile,
+        netcdf_variables,
+        verbose = true,
```
charleskawczynski, line 365: We probably don't want to leave this as `verbose=true`

regression_tests/mse_tables.jl
```diff
@@ -128,18 +128,6 @@ all_best_mse["edmf_bomex"][(:c, :turbconv, :up, 1, :ρae_tot)] = 0
 all_best_mse["edmf_bomex"][(:c, :turbconv, :up, 1, :ρaq_tot)] = 0
 all_best_mse["edmf_bomex"][(:f, :turbconv, :up, 1, :w, :components, :data, 1)] = 0
 #
-all_best_mse["toml_edmf_bomex"] = OrderedCollections.OrderedDict()
```
charleskawczynski, line 131: Why is this getting removed?

src/parameterized_tendencies/gravity_wave_drag/orographic_gravity_wave_helper.jl
```diff
@@ -13,8 +13,9 @@ using Statistics: mean
     - earth_radius: radius of the Earth
 """
 function calc_orographic_tensor(elev, χ, lon, lat, earth_radius)
-    @info "Comuting T tensor..."
+    @info "Computing T tensor..."
     FT = eltype(elev)
+    @show FT
```
charleskawczynski, line 18:
```suggestion
```

src/solver/yaml_helper.jl
```diff
@@ -1,24 +1,149 @@
...
-    cmds = map(x -> x isa AbstractString ? x : join(x), cmds)
-    if filter_name ≠ nothing
-        filter!(x -> occursin(filter_name, x), cmds)
+    default_config_dict()
+
+Loads the default configuration from files into a Dict for use in AtmosConfig().
+"""
+function default_config_dict()
+    default_config_file =
+        joinpath(pkgdir(CA), ".buildkite", "default_config.yml")
```
charleskawczynski, line 11:
```suggestion
        joinpath("..", "..", ".buildkite", "default_config.yml")
```
I think that's the right way to do the relative path? Or, we could try to figure out the `pkgdir(CA)` in a safe way (please see https://github.com/CliMA/policies/wiki/Software-Design-Patterns-(SDPs)#avoid-using-somepkg--import-somepkg-in-src).

test/coupler_compatibility.jl
```diff
@@ -33,10 +33,12 @@ const T2 = 290
...
 
 @testset "Coupler Compatibility (Hacky Version)" begin
-    # Initialize a model. The value of --surface_setup is irrelevant, since it
+    # Initialize a model. The value of surface_setup is irrelevant, since it
     # will get overwritten.
-    config = CA.AtmosConfigArgs(;
-        args = ["--initial_condition", "DryBaroclinicWave"],
+    config = CA.AtmosConfig(;
+        config_dict = Dict{Any, Any}(
+            "initial_condition" => "DryBaroclinicWave",
```
charleskawczynski, line 40: I love this!

src/solver/yaml_helper.jl
```diff
@@ -1,24 +1,149 @@
...
+end
+
+
+"""
+    non_default_config_entries(config)
+
+Given a configuration Dict, returns a Dict of the non-default values.
+"""
+function non_default_config_entries(config)
+    defaults = default_config_dict()
```
charleskawczynski, line 72:
```suggestion
function non_default_config_entries(config, defaults = default_config_dict())
```
I think this just tends to be more flexible

src/solver/types.jl
```diff
@@ -375,123 +375,58 @@ struct AtmosConfig{FT, TD, PA, C}
...
-    s = argparse_settings();
-    args = String[],
-    parsed_args = parse_commandline(args, s),
-    comms_ctx = get_comms_context(parsed_args),
-)
-    @info "Running ClimaAtmos with default argparse settings + $args"
-    return AtmosConfig(s; parsed_args, comms_ctx)
+function AtmosCoveragePerfConfig(s = argparse_settings())
+    parsed_args = parse_commandline(s)
+    config_dict = merge(YAML.load_file.(parsed_args["config_file"])...)
```
charleskawczynski, line 390: This is a bit confusing, is `config_file` a list of files? This is broadcasting `load_file` over it and then merging the dicts

src/solver/types.jl
```diff
@@ -375,123 +375,58 @@ struct AtmosConfig{FT, TD, PA, C}
...
+    return AtmosConfig(; s, config_dict = config_dict)
 end
 
-function AtmosConfig(
-    s = argparse_settings();
+function AtmosConfig(;
+    s = argparse_settings(),
     parsed_args = parse_commandline(s),
-    comms_ctx = get_comms_context(parsed_args),
+    config_dict = nothing,
```
charleskawczynski, line 406: Just to note, changing the kwarg name is a breaking change (which is fine, just want to note it), I've added the breaking label.

src/solver/types.jl
nefrathenrici: I think `config_file` can contain either be a single filepath or a list of filepaths. I thought that might be more flexible, but now it seems like it should be restricted to just one.

src/solver/yaml_helper.jl
```diff
@@ -1,24 +1,149 @@
...
+    for override_config in override_configs
+        for (k, v) in override_config
+            !haskey(default_config, k) &&
+                error("Key `$k` is not in the default configuration.")
+            default_type = typeof(default_config[k])
+            config[k] = isnothing(default_config[k]) ? v : default_type(v)
+        end
+    end
+    return config
+end
```
charleskawczynski, line 63: It might be nice if there was a version that did not operate over a vector input

src/solver/yaml_helper.jl
```diff
@@ -1,24 +1,149 @@
...
+
+"""
+    config_from_target_job(target_job)
+
+Given a job id string, returns the configuration for that job.
+Does not include the default configuration dictionary.
+"""
+function config_from_target_job(target_job)
+    for entry in configs_per_job_id(joinpath(pkgdir(CA), ".buildkite"))
```
charleskawczynski, line 34: Use of `CA` (see other comments about `import ClimaAtmos as CA` from within `src/`).

perf/tabulate_perf_summary.jl
```diff
@@ -27,19 +27,20 @@ all metrics can be found in `get_summary`.
...
+    ]
     return job_ids
 end
 
 function combine_PRs_performance_benchmarks(path)
     job_ids = get_job_ids(
-        joinpath(ca_dir, ".buildkite", "pipeline.yml");
-        filter_name = "--perf_summary true",
+        joinpath(ca_dir, ".buildkite");
+        filter = ("perf_summary" => true),
```
charleskawczynski, line 43: I prefer the former name (`filter_name`), or something similarly descriptive to avoid overwriting the existing Base function (`Base.filter`)

charleskawczynski:
From the REPL scripts:

```julia
### Buildkite job `longrun_aquaplanet_rhoe_equil_gray`

using Revise; import ClimaAtmos as CA;

config = Dict();
config["moist"] = "equil";
config["dt_save_to_disk"] = "10days";
config["dt"] = "450secs";
config["surface_setup"] = "DefaultExchangeCoefficients";
config["fps"] = 30;
config["t_end"] = "600days";
config["vert_diff"] = "true";
config["precip_model"] = "0M";
config["dt_save_to_sol"] = "1days";
config["job_id"] = "longrun_aquaplanet_rhoe_equil_gray";
config["rad"] = "gray";

```
won't work. I think we need to either change the driver logic:
```julia
if !(@isdefined config)
    config = CA.AtmosConfig()
end
```
or maybe define `CA.get_integrator(::AbstractDict)` so that it calls `CA.get_integrator(CA.AtmosConfig(config::Dict))` or something like that.

perf/benchmark_dump.jl
nefrathenrici: The toml for this job is [here](https://github.com/CliMA/ClimaAtmos.jl/blob/82f828cd29a0cc51e97b60299ef826e3fa039a5d/.buildkite/perf_configs/gpu_implicit_barowave_wrt_h_elem.yml) and it's used in the pipeline [here](https://github.com/CliMA/ClimaAtmos.jl/blob/82f828cd29a0cc51e97b60299ef826e3fa039a5d/.buildkite/pipeline.yml#L893-L899).

charleskawczynski: Ah, ok, thanks, I just didn't grep it properly.

nefrathenrici:
@charleskawczynski 

For the repl script, what about this?
```
using Revise; import ClimaAtmos as CA;

config_dict = Dict();
config_dict["moist"] = "equil";
config_dict["dt_save_to_disk"] = "10days";
config_dict["dt"] = "450secs";
config_dict["surface_setup"] = "DefaultExchangeCoefficients";
config_dict["fps"] = 30;
config_dict["t_end"] = "600days";
config_dict["vert_diff"] = "true";
config_dict["precip_model"] = "0M";
config_dict["dt_save_to_sol"] = "1days";
config_dict["job_id"] = "longrun_aquaplanet_rhoe_equil_gray";
config_dict["rad"] = "gray";

config = CA.AtmosConfig(config_dict);

```
It might still be nice to make a convenience function for `get_integrator`

charleskawczynski:
> For the repl script, what about this? ....

That looks good to me.

src/solver/yaml_helper.jl
```diff
@@ -1,24 +1,158 @@
...
 end
+
+"""
+    job_id_from_config(config)
+
+Returns a unique name (`String`) given
+ - `config` the Dict containing the model configuration
+"""
+function job_id_from_config(config)
+    defaults = default_config_dict()
```
charleskawczynski, line 131:
```suggestion
function job_id_from_config(config, defaults = default_config_dict())
```

charleskawczynski (approved):

szy21:
@nefrathenrici Could you update the script for running simulations on the central cluster as well: https://github.com/CliMA/ClimaAtmos.jl/blob/main/examples/README.md? Thanks

PR #1901, opened 2023-07-20
Add CPU/GPU reference test

### LINKED ISSUE #1864: Dry Held-Suarez: compare CPU vs GPU results
Add GPU test case for dry Held-Suarez, and compare results.

Add mechanism to check that we get the same results on both CPU and GPU.

charleskawczynski:
This PR:
 - Adds a CPU/GPU job running the dry baro wave (for a reasonable amount of physical time)
 - Adds a script, `post_processing/compare_outputs.jl`, which compares the hdf5 files of two runs, prints the compared fields, and the maximum error
 - Fixes a bug where `n_steps_per_cycle_per_cb` can lead to `InexactError` exceptions

Closes #1864.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -788,23 +788,23 @@ steps:
         agents:
           slurm_mem: 20GB
 
-  - group: "CUDA"
+  - group: "GPU"
     steps:
 
-      - label: "CUDA: CUDA implicit baro wave"
-        key: "cuda_implicit_barowave_ref"
+      - label: "GPU: GPU implicit baro wave"
+        key: "gpu_implicit_barowave_ref"
         command:
           - "julia --project -e 'using CUDA; CUDA.versioninfo()'"
-          - "julia --color=yes --project=examples examples/hybrid/driver.jl --job_id cuda_implicit_barowave_ref --dt 580secs --t_end 10days --dt_save_to_disk 2days --initial_condition DryBaroclinicWave --dt_save_to_sol Inf"
-        artifact_paths: "cuda_implicit_barowave_ref/*"
+          - "julia --color=yes --project=examples examples/hybrid/driver.jl --job_id gpu_implicit_barowave_ref --dt 580secs --t_end 10days --dt_save_to_disk 2days --initial_condition DryBaroclinicWave --dt_save_to_sol Inf"
+        artifact_paths: "gpu_implicit_barowave_ref/*"
         agents:
           slurm_gpus: 1
 
-      - label: "CUDA: compare with CPU"
-        command: "julia --color=yes --project=examples post_processing/compare_outputs.jl --output_folder_1 sphere_baroclinic_wave_rhoe/ --output_folder_2 cuda_implicit_barowave_ref --t_end 10days --compare_state false --compare_diagnostics false"
+      - label: "GPU: compare with CPU"
+        command: "julia --color=yes --project=examples post_processing/compare_outputs.jl --output_folder_1 sphere_baroclinic_wave_rhoe/ --output_folder_2 gpu_implicit_barowave_ref --t_end 10days --compare_state false --compare_diagnostics false"
         depends_on:
           - "sphere_baroclinic_wave_rhoe"
-          - "cuda_implicit_barowave_ref"
+          - "gpu_implicit_barowave_ref"
 
   - group: "GPU Performance"
     steps:
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -179,8 +179,9 @@ steps:
           slurm_mem: 20GB
 
       - label: ":computer: baroclinic wave (ρe)"
+        key: sphere_baroclinic_wave_rhoe
         command:
-          - julia --color=yes --project=examples examples/hybrid/driver.jl --job_id sphere_baroclinic_wave_rhoe --dt 580secs --dt_save_to_disk 2days --initial_condition DryBaroclinicWave --regression_test true
+          - julia --color=yes --project=examples examples/hybrid/driver.jl --job_id sphere_baroclinic_wave_rhoe --dt 580secs --t_end 10days --dt_save_to_disk 2days --initial_condition DryBaroclinicWave --regression_test true
           - julia --color=yes --project=examples post_processing/remap/remap_pipeline.jl --data_dir sphere_baroclinic_wave_rhoe --out_dir sphere_baroclinic_wave_rhoe
           - julia --color=yes --project=examples post_processing/plot/plot_pipeline.jl --nc_dir sphere_baroclinic_wave_rhoe --fig_dir sphere_baroclinic_wave_rhoe --case_name dry_baroclinic_wave
           - julia --color=yes --project=examples regression_tests/test_mse.jl --job_id sphere_baroclinic_wave_rhoe --out_dir sphere_baroclinic_wave_rhoe
@@ -786,6 +787,24 @@ steps:
         artifact_paths: "edmf_bomex_consistency/*"
         agents:
           slurm_mem: 20GB
+
+  - group: "CUDA"
+    steps:
+
+      - label: "CUDA: CUDA implicit baro wave"
+        key: "cuda_implicit_barowave_ref"
+        command:
+          - "julia --project -e 'using CUDA; CUDA.versioninfo()'"
+          - "julia --color=yes --project=examples examples/hybrid/driver.jl --job_id cuda_implicit_barowave_ref --dt 580secs --t_end 10days --dt_save_to_disk 2days --initial_condition DryBaroclinicWave --dt_save_to_sol Inf"
+        artifact_paths: "cuda_implicit_barowave_ref/*"
+        agents:
+          slurm_gpus: 1
+
+      - label: "CUDA: compare with CPU"
+        command: "julia --color=yes --project=examples post_processing/compare_outputs.jl --output_folder_1 sphere_baroclinic_wave_rhoe/ --output_folder_2 cuda_implicit_barowave_ref --t_end 10days --compare_state false --compare_diagnostics false"
+        depends_on:
+          - "sphere_baroclinic_wave_rhoe"
+          - "cuda_implicit_barowave_ref"
 
   - group: "GPU Performance"
     steps:
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -358,6 +358,7 @@ steps:
           CLIMACORE_DISTRIBUTED: "MPI"
         agents:
           slurm_ntasks: 2
+          slurm_mem: 20GB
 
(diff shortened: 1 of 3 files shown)


.buildkite/pipeline.yml
```diff
@@ -787,6 +787,29 @@ steps:
...
+        command:
+          - "julia --project -e 'using CUDA; CUDA.versioninfo()'"
+          - "julia --color=yes --project=examples examples/hybrid/driver.jl --job_id cuda_implicit_barowave_ref --dt 580secs --t_end 10days --dt_save_to_disk 2days --initial_condition DryBaroclinicWave --dt_save_to_sol Inf"
+        artifact_paths: "cuda_implicit_barowave_ref/*"
+        agents:
+          slurm_gpus: 1
+
+      - label: "CUDA: CPU implicit baro wave"
+        key: "cpu_implicit_barowave_ref"
+        command: "julia --color=yes --project=examples examples/hybrid/driver.jl --job_id cpu_implicit_barowave_ref --dt 580secs --t_end 10days --dt_save_to_disk 2days --initial_condition DryBaroclinicWave --dt_save_to_sol Inf"
```
valeriabarra, line 804: Why do we need to add a new CPU baroclinic wave as a reference? Can't we use one of the baroclinic wave integration tests we already have in CI?

post_processing/compare_outputs.jl
```diff
@@ -0,0 +1,91 @@
...
+        help = "Output file for job 2"
+        arg_type = String
+        "--t_end"
+        help = "Simulation end time. This must match the target jobs"
+        arg_type = String
+        "--state_working"
+        help = "A bool indicating that the state test is working [`true` [default], `false`]"
+        arg_type = Bool
+        default = true
+        "--diagnostics_working"
```
valeriabarra, line 23: I'm a bit confusing on how a user would use this option. Do they need to know if the diagnostics is working before hand? Is this just for us? Sorry, it's not that clear from the help string.

post_processing/compare_outputs.jl
```diff
@@ -0,0 +1,91 @@
...
+end
+Y_1 = get_data(parsed_args["output_folder_1"], "Y");
+Y_2 = get_data(parsed_args["output_folder_2"], "Y");
+D_1 = get_data(parsed_args["output_folder_1"], "diagnostics");
+D_2 = get_data(parsed_args["output_folder_2"], "diagnostics");
+
+compare(a::FV, b::FV, pn0 = "") where {FV <: Fields.FieldVector} =
+    compare(true, a, b, pn0)
+function _compare(zero_error, a::FV, b::FV, pn0 = "") where {FV}
+    for pn in propertynames(a)
```
valeriabarra, line 52: Should we assert first that the lengths of the `propertynames` for both `a` and `b`, or the lengths of the FieldVectors are the same?

post_processing/compare_outputs.jl
```diff
@@ -0,0 +1,91 @@
...
+
+compare(zero_error, a::FV, b::FV, pn0 = "") where {FV <: Fields.FieldVector} =
+    _compare(zero_error, a, b, pn0)
+compare(a::F, b::F, pn0 = "") where {F <: Fields.Field} =
+    compare(true, a, b, pn0)
+function compare(zero_error, a::F, b::F, pn0 = "") where {F <: Fields.Field}
+    isempty(propertynames(a)) || return _compare(zero_error, a, b, pn0)
+    err = abs.(Array(parent(a)) .- Array(parent(b)))
+    if !(maximum(err) == 0.0)
+        @warn "--- Failed field comparison $pn0"
```
valeriabarra, line 68: Should this be an `@error` more than a warning?

.buildkite/pipeline.yml
charleskawczynski: I had thought about that, but these jobs are primarily for comparing CPU/GPU implementations, I figured it may be easier to just leave it decoupled from the other jobs

post_processing/compare_outputs.jl
charleskawczynski: Yeah, we should probably improve the name. It basically indicates that the diagnostics are successfully compared, meaning it's tested and there's zero difference between the two outputs. Maybe `test_diagnostics` or `compare_diagnostics`?

post_processing/compare_outputs.jl
charleskawczynski: They are enforced by type diagonalization

post_processing/compare_outputs.jl
valeriabarra: Oh, yes, I like `test_diagnostics` or `compare_diagnostics` much better!

post_processing/compare_outputs.jl
charleskawczynski: We could use `@error`, but we already error out if `compare` returns `false`. I'm fine with either one. Also, I'm not sure how practical it is that we have exactly zero error. The goal of the PR is to first perform the comparison. It seems that there's some differences between the CPU/GPU implementations. I imagine we'll have a separate effort to minimize this difference if it's deemed important.

.buildkite/pipeline.yml
charleskawczynski: Any suggestions?

valeriabarra: What are the main differences in the setups with the proposed job and one of the existing ones? Is it `t_end` mainly or something else?

post_processing/compare_outputs.jl
valeriabarra: Sounds reasonable

.buildkite/pipeline.yml
charleskawczynski: I don't think there are differences, actually, but I think the purpose of the other job is for visualizing the output, and maybe doing science?

post_processing/compare_outputs.jl
charleskawczynski: Ok, I'll just leave it as a warning for now

.buildkite/pipeline.yml
valeriabarra: I see. I would not be too excited to add another ad-hoc CPU test that it does not test any different functionality, but it's only for the purpose of comparing results with the GPU test. We often find ourselves in the position to reduce CI jobs. If it is ok with @szy21 , it's fine by me.

valeriabarra:
> This PR:

> Closes #1864.

Just note that this PR does not really close #1864 (which talks about HS test). Either we modify the Issue description or we open a new one for BW.

.buildkite/pipeline.yml
simonbyrne: Can we just use the results from the existing job?

charleskawczynski: Sure, I can delete the duplicate

valeriabarra:
Some of the errors in this job CI are pretty high now, e.g., `maximum(err) = 11.485352f0`. Are we sure we don't want to `error` in that case? We should probably decide a tolerance for `@warn` Vs error

.buildkite/pipeline.yml
```diff
@@ -787,6 +788,24 @@ steps:
         agents:
           slurm_mem: 20GB
 
+  - group: "CUDA"
+    steps:
+
+      - label: "CUDA: CUDA implicit baro wave"
```
valeriabarra, line 794: I'm noticing that we labeled other GPU jobs with the `GPU` label rather than `CUDA`. I think we should be consistent.

.buildkite/pipeline.yml
```diff
@@ -787,6 +788,24 @@ steps:
...
+        key: "cuda_implicit_barowave_ref"
+        command:
+          - "julia --project -e 'using CUDA; CUDA.versioninfo()'"
+          - "julia --color=yes --project=examples examples/hybrid/driver.jl --job_id cuda_implicit_barowave_ref --dt 580secs --t_end 10days --dt_save_to_disk 2days --initial_condition DryBaroclinicWave --dt_save_to_sol Inf"
+        artifact_paths: "cuda_implicit_barowave_ref/*"
+        agents:
+          slurm_gpus: 1
+
+      - label: "CUDA: compare with CPU"
```
valeriabarra, line 803: Same as above.

charleskawczynski:
> Some of the errors in this job CI are pretty high now, e.g., `maximum(err) = 11.485352f0`. Are we sure we don't want to `error` in that case? We should probably decide a tolerance for `@warn` Vs error

Agreed. I think this is part of the next step, though: determine why we have large differences between the implementations. For now, merging this will serve as a monitor

.buildkite/pipeline.yml
charleskawczynski: sure, we can change the name

simonbyrne:
@valeriabarra would you be able to look at the files to see what the big difference is?

valeriabarra:
> @valeriabarra would you be able to look at the files to see what the big difference is?

Which file do you mean? This is currently using the `driver.jl` file with the `DryBaroclinicWave` IC. It is generating huge differences in `uₕ` (both components) of the order of `maximum(err) = 11921.125f0` and `maximum(err) = 11967.109f0` (respectively), `maximum(err) = 11.485352f0` for `ρe_tot` and `13.78125f0` for `specific_enthalpy`. Other fields have less remarkable differences but still sensitive

valeriabarra:
The physical velocities have a less remarkable difference, but still notable

simonbyrne:
@valeriabarra can you plot them, and the difference?

post_processing/compare_outputs.jl
valeriabarra: I would still at least make this a soft fail in the Buildkite Job, otherwise it is misleading in my opinion

valeriabarra:
I checked the plots of the GPU and CPU job, separately, and there are no noticeable differences. 

- GPU: 
    - `rho`, day 10, level 1:
rho_day10_level1

   - `rho_e_tot`, day 10, level 1:
rho_e_tot_day10_level1

   - `u`, day 10, level 1:
u_day10_level1

   - `w`, day 10, level 1+half:
w_day10_level1

- CPU: 
    - `rho`, day 10, level 1:
rho_day10_level1

   - `rho_e_tot`, day 10, level 1:
rho_e_tot_day10_level1


   - `u`, day 10, level 1:
u_day10_level1


   - `w`, day 10, level 1+half:
w_day10_level1

But I am not sure how to plot the difference between the different fields in the same process. I have to figure out how to reconfigure the `comms_ctx` in `AtmosConfig` to be able to rerun using the CPU/GPU and store the different CPU/GPU solution fields to plot the difference in the same process. 

But these plot look identical with the eyeball norm.

charleskawczynski:
@valeriabarra, can you take over this PR and merge when you're ready?

valeriabarra:
> @valeriabarra, can you take over this PR and merge when you're ready?

What is left to do?

charleskawczynski:
> What is left to do?

I think it's ready to be merged unless you prefer to make any changes

Issue #1897, opened 2023-07-19
Add mixing length for diagnostic EDMFX


szy21:
- [x] Add diffusive SGS flux #1840 
- [x] Add mixing length with prescribed shear #1891 
- [x] Add shear

PR #1906, opened 2023-07-23
add shear production to diagnostic edmf

szy21:
## Purpose 

Adds shear production to diagnostic edmf. Only includes vertical gradients for now.
#1897 

## To-do

- [ ] fix allocation
- [x] Add rico

## Content

----

### DIFF
--- a/src/cache/diagnostic_edmf_precomputed_quantities.jl
+++ b/src/cache/diagnostic_edmf_precomputed_quantities.jl
@@ -490,8 +490,13 @@ function set_diagnostic_edmf_precomputed_quantities!(Y, p, t)
             Y.c.ρ,                                                             # ρ
         ),
     )
-    @. ᶜshear² = $(FT(1e-4))
 
+    # TODO: This is not correct with topography or with grid stretching
+    ᶠu⁰ = p.ᶠtemp_C123
+    @. ᶠu⁰ = C123(ᶠinterp(Y.c.uₕ)) + C123(ᶠu³⁰)
+    ct3_unit = p.ᶜtemp_CT3
+    @. ct3_unit = CT3(Geometry.WVector(FT(1)), ᶜlg)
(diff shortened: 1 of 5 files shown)


src/cache/diagnostic_edmf_precomputed_quantities.jl
```diff
@@ -484,9 +484,14 @@ function set_diagnostic_edmf_precomputed_quantities!(Y, p, t)
             Y.c.ρ,                                                             # ρ
         ),
     )
-    @. ᶜshear² = FT(1e-4)
+
+    ᶠu⁰ = p.ᶠtemp_C123
+    @. ᶠu⁰ = C123(ᶠinterp(Y.c.uₕ)) + C123(ᶠu³⁰)
+    ct3_unit = p.ᶜtemp_CT3
+    @. ct3_unit = CT3(Geometry.WVector(FT(1)), ᶜlg)
+    @. ᶜshear² = norm_sqr(adjoint(CA.ᶜgradᵥ(ᶠu⁰)) * ct3_unit)
```
szy21, line 492: @charleskawczynski I copied this from the old EDMF [here](https://github.com/CliMA/ClimaAtmos.jl/blob/199f4c90617bb5fb90148a946e31ce9134030627/src/TurbulenceConvection_deprecated/update_aux.jl#L415-L417) but it allocates a lot. Is there a way to not allocate here?

charleskawczynski: There's not really anything wrong with this code-- these allocations are due to ClimaCore issues, some hopefully will be fixed soon

szy21: Ah, ok. It increases the allocation by a factor of 55 though:( Does this look like the simplest way to calculate the approximated shear production (∂u∂z^2 + ∂v∂z^2 + ∂w∂z^2)?

PR #1913, opened 2023-07-24
Refactor quad loop

charleskawczynski:
This is a step towards putting the quadrature loop into the form: `quad(f, m, cov, dist)`, which evaluates an integral `int pdf(x; m, cov) .* f(x) dx`. Even just with this PR, I think the physics are more decoupled from the integration. There is a TODO added, which I'm not yet sure how to address:

```julia
σ_q_lim = -qt_mean / (sqrt2 * env_thermo_quad.a[1]) # TODO: is this correct?
```

For now,t there's an inference issue which requires us to type cast in `compute_cloud_fraction`, but we should be able to revert it once we dig into what's wrong there.

### DIFF
--- a/src/prognostic_equations/cloud_fraction.jl
+++ b/src/prognostic_equations/cloud_fraction.jl
@@ -22,13 +22,13 @@ end
 function compute_cloud_fraction(
     env_thermo_quad,
     thermo_params,
-    ᶜp,
+    ᶜp::FT,
     q_tot,
     θ_liq_ice,
     qt′qt′,
     θl′θl′,
     θl′qt′,
-)
+) where {FT}
     vars = (;
         p_c = ᶜp,
         qt_mean = q_tot,
@@ -37,18 +37,20 @@ function compute_cloud_fraction(
         θl′θl′,
         θl′qt′,
     )
-    return quad_loop(env_thermo_quad, vars, thermo_params).cf
+    return quad_loop(
+        env_thermo_quad,
+        env_thermo_quad.quadrature_type,
+        vars,
+        thermo_params,
+    )::FT
 end
 
-function quad_loop(env_thermo_quad::SGSQuadrature, vars, thermo_params)
-
-    env_len = 8
-    i_ql, i_qi, i_T, i_cf, i_qt_sat, i_qt_unsat, i_T_sat, i_T_unsat = 1:env_len
-
-    quadrature_type = env_thermo_quad.quadrature_type
-    quad_order = quadrature_order(env_thermo_quad)
-    χ = env_thermo_quad.a
-    weights = env_thermo_quad.w
+function quad_loop(
+    env_thermo_quad::SGSQuadrature,
+    quadrature_type::GaussianQuad,
+    vars,
+    thermo_params,
+)
 
     # qt - total water specific humidity
     # θl - liquid ice potential temperature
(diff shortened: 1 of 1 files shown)


src/prognostic_equations/cloud_fraction.jl
```diff
@@ -96,67 +88,53 @@ function quad_loop(env_thermo_quad::SGSQuadrature, vars, thermo_params)
...
-                inner_env[i_T_sat] += T * weights[m_h] * sqpi_inv
-            else
-                inner_env[i_qt_unsat] += qt_hat * weights[m_h] * sqpi_inv
-                inner_env[i_T_unsat] += T * weights[m_h] * sqpi_inv
-            end
-        end
-
-        for idx in 1:env_len
-            outer_env[idx] += inner_env[idx] * weights[m_q] * sqpi_inv
+            inner_env = inner_env ⊞ f(χ[m_h]) ⊠ weights[m_h] ⊠ FT(1 / sqrt(π))
```
szy21, line 135: Why do we need RecursiveApply?

szy21:
It might be good if @tapios takes a look at this as well. If I understand Tapio's comments correctly, what we want is a function f(qt_hat, thetali_hat) rather than f(χ). But qt_hat and thetali_hat are just a function of χ and mean values so maybe this is ok as well? In any case this seems to be a step towards decoupling the pdf and the physical function, thanks!

tapios:
It's going in the right direction but can go further. I want to decouple all thermodynamics from the quadrature. That is, write a function that evaluates the bivariate quadrature `int pdf(x_1, x_2, parameters; m, cov) .* f(x_1, x_2) dx_1 dx_2` (my previous x was supposed to be a 2-vector). Then evaluate this with a function such as `f(q_t, theta_li) = heaviside (q_t - q_v^*(q_t, theta_li, pars))`, where `pars`, this case, would include density. That way, the same quadrature function could be used for any combination of thermodynamic variables we choose, and no explicit thermodynamic evaluations (other than those necessary to evaluate `f`) would be needed in the quadrature function itself.

src/prognostic_equations/cloud_fraction.jl
charleskawczynski: This will allow us to compute quadrature over arbitrary variables in one pass (e.g., a vector of scalars, a NamedTuple of scalars, structs etc.)

charleskawczynski:
> It's going in the right direction but can go further. I want to decouple all thermodynamics from the quadrature. That is, write a function that evaluates the bivariate quadrature `int pdf(x_1, x_2, parameters; m, cov) .* f(x_1, x_2) dx_1 dx_2` (my previous x was supposed to be a 2-vector). Then evaluate this with a function such as `f(q_t, theta_li) = heaviside (q_t - q_v^*(q_t, theta_li, pars))`, where `pars`, this case, would include density. That way, the same quadrature function could be used for any combination of thermodynamic variables we choose, and no explicit thermodynamic evaluations (other than those necessary to evaluate `f`) would be needed in the quadrature function itself.

Agreed, I thought that this would be a good intermediate step, I think we can make the remaining changes in a follow up PR.

charleskawczynski:
Ok, inference was really problematic in that function because the compiler was boxing many things (due to redefining variables and the compiler didn't seem to like defining different closures inside union split branches). So, I removed the log normal distribution (for now), changed a few variable names, and added some type casting to fix type instabilities (`eps(Float32) isa Float64`!)

charleskawczynski:
@szy21, can you take a second look? This may make this function significantly faster too (it was previously _very_ expensive, about as expensive as `step!`!)

charleskawczynski:
In case I need to re-review, this is what I was using to debug via JET:

```julia
import Revise
import ClimaCore
import FastGaussQuadrature
import Thermodynamics as TD
import CLIMAParameters as CP
if !(@isdefined TU)
    include(
        joinpath(pkgdir(ClimaCore), "test", "TestUtilities", "TestUtilities.jl"),
    );
end
import .TestUtilities as TU;
space = TU.CenterExtrudedFiniteDifferenceSpace(Float64);
f = fill((; ᶜts = zero(TD.PhaseEquil{Float64}), ᶜp = Float64(101325)), space);
@. f.ᶜts = TD.PhaseEquil_ρTq(thermo_params, 1.0, 300.0, 0.001);
@. f.ᶜp = TD.air_pressure(thermo_params, f.ᶜts);


include("src/cache/precomputed_quantities.jl")
include("src/prognostic_equations/cloud_fraction.jl")

aliases = string.(fieldnames(TD.Parameters.ThermodynamicsParameters));
toml_dict = CP.create_toml_dict(Float64; dict_type = "alias");
pairs = CP.get_parameter_values!(toml_dict, aliases, "Thermodynamics");
thermo_params = TD.Parameters.ThermodynamicsParameters{Float64}(; pairs...);
Base.broadcastable(x::SGSQuadrature) = tuple(x)
env_thermo_quad = SGSQuadrature(Float64);
get_cloud_fraction(thermo_params, env_thermo_quad, f.ᶜp, f.ᶜts);

using JET
include("src/prognostic_equations/cloud_fraction.jl"); JET.@test_opt get_cloud_fraction(thermo_params, env_thermo_quad, f.ᶜp, f.ᶜts);
```

PR #1916, opened 2023-07-25
add tke equation for diagnostic edmf

### LINKED ISSUE #1896: Add TKE equations for diagnostic EDMFX

szy21:
## Purpose 

Closes #1896 

## To-do

- [ ] add boundary condition for tke
- [ ] see why it's unstable

## Content

----


src/prognostic_equations/advection.jl
```diff
@@ -173,5 +173,22 @@ function explicit_vertical_advection_tendency!(Yₜ, Y, p, t)
             end
         end
     end
+
+    # if p.atmos.turbconv_model isa DiagnosticEDMFX
```
szy21, line 177: Ignore this. Waiting for #1917

src/prognostic_equations/edmfx_tke.jl
```diff
@@ -0,0 +1,62 @@
...
+    c_m = TCP.tke_ed_coeff(turbconv_params)
+    vkc = TCP.von_karman_const(turbconv_params)
+    (; ᶜentr_detrʲs, ᶠu³ʲs) = p
+    (; ᶠu³⁰, ᶜshear², ᶜlinear_buoygrad, ᶜtke⁰, ᶜmixing_length, ᶜK_u, ᶜK_h) = p
+    (; ustar) = p.sfc_conditions
+    ᶠgradᵥ = Operators.GradientC2F()
+
+    ᶠρaK_u = p.ᶠtemp_scalar
+    ρatke_flux = p.sfc_temp_C3
+    if true #p.atmos.prognostic_tke
```
szy21, line 30: Ignore this.

src/prognostic_equations/edmfx_tke.jl
```diff
@@ -0,0 +1,62 @@
...
+        # turbulent transport
+        @. ᶠρaK_u[colidx] = ᶠinterp(Y.c.ρ[colidx]) * ᶠinterp(ᶜK_u[colidx])
+        @. ρatke_flux[colidx] = C3(FT(0))
```
szy21, line 33: Need to add surface fluxes. It's a bit annoying as 
```
ρatke_flux = WVector(Fields.level(Y.c.ρ, 1) * constant * ustar^2 * norm(Fields.level(Y.c.uₕ), 1))
```
(or convert it to Covariant3Vector, which will be more correct) But the variables are not on the same spaces (some are on centers and some are on faces), so we may need to use the `DataLayouts`? @charleskawczynski @dennisYatunin any idea?

src/prognostic_equations/advection.jl
szy21: Also, this is allocating a lot.

src/prognostic_equations/edmfx_tke.jl
tapios: This, it seems, is a generic problem with surface fluxes that should have a generic solution. 

Also, the horizontal velocity norm should be that of the velocity parallel to the surface (again, should be same for all surface fluxes). If `uₕ` consists of the covariant velocity components, it's not quite right (it should be the velocity vector consisting of the contravariant components--is that what it is?)

src/prognostic_equations/edmfx_tke.jl
```diff
@@ -0,0 +1,62 @@
...
+        ᶜdivᵥ_ρatke = Operators.DivergenceF2C(
+            top = Operators.SetValue(C3(FT(0))),
+            bottom = Operators.SetValue(ρatke_flux[colidx]),
```
tapios, line 36: It would be good to make clear that this is the divergence of just the diffusive flux, and the boundary condition only applies to that diffusive flux (not the advective flux).

src/prognostic_equations/edmfx_tke.jl
```diff
@@ -0,0 +1,62 @@
...
+        )
+        @. Yₜ.c.sgs⁰.ρatke[colidx] -= ᶜdivᵥ_ρatke(-(ᶠρaK_u[colidx] * ᶠgradᵥ(ᶜtke⁰[colidx])))
```
tapios, line 38: Maybe remove the double minus?

src/prognostic_equations/edmfx_tke.jl
```diff
@@ -0,0 +1,62 @@
...
+        # shear production
+        @. Yₜ.c.sgs⁰.ρatke[colidx] +=
+            2 * Y.c.ρ[colidx] * ᶜK_u[colidx] * ᶜshear²[colidx]
```
tapios, line 41: Is `Y.c.ρ[colidx]` including the a^(0) factor?

tapios (approved):
LGTM

src/prognostic_equations/edmfx_tke.jl
szy21: No, Y.c.uₕ is the covariant velocity (horizontal). You are right this is not correct, I just didn't write the equation carefully. Our current implementation of surface flux should be correct. `ρatke_flux` is also not a WVector (vertical), instead it should be a Covariant3Vector, but with the physical magnitude as in the equation, right?

src/prognostic_equations/edmfx_tke.jl
szy21: No, because this is only for diagnostic EDMF, where \rhoa^(0) ~ \rho_gm. For prognostic EDMF, which we haven't added, this will include the a^(0) factor.

src/prognostic_equations/edmfx_tke.jl
charleskawczynski: > But the variables are not on the same spaces (some are on centers and some are on faces), so we may need to use the DataLayouts? @charleskawczynski @dennisYatunin any idea?

On the GPU we'll need to, because `bycolumn` is functionally different (https://github.com/CliMA/ClimaCore.jl/commit/b1bdf1e6db35b9ae3fb6a4fe358f27a3df4c840d). Maybe @simonbyrne can re-visit making `bycolumn` on the GPU functionally the same (and maintain performance). Doing that would be really nice because then we can just use bycolumn and use field broadcasting (PointFields can broadcast like scalars do, so they don't need to be on the same datalayout).

src/prognostic_equations/edmfx_tke.jl
tapios: Would be nice to use the same code for both, and just set a^0=1 for diagnostic.

szy21: It's not very easy to unify prognostic and diagnostic edmf now, as they have several different variables. But it's on my list and will be one of our priorities after 8/7.

charleskawczynski (approved):

szy21:
This is not stable, but I will merge it in and try to stabilize it.

PR #1917, opened 2023-07-25
Add toggle for prognostic TKE and fix TKE advection

dennisYatunin:
## Purpose 

This PR adds the `--prognostic_tke` command-line option, which currently just toggles the advection of TKE. It also adds the vertical advection of TKE and cleans up the advection tendency functions. Related to #1916.

## Content

- Add a boolean type parameter to `EDMFX` and `DiagnosticEDFMX`, and set it to `parsed_args["prognostic_tke"]`.
- Define the `use_prognostic_tke` function to extract this type parameter (or to return `false` when there is no `turbconv_model`).
- Modify `horizontal_advection_tendency!` as follows:
    - Unify the horizontal advection of TKE for `EDMFX` and `DiagnosticEDFMX`.
    - Ensure that TKE is only advected when `use_prognostic_tke` returns `true`.
- Modify `explicit_vertical_advection_tendency!` as follows:
    - Extract all cached values at the beginning of this function.
    - Wrap all vertical operations in calls to `bycolumn`. Since there are two sets of vertical operations separated by horizontal operations, two calls to `bycolumn` are needed.
    - Add the vertical advection of TKE.
    - Ensure that TKE is only advected when `use_prognostic_tke` returns `true`.
    - Add comments to indicate which parts of this function will need to be moved to `implicit_vertical_advection_tendency!`.
- Add `--prognostic_tke true` to the `diagnostic_edmfx_aquaplanet` job. Since the initial TKE is 0 for this job, the simulation will not crash when TKE is advected.
- Remove `ᶜρa⁰` from the cache for `DiagnosticEDMFX` because it is not being used (we approximate it with `Y.c.ρ`).

----

### DIFF
--- a/src/prognostic_equations/advection.jl
+++ b/src/prognostic_equations/advection.jl
@@ -9,11 +9,11 @@ import ClimaCore.Geometry as Geometry
 function horizontal_advection_tendency!(Yₜ, Y, p, t)
     n = n_mass_flux_subdomains(p.atmos.turbconv_model)
     (; ᶜu, ᶜK, ᶜp, ᶜΦ, ᶜp_ref) = p
-    if p.atmos.turbconv_model isa EDMFX
-        (; ᶜu⁰, ᶜuʲs) = p
-    end
-    if p.atmos.turbconv_model isa DiagnosticEDMFX
+    if p.atmos.turbconv_model isa AbstractEDMF
         (; ᶜu⁰) = p
+    end
+    if p.atmos.turbconv_model isa EDMFX
+        (; ᶜuʲs) = p
     end
 
     @. Yₜ.c.ρ -= divₕ(Y.c.ρ * ᶜu)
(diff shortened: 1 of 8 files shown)


dennisYatunin:
bors try

src/prognostic_equations/advection.jl
```diff
@@ -85,8 +81,17 @@ function explicit_vertical_advection_tendency!(Yₜ, Y, p, t)
     (; ᶜu, ᶠu³, ᶜK, ᶜf) = p
     ᶜω³ = p.ᶜtemp_CT3
     ᶠω¹² = p.ᶠtemp_CT12
+    if p.atmos.turbconv_model isa AbstractEDMF
+        (; ᶠu³⁰, edmfx_upwinding) = p
+        (; dt) = p.simulation
+        ᶜρa⁰ = p.atmos.turbconv_model isa DiagnosticEDMFX ? Y.c.ρ : p.ᶜρa⁰
```
szy21, line 87: Maybe this breaks the allocation?

src/prognostic_equations/advection.jl
```diff
@@ -172,6 +181,19 @@ function explicit_vertical_advection_tendency!(Yₜ, Y, p, t)
...
         end
+
+        # TODO: Move this to implicit_vertical_advection_tendency!.
+        if use_prognostic_tke(p.atmos.turbconv_model)
+            vertical_transport!(
+                Yₜ.c.sgs⁰.ρatke[colidx],
+                ᶜJ[colidx],
+                ᶜρa⁰[colidx],
+                ᶠu³⁰[colidx],
+                Y.c.sgs⁰.ρatke[colidx],
```
szy21, line 192: This should be `tke`, not `ρatke` I think.

dennisYatunin:
bors try

dennisYatunin:
bors try

dennisYatunin:
bors try

dennisYatunin:
bors try

PR #1919, opened 2023-07-26
Fix moist baro wave

charleskawczynski:
This PR fixes the moist baro wave


charleskawczynski:
@szy21, can you try edmf on the gpu once this merges?

szy21:
Sounds good. Do I just follow the example in buildkite?

charleskawczynski:
> Sounds good. Do I just follow the example in buildkite?

Yeah, you should be able to just use the edmf example and add GPU resources (`--device = "auto"` by default, so it can have the exact same CL model flags)

szy21:
I think it's merged, but for some reason this PR is not closed.

charleskawczynski:
Yeah, that's a bit odd. closing.

PR #1920, opened 2023-07-26
surface condition and edmf hack to meet Aug 7 deadline

szy21:
## Purpose 

Sorry :(
We are not really sorry
-- Zhaoyi, Dennis, Lenka

@dennisYatunin will take this on

## Content

----

### DIFF
--- a/perf/flame.jl
+++ b/perf/flame.jl
@@ -61,7 +61,7 @@ allocs_limit["flame_perf_target_diagnostic_edmfx"] = 630384
 allocs_limit["flame_perf_target_edmf"] = 7459435344
 allocs_limit["flame_perf_target_threaded"] = 6175664
 allocs_limit["flame_perf_target_callbacks"] = 49850232
-allocs_limit["flame_perf_gw"] = 4933595680
+allocs_limit["flame_perf_gw"] = 4933596128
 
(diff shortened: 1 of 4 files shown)


szy21:
@dennisYatunin Could you add the test and merge it today?

szy21:
Actually another thing: I am not sure if this hack works for restarting a simulation (as t won't be zero), so maybe it's worth checking.

LenkaNovak:
usually `reinit!` allows you to deal with that but def worth checking before merging...

LenkaNovak:
trying this branch locally seems to work, but the `@info here` needs to be removed lol

szy21:
Oh, yes. Dennis is not here yet so I don't think he will get to it until late afternoon. I would be happy to clean it up and merge it without the test, if you want it now. Maybe Dennis can add test later.

LenkaNovak:
Late afternoon with a test should be ok, thanks! I can work with this branch for now.

dennisYatunin:
bors try

dennisYatunin:
I added a unit test and made it so that this will (probably) work with restart files. I'll merge this in if it passes CI.

dennisYatunin:
bors try

PR #1922, opened 2023-07-27
Update manifest files

charleskawczynski:
This PR updates the dependencies, which includes https://github.com/CliMA/ClimaCore.jl/pull/1392


charleskawczynski:
Can someone (@nefrathenrici, or @valeriabarra, or @juliasloan25) review the results in here to make sure the physical results still look okay?

valeriabarra:
The BW and HS tests have no difference at all (when compared to the results on `main`). The aquaplanet test has some small differences. But I am not the best person to tell if those are significant or not. Perhaps @szy21 ?

szy21:
The aquaplanet looks fine to me.

PR #1923, opened 2023-07-27
removes unnecessary variables from get_spaces_restart

LanaLubecke:
## Purpose 

Removes unnecessary variables from get_spaces_restart so that the function works with box simulations.

## Content

----


szy21:
Hi @LanaLubecke, Could you please test if restart works for a spherical run with this branch? Both Dennis and I don't think it will break, but we don't want to risk anything for these two weeks as we have a tight deadline. Thanks!

LanaLubecke:
Yes I will test that
________________________________

szy21 (approved):

PR #1928, opened 2023-07-28
Fix some driver inference failures

charleskawczynski:
This PR fixes some inference failures in the driver.


charleskawczynski:
I think the main latency issue is compiling the init, closing.

PR #1935, opened 2023-07-31
check NaN in tendencies

szy21:
## Purpose 

There have been a few cases that the simulation has NaNs but still runs. This PR adds a check for NaNs in tendencies.

## Content

----

### DIFF
--- a/src/prognostic_equations/implicit/implicit_tendency.jl
+++ b/src/prognostic_equations/implicit/implicit_tendency.jl
@@ -31,6 +31,7 @@ function implicit_tendency!(Yₜ, Y, p, t)
             zero_velocity_tendency!(Yₜ, Y, p, t, colidx)
         end
     end
+    any(isnan, Yₜ) && error("Found NaN in implicit tendency")
     return nothing
 end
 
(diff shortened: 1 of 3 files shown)


charleskawczynski (approved):

szy21:
Hmm, this increases allocation a lot: https://buildkite.com/clima/climaatmos-ci/builds/11772#0189ae1e-372a-49c8-a7e3-0f5ba68b5f72

charleskawczynski:
> Hmm, this increases allocation a lot: https://buildkite.com/clima/climaatmos-ci/builds/11772#0189ae1e-372a-49c8-a7e3-0f5ba68b5f72

That's a bummer, `isnan(::Function, ::FieldVector)` must be type-unstable. One thing we could do is add a debugger type into the model, and put these sort of checks inside conditionals: `if debugger isa DebugAtmos`. This would be statically removed when it's not `DebugAtmos`, and we could frankly then sprinkle type-unstable debug statements virtually anywhere (maybe except inside broadcast expressions if ClimaCore can't handle them).

Or we just increase the allocation limit, I'm fine with this for the short term.

szy21:
I'm not in a hurry to merge this, as in most cases NaNs will be captured by saturation adjustment. It would still be good to stop the simulation when the tendency has NaN though, as we can examine the state when that happens. I would be happy to try a better way sometime after this week.

szy21:
Superseded by #3032

PR #1936, opened 2023-08-01
Update dycoms radiation tendency

dennisYatunin:
## Purpose 

This PR updates the DYCOMS RF01 radiation tendency, and it adds a new test case to CI that runs DYCOMS with diagnostic EDMFX.

## Content

- Add the `column_indefinite_integral!` and `column_mapreduce!` functions. These should be moved into ClimaCore as soon as possible, but they are defined here for now so that we can merge this in quickly.
- Rewrite the DYCOMS radiation tendency so that it only performs generic ClimaCore operations (i.e., no accessing `parent`, no manual loops over levels, no ODE solves, etc.), and so that it works for any moist atmosphere model (i.e., no relying on the old EDMF cache).
- Add some missing EDMF initial conditions to the list of valid initial condition strings in `type_getters.jl`.
- Add a new test case to CI that runs DYCOMS in a box with diagnostic EDMFX.

----


dennisYatunin:
bors try

szy21 (approved):

PR #1939, opened 2023-08-01
add grid-mean cloud fraction to radiation

szy21:
## Content

----


szy21:
I think this does not change MSE because there is no cloud. @dennisYatunin Could you please run this branch and main for `sphere_baroclinic_wave_rhoe_equilmoist` with `--rad allskywithclear` for 10 days, and compare if the results are different? Thanks!

szy21:
@valeriabarra Let's tag a release after this is merged and update the coupler, thanks!

PR #1944, opened 2023-08-03
Update surface conditions dispatch

etmaca5:
## Purpose 

Create a dispatch for update surface conditions which simplifies the code

----

### DIFF
--- a/src/surface_conditions/surface_conditions.jl
+++ b/src/surface_conditions/surface_conditions.jl
@@ -5,137 +5,69 @@ Updates the value of `p.sfc_conditions` based on the current state `Y` and time
 `t`. This function will only update the surface conditions if the surface_setup
 is not a PrescribedSurface.
 """
+
 function update_surface_conditions!(Y, p, t)
+    # Need to extract the field values so that we can do
+    # a DataLayout broadcast rather than a Field broadcast
+    # because we are mixing surface and interior fields
     if isnothing(p.sfc_setup)
         p.is_init[] && set_dummy_surface_conditions!(p)
         return
     end
-    # Need to extract the field values so that we can do
-    # a DataLayout broadcast rather than a Field broadcast
-    # because we are mixing surface and interior fields
     sfc_local_geometry_values = Fields.field_values(
         Fields.level(Fields.local_geometry_field(Y.f), Fields.half),
     )
     int_local_geometry_values =
         Fields.field_values(Fields.level(Fields.local_geometry_field(Y.c), 1))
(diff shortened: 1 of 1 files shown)


dennisYatunin (changes requested):
some suggested changes

dennisYatunin (approved):
Looks good now!

etmaca5:
bors cancel

src/surface_conditions/surface_conditions.jl
```diff
@@ -5,138 +5,70 @@ Updates the value of `p.sfc_conditions` based on the current state `Y` and time
...
+
+surface_state(sfc_setup_wrapper::SurfaceState, _, _, _) = sfc_setup_wrapper
+
+surface_state(
+    sfc_setup_wrapper::Function,
+    sfc_local_geometry_values,
+    int_z_values,
+    t,
+) = sfc_setup_wrapper(sfc_local_geometry_values.coordinates, int_z_values, t)
+
```
szy21, line 71: I'm a bit confused how the dispatch works here. Could you explain a bit more? (Feel free to merge, I just want to make sure I understand).

etmaca5: Of course! Because we cannot store a function in a variable this surface_state() function is used in the case where surface setup is a function (the Rico simulation for example) and it passes a changed surface_setup at the last moment as a parameter to the surface_state_to_conditions() function. 

In the other two cases: the coupler case and the default case we change/ wrap the surface setup using the sfc_setup_wrapper() function and store them using a variable wrapped_sfc_setup.

I'm not sure if that clarifies what it's doing?

szy21: I guess I'm confused about what the function `sfc_setup_wrapper(sfc_local_geometry_values.coordinates, int_z_values, t)` is?

etmaca5: I believe that was in the code before in a slightly different format, I'm not 100% sure how it works. I think it casts the default surface setup type to each coordinate, but I don't really know how. I think Dennis probably knows.

etmaca5: Oh also it might be unclear that the variable passed into the surface_setup() function has the same name as a complete different function.

szy21: Yeah that's probably why I am confused I think. I can ask Dennis tomorrow, thanks!

etmaca5: Do you want me to change the variable name in another PR and merge that quickly?

szy21: That might be good, thanks!

PR #1947, opened 2023-08-04
GPU-CPU comparison: Use relative error

### LINKED ISSUE #1946: Use relative error in GPU-CPU comparisons
We're currently using an absolute error for the GPU-CPU comparisons, but some of the fields vary greatly in order of magnitudes compared to others. 

I think we will have a clearer metric if we used a relative error.

Solution suggested: use a relative error

valeriabarra:
## Purpose 
This PR modifies the error calculation in the GPU-CPU comparison. It uses a relative error rather than an absolute one. 

Closes #1946 

## Content
- Solution implemented
     - [x] Modified error calculation in GPU-CPU comparison to use relative error 

Review checklist

I have:
- followed the codebase contribution guide: https://clima.github.io/ClimateMachine.jl/latest/Contributing/
- followed the style guide: https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/
- followed the documentation policy: https://github.com/CliMA/policies/wiki/Documentation-Policy
- checked that this PR does not duplicate an open PR.

In the Content, I have included 
- relevant unit tests, and integration tests, 
- appropriate docstrings on all functions, structs, and modules, and included relevant documentation.

----

### DIFF
--- a/post_processing/compare_outputs.jl
+++ b/post_processing/compare_outputs.jl
@@ -4,6 +4,8 @@ import ClimaAtmos as CA
 import ClimaComms
 import ClimaCore.InputOutput as InputOutput
 import ClimaCore.Fields as Fields
+using LinearAlgebra: norm
+
 function parse_commandline()
     s = ArgParse.ArgParseSettings()
     ArgParse.@add_arg_table s begin
@@ -63,12 +65,12 @@ compare(a::F, b::F, pn0 = "") where {F <: Fields.Field} =
     compare(true, a, b, pn0)
 function compare(zero_error, a::F, b::F, pn0 = "") where {F <: Fields.Field}
     isempty(propertynames(a)) || return _compare(zero_error, a, b, pn0)
-    err = abs.(Array(parent(a)) .- Array(parent(b)))
-    if !(maximum(err) == 0.0)
-        @warn "--- Failed field comparison $pn0"
-        @show a
-        @show b
-        @show maximum(err)
+    abs_err = abs.(Array(parent(a)) .- Array(parent(b)))
+    rel_err =
+        norm(Array(parent(a)) .- Array(parent(b))) ./ norm(Array(parent(a)))
+    if !(maximum(abs_err) == 0.0) || !(maximum(rel_err) == 0.0)
+        @warn "--- Failed field comparison $pn0" a b max_abs_err =
+            maximum(abs_err) rel_err
         zero_error = false
     end
     return zero_error


valeriabarra:
bors try

charleskawczynski (commented):
Can we compute and show both? There are cases where that would be helpful

valeriabarra:
> Can we compute and show both? There are cases where that would be helpful

Like which cases are you thinking of?

charleskawczynski:
> Like which cases are you thinking of?

When we have tracers with no forcing-- we'll end up seeing NaNs for the relative error.

valeriabarra:
I'm thinking that some fields vary greatly in order of magnitudes compared to others, and have a very large magnitude (e.g., total energy). To me relative errors are always more informative, but we can keep both.

post_processing/compare_outputs.jl
```diff
@@ -63,12 +63,14 @@ compare(a::F, b::F, pn0 = "") where {F <: Fields.Field} =
     compare(true, a, b, pn0)
 function compare(zero_error, a::F, b::F, pn0 = "") where {F <: Fields.Field}
     isempty(propertynames(a)) || return _compare(zero_error, a, b, pn0)
-    err = abs.(Array(parent(a)) .- Array(parent(b)))
-    if !(maximum(err) == 0.0)
+    abs_err = abs.(Array(parent(a)) .- Array(parent(b)))
+    rel_err = abs.((Array(parent(a)) .- Array(parent(b))) ./ Array(parent(a)))
```
charleskawczynski, line 67: I might suggest doing a norm in the denominator, otherwise any element that has zero will give us NaN

valeriabarra: What norm do you suggest? L-2?

charleskawczynski: I think L2 is fine, we just need some ballpark that's likely not zero-- and if it is, we'll know we have a trivial (all zeros) field.

valeriabarra: Like 
```
rel_err = norm.( Array(parent(a)) .- Array(parent(b)) ) ./ norm.( Array(parent(a)) )
```
?

charleskawczynski: Well, the maximum is computed below, and you don't want to broadcast norm (that would call norm over scalars), so perhaps:

```julia
rel_err = Array(parent(a)) .- Array(parent(b)) ./ norm( Array(parent(a)) )
```

valeriabarra: I would still keep the norm in the nominator. Otherwise we have a sort of mixed error definition (with the max below). By the way, the comparison jobs in CI are now not showing anything (even using a bitwise or). Can you think of why?

charleskawczynski: > I would still keep the norm in the nominator. Otherwise we have a sort of mixed error definition
What's the error?

CI is not showing anything because we first check `if !(maximum(abs_err) == 0.0) | !(maximum(rel_err) == 0.0)`, so this must be returning false.

charleskawczynski: > I would still keep the norm in the nominator. Otherwise we have a sort of mixed error definition (with the max below).

Computing the norm should result in a scalar variable, meaning that we wouldn't need a `maximum` call later.

valeriabarra: > CI is not showing anything because we first check `if !(maximum(abs_err) == 0.0) | !(maximum(rel_err) == 0.0)`, so this must be returning `false`?

That was not the reason why. I ran interactively in the REPL and saw the error message 
```
ERROR: UndefVarError: norm not defined
```

I am not sure why this was suppressed both in CI and when run locally (non interactively). For some reason the `stderr` was not redirected to `stdout`? I have never seen this behavior before. Took a while to figure out, because it only showed up in REPL mode. 

Now I fixed it. And it is printing the values correctly.

examples/Manifest.toml
```diff
@@ -262,7 +262,7 @@ version = "0.1.8"
 deps = ["ArgParse", "ArtifactWrappers", "Artifacts", "AtmosphericProfilesLibrary", "CLIMAParameters", "CUDA", "ClimaComms", "ClimaCore", "ClimaTimeSteppers", "CloudMicrophysics", "Colors", "Dates", "Dierckx", "DiffEqBase", "DiffEqCallbacks", "Distributions", "DocStringExtensions", "FastGaussQuadrature", "ImageFiltering", "Insolation", "Interpolations", "IntervalSets", "JLD2", "LambertW", "LinearAlgebra", "Logging", "NCDatasets", "NVTX", "OrdinaryDiffEq", "Pkg", "Printf", "RRTMGP", "Random", "RootSolvers", "StaticArrays", "Statistics", "StatsBase", "SurfaceFluxes", "TerminalLoggers", "Test", "Thermodynamics", "YAML"]
 path = ".."
 uuid = "b2c96348-7fb7-4fe0-8da9-78d88439e717"
-version = "0.15.4"
+version = "0.15.6"
```
valeriabarra, line 265: The modified Manifest.toml file accidentally slipped and should have not been part of this PR. I am going to fix this.

valeriabarra:
@charleskawczynski , this is passing and printing the expected values (I fixed the `UndefVarError` problem that was not showing up in CI stdout before). 
When approved, I will merge it. Thanks

post_processing/compare_outputs.jl
```diff
@@ -63,12 +65,15 @@ compare(a::F, b::F, pn0 = "") where {F <: Fields.Field} =
...
+    abs_err = abs.(Array(parent(a)) .- Array(parent(b)))
+    rel_err =
+        norm(Array(parent(a)) .- Array(parent(b))) ./ norm(Array(parent(a)))
+    if !(maximum(abs_err) == 0.0) || !(maximum(rel_err) == 0.0)
         @warn "--- Failed field comparison $pn0"
         @show a
         @show b
-        @show maximum(err)
+        @show maximum(abs_err)
+        @show rel_err
```
simonbyrne, line 76: rather than using `@show`, it would be better to  make these arguments to `@warn`

post_processing/compare_outputs.jl
```diff
@@ -63,12 +65,15 @@ compare(a::F, b::F, pn0 = "") where {F <: Fields.Field} =
     compare(true, a, b, pn0)
 function compare(zero_error, a::F, b::F, pn0 = "") where {F <: Fields.Field}
     isempty(propertynames(a)) || return _compare(zero_error, a, b, pn0)
-    err = abs.(Array(parent(a)) .- Array(parent(b)))
-    if !(maximum(err) == 0.0)
+    abs_err = abs.(Array(parent(a)) .- Array(parent(b)))
+    rel_err =
+        norm(Array(parent(a)) .- Array(parent(b))) ./ norm(Array(parent(a)))
```
simonbyrne, line 70: any reason we're doing this on the arrays rather than the fields? @charleskawczynski ?

post_processing/compare_outputs.jl
valeriabarra: Do you suggest to use multiple `@warn` lines? Like: 
```suggestion
        @warn "First field: $a"
        @warn "Second field: $b"
        @warn "Max absolute error $maximum(abs_err)"
        @warn "Relative error $rel_err"
```
? 
That would be a bunch of warnings

simonbyrne: no, you can pass args to `@warn` (like all logging macros). They can be either `key=arg`, or simply `arg` (where the key will be the name of the arg)
```
     @warn "Failed field comparison $pn0" a b max_abs_err=maximum(abs_err) rel_arr
```

simonbyrne: See https://docs.julialang.org/en/v1/stdlib/Logging/#Logging.@logmsg

simonbyrne (approved):

PR #1952, opened 2023-08-07
Temporarily disable profiling in scaling pipeline

sriharshakandala:
Temporarily disable profiling in scaling pipeline
Increase job times

----

### DIFF
--- a/.buildkite/scaling/pipeline.sh
+++ b/.buildkite/scaling/pipeline.sh
@@ -7,7 +7,7 @@ high_res_process_counts=(1 2 4 8 16 32 64 128)
 FT="Float32"
 resolutions=("low" "mid" "high")
 max_procs_per_node=16 # limit this artificially for profiling
-profiling=enable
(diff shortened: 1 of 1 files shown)


simonbyrne:
Does the profiling add overhead?

sriharshakandala:
The most expensive simulations are the `1 or 2` process ones. If we are ok with changing the scaling reference to `2` or `4` process run for high resolution simulations, that would help speedup the pipeline! (assuming perfect scaling upto `2` or `4` processes!)

simonbyrne:
The other option would be to reduce how long we run the simulation (for the smaller numbers of procs).

sriharshakandala:
> The other option would be to reduce how long we run the simulation (for the smaller numbers of procs).

We were targeting about `2000` samples for reducing variablility during the averaging process last year. But we can retry a smaller number!

simonbyrne:
That's fair.

PR #1953, opened 2023-08-07
limit mixing length by smag-lilly scale

szy21:
## Content

----


szy21:
@tapios Is this what you suggested? Would dz be too large for limiting the mixing length in the upper atmosphere?

tapios:
> @tapios Is this what you suggested? Would dz be too large for limiting the mixing length in the upper atmosphere?

The Smagorinsky-Lilly length scale has the properties we want (dz limited by stratification). I suspect (but need more time to think this through) the right thing to do is limit the mixing length by that length scale, rather than having that in the ensemble over which the smooth minimum is taken.

PR #1956, opened 2023-08-08
add Prandtl number calculation

### LINKED ISSUE #1955: Add prandtl number calculation

szy21:
## Purpose 

Closes #1955 

## Content

----


trontrytel (approved):
LGTM! 

I wonder if `gradient_richardson_number` is ever needed other than for computing `turbulent_prandtl_number`? If so maybe we don't have to define a separate function for it.

szy21:
> LGTM!
> 
> I wonder if `gradient_richardson_number` is ever needed other than for computing `turbulent_prandtl_number`? If so maybe we don't have to define a separate function for it.

It doesn't seem we use it anywhere else. I will just move it into the prandtl number calculation then.

PR #1959, opened 2023-08-08
Set qt to zero for dry EDMF cases and add GABLS

### LINKED ISSUE #1960: Add GABLS for diagnostic EDMF
Add GABLS, make diagnostic EDMF work for dry case or set qt to zero.

szy21:
## Purpose 

Add GABLS and set qt to zero in dry edmf cases.
Closes #1960 

## Content

----

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -439,6 +439,19 @@ steps:
         artifact_paths: "diagnostic_edmfx_test_box/*"
         agents:
           slurm_mem: 20GB
+      
+      - label: ":genie: Diagnostic EDMFX GABLS in a box"
+        command: >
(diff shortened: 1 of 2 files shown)


szy21:
This seems like an easy way to make EDMF run dry case with qt=0. @trontrytel Let me know what you think.

trontrytel:
looks good to me. But it also looks like Bomex box breaks from `--nh_poly 1`

szy21:
> looks good to me. But it also looks like Bomex box breaks from `--nh_poly 1`

Changing the domain size (which has a similar effect to changing hyperviscosity) make Bomex work but break Rico:( I am a bit unsure about using hyperdiffusion for poly order 1, and also it doesn't seem to help reduce ci time much, so I will just keep the default poly order 3 for now.

PR #1963, opened 2023-08-09
Add TRMM for diagnostic EDMF

szy21:
## Purpose 

Adds TRMM with diagnostic EDMF. Also makes struct MoninObukhov always contain roughness.

## Content

----

### DIFF
--- a/src/surface_conditions/surface_setups.jl
+++ b/src/surface_conditions/surface_setups.jl
@@ -101,9 +101,10 @@ function (::Bomex)(params)
     q_vap = FT(0.02245)
     θ_flux = FT(8e-3)
     q_flux = FT(5.2e-5)
+    z0 = FT(1e-4)
     ustar = FT(0.28)
     fluxes = θAndQFluxes(; θ_flux, q_flux)
-    parameterization = MoninObukhov(; fluxes, ustar)
+    parameterization = MoninObukhov(; z0, fluxes, ustar)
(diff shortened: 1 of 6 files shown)


src/surface_conditions/surface_setups.jl
```diff
@@ -101,9 +101,10 @@ function (::Bomex)(params)
     q_vap = FT(0.02245)
     θ_flux = FT(8e-3)
     q_flux = FT(5.2e-5)
+    z0 = FT(1e-4)
```
trontrytel, line 104: why not zero?

szy21: You can't use zero as it is used in the denominator in surface flux. 1e-4 for roughness is a typical value over ocean.

trontrytel: Or maybe it's a different variable than what I think it is/ Maybe a longer name?

trontrytel (approved):

src/surface_conditions/surface_setups.jl
szy21: Right, it is roughness, and it is commonly noted as z0 in Monin-Obukhov.

szy21: I can add a comment on the MoninObukhov struct to make this more clear, if that's helpful?

PR #1966, opened 2023-08-10
Pipeline for GPU target simulations

sriharshakandala:
Pipeline for GPU target simulations

----

### DIFF
--- a/config/gpu_configs/target_gpu_implicit_baroclinic_wave.yml
+++ b/config/gpu_configs/target_gpu_implicit_baroclinic_wave.yml
@@ -1,0 +1,7 @@
+h_elem: 30
+initial_condition: "DryBaroclinicWave"
+t_end: "1days"
+z_elem: 45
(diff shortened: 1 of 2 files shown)


bloops (approved):

config/gpu_configs/target_gpu_implicit_baroclinic_wave.yml
```diff
@@ -0,0 +1,7 @@
+h_elem: 30
+initial_condition: "DryBaroclinicWave"
+t_end: "1days"
+z_elem: 45
+dt: "50secs"
```
bloops, line 5: do we know for sure that dt=50secs would not lead the simulation to blow up? (Perhaps write the corresponding CFL condition numbers in a comment?)

sriharshakandala: We should verify this using a long run!

PR #1969, opened 2023-08-11
Set apply_limiter default to false

szy21:
## Purpose 

Set `apply_limiter` default to false as that is most of the case. Also fixes some missing `apply_limiter` argument when converting cli to yaml. cc @valeriabarra @nefrathenrici 

## Content

----


config/default_configs/default_perf.yml
```diff
@@ -1,4 +1,5 @@
 moist: "equil"
+apply_limiter: true
```
valeriabarra, line 2: Why do we want to use limiters in the perf jobs?

config/perf_configs/flame/perf_gw.yml
```diff
@@ -15,3 +15,4 @@ precip_model: "0M"
 alpha_rayleigh_uh: 0.0
 job_id: "flame_perf_gw"
 moist: "equil"
+apply_limiter: true
```
valeriabarra, line 18: same as above

config/default_configs/default_perf.yml
szy21: Oh, I just want to keep it the same as before (We probably want to check the performance of the limiter?). But we don't have to.

config/perf_configs/flame/perf_gw.yml
szy21: This one can be removed I think.

config/default_configs/default_perf.yml
szy21: @nefrathenrici @charleskawczynski Do we want this to be true or false in perf?

nefrathenrici: @charleskawczynski will know for sure, but I assume we should keep it the same for consistency.

config/default_configs/default_config.yml
```diff
@@ -182,7 +182,7 @@ trunc_stack_traces:
   value: true
 apply_limiter:
   help: "Apply a horizontal limiter to every tracer [`true` (default), `false`]"
```
nefrathenrici, line 184:
```suggestion
  help: "Apply a horizontal limiter to every tracer [`false` (default), `true`]"
```

nefrathenrici (commented):
Thanks for taking care of this!

config/default_configs/default_perf.yml
valeriabarra: Charlie is out now and he will be all next week. So I guess for now we want to keep it the same, and then when we have some clarity on why it was on for the perf jobs, we can always change it, if needed.

szy21: Sounds good to me, thanks!

valeriabarra (approved):

PR #1981, opened 2023-08-16
Renaming surface condition variables

etmaca5:
## Purpose 

Changing the name of a variable so that it doesn't share the same name as another function

----


szy21 (approved):
This is definitely more clear I think, thanks!

PR #1987, opened 2023-08-16
Setting maximum iteration for diagnostic edmf to 5

szy21:
## Content

----


szy21:
Actually Bomex results change a bit too much I think. I'm trying tolerance 0.0003 (~0.1 K).

szy21:
The changes in the results are smaller with maximum iteration 8 and tolerance 0.0003 (~0.1K). I think it's good for now. I will come back to the oscillations in Bomex after #1994 and implicit solver.

trontrytel (approved):

PR #2000, opened 2023-08-21
Add nonhydrostatic pressure drag for diagnostic edmf

### LINKED ISSUE #1800: Add nonhydrostatic pressure drag for diagnostic edmfx
We can use the existing ᶠupdraft_nh_pressure function, but need to convert everything to the contravariant 3 component.
- [x] Write out the equation and double check with Tapio
- [ ] Implement

szy21:
## Purpose 

Please see the content for the changes. The oscillation in Bomex looks worse, but I don't think it is related to this change. I will revisit it after changing the detrainment to deal with updraft top. The updraft height in TRMM decreases as expected. I need to reduce the dt for the aquaplanet run to make it stable, but the dt for that simulation was probably too large. I think we can come back to the aquaplanet after implicit diffusion and dissipation. Other cases don't change much.

Closes #1800 

## To-do

- Fix allocation

## Content

- Fixes a bug in pressure drag closure
- Change the denominator in buoyancy from \rho to \rho^j
- Update docs

----


src/prognostic_equations/edmfx_closures.jl
```diff
@@ -29,23 +29,23 @@ end
    - nh_presssure_flag - bool flag for if we want/don't want to compute
                          pressure drag
    - ᶠlg - local geometry (needed to compute the norm inside a local function)
-   - ᶠbuoyʲ - updraft buoyancy
-   - ᶠu₃ʲ, ᶠu₃⁰ - Covariant3Vector velocity for updraft and environment
+   - ᶠbuoyʲ - covariant3 or contravariant3 updraft buoyancy
+   - ᶠu3ʲ, ᶠu3⁰ - covariant3 or contravariant3 velocity for updraft and environment
```
szy21, line 33: I use u3 to note that it could be either covariant3 or contravariant3. Let me know if you have naming suggestion :)

src/cache/diagnostic_edmf_precomputed_quantities.jl
```diff
@@ -310,6 +314,19 @@ function set_diagnostic_edmf_precomputed_quantities!(Y, p, t)
...
             )
 
+            @. nh_pressureʲ_prev_level = ᶠupdraft_nh_pressure(
+                params,
+                p.atmos.edmfx_nh_pressure,
+                local_geometry_prev_halflevel,
+                -∇Φ³_prev_level * (ρʲ_prev_level - ρ_prev_level) / ρʲ_prev_level,
+                u³ʲ_prev_halflevel,
+                u³⁰_prev_halflevel,
+                FT(8000),
```
szy21, line 324: I will add a todo that it is better to use updraft top here.

trontrytel (approved):

src/cache/diagnostic_edmf_precomputed_quantities.jl
tapios: Can you at least make it a parameter for now, rather than a fixed number? And setting it to the scale height (RT/g with a reference T) would make sense.

src/cache/diagnostic_edmf_precomputed_quantities.jl
```diff
@@ -365,6 +382,18 @@ function set_diagnostic_edmf_precomputed_quantities!(Y, p, t)
...
+                (
+                    1 /
+                    (local_geometry_halflevel.J * local_geometry_halflevel.J)
+                ) * (
+                    local_geometry_prev_level.J *
+                    local_geometry_prev_level.J *
+                    2 *
+                    nh_pressureʲ_data_prev_level
+                )
+
```
tapios, line 396: What is the scale_factor in the next line?

src/prognostic_equations/edmfx_closures.jl
tapios: It would be good to make clear in the docstring which is used when.

tapios (approved):

src/cache/diagnostic_edmf_precomputed_quantities.jl
szy21: That's not a good name. It's the threshold for setting everything to grid mean.

src/cache/diagnostic_edmf_precomputed_quantities.jl
szy21: I can set it to R_d * T_surf_ref / g then. All these three constants are in CLIMAParameters. Does that sound good to you?

tapios: Yes

src/cache/diagnostic_edmf_precomputed_quantities.jl
szy21: I changed it to `minimum_value`.

PR #2006, opened 2023-08-21
Use 3D curl in vector hyperdiffusion

simonbyrne:
## Purpose 
This switches to using the 3D spectral element curl (https://github.com/CliMA/ClimaCore.jl/pull/1425) in hyperdiffusion.

## To-do
- [ ] check that it is non-breaking

## Content

----


simonbyrne:
https://buildkite.com/clima/climaatmos-longruns/builds/1161

szy21 (approved):
Is the behavior change as expected?

simonbyrne:
There's probably some small numerical change, but it shouldn't be behavior changing.

szy21:
> There's probably some small numerical change, but it shouldn't be behavior changing.

Right, I meant the mse change. The results look good to me.

simonbyrne:
Longruns available here: https://buildkite.com/clima/climaatmos-longruns/builds/1166

@szy21 do they look okay?

szy21:
The longrun looks good to me.

PR #2007, opened 2023-08-22
Config files for multi-year, high-topped, dry Held-Suarez simulation

### LINKED ISSUE #1990: Multi-year dry Held-Suarez simulation
**Requirements** 

- 55 km domain top
- 200km horizontal resolution
- 63 levels
- Held-Suarez forcing

- [ ] Merge completed simulation configuration into long-run buildkite pipeline.

akshaysridhar:
new file:   longrun_dry-hs_rhoe_hightop.yml

Linked Issue #1990

### DIFF
--- a/config/longrun_configs/longrun_hs_rhoe_dry_nz63_55km_rs35km.yml
+++ b/config/longrun_configs/longrun_hs_rhoe_dry_nz63_55km_rs35km.yml
@@ -1,0 +1,13 @@
+dz_bottom: 30.0
+dz_top: 3000.0
+dt_save_to_disk: "10days"
+t_end: "300days"
+h_elem: 16
+z_elem: 63
+z_max: 55000.0
+kappa_4: 1.0e16
+forcing: "held_suarez"
+dt: "300secs"
+rayleigh_sponge: true
+zd_rayleigh: 35000.0
+job_id: "longrun_hs_rhoe_equilmoist_nz63_0M_55km_rs35km"
--- a/.buildkite/longruns/pipeline.yml
+++ b/.buildkite/longruns/pipeline.yml
@@ -274,6 +274,15 @@ steps:
 
   - group: "Milestone 1.5.2a"
     steps:
+      
+      - label: ":computer: held-suarez, dry, high-topped (55km), high-sponge(35km), helem_16 np_3"
+          - srun julia --project=examples examples/hybrid/driver.jl --config_file $CONFIG_PATH/longrun_hs_rhoe_dry_nz63_55km_rs35km.yml
+          - julia --color=yes --project=examples post_processing/remap/remap_pipeline.jl --data_dir longrun_hs_rhoe_dry_nz63_55km_rs35km --out_dir longrun_hs_rhoe_dry_nz63_55km_rs35km
+        artifact_paths: "longrun_hs_rhoe_dry_nz63_55km_rs35km.yml/*"
+        env:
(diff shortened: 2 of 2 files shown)


szy21:
You don't need any sponge to make it stable? That's good!

akshaysridhar:
> You don't need any sponge to make it stable? That's good!

I'm testing a series of these configurations and will summarise the best outcomes over 1000day runs.

akshaysridhar:
> You don't need any sponge to make it stable? That's good!

In this specific config file instance the `default` arg of `alpha_rayleigh_w` should still be applied - if it isn't this would be a separate issue.

szy21:
> > You don't need any sponge to make it stable? That's good!
> 
> In this specific config file instance the `default` arg of `alpha_rayleigh_w` should still be applied - if it isn't this would be a separate issue.

It won't be applied if rayleigh_sponge is false.

szy21:
But the dry simulation wouldn't stop with NaNs, so it might be worth checking the results.

akshaysridhar:
> > > You don't need any sponge to make it stable? That's good!
> > 
> > 
> > In this specific config file instance the `default` arg of `alpha_rayleigh_w` should still be applied - if it isn't this would be a separate issue.
> 
> It won't be applied if rayleigh_sponge is false.

Ah, yes - noted.

akshaysridhar:
> But the dry simulation wouldn't stop with NaNs, so it might be worth checking the results.

Yeah output frequency is `10days` and I can query the hdf5 for `NaNs` offline.

akshaysridhar:
@szy21 sponge arg is fixed now.

szy21:
Thanks! You need to change `zd_rayleigh` as well though, the default is 15000 and is too low.

szy21:
> Ah, yes - noted. Is the default sponge arg also something we want to change ?

I think we can set it to false for now, as there are still a lot of 30km simulations that don't require a sponge. Once we move to default 55km we can change the sponge, if needed.

akshaysridhar:
> Thanks! You need to change `zd_rayleigh` as well though, the default is 15000 and is too low.

Yeah referring to the moist sims for this one - in those the value for this arg is 35km.

szy21:
It might be good to make the job_id of these two experiments more consistent, for clarity.

akshaysridhar:
At day 270 
```
julia> Y.c.ρ |> extrema
(0.000228748f0, 1.3460394f0)
```

config/longrun_configs/longrun_hs_rhoe_dry_nz63_55km_rs35km.yml
```diff
@@ -0,0 +1,13 @@
...
+t_end: "300days"
+h_elem: 16
+z_elem: 63
+z_max: 55000.0
+kappa_4: 1.0e16
+forcing: "held_suarez"
+dt: "300secs"
+rayleigh_sponge: true
+zd_rayleigh: 35000.0
+job_id: "longrun_dry-hs_rhoe_highres_hightop"
```
szy21, line 13:
```suggestion
job_id: "longrun_hs_rhoe_equilmoist_nz63_0M_55km_rs35km"
```
(otherwise the artifact folder will have a problem I think)

szy21: I will make the change and merge.

PR #2010, opened 2023-08-22
Fix config name in longruns

simonbyrne:
## Purpose 
This was using the wrong name

## Content

----


nefrathenrici (approved):
Thanks!

PR #2011, opened 2023-08-22
Add 0-moment micro to diagnostic EDMFX

trontrytel:
This PR couples 0-moment microphysics scheme to diagnostic EDMFX


src/cache/diagnostic_edmf_precomputed_quantities.jl
```diff
@@ -310,6 +326,23 @@ function set_diagnostic_edmf_precomputed_quantities!(Y, p, t)
...
 
+            # TODO:
+            # - we should also sum sgs contributions to energy and apply them
+            #   the same way as qt
+            # - code up the energy tendency and re-use it in here, depracated and precipitation
+
+            # Updraft q_tot sources from precipitation formation
+            # To be applied in updraft continuity, moisture and energy
+            # for updrafts and grid mean
+            @. S_q_totʲ_prev_level = -min(
```
szy21, line 337: I think we guarantee that remove_precipitation is smaller than q_tot / dt by setting _τ_precip to dt, but maybe no harm to do it again here.

src/parameterized_tendencies/microphysics/precipitation.jl
```diff
@@ -63,6 +63,21 @@ function compute_precipitation_cache!(
...
+    colidx,
+    ::Microphysics0Moment,
+    ::DiagnosticEDMFX,
+)
+    (; ᶜS_ρq_tot, ᶜS_q_tot⁰, ᶜS_q_totʲs) = p
+    FT = Spaces.undertype(axes(Y.c))
+
+    @. ᶜS_ρq_tot[colidx] =
+        Y.c.ρ[colidx] *
+        (sum(p.ᶜS_q_totʲs[colidx]) + p.ᶜS_q_tot⁰[colidx])
```
szy21, line 78: It should be something like ᶜS_q_totʲs * ρaʲs in the sum?

szy21 (commented):
Looking good. I was hoping we can have something like `calculate_qt_tendency` and dispatch on the microphysics/precipitation scheme. And for now we can have nothing or 0M, so we can still support e.g. Bomex where we don't have precipitation. And we can add 1M in the future.

szy21:
What's the advantage of caching `ᶜS_e_totʲs_helper` over caching `ᶜS_e_totʲs`?

trontrytel:
> What's the advantage of caching `ᶜS_e_totʲs_helper` over caching `ᶜS_e_totʲs`?

Initially I thought I would be able to hide the different options via dispatch over `compute_precipitation_cache!` and would not have to modify the `precipitation_tendency!` itself. So I was trying to keep to a similar pattern of `S_q_tot * rhoa * S_e_tot`. But I gave up on that, and now I don't modify the cache and instead have an ugly if/else in the actual tendency. 

So from that point of view I can instead cache `ᶜS_q_totʲs * ᶜS_e_totʲs_helper` . I can make the change if you think it's cleaner this way.

trontrytel:
All the cases work now. From the precipitating ones:
- TRMM looks very patchy (what a great analysis on my side...)
- Rico looks non-existent

I'll tackle allocations tomorrow, though right now I don't see any obvious culprits

src/cache/diagnostic_edmf_precomputed_quantities.jl
```diff
@@ -489,6 +525,16 @@ function set_diagnostic_edmf_precomputed_quantities!(Y, p, t)
...
+        # Environment precipitation sources (to be applied to grid mean)
+        @. S_q_tot⁰_prev_level = q_tot_precipitation_sources(
+            precip_model,
+            thermo_params,
+            microphys_params,
+            dt,
+            q_tot_prev_level,
+            ts_prev_level,
+        )
+
```
szy21, line 537: Since we don't need this in the integral, we can actually move it outside the loop and calculate the environment precipitation of the entire 3d field.

trontrytel: Sounds good. Will update tomorrow

szy21 (approved):

trontrytel:
@charleskawczynski - I applied your suggestions regarding the way I was summing things in precipitation tendency and with the FT. However it's still allocating. Is that level of allcocs acceptable? Or would you have any more hints for me?

`  Evaluated: 650640 ≤ 637328`

trontrytel:
FYI: Just adding 3 variables to cache 

```
            ᶜS_q_totʲs = similar(Y.c, NTuple{n, FT}),
            ᶜS_q_tot⁰ = similar(Y.c, FT),
            ᶜS_e_totʲs_helper = similar(Y.c, NTuple{n, FT}),
```
            
and then unpacking them:

```
    (; ᶜS_q_totʲs, ᶜS_q_tot⁰, ᶜS_e_totʲs_helper) = p
```

Increases allocations by

`Evaluated: 657168 ≤ 637328`

See this PR: https://github.com/CliMA/ClimaAtmos.jl/pull/2031

I'll test now just adding them, without unpacking.

charleskawczynski:
Interesting, I wonder if we're running into https://github.com/CliMA/ClimaCore.jl/issues/1015?

charleskawczynski:
I think it might be better / more ideal if we could aim to compute more things on the fly, from a minimal number of variables, if possible. Otherwise, it's probably fine to just increase the allocation limit so that you're not held back by this.

trontrytel:
> I think it might be better / more ideal if we could aim to compute more things on the fly, from a minimal number of variables, if possible. Otherwise, it's probably fine to just increase the allocation limit so that you're not held back by this.

Yes, I agree. Unfortunately we have to cache those ones. Otherwise we would have to redo the whole EDMF diagnostic loop to get the SGS contributions.

I'll test if it has something to do with unpacking rather than caching. But I doubt it

PR #2012, opened 2023-08-23
Avoid unnecessary cache recomputation

charleskawczynski:
This PR defines `CacheState` and uses it to avoid unnecessary cache re-computations by setting a Bool at the beginning of each function that ClimaTimesteppers calls, and then resetting it at the end of `set_precomputed_quantities!`.

This PR should eliminate at least one call to `set_precomputed_quantities!` (the one after `Wfact!`), but in theory, this can eliminate more.


charleskawczynski:
Superseded by timestepping efforts

PR #2017, opened 2023-08-25
Remove case-specific overrides from parameter creation

nefrathenrici:
### Content
- Removes case-specific parameter overrides. This changes simulation results, which have been checked by @szy21 (including longruns). Relevant builds: [default pipeline](https://buildkite.com/clima/climaatmos-ci/builds/12757), [longruns](https://buildkite.com/clima/climaatmos-longruns/builds/1170)
- The exception to the above is `τ_precip`, which is still set to `dt` in all cases. This is the same behavior as before, and but can now be switched off by setting the new config flag `override_τ_precip = false`
- Adds a function `create_parameter_struct` which takes in a parameter struct type, any nested parameter structs, and constructs the parameters. This cleans up a lot of boilerplate from `create_parameter_set`
- Removes unused parameters: `ug`, `vg`, `f`, and `Cd`. It was hard to grep for usage of `f` but all of my local tests work.
- Sets `f_plane_coriolis_frequency`via CLIMAParameters instead of a hardcoded value. The default value is unchanged


Sbozzolo:
Where/how is `τ_precip` used? I don't see any reference to it in `ClimaAtmos` except in setting the parameter

nefrathenrici:
> Where/how is `τ_precip` used? I don't see any reference to it in `ClimaAtmos` except in setting the parameter

@Sbozzolo It's used in CloudMicrophysics and is used in periodically removing precipitation in the 0Moment microphysics scheme. It's kind of an exception since it is always set to dt. There is likely a better way of passing it through to CloudMicrophysics but I just want to get rid of the case-specific overrides for now!

Sbozzolo:
> > Where/how is `τ_precip` used? I don't see any reference to it in `ClimaAtmos` except in setting the parameter
> 
> @Sbozzolo It's used in CloudMicrophysics and is used in periodically removing precipitation in the 0Moment microphysics scheme. It's kind of an exception since it is always set to dt. There is likely a better way of passing it through to CloudMicrophysics but I just want to get rid of the case-specific overrides for now!

Thanks! How is it passed to CloudMicrophysics? If it is always set to `dt`, maybe we can implement it as a callback that executes at the end of each integration step? 

(My interest in this stems from the fact that I am studying how to remove explicit references to `parsed_args` configuration stage.)

szy21:
It would be better to just ask the user to specify both `τ_precip` and `dt`. They don't have to be the same.

nefrathenrici:
@Sbozzolo I'm not sure how it would work as a callback, since it needs to be passed in as a parameter to CloudMicrophysics, but I'm open to changing how it's set.

src/parameters/create_parameters.jl
```diff
@@ -4,54 +4,48 @@ import CLIMAParameters as CP
...
-)
-    intersect_keys = intersect(keys(defaults), keys(overrides))
-    intersect_vals = getproperty.(Ref(overrides), intersect_keys)
-    intersect_overrides = (; zip(intersect_keys, intersect_vals)...)
-    return merge(defaults, intersect_overrides)
-end
+function create_parameter_set(config::AtmosConfig)
+    # Helper function that creates a parameter struct. If a struct has nested 
+    # parameter structs, they must be passed to subparam_structs as a NamedTuple.
+    function create_parameter_struct(param_struct; subparam_structs = (;))
```
charleskawczynski, line 17: Does this need to be defined in `create_parameter_set`?

docs/src/config.md
```diff
@@ -21,7 +21,7 @@ debugging_tc: true
 surface_setup: Bomex
 turbconv_case: Bomex
 t_end: 6hours
-turbconv: "edmf"
+turbconv: edmf
```
charleskawczynski, line 24: This was still working, correct?

src/parameters/create_parameters.jl
nefrathenrici: It doesn't need to be, but it makes the syntax nicer since it automatically grabs the correct `toml_dict` and `parsed_args` without them needing to be passed in.

docs/src/config.md
nefrathenrici: yes, both `"edmf"` and `edmf` get parsed to the String `edmf`, so the quotes aren't needed.

PR #2026, opened 2023-08-28
Remove Manifest.toml in the examples folder

Sbozzolo:
The Manifest.toml is automatically generated and machine-depended. Using the provided Manifest.toml may lead to warnings and errors. We are already not including the Manifests for the other environments.


Sbozzolo:
Thanks. I think my issues with moving to Julia 1.9 were due to the fact that I was using this Manifest. Should I assume that this Manifest is not intended for direct use?

charleskawczynski:
> I think my issues with moving to Julia 1.9 were due to the fact that I was using this Manifest.

We can update the manifest using the package manager

> Should I assume that this Manifest is not intended for direct use?
Yes, it can be used for direct use, we may just need to update the versions.

Sbozzolo:
Thank you for the explanation! 

I closed the PR, but I think that there is an opportunity to improve our documentation/messaging. I interpret the word "examples" as something that I can just take and run, without the restriction of having to use a specific version of Julia.

PR #2034, opened 2023-08-29
make scaling pipeline plot on failure

nefrathenrici:
The plotting jobs in the scaling pipeline don't run if a prior job fails, this needs to be fixed.

Test scaling build: https://buildkite.com/clima/climaatmos-scaling/builds/506#018a427c-39c0-45f1-a03c-0a44b6a7c722

### DIFF
--- a/.buildkite/scaling/pipeline.sh
+++ b/.buildkite/scaling/pipeline.sh
@@ -198,7 +198,8 @@ fi
 done
 
 cat << EOM
-    - wait
+    - wait: ~
+      continue_on_failure: true
 
     - label: ":chart_with_upwards_trend:"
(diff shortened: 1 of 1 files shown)


simonbyrne (approved):

simonbyrne:
Does the script still work in this case?

nefrathenrici:
> Does the script still work in this case?

@simonbyrne I think the plotting script will only break if there are no files to plot from.  We can see if it works with just one file when this [job](https://buildkite.com/clima/climaatmos-scaling/builds/506#018a427c-3976-4a2f-a20e-43e41f6bf96c) finishes.

PR #2044, opened 2023-08-31
Update to new implicit solver interface

dennisYatunin:
## Purpose 

Updates ClimaAtmos to use the new implicit solver interface provided in `ClimaCore.MatrixFields` (CliMA/ClimaCore.jl#1326, CliMA/ClimaCore.jl#1399, and CliMA/ClimaCore.jl#1436).

## Content

- Replace `wfact.jl` and `linsolve.jl` with `implicit_solver.jl`. Most of the code from `wfact.jl` has been rewritten using the new `FieldName` and `FieldMatrix` infrastructure, while most of the code from `linsolve.jl` has been moved into the definition of `FieldMatrixSolver` in `ClimaCore`.
- Modify the derivatives with respect to the vertical velocity `u₃` to use `AxisTensor`s and `AdjointAxisTensor`s. (We were previously replacing these derivatives with similar derivatives with respect to `u₃_data`, which made the code unnecessarily complicated.)
- Add a detailed docstring for the `ImplicitEquationJacobian` object. (This was formerly called `SchurComplementW`, but it needed to be renamed because we will not always use the "Schur complement solve" algorithm in the future.) Simplify and update the comments in the `wfact!` function to reflect the new code.
- Bump the ref counter to account for 12 small MSE table changes.
- Update the code in `test/implicit/debugging_tools.jl`. Although this still isn't a unit test, as the code is not currently modular enough to write a proper test, it can be very helpful when debugging the matrix blocks in the `ImplicitEquationJacobian`.

----


dennisYatunin:
bors try

szy21 (approved):

dennisYatunin:
bors cancel

PR #2053, opened 2023-09-03
trying to get rid of MSLP overwrite

trontrytel:


trontrytel:
cc: @szy21 - could you take a look when you have the time? I basically changed `MSLP` to `p_ref_theta` any time I saw it next to potential temperature. But I wasn't sure about the Held-Suarez forcing and some of the initial conditions.

I think the changed cases in the CI look fine. One timed-out on me. I'll retry.

cc: @nefrathenrici - looks promising!

src/initial_conditions/initial_conditions.jl
szy21, line 1: I think it might be more consistent to change MSLP to p_ref_theta in the initial conditions in this file as it is used in `p = p_0 * π_exn^(cp_d / R_d)`, so `π_exn = (p / p_0)^(R_d/cp_d)` when `p_0 = p_ref_theta`. It probably doesn't matter much though, as it is just the initial condition.

szy21 (approved):
The changes look good to me, thanks!

src/initial_conditions/initial_conditions.jl
trontrytel: I kept the MSLP in the isothermal one and in the baroclinic wave initial conditions, but changed the rest. Lets see what the CI thinks about that

charleskawczynski:
Amazing work, everyone! 👏👏👏 

I tried doing this years ago, and was unsuccessful, it’s great to see that this is finally fixed!

Issue #2043, opened 2023-08-31
OKR O.3.2.2: A flexible diagnostic module for ClimaAtmos


Sbozzolo:
# The Climate Modeling Alliance
### Purpose

This SDI proposes to add a flexible module to compute arbitrary diagnostics from the simulation.

We want to be able to:
- compute arbitrary diagnostics with possibly arbitrary reductions in time (e.g., arithmetic average, min, max) on arbitrary intervals (e.g., daily, monthly),
- let users specify what they want/where/when to output.

Our goals are:
- it has be easy for a users to specify what they want to output and for developers to add new diagnostics,
- diagnostics that are not used should not be computed,
- there should be at least one way for users/developers to achieve complex behaviors that are not pre-defined in `ClimaAtmos` (e.g., adding a new reduction operation),
- our abstractions should not significantly degrade performance.

### Cost/Benefits/Risks

Diagnostics are currently hard-coded, so this is an important step towards a general and usable `ClimaAtmos`.

A proof-of-concept implementation is already available, and some of the challenges have been addressed.

A possible performance problem with the design outlined below is that diagnostics cannot trivially use information computed in other diagnostics.

### People and Personnel
- Lead: @sbozzolo
- Collaborators: @szy21 @nefrathenrici
- Reviewers:

This design was discussed with @simonbyrne

### Components

Diagnostics are implemented as callbacks in the integrator. At fixed intervals of (integration) time, the diagnostics are computed from the state are output to disk. In an nutshell, this SDI discusses a module to produce a list of callbacks.

We will initially focus on point-wise operations and HDF5 files. Online remapping and producing NetCDF files can be implemented as a different `output_writer` (see below) and will be tackled after the main infrastructure is put in place.

#### The low-level details

(Snippets of code below are to be considered pseudo-code.)

##### `DiagnosticVariable`

We represent a diagnostic variable as `struct`s that look like (roughly following `ClimateMachine`)

```julia
struct DiagnosticVariable
    short_name::String
    long_name::String
    units::String
    description::String
    compute_from_integrator::Function
end
```

[...]

PR #2064, opened 2023-09-06
Add diagnostic module

Sbozzolo:
This PR adds a new diagnostic module that roughly follows what described in #2043.


src/diagnostics/reduction_identities.jl
```diff
@@ -0,0 +1,19 @@
...
+# reduction. This, by definition, is the identity of the operation. The identity of the operation
+# + is 0 because x + 0 = x for every x.
+#
+# We have to know the identity for every operation we want to support. Of course, users are
+# welcome to define their own by adding new methods to identity_of_reduction.
+
+identity_of_reduction(::Val{max}) = -Inf
+identity_of_reduction(::Val{min}) = +Inf
+identity_of_reduction(::Val{+}) = 0
+identity_of_reduction(::Val{*}) = 1
```
simonbyrne, line 19:
```suggestion
identity_of_reduction(::typeof(max)) = -Inf
identity_of_reduction(::typeof(min)) = +Inf
identity_of_reduction(::typeof(+)) = 0
identity_of_reduction(::typeof(*)) = 1
```

src/diagnostics/Diagnostics.jl
```diff
@@ -0,0 +1,634 @@
...
+        # we don't have a reduction, we don't have to do anything)
+        #
+        # ClimaAtmos defines methods for identity_of_reduction for standard
+        # reduction_time_func in reduction_identities.jl
+        reset_accumulator! =
+            isa_reduction ?
+            () -> begin
+                # identity_of_reduction works by dispatching over Val{operation}
+                identity =
+                    identity_of_reduction(Val(diag.reduction_time_func))
```
simonbyrne, line 583:
```suggestion
                    identity_of_reduction(diag.reduction_time_func)
```

src/diagnostics/Diagnostics.jl
```diff
@@ -0,0 +1,634 @@
...
+        # which that will be the newly computed one). If we have a reduction, we apply it
+        # point-wise
+        reduction = isa_reduction ? diag.reduction_time_func : (_, y) -> y
+
+        # If we have a reduction, we have to reset the accumulator to its neutral state. (If
+        # we don't have a reduction, we don't have to do anything)
+        #
+        # ClimaAtmos defines methods for identity_of_reduction for standard
+        # reduction_time_func in reduction_identities.jl
+        reset_accumulator! =
```
simonbyrne, line 578: I'm a bit wary of having too many closures:
- they make stack traces harder to read (anonymous functions don't get nice names, names get mangled)
- they can capture variables unexpectedly, making it harder to figure out what is in scope

As an alternative, I would propose defining a function at top-level
```julia
function reset_accumulator!(diag_storage, diag::ScheduledDiagnostic)
    if !isnothing(diag.reduction_time_func) # this check should be inferred at compile time, so function would become a no-op
        # identity_of_reduction works by dispatching over operation
        identity = identity_of_reduction(diag.reduction_time_func)
        float_type = eltype(diag_storage)
        identity_ft = convert(float_type, identity)
        diag_storage .= identity_ft
    end
end
```

src/diagnostics/Diagnostics.jl
```diff
@@ -0,0 +1,634 @@
+# Diagnostics.jl
```
simonbyrne, line 1: I think it makes sens to make this a module
```suggestion
module Diagnostics
# Diagnostics.jl
```

src/diagnostics/Diagnostics.jl
```diff
@@ -0,0 +1,634 @@
...
+                    output_callback,
+                    diag.output_every,
+                    skip_first = true,
+                ),
+            ],
+        )
+    end
+
+    return callbacks
+end
```
simonbyrne, line 634:
```suggestion
end
end # module
```

src/diagnostics/Diagnostics.jl
simonbyrne: And similarly
```julia
function apply_reduction!(diag_storage, reductfn, value)
   diag_storage .= reductfn.(diag_storage, value)
end
function apply_reduction!(diag_storage, reductfn::Nothing, value)
   diag_storage .= value
end
```

src/diagnostics/Diagnostics.jl
```diff
@@ -0,0 +1,634 @@
...
+                float_type = eltype(storage[diag])
+                identity_ft = convert(float_type, identity)
+                storage[diag] .= identity_ft
+            end : () -> nothing
+
+        compute_callback =
+            integrator -> begin
+                # FIXME: Change when ClimaCore overrides .= for us to avoid multiple allocations
+                value = variable.compute_from_integrator(integrator, nothing)
+                storage[diag] .= reduction.(storage[diag], value)
```
simonbyrne, line 594:
```suggestion
                apply_reduction!(storage[diag], diag.reduction_time_func, value)
```

src/diagnostics/Diagnostics.jl
```diff
@@ -0,0 +1,634 @@
...
+        output_callback =
+            integrator -> begin
+                # Any operations we have to perform before writing to output?
+                # Here is where we would divide by N to obtain an arithmetic average
+                diag.pre_output_hook!(storage[diag], counters[diag])
+
+                # Write to disk
+                diag.output_writer(storage[diag], diag, integrator)
+
+                reset_accumulator!()
```
simonbyrne, line 608:
```suggestion
                reset_accumulator!(storage[diag], diag)
```

src/diagnostics/defaults/moisture_model.jl
```diff
@@ -0,0 +1,22 @@
+# FIXME: Gabriele added this as an example. Put something meaningful here!
+function get_default_diagnostics(::T) where {T <: DryModel}
+    return vcat(
+        get_daily_average("air_density"),
```
simonbyrne, line 4: I'm not sure about the variadic option. Part of me thinks it would be better to return a single value, and require users to broadcast, e.g.
```julia
        get_daily_average.(["air_density", "air_pressure"])
```

src/diagnostics/default_diagnostics.jl
```diff
@@ -0,0 +1,122 @@
...
+                variable = ALL_DIAGNOSTICS[long_name],
+                compute_every = :timestep,
+                output_every = period, # seconds
+                reduction_time_func = reduction,
+                output_writer = output_writer,
+                pre_output_hook! = pre_output_hook!,
+            ) for long_name in long_names
+        ]
+    end
+end
```
simonbyrne, line 57:
```suggestion
function diagnostic_reduction(
    long_name,
    period,
    reduction;
    pre_output_hook! = (accum, count) -> nothing,
)
      ScheduledDiagnosticTime(
          variable = ALL_DIAGNOSTICS[long_name],
          compute_every = :timestep,
          output_every = period, # seconds
          reduction_time_func = reduction,
          output_writer = output_writer,
          pre_output_hook! = pre_output_hook!,
      ) 
end

hourly_average(long_name) = diagnostic_reduction(long_name, 60*60, +; pre_output_hook=average_periodic)
```

src/diagnostics/default_diagnostics.jl
```diff
@@ -0,0 +1,122 @@
...
+
+function average_pre_output_hook!(accum, counter)
+    @. accum = accum / counter
+    nothing
+end
+
+"""
+    get_daily_max(long_names...; output_writer = HDF5Writer())
```
simonbyrne, line 65:
```suggestion
    get_daily_max(long_names...; output_writer = HDF5Writer())
```
Not sure `get_XX` is a good name, since it doesn't "get" anything.  `daily_max` i thing is fine?

docs/make_diagnostic_table.jl
```diff
@@ -0,0 +1,28 @@
...
+# have that, we will assume that we are running from the project root. If this
+# code is run from anywhere but these two places, mkdocs will fail to find
+# availbale_diagnostics.md
+prefix = basename(pwd()) == "docs" ? "" : "docs/"
+
+out_path = "$(prefix)src/available_diagnostics.md"
+
+open(out_path, "w") do file
+
+    write(file, "# Available diagnostics\n\n")
```
simonbyrne, line 16:
```suggestion
    write(file, "# Available diagnostic variables\n\n")
```

src/diagnostics/diagnostics.jl
```diff
@@ -0,0 +1,681 @@
...
+                             one will be allocated. To avoid extra allocations, this
+                             function should perform the calculation in-place (i.e., using
+                             `.=`).
+"""
+Base.@kwdef struct DiagnosticVariable{T}
+    short_name::String
+    long_name::String
+    units::String
+    comments::String
+    compute_from_integrator!::T
```
charleskawczynski, line 93: The entire integrator cache is stored in `integrator.p`, maybe we can remove a few layers of complexity if we start off assuming that we have something like `integrator.p.X` where `X` are the components defined in this PR? That way, we can dispatch on them, and avoid need / entanglement of the integrator.

docs/src/diagnostics.md
```diff
@@ -0,0 +1,277 @@
...
+
+function compute_relative_humidity_from_integrator!(
+    out,
+    integrator,
+    moisture_model::T,
+) where {T <: Union{EquilMoistModel, NonEquilMoistModel}}
+    # FIXME: Avoid extra allocations when ClimaCore overloads .= for this use case
+    # We will want: out .= integrator.u.c.ρ
+    thermo_params = CAP.thermodynamics_params(integrator.p.params)
+    return TD.relative_humidity.(thermo_params, integrator.p.ᶜts)
```
charleskawczynski, line 248: Can we address these TODOs before merging? Fixing this may have a big impact on the overall design.

src/diagnostics/diagnostics.jl
Sbozzolo: I don't think I understand what you are proposing. Can you share some pseudo code of what you have in mind?

I do agree that we can remove the reference to the integrator in the definition of the diagnostic variable and take in `state`, and `params` directly. And I like the idea because it feel more natural and motivated to compute something from a physical state as opposed to an integrator, which should probably be considered an implementation detail. 

So, `compute_from_integrator!(out, integrator)` might become `!compute(out, state, params)`. Then, we call this function with the correct arguments when we set up the callbacks.

src/diagnostics/core_diagnostics.jl
```diff
@@ -0,0 +1,55 @@
...
+    # We will want: out .= integrator.u.c.ρ
+    thermo_params = CAP.thermodynamics_params(integrator.p.params)
+    return TD.relative_humidity.(thermo_params, integrator.p.ᶜts)
+end
+
+compute_relative_humidity_from_integrator!(out, integrator) =
+    compute_relative_humidity_from_integrator!(
+        out,
+        integrator,
+        integrator.p.atmos,
```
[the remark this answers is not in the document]
Sbozzolo: Thanks!

src/diagnostics/diagnostics.jl
charleskawczynski: > I do agree that we can remove the reference to the integrator in the definition of the diagnostic variable and take in `state`, and `params` directly. And I like the idea because it feel more natural and motivated to compute something from a physical state as opposed to an integrator, which should probably be considered an implementation detail.

I'm proposing exactly this. Basically, rather than having functions in here that take the integrator (`foo(integrator)`), we instead pass in the parts (e.g., `foo(u, p, t)`)

szy21:
Thanks! I will update the core_diagnostics and default_diagnostics this Friday.

src/diagnostics/core_diagnostics.jl
```diff
@@ -16,19 +27,160 @@ add_diagnostic_variable!(
...
+add_diagnostic_variable!(
+    short_name = "air_temperature",
+    long_name = "Air Temperature",
+    units = "K",
+    comments = "Temperature of air",
+    compute! = (out, state, cache, time) -> begin
+        # FIXME: Avoid extra allocations when ClimaCore overloads .= for this use case
+        # We will want: out .= integrator.u.c.ρ
+        thermo_params = CAP.thermodynamics_params(cache.params)
+        return copy(TD.air_temperature.(thermo_params, cache.ᶜts))
```
Sbozzolo, line 89: We don't need copies here because `TD.air_temperature.` already allocates new memory (same for other functions below)

src/diagnostics/core_diagnostics.jl
```diff
@@ -16,19 +27,160 @@ add_diagnostic_variable!(
...
+    long_name = "Vertical component of relative vorticity",
+    units = "s^-1",
+    comments = "Vertical component of relative vorticity",
+    compute! = (out, state, cache, time) -> begin
+        # FIXME: Avoid extra allocations when ClimaCore overloads .= for this use case
+        # We will want: out .= integrator.u.c.ρ
+        ᶜvort = @. Geometry.WVector(curlₕ(state.c.uₕ))
+        if cache.do_dss
+            Spaces.weighted_dss!(ᶜvort)
+        end
```
Sbozzolo, line 138: It would be nice to have a comment explaining why we have this if-else

src/diagnostics/core_diagnostics.jl
```diff
@@ -38,24 +190,245 @@ function compute_relative_humidity!(
...
+
+###
+# Surface temperature (2d)
+###
+add_diagnostic_variable!(
+    short_name = "surface_temperature",
+    long_name = "Surface Temperature",
+    units = "K",
+    comments = "Temperature of the surface",
+    compute! = TD.air_temperature.(thermo_params, cache.sfc_conditions.ts),
```
Sbozzolo, line 257: `compute!` has to take four specific arguments, so we have to wrap it around another function

src/diagnostics/core_diagnostics.jl
```diff
@@ -38,24 +190,245 @@ function compute_relative_humidity!(
...
+###
+# Eastward surface drag component (2d)
+###
+function drag_vector!(state, cache)
+    sfc_local_geometry =
+    Fields.level(Fields.local_geometry_field(state.f), Fields.half)
+    surface_ct3_unit =
+        CT3.(unit_basis_vector_data.(CT3, sfc_local_geometry))
+    (; ρ_flux_uₕ) = cache.sfc_conditions
+    sfc_flux_momentum =
```
Sbozzolo, line 269: Adding a `return` statement would make the intentions more clear

Sbozzolo:
I think we can avoid using `Val`s. My impression is that the `Val`s here are mostly used to avoid repeating the base case with the error message when there's a mismatch between the diagnostic we want to compute and the model. 
Each diagnostic has to define its method for `compute_variable!` and wrap it around a function with the correct signature (e.g., `compute_specific_humidity!(...) =
    compute_variable!(Val(:specific_humidity), ...)`).

To me, it feels like that adds unnecessary type complexity and that these shouldn't be methods but different functions. With the exception of the error message, all the `compute_variable!(Val{XXX})` are conceptually distinct and end up in a named function.

What I propose is to add a function
```julia
error_diagnostic_variable(variable, model::T, message = "Cannot compute $variable with model = $T") where {T} = error(message)
```
Then, we define two methods for each function, and call `error_diagnostic_variable` in the base one. This allows us to easily define custom error messages for cases where we want to be more informative.

Considering that this would remove all the `Val`s, the total number of lines of code will be equal or fewer.

What do you think?

szy21:
I think @LenkaNovak added the core diagnostics and the `Val`s? I will take a closer look later this week.

Sbozzolo:
> I think @LenkaNovak added the core diagnostics and the `Val`s? I will take a closer look later this week.

Yes, we will figure it out.

It would be super useful to have a paragraph to explain very clearly our naming conventions. Would be able to write that too?

src/diagnostics/core_diagnostics.jl
LenkaNovak: `do_dss` is probably dependent on the discretization, but maybe @sriharshakandala or @dennisYatunin will know more.

szy21:
> > I think @LenkaNovak added the core diagnostics and the `Val`s? I will take a closer look later this week.
> 
> Yes, we will figure it out.
> 
> It would be super useful to have a paragraph to explain very clearly our naming conventions. Would be able to write that too?

Yes, we need to change the `short_name` and `long_name` in the core diagnostics, following CMIP. I can do that on Friday, and add the documentation on our naming convention.

LenkaNovak:
> I think we can avoid using `Val`s. My impression is that the `Val`s here are mostly used to avoid repeating the base case with the error message when there's a mismatch between the diagnostic we want to compute and the model. Each diagnostic has to define its method for `compute_variable!` and wrap it around a function with the correct signature (e.g., `compute_specific_humidity!(...) = compute_variable!(Val(:specific_humidity), ...)`).
> 
> To me, it feels like that adds unnecessary type complexity and that these shouldn't be methods but different functions. With the exception of the error message, all the `compute_variable!(Val{XXX})` are conceptually distinct and end up in a named function.
> 
> What I propose is to add a function
> 
> ```julia
> error_diagnostic_variable(variable, model::T, message = "Cannot compute $variable with model = $T") where {T} = error(message)
> ```
> 
> Then, we define three methods for each function, and call `error_diagnostic_variable` in the base one. This allows us to easily define custom error messages for cases where we want to be more informative.
> 
> ```julia
> compute_relative_humidity!(out, state, cache, time) =
>     compute_relative_humidity!(out, state, cache, time, cache.model)
> compute_relative_humidity!(_, _, _, _, model::T) where {T} = error_diagnostic_variable("relative_humidity", model)
> 
> function compute_relative_humidity!(out, state, cache, time, ::T) where {T <: Union{EquilMoistModel, NonEquilMoistModel}}
>     thermo_params = CAP.thermodynamics_params(cache.params)
>     return TD.relative_humidity.(thermo_params, cache.ᶜts)
> end
> ```
> 
> There is more boilerplate.
> 
> Considering that this would remove all the `Val`s, the total number of lines of code will be equal or fewer.
> 
> What do you think?

For us the `Val` method has worked pretty well in the coupler, because it allows the user easily to define custom diagnostics and dispatch general methods on them, without having to define new concrete type each time. The `Val` method also avoids having to define the equivalent of `compute_relative_humidity!(_, _, _, _, model::T) where {T} = error_diagnostic_variable("relative_humidity", model)` for each variable separately, thus avoiding repetition. But if we want to be able to have customizable error messages for each diagnostic, then I agree that `Val` method makes less sense.

Sbozzolo:
> > I think we can avoid using `Val`s. My impression is that the `Val`s here are mostly used to avoid repeating the base case with the error message when there's a mismatch between the diagnostic we want to compute and the model. Each diagnostic has to define its method for `compute_variable!` and wrap it around a function with the correct signature (e.g., `compute_specific_humidity!(...) = compute_variable!(Val(:specific_humidity), ...)`).
> > To me, it feels like that adds unnecessary type complexity and that these shouldn't be methods but different functions. With the exception of the error message, all the `compute_variable!(Val{XXX})` are conceptually distinct and end up in a named function.
> > What I propose is to add a function
> > ```julia
> > error_diagnostic_variable(variable, model::T, message = "Cannot compute $variable with model = $T") where {T} = error(message)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Then, we define three methods for each function, and call `error_diagnostic_variable` in the base one. This allows us to easily define custom error messages for cases where we want to be more informative.
> > ```julia
> > compute_relative_humidity!(out, state, cache, time) =
> >     compute_relative_humidity!(out, state, cache, time, cache.model)
> > compute_relative_humidity!(_, _, _, _, model::T) where {T} = error_diagnostic_variable("relative_humidity", model)
> > 
> > function compute_relative_humidity!(out, state, cache, time, ::T) where {T <: Union{EquilMoistModel, NonEquilMoistModel}}
> >     thermo_params = CAP.thermodynamics_params(cache.params)
> >     return TD.relative_humidity.(thermo_params, cache.ᶜts)
> > end
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > There is more boilerplate.
> > Considering that this would remove all the `Val`s, the total number of lines of code will be equal or fewer.
> > What do you think?
> 
> For us the `Val` method has worked pretty well in the coupler, because it allows the user easily to define custom diagnostics and dispatch general methods on them, without having to define new concrete type each time. The `Val` method also avoids having to define the equivalent of `compute_relative_humidity!(_, _, _, _, model::T) where {T} = error_diagnostic_variable("relative_humidity", model)` for each variable separately, thus avoiding repetition. But if we want to be able to have customizable error messages for each diagnostic, then I agree that the`Val` method makes less sense.

I think that users would not be able to define custom diagnostics and dispatch methods simply by using `Val` in this module. The `compute!` function must take the given arguments `out, state, cache, time`. So, one would always have to define a function with that signature.

If users want to define a new diagnostic, they will have to define an object `DiagnosticVariable` and provide the `compute!` function. 

(Also, we can always have customized error messages using `Val` by defining a specialized method for those cases where we want customized errors)

src/diagnostics/core_diagnostics.jl
sriharshakandala: The `do_dss` checks if the underlying `hspace` is a spectral element space. https://github.com/CliMA/ClimaAtmos.jl/blob/main/src/cache/cache.jl#L64
It is not immediately clear to me why a dss is needed here. May be @dennisYatunin knows!

LenkaNovak:
> > > I think we can avoid using `Val`s. My impression is that the `Val`s here are mostly used to avoid repeating the base case with the error message when there's a mismatch between the diagnostic we want to compute and the model. Each diagnostic has to define its method for `compute_variable!` and wrap it around a function with the correct signature (e.g., `compute_specific_humidity!(...) = compute_variable!(Val(:specific_humidity), ...)`).
> > > To me, it feels like that adds unnecessary type complexity and that these shouldn't be methods but different functions. With the exception of the error message, all the `compute_variable!(Val{XXX})` are conceptually distinct and end up in a named function.
> > > What I propose is to add a function
> > > ```julia
> > > error_diagnostic_variable(variable, model::T, message = "Cannot compute $variable with model = $T") where {T} = error(message)
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Then, we define three methods for each function, and call `error_diagnostic_variable` in the base one. This allows us to easily define custom error messages for cases where we want to be more informative.
> > > ```julia
> > > compute_relative_humidity!(out, state, cache, time) =
> > >     compute_relative_humidity!(out, state, cache, time, cache.model)
> > > compute_relative_humidity!(_, _, _, _, model::T) where {T} = error_diagnostic_variable("relative_humidity", model)
> > > 
> > > function compute_relative_humidity!(out, state, cache, time, ::T) where {T <: Union{EquilMoistModel, NonEquilMoistModel}}
> > >     thermo_params = CAP.thermodynamics_params(cache.params)
> > >     return TD.relative_humidity.(thermo_params, cache.ᶜts)
> > > end
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > There is more boilerplate.
> > > Considering that this would remove all the `Val`s, the total number of lines of code will be equal or fewer.
> > > What do you think?
> > 
> > 
> > For us the `Val` method has worked pretty well in the coupler, because it allows the user easily to define custom diagnostics and dispatch general methods on them, without having to define new concrete type each time. The `Val` method also avoids having to define the equivalent of `compute_relative_humidity!(_, _, _, _, model::T) where {T} = error_diagnostic_variable("relative_humidity", model)` for each variable separately, thus avoiding repetition. But if we want to be able to have customizable error messages for each diagnostic, then I agree that the`Val` method makes less sense.
> 
> I think that users would not be able to define custom diagnostics and dispatch methods simply by using `Val` in this module. The `compute!` function must take the given arguments `out, state, cache, time`. So, one would always have to define a function with that signature.

Why do you think that? Not sure exactly whether this is what you mean, but the idea is/was that there is a predefined general method in the module that can dispatch on any `Val`, even if the `Val{:variable}` is defined outside of the module by the user. We dispatch on Val a lot in the coupler e.g. [here](https://github.com/CliMA/ClimaCoupler.jl/blob/c4356366df6e01b78e7bcf85f54ca78d0f0fa386/src/Interfacer.jl#L79). In this case though, it was purely to reduce having to repeat the error message, but from what you say that's not an issue, since we want to be able to customize errors, so no worries. 
> 
> If users want to define a new diagnostic, they will have to define an object `DiagnosticVariable` and provide the `compute!` function.
Sounds good!
> (Also, we can always have customized error messages using `Val` by defining a specialized method for those cases where we want customized errors)
True, we could do that, but at that point we might as well use your method above. ;-)

Sbozzolo:
> Why do you think that? Not sure exactly whether this is what you mean, but the idea is/was that there is a predefined general method in the module that can dispatch on any Val, even if the Val{:variable} is defined outside of the module by the user. We dispatch on Val a lot in the coupler e.g. [here](https://github.com/CliMA/ClimaCoupler.jl/blob/c4356366df6e01b78e7bcf85f54ca78d0f0fa386/src/Interfacer.jl#L79). In this case though, it was purely to reduce having to repeat the error message, but from what you say that's not an issue, since we want to be able to customize errors, so no worries.

I was specifically referring to the current implementation in `ClimaAtmos`. If you just add a method `compute_variable!(Val{:new_variable})`, you would not be able to use it unless you also define a `DiagnosticVariable` that calls that method with a specific signature.

src/diagnostics/core_diagnostics.jl
dennisYatunin: Not sure, but my guess is that `curlₕ.(Y.c.uₕ)` generates a Field that is discontinuous across element boundaries (unless we're in a column, in which case `do_dss` is false), and we only want to use continuous Fields for diagnostics.

Sbozzolo: Can `weighted_dss!` automatically recognize if we are in a space where dss makes sense or not?

dennisYatunin: Not currently. We could move this if-statement into `weighted_dss!` in ClimaCore, thought I feel like that would be hiding a bit too much under the hood. Since DSS only makes sense on a spectral element discretization, it would be confusing to see a call to `weighted_dss!` on a column space, even if it gets reduced to a no-op in ClimaCore.

Sbozzolo: I think there's an opportunity here to make things more clear. If one doesn't know what DSS is, this code is completely obscure. If one knows what DSS is, the intent of the code is still obscure. Maybe `weighted_dss!` could be renamed to `enforce_continuity!` ? 

I don't know what's the best way.

dennisYatunin: You could definitely make an `enforce_continuity!` function in ClimaAtmos that dispatches based on the type of the space, if you want to hide this if-statement. Renaming this in ClimaCore might not be such a good idea, since `enforce_continuity!` is less specific than `weighted_dss!` and could lead to ambiguity in the future. @sriharshakandala, thoughts on renaming `weighted_dss!` in ClimaCore?

sriharshakandala: In this case, a name like `smoothen` might make sense as this is what is being done here! I believe we already made a dss a no-op for non-spectral element fields in ClimaCore!

sriharshakandala: However, at other locations, `weighted_dss` makes more sense! We need to rename `dss_op` to `weighted_dss_op` since the underlying operation is weighted dss (i.e. M_-1 * dss(field)) and not simple dss.

dennisYatunin: Ah, I didn't know dss can be a no-op now. In that case, the only use of `do_dss` in ClimaAtmos should be to determine whether or not to allocate a cache for dss, and you can probably remove the if-statement here. As for the name change, maybe we could run that by the rest of the group first; I'm sure people will have lots of opinions!

Sbozzolo:
Attached is a simple file that checks that the new diagnostics are the same as the old ones.

For `helz_surez_equilmost`, I get:
```
ERROR DAY 0
air_pressure : 0.0
relative_humidity : 0.0
surface_evaporation : 0.0
surface_energy_flux : 0.0
air_temperature : 0.0
air_potential_temperature : 0.0
specific_humidity : 0.0
surface_temperature : 0.0
surface_specific_humidity : 0.0
eastward_wind : 0.0
northward_wind : 0.0
vertical_wind : 0.0
ERROR DAY 2
air_pressure : 0.0
relative_humidity : 0.0
surface_evaporation : 0.0
surface_energy_flux : 0.0
air_temperature : 0.0
air_potential_temperature : 0.0
specific_humidity : 0.0
surface_temperature : 0.0
surface_specific_humidity : 0.0
eastward_wind : 0.0
northward_wind : 0.0
vertical_wind : 0.0
```
Perfect agreement.

[compare_diag.py.txt](https://github.com/CliMA/ClimaAtmos.jl/files/12602010/compare_diag.py.txt)

docs/src/diagnostics.md
```diff
@@ -0,0 +1,306 @@
...
+accumulator in place. The return value of `pre_output_hook!` is discarded. An
+example of `pre_output_hook!` to compute the arithmetic average is
+`pre_output_hook!(acc, N) = @. acc = acc / N`. `ClimaAtmos` provides an alias to
+the function needed to compute averages `ClimaAtmos.average_pre_output_hook!`.
+
+For custom reductions, it is necessary to also specify the identity of operation
+by defining a new method to `identity_of_reduction`. `identity_of_reduction` is
+a function that takes a `Val{op}` argument, where `op` is the operation for
+which we want to define the identity. For instance, for the `max`,
+`identity_of_reduction` would be `identity_of_reduction(::Val{max}) = -Inf`. The
```
simonbyrne, line 125:
```suggestion
a function that takes a `op` argument, where `op` is the operation for
which we want to define the identity. For instance, for the `max`,
`identity_of_reduction` would be `identity_of_reduction(::typeof(max)) = -Inf`. The
```

src/diagnostics/writers.jl
```diff
@@ -0,0 +1,161 @@
...
+  with different reductions)
+- All variables in one file/each variable in its own file
+- All timesteps in one file/each timestep in its own file
+- Writing the correct attributes
+- Overriding simulation.output_dir (e.g., if the path starts with /)
+- ...more features/options
+
+"""
+
+function HDF5Writer()
```
simonbyrne, line 33: I know it's in a different namespace, but it's a bit confusing that this returns a function rather than a `ClimaCore.InputOutput.HDF5Writer` object.

simonbyrne: What if instead of passing `ClimaAtmos.Diagnostics.HDF5Writer()`, you instead passed `ClimaCore.InputOutput.HDF5Writer` (i.e. the type), and then we dispatched on that?

Sbozzolo: We can do that, but wouldn't we have to prepare a `ClimaCore.InputOutput.HDF5Writer` for each file we want to write?

In general, I agree that the writer interface can be significantly improved

src/diagnostics/diagnostics_utils.jl
```diff
@@ -0,0 +1,75 @@
...
+            end
+            if hours > 0
+                period *= "$(hours)h_"
+            end
+            if minutes > 0
+                period *= "$(minutes)m_"
+            end
+            if seconds > 0
+                period *= "$(seconds)s_"
+            end
```
charleskawczynski, line 65:
```suggestion
            days > 0 && (period *= "$(days)d_")
            hours > 0 && (period *= "$(hours)h_")
            minutes > 0 && (period *= "$(minutes)m_")
            seconds > 0 && (period *= "$(seconds)s_")
```
The compactness of single line if-statements might be nice to use here.

src/diagnostics/diagnostics_utils.jl
```diff
@@ -0,0 +1,75 @@
...
+
+Return a compact, unique-ish, identifier generated from the given information.
+
+`output_every` is interpreted as in seconds if `units_are_seconds` is `true`. Otherwise, it
+is interpreted as in units of number of iterations.
+
+This function is useful for filenames and error messages.
+
+ """
+
```
charleskawczynski, line 23:
```suggestion
```
Newlines between doc strings and their functions are not allowed

src/diagnostics/diagnostics_utils.jl
Sbozzolo: I did this originally and got this error `syntax: invalid assignment location`

src/diagnostics/defaults/moisture_model.jl
```diff
@@ -0,0 +1,19 @@
+# FIXME: Gabriele added this as an example. Put something meaningful here!
+function get_default_diagnostics(::T) where {T <: DryModel}
```
charleskawczynski, line 2:
```suggestion
function get_default_diagnostics(::DryModel)
```
We don't need to force specialization on this

src/diagnostics/diagnostics_utils.jl
Sbozzolo:
```julia
a = "a"
true && a*="b"
```

```
ERROR: syntax: invalid assignment location "true && a" around REPL[9]:1
```

src/diagnostics/default_diagnostics.jl
```diff
@@ -0,0 +1,194 @@
...
+    common_diagnostics(60 * 60, min, output_writer, short_names)[1]
+
+"""
+    hourly_averages(short_names...; output_writer = HDF5Writer())
+
+
+Return a list of `ScheduledDiagnostics` that compute the hourly average for the given variables.
+"""
+
+# An average is just a sum with a normalization before output
```
charleskawczynski, line 168:
```suggestion
# An average is just a sum with a normalization before output
"""
```
Newlines between doc strings and their functions are not allowed. It looks like there may be a bunch of these, let's apply elsewhere, too.

src/diagnostics/default_diagnostics.jl
```diff
@@ -0,0 +1,194 @@
...
+end
+
+"""
+    daily_maxs(short_names...; output_writer = HDF5Writer())
+
+
+Return a list of `ScheduledDiagnostics` that compute the daily max for the given variables.
+"""
+daily_maxs(short_names...; output_writer = HDF5Writer()) =
+    common_diagnostics(24 * 60 * 60, max, output_writer, short_names...)
```
charleskawczynski, line 71: Won't users generally want to have a single writer, so that we can efficiently read/write to disc? If a user calls any of these functions without the output writer, then what's the default behavior? I guess I'm wondering if we should just make `output_writer` non-optional.

src/diagnostics/Diagnostics.jl
```diff
@@ -0,0 +1,40 @@
...
+# Abbreviations (following utils/abbreviations.jl)
+const curlₕ = Operators.Curl()
+const CT3 = Geometry.Contravariant3Vector
+const ᶜinterp = Operators.InterpolateF2C()
+# TODO: Implement proper extrapolation instead of simply reusing the first
+# interior value at the surface.
+const ᶠinterp = Operators.InterpolateC2F(
+    bottom = Operators.Extrapolate(),
+    top = Operators.Extrapolate(),
+)
```
charleskawczynski, line 36: Can we just reuse `utils/abbreviations.jl`? (and do `import ..`)

src/diagnostics/defaults/moisture_model.jl
Sbozzolo: Right

src/diagnostics/Diagnostics.jl
Sbozzolo: If we want to do so, we should move out the definitions of `AxisTensor` and `convert` from `abbreviations.jl`, otherwise we get a redefinition.

src/diagnostics/default_diagnostics.jl
Sbozzolo: Yes, at the moment each diagnostic would get its own writer if you don't specify anything. But I think that I should work on this, and Simon also had comments on the writers.

src/diagnostics/diagnostics_utils.jl
charleskawczynski: parenthesis are needed for assignments:
```julia
julia> a = "a"
"a"

julia> true && a*="b"
ERROR: syntax: invalid assignment location "true && a" around REPL[2]:1
Stacktrace:
 [1] top-level scope
   @ REPL[2]:1

julia> true && (a*="b")
"ab"
```

src/diagnostics/Diagnostics.jl
charleskawczynski: Ah, right, we need to get rid of those..

docs/src/diagnostics.md
simonbyrne:
```suggestion
    # FIXME: Avoid extra allocations when ClimaCore overloads .= for this use case
    # We will want: out .= integrator.u.c.ρ
    thermo_params = CAP.thermodynamics_params(integrator.p.params)
    if isnothing(out)
        return TD.relative_humidity.(thermo_params, integrator.p.ᶜts)
    else
        return out .= TD.relative_humidity.(thermo_params, integrator.p.ᶜts)
    end
```

src/diagnostics/diagnostic.jl
```diff
@@ -0,0 +1,694 @@
...
+
+        compute_callback =
+            integrator -> begin
+                # FIXME: Change when ClimaCore overrides .= for us to avoid multiple allocations
+                storage[diag] .= variable.compute!(
+                    storage[diag],
+                    integrator.u,
+                    integrator.p,
+                    integrator.t,
+                )
```
simonbyrne, line 644:
```suggestion
                variable.compute!(
                    storage[diag],
                    integrator.u,
                    integrator.p,
                    integrator.t,
                )
```

or if you really want to be pedantic
```
                storage[diag] = variable.compute!(
                    storage[diag],
                    integrator.u,
                    integrator.p,
                    integrator.t,
                )
```

src/diagnostics/core_diagnostics.jl
```diff
@@ -0,0 +1,442 @@
...
+###
+add_diagnostic_variable!(
+    short_name = "air_density",
+    long_name = "Air Density",
+    units = "kg m^-3",
+    comments = "Density of air, a prognostic variable",
+    compute! = (out, state, cache, time) -> begin
+        # FIXME: Avoid extra allocations when ClimaCore overloads .= for this use case
+        # We will want: out .= integrator.u.c.ρ
+        return copy(state.c.ρ)
```
simonbyrne, line 60:
```suggestion
        if isnothing(out)copy(state.c.ρ)
```

src/diagnostics/writers.jl
```diff
@@ -0,0 +1,160 @@
...
+        var = diagnostic.variable
+        time = integrator.t
+
+        # diagnostic here is a ScheduledDiagnosticIteration. If we want to obtain a
+        # descriptive name (e.g., something with "daily"), we have to pass the timestep as
+        # well
+
+        output_path = joinpath(
+            integrator.p.simulation.output_dir,
+            "$(diagnostic.output_name)_$time.h5",
```
simonbyrne, line 45: I think by default we should write everything at the same output interval to one file.

src/diagnostics/writers.jl
```diff
@@ -0,0 +1,160 @@
...
+        )
+
+        hdfwriter = InputOutput.HDF5Writer(output_path, integrator.p.comms_ctx)
+        InputOutput.HDF5.write_attribute(hdfwriter.file, "time", time)
+        InputOutput.HDF5.write_attribute(
+            hdfwriter.file,
+            "long_name",
+            var.long_name,
+        )
```
simonbyrne, line 54: I think attributes should be attached to the dataset (field) that you're writing to.

simonbyrne:
Overall, I think this is mostly ready. Still need to
- Figure out what "long_name" and "short_name" should be (which is the CF standard name? which is unique?)
- performance and allocations: it would be good to see what the overhead is in a profile (especially the reduction ones which are called every time step)

simonbyrne:
From discussion today:

Variable should have
- short name / id: unique
- standard name (optional): this may not be unique
- long name: more descriptive

e.g.

| short name | standard name | long name |
|-----|----------|------------------------|
| `ta` | `air_temperature` | `Air Temperature` |
| `tas` | `air_temperature` | `Near Surface Air Temperature` |
| `ua` | `eastward_wind` | `Eastward Wind` |

Output variables would use a then inherit these:
- short name: should still be unique, inherit name from reduction (e.g. `tasmax`/`tasmin`)
   - this is used as the dataset / variable name
- standard name: same as variable (e.g. `air_temperature` for both)
   - written as attribute
- long name: should include reduction info (e.g. `Daily Maximum Near-Surface Air Temperature`)
   - written as attribute
- should also inherit other info (units, etc)

src/diagnostics/defaults/moisture_model.jl
```diff
@@ -0,0 +1,19 @@
...
+            output_every = 86400, # seconds
+            reduction_time_func = min,
+            output_writer = HDF5Writer(),
+        ),
+        ScheduledDiagnosticIterations(
+            variable = get_diagnostic_variable("air_density"),
+            compute_every = 1,
+            output_every = 1, # iteration
+            output_writer = HDF5Writer(),
+        ),
```
simonbyrne, line 17:
```
ScheduldDiagnosticTime(
    variable = get_diagnostic_variable("tas"),
    compute_every = :timestep,
    output_every = 86400,
    reduction_time_func = max,
    name = "tasmax",
    output_writer = HDF5Writer(),
)
```

src/diagnostics/default_diagnostics.jl
```diff
@@ -0,0 +1,236 @@
...
+
+    append!(defaults, core_default_diagnostics())
+
+    for field in fieldnames(AtmosModel)
+        def_model = default_diagnostics(getfield(model, field))
+        append!(defaults, def_model)
+    end
+
+    return defaults
+end
```
charleskawczynski, line 26:
```suggestion
function default_diagnostics(model::AtmosModel)
    # TODO: Probably not the most elegant way to do this...
    return [
        core_default_diagnostics(),
        map(fieldnames(AtmosModel)) do field
            default_diagnostics(getfield(model, field))
        end...
    ]
end
```

simonbyrne:
Can you resolve the comments that are closed?

charleskawczynski:
Let's also make sure to add a flame graph showing the runtime performance. We can add a separate flame script if that's easier.

Sbozzolo:
> Let's also make sure to add a flame graph showing the runtime performance. We can add a separate flame script if that's easier.

Yes, it's my last item to tackle:)

Sbozzolo:
> > Let's also make sure to add a flame graph showing the runtime performance. We can add a separate flame script if that's easier.
> 
> Yes, it's my last item to tackle:)

Flamegraph added, so I think that everything is in place for this to be merged.

Sbozzolo:
@szy21 @simonbyrne @charleskawczynski, if you want to have another look

config/model_configs/sphere_aquaplanet_rhoe_equilmoist_allsky_gw_raw_zonallyasymmetric.yml
szy21, line 1: This job only runs for 18hours, so it won't output the default daily diagnostics. We can add default diagnostics for this one instead: `sphere_held_suarez_rhoe_equilmoist_hightop_sponge`.

src/diagnostics/radiation_diagnostics.jl
```diff
@@ -0,0 +1,307 @@
+# This file is included in Diagnostics.jl
+
+# Long/short-wave fluxes
```
szy21, line 3:
```suggestion
# Radiative fluxes
```

src/diagnostics/radiation_diagnostics.jl
```diff
@@ -0,0 +1,307 @@
...
+
+###
+# Donwelling shortwave radiation (3d)
```
szy21, line 6:
```suggestion
# Downwelling shortwave radiation (3d)
```

src/diagnostics/radiation_diagnostics.jl
```diff
@@ -0,0 +1,307 @@
...
+    short_name = "rlu",
+    long_name = "Upwelling Longwave Radiation",
+    standard_name = "surface_upwelling_longwave_flux_in_air",
+    units = "W m^-2",
+    comments = "Upwelling longwave radiation",
+    compute! = compute_rlu!,
+)
+
+###
+# Donwelling clear sky shortwave radiation (3d)
```
szy21, line 158:
```suggestion
# Downwelling clear sky shortwave radiation (3d)
```

src/diagnostics/default_diagnostics.jl
```diff
@@ -0,0 +1,248 @@
...
+##################
+function default_diagnostics(::RRTMGPI.AbstractRRTMGPMode)
+    allsky_diagnostics = ["rsd", "rsu", "rld", "rlu"]
+
+    return [daily_averages(allsky_diagnostics...)...]
+end
+
+
+function default_diagnostics(::RRTMGPI.AllSkyRadiationWithClearSkyDiagnostics)
+    allsky_diagnostics = ["rsdcs", "rsucs", "rldcs", "rlucs"]
```
szy21, line 245:
```suggestion
    clearsky_diagnostics = ["rsdcs", "rsucs", "rldcs", "rlucs"]
```

src/diagnostics/default_diagnostics.jl
```diff
@@ -0,0 +1,248 @@
...
+
+    return [daily_averages(allsky_diagnostics...)...]
```
szy21, line 247:
```suggestion
    return [daily_averages(clearsky_diagnostics...)...]
```

src/diagnostics/default_diagnostics.jl
```diff
@@ -0,0 +1,248 @@
...
+hourly_average(short_names; output_writer = HDF5Writer()) =
+    hourly_averages(short_names; output_writer)[1]
+
+# Include all the subdefaults
+
+########
+# Core #
+########
+function core_default_diagnostics()
+    core_diagnostics = ["ts", "thetaa", "rhoa", "ua", "va", "wa"]
```
szy21, line 195:
```suggestion
    core_diagnostics = ["ts", "ta", "thetaa", "pfull", "rhoa", "ua", "va", "wa"]
```

src/diagnostics/core_diagnostics.jl
```diff
@@ -0,0 +1,475 @@
...
+function compute_hus!(
+    out,
+    state,
+    cache,
+    time,
+    moisture_model::T,
+) where {T <: Union{EquilMoistModel, NonEquilMoistModel}}
+    thermo_params = CAP.thermodynamics_params(cache.params)
+    if isnothing(out)
+        return TD.total_specific_humidity.(thermo_params, cache.ᶜts)
```
szy21, line 250:
```suggestion
        return state.c.ρq_tot ./ state.c.ρ
```

src/diagnostics/core_diagnostics.jl
```diff
@@ -0,0 +1,475 @@
...
+    else
+        out .= TD.total_specific_humidity.(thermo_params, cache.ᶜts)
```
szy21, line 252:
```suggestion
        out .= state.c.ρq_tot ./ state.c.ρ
```
(This is not the same as the current diagnostics, but is better. If you want the diagnostics to be the same as the current one you can leave this unchanged - I can change it in a future PR.)

szy21 (approved):
Thank you Gabriele! Just a few more minor comments.

Sbozzolo:
> Thank you Gabriele! Just a few more minor comments.

All taken care of!

charleskawczynski:
Yep, I’ll take a look once the flame graph is generated!

src/ClimaAtmos.jl
```diff
@@ -110,6 +110,8 @@ include(joinpath("prognostic_equations", "limited_tendencies.jl"))
 
 include(joinpath("callbacks", "callbacks.jl"))
 
+include(joinpath("diagnostics", "Diagnostics.jl"))
+
```
charleskawczynski, line 114:
```suggestion
import .Diagnostics as CAD
```
I might suggest putting this import here

src/diagnostics/diagnostic.jl
```diff
@@ -0,0 +1,734 @@
...
+    identity_ft = convert(float_type, identity)
+    fill!(parent(diag_accumulator), identity_ft)
+end
+
+# When the reduction is nothing, we do not need to accumulate anything
+accumulate!(_, _, reduction_time_func::Nothing) = nothing
+
+# When we have a reduction, apply it between the accumulated value one
+function accumulate!(diag_accumulator, diag_storage, reduction_time_func)
+    diag_accumulator .= reduction_time_func.(diag_accumulator, diag_storage)
```
charleskawczynski, line 640:
```suggestion
    diag_accumulator .= reduction_time_func.(diag_accumulator, diag_storage)
    return nothing
```

charleskawczynski (approved):
I made one suggestion that might fix one of the red boxes. Inference doesn't look too bad. For profiling sake, it might be nice if we changed some of those anonymous functions to named functions (it's helpful to more quickly identify which flames are red). Overall I'm fine with merging and we can always try to fix the remaining inference failures later.

Screen Shot 2023-09-22 at 4 23 38 PM

PR #2065, opened 2023-09-06
Test that set_precomputed_quantities is idempotent

charleskawczynski:
This PR tests the hypothesis that `set_precomputed_quantities` is idempotent.


charleskawczynski:
Looks like it was idempotent. We should probably add a test, but closing for now.

PR #2069, opened 2023-09-06
Adds env var for job name per buildkite step

akshaysridhar:
For each `longrun` buildkite label, this PR adds a `JOB_NAME` environment variable to the pipeline, which then assigns the yaml file,  data / output directories, figure directories, artifact paths. This replaces the current system where we have repeated references to the same verbose file / job name.


nefrathenrici (approved):
Thank you!

PR #2070, opened 2023-09-06
Remove perf_mode

Sbozzolo:
This commit removes `perf_mode`, as requested by several `TODO`s sprinkled around the code. It seems that currently `perf_mode` does nothing.

### DIFF
--- a/src/initial_conditions/atmos_state.jl
+++ b/src/initial_conditions/atmos_state.jl
@@ -59,7 +59,7 @@ grid_scale_center_variables(ls, atmos_model) = (;
     uₕ = C12(ls.velocity, ls.geometry),
     energy_variables(ls)...,
     moisture_variables(ls, atmos_model.moisture_model)...,
-    precip_variables(ls, atmos_model.precip_model, atmos_model.perf_mode)...,
+    precip_variables(ls, atmos_model.precip_model)...,
 )
 
 energy_variables(ls) = (;
@@ -116,15 +116,12 @@ moisture_variables(ls, ::NonEquilMoistModel) = (;
     ρq_ice = ls.ρ * TD.ice_specific_humidity(ls.thermo_params, ls.thermo_state),
 )
 
-# TODO: Remove perf_mode. Currently, adding tracers hurts performance.
-precip_variables(ls, ::NoPrecipitation, ::PerfStandard) = (;)
-precip_variables(ls, ::Microphysics0Moment, ::PerfStandard) = (;)
-precip_variables(ls, ::Microphysics1Moment, ::PerfStandard) = (;
+precip_variables(ls, ::NoPrecipitation) = (;)
+precip_variables(ls, ::Microphysics0Moment) = (;)
+precip_variables(ls, ::Microphysics1Moment) = (;
     ρq_rai = ls.ρ * ls.precip_state.q_rai,
     ρq_sno = ls.ρ * ls.precip_state.q_sno,
 )
-precip_variables(ls, _, ::PerfExperimental) =
-    (; ρq_rai = zero(eltype(ls)), ρq_sno = zero(eltype(ls)))
 
 # We can use paper-based cases for LES type configurations (no TKE)
 # or SGS type configurations (initial TKE needed), so we do not need to assert
--- a/config/default_configs/default_config.yml
+++ b/config/default_configs/default_config.yml
@@ -187,9 +187,6 @@ precip_model:
 cloud_model:
   help: "Cloud model [`grid_scale`, `quadrature` (default)]"
   value: "quadrature"
-perf_mode:
-  help: "A flag for analyzing performance [`PerfStandard` (default), `PerfExperimental`]"
-  value: "PerfStandard"
 perf_summary:
   help: "Flag for collecting performance summary information"
   value: false
--- a/config/perf_configs/flame_perf_target_tracers.yml
+++ b/config/perf_configs/flame_perf_target_tracers.yml
@@ -8,10 +8,9 @@ vert_diff: "true"
 approximate_linear_solve_iters: 2
 h_elem: 12
 forcing: "held_suarez"
-precip_model: "0M"
+precip_model: "1M"
 dt_save_to_sol: "Inf"
 rad: "allskywithclear"
 warn_allocations_diagnostics: true
 log_progress: false
-perf_mode: "PerfExperimental"
 implicit_diffusion: false
--- a/perf/flame.jl
+++ b/perf/flame.jl
@@ -38,7 +38,7 @@ ProfileCanvas.html_file(joinpath(output_dir, "flame.html"), results)
 
 allocs_limit = Dict()
 allocs_limit["flame_perf_target"] = 1_160_723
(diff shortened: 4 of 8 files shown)


charleskawczynski:
Yeah, I'm not proud of the name or what it did. I added this.

For lack of a better name, and way to hook this in, `perf_mode` basically turns on a bunch of tracers in order to test that some of the tendencies for those tracers are well inferred. During this time, there were a lot of inference issues, and I thought that I may need to multipurpose this flag by turning on a bunch of features and ensuring that we keep the number of JET failures limited to some fixed number that we could drive down. I never ended up multi-purposing it though-- it turned out to be a pretty isolated edge case.

charleskawczynski:
All that said, I do think we need to exercise this code and test that it's well inferred. Now that we can create an integrator in a few lines, it is practical that we could add it as a unit test

charleskawczynski:
In particular, [this loop](https://github.com/CliMA/ClimaAtmos.jl/blob/7b777bc61eb7cd278e5e3b6916d5299d64997a39/src/prognostic_equations/implicit/schur_complement_W.jl#L345-L351) was poorly inferred and was tanking performance.

charleskawczynski:
If there's some way that you can remove `perf_mode` and add a performance unit test using JET.jl's `@test_opt`, that would be great.

Sbozzolo:
> Yeah, I'm not proud of the name or what it did, but I added this code.
> 
> For lack of a better name, and way to hook this in, `perf_mode` basically turns on a bunch of tracers in order to test that some of the tendencies for those tracers are well inferred. During this time, there were a lot of inference issues, and I thought that I may need to multipurpose this flag by turning on a bunch of features and ensuring that we keep the number of JET failures limited to some fixed number that we could drive down. I never ended up multi-purposing it though-- it turned out to be a pretty isolated edge case.

Is it still the case in the current code base? I don't see how this is done. There are not many references to `perf_mode` in the entire repository.

charleskawczynski:
> Is it still the case in the current code base? If yes, I don't see how this is done. There are not many references to perf_mode in the entire repository.

Yes, it's exercised in this [flame graph job](https://github.com/CliMA/ClimaAtmos.jl/blob/b0d14a2d7b12e4cb24b9b3a057466da92a69cd6a/.buildkite/pipeline.yml#L932-L938), which is turned on in the yaml included [here](https://github.com/CliMA/ClimaAtmos.jl/blob/b0d14a2d7b12e4cb24b9b3a057466da92a69cd6a/config/perf_configs/flame/perf_target_tracers.yml)

Sbozzolo:
> 

How does it work? Looking at the diff, I cannot find code that would be executed differently if you set `perf_mode` to `Standard` or `Experimental`.

However, when I look at the flame graphs, I do see a difference.

src/initial_conditions/atmos_state.jl
```diff
@@ -102,12 +102,9 @@ moisture_variables(ls, ::NonEquilMoistModel) = (;
     ρq_ice = ls.ρ * TD.ice_specific_humidity(ls.thermo_params, ls.thermo_state),
 )
 
-# TODO: Remove perf_mode. Currently, adding tracers hurts performance.
-precip_variables(ls, ::NoPrecipitation, ::PerfStandard) = (;)
-precip_variables(ls, ::Microphysics0Moment, ::PerfStandard) = (;)
-precip_variables(ls, ::Microphysics1Moment, ::PerfStandard) =
-    (; ρq_rai = zero(eltype(ls)), ρq_sno = zero(eltype(ls)))
```
charleskawczynski, line 109: @Sbozzolo, this deleted function is where the tracers are added. It's difficult to see in the diff, but in the main branch it looks like

```julia
precip_variables(ls, _, ::PerfExperimental) =
    (; ρq_rai = zero(eltype(ls)), ρq_sno = zero(eltype(ls)))
```

Sbozzolo: Aren't the two identical?

Sbozzolo:
This is a possible way to free up the code base from all the `Perf` stuff while having a test with `0Microphysics0Moment` that adds `ρq_rai` and `ρq_sno` to the state variables.

The solution is not going to be clean, but in my opinion it is worth it because it removes unnecessary code and complexity in `ClimaAtmos`.

Currently, the only place where `PerfExperimental` is used in a a test `flame_perf_target_tracers`. This test uses the default performance configuration which has `0Microphysics0Moment` precipitation model and adds two variables `ρq_rai` and `ρq_sno`, which normally does not have them. Presumably, these variables are then evolved.

My proposal is to to perform text substitution in the source code to add those two variables for that specific test. There are many ways to do this with shell commands. A very simple one would be
```
sed -i 's/precip_variables(ls, ::Microphysics0Moment, ::PerfStandard) = (;)/precip_variables(ls, ::Microphysics0Moment, ::PerfStandard) = (; ρq_rai = zero(eltype(ls)), ρq_sno = zero(eltype(ls)))/' src/initial_conditions/atmos_state.jl
```
This is not the most robust because it matches the line exactly, but it might do the job.

An added complication is that since we change the source code, we need to recompile (without affecting the other runs).

So, at the end, the section in the `pipeline.yml` would something look like:
```bash
# Monkey-patch ClimaAtmos to add tracers to a Microphysics0Moment simulation
`sed -i 's/precip_variables(ls, ::Microphysics0Moment, ::PerfStandard) = (;)/precip_variables(ls, ::Microphysics0Moment, ::PerfStandard) = (; ρq_rai = zero(eltype(ls)), ρq_sno = zero(eltype(ls)))/' src/initial_conditions/atmos_state.jl`
# Create a new temporary depot so that we can precompile the new code
export JULIA_DEPOT_PATH="$(mktemp -d)"

julia --project=perf -e 'using Pkg; Pkg.instantiate(;verbose=true)'
julia --project=perf -e 'using Pkg; Pkg.precompile()'
julia --project=perf -e 'using Pkg; Pkg.status()'

julia --color=yes --project=perf perf/flame.jl --config_file  $PERF_CONFIG_PATH/flame/perf_target_tracers.yml
```

This is certainly not the most beautiful solution, but I think it is still worth it. `perf_mode` is already a fancy way to patch the code, with this new approach at least we clean up the main code.

src/initial_conditions/atmos_state.jl
charleskawczynski: No, the second argument is more generic, so it catches basically any precip model (even if it's nothing).

charleskawczynski:
I really don't like this solution. I prefer we stick to something simpler.

Sbozzolo:
Another possibility might be to redefine the method `precip_variables(ls, ::Microphysics0Moment)` in the test. Something like (pseudo code)
```julia
if parsed_args["perf_mode"] == "experimental"
    ClimaAtmos.eval(:(precip....))
```
The drawback is that this would still keep the `perf_mode` flag (or equivalent) around.

Sbozzolo:
Now that we have the 1M microphysics, the two tracers that were artificially added are now organically there.

charleskawczynski (approved):
> Now that we have the 1M microphysics, the two tracers that were artificially added are now organically there.

Excellent, thank you!

PR #2078, opened 2023-09-08
Treat idealized_{insolation,clouds} as idealized_h20

Sbozzolo:
This commit puts `idealized_insoluation` and `idealized_clouds` on the same footing as `idealized_h20`. As a result, it also removes the need to use `parsed_args` in the computation of the cache.


szy21 (approved):
Thanks!

PR #2079, opened 2023-09-08
Try to fix FieldVector inference

charleskawczynski:
This PR is an attempt to fix the very bad remaining inference failure in the TC step! function, by calling a different FieldVector constructor to avoid the inference failure in (what I assume is `promote_type(map(RecursiveArrayTools.recursive_bottom_eltype, values)...)`).

Inference currently looks like this in the EDMF flame graph:
Screen Shot 2023-09-08 at 11 09 07 AM

Let's see if this PR has any impact.

cc @LenkaNovak, @akshaysridhar 

This is why I suggested adding JET tests in https://github.com/CliMA/ClimaCore.jl/pull/1441, it will work, but it will be impractically slow. This flame graph is for a column, I imagine the sphere would be much worse.


szy21:
We will delete this folder soon (hopefully next week). I wonder if we need to fix this, or it it a more general issue?

charleskawczynski:
> We will delete this folder soon (hopefully next week). I wonder if we need to fix this, or it it a more general issue?

Sounds good to me, this is more to see if these changes will fix the issue in light of https://github.com/CliMA/ClimaCore.jl/pull/1441

PR #2084, opened 2023-09-11
Improve docs, add convenience func for setting config file

nefrathenrici:
Content
- Adds better docs on configuring the model
- Adds `run_model_from_file`, a convenience function for running the model from a file

### DIFF
--- a/docs/src/config.md
+++ b/docs/src/config.md
@@ -1,3 +1,27 @@
+# Running the model with custom configurations
+There are two main ways of setting the model configuration: with a file or a dictionary.
+
+### Runing from a file
+Given a configuration file, you can run the model from the terminal using:
+```
+julia --project=examples examples/driver.jl --config_file <filepath>`
+```
+If you are running the model from the REPL and want to rerun the model 
+with a new configuration file without recompiling,
+a convenience function `run_model_from_file(config_file)` is defined. 
+
+### Running from a dictionary
+You can also set the config interactively, but this is more involved:
+```julia
+config_dict = Dict(
+"FLOAT_TYPE" => Float64,
+"z_max" => 3000.0,
+"z_elem" => 60,
+"dt" => "20secs",
+)
+config = ClimaAtmos.AtmosConfig(; config_dict)
+include("examples/hybrid/driver.jl")
+```
 
 # Creating custom configurations
 To create a custom configuration, first make a .yml file.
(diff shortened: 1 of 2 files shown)


Sbozzolo:
Under which conditions does this avoid recompilation?

nefrathenrici:
> Under which conditions does this avoid recompilation?

@Sbozzolo Using [Revise](https://timholy.github.io/Revise.jl/stable/) can help avoid restarting the REPL after making code changes, since it does some magic to recompile changed code (except types, to my knowledge). This function just makes it easier to change the model config when using Revise, before it was a bit clunky.

src/solver/yaml_helper.jl
```diff
@@ -162,3 +162,23 @@ function job_id_from_config(config, defaults = default_config_dict())
...
+    config = CA.AtmosConfig(; config_dict)
+    driver_path = joinpath(
+        dirname(@__FILE__),
+        "..",
+        "..",
+        "examples",
+        "hybrid",
+        "driver.jl",
+    )
+    include(driver_path)
```
charleskawczynski, line 183: I believe `include` is only allowed in global scope (or, at least it's not allowed inside functions)

nefrathenrici: Yeah, it's just running the driver file in global scope, I need to rethink this. Thanks for catching that!

Sbozzolo:
This change seems to imply that `examples/hybrid/driver.jl` is the correct way to use `ClimaAtmos`. However, the driver does a lot of stuff that is unrelated to actually running the code. Moreover, working with `ARGS` feels like a workaround. Might as well have a second constructor for `AtmosConfig` that directly takes a file path and calls the normal constructor with `parsed_args=Dict("config_file" => PATH)`.

nefrathenrici:
> This change seems to imply that `examples/hybrid/driver.jl` is the correct way to use `ClimaAtmos`. However, the driver does a lot of stuff that is unrelated to actually running the code. Moreover, working with `ARGS` feels like a workaround. Maybe having a second constructor for `AtmosConfig` that directly takes a file path and calls the normal constructor with `parsed_args=Dict("config_file" => PATH)`?

Thanks @Sbozzolo, that does seems like a better way - something like this:
```
function run_model_from_file(config_file)
    config = CA.AtmosConfig(; parsed_args = ["--config_file", config_file])
    return CA.solve_atmos!(integrator)
end
```
I have two hesitations with this: we would need to set the random seed, and people may want to full driver behavior rather than just running the solver - I will have to ask around.

Sbozzolo:
I think that a function that calls the driver is not something that should be part of `/src`. At this stage, the driver is fundamentally a development tool, that even has ties to CI and Caltech's cluster. If we think of `ClimaAtmos` as a library (as the GitHub description says), then the driver would be a user of such library. 

The code that you sketched does not depend on the driver, so it would be fine (and I would use it). But if people want to have a one-liner to run the driver, I personally think that we should instead show a function that those users can add to their `startup.jl`. 

(Also, `parsed_args` in the constructor of `ClimaAtmos` should be a dictionary.)

nefrathenrici:
bors try

Sbozzolo:
Is this still relevant or can we close it?

PR #2085, opened 2023-09-13
add momentum diffusion for diagnostic edmf

szy21:
## Content

----


perf/jet_report_nfailures.jl
```diff
@@ -20,7 +20,7 @@ t = integrator.t;
 Yₜ = similar(Y);
 ref_Y = similar(Y);
 #! format: off
-n["step!"]                                       = @n_failures SciMLBase.step!(integrator);
+#n["step!"]                                       = @n_failures SciMLBase.step!(integrator);
```
szy21, line 23: revert this change before merging

tapios (approved):
It looks good to me. There are several code constructs I do not fully understand. It would be good for @trontrytel or @simonbyrne to check.

src/prognostic_equations/edmfx_sgs_flux.jl
```diff
@@ -194,15 +194,42 @@ function edmfx_sgs_flux_tendency!(
...
+                Y.c.ρ[colidx],
+            ),
+        )
         ᶜdivᵥ_uₕ = Operators.DivergenceF2C(
             top = Operators.SetValue(C3(FT(0)) ⊗ C12(FT(0), FT(0))),
             bottom = Operators.SetValue(sfc_conditions.ρ_flux_uₕ[colidx]),
         )
         @. Yₜ.c.uₕ[colidx] -=
-            ᶜdivᵥ_uₕ(-(ᶠρaK_u[colidx] * ᶠgradᵥ(Y.c.uₕ[colidx]))) / Y.c.ρ[colidx]
+            ᶜdivᵥ_uₕ(-(FT(0) * ᶠgradᵥ(Y.c.uₕ[colidx]))) / Y.c.ρ[colidx]
```
simonbyrne, line 232: I think we should be able to combine these into one operation.

src/prognostic_equations/edmfx_sgs_flux.jl
```diff
@@ -194,15 +194,42 @@ function edmfx_sgs_flux_tendency!(
             )
         end
 
+        # mass flux
+        ᶠinterp_u₃ = Operators.InterpolateC2F(
+            top = Operators.SetValue(C3(FT(0))),
+            bottom = Operators.SetValue(C3(FT(0))),
```
simonbyrne, line 200: not sure this is right for topography?

simonbyrne: Actually i think this will get ignored.

src/utils/utilities.jl
```diff
@@ -94,6 +94,28 @@ function compute_strain_rate!(ϵ::Fields.Field, u::Fields.Field)
...
+"""
+    compute_strain_rate_face!(ϵ::Field, u::Field)
+
+Compute the strain_rate at cell faces, storing in `ϵ` from
+velocity at cell centers.
+"""
+function compute_strain_rate_face!(ϵ::Fields.Field, u::Fields.Field)
+    @assert eltype(u) <: C123
+    ∇ᵥuvw_boundary =
+        Geometry.outer(Geometry.WVector(0), Geometry.UVWVector(0, 0, 0))
```
trontrytel, line 106: Will it always be zero BC?

szy21: I think so. What we have here needs to be consistent with surface flux, which is applied elsewhere, so it should be zero here. If Simon combines the two operations he mentioned above, then whatever we have here will be ignored, as the boundary condition of the `ᶜdivᵥ` operator will override this.

src/prognostic_equations/edmfx_sgs_flux.jl
```diff
@@ -194,15 +194,42 @@ function edmfx_sgs_flux_tendency!(
...
+        @. Yₜ.c.uₕ[colidx] -= C12(
+            ᶜdivᵥ_uₕ(-(2 * ᶠρaK_u[colidx] * ᶠstrain_rate[colidx])) /
+            Y.c.ρ[colidx],
+        )
+        @. Yₜ.f.u₃[colidx] -= ᶠinterp_u₃(
+            C3(
+                ᶜdivᵥ(-(2 * ᶠρaK_u[colidx] * ᶠstrain_rate[colidx])) /
+                Y.c.ρ[colidx],
+            ),
+        )
```
simonbyrne, line 232: @szy21 does this make sense? I think it should work.

We may want the same for the `Yₜ.f.u₃` term

perf/jet_report_nfailures.jl
szy21, line 23:
```suggestion
n["step!"]                                       = @n_failures SciMLBase.step!(integrator);
```

PR #2087, opened 2023-09-13
Remove hardcoded parameters, minor improvements for parameters-as-functions

nefrathenrici:
Content
- Removes `entr_coeff` `detr_coeff` and `entr_tau` from the config, makes them parameters
- - Update `edmfx_bomex_box_v2.toml` to include the `detr_coeff` change
- Adds  parameters for previously hardcoded values in the held suarez `forcing_tendency`, update the parameter struct accordingly
- Removes `SpongeParameters` - I think we can keep these in the top-level struct
- Improves the parameter functions, auto-generate them using `@eval`
- Adds some tests to ensure parameters are properly read in. I introduced a bug in the atmos tomls where different parameters had the same alias, leading to parameters not being properly set. This bug has been fixed by [ClimaParameters #136](https://github.com/CliMA/CLIMAParameters.jl/pull/136), which also includes an error check for unique aliases.

Note: This does not get rid of all hardcoded parameters, only the high-priority ones.

### DIFF
--- a/src/parameters/Parameters.jl
+++ b/src/parameters/Parameters.jl
@@ -10,34 +10,30 @@ const ACAP = AbstractClimaAtmosParameters
 
 Base.broadcastable(param_set::ACAP) = tuple(param_set)
 
-Base.@kwdef struct ClimaAtmosParameters{FT, TP, RP, IP, MPP, SFP, TCP, SP} <:
-                   ACAP
-    Omega::FT
-    f_plane_coriolis_frequency::FT
-    planet_radius::FT
-    astro_unit::FT
+Base.@kwdef struct ClimaAtmosParameters{FT, TP, RP, IP, MPP, SFP, TCP} <: ACAP
     thermodynamics_params::TP
     rrtmgp_params::RP
     insolation_params::IP
     microphysics_params::MPP
-    surfacefluxes_params::SFP
+    surface_fluxes_params::SFP
     turbconv_params::TCP
-    sponge_params::SP
+    Omega::FT
+    f_plane_coriolis_frequency::FT
+    planet_radius::FT
+    astro_unit::FT
     entr_tau::FT
     entr_coeff::FT
     detr_coeff::FT
-    # TODO: Figure out a better place for these held-suarez parameters
+    C_E::FT
+    # Held Suarez
     ΔT_y_dry::FT
     ΔT_y_wet::FT
-    C_E::FT
-end
-
-"""
-    SpongeParameters{FT}
-
-Parameters for both the viscous and rayleigh sponge.
-"""
-Base.@kwdef struct SpongeParameters{FT}
+    σ_b::FT
+    T_equator_dry::FT
+    T_equator_wet::FT
+    T_min_hs::FT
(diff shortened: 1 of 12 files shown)


szy21 (approved):
Looks good, thanks! In general, how do we determine if some parameters (e.g. sponge) should be in the top-level struct?

nefrathenrici:
> Looks good, thanks! In general, how do we determine if some parameters (e.g. sponge) should be in the top-level struct?

For now, we can just keep adding parameters to the top level because there aren't too many. If we keep adding parameters, we should probably find better ways to break them up.

trontrytel (approved):

src/parameters/Parameters.jl
```diff
@@ -85,4 +59,11 @@ detr_coeff(ps::ACAP) = ps.detr_coeff
 day(ps::ACAP) = IP.day(insolation_params(ps))
 tot_solar_irrad(ps::ACAP) = IP.tot_solar_irrad(insolation_params(ps))
 
+# Define parameters as functions
+# Place derived parameters in exceptions vector
+exceptions = []
+for var in filter(x -> !(x in exceptions), fieldnames(ClimaAtmosParameters))
```
charleskawczynski, line 65:
```suggestion
for var in fieldnames(ClimaAtmosParameters)
```
I'd recommend removing the exceptions part

charleskawczynski: (sorry I'm late to review this)

nefrathenrici: Oops, I can add this in #2112 or as a standalone PR

src/parameters/create_parameters.jl
```diff
@@ -80,13 +79,13 @@ function create_parameter_set(config::AtmosConfig)
...
-            sponge_params,
         ),
     )
-    # TODO: Add parameter logging option from config
-    # logfilepath = joinpath(@__DIR__, "logfilepath_$FT.toml")
-    # CP.log_parameter_information(toml_dict, logfilepath)
+    if parsed_args["log_params"]
+        logfilepath = joinpath(@__DIR__, "$(parsed_args["job_id"])_$FT.toml")
+        CP.log_parameter_information(toml_dict, logfilepath)
+    end
```
Sbozzolo, line 89: From the help of `@__DIR__`: `  Expand to a string with the absolute path to the directory of the file containing the macrocall`, so I think that writes in the folder where this file particular lives, which might be something like `.julia/packages/ClimaAtmos/packages/XASy1234/src/`. 

Also, it would be really amazing if we didn't access `parsed_args` in this function, so that we can decouple the generation of the parameters with the parsed arguments.

Sbozzolo: The function to log the parameters could be moved to the `AtmosConfig` constructor so that the only reference to `parsed_args` in this function is for the precipitation.

nefrathenrici: Thanks for catching this! I can move it. I will try to get rid of the tau precip override next week, so if everything goes well the `parsed_args` will be separated out.

PR #2090, opened 2023-09-14
switch to using NVTX annotate instead of range

simonbyrne:
## Purpose 
Switches to `NVTX.@annotate`, which applies at the definition, and makes sure we don't miss any precomputed quantities calls.

## Content

----


charleskawczynski (approved):

simonbyrne:
A bit clearer what is going on now:
Screenshot 2023-09-14 at 10 00 01 PM

PR #2091, opened 2023-09-15
Use weak divergence in scalar advection

simonbyrne:
## Purpose 
If we use weak divergence for advection, then no mass is transported between elements until the DSS operation is performed. This means that we can guarantee that the total mass in the element is non-negative (and so don't need to borrow horizontally).

## Content

----


szy21:
bors try

src/prognostic_equations/advection.jl
```diff
@@ -59,15 +59,23 @@ function horizontal_tracer_advection_tendency!(Yₜ, Y, p, t)
...
     end
 
+    horizontal_tracer_advection_tendency_turbconv!(
+        Yₜ,
+        Y,
+        p,
+        t,
+        p.atmos.turbconv_model,
+    )
+
```
simonbyrne, line 72:
```suggestion
```

PR #2092, opened 2023-09-15
remove DSS on remaining tendency

simonbyrne:
## Purpose 
This shouldn't be needed, as we do it in the timestepper.

## Content

----


szy21:
This changes MSE. Is it as expected?

simonbyrne:
Should only be at machine error

szy21 (approved):
The results look good to me.

PR #2112, opened 2023-09-18
Simplify configurations for performance jobs

nefrathenrici:
Content
- Splits up `AtmosConfig` constructor, for a clearer flow and dispatch on `Dicts`, `Strings` (filepath) and `Nothing`. This allows for circumventing the CLI and adds another entry point in the config generation flow. This is technically a breaking change, and might need to be changed before merging. 
- Add `perf/common.jl` with two new functions to easily generate configurations for performance jobs. `AtmosCoveragePerfConfig` takes in a config dictionary, overrides the default performance configuration, and passes it to `AtmosConfig`. `TargetJobConfig` takes in a job id, and constructs its corresponding `AtmosConfig` 

- Rework `flame.jl`, `jet_test_nfailures.jl`, and `benchmark.jl` to maintain continuity for the perf tests. Apply this change for the relevant buildkite jobs in `pipeline.yml`.
- Add tests for `AtmosCoveragePerfConfig`


Sbozzolo:
> AtmosCoveragePerfConfig is really only needed for performance analysis. I think we could move it outside of src and into, e.g., perf/common.jl.

I agree: `AtmosCoveragePerfConfig` is nothing fundamental, so I agree that is should not be part of the core library.

> * Run a target job (presumably, someone reports that a particular job is slow, and it's helpful to be able to somehow target that one job with `job_id`)

The current behavior is to merge `default_perf` with `target_job`, which makes profiling a particular job next to impossible. I think that when we want to run a target job, we should not add anything extra.

nefrathenrici:
> The current behavior is to merge `default_perf` with `target_job`, which makes profiling a particular job next to impossible. I think that when we want to run a target job, we should not add anything extra.
There is actually already a function `config_from_target_job` which can be passed in to `AtmosConfig`.
```
config_dict = CA.config_from_target_job("edmf_trmm")
config = CA.AtmosConfig(; config_dict)
```
cc @Sbozzolo

Sbozzolo:
See also: https://github.com/CliMA/ClimaAtmos.jl/pull/2157

.buildkite/pipeline.yml
```diff
@@ -1024,23 +1024,23 @@ steps:
...
+          flame_perf_gw
         artifact_paths: "flame_perf_gw/*"
         agents:
           slurm_mem: 20GB
 
       - label: ":fire: Flame graph: perf target (diagnostics)"
         command: >
           julia --color=yes --project=perf perf/flame.jl
-          --config_file $PERF_CONFIG_PATH/flame/diagnostics.yml
+          flame_perf_diagnostics
```
charleskawczynski, line 1043: So is this just using ARGS now?

charleskawczynski (approved):
I think this generally looks better

.buildkite/pipeline.yml
nefrathenrici: Yes, but we could use argparse. Do you think that would be clearer?

charleskawczynski: The only benefit I see is that it's keyword based rather than positional. It's not a big issue if there's a small number of arguments, but 🤷🏻

test/config.jl
```diff
@@ -0,0 +1,39 @@
...
+    @test config.parsed_args["dt"] == "1secs"
+    # Test defaults not overridden by `default_perf`
+    @test config.parsed_args["y_elem"] == 6
+
+    # Test with `target_job`
+    config = AtmosCoveragePerfConfig(;
+        config_dict = CA.config_from_target_job("sphere_baroclinic_wave_rhoe"),
+    )
+    # Target job config overridden by `default_perf`
+    @test config.parsed_args["dt"] == "1secs"
```
Sbozzolo, line 19: Shouldn't `sphere_baroclinic_wave_rhoe` have the precedence over the default (as per docstring), so that `dt != "1secs"`?

perf/benchmark.jl
```diff
@@ -1,7 +1,16 @@
 import Random
 Random.seed!(1234)
 import ClimaAtmos as CA
-config = CA.AtmosCoveragePerfConfig()
+
+include("common.jl")
+
+target_job = ARGS[1]
+job_id = get(ARGS, 2, target_job)
```
Sbozzolo, line 8: If we are using `ARGS`, I think that we should check that the correct number of arguments is passed and provide an error message explaning the various arguments. Something like:
```julia
length(ARGS) < 2 && error("Usage: benchmark.jl <target_job> <job_id>
```

test/config.jl
nefrathenrici: Sorry, the docstring is incorrect - the default perf config needs to take priority to keep existing performance tests consistent.

Priority:
1. Default perf configuration (to increase coverage)
2. Configuration from the given config file/dict
3. Default configuration

perf/benchmark.jl
nefrathenrici: That's a good idea - I think we should just use argparse to ensure CLI input is correct since this will reduce repeated code.

test/config.jl
Sbozzolo: If the target_job is overridden, what is its use? 
```
    config = AtmosCoveragePerfConfig(;
        config_dict = CA.config_from_target_job("sphere_baroclinic_wave_rhoe"),
    )
```
Is for a moist configuration with vertical diffiusion, Held Suarzed forcing, precipitation, radiation.

The only meaningful configuration that is left from `sphere_baroclinic_wave_rhoe` is that the initial condition is `DryBaroclinicWave`.

perf/benchmark.jl
Sbozzolo: I think that this is fine. If you are using only two arguments, there's not much confusion and make things more straightforward. You also remove the dependency of this script on `ArgParse.jl` which hasn't been updated in more than 2 years.

test/config.jl
nefrathenrici: If a user wants to test the performance of a specific job, they can use `TargetJobConfig`.
`AtmosCoveragePerfConfig` is essentially only used to create configurations for the existing performance tests.

perf/benchmark.jl
nefrathenrici: That sounds good to me. We could even get rid of `ArgParse` in the driver if we just take one argument.

test/config.jl
Sbozzolo: Given that, what I would recommend to reframe `AtmosCoveragePerfConfig` as being: simulate everything, optionally add more configuration using the `config_dict` (even if it is very unusual that a user provided option is ignored in favor of a default one). Then, we remove all the references to target_jobs in `AtmosCoveragePerfConfig` and only add those options that we care about for our tests. 

My point is that the notion of target_job attached in any way to an `AtmosCoveragePerfConfig` that enables all the physics is really confusing and counter intuitive. I think that my proposed change would make the intent clear: instead of "targeting a edmf job" we "add emdf" to our configuration.

nefrathenrici: I agree that the current way we use `AtmosCoveragePerfConfig` is not intuitive to users, but it is needed to maintain our tests. This should be changed, but I think that this would need to be part of a larger discussion of our performance testing practices, since it would break continuity on said tests.

Sbozzolo: Yes, what I am saying is that we maintain our tests by explicitly adding those options that we need.

For example, if we had a `sphere_baroclinic_rhoe` test, we would just need to add `initial_conditions = DryBaroclinicWave`.

perf/flame.jl
```diff
@@ -1,21 +1,15 @@
...
-job_id = get(ARGS, 2, target_job)
-
-config_dict =
-    target_job != "default" ? CA.config_from_target_job(target_job) :
-    CA.default_config_dict()
-
-# Need to set internal job_id for diagnostics saved by model
-config_dict["job_id"] = job_id
-
+length(ARGS) != 1 && error("Usage: benchmark.jl <config_file>")
```
Sbozzolo, line 8:
```suggestion
length(ARGS) != 1 && error("Usage: flame.jl <config_file>")
```

perf/jet_test_nfailures.jl
```diff
@@ -1,16 +1,13 @@
...
 include("common.jl")
 
-length(ARGS) < 2 && error("Usage: benchmark.jl <target_job> <job_id>")
-target_job = ARGS[1]
-job_id = get(ARGS, 2, target_job)
-
-config_dict =
-    target_job != "default" ? CA.config_from_target_job(target_job) :
-    CA.default_config_dict()
+length(ARGS) != 1 && error("Usage: benchmark.jl <config_file>")
```
Sbozzolo, line 8:
```suggestion
length(ARGS) != 1 && error("Usage: jet_test_nfailures.jl <config_file>")
```
Or `basename(@__FILE__)`

perf/common.jl
```diff
@@ -12,14 +12,9 @@ The configuration precedence is as follows:
 There is an exception for the job `flame_perf_target_edmf`, which requires that
```
Sbozzolo, line 12: The docstring should be updated

nefrathenrici: I will clean this up, just wanted to run CI to ensure the config is being set correctly. Thanks for looking it over!

PR #2115, opened 2023-09-19
Avoid FieldVector outer constructor inference

charleskawczynski:
I realized that most of the allocations in the callbacks are actually coming from _inference_ in the FieldVector outer constructor:

Screen Shot 2023-09-19 at 11 08 19 AM

This PR should (hopefully) avoid this, by calling the outer constructor where the undertype is pre-specified.

### DIFF
--- a/perf/flame.jl
+++ b/perf/flame.jl
@@ -54,14 +54,14 @@ allocs = @allocated SciMLBase.step!(integrator)
 @info "`allocs ($job_id)`: $(allocs)"
 
 allocs_limit = Dict()
-allocs_limit["flame_perf_target"] = 4384
-allocs_limit["flame_perf_target_tracers"] = 204016
-allocs_limit["flame_perf_target_edmfx"] = 304064
-allocs_limit["flame_perf_target_diagnostic_edmfx"] = 685456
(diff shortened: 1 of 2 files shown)


charleskawczynski:
Also, I plan to update the flame allocation tests once we get numbers from CI.

charleskawczynski:
Allocations in the callbacks flame in this PR:

Screen Shot 2023-09-19 at 11 47 51 AM

simonbyrne:
should we just `@nospecialize` the constructor?

charleskawczynski:
> should we just `@nospecialize` the constructor?

Will that help? I think it's the recursive `promote_type` (`T = promote_type(map(RecursiveArrayTools.recursive_bottom_eltype, values)...,)`) that's type unstable. I think it's okay so long we avoid it in hot loops 🤷🏻

charleskawczynski:
> I realized that most of the allocations in the callbacks are actually coming from inference in the FieldVector outer constructor:

Correction: most of the allocations in the callbacks that `Profile.allocs` sees (see https://github.com/JuliaLang/julia/issues/50187)

PR #2121, opened 2023-09-20
change edmfx advection

### LINKED ISSUE #2106: Switch advection for edmfx from (rho a) * tracer to rho * (a * tracer)

trontrytel:
Closes #2106 

I wasn't sure if we also want to change the tke advection?


config/model_configs/edmfx_adv_test_box.yml
```diff
@@ -1,18 +1,21 @@
...
-dz_bottom: 30.0
 ode_algo: "SSP33ShuOsher"
+edmfx_upwinding: "first_order"
 config: "box"
-advection_test: true
-kappa_4: 1.0e8
-job_id: "edmfx_adv_test_box"
 moist: "equil"
+hyperdiff: "true"
+kappa_4: .0e8
```
szy21, line 10:
```suggestion
kappa_4: 1.0e8
```

src/prognostic_equations/advection.jl
```diff
@@ -139,24 +140,28 @@ NVTX.@annotate function explicit_vertical_advection_tendency!(Yₜ, Y, p, t)
 
         # TODO: Move this to implicit_vertical_advection_tendency!.
         for j in 1:n
-            @. ᶜ1[colidx] = one(Y.c.ρ[colidx])
+            @. ᶜtmp_scalar[colidx] =
```
szy21, line 143: I might call it something like ᶜa_scalar to make it more clear.

toml/edmfx_box_advection.toml
```diff
@@ -0,0 +1,15 @@
+[C_E]
+alias = "C_E"
+value = 0.044
+type = "float"
```
szy21, line 4: This is not needed as `vert_diff` is false in this test.

toml/edmfx_box_advection.toml
```diff
@@ -0,0 +1,15 @@
...
+
+[EDMF_surface_area]
+alias = "surface_area"
+value = 0
+type = "float"
+
+[EDMF_min_area]
+alias = "min_area"
+value = 1.0e-2
```
szy21, line 13: This seems pretty large. The result of the advection test looks a bit strange, not sure if it's related to this. We can talk more tomorrow.

config/model_configs/edmfx_adv_test_box.yml
trontrytel: Thank you!

trontrytel:
Here is another clue. I'm plotting the difference between the grid mean `ρq_tot` and the area weighted contributions from the drafts computed in two ways: Updraft contribution based on the state `Y` and based on the cache `p`. 

You can see that the cache based version has a big spike where we have the problematic secondary peak in `q_tot`.

Updraft computed from the state: (`Y.c.ρq_tot - Y.c.sgsʲs[1].ρaq_tot - p.ᶜρa⁰ * p.ᶜspecific⁰.q_tot`):

testing_from_Y

Updraft computed from the cache: (`Y.c.ρq_tot - Y.c.sgsʲs[1].ρa * p.ᶜspecificʲs[1].q_tot - p.ᶜρa⁰ * p.ᶜspecific⁰.q_tot`):

testing_from_p

trontrytel:
Here is how the weight `= sgs_weight_function(ρa / ρ, turbconv_model.a_half)` looks like for updraft and environment

weight

tapios:
> Here is another clue. I'm plotting the difference between the grid mean `ρq_tot` and the area weighted contributions from the drafts computed in two ways: Updraft contribution based on the state `Y` and based on the cache `p`.

Looks like state and cache are out of sync. May this get better with @charleskawczynski's refactoring of the precomputed quantities to postcomputed quantities?

tapios:
> > Looks like state and cache are out of sync. May this get better with @charleskawczynski's refactoring of the precomputed quantities to postcomputed quantities?
> 
> Maybe. It's more likely from that ρaq_tot is not equal to ρa * q_tot from the regularization I think.

Can we run this test with just the limiters on entrainment/detrainment I sketched, without any additional regularization?

trontrytel:
> Can we run this test with just the limiters on entrainment/detrainment I sketched, without any additional regularization?

This test is only with advection. There is no entrainment/detrainment here

tapios:
> > Can we run this test with just the limiters on entrainment/detrainment I sketched, without any additional regularization?
> 
> This test is only with advection. There is no entrainment/detrainment here

So entrainment=detrainment=0?

tapios:
> > So entrainment=detrainment=0?
> 
> Yes. The test was meant to look only at advection issues. No other tendencies are applied.
> 
> I think that the issue is in the regularization function. I'll try to track it today

Sounds good. Maybe the test is more meaningful with small but nonzero entrainment, detrainment, with the limiter functions on them, to ensure that area fraction remains bounded and updraft properties relax to grid means.

trontrytel:
Can we merge this change as it is? No matter what we do with the regularization, we still want to switch our advection. 

Merging it in would help me with testing for other cases. cc: @tapios @szy21

szy21:
Sounds good to me.

trontrytel:
Below are results with/without using grid mean density `Y.c.rho` when doing upwind reconstructions for edmfx variables. I'm running the test without any moisture, to highlight that the problem is not with the phase change. The last set of plots is the advection test for our target vertical resolution and using the `min_area = 1e3`. 

- Changing the density in the advection reconstructions doesn't change the results much.
- When switching to the target resolution and increasing the `min_area`, the problem is less severe. 

I'm going to merge those settings as the default for prognostic edmfx and move on to the next test case (GABLS).

Using grid mean density in upwind reconstructions:

final_profiles_zoom_in_upwind_recs

Using updraft density in upwind reconstructions:

final_profiles_zoom_in_no_upwind_recs

Target vertical grid:

final_profiles

PR #2125, opened 2023-09-20
add the filter for GABLS

trontrytel:
closing #2124


szy21 (approved):
Looks good to me. Just want to double check that this filter is approved by @tapios :) Should we start to keep track of the filters we have somewhere?

Btw when I tried a quick hack for setting e_int to be the same it didn't work, but maybe I did something wrong.

tapios (commented):
I do not particularly like this zeroing out approach, but I'm ok with having it if there is no more principled alternative. 

It should suffice to have entrainment go to zero as a -> 0. Did you try an approach that does that? In some experiments, I found that limiting entrainment by multiplying it with `2*harmonic_mean([max(a, 0); max(1-a, 0)]))` (while increasing entrainment overall) leads to bounded solutions for `a`. Could you try this here?

tapios:
> > I do not particularly like this zeroing out approach, but I'm ok with having it if there is no more principled alternative.
> > It should suffice to have entrainment go to zero as a -> 0. Did you try an approach that does that? In some experiments, I found that limiting entrainment by multiplying it with `2*harmonic_mean([max(a, 0); max(1-a, 0)]))` (while increasing entrainment overall) leads to bounded solutions for `a`. Could you try this here?
> 
> I will try it. But I think that here the issue is not that the area does not go to zero. But rather that the area is close to zero, but the updraft still has different properties than the grid mean

Is the issue the updraft buoyancy?

trontrytel:
> Is the issue the updraft buoyancy?

The way I understand it is that if a<<1 we are setting the updraft e_tot to be the same as the grid mean. But if the updraft vertical velocity is different that will make internal energy different and will cause things to diverge. I'm not sure what is causing the initial divergence. 

Here are the timeseries and final profiles from GABLS without the filter I'm trying to add. 

GABLS_contours
GABLS_final_profiles

szy21:
It's probably not related, I just would like to comment here so I won't forget: I think it would be better to change this [line](https://github.com/CliMA/ClimaAtmos.jl/blob/ec03d4e3f5021360ff591edde657e00bb921f638/src/prognostic_equations/edmfx_entr_detr.jl#L265) to `ᶠinterp(ᶜentrʲs.:($$j)[colidx]) * Y.f.u₃[colidx]`, so that we can make sure the tendency of u_3 at the bottom face is zero if both grid-mean and updraft u_3 is zero (as they should be, when there is no topography).

trontrytel:
In the end I'm:
- reducing the dt to 5s, 
- adding another entr/detr option: a constant coefficient but limited by the harmonics mean of (a, 1-a). This is so that we can easily test it for other cases.

This doesn't solve the problem. Our GABLS simulation continues to diverge. But maybe we can merge this setup in, while we try to understand how to make the scheme more stable. 

Do we want to keep the harmonics limiter as default?

With harmonics:

contours_w_harmonics

Without harmonics:

contours_wo_harmonics

szy21 (approved):

PR #2127, opened 2023-09-20
add a function to get draft area

trontrytel:


src/prognostic_equations/pressure_work.jl
```diff
@@ -8,7 +8,8 @@ function pressure_work_tendency!(Yₜ, Y, p, t, colidx, ::EDMFX)
     for j in 1:n
         if :ρe_tot in propertynames(Y.c)
             @. Yₜ.c.sgsʲs.:($$j).ρae_tot[colidx] -=
-                ᶜp[colidx] / Y.c.ρ[colidx] * Yₜ.c.sgsʲs.:($$j).ρa[colidx]
+                ᶜp[colidx] *
+                draft_area(Yₜ.c.sgsʲs.:($$j).ρa[colidx], Y.c.ρ[colidx])
```
szy21, line 12: I would keep this the same, as this is the tendency of ρa * (p / ρ). Although the results are the same this is not really area fraction.

szy21 (approved):

PR #2131, opened 2023-09-20
Remove unused fieldvector

charleskawczynski:
Followup to #2130

### DIFF
--- a/src/cache/cache.jl
+++ b/src/cache/cache.jl
@@ -87,7 +87,6 @@ function default_cache(
         test,
         moisture_model = atmos.moisture_model,
         model_config = atmos.model_config,
-        Yₜ = similar(Y), # only needed when using increment formulation
         limiter,
         ᶜΦ,
         ᶠgradᵥ_ᶜΦ = ᶠgradᵥ.(ᶜΦ),
--- a/src/callbacks/callbacks.jl
+++ b/src/callbacks/callbacks.jl
@@ -266,7 +266,7 @@ NVTX.@annotate function compute_diagnostics(integrator)
     if p.atmos.precip_model isa Microphysics0Moment
         (; ᶜS_ρq_tot, col_integrated_rain, col_integrated_snow) = p
         Fields.bycolumn(axes(Y.c)) do colidx
-            precipitation_tendency!(p.Yₜ, Y, p, t, colidx, p.precip_model)
+            precipitation_tendency!(nothing, Y, p, t, colidx, p.precip_model)
         end # TODO: Set the diagnostics without computing the tendency.
         precip_diagnostic = (;
             precipitation_removal = ᶜS_ρq_tot,
--- a/src/parameterized_tendencies/microphysics/precipitation.jl
+++ b/src/parameterized_tendencies/microphysics/precipitation.jl
@@ -105,8 +105,10 @@ function precipitation_tendency!(
     thermo_params = CAP.thermodynamics_params(params)
     cm_params = CAP.microphysics_params(params)
     compute_precipitation_cache!(Y, p, colidx, precip_model, turbconv_model)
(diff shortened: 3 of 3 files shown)


charleskawczynski:
Why is
```julia
            precipitation_tendency!(p.Yₜ, Y, p, t, colidx, p.precip_model)
        end # TODO: Set the diagnostics without computing the tendency.
```
used in the diagnostics? cc @trontrytel . Can we split this into precompute and add a function that adds fluxes so that we can remove `p.Yₜ`?

trontrytel:
Not sure why? Not wanting to cache more things and or duplicate the code?

I don't have the bandwidth to fix this this week. Feel free to open an issue and tag me on it. I can try getting it done before the next sprint

> used in the diagnostics? cc @trontrytel . Can we split this into precompute and add a function that adds fluxes so that we can remove `p.Yₜ`?

szy21:
I think we might use it to make sure the precipitation tendency is consistent with the state, as we call `set_precomputed_quantities!` in the diagnostics which updates the thermo state. It might be ok to remove it (i.e. I think it's ok if the diagnostics we save are slightly inconsistent, but maybe other people have different opinions.)

charleskawczynski:
> I think we might use it to make sure the precipitation tendency is consistent with the state, as we call `set_precomputed_quantities!` in the diagnostics which updates the thermo state. It might be ok to remove it (i.e. I think it's ok if the diagnostics we save are slightly inconsistent, but maybe other people have different opinions.)

I think it's fine to precompute things needed for precipitation tendencies, but the precip tendendencies themselves don't seem to be used in the diagnostics-- Also if they are, then I think we can get them from the integrator?

charleskawczynski:
@trontrytel, for now, I've just wrapped the tendency updates in `if`-blocks (`if !isnothing(Yₜ); Yₜ.var += ...; end`), so that we can pass in `nothing` from diagnostics-- all other quantities are computed (so `p` should be appropriately updated). Is this okay for now?

szy21:
> > I think we might use it to make sure the precipitation tendency is consistent with the state, as we call `set_precomputed_quantities!` in the diagnostics which updates the thermo state. It might be ok to remove it (i.e. I think it's ok if the diagnostics we save are slightly inconsistent, but maybe other people have different opinions.)
> 
> I think it's fine to precompute things needed for precipitation tendencies, but the precip tendendencies themselves don't seem to be used in the diagnostics-- Also if they are, then I think we can get them from the integrator?

We calculate col_integrated_rain and col_integrated_snow in `precipitation_tendency!`, that's probably why we call it here. We can move them somewhere else though - we don't need them in the tendency.

trontrytel:
> @trontrytel, for now, I've just wrapped the tendency updates in `if`-blocks (`if !isnothing(Yₜ); Yₜ.var += ...; end`), so that we can pass in `nothing` from diagnostics-- all other quantities are computed (so `p` should be appropriately updated). Is this okay for now?

Sounds good. Apologies it's such a mess over there. It would be great to allocate some time and straighten things up (looking at myself pointedly).

The logic we want would be to put `compute_precipitation_cache` somewhere in `precomputed_quantities`, instead of calling it from the tendency? And then use it from both the tendency and diagnostics?

charleskawczynski:
> Sounds good. Apologies it's such a mess over there. It would be great to allocate some time and straighten things up (looking at myself pointedly).
> 
> The logic we want would be to put `compute_precipitation_cache` somewhere in `precomputed_quantities`, instead of calling it from the tendency? And then use it from both the tendency and diagnostics?

No problem. I'm just trying to incrementally improve on https://github.com/CliMA/ClimaCore.jl/issues/1467 for now. It's not clear to me how to best structure things, but there's definitely room for improvement.

PR #2135, opened 2023-09-21
Ignore HDF5 in JET reports

charleskawczynski:
HDF5 has many many runtime dispatch warnings in JET, and while I do think that some of them could be fixed, IIRC others are unavoidable due to working with generic types. So, this PR ignores that module to reduce noise in the JET report.

A build of failures are here: https://buildkite.com/clima/climaatmos-ci/builds/13072#018ab62d-8411-4cf8-a1b9-c29547cdef3b


charleskawczynski:
Ah, this requires adding HDF5 as a direct dependency. Maybe we can just do this in the perf env?

simonbyrne:
Sure, that seems reasonable.

charleskawczynski:
Actually, we're running these tests in the examples, so let's just add to the other environments without a compat entry so that we stay in sync with ClimaCore's version.

charleskawczynski:
I'm going to move the jet tests to the perf scripts, we'll be monitoring fewer cases, but we can add new ones as needed.

PR #2141, opened 2023-09-22
Update dependencies

charleskawczynski:
This PR upgrades the dependencies, including ClimaTimeSteppers. The new version fuses some operations, resulting in precision error differences.


szy21:
Ah, I just did #2140, but I didn't update ClimaTimesteppers I think.

charleskawczynski:
> Ah, I just did https://github.com/CliMA/ClimaAtmos.jl/pull/2140, but I didn't update ClimaTimesteppers I think.

No problem, we need to bump the ref counter for this one anyway

PR #2144, opened 2023-09-22
Hopefully remove last references to Julia 1.8

Sbozzolo:


bors.toml
```diff
@@ -1,7 +1,4 @@
 status = [
-  # "ci 1.8.5 - ubuntu-latest",
```
charleskawczynski, line 2: It looks like CI is working on ubuntu and macos. Let's uncomment this and update 1.8.5 to 1.9

charleskawczynski: Buildkite is a better success metric, but we do have some tests in test/runtests.jl that are not run on buildkite, so we really should be checking that.

charleskawczynski (approved):

PR #2149, opened 2023-09-22
Move parameter logging to `get_integrator`

nefrathenrici:
From @Sbozzolo's comment on #2087

### DIFF
--- a/src/parameters/create_parameters.jl
+++ b/src/parameters/create_parameters.jl
@@ -70,7 +70,7 @@ function create_parameter_set(config::AtmosConfig)
         TCP.TurbulenceConvectionParameters;
         subparam_structs = (; microphys_params, surf_flux_params),
     )
-    param_set = create_parameter_struct(
+    return create_parameter_struct(
         CAP.ClimaAtmosParameters;
         subparam_structs = (;
             thermodynamics_params = thermo_params,
@@ -81,9 +81,4 @@ function create_parameter_set(config::AtmosConfig)
             turbconv_params,
         ),
     )
-    if parsed_args["log_params"]
-        logfilepath = joinpath(@__DIR__, "$(parsed_args["job_id"])_$FT.toml")
-        CP.log_parameter_information(toml_dict, logfilepath)
-    end
-    return param_set
 end
--- a/src/parameters/create_parameters.jl
+++ b/src/parameters/create_parameters.jl
@@ -7,9 +7,6 @@ import SurfaceFluxes.UniversalFunctions as UF
 import Insolation.Parameters as IP
 import Thermodynamics as TD
 import CloudMicrophysics as CM
-# TODO: Remove these imports?
-import ClimaCore
-import ClimaCore as CC
 
 function create_parameter_set(config::AtmosConfig)
     # Helper function that creates a parameter struct. If a struct has nested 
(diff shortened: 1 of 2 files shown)


src/solver/types.jl
```diff
@@ -459,6 +460,14 @@ function AtmosConfig(;
...
         @info "Running ClimaCore in unthreaded mode."
     end
+    if config["log_params"]
+        default_output = haskey(ENV, "CI") ? job_id : joinpath("output", job_id)
+        out_dir = parsed_args["output_dir"]
+        output_dir = isnothing(out_dir) ? default_output : out_dir
+        mkpath(output_dir)
+        logfilepath = joinpath(output_dir, "$(parsed_args["job_id"])_$FT.toml")
+        CP.log_parameter_information(toml_dict, logfilepath)
+    end
```
Sbozzolo, line 470: You could move this function to `get_integrator`, where the `output_dir` is already defined (after `get_simulation`)

src/parameters/create_parameters.jl
```diff
@@ -7,9 +7,6 @@ import SurfaceFluxes.UniversalFunctions as UF
 import Insolation.Parameters as IP
 import Thermodynamics as TD
 import CloudMicrophysics as CM
-# TODO: Remove these imports?
-import ClimaCore
-import ClimaCore as CC
```
Sbozzolo, line 12: What's the purpose of this change with regards to moving parameter logging?

nefrathenrici: They don't seem to serve any purpose, I think they are old fragments that someone forgot to delete.

src/solver/types.jl
nefrathenrici: TY!

src/parameters/create_parameters.jl
Sbozzolo: Maybe they should be removed in a different commit if they are not logically related to the change under consideration?

src/solver/type_getters.jl
```diff
@@ -586,6 +586,11 @@ function get_simulation(config::AtmosConfig, comms_ctx)
     output_dir = isnothing(out_dir) ? default_output : out_dir
     mkpath(output_dir)
 
+    if parsed_args["log_params"]
+        logfilepath = joinpath(output_dir, "$(job_id)_$FT.toml")
+        CP.log_parameter_information(toml_dict, logfilepath)
+    end
+
```
Sbozzolo, line 593: I would have put this directly in the `get_integrator`. I am not sure if logging the parameters logically belongs to to the `get_simulation` function. 

Also, maybe we can have the word `param`(eters) in the name of the output file

src/solver/type_getters.jl
```diff
@@ -787,6 +787,11 @@ function get_integrator(config::AtmosConfig)
     atmos = get_atmos(config, params)
     numerics = get_numerics(config.parsed_args)
     simulation = get_simulation(config, config.comms_ctx)
+    simulation.output_dir
```
Sbozzolo, line 790: I think you might have forgot to remove this line

src/solver/type_getters.jl
```diff
@@ -787,6 +787,11 @@ function get_integrator(config::AtmosConfig)
...
+    if config.parsed_args["log_params"]
+        filepath = joinpath(output_dir, "$(job_id)_parameters.toml")
```
Sbozzolo, line 792: I think this will fail because `output_dir` is not defined here.

src/solver/type_getters.jl
nefrathenrici: Yes, this is a typo. thanks!

Sbozzolo (approved):

charleskawczynski (approved):

Issue #2093, opened 2023-09-15
Configuration priorities in `AtmosCoveragePerfConfig` are surprising


Sbozzolo:
[`AtmosCoveragePerfConfig` ](https://github.com/CliMA/ClimaAtmos.jl/blob/f751e691c05ae935ff30d82b7ef804021f5355ea/src/solver/types.jl#L376-L410)advertises itself as creating a `model configuration for many performance tests`. The docstring says that the priorities for the configuration s are:
```
Creates a config from the following in order of top priority to last:
1. Configuration from the given config file/dict
2. Default perf configuration
3. Target job configuration
4. Default configuration
```
This is very unexpected. 

The flag `target_job` says: `An (optional) job to target for analyzing performance`. So, one would expect that something like
```
config = CA.AtmosCoveragePerfConfig(config_dict = Dict("target_job" => "sphere_baroclinic_wave_rhoe_equilmoist"))
```
would run the `sphere_baroclinic_wave_rhoe` example. Instead, the default `perf` configuration has higher priority, so, the following configurations are changed (from `default_perf.yml`):
```
moist: "equil"
apply_limiter: true
z_elem: 25
dt: "1secs"
surface_setup: "DefaultExchangeCoefficients"
t_end: "10secs"
vert_diff: "true"
h_elem: 12
forcing: "held_suarez"
precip_model: "0M"
dt_save_to_sol: "Inf"
rad: "allskywithclear"
```
At this point, not much is left of `sphere_baroclinic_wave_rhoe` and a completely different configuration is being profiled. 

Should `default_perf.yml` be ignored when a `target_job` is passed? Maybe `default_perf` should be thought not as a default, but as a `all_perf`, ie, something that actives several parts of the code so that the benchmark is closer to a realistic complete run.

Sbozzolo:
In addition to the above, `ClimaAtmos.AtmosCoveragePerfConfig()` fails because `YAML.load_file(parsed_args["config_file"])) is called with `Nothing` as argument.

This is not the same behavior as `AtmosConfig`, which uses the default configuration when called with no options.

charleskawczynski:
[...]

PR #2157, opened 2023-09-25
Move perf funcs to perf folder

charleskawczynski:
This PR moves some of perf funcs to the perf folder, since this isn't needed by users.


charleskawczynski:
@Sbozzolo, would you say that this closes #2093 ?

perf/common.jl
```diff
@@ -0,0 +1,32 @@
...
+    AtmosCoveragePerfConfig()
+    AtmosCoveragePerfConfig(; s, config_dict)
+
+Creates a model configuration that covers many physics components.
+
+The configuration precedence is as follows:
+1. Configuration from the given config file/dict (highest precedence)
+2. Default perf configuration (to increase coverage)
+4. Default configuration (lowest precedence)
+"""
```
Sbozzolo, line 11: This should be updated: `s` is no longer an accepted argument, and `4.` should become `3.`

src/solver/types.jl
```diff
@@ -439,6 +403,10 @@ function AtmosConfig(;
         @info "Using default configuration"
         default_config_dict()
     end
+    return AtmosConfig(config; comms_ctx)
+end
+
+function AtmosConfig(config::Dict; comms_ctx = nothing)
```
Sbozzolo, line 409: It would be useful to add docstrings for these functions since they are user-facing.

Sbozzolo (commented):
It looks good. 

I still find this system to configure a simulation quite convoluted and difficult to follow, but this PR is step towards reducing the complexity. For instance, this change allows me to directly pass a dictionary to `AtmosConfig`, which I really like because I can see exactly what configurations I am setting.

Sbozzolo (approved):

PR #2160, opened 2023-09-26
reduce t_end in edmfx_trmm_box

szy21:
## Purpose 

So that it won't time out.

## Content

----


szy21:
cc @trontrytel

Issue #2015, opened 2023-08-24
Only call set_precomputed_quantities! once in the implict stage


simonbyrne:
Currently we call `set_precomputed_quantities!` in the `Wfact!`:
https://github.com/CliMA/ClimaAtmos.jl/blob/17c5a2f13cc01b013c1f1bcd9a8857943a147b8d/src/prognostic_equations/implicit/wfact.jl#L64
and once again in the `implicit_tendency!`:
https://github.com/CliMA/ClimaAtmos.jl/blob/17c5a2f13cc01b013c1f1bcd9a8857943a147b8d/src/prognostic_equations/implicit/implicit_tendency.jl#L12

But the state should be the same in both cases, so this should be unnecessary.

simonbyrne:
This appears to break the SSP tests, since they don't quite follow the same logic.

I think the general solution here is to move `set_precomputed_quantities!` so that it is applied _after_ each stage update: in this way, our cache will always be in sync with the current state (which means that we don't need to call it in the diagnostics).

simonbyrne:
We also call it twice in the explicit stage: once in `limited_tendencies!`:
https://github.com/CliMA/ClimaAtmos.jl/blob/f751e691c05ae935ff30d82b7ef804021f5355ea/src/prognostic_equations/limited_tendencies.jl

and once in `remaining_tendencies!`:
https://github.com/CliMA/ClimaAtmos.jl/blob/f751e691c05ae935ff30d82b7ef804021f5355ea/src/prognostic_equations/remaining_tendency.jl#L7

With the ARS343, the second is called directly after the first. Can we safely get rid of it in the second one?

simonbyrne:
@dennisYatunin ?

dennisYatunin:
Yes, the two explicit tendencies are computed for the same state, so their precomputed quantities should be identical. The order in which the tendencies are evaluated is implementation-dependent, though, so it would definitely be safer to separate out the precomputed quantities computation and guarantee that it will only be called once.

PR #2161, opened 2023-09-27
Update manifest files, overload ClimaTimeSteppers

charleskawczynski:
This PR upgrades to the latest dependencies and overloads ClimaTimeSteppers methods so that we can more easily iterate on fixing #2015. (cc @simonbyrne)


charleskawczynski:
This will allow us to more easily iterate on tackling #2015.

charleskawczynski:
We can push this back to ClimaTimeSteppers once we iron out the issues.

PR #2164, opened 2023-09-28
limit mixing length by z - z_sfc

szy21:
## Purpose 

Replace the upper limit of mixing length (l_max, default 1e6) by z - z_sfc.

## Content

----


szy21:
@tapios Is it ok to do this, or is it better to limit the smooth minimum by z - z_sfc? It probably doesn't make any difference. I did this so that we can remove `l_max` altogether.

tapios (approved):
This is fine.

PR #2172, opened 2023-09-29
Update compats and dependencies

charleskawczynski:
This PR updates CUDA, SciMLBase and StatsBase compats, and the dependencies.


charleskawczynski:
Supersedes #2151.

PR #2176, opened 2023-10-01
Fix docs

### LINKED ISSUE #2068: DocumenterCitation error in generating documentation
The documentation runs with a non-fatal error (see, e.g., https://github.com/CliMA/ClimaAtmos.jl/actions/runs/6101954504/job/16559376893#step:5:8). I tracked this down as an incompatibility between `doctest` and `DocumenterCitations`. The error is reported [upstream](https://github.com/JuliaDocs/DocumenterCitations.jl/issues/34).

charleskawczynski:
I think that this should close #2068. Documenter is now strict by default, so that keyword was removed.

### DIFF
--- a/docs/Manifest.toml
+++ b/docs/Manifest.toml
@@ -2,7 +2,7 @@
 
 julia_version = "1.9.3"
 manifest_format = "2.0"
-project_hash = "7fc686bc0a71a5f83e48b1285d03c4910b039279"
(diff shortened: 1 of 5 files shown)


charleskawczynski:
@Sbozzolo, please reopen #2068 if it doesn't close it. I think this should at least fix the broken docs.

PR #2179, opened 2023-10-02
Add support for online remapping

Sbozzolo:
This PR 
- Transforms writers to standalone objects instead of functions
- Changes the diagnostics to have one writer for all the diagnostics
- Adds support to a general NetCDF writer

This PR requires https://github.com/CliMA/ClimaCore.jl/pull/1475 to be merged.


Sbozzolo:
Topography available as diagnostic `orog`:

display

post_processing/plot/plot_everything.jl
```diff
@@ -0,0 +1,236 @@
+# plot_everything.jl: a simple Julia script to make a suite of plots for quick inspection as
+# part of CI.
+
+if length(ARGS) < 1
+    error("Run as plot_everything.jl <path of datadir>")
+end
+
+using Statistics # Used for `mean`
+using NCDatasets
+using CairoMakie
```
charleskawczynski, line 10: Can we use the `import` pattern here instead, and qualify where things are coming from?

Sbozzolo:
With https://github.com/CliMA/ClimaCore.jl/pull/1497, this PR has full support for topography.

Example: `z` is from sea level (with wrong titles, `level` -> `elevation from MSL` )
rhoa_inst_1
`z` is from surface
rhoa_inst_1

akshaysridhar:
@Sbozzolo cc @nefrathenrici 
In this example I'm using this PR branch to output mean values for a list of user-defined variables: 
```
julia> get_filenames(dir)
[ Info: Detected: 24 total files in directory
24-element Vector{Any}:
 "cli_6.0h_average.nc"
 "clw_6.0h_average.nc"
 "evspsbl_6.0h_average.nc"
 "ha_6.0h_average.nc"
 "hfes_6.0h_average.nc"
 "hur_6.0h_average.nc"
 "hus_6.0h_average.nc"
 "hussfc_6.0h_average.nc"
 "pfull_6.0h_average.nc"
 ⋮
 "rsd_6.0h_average.nc"
 "rsdcs_6.0h_average.nc"
 "rsu_6.0h_average.nc"
 "rsucs_6.0h_average.nc"
 "ta_6.0h_average.nc"
 "thetaa_6.0h_average.nc"
 "ts_6.0h_average.nc"
 "ua_6.0h_average.nc"
 "va_6.0h_average.nc"

julia> generate_all_summary_plots(dir)
[ Info: Detected: 24 total files in directory
[ Info: Detected: 24 unique variables
```

Taking `ta` as an example (e.g. from Nat's observations in his experiments), for  "ta_6.0h_average.nc", we have: 
```

julia> data["ta"]
ta (2 × 180 × 80 × 10)
  Datatype:    Float32
  Dimensions:  time × lon × lat × z
  Attributes:
   long_name            = Air Temperature, average within 6.0 Hour(s)average
   units                = K
   comments             = 

julia> data["ta"] |> extrema
(0.0f0, 5.2507515f0)
```

@nefrathenrici  is this similar to what you see as well ?

Sbozzolo:
Thank you! I reverted the latest change the introduced this regression.

Sbozzolo:
Closes #2269

akshaysridhar:
There seem to be issues with this list of variables 
```
ha
cli
husup
clw
```

akshaysridhar:
Thanks @Sbozzolo. Above issue with specific variables is resolved -> ClimaAtmos from registry was being picked up as opposed to local dir / dev pkg.

akshaysridhar:
Screenshot 2023-11-01 at 2 40 53 PM
Plotting only lat-z for select fields; we can use some version of this sort of a chart for CI in the future

config/model_configs/box_hydrostatic_balance_rhoe.yml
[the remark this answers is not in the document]
Sbozzolo: If you don't specify anything, the default is to output instantaneous variables.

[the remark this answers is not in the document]
akshaysridhar: Noted,thanks!

akshaysridhar (commented):
In this job (I use this as an example / demo here - I haven't gone through all buildkite outputs at this stage) : https://buildkite.com/clima/climaatmos-ci/builds/15109#018c03a8-9e72-4fbc-aedb-8d0bff3c0374

We have, for instance, output files named 
`hus_1d_average.nc`
`ta_1d_average.nc`
`hussfc_1d_average.nc`

However the dimensions of, for instance, `hus` are 
```
  hus   (4 × 180 × 80 × 10)
    Datatype:    Float32 (Float32)
    Dimensions:  time × lon × lat × z
```
Perhaps this is a file naming issue (`average`) as the default reduction is `instantaneous` or `no reduction` (?)

Sbozzolo:
> In this job (I use this as an example / demo here - I haven't gone through all buildkite outputs at this stage) : https://buildkite.com/clima/climaatmos-ci/builds/15109#018c03a8-9e72-4fbc-aedb-8d0bff3c0374
> 
> We have, for instance, output files named `hus_1d_average.nc` `ta_1d_average.nc` `hussfc_1d_average.nc`
> 
> However the dimensions of, for instance, `hus` are
> 
> ```
>   hus   (4 × 180 × 80 × 10)
>     Datatype:    Float32 (Float32)
>     Dimensions:  time × lon × lat × z
> ```
> 
> Perhaps this is a file naming issue (`average`) as the default reduction is `instantaneous` or `no reduction` (?)

The simulation is 4 days long, so you have 4 computations of a daily average. The default reduction is `no reduction`, but there are also a bunch of default diagnostics defined in `default_diagnostics.jl` (which are averages).

perf/flame.jl
```diff
@@ -59,17 +60,17 @@ allocs = @allocated SciMLBase.step!(integrator)
 @info "`allocs ($job_id)`: $(allocs)"
 
 allocs_limit = Dict()
-allocs_limit["flame_perf_target"] = 4656
-allocs_limit["flame_perf_target_tracers"] = 204288
-allocs_limit["flame_perf_target_edmfx"] = 253440
-allocs_limit["flame_perf_diagnostics"] = 3016328
-allocs_limit["flame_perf_target_diagnostic_edmfx"] = 920960
+allocs_limit["flame_perf_target"] = 10_155_936
```
charleskawczynski, line 63: I think this needs to be fixed

Sbozzolo: When I first looked into this, I found that there were a lot of allocations in NCDatasets. I'll look again because this is probably not the case for this particular run (since we don't write to disk)

Sbozzolo: Ah no, now I remember. The allocations come from these functions:
```
RRTMGPI.array2field(
            FT.(cache.radiation.radiation_model.face_clear_sw_flux_dn),
            axes(state.f),
        )
```

charleskawczynski: Hm, I wonder if it's just the `FT` (https://github.com/CliMA/ClimaCore.jl/issues/1372), so maybe this can easily be fixed with `$(FT(...))`?

Sbozzolo: The other allocating diagnostics have

```
out .= Geometry.UVector.(cache.precomputed.ᶜu).components.data.:1
```

or

```
out .= Geometry.VVector.(cache.precomputed.ᶜu).components.data.:1
```

How do I deal with this?

simonbyrne: We need a function like 
https://github.com/CliMA/ClimaCore.jl/blob/30a42ece976016c713c70a5e54bc3fb41568ef11/src/Geometry/conversions.jl#L175

simonbyrne: An easy workaround for now:
```julia
u_component(u::LocalVector) = u.u
v_component(u::LocalVector) = u.v
w_component(u::LocalVector) = u.w
out .= u_component.(Geometry.UVector.(cache.precomputed.ᶜu))
```

Sbozzolo: This worked and dramatically reduced allocations! :)

Sbozzolo: Among the remaining allocations I am finding:
```
julia> @allocated integrator.p
6128

julia> @allocated integrator.u
160

julia> @allocated integrator.t
128
```
Which I find quite interesting and I don't understand. 

I also added a check for excessive allocations in the constructor of the simulation.

The EDMF case still allocates a lot. The same diagnostics allocate much more compared to the non EDMF case. I am still debugging this.

Sbozzolo: Here are where the EDMF allocations happen:

Function # 68 is the one at line 699 on the right panel

image

simonbyrne: This could simply be dynamic dispatch? Not sure there is much you can do about this.

PR #2184, opened 2023-10-03
If TOML path is not found, try prepending pkgdir to the path

### LINKED ISSUE #2167: Reading parameters from `toml` in the yaml file assumes that the toml is in the working directory
Steps to reproduce:
```julia
import ClimaAtmos as CA
cd("/tmp")
CA.AtmosConfig(; config_dict = CA.config_from_target_job("sphere_aquaplanet_rhoe_equilmoist_allsky_gw_raw_zonallyasymmetric"))
```
Returns this errror:
```
"toml/sphere_aquaplanet_rhoe_equilmoist_allsky_gw_raw_zonallyasymmetric.toml": No such file
```
As a result, it is not practical to run the various configurations from anywhere but the root of the `ClimaAtmos` project.

I am not sure what is the cleanest way to fix this. On one side, reading the toml distributed with `ClimaAtmos` defies the purpose of having them as a configuration. On the other hand, this makes impossible to run the configurations unless one clones the `ClimaAtmos` repo and runs the code from the root of the project.

nefrathenrici:
This is needed to fix #2167, where configurations in `config` that used a TOML file could only be run from `pkgdir(CA)`. 

Content
- prepend `pkgdir(CA)` to the TOML path if the given TOML path does not point to a file.
- Remove support for multiple TOML files. I don't think that this is useful, and if anything could cause issues.
-

### DIFF
--- a/src/solver/types.jl
+++ b/src/solver/types.jl
@@ -430,11 +430,12 @@ function AtmosConfig(config::Dict; comms_ctx = nothing)
 
     # Check that config["toml"] is a filepath and the file does not exist
     if config["toml"] isa AbstractString && !isfile(config["toml"])
-        @info "Could not find TOML file at path $(joinpath(pwd(), config["toml"]))"
-        prepend_path = joinpath(dirname(@__FILE__), "..", "..", config["toml"])
-        if isfile(prepend_path)
-            @info "Prepending `pkgdir(CA)` to relative TOML file path"
-            config["toml"] = prepend_path
+        @info "Could not find TOML file at path `$(joinpath(pwd(), config["toml"]))`"
+        toml_path_in_atmos_dir =
+            joinpath(dirname(@__FILE__), "..", "..", config["toml"])
+        if isfile(toml_path_in_atmos_dir)
+            @info "Using parameter file at `$toml_path_in_atmos_dir`"
+            config["toml"] = toml_path_in_atmos_dir
         end
     end
     toml_dict = CP.create_toml_dict(FT; override_file = config["toml"])
--- a/src/solver/types.jl
+++ b/src/solver/types.jl
@@ -427,10 +427,18 @@ the default configurations set in `default_config_dict()`.
 function AtmosConfig(config::Dict; comms_ctx = nothing)
     config = override_default_config(config)
     FT = config["FLOAT_TYPE"] == "Float64" ? Float64 : Float32
-    toml_dict = CP.create_toml_dict(
-        FT;
-        override_file = CP.merge_toml_files(config["toml"]),
-    )
+
+    # Check that config["toml"] is a filepath and the file does not exist
+    if config["toml"] isa AbstractString && !isfile(config["toml"])
+        @info "Could not find TOML file at path $(joinpath(pwd(), config["toml"]))"
+        prepend_path = joinpath(dirname(@__FILE__), "..", "..", config["toml"])
+        if isfile(prepend_path)
+            @info "Prepending `pkgdir(CA)` to relative TOML file path"
+            config["toml"] = prepend_path
+        end
+    end
+    toml_dict = CP.create_toml_dict(FT; override_file = config["toml"])
+
     # TODO: is there a better way? We need a better
     #       mechanism on the ClimaCore side.
     if config["trunc_stack_traces"]
--- a/test/config_test.yml
+++ b/test/config_test.yml
@@ -1,0 +1,1 @@
+toml: "test/parameter_tests.toml"
--- a/config/default_configs/default_config.yml
+++ b/config/default_configs/default_config.yml
@@ -232,8 +232,8 @@ subsidence:
   help: "Subsidence [`nothing` (default), `Bomex`, `LifeCycleTan2018`, `Rico`, `DYCOMS`]"
   value: ~
 toml:
-  help: "TOML file(s) used to override model parameters"
-  value: []
+  help: "TOML file used to override model parameters"
+  value: ~
 prognostic_tke:
   help: "A flag for prognostic TKE [`false` (default), `true`]"
   value: false
--- a/config/gpu_configs/central_gpu_hs_rhoe_equilmoist_nz63_0M_55km_rs35km.yml
+++ b/config/gpu_configs/central_gpu_hs_rhoe_equilmoist_nz63_0M_55km_rs35km.yml
@@ -13,4 +13,4 @@ precip_model: "0M"
 rayleigh_sponge: true
 forcing: "held_suarez"
 job_id: "central_gpu_hs_rhoe_equilmoist_nz63_0M_55km_rs35km"
-toml: [toml/longrun_hs_rhoe_equilmoist_nz63_0M_55km_rs35km.toml]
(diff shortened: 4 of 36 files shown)


Sbozzolo:
What was the point of having lists in the `toml` entry?

nefrathenrici:
> What was the point of having lists in the `toml` entry?

@Sbozzolo It enabled multiple parameter files to be passed in. I thought that this flexibility might be useful, but it has not been. Since it is not used and could cause issues with multiple parameter sources, it seems better to remove for now.

Sbozzolo:
> > What was the point of having lists in the `toml` entry?
> 
> @Sbozzolo It enabled multiple parameter files to be passed in. I thought that this flexibility might be useful, but it has not been. Since it is not used and could cause issues with multiple parameter sources, it seems better to remove for now.

Thanks! Supporting a list will aksi introduce the problem of what to do when there's a conflict. So, I guess we can postpone supporting this until we find that we need to.

src/solver/types.jl
```diff
@@ -427,10 +427,18 @@ the default configurations set in `default_config_dict()`.
...
     FT = config["FLOAT_TYPE"] == "Float64" ? Float64 : Float32
-    toml_dict = CP.create_toml_dict(
-        FT;
-        override_file = CP.merge_toml_files(config["toml"]),
-    )
+
+    # Check that config["toml"] is a filepath and the file does not exist
+    if config["toml"] isa AbstractString && !isfile(config["toml"])
+        @info "Could not find TOML file at path $(joinpath(pwd(), config["toml"]))"
+        prepend_path = joinpath(dirname(@__FILE__), "..", "..", config["toml"])
```
Sbozzolo, line 434: I think the variable name `prepend_path` doesn't describe what the variable actually is. Maybe `toml_path_in_climaatmos_dir` or something like that?

src/solver/types.jl
```diff
@@ -427,10 +427,18 @@ the default configurations set in `default_config_dict()`.
...
+        if isfile(prepend_path)
+            @info "Prepending `pkgdir(CA)` to relative TOML file path"
```
Sbozzolo, line 436: Why not `Using bundled TOML file", which is a more user-friendly warning?

Sbozzolo (approved):

test/config_test.yml
```diff
@@ -0,0 +1 @@
+toml: "test/parameter_tests.toml"
```
Sbozzolo, line 1: Can you add a newline at the end of this file?

src/solver/types.jl
nefrathenrici: @charleskawczynski I don't think there's a better way of doing this in `src`. I am happy to include a commented snippet for devs copy-pasting into the REPL, something like `prepend_path = joinpath(pkgdir(CA), config["toml"])`

charleskawczynski: IMO, that means that here is a problem with the way we are doing this. I think that goes along the lines of @simonbyrne's suggestion to avoid grabbing things outside of `src/`. Maybe this logic can be extracted and moved into the driver, and we can simplify how `AtmosConfig` works?

nefrathenrici:
> > > What was the point of having lists in the `toml` entry?
> > 
> > 
> > @Sbozzolo It enabled multiple parameter files to be passed in. I thought that this flexibility might be useful, but it has not been. Since it is not used and could cause issues with multiple parameter sources, it seems better to remove for now.
> 
> Thanks! Supporting a list will aksi introduce the problem of what to do when there's a conflict. So, I guess we can postpone supporting this until we find that we need to.

There is already defined functionality in `CLIMAParameters.merge_toml_files()`, which is to warn the user about the overrides and merge parameters with the same priority as `merge` (rightmost argument takes first priority). See [here](https://github.com/CliMA/CLIMAParameters.jl/blob/main/src/file_parsing.jl#L268-L292)

src/solver/types.jl
nefrathenrici: IMO "bundled" is not clear. Are you ok with "Using TOML at $prepend_path"?

src/solver/types.jl
nefrathenrici: I don't see how that would work - running from anywhere with any configuration/parameter files isn't really needed in the driver, which is almost always run from `pkgdir(CA)`, but is needed for `AtmosConfig`.
It seems like the only way to make this work with the current system is this partial fix which assumes that parameter paths are relative to `pkgdir(CA)`. If that is not acceptable, we will probably need to rework with how parameters are taken in.

src/solver/types.jl
Sbozzolo: > IMO "bundled" is not clear. Are you ok with "Using TOML at $prepend_path"?

For users, that would result in

`Using TOML at `/home/sbozzolo/.julia/packages/1.9/ClimaAtmos/asd23/toml/very_long_name.toml`

The relevant information here I think is that we are using the one that comes when you install ClimaAtmos, but the path is not needed. I think we shouldn't need to print out this internal piece of information.

I don't know what's the best verbiage, maybe `Using default TOML file for job $job_id`

src/solver/types.jl
```diff
@@ -427,10 +427,19 @@ the default configurations set in `default_config_dict()`.
...
+    if config["toml"] isa AbstractString && !isfile(config["toml"])
+        @info "Could not find TOML file at path `$(joinpath(pwd(), config["toml"]))`"
+        toml_path_in_atmos_dir =
+            joinpath(dirname(@__FILE__), "..", "..", config["toml"])
+        if isfile(toml_path_in_atmos_dir)
+            @info "Using parameter file at `$toml_path_in_atmos_dir`"
+            config["toml"] = toml_path_in_atmos_dir
+        end
+    end
+    toml_dict = CP.create_toml_dict(FT; override_file = config["toml"])
```
charleskawczynski, line 441: This is mutating `config`, which means that `AtmosConfig` should be `AtmosConfig!`! I think it's better that we try to keep each step immutable. Why can't we just put this code into the driver? It's not a lot of code, and `AtmosConfig` will work in a much simpler way. Then, instead of `dirname(@__FILE__)` we can use `pkgdir(CA)`

src/solver/types.jl
nefrathenrici: I see the path being useful for users looking at a run log and being able to see which parameter filepath was used. If we don't print the path, it requires a bit of detective work for a user to figure out which file was used in a simulation.

src/solver/types.jl
Sbozzolo: I do agree with @charleskawczynski that ideally stuff in `src/` should not depend on stuff outside of it.

I think that `src/` should be treated as a library, so we when we mix it with stuff outside we are entangling library with an application of it.

> It seems like the only way to make this work with the current system is this partial fix which assumes that parameter paths are relative to pkgdir(CA)

I think that this is pretty much equivalent to hardcoding the parameters from a user point of view. (The point of view I take is 'how would this work if I were to just Pkg.add("ClimaAtmos")?' )

src/solver/types.jl
nefrathenrici: Sorry, I misunderstood the earlier comment. I agree `AtmosConfig()` being in the driver would make things a lot simpler, but how would a user create an `AtmosConfig` struct without including the driver?

Issue #1980, opened 2023-08-15
O1.2.1 Make ClimaAtmos GPU compatible


sriharshakandala:
# The Climate Modeling Alliance
### Purpose
Make `ClimaAtmos.jl` package GPU compatible.

### People and Personnel
- Lead: @sriharshakandala 
- Collaborators: @szy21 , @akshaysridhar, @trontrytel  
- Reviewers: @simonbyrne 

### CC
@simonbyrne @cmbengue @tapios 

```[tasklist]
### `RRTMGP.jl`
- [ ] Support single precision calculations https://github.com/CliMA/RRTMGP.jl/issues/337
- [ ] Tuning RRTMGP on GPU
- [ ] Tune/refactor RRTMGP interface in ClimaAtmos.jl as needed
- [ ] https://github.com/CliMA/RRTMGP.jl/pull/397
- [ ] https://github.com/CliMA/RRTMGP.jl/issues/400
```

```[tasklist]
### `SurfaceFluxes.jl`
- [ ] https://github.com/CliMA/SurfaceFluxes.jl/issues/113
- [ ] Tune/refactor SurfaceFluxes.jl interface in `ClimaAtmos.jl` as necessary
- [ ] https://github.com/CliMA/ClimaAtmos.jl/issues/2100
- [ ] https://github.com/CliMA/ClimaAtmos.jl/issues/2101
```

```[tasklist]
### `CloudMicroPhysics.jl`
- [ ] https://github.com/CliMA/CloudMicrophysics.jl/issues/164
- [ ] https://github.com/CliMA/CloudMicrophysics.jl/issues/173
- [ ] https://github.com/CliMA/CloudMicrophysics.jl/issues/153
- [ ] https://github.com/CliMA/CloudMicrophysics.jl/issues/219
```

```[tasklist]
### Diagnostic EDMF
```

```[tasklist]
### Insolation
- [ ] Use GPU compatible interpolation
- [ ] https://github.com/CliMA/Insolation.jl/pull/51
- [ ] https://github.com/CliMA/Insolation.jl/pull/60
```

```[tasklist]
### ClimaAtmos.jl
- [ ] https://github.com/CliMA/ClimaAtmos.jl/issues/2258
- [ ] https://github.com/CliMA/ClimaAtmos.jl/issues/2257
- [ ] https://github.com/CliMA/ClimaAtmos.jl/issues/2285
[...]

PR #2190, opened 2023-10-04
Fix moist gpu example failure for mixed spaces expression

### LINKED ISSUE #2191: Mismatched spaces error in vertical_diffusion_boundary_layer_tendency!
Current failure in the moist model
https://buildkite.com/clima/climaatmos-target-gpu-simulations/builds/88#018af76a-8b21-411b-a348-5aa181011b5f/138-220

which is here:
https://github.com/CliMA/ClimaAtmos.jl/blob/641cdeb37295822760fc74f9671ff73f06950bee/src/prognostic_equations/vertical_diffusion_boundary_layer.jl#L92-L98

From what I can tell, the problem is that we have fields at two different levels (one at level 1, one at level half).

For now, I think the easiest option might be to simply get rid of the error. Hopefully we can address this properly as part of https://github.com/CliMA/ClimaCore.jl/issues/1120

charleskawczynski:
This PR fixes the next breaking piece, blocking us from #1980. Closes #2191.

### DIFF
--- a/src/prognostic_equations/vertical_diffusion_boundary_layer.jl
+++ b/src/prognostic_equations/vertical_diffusion_boundary_layer.jl
@@ -84,10 +84,9 @@ function vertical_diffusion_boundary_layer_tendency!(
 
     FT = eltype(Y)
     interior_uₕ = Fields.level(Y.c.uₕ, 1)
-    interior_coordinates = Fields.level(Fields.coordinate_field(Y.c), 1)
     ᶠp = ᶠρK_E = p.ᶠtemp_scalar
     @. ᶠp[colidx] = ᶠinterp(ᶜp[colidx])
-    ᶜΔz_surface = Fields.Δz_field(interior_coordinates)
+    ᶜΔz_surface = Fields.Δz_field(interior_uₕ)
     @. ᶠρK_E[colidx] =
         ᶠinterp(Y.c.ρ[colidx]) * eddy_diffusivity_coefficient(
             C_E,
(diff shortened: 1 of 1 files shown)


src/prognostic_equations/vertical_diffusion_boundary_layer.jl
```diff
@@ -85,15 +85,14 @@ function vertical_diffusion_boundary_layer_tendency!(
     FT = eltype(Y)
     interior_uₕ = Fields.level(Y.c.uₕ, 1)
     interior_coordinates = Fields.level(Fields.coordinate_field(Y.c), 1)
```
szy21, line 87:
```suggestion
```

szy21 (approved):

src/prognostic_equations/vertical_diffusion_boundary_layer.jl
charleskawczynski: Ah, this is still used to get `ᶜΔz_surface`, but I'll grab it from `interior_uₕ` instead.

szy21: Ah ok, thanks! I was on my phone and didn’t see the whole file 😂

szy21: Actually, does Δz_field for a center field give the interval between cell center or cell face?

charleskawczynski: > Actually, does Δz_field for a center field give the interval between cell center or cell face?

I had thought about this, I think it would only make sense that it is the distance between cell faces, because the data lives on cell centers (and it should be centered / symmetric about cell centers). I should verify that, though. It returns `lg.∂x∂ξ.components.data.:9` for us, @simonbyrne, can you confirm that this is correct?

The MSE tables didn't need updated, so, if we have regression tests that use this, we should be in the clear.

szy21: We do have regression test for this, this should be correct then.

PR #2193, opened 2023-10-05
Update config for GPU DryBaroclinicWave benchmarks

sriharshakandala:
## Purpose 
Update z_elem to 63 for GPU DryBaroclinicWave benchmarks.

----


szy21:
Setting z_max to 55000 may cause unphysical initial condition for baroclinic wave. For testing the performance maybe keep z_max as 30000.

simonbyrne (approved):

PR #2195, opened 2023-10-05
Fix h-scaling plots

charleskawczynski:
I can't see the x/y axes in the h-scaling plots, this PR should fix that. The plot looks fine when I run locally but, for some reason, it's pretty truncated on the buildkite artifacts.


bloops (approved):

bloops:
Thanks for the fix!

PR #2200, opened 2023-10-05
add remaining diagnostics to the new diagnostics package

szy21:
## Purpose 

Add EDMFX diagnostics and add some default diagnostics when running simulations with EDMFX. I think this includes all the important diagnostics we want to have.

## To-do

- [x] Check if the diagnostics are physically correct

## Content

----

### DIFF
--- a/config/model_configs/diagnostic_edmfx_bomex_box.yml
+++ b/config/model_configs/diagnostic_edmfx_bomex_box.yml
@@ -27,3 +27,58 @@ z_stretch: false
 dt: "10secs" 
 t_end: "6hours" 
 dt_save_to_disk: "10mins"
+diagnostics:
+  - short_name: ts
+    period: 10mins
+  - short_name: ta
+    period: 10mins
+  - short_name: thetaa
+    period: 10mins
+  - short_name: ha
+    period: 10mins
+  - short_name: pfull
+    period: 10mins
+  - short_name: rhoa
+    period: 10mins
+  - short_name: ua
+    period: 10mins
+  - short_name: va
+    period: 10mins
+  - short_name: wa
+    period: 10mins
+  - short_name: hur
+    period: 10mins
+  - short_name: hus
+    period: 10mins
+  - short_name: clw
+    period: 10mins
+  - short_name: cli
+    period: 10mins
+  - short_name: hussfc
+    period: 10mins
+  - short_name: evspsbl
+    period: 10mins
+  - short_name: arup
+    period: 10mins
+  - short_name: waup
+    period: 10mins
+  - short_name: taup
+    period: 10mins
+  - short_name: thetaaup
+    period: 10mins
+  - short_name: haup
+    period: 10mins
+  - short_name: husup
+    period: 10mins
+  - short_name: hurup
+    period: 10mins
+  - short_name: clwup
+    period: 10mins
+  - short_name: cliup
+    period: 10mins
+  - short_name: waen
+    period: 10mins
+  - short_name: tke
+    period: 10mins
+  - short_name: lmix
+    period: 10mins
--- a/src/diagnostics/core_diagnostics.jl
+++ b/src/diagnostics/core_diagnostics.jl
@@ -47,14 +47,13 @@ error_diagnostic_variable(variable, model::T) where {T} =
     error_diagnostic_variable("Cannot compute $variable with model = $T")
 
 ###
-# Rho (3d)
+# Density (3d)
 ###
 add_diagnostic_variable!(
     short_name = "rhoa",
     long_name = "Air Density",
     standard_name = "air_density",
     units = "kg m^-3",
-    comments = "Density of air",
     compute! = (out, state, cache, time) -> begin
         if isnothing(out)
             return copy(state.c.ρ)
@@ -129,7 +128,6 @@ add_diagnostic_variable!(
     long_name = "Air Temperature",
     standard_name = "air_temperature",
     units = "K",
-    comments = "Temperature of air",
     compute! = (out, state, cache, time) -> begin
         thermo_params = CAP.thermodynamics_params(cache.params)
         if isnothing(out)
@@ -148,7 +146,6 @@ add_diagnostic_variable!(
     long_name = "Air Potential Temperature",
     standard_name = "air_potential_temperature",
     units = "K",
-    comments = "Potential temperature of air",
     compute! = (out, state, cache, time) -> begin
         thermo_params = CAP.thermodynamics_params(cache.params)
         if isnothing(out)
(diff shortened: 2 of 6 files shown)


config/model_configs/diagnostic_edmfx_bomex_box.yml
```diff
@@ -26,4 +26,5 @@ z_max: 3e3
 z_stretch: false
 dt: "10secs" 
 t_end: "6hours" 
-dt_save_to_disk: "10mins"
\ No newline at end of file
+dt_save_to_disk: "10mins"
+output_default_diagnostics: true
```
Sbozzolo, line 30: Can you please add a newline?

src/diagnostics/default_diagnostics.jl
```diff
@@ -92,7 +92,7 @@ daily_max(short_names; output_writer = HDF5Writer()) =
 Return a list of `ScheduledDiagnostics` that compute the daily min for the given variables.
 """
 daily_mins(short_names...; output_writer = HDF5Writer()) =
-    common_diagnostics(24 * 60 * 60, min, output_writer, short_names...)
+    common_diagnostics(24 * 60 * 60, nothing, output_writer, short_names...)
```
Sbozzolo, line 95: I think this should be `min`. Why did you change to `nothing`?

src/diagnostics/default_diagnostics.jl
```diff
@@ -186,6 +186,24 @@ Return a `ScheduledDiagnostics` that computes the hourly average for the given v
...
     hourly_averages(short_names; output_writer)[1]
 
+"""
+    tenmin_insts(short_names...; output_writer = HDF5Writer())
+
+
+Return a list of `ScheduledDiagnostics` that compute the 10 min instantaneous output for the given variables.
+"""
+tenmin_insts(short_names...; output_writer = HDF5Writer()) =
+    common_diagnostics(10 * 60, nothing, output_writer, short_names...)
```
Sbozzolo, line 196: I think that this will return in some warnings being raised. With `common_diagnostics` you set `output_every`, but `compute_every` is set to `:timestep`. If you don't want time reductions, `compute_every` will be adjusted to be the same as `output_every`, but a warning will be produced.

So, I think that you have to change this to
```suggestion
function tenmin_insts(short_names...; output_writer = HDF5Writer()) 
period = 10 * 60 # seconds
        return [
        ScheduledDiagnosticTime(
            variable = get_diagnostic_variable(short_name),
            compute_every = period,
            output_every = period, 
            output_writer = output_writer,
        ) for short_name in short_names
    ]

end
```

Sbozzolo (commented):
Looks good!

My main comment (other than the tiny details below) is that you the "comment" field in your variables is not really useful since you just paraphrase the `long_name`.

Maybe we can copy and paste the comments from the airtable with all the CMIP quantities?

src/diagnostics/default_diagnostics.jl
szy21: Oh, this is a typo, thanks!

src/diagnostics/core_diagnostics.jl
```diff
@@ -261,6 +261,70 @@ add_diagnostic_variable!(
...
+        return TD.ice_specific_humidity.(thermo_params, cache.ᶜts)
+    else
+        out .= TD.ice_specific_humidity.(thermo_params, cache.ᶜts)
+    end
+end
+
+add_diagnostic_variable!(
+    short_name = "cli",
+    long_name = "Mass Fraction of Cloud Ice",
+    standard_name = "mass_fraction_of_cloud_icein_air",
```
szy21, line 322:
```suggestion
    standard_name = "mass_fraction_of_cloud_ice_in_air",
```

src/diagnostics/default_diagnostics.jl
szy21: Thanks!

szy21:
> Maybe we can copy and paste the comments from the airtable with all the CMIP quantities?

All the EDMFX specific diagnostics don't have a corresponding variable in the CMIP table, as they don't use the same convection scheme in CMIP. The comments for all updraft variables (diagnostics ending with `up`) are mostly to just describe that the diagnostics is for the first updraft. The comments for environment variables are basically the `long_name`, maybe I can just remove them?

Sbozzolo:
> > Maybe we can copy and paste the comments from the airtable with all the CMIP quantities?
> 
> All the EDMFX specific diagnostics don't have a corresponding variable in the CMIP table, as they don't use the same convection scheme in CMIP. The comments for all updraft variables (diagnostics ending with `up`) are mostly to just describe that the diagnostics is for the first updraft. The comments for environment variables are basically the `long_name`, maybe I can just remove them?

When I look the [table](https://airtable.com/appYNLuWqAgzLbhSq/shrKcLEdssxb8Yvcp/tblL7dJkC3vl5zQLb), I see for example for `cli`:
```
Includes both large-scale and convective cloud. Calculate as the mass of cloud liquid water in the grid cell divided by the mass of air (including the water in all phases) in the grid cells. Precipitating hydrometeors are included ONLY if the precipitating hydrometeors affect the calculation of radiative transfer in model.
```

I don't know how much detail we should have, but I think that having more is better than having less (especially if the variable is never going to change).

szy21:
> 
> When I look the [table](https://airtable.com/appYNLuWqAgzLbhSq/shrKcLEdssxb8Yvcp/tblL7dJkC3vl5zQLb), I see for example for `cli`:
> 
> ```
> Includes both large-scale and convective cloud. Calculate as the mass of cloud liquid water in the grid cell divided by the mass of air (including the water in all phases) in the grid cells. Precipitating hydrometeors are included ONLY if the precipitating hydrometeors affect the calculation of radiative transfer in model.
> ```
> 
> I don't know how much detail we should have, but I think that having more is better than having less (especially if the variable is never going to change).

Oh, right, I thought you meant the EDMFX variables. Yes, I can add more comments to variables like `clw` and `cli`. One reason I didn't add it is that the comments from the table don't apply to our model yet. For example, for `cli`, we don't have convective cloud, or precipitating hydrometeors. But maybe we can just add the comments and change how the diagnostics are calculated later?

Sbozzolo:
> > When I look the [table](https://airtable.com/appYNLuWqAgzLbhSq/shrKcLEdssxb8Yvcp/tblL7dJkC3vl5zQLb), I see for example for `cli`:
> > ```
> > Includes both large-scale and convective cloud. Calculate as the mass of cloud liquid water in the grid cell divided by the mass of air (including the water in all phases) in the grid cells. Precipitating hydrometeors are included ONLY if the precipitating hydrometeors affect the calculation of radiative transfer in model.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I don't know how much detail we should have, but I think that having more is better than having less (especially if the variable is never going to change).
> 
> Oh, right, I thought you meant the EDMFX variables. Yes, I can add more comments to variables like `clw` and `cli`. One reason I didn't add it is that the comments from the table don't apply to our model yet. For example, for `cli`, we don't have convective cloud, or precipitating hydrometeors. But maybe we can just add the comments and change how the diagnostics are calculated later?

I think we should describe what we have, and update the description when we update the diagnostic.

szy21:
@Sbozzolo I updated the comment of several variables following the CMIP air table, and removed the comment of some variables where they are similar to the long name. I also set the default EDMF diagnostics to daily average, and included an example yaml file in ci that calculates 10-min instantaneous EDMF diagnostics. The results look physical. Would you like to take another look?

Sbozzolo:
> @Sbozzolo I updated the comment of several variables following the CMIP air table, and removed the comment of some variables where they are similar to the long name. I also set the default EDMF diagnostics to daily average, and included an example yaml file in ci that calculates 10-min instantaneous EDMF diagnostics. The results look physical. Would you like to take another look?

It looks good to me. The only thing: could you split the extremely long strings to multiline strings?

Also, I didn't realize how many EDMF examples we are running. In the future, if we want the whole suite of diagnostics for all the EMDF examples, we will probably have to define them as default in the code. (But this is good for now).

szy21:
> It looks good to me. The only thing: could you split the extremely long strings to multiline strings?

I changed some of the comment to multiline strings. Is that what you mean?

> Also, I didn't realize how many EDMF examples we are running. In the future, if we want the whole suite of diagnostics for all the EMDF examples, we will probably have to define them as default in the code. (But this is good for now).

Yeah, I think it would be good if we don't need to specify them for each job. I wonder if the config can take multiple yaml files now (it could before I think). If so, we can have one common diagnostic yaml file for the current edmf examples.

simonbyrne:
Part of #2200

PR #2199, opened 2023-10-05
Upgrade CUDA

charleskawczynski:
This PR upgrades the dependencies. Notably, to CUDA 5.0

PR #2201, opened 2023-10-05
Fix CUDA compat, upgrade deps

charleskawczynski:
This PR fixes the CUDA compat, until we get #2199 working, and upgrades the dependencies (to use the latest ClimaTimeSteppers).

I already ran CI with the latest ClimaTimeSteppers branch here: https://buildkite.com/clima/climaatmos-ci/builds/13626 and results looked like they match within machine precision, so I'm just going to increment the ref counter and r+.


charleskawczynski:
I'm going to override and manually merge (since the ref counter change implies that data has been moved on central).

PR #2209, opened 2023-10-06
add constant detrainment option

szy21:
## Purpose 

Adds constant timescale detrainment, as suggested by Tapio. I tried running bomex with constant timescale entrainment and detrainment with the same time scale, and the results look terrible (see comments), so I didn't change the configuration file in the end. But this option is still good to have for testing as it is the simplest detrainment closure.

## Content

----


szy21:
Here is bomex with constant entrainment and detrainment with a timescale of 900s.
image

szy21:
I'm going to merge this in as I would like to use it for some other tests. I didn't change any of the current EDMFX configurations.

PR #2210, opened 2023-10-06
update dependencies

szy21:
## Content

----


szy21:
@charleskawczynski This updates `SciMLBase` and `CLIMAParameters` and changes MSE. Does it make sense? The update to `CLIMAParameters` shouldn't change anything, but maybe `SciMLBase` could change?

szy21:
Actually the MSE error seems random.

PR #2211, opened 2023-10-07
Fix some type piracy

charleskawczynski:
This PR:
 - Removes overloading of Fields on Integers (which is type piracy, and results in invalidations)
 - Applies the minimalistic fixes by interpolating integers into getproperty calls.

As a compromise, this PR only removes `getindex` on `Field`s and `Integer`s:
```julia
Base.@propagate_inbounds Base.getindex(field::FDFields, i::Integer) =
    Base.getproperty(field, i)
```
Which is the worst of the type pirated methods. The next will be for `Base.getindex(field::FaceFields, i::CCO.PlusHalf)` and `Base.setindex!(field::FaceFields, v, i::CCO.PlusHalf)`, however, that requires adding `Spaces.level` and such to a lot of places, which is just a bit more work.


charleskawczynski:
Interestingly, Aqua is not catching that we're committing type piracy here.

PR #2215, opened 2023-10-09
Fix spatially varying boundary conditions on GPU

simonbyrne:
## Content

----

### DIFF
--- a/config/gpu_configs/gpu_hs_rhoe_equilmoist_nz63_0M_55km_rs35km.yml
+++ b/config/gpu_configs/gpu_hs_rhoe_equilmoist_nz63_0M_55km_rs35km.yml
@@ -1,12 +1,12 @@
 dt_save_to_disk: "10days"
-dt: "150secs"
-t_end: "300days"
-h_elem: 16
+dt: "100secs"
+t_end: "1days"
+h_elem: 30
 z_elem: 63
 dz_bottom: 30.0
 dz_top: 3000.0
 z_max: 55000.0
-kappa_4: 2.0e16
+kappa_4: 1.0e15
 vert_diff: "true"
 moist: "equil"
 precip_model: "0M"
--- a/.buildkite/gpu_pipeline/pipeline.yml
+++ b/.buildkite/gpu_pipeline/pipeline.yml
@@ -61,7 +61,6 @@ steps:
         artifact_paths: "gpu_aquaplanet_dyamond/*"
         agents:
           slurm_gpus: 1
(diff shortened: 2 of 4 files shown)


szy21:
Try use `kappa_4: 1.0e15`

sriharshakandala:
Can we remove the `soft_fail: true` flag in the CI pipeline!

config/gpu_configs/gpu_hs_rhoe_equilmoist_nz63_0M_55km_rs35km.yml
```diff
@@ -1,12 +1,12 @@
 dt_save_to_disk: "10days"
-dt: "150secs"
-t_end: "300days"
-h_elem: 16
+dt: "100secs"
+t_end: "1days"
```
charleskawczynski, line 3: I suggest we go at least 10 days, so that the callbacks are exercised.

simonbyrne: what callbacks need to be exercised?

sriharshakandala (approved):

config/gpu_configs/gpu_hs_rhoe_equilmoist_nz63_0M_55km_rs35km.yml
charleskawczynski: The default are probably fine, which will probably include a bunch of model-specific pieces (e.g., diagnostics for moisture etc.).

simonbyrne:
> Can we remove the `soft_fail: true` flag in the CI pipeline!

Done

config/gpu_configs/gpu_hs_rhoe_equilmoist_nz63_0M_55km_rs35km.yml
simonbyrne: can we just reduce the `dt_save_to_disk`?

simonbyrne: okay, but that doesn't need to be in the perf target. We should make sure it is in the CI though.

charleskawczynski: > can we just reduce the dt_save_to_disk?

Yeah, that works too

PR #2218, opened 2023-10-09
Update CUDA - take 2

charleskawczynski:
This PR upgrades the dependencies. Notably, to CUDA 5.0. Supersedes #2199.


simonbyrne:
Is it related to
```
┌ Warning: @cuprint does not support values of type Nothing
└ @ CUDA /central/scratch/esm/slurm-buildkite/climaatmos-ci/depot/cpu/packages/CUDA/nbRJk/src/device/intrinsics/output.jl:170
```

simonbyrne:
Maybe try handling the `T_guess::Nothing` case differently here:
https://github.com/CliMA/Thermodynamics.jl/blob/b49cb7249d275b6f3d7a0aafa0eed700b9091147/src/config_numerical_method.jl#L24-L33

charleskawczynski:
Maybe? I can add a hot fix. I'll probably need to fix Thermodynamics CUDA CI, first (assuming that it's breaking like RootSolvers)

charleskawczynski:
Alright, I've released a new version of Thermodynamics that doesn't potentially print `nothing`. Let's see how this goes 🤞🏻

charleskawczynski:
That was it!

PR #2229, opened 2023-10-11
set u_3 to 0 at the top cell face

### LINKED ISSUE #2228: Set u^3 and u_3 to be zero at the top cell face
Currently, u^3 and u_3 are not zero at the top cell face. We should set them to zero.

szy21:
## Purpose 

Closes #2228 
Increases ref contour

## Content

----


simonbyrne (approved):
This seems fine.

szy21:
hmm this changes MSE. The results look identical so it's probably just round off errors. @simonbyrne does this make sense?

PR #2237, opened 2023-10-13
clean up longrun

szy21:
## Purpose 

Removes `longrun_aquaplanet_rhoe_equilmoist_nz63_0M_55km_rs35km_clearsky` and only keeps the time-varying one. Also some other minor fixes.

cc @akshaysridhar 

## Content

----


akshaysridhar:
ty @szy21 I'm testing the new diagnostics and plotting options today (there is a running CI branch, will move towards that in each of existing longruns over the next week)

PR #2239, opened 2023-10-13
Upgrade ClimaCore

charleskawczynski:
This PR:
 - Removes the CUDA 5 compat entry (we need CUDA 4 to upgrade ClimaCore)
 - Updates the ClimaCoreMakie compat
 - Updates the manifest files


sriharshakandala:
Looks good to me!

sriharshakandala (approved):

simonbyrne:
As @Sbozzolo pointed put on Thursday, we should be bumping the compat requirements for ClimaCore

PR #2242, opened 2023-10-16
Remove soft fail from gpu HS

charleskawczynski:
This PR removes the soft fail from the GPU Held-Suarez job.


charleskawczynski:
Superseded by #2215

PR #2244, opened 2023-10-17
Clean up cache

### LINKED ISSUE #2217: Replace cache namedtuple with explicit struct
Currently, the integrator `cache` is a complex heterogeneous `NamedTuple`. The `cache` is partially flat, partially nested. For instance, `precomputed_quantities(Y, atmos)` is unpacked into the `cache`, but `simulation` is not. The details are not super easy to follow given the mix of splatting, unpacking, and merging that occur when building the cache.

Here's some of the fields that are in the cache:
```
    is_init
    simulation
    spaces
    atmos
    comms_ctx
    sfc_setup
    test
    moisture_model
    model_config
    Yₜ
    limiter
    ᶜΦ
    ᶠgradᵥ_ᶜΦ
    ᶜρ_ref
    ᶜp_ref
    ᶜT
    ᶜf
    ∂ᶜK∂ᶠu₃_data
    params
    energy_upwinding
    tracer_upwinding
    density_upwinding
    edmfx_upwinding
    do_dss
    ghost_buffer
    net_energy_flux_toa
    net_energy_flux_sfc
    env_thermo_quad
    ᶜspecific
    ᶜu
    ᶠu³
    ᶜK
    ᶜts
    ᶜp
    ᶜh_tot
    sfc_conditions
    ᶠtemp_scalar
    ᶜtemp_scalar
    ᶜtemp_scalar_2
    temp_data_level
    temp_data_level_2
    temp_data_level_3
    ᶜtemp_CT3
    ᶠtemp_CT3
    ᶠtemp_CT12
    ᶠtemp_CT12ʲs
    ᶠtemp_C123
    ᶜtemp_UVWxUVW
    sfc_temp_C3
    ᶜ∇²u
    ᶜ∇²specific_energy
    ᶜ∇²specific_tracers
    hyperdiffusion_ghost_buffer
    ᶜ∇²uʲs
    center_space
    radiation_model
    rayleigh_sponge_cache
    viscous_sponge_cache
    precipitation_cache
    subsidence_cache
    large_scale_advection_cache
    edmf_coriolis_cache
    forcing_cache
    radiation_cache
    non_orographic_gravity_wave_cache
    orographic_gravity_wave_cache
    edmfx_nh_pressure_cache
    Δt
    turbconv_cache
```

Some of the `cache` items are always added (e.g., `non_orographic_gravity_wave_cache`), others are added conditionally. Some fields in the cache are directly controlled by flags in `parsed_args` (e.g. `use_reference_state`, `test_dycore_consistency`).

The cache also contains information that is not related to the model (e.g., `output_dir`), information that is available elsewhere (e.g., ` Δt`), or that is redundant (`model_config = atmos.model_config`).

Some values are hardcoded in the computation of the cache (e.g., `T_ref = FT(255)`), others are added with possibly fragile checks (e.g., `ᶜf` is set by checking if `ᶜcoord` is a `LatLongZPoint`, and otherwise set using `f_plane_coriolis_frequency`).

Sbozzolo:
I went ahead and did a large first round of refactoring of the cache. This is what I have done:
- Removed the items that can be trivially removed from the cache
- Organized the cache by giving it a clear nested structure [see also, 1] and trying to establish ownership for the various quantities. The cache is now much simpler. It looks like:
```julia
p = (;
        simulation,
        atmos,
        core,
        sfc_setup,
        params,
        precomputed,
        scratch,
        hyperdiff,
        rayleigh_sponge,
        viscous_sponge,
        precipitation,
        subsidence,
        large_scale_advection,
        edmf_coriolis,
        forcing,
        non_orographic_gravity_wave,
        orographic_gravity_wave,
        radiation_mode,
        turbconv,
        ghost_buffer,
        Δt = dt,
    )
```
- The temporary variables are all in `scratch`
- Note that now it is no longer possible to access variables like `ᶜu` from `p`. It has to be `p.precomputed`. I am not super excited about this change and its impact on existing code. I also think that the ideal change would be break down precomputed and make it more modular. However, precomputed is one of the most involved components of the current cache (hyperdiff gets also some conditional quantities depending on EDMF).
- Defined a standard pattern to obtain the cache for a given. For all the fields in `AtmosModel` except two, now the pattern is something like `precipitation = precipitation_cache(Y, atmos)`. Dispatch over details of the submodel is used internally (for instance `precipitation_cache(Y, atmos) = precipitation_cache(Y, atmos, atmos.precip_model)`). The two items that could not fit this pattern are: (1) radiation, depending on the model more arguments are needed, (2) turbconv_model, it currently uses `parsed_args` for `"imex_edmf_turbconv"`, `"imex_edmf_gm"`, and so on (inside `dycore_equations_deprecated`, which is why I didn't bother fixing it too much).
- Defined a `generate_limiters_func!` that returns `limiters_func!` depending on the model (I did this also to remove the limiter from the cache).
- Moved the details of the numerics inside the `AtmosModel`, with a new `AtmosNumerics` type.
- Removed `radiation_model` (now everything is in `radiation_mode`).
- Cleaned up (a little bit) the hack for dummy surfaces.
- Removed unused variables here and there.
 
Easier steps that can be taken next:
- `sfc_setup` is used exclusively in `update_surface_conditions!` in `set_precomputed_quantities`.
- `simulation` is pretty much only used for `start_date` for the insolation (+ `job_id` in the perf/post processing)

Difficult step that can be taken next:
- `precomputed_quantities` is involved, especially with all the EDMF stuff. The `precomputed` and the `hyperdiff` tuples break the idea of ownership because it contains all sorts of variables that logically should belong to other fields. This could be refactored, but it requires thinking about the mechanism for (pre)computing variables.
- Move all the named tuples to structs (see also [1])

Open questions:
- `core` contains `ᶜΦ, ᶠgradᵥ_ᶜΦ, ᶜρ_ref, ᶜp_ref, ᶜT, ᶜf, ∂ᶜK∂ᶠu₃_data`. Is `core` the best name? Should all these variables be here?
- `T_ref` is hardcoded to `FT(255)`. Is this correct? 
- Can we remove this block: https://github.com/CliMA/ClimaAtmos.jl/blob/2da96b14c8ae24ac4474cad6b56162e8a74b90f3/src/cache/precomputed_quantities.jl#L315-L328?
- Is this code correct and relevant for all the configurations?
https://github.com/CliMA/ClimaAtmos.jl/blob/2da96b14c8ae24ac4474cad6b56162e8a74b90f3/src/cache/cache.jl#L54-L61
- Can we remove this hack? https://github.com/CliMA/ClimaAtmos.jl/blob/2da96b14c8ae24ac4474cad6b56162e8a74b90f3/src/cache/temporary_quantities.jl#L61-L62
- Is there a more explicit way to achieve what this code is doing (this pattern is common throughout the code and I think that it is just checking energy_mode)
https://github.com/CliMA/ClimaAtmos.jl/blob/2da96b14c8ae24ac4474cad6b56162e8a74b90f3/src/prognostic_equations/advection.jl#L26

Cleaning up stuff and organizing the cache more neatly while still using `NamedTuples` led to a 5-10 % overall improvement in latency (which compounds the ClimaCore fix for mutable spaces). Moving the cache from a `NamedTuple` to a struct of `NamedTuples` bought an additional 5-10 % of improvement in latency. As a result, merging this PR reduces the compile time for jobs in the CI by up to 9 minutes.

As an experiment, I turned one of the `NamedTuples` inside the cache into a struct (the `core` one). That led to a percent-level improvement in latency, not enough to stand up the experimental noise, but enough to give another hint that our `NamedTuples` are not the best for compilation.

[1] There's some [lost julia knowledge](https://github.com/JuliaLang/julia/issues/37642) about (named) tuples with many elements being slow. We should probably just start avoid large named tuples.

Closes #2217 

CC: @charleskawczynski @LenkaNovak @szy21 

Todo:
- [ ] Two EDMFX tests are timing out. Figure out what's happening


Sbozzolo:
I am a little stumped on what to do with the tests that timeout. I cannot reproduce on my laptop and on the cluster, so I am extrapolating some guesses. I possibly narrowed down the problem to an inference issue in the EDMFX precomputed variables. The tests time out because julia cannot even compile the code.

`set_edmf_precomputed_quantities!` is indeed a big and complex function, but I don't understand what change led to this explosion in inference time on the cluster.

szy21:
All EDMF cases call the same `set_edmf_precomputed_quantities!`, so it is also confusing that only two of them time out.

Sbozzolo:
> All EDMF cases call the same `set_edmf_precomputed_quantities!`, so it is also confusing that only two of them time out.

At a closer inspection, all the EDMF cases are affected (with compile and run time slowdowns)

charleskawczynski (commented):
Can we break this PR down into smaller, easily merge-able parts? My suggestion of some good peel off PRs would be:
 - Remove `comms_ctx` from `p`, and get from the space instead.
 - Rename `param_set` to `params` (this will be pretty big on its own). Also, I think we use the name `params` for some other data structure elsewhere, so we'll need to look out for name collisions and decide on how to resolve them.
 - Remove the `do_dss` bool and compute from the space
 - Struct-based changes. I might even suggest breaking this PR down into smaller parts.

Sbozzolo:
> Can we break this PR down into smaller, easily merge-able parts? My suggestion of some good peel off PRs would be:
> 
> * Remove `comms_ctx` from `p`, and get from the space instead.
> * Rename `param_set` to `params` (this will be pretty big on its own). Also, I think we use the name `params` for some other data structure elsewhere, so we'll need to look out for name collisions and decide on how to resolve them.
> * Remove the `do_dss` bool and compute from the space
> * Struct-based changes. I might even suggest breaking this PR down into smaller parts.

`param_set` was set to be exactly what we call `params` elsewhere (if I follow the chain correct), so I renamed to be more consistent across the codebase.

> Can we break this PR down into smaller, easily merge-able parts? 

Most of the commit should work as they are, and there's only a couple of commits that implement more than one change. So, it should be easy to peel off what we want. However, I first want to finalize the larger change (fixing the compilation problems)

Sbozzolo:
Closed in favor of #2287

PR #2246, opened 2023-10-17
Add patch version to some of the CliMA dependencies

Sbozzolo:


nefrathenrici:
Is there a reason to fix the compat to such specific versions? The package manager chooses the highest compatible version without the patch version.

Sbozzolo:
> Is there a reason to fix the compat to such specific versions? The package manager chooses the highest compatible version without the patch version.

Yes, see https://github.com/CliMA/ClimaComms.jl/pull/64.

In a few words, when you say `compat = "0.7.23"`, you mean "version in [0.7.23, 0.8.0)". You are not pinning a specific version, but you are guaranteeing a minimum version.

PR #2265, opened 2023-10-20
Fix Available Diagnostics docs table

nefrathenrici:
The [available diagnostics](https://clima.github.io/ClimaAtmos.jl/dev/available_diagnostics/) page is broken.

Main issue: multi-line strings do not work for the Documenter.jl markdown tables, this turns all comments into single line strings to fix the table.

Second issue: some of the standard names are so long that they effectively hide the description/comments column, since a user would have to scroll to see them.

I think we should switch to PrettyTables, since this provides more control over the formatting available diagnostics.


Sbozzolo:
Let me try to come up with a full solution. I don't want to have lines that super long in the source code.

Sbozzolo:
Documentar has this

image

So, we should try to make the table scrollable (and fix the multiline comments)

Sbozzolo (approved):
Thank you!

PR #2274, opened 2023-10-23
separate precomputed quantities in 3 funcs

trontrytel:
Following what we discussed @szy21 , @Sbozzolo  this PR splits prognostic edmfx precomputed quantities into 3 steps
1) updating environment
2) updating updraft thermo state and BC
3) updating the closures.

I'm not sure if it actually makes anything better though. Or if you want to split the individual closures.

### DIFF
--- a/src/cache/precomputed_quantities.jl
+++ b/src/cache/precomputed_quantities.jl
@@ -154,7 +154,7 @@ end
 """
     set_velocity_at_top!(Y, turbconv_model)
 
-Modifies `Y.f.u₃` so that `u₃` is 0 at the model top. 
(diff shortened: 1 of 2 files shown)


szy21 (approved):
Thanks! Should we do it for diagnostic edmf as well? (It doesn't have to be in this PR.)

trontrytel:
> Thanks! Should we do it for diagnostic edmf as well? (It doesn't have to be in this PR.)

Yes, I can do it in the next one. Just wasn't sure if this is actually helpful or not

Sbozzolo:
It's a step in the right direction. Thanks :)

PR #2278, opened 2023-10-24
refactor entrainment and detrainment

szy21:
## Purpose 

As discussed with Tapio, we should use some simple function forms to test entrainment and detrainment. To simplify the testing, this PR mainly combines different entrainment and detrainment into `GeneralizedEntrainment` and `GeneralizedDetrainment`. We can then test different entrainment and detrainment by using different parameters. The function forms are:
Entrainment: 1/timescale + c1 * w / z + minimum area limiter
Detrainment: 1/timescale + c2 * w + c3 * min(b, 0) / w + maximum area limiter

## To-do

- [x] Replace hard-coded numbers in area limiters with parameters after #2276 

## Content

----


szy21:
@trontrytel There is one problem when including the constant timescale entrainment or detrainment this way. As we are using 1 / timescale it will never be 0. Right now I am just setting the timescale to be very large. We can also change the timescale CLIMAParameters to 1 / timescale so it can be set to zero. What do you think?

src/prognostic_equations/edmfx_entr_detr.jl
```diff
@@ -103,10 +103,17 @@ function entrainment(
     ᶜRH⁰::FT,
     ᶜbuoy⁰::FT,
     dt::FT,
-    ::ConstantCoefficientEntrainment,
+    ::GeneralizedEntrainment,
 ) where {FT}
+    entr_tau = CAP.entr_tau(params)
     entr_coeff = CAP.entr_coeff(params)
-    entr = min(entr_coeff * abs(ᶜwʲ) / (ᶜz - z_sfc), 1 / dt)
+    min_area_limiter = FT(1e-3) * exp(-10 * ᶜaʲ)
```
szy21, line 110: I will change these, and other hard-coded numbers to parameters after #2276

PR #2282, opened 2023-10-24
Increase job memory for perf jobs and lower allocations

Sbozzolo:


charleskawczynski (approved):

charleskawczynski:
Let's just fix the memory request to start with, since that seems to be needed to fix CI (I'll update the PR since I'd like to try and get some other things merged)

PR #2283, opened 2023-10-24
Remove CUDA as direct dependency

charleskawczynski:
This PR removes CUDA as a direct dependency, since it's no longer needed. In order to instantiate environments with CUDA, I've added `ClimaAtmos.jl/cuda_env/Project.toml`, and added CUDA and ClimaCore (without a specified version) as a dependency. This way, it will sync with whatever version is specified in ClimaCore. I'm not sure if this is the best way to do this, so I'm open to other suggestions!


charleskawczynski:
I may peel off the last commit, since that's less config oriented.

charleskawczynski:
Supersedes #2264.

szy21:
This looks fine to me, but I don't understand julia packages enough to know if this is the best way. @sriharshakandala Could you take a look?

szy21 (approved):

charleskawczynski:
Ok, well, this seems better to me. So, if nobody else has any input I suppose we can move forward with it.

simonbyrne:
Seems like adding a whole new environment may have defeated the purpose here: it adds a lot of complexity...

PR #2286, opened 2023-10-24
Separate diagnostic edmf precomputed quants int 4 funcs

trontrytel:


src/cache/diagnostic_edmf_precomputed_quantities.jl
```diff
@@ -632,21 +642,45 @@ function set_diagnostic_edmf_precomputed_quantities!(Y, p, t)
...
+        S_q_totʲ_level = F.field_values(F.level(ᶜS_q_totʲ, i_top))
+        S_e_totʲ_helper_level = F.field_values(F.level(ᶜS_e_totʲ_helper, i_top))
         @. S_q_totʲ_level = 0
         @. S_e_totʲ_helper_level = 0
     end
+    return nothing
+end
+
+"""
+    set_diagnostic_edmf_precomputed_quantities!(Y, p, t)
```
szy21, line 663:
```suggestion
    set_diagnostic_edmf_precomputed_quantities_closures!(Y, p, t)
```

src/cache/diagnostic_edmf_precomputed_quantities.jl
```diff
@@ -632,21 +642,45 @@ function set_diagnostic_edmf_precomputed_quantities!(Y, p, t)
...
+
+Updates the precomputed quantities stored in `p` for diagnostic edmfx.
```
szy21, line 665:
```suggestion
Updates the environment closures in precomputed quantities stored in `p` for diagnostic edmfx.
```

src/cache/diagnostic_edmf_precomputed_quantities.jl
```diff
@@ -617,10 +615,22 @@ function set_diagnostic_edmf_precomputed_quantities!(Y, p, t)
             turbconv_model,
         )
     end
+    return nothing
+end
+
+"""
+    set_diagnostic_edmf_precomputed_quantities_top_bc!(Y, p, t)
+
+Updates the precomputed quantities stored in `p` for diagnostic edmfx.
```
szy21, line 624:
```suggestion
Updates the top boundary condition of precomputed quantities stored in `p` for diagnostic edmfx.
```

src/cache/diagnostic_edmf_precomputed_quantities.jl
```diff
@@ -632,21 +642,45 @@ function set_diagnostic_edmf_precomputed_quantities!(Y, p, t)
...
     end
+    return nothing
+end
+
+"""
+    set_diagnostic_edmf_precomputed_quantities!(Y, p, t)
+
+Updates the precomputed quantities stored in `p` for diagnostic edmfx.
+"""
+function set_diagnostic_edmf_precomputed_quantities_closures!(Y, p, t)
```
szy21, line 667: (Just want to note that it's more accurate to call it something like environment closures as the closures related to updraft, including entrainment, detrainment, nonhydrostatic pressure drag, and microphysics are in the integral. But I'm fine with the current name.)

szy21 (approved):
Thanks! I prefer to keep `Fields` to be consistent with the rest of the code base. But I'm fine with this change if @Sbozzolo thinks it's a good idea.

Sbozzolo:
> Thanks! I prefer to keep `Fields` to be consistent with the rest of the code base. But I'm fine with this change if @Sbozzolo thinks it's a good idea.

`Fields` would make search-and-replace operations easier, but I can see that it is highly redundant, so I don't have strong opinions.

PR #2281, opened 2023-10-24
Cache: clean up viscous and rayleigh

Sbozzolo:
- Establish the pattern: `cache(Y, atmos)`
- Clean up a couple of things

PR #2273, opened 2023-10-23
Move temporary variables in cache in `p.scratch`

Sbozzolo:
To be merged after https://github.com/CliMA/ClimaAtmos.jl/pull/2270

PR #2270, opened 2023-10-23
Add AtmosNumerics

Sbozzolo:
Group a few numeric choices into a single object.

PR #2259, opened 2023-10-19
Clean up: remove underused quantities

Sbozzolo:
Peel off #2244.

PR #2287, opened 2023-10-25
Clean more caches

Sbozzolo:
Clean up cache for:
- precipitation
- edmfx_nh_pressure
- subsidence
- EDMFCoriolis
- forcing_type
- non_orographic_gravity_wave
- orographic_gravity_wave
- hyperdiffiusion
- radiation
- limiters

Add surface_setup to AtmosModel and removes dt from simulation

A quick test seem to indicate that this PR reduces compile time by 5-10 % (or more in some cases)


Sbozzolo:
To be merged after #2281

src/parameterized_tendencies/gravity_wave_drag/orographic_gravity_wave.jl
```diff
@@ -16,23 +16,31 @@ Not yet included in our codebase
...
 
 orographic_gravity_wave_tendency!(Yₜ, Y, p, t, ::Nothing) = nothing
 
 include(joinpath(pkgdir(ClimaAtmos), "artifacts", "artifact_funcs.jl"))
 
-function orographic_gravity_wave_cache(ogw::OrographicGravityWave, Y, radius)
+function orographic_gravity_wave_cache(Y, ogw::OrographicGravityWave)
+
+    @assert Spaces.horizontal_space(axes(Y.c)).topology.mesh.domain isa
+            Domains.SphereDomain
```
charleskawczynski, line 31:
```suggestion
    @assert Spaces.horizontal_space(axes(Y.c)).topology.mesh.domain isa
            Domains.SphereDomain
```
is there a more direct accessor function we can use here?

Sbozzolo: We have `domain`, which takes in a `Mesh`, but it wouldn't be much better I think

Sbozzolo:
I take back the compile-time benefits. They are not uniform, and there's a regression for some EDMF cases (not all).

I am investigating this.

examples/hybrid/driver.jl
```diff
@@ -117,7 +117,8 @@ end
 if config.parsed_args["check_conservation"]
     FT = Spaces.undertype(axes(sol.u[end].c.ρ))
     @test sum(sol.u[1].c.ρ) ≈ sum(sol.u[end].c.ρ) rtol = 50 * eps(FT)
-    @test sum(sol.u[1].c.ρe_tot) +
-          (p.net_energy_flux_sfc[][] - p.net_energy_flux_toa[][]) ≈
-          sum(sol.u[end].c.ρe_tot) rtol = 100 * eps(FT)
+    @test sum(sol.u[1].c.ρe_tot) + (
+        p.radiation_mode.net_energy_flux_sfc[][] -
+        p.radiation_mode.net_energy_flux_toa[][]
```
szy21, line 122: Would this cause a problem when running simulations without radiation? In any case, I think it would be better to put `net_energy_flux_sfc` and `net_energy_flux_toa` in the default cache, as they are used for the total energy flux at the surface and TOA. Currently, we only include radiative fluxes, but we should still have these two variables when running simulations without radiation.

Sbozzolo: Okay, I thought that they were needed only for radiation. I'll revert this.

src/parameterized_tendencies/gravity_wave_drag/orographic_gravity_wave.jl
charleskawczynski: Ah, bummer. We should probably add something to ClimaCore to hide internal structure a bit more.

szy21:
@dennisYatunin may want to take a look at the changes to radiation cache.

src/diagnostics/radiation_diagnostics.jl
```diff
@@ -58,12 +58,12 @@ function compute_rsu!(
     FT = eltype(state)
     if isnothing(out)
         return RRTMGPI.array2field(
-            FT.(cache.radiation_model.face_sw_flux_up),
+            FT.(cache.atmos.radiation_mode.face_sw_flux_up),
```
szy21, line 61:
```suggestion
            FT.(cache.radiation.radiation_model.face_sw_flux_up),
```
And for all the changes below. The change here should be the same as the one in line 23?

src/parameterized_tendencies/radiation/radiation.jl
```diff
@@ -239,16 +241,17 @@ end
...
     NT = NamedTuple{(:z, :ρ, :q_tot), NTuple{3, FT}}
     return (;
-        radiation_model = radiation_mode,
         ᶜκρq = similar(Y.c, FT),
         ∫_0_∞_κρq = similar(Spaces.level(Y.c, 1), FT),
         ᶠ∫_0_z_κρq = similar(Y.f, FT),
         isoline_z_ρ_q = similar(Spaces.level(Y.c, 1), NT),
         ᶠradiation_flux = similar(Y.f, Geometry.WVector{FT}),
+        net_energy_flux_toa = [Geometry.WVector(FT(0))],
+        net_energy_flux_sfc = [Geometry.WVector(FT(0))],
```
szy21, line 254:
```suggestion
```

src/parameterized_tendencies/radiation/radiation.jl
```diff
@@ -327,11 +331,15 @@ end
...
 
-function radiation_model_cache(Y, params, radiation_mode::RadiationTRMM_LBA;)
+function radiation_model_cache(Y, radiation_mode::RadiationTRMM_LBA)
     FT = Spaces.undertype(axes(Y.c))
-    # TODO: should we distinguish between radiation mode vs model?
-    return (; ᶜdTdt_rad = similar(Y.c, FT), radiation_model = radiation_mode)
+    return (;
+        ᶜdTdt_rad = similar(Y.c, FT),
+        net_energy_flux_toa = [Geometry.WVector(FT(0))],
+        net_energy_flux_sfc = [Geometry.WVector(FT(0))],
```
szy21, line 339:
```suggestion
```

Sbozzolo:
> @dennisYatunin may want to take a look at the changes to radiation cache.

We agreed to remove one level of nesting to the RRTMPGI cache and rename `radiation_mode` to `radiation_model` in a future PR.

szy21 (approved):
Thanks!

Sbozzolo:
I thought I identified and reverted the issue for the slowdown, but that's not the case. 

While on my computer, the PR leads to consistent performance/latency improvements, this is not the case on the cluster. Quite the opposite, this PR introduces some noticeable regressions. 

I would like to still merge this (and the other, #2296) PR. They clean up and improve the cache, and they are a lot of work to maintain because pretty much every change touches the cache, leading to conflicts that are time consuming to solve. 

Debugging this issue is particularly annoying because seemingly there's no correlation between my changed and the outcome (+ I need to run stuff on the cluster to clearly see the effects). I suspect that some of the newly introduce (and non identified) problems will be automatically solved by the mutable spaces. 

What do you think?
@szy21 , @charleskawczynski , @dennisYatunin

szy21:
How much regression did you get on the cluster? If it's not too bad (e.g. ci can still finish in a reasonable time), then I think it's ok to merge it in.

Sbozzolo:
You can compare 

https://buildkite.com/clima/climaatmos-ci/builds/14418#_  (which would be the "final" PR)

With master:

https://buildkite.com/clima/climaatmos-ci/builds/14401#_

Some jobs take 10 more minutes:

https://buildkite.com/clima/climaatmos-ci/builds/14418#018b6e11-f634-4293-b119-d2b0d8b59bac

vs

https://buildkite.com/clima/climaatmos-ci/builds/14401#018b6d70-d498-49e9-a8cf-8a3eab19e45c

I think that this is all in compiling `solve_atmos!`. If you look at everything else (including SYPD), the new PR is better.

szy21:
hmm, indeed a lot of jobs are slower. @charleskawczynski What do you think?

Sbozzolo:
Maybe we can start by merging this PR, and wait until we have the mutable spaces to merge #2296 ?

This current PR does introduce some additional costs, but nowhere near as much as #2296.

(And I do think that most of the additional will go away with the mutable spaces)

charleskawczynski:
My 2 cents is: break this PR down. Two advantages are: 1) it's much easier to review smaller PRs, and reviews tend to be more thorough (I can't admit that I would review this PR carefully). and 2) we'll actually learn _what_ is causing the regression, rather than sweeping under the rug.

Sbozzolo:
Each commit is self sufficient, working, and implement just one change. You can go through them one by one.  I tried having small PRs with #2273, #2270, #2281, but it took days to get these very small changes merged, with a lot of time on my side spent rebasing every time `main` changed. 

This PR implements one single logical change in multiple commits.

PR #2296 consists of only 2 commits (on top of all these ones). I know that the slow down is in those two, but even with that, I cannot fix it.

szy21:
Would we be able to know which commit causes the regression?

charleskawczynski:
> but it took days to get these very small changes merged

I think that time is dominated by time to review. (e.g., https://github.com/CliMA/ClimaAtmos.jl/pull/2283 is pretty small and still waiting on review).

Sbozzolo:
> Would we be able to know which commit causes the regression?

Yes, I'll find it.

> > but it took days to get these very small changes merged
> 
> I think that time is dominated by time to review. (e.g., #2283 is pretty small and still waiting on review).

I cannot turn this PR into several concurrent PRs because they all change the `cache.jl` file (at least), leading to conflict that I'd have to solve for each of them every time one gets merged. So, to avoid internal conflicts, I'd have to submit multiple serial PRs. In that case, I'd have to solve for the conflict that arise from all the other PRs that get merged (the cache is used everywhere in the code, so the probability of conflicts is very high). I am not saying that it cannot be done, 
it just feels slow and inefficient when I know I have all these changes ready. 

#2259 is an example of fast turnaround time: 12 minutes from commit to review, but almost 3 hours from commit to merge.

However, I absolutely agree with your comment about ease of review. For that, I'd ask you to look at each commit as if it was a different PR. I take a lot of care to constantly rebase the code to have good and atomic commits.

(I am not reviewing #2283 because I don't understand its implications so I cannot make meaningful comments)

src/prognostic_equations/limited_tendencies.jl
```diff
@@ -1,32 +1,23 @@
...
-NVTX.@annotate function limiters_func!(Y, p, t, ref_Y)
-    (; limiter) = p
-    n = n_mass_flux_subdomains(p.atmos.turbconv_model)
-    if !isnothing(limiter)
+function generate_limiters_func!(limiter::Nothing)
+    limiters_func! = (Y, p, t, ref_Y) -> nothing
+    return limiters_func!
+end
+
+function generate_limiters_func!(limiter)
```
charleskawczynski, line 13: I prefer we avoid this closure. What is this needed for?

perf/flame.jl
```diff
@@ -66,7 +66,7 @@ allocs_limit["flame_perf_diagnostics"] = 3016328
 allocs_limit["flame_perf_target_diagnostic_edmfx"] = 920960
 allocs_limit["flame_perf_target_threaded"] = 5857808
 allocs_limit["flame_perf_target_callbacks"] = 46407936
-allocs_limit["flame_perf_gw"] = 4844555632
+allocs_limit["flame_perf_gw"] = 4865411440
```
charleskawczynski, line 69: Can we fix this allocation so that it's not increasing?

src/prognostic_equations/limited_tendencies.jl
Sbozzolo: `limiters_func!` has a specific signature. If I want to dispatch over types of limiters (instead of checking the boolean `apply_limiter`, I need to return it as a function. This allows me to remove `limiter` from the cache, as `limiter` is used exclusively here.

Sbozzolo:
This buildkite is fast: https://buildkite.com/clima/climaatmos-ci/builds/14450

Leaving only one commit as culprit: d8ff6796

I am checking that that's indeed the case.

src/prognostic_equations/limited_tendencies.jl
charleskawczynski: Couldn't we instead change `limiters_func!(Y, p, t, ref_Y)` to `limiters_func!(::QuasiMonotoneLimiter, Y, p, t, ref_Y)` and define
```julia
limiters_func!(Y, p, t, ref_Y) = limiters_func!(p.limiter, Y, p, t, ref_Y)
limiters_func!(::Nothing, args...) = nothing
```
?
This would require fewer changes across the codebase, too.

charleskawczynski: And why do we need to remove limiter from the cache?

Sbozzolo: > And why do we need to remove limiter from the cache?

Because the limiter is used exclusively in `limiters_func!`, so if we want to reduce the variables we carry around in the cache, this is a good place to do that. But we need to capture `limiter` to do that.

Sbozzolo:
Yes, in this PR https://github.com/CliMA/ClimaAtmos.jl/commit/d8ff67968dd2cfedb81950df30e1b23c3fba50e0 leads to an increase in compile time.

I reverted that change.

src/prognostic_equations/limited_tendencies.jl
charleskawczynski: We can still reduce variables in the cache without removing this one. I'd prefer we leave this in the cache-- one disadvantage of doing this is that it becomes difficult to interactively inspect the `QuasiMonotoneLimiter` object (which can be helpful for debugging), since it only exists in the anonymous struct generated by the closure.

charleskawczynski: @kmdeck, is this something that you've found in ClimaLSM? (I think this pattern may be pretty heavily used there)

Sbozzolo: Okay, I don't have strong feelings about this.

Sbozzolo: Rebased and fixed this. I put the limiter inside `p.numerics`.

src/diagnostics/radiation_diagnostics.jl
```diff
@@ -96,12 +96,12 @@ function compute_rld!(
     FT = eltype(state)
     if isnothing(out)
         return RRTMGPI.array2field(
-            FT.(cache.radiation_model.face_lw_flux_dn),
+            FT.(cache.atmos.radiation.radiation_model.face_lw_flux_dn),
```
szy21, line 99: Is `cache.atmos.radiation` the same as `cache.radiation`? (I was wondering why this line is different from e.g. line23)

On a side note, we didn't test all the diagnostics in ci, so it would be hard to catch things like this.

szy21:
It seems the compile time is back to normal on the cluster now. Feel free to merge it once others' comments are addressed.

Sbozzolo:
In https://github.com/CliMA/ClimaAtmos.jl/pull/2244/commits I add [one additional commit](https://github.com/CliMA/ClimaAtmos.jl/pull/2244/commits/58194fd6fed20f627ead5763af07d1b3b13856e7)  to this PR to complete the refactor of the cache by moving all the precomputed quantities in `p.precomputed`. Overall compile time is nearly identical to the master branch (https://buildkite.com/clima/climaatmos-ci/builds/14519). I would like to merge that and close the chapter of "cache refactoring".

@szy21 do you want to have a look at [that commit](https://github.com/CliMA/ClimaAtmos.jl/pull/2244/commits/58194fd6fed20f627ead5763af07d1b3b13856e7) too? If it the change is okay, I will move it over to this PR and merge it.

szy21:
The commit looks good to me, thanks! I need to think a bit more about what should be in `core` and what should be in `precomputed`, but we don't need to wait for that to merge this PR.

src/prognostic_equations/limited_tendencies.jl
kmdeck: @charleskawczynski let me know if I am not answering your question...

we use closures like this a lot:
```
function make_update_aux(model::MyModelType)
    function update_aux!(p,Y,t)
        #update elements in p in place based on type of `model` (simple dispatch) but also
        # using parameters, etc that are stored in `model` (closure), Y, p
    end
    return update_aux!
end
```

I agree that they are kind of annoying to write tests for because you need to make the `model` in order to test it. but we also need to contruct Y and p with the necessary structure to debug/write tests for, so the closure isnt the only annoying part.

kmdeck: and, I guess the alternative would be to pass `model` as an argument, which also doesnt get around having to make a `MyModelType` instance

PR #2296, opened 2023-10-26
Turn cache into struct

### LINKED ISSUE #2217: Replace cache namedtuple with explicit struct

Sbozzolo:
Closes #2217

To be merged after #2287

### DIFF
--- a/src/cache/cache.jl
+++ b/src/cache/cache.jl
@@ -1,3 +1,85 @@
+struct AtmosCache{
+    FT <: AbstractFloat,
+    SIM,
+    AM,
+    NUM,
+    CAP,
+    COR,
+    SFC,
+    GHOST,
+    ENV,
+    PREC,
+    SCRA,
+    HYPE,
+    DSS,
+    RS,
+    VS,
+    PR,
+    SUB,
+    LSAD,
+    EDMFCOR,
+    FOR,
+    NONGW,
+    ORGW,
+    RAD,
+    NETFLUXTOA,
+    NETFLUXSFC,
+}
+    """Timestep of the simulation (in seconds). This is also used by callbacks and tendencies"""
+    dt::FT
+
+    """NamedTuple with job_id, output_dir (used in post), and start_date (used for insolation)."""
+    simulation::SIM
+
+    """AtmosModel"""
+    atmos::AM
+
+    """Limiter"""
+    numerics::NUM
+
+    """ClimaAtmosParameters that have to be used"""
+    params::CAP
+
+    """Variables that are used generally, such as ᶜρ_ref, ᶜΦ"""
+    core::COR
+
+    """Used by update_surface_conditions! in set_precomputed_quantities! and coupler"""
+    sfc_setup::SFC
+
+    """Center and face ghost buffers used by DSS"""
+    ghost_buffer::GHOST
+
+    env_thermo_quad::ENV
+
+    """Quantities that are updated with set_precomputed_quantities!"""
+    precomputed::PREC
+
+    """Pre-allocated areas of memory to store temporary values"""
+    scratch::SCRA
+
+    """Hyperdiffision quantities for grid and subgrid scale quantities, potentially with
+       ghost buffers for DSS"""
+    hyperdiff::HYPE
+
+    do_dss::DSS
+
+    """Additional parameters used by the various tendencies"""
+    rayleigh_sponge::RS
+    viscous_sponge::VS
+    precipitation::PR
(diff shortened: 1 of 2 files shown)


Sbozzolo:
For some reason, this PR slows down compilation time a lot:

https://buildkite.com/clima/climaatmos-ci/builds/15102

Sbozzolo:
Found an incantation that is more reasonable, still a little bit slower than without struct (but part of it is probably just the `do_dss`):

https://buildkite.com/clima/climaatmos-ci/builds/15104

src/cache/cache.jl
```diff
@@ -1,3 +1,86 @@
+Base.@kwdef struct AtmosCache{
```
sriharshakandala, line 1: Is there a reason we need `Base.@kwdef` here? i.e. do we have default values for most fields? 
Otherwise a normal `struct` should suffice!

sriharshakandala (approved):

src/cache/cache.jl
Sbozzolo: I used @kwdef because the order of the parameters is arbitrary and I didn't want users to have to strictly follow the order. In the `coupler_compatibility` test I use use keyword constructor.

sriharshakandala: I do not see a specific advantage with this in terms of user convenience! 
We still need to specify (and remember to specify) all fields anyways! 
And a keyword constructor is convenient if we can specify a few and use defaults for others!

Sbozzolo: It is not required for compatibility. I just saw it as a way to "forget about the order" of arguments. Without this constructor, we would have to provide the arguments in the correct order, and since there is no given order, it felt natural to make them keyword.

I don't have strong feelings about this and I can change it to positional only if you prefer.

Issue #2294, opened 2023-10-26
Test buoyancy terms in prognostic edmf


szy21:
The updraft buoyancy we use in momentum equation, detrainment, and pressure drag for prognostic edmf are all different. We should check what is more physical and/or gives better results.

PR #2298, opened 2023-10-26
Change buoyancy term in prognostic edmf

szy21:
## Purpose 

Tests #2294 

Currently, buoyancy of a subdomain j is defined as -(\rho^j - \rho_ref) / \rho^j * g. And the buoyancy difference between updraft and environment is used in detrainment and entrainment closures. This is not ideal as buoyancy of updraft and environment will both be -g if we don't use a reference state (\rho_ref = 0). I changed the updraft buoyancy to be -(\rho^j - \rho_gm) / \rho^j * g. I set environment buoyancy to zero for now to keep the function structure of entrainment and detrainment closures the same. To be consistent, I use grid-mean velocity in the b/w detrainment, i.e. b^j / (w^j - w_gm), where b^j uses the new buoyancy definition.

I also changed the buoyancy in updraft u_3 equation to -(\rho^j - \rho_gm) / \rho^j * \grad \Phi. I'm less sure about this. I prefer this form as it would be easier for me to understand the tendencies of updraft u_3, but this is not as accurate as what we had before. I don't have a strong preference for this.

I hope this change, together with #2304, will make it easier to figure out why edmf is breaking.

## Content

----


tapios (approved):
LGTM. But can you please add a docstring documenting what is going on here (rather than just the TODO).

PR #2299, opened 2023-10-26
New CloudMicrophysics parameters interface

trontrytel:
This PR:
- Tests the new `CM.jl` parameters interface (currently on the main branch. We should have the release once we are confident that things for for `CA.jl`)
- removes the obsolete `TurbulenceConvection/Parameters.jl` file


src/parameterized_tendencies/microphysics/precipitation.jl
```diff
@@ -146,122 +145,122 @@ end
...
-            error(
-                "1-moment microphysics can only be coupled to ρe_tot energy variable",
-            )
-        end
-    end
-
-    precipitation_advection_tendency!(Yₜ, Y, p, colidx, precip_model)
-    return nothing
-end
+#function precipitation_cache(Y, precip_model::Microphysics1Moment)
```
trontrytel, line 148: I don't think we use it anywhere. I would prefer to first update things properly for the 0-moment microphysics and the code up the 1-moment microphysics

src/parameters/create_parameters.jl
```diff
@@ -36,10 +26,12 @@ function create_parameter_set(config::AtmosConfig)
         create_parameter_struct(TD.Parameters.ThermodynamicsParameters)
     rrtmgp_params = create_parameter_struct(RP.RRTMGPParameters)
     insolation_params = create_parameter_struct(IP.InsolationParameters)
-    microphys_params = create_parameter_struct(
-        CM.Parameters.CloudMicrophysicsParameters;
-        subparam_structs = (; thermo_params),
-    )
+
+    toml_dict["τ_precip"]["value"] = FT(CA.time_to_seconds(parsed_args["dt"]))
```
trontrytel, line 30: Probably a temporary solution? The alternative is for me to make the CloudMicrophysics structs be `kwdef`

src/parameters/Parameters.jl
```diff
@@ -4,17 +4,46 @@ import ClimaCore as CC
...
 abstract type AbstractClimaAtmosParameters end
 const ACAP = AbstractClimaAtmosParameters
 
+abstract type AbstractTurbulenceConvectionParameters end
+const ATCP = AbstractTurbulenceConvectionParameters
+
 Base.broadcastable(param_set::ACAP) = tuple(param_set)
+Base.broadcastable(param_set::ATCP) = tuple(param_set)
+
+Base.@kwdef struct TurbulenceConvectionParameters{FT} <: ATCP
```
trontrytel, line 18: Not sure if we want to keep edmf parameters in a separate struct or have them be part of ClimaAtmos parameters below. Right now things are mixed up a little. I can clean it up in a separate PR, once we decide @szy21

trontrytel: At least right now we don't put `microphysics` parameters inside `turbconv` parameters so that we can create `thermodynamics` parameters...

szy21: I think it would be good to have edmf parameters in a separate struct. We can do it after this.

src/parameters/create_parameters.jl
sriharshakandala: As we discussed, using `@kwdef` does not seem logical when default values are not being set! Updating the value for the dict key looks like the more natural thing to do!

src/parameters/Parameters.jl
trontrytel: Actually, I think I'll do it before. Stay tuned ;)

PR #2301, opened 2023-10-27
Add barowave jfnk flame graph

charleskawczynski:
This PR adds a flame graph for the baro wave JFNK, which is currently experiencing a lot of allocations and inference issues.


charleskawczynski:
Merging this will make it easier to try out potential fixes.

PR #2302, opened 2023-10-27
Delete old TC params struct

trontrytel:
This PR:
- Deletes the old EDMF parameter struct
- Creates a new parameter struct for `TurbulenceConvectionParameters` parameters. This struct is flat and does not store any thermodynamics or microphysics parameters.
- Moves all the EDMF relevant parameters out of `ClimaAtmosParameters` and into `TurbulenceConvectionParameters`

@szy21 - could you double check if I moved all the edmf parameters?
@nefrathenrici - does that look reasonable?


src/parameters/Parameters.jl
```diff
@@ -51,6 +75,15 @@ end
 
 Base.eltype(::ClimaAtmosParameters{FT}) where {FT} = FT
 
+thermodynamics_params(ps::ATCP) = ps.thermodynamics_params
+surface_fluxes_params(ps::ATCP) = ps.surface_fluxes_params
```
nefrathenrici, line 79:
```suggestion
```

I don't think these are needed, they should auto-generated by this snippet below
```
for var in fieldnames(ClimaAtmosParameters)
    @eval $var(ps::ACAP) = ps.$var
end
```

src/parameters/Parameters.jl
```diff
@@ -51,6 +75,15 @@ end
...
+
+# Forward TurbulenceConvection parameters
+const TCPS = TurbulenceConvectionParameters
+for fn in fieldnames(TCPS)
+    @eval $(fn)(ps::ATCP) = ps.$(fn)
+end
```
nefrathenrici, line 85: Could we add parameter forwarding for `turbconv_params` like we do for the thermo params?
Then we can do
`detr_coeff = CAP.detr_coeff(params)` instead of 
```
turbconv_params = CAP.turbconv_params(params)
`detr_coeff = CAP.detr_coeff(turbconv_params)`
```

nefrathenrici (approved):
Looks good! It seems like some TC fragments are still in the docs.

src/parameters/Parameters.jl
```diff
@@ -4,11 +4,45 @@ import ClimaCore as CC
...
+    entr_tau::FT
+    entr_coeff::FT
+    detr_tau::FT
+    detr_coeff::FT
+    detr_buoy_coeff::FT
+    min_area_limiter_scale::FT
+    min_area_limiter_power::FT
+    max_area_limiter_scale::FT
+    max_area_limiter_power::FT
+    c_smag::FT
```
szy21, line 43: We can keep `c_smag` in ClimaAtmosParameters. This is not EDMF specific, and will be used for LES in the future.

src/parameters/Parameters.jl
```diff
@@ -4,11 +4,45 @@ import ClimaCore as CC
...
+    updraft_number::Int
+end
 
 Base.@kwdef struct ClimaAtmosParameters{FT, TP, RP, IP, MPP, SFP, TCP} <: ACAP
```
szy21, line 47: Should we remove TCP from this struct?

szy21 (approved):
Thanks!

sriharshakandala (approved):

sriharshakandala:
Please fix docs issues and rebase (as needed). Thanks.

src/parameters/Parameters.jl
trontrytel: I think it should stay there. It's just now it's for the `TurbulenceConvectionParameters`struct I create above. I still make it be part of `ClimaAtmosParameters` the same as the thermo and surface fluxes parameters, etc.

src/parameters/Parameters.jl
trontrytel: Done!

src/parameters/Parameters.jl
nefrathenrici:
```
for var in fieldnames(TCPS)
    @eval $var(ps::ACAP) = $var(turbconv_params(ps))
 ```

PR #2305, opened 2023-10-28
limix mixing length with smagorinsky length scale

szy21:
## Purpose 

Uses smagorinsky length scale as lower limit of mixing length.

## Content

----


szy21:
superseded by #2413

PR #2337, opened 2023-11-06
change edmfx sgs mass flux to advect a*tracer

### LINKED ISSUE #2335: Change the numerical scheme in sgs mass flux
We would like to advect a*tracer instead of tracer in sgs mass flux, and add an option for unwinding or central.

szy21:
## Purpose 

Also adds an option for the numerical scheme for edmfx sgs mass flux. (Note: This PR is only changing the numerics of sgs mass flux. I will have another PR soon that adds a divergence term in GeneralizedDetrainment, adds a few more cases, and maybe changes the default detrainment for all the cases.)

Closes #2335 

## Content

----

### DIFF
--- a/config/default_configs/default_edmf_config.yml
+++ b/config/default_configs/default_edmf_config.yml
@@ -18,7 +18,10 @@ edmfx_detr_model:
   help: "EDMFX detrainment closure.  [`nothing` (default), `PiGroups`, `Generalized`, `GeneralizedHarmonics`]"
   value: ~
 edmfx_upwinding:
-  help: "EDMFX upwinding mode [`none` (default), `first_order` , `third_order`, `boris_book`, `zalesak`]"
+  help: "EDMFX upwinding mode [`none` (default), `first_order`, `third_order`]"
+  value: none
+edmfx_sgsflux_upwinding:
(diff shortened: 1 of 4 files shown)


config/default_configs/default_edmf_config.yml
```diff
@@ -18,7 +18,10 @@ edmfx_detr_model:
   help: "EDMFX detrainment closure.  [`nothing` (default), `PiGroups`, `Generalized`, `GeneralizedHarmonics`]"
   value: ~
 edmfx_upwinding:
-  help: "EDMFX upwinding mode [`none` (default), `first_order` , `third_order`, `boris_book`, `zalesak`]"
+  help: "EDMFX upwinding mode [`none` (default), `first_order`, `third_order`, `boris_book`, `zalesak`]"
+  value: none
+edmfx_sgsflux_upwinding:
+  help: "EDMFX SGS mass flux upwinding mode [`none` (default), `first_order`, `third_order`, `boris_book`, `zalesak`]"
```
trontrytel, line 24: I think the `zalesak` didn't compile for me what I tried to use it in EDMF. Maybe we should just suggest the choice between 1st and 3rd order?

trontrytel (approved):

PR #1668, opened 2023-05-15
Move from bors to GitHub Merge Queue

charleskawczynski:
This PR removes bors, as it is being deprecated, and applies a small fix in preparation of using the GitHub Merge Queue.

Closes #1667.

PR #2338, opened 2023-11-07
Update github workflow for merge queue

szy21:
## Purpose 

Follows #1668 

## Content

----


simonbyrne:
Don't need to build on `main` either (since it will get built on the queue)

PR #2348, opened 2023-11-09
use MSE as a prognostic variable for prognostic edmf

### LINKED ISSUE #2251: Test different energy variable in prognostic EDMF
Currently, prognostic EDMF uses h_tot as prognostic variable. We want to try MSE or theta_li

szy21:
## Purpose 

Change subdomain prognostic variable from h_tot to MSE. I think this is mostly ready. One thing I want to check is whether we can remove some caches (e.g. only save mse and calculate h_tot on the fly). I may also need to fix allocations for some cases.

Closes #2251 

## To-do

- [ ] Try removing h_tot in the cache
- [x] Fix allocations

## Content

----

### DIFF
--- a/src/prognostic_equations/advection.jl
+++ b/src/prognostic_equations/advection.jl
@@ -33,9 +33,9 @@ NVTX.@annotate function horizontal_advection_tendency!(Yₜ, Y, p, t)
 
     if p.atmos.turbconv_model isa PrognosticEDMFX
         for j in 1:n
-            @. Yₜ.c.sgsʲs.:($$j).h_tot -=
-                wdivₕ(Y.c.sgsʲs.:($$j).h_tot * ᶜuʲs.:($$j)) -
-                Y.c.sgsʲs.:($$j).h_tot * wdivₕ(ᶜuʲs.:($$j))
+            @. Yₜ.c.sgsʲs.:($$j).mse -=
+                wdivₕ(Y.c.sgsʲs.:($$j).mse * ᶜuʲs.:($$j)) -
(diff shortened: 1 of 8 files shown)


src/prognostic_equations/advection.jl
```diff
@@ -132,37 +138,41 @@ NVTX.@annotate function explicit_vertical_advection_tendency!(Yₜ, Y, p, t)
...
-                )
-
-                vertical_advection!(
-                    Yₜ.c.sgsʲs.:($j).q_tot[colidx],
-                    ᶠu³ʲs.:($j)[colidx],
-                    Y.c.sgsʲs.:($j).q_tot[colidx],
-                    edmfx_upwinding,
-                )
-            end
+        for j in 1:n
```
szy21, line 150: The if-end block is not needed as n is zero if turbconv_model is not PrognosticEDMFX

src/prognostic_equations/advection.jl
```diff
@@ -132,37 +138,41 @@ NVTX.@annotate function explicit_vertical_advection_tendency!(Yₜ, Y, p, t)
                     ᶠinterp(ᶜρʲs.:($$j)[colidx] - Y.c.ρ[colidx]) *
                     ᶠgradᵥ_ᶜΦ[colidx]
                 ) / ᶠinterp(ᶜρʲs.:($$j)[colidx])
+
+            # buoyancy term in mse equation
+            @. Yₜ.c.sgsʲs.:($$j).mse[colidx] +=
```
szy21, line 143: We can talk about what is the best way for interpolations in this equation.

szy21:
I removed draft and environment h_tot from the cache.

trontrytel (approved):

PR #2349, opened 2023-11-09
Stability dependent boundary-layer diffusion profile 

### LINKED ISSUE #2332: Add a more realistic vertical diffusion
Currently, the vertical diffusion assumes a fixed boundary layer depth and does not depend on the stability of the surface layer. This is unphysical and may lead to unphysical atmospheric dynamics. We should implement something like [Frierson 2007](https://journals.ametsoc.org/view/journals/atsc/63/10/jas3753.1.xml).

akshaysridhar:
Closes #2332 

Corresponding issue in ClimaCoupler: 
https://github.com/CliMA/ClimaCoupler.jl/issues/485

### DIFF
--- a/src/cache/precomputed_quantities.jl
+++ b/src/cache/precomputed_quantities.jl
@@ -111,6 +111,9 @@ function precomputed_quantities(Y, atmos)
     vert_diff_quantities = if atmos.vert_diff isa VerticalDiffusion
         ᶜK_h = similar(Y.c, FT)
         (; ᶜK_u = ᶜK_h, ᶜK_h) # ᶜK_u aliases ᶜK_h because they are always equal.
+    elseif atmos.vert_diff isa FriersonDiffusion
+        ᶜK_h = similar(Y.c, FT)
+        (; ᶜK_u = ᶜK_h, ᶜK_h) # ᶜK_u aliases ᶜK_h because they are always equal.
     else
         (;)
     end
@@ -285,6 +288,110 @@ function eddy_diffusivity_coefficient(C_E, norm_v_a, z_a, p)
     K_E = C_E * norm_v_a * z_a
     return p > p_pbl ? K_E : K_E * exp(-((p_pbl - p) / p_strato)^2)
 end
+function eddy_diffusivity_coefficient(
+    z::FT,
+    z₀,
+    f_b::FT,
+    h::FT,
+    uₐ,
+    C_E::FT,
+    Ri::FT,
+    Ri_a::FT,
+    Ri_c::FT,
+    κ::FT,
+) where {FT}
+    # Equations (17), (18)
+    if z < f_b * h
+        K_b =
+            compute_surface_layer_diffusivity(z, z₀, κ, C_E, Ri, Ri_a, Ri_c, uₐ)
+        return K_b
+    elseif f_b * h < z < h
+        K_b = compute_surface_layer_diffusivity(
+            f_b * h,
+            z₀,
+            κ,
+            C_E,
+            Ri,
+            Ri_a,
+            Ri_c,
+            uₐ,
+        )
+        K = K_b * (z / f_b / h) * (1 - (z - f_b * h) / (1 - f_b) / h)^2
+        return K
+    else
+        return FT(0)
+    end
+end
+
+### Frierson (2006) diffusion
+function compute_boundary_layer_height!(
+    h_boundary_layer,
+    f_b::FT,
+    dz,
+    Ri_local,
+    Ri_c::FT,
+    Ri_a,
+) where {FT}
+    Fields.bycolumn(axes(Ri_local)) do colidx
+        @inbounds for il in 1:Spaces.nlevels(axes(Ri_local[colidx]))
+            h_boundary_layer[colidx] .=
+                ifelse.(
+                    Fields.Field(
+                        Fields.field_values(Fields.level(Ri_local[colidx], il)),
+                        axes(h_boundary_layer[colidx]),
+                    ) .< Ri_c,
+                    Fields.Field(
+                        Fields.field_values(Fields.level(dz[colidx], il)),
+                        axes(h_boundary_layer[colidx]),
+                    ),
+                    h_boundary_layer[colidx],
+                )
+        end
+    end
+end
+
+function compute_bulk_richardson_number(
+    θ_v,
+    θ_v_a,
+    norm_ua,
+    grav,
+    z::FT,
+) where {FT}
+    return (grav * z) * (θ_v - θ_v_a) / (θ_v_a * (norm_ua)^2 + FT(1))
+end
+function compute_exchange_coefficient(Ri_a, Ri_c, zₐ, z₀, κ::FT) where {FT}
+    # Equations (12), (13), (14)
+    if Ri_a < FT(0)
+        return κ^2 * (log(zₐ / z₀))^(-2)
+    elseif FT(0) < Ri_a < Ri_c
+        return κ^2 * (log(zₐ / z₀))^(-2) * (1 - Ri_a / Ri_c)^2
+    else
+        return FT(0)
+    end
+end
+
+function compute_surface_layer_diffusivity(
+    z::FT,
+    z₀::FT,
+    κ::FT,
+    C_E::FT,
+    Ri::FT,
+    Ri_a::FT,
+    Ri_c::FT,
+    norm_uₐ,
+) where {FT}
+    # Equations (19), (20)
+    if Ri_a < FT(0)
+        return κ * norm_uₐ * sqrt(C_E) * z
+    else
+        return κ *
+               norm_uₐ *
+               sqrt(C_E) *
+               z *
+               (1 + Ri / Ri_c * (log(z / z₀) / (1 - Ri / Ri_c)))^(-1)
+    end
+end
+###
 
 """
     set_precomputed_quantities!(Y, p, t)
@@ -373,6 +480,89 @@ NVTX.@annotate function set_precomputed_quantities!(Y, p, t)
             norm(interior_uₕ),
             ᶜΔz_surface / 2,
             ᶜp,
+        )
+    elseif vert_diff isa FriersonDiffusion
+        (; ᶜK_h, sfc_conditions, ᶜts) = p.precomputed
+        (; params) = p
+        interior_uₕ = Fields.level(Y.c.uₕ, 1)
+        κ = CAP.von_karman_const(params)
+        grav = CAP.grav(params)
+        FT = Spaces.undertype(axes(ᶜK_h))
+        z₀ = FT(1e-5)
+        Ri_c = FT(1.0)
+        f_b = FT(0.10)
+
+        # Prepare scratch vars
+        ᶠρK_E = p.scratch.ᶠtemp_scalar
+        θ_v = p.scratch.ᶜtemp_scalar
+        Ri = p.scratch.ᶜtemp_scalar_2
+        dz_local = p.scratch.ᶜtemp_scalar_3
+        θ_v_sfc = p.scratch.ᶠtemp_field_level
+        Ri_a = p.scratch.temp_field_level
+        z_local = p.scratch.temp_data
+        z_sfc = p.scratch.temp_data_face_level
+        ᶜθ_v_sfc = C_E = p.scratch.temp_field_level_2
+        h_boundary_layer = p.scratch.temp_field_level_3
+        ᶠts_sfc = sfc_conditions.ts
+        ᶜz = Fields.coordinate_field(Y.c).z
+        interior_uₕ = Fields.level(Y.c.uₕ, 1)
+        ᶜΔz_surface = Fields.Δz_field(interior_uₕ)
+        @. θ_v = TD.virtual_pottemp(thermo_params, ᶜts)
+        @. θ_v_sfc = TD.virtual_pottemp(thermo_params, ᶠts_sfc)
+        θ_v_a = Fields.level(θ_v, 1)
+
+        ## Compute boundary layer height
+
+        ## TODO: Cache elevation field?
+        z_local .= Fields.field_values(Fields.coordinate_field(Y.c).z)
+        z_sfc .= Fields.field_values(
+            Fields.level(Fields.coordinate_field(Y.f).z, half),
+        )
+        @. z_local = z_local - z_sfc
+        dz_local .= Fields.Field(z_local, axes(Y.c))
+        ᶜθ_v_sfc .=
+            Fields.Field(Fields.field_values(θ_v_sfc), axes(interior_uₕ))
+
+        @. Ri = compute_bulk_richardson_number(
+            θ_v,
+            θ_v_a,
+            norm(Y.c.uₕ),
+            grav,
+            dz_local,
+        )
+        @. Ri_a = compute_bulk_richardson_number(
+            θ_v_a,
+            ᶜθ_v_sfc,
+            norm(interior_uₕ),
+            grav,
+            ᶜΔz_surface / 2,
+        )
+
+        #### Detect 𝒽, boundary layer height per column
+        h_boundary_layer = f_b .* Fields.level(ᶜz, Spaces.nlevels(axes(Y.c)))
+        compute_boundary_layer_height!(
+            h_boundary_layer,
+            f_b,
+            dz_local,
+            Ri,
+            Ri_c,
+            Ri_a,
+        )
+
+        ## Exchange coefficients
+        @. C_E =
+            compute_exchange_coefficient(Ri_a, Ri_c, ᶜΔz_surface ./ 2, z₀, κ)
+        @. ᶜK_h = eddy_diffusivity_coefficient(
+            dz_local,
+            z₀,
+            f_b,
+            h_boundary_layer,
+            norm(interior_uₕ),
+            C_E,
+            Ri,
+            Ri_a,
+            Ri_c,
+            κ,
         )
     end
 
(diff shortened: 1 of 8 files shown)


src/prognostic_equations/vertical_diffusion_boundary_layer.jl
```diff
@@ -77,23 +84,47 @@ function vertical_diffusion_boundary_layer_tendency!(
...
+    @. θ_v = TD.virtual_pottemp(thermo_params, ᶜts)
     interior_uₕ = Fields.level(Y.c.uₕ, 1)
-    ᶠp = ᶠρK_E = p.scratch.ᶠtemp_scalar
-    @. ᶠp[colidx] = ᶠinterp(ᶜp[colidx])
+    θ_a = Fields.level(deepcopy(θ_v), 1)
+    θ_v_sfc = @. TD.virtual_pottemp(thermo_params, ᶜts_sfc)
     ᶜΔz_surface = Fields.Δz_field(interior_uₕ)
+
+    ### Detect 𝒽, boundary layer height per column.
+    @. Ri[colidx] = compute_bulk_richardson_number(θ_v[colidx], θ_a[colidx], norm(interior_uₕ[colidx]), grav, ᶜz[colidx])
```
szy21, line 114: I think what might happen is that Ri is 0 for some colidx, and θ_v is overwritten as they point to the same ᶜtemp_scalar. It may not be a problem as θ_v is not used after this, but if you may see these zero values if you print θ_v after this.

szy21, line 114: The windspeed in this function should be the value at level z, not level 1 (interior_uₕ) I think.

akshaysridhar: Agreed, and fixed.

src/prognostic_equations/vertical_diffusion_boundary_layer.jl
```diff
@@ -77,23 +84,47 @@ function vertical_diffusion_boundary_layer_tendency!(
...
+    
+    z₀ = FT(3.21e-5)
+    f_b = FT(0.1)
+    κ = FT(0.4)
+    Ri_c = FT(1)
 
     ᶠgradᵥ = Operators.GradientC2F() # apply BCs to ᶜdivᵥ, which wraps ᶠgradᵥ
+    
+    ᶠρK_E = p.scratch.ᶠtemp_scalar
+    θ_v = p.scratch.ᶜtemp_scalar
```
akshaysridhar, line 99: @simonbyrne I think my issue is related to the use of scratch variables here - theta_v is then used as an argument in line 115. Ri_a is a scalar field at the first interior center level.

akshaysridhar: Assigning potential temperature columnwise to `@. θ_v[colidx]` now; and then calling `Fields.level(θ_v, 1)`to define `θ_a` seems to fix this issue. `deepcopy` is no longer used.

szy21:
Should z be z - z_sfc throughout this PR?

akshaysridhar:
> Should z be z - z_sfc for most places in this PR?

Yup - including here: `77     |   |   h_boundary_layer = Spaces.level(ᶜz[colidx], level)[]` > Currently testing this in the `bomex_box_rhoe.jl` offline.

src/prognostic_equations/vertical_diffusion_boundary_layer.jl
```diff
@@ -17,37 +17,89 @@ import ClimaCore.Operators as Operators
...
+    θ_v::FT,
+    θ_a,
+    norm_ua,
+    grav,
+    z,
+) where {FT}
+    norm²_ua = norm_ua^2# SurfaceFluxes uses FT(1) gustiness #TODO Add to ClimaParameters?
+    return (grav * z) * (θ_v - θ_a) / (θ_a * norm²_ua + eps(FT))
+end
+function compute_diffusion_coefficient(
```
szy21, line 34:
```suggestion
function compute_drag_coefficient(
```
(I think this is more accurate)

szy21: Also, I wonder if we can get the exchange coefficient from surface fluxes. That might be more physically consistent. We don't have to do this now though.

src/prognostic_equations/vertical_diffusion_boundary_layer.jl
```diff
@@ -17,37 +17,89 @@ import ClimaCore.Operators as Operators
...
+    Ri_a::FT,
+    Ri_c::FT,
+    zₐ::FT,
+    z₀::FT,
+    κ::FT,
+) where {FT}
+    # Equations (12), (13), (14)
```
szy21, line 41: Add which paper? (Or better, add documentation, but I'm terrible at it so I won't force you to do that:)

src/prognostic_equations/vertical_diffusion_boundary_layer.jl
```diff
@@ -17,37 +17,89 @@ import ClimaCore.Operators as Operators
...
+    Ri::FT,
+    Ri_a::FT,
+    Ri_c::FT,
+    κ::FT,
+) where {FT}
+    # Equations (17), (18)
+    K = FT(0)
+    if z < f_b * h
+        K_b = compute_surface_layer_diffusivity(z, z₀, κ, C_E, Ri, Ri_c, Ri_a, uₐ)
+        K = K_b
```
szy21, line 88:
```suggestion
        K = compute_surface_layer_diffusivity(z, z₀, κ, C_E, Ri, Ri_c, Ri_a, uₐ)
```

src/prognostic_equations/vertical_diffusion_boundary_layer.jl
```diff
@@ -17,37 +17,89 @@ import ClimaCore.Operators as Operators
...
+    uₐ,
+    C_E::FT,
+    Ri::FT,
+    Ri_a::FT,
+    Ri_c::FT,
+    κ::FT,
+) where {FT}
+    # Equations (17), (18)
+    K = FT(0)
+    if z < f_b * h
```
szy21, line 86:
```suggestion
    if z <= f_b * h
```

src/prognostic_equations/vertical_diffusion_boundary_layer.jl
```diff
@@ -17,37 +17,89 @@ import ClimaCore.Operators as Operators
...
+end
+function compute_diffusion_coefficient(
+    Ri_a::FT,
+    Ri_c::FT,
+    zₐ::FT,
+    z₀::FT,
+    κ::FT,
+) where {FT}
+    # Equations (12), (13), (14)
+    if Ri_a < FT(0)
```
szy21, line 42:
```suggestion
    if Ri_a <= FT(0)
```
(Just to be safe)

src/prognostic_equations/vertical_diffusion_boundary_layer.jl
```diff
@@ -17,37 +17,89 @@ import ClimaCore.Operators as Operators
...
+    z₀::FT,
+    κ::FT,
+    C_E::FT,
+    Ri::FT,
+    Ri_c::FT,
+    Ri_a::FT,
+    norm_uₐ,
+) where {FT}
+    # Equations (19), (20)
+    if Ri_a < FT(0)
```
szy21, line 62:
```suggestion
    if Ri_a <= FT(0)
```
(To be consistent with the previous function)

src/prognostic_equations/vertical_diffusion_boundary_layer.jl
```diff
@@ -77,23 +129,87 @@ function vertical_diffusion_boundary_layer_tendency!(
...
+    thermo_params = CAP.thermodynamics_params(params)
+    (; ᶜts, ᶜp, ᶜspecific, sfc_conditions) = p.precomputed # assume ᶜts and ᶜp have been updated
+
+    κ = CAP.von_karman_const(params)
+    grav = CAP.grav(params)
+    
+    # TODO : Move to ClimaParameters
+    z₀ = FT(3.21e-5)
+    f_b = FT(0.1)
+    Ri_c = FT(1)
```
szy21, line 142: Feel free to add them to CLIMAParameters and tag a release.

src/prognostic_equations/vertical_diffusion_boundary_layer.jl
```diff
@@ -77,23 +129,87 @@ function vertical_diffusion_boundary_layer_tendency!(
...
+    ᶠρK_E = p.scratch.ᶠtemp_scalar
+    θ_v = p.scratch.ᶜtemp_scalar
+    Ri = p.scratch.ᶜtemp_scalar_2
+    dz_local = p.scratch.ᶜtemp_scalar_3
+    θ_v_sfc = p.scratch.ᶠtemp_field_level
+    Ri_a = p.scratch.temp_field_level_2
+
+    # Compute boundary layer height
     FT = eltype(Y)
+    ᶠts_sfc = sfc_conditions.ts
```
szy21, line 155:
```suggestion
    ts_sfc = sfc_conditions.ts
```
Our convention is to only use ᶜ and ᶠ for 3D fields. We didn't follow this convention very well though.

src/prognostic_equations/vertical_diffusion_boundary_layer.jl
```diff
@@ -77,23 +129,87 @@ function vertical_diffusion_boundary_layer_tendency!(
...
+    zₐ = Fields.level(ᶜz, 1)
     interior_uₕ = Fields.level(Y.c.uₕ, 1)
-    ᶠp = ᶠρK_E = p.scratch.ᶠtemp_scalar
-    @. ᶠp[colidx] = ᶠinterp(ᶜp[colidx])
     ᶜΔz_surface = Fields.Δz_field(interior_uₕ)
-    @. ᶠρK_E[colidx] =
-        ᶠinterp(Y.c.ρ[colidx]) * eddy_diffusivity_coefficient(
-            C_E,
+    @. θ_v[colidx] = TD.virtual_pottemp(thermo_params, ᶜts[colidx])
+    θ_a = Fields.level(θ_v[colidx], 1)
```
szy21, line 161:
```suggestion
    θ_v_a = Fields.level(θ_v[colidx], 1)
```
(I think this is more clear)

src/prognostic_equations/vertical_diffusion_boundary_layer.jl
```diff
@@ -77,23 +129,87 @@ function vertical_diffusion_boundary_layer_tendency!(
...
+    )
+    
+    h_boundary_layer = FT(0)
+    for il in 1:Spaces.nlevels(axes(ᶜz))
+        if Spaces.level(Ri[colidx], il)[] < Ri_c
+            h_boundary_layer = Spaces.level(dz_local[colidx], il)[]
+        end
+    end
+
+    C_E = @. compute_diffusion_coefficient(Ri_a, Ri_c, ᶜΔz_surface[colidx], z₀, κ)
```
szy21, line 198: This will allocate I think.

src/utils/common_spaces.jl
```diff
@@ -103,7 +103,7 @@ function make_hybrid_spaces(
         face_space = Spaces.ExtrudedFiniteDifferenceSpace(
             h_space,
             z_face_space,
-            Hypsography.LinearAdaption(z_surface),
+            Hypsography.SLEVEAdaption(z_surface, FT(0.75), FT(10)),
```
szy21, line 106: Change back to default before merging?

src/prognostic_equations/vertical_diffusion_boundary_layer.jl
```diff
@@ -77,23 +129,87 @@ function vertical_diffusion_boundary_layer_tendency!(
...
+    Ri_c = FT(1)
 
     ᶠgradᵥ = Operators.GradientC2F() # apply BCs to ᶜdivᵥ, which wraps ᶠgradᵥ
 
+    ᶠρK_E = p.scratch.ᶠtemp_scalar
+    θ_v = p.scratch.ᶜtemp_scalar
+    Ri = p.scratch.ᶜtemp_scalar_2
+    dz_local = p.scratch.ᶜtemp_scalar_3
+    θ_v_sfc = p.scratch.ᶠtemp_field_level
+    Ri_a = p.scratch.temp_field_level_2
```
szy21, line 151:
```suggestion
    Ri_a = p.scratch.temp_field_level
```
`temp_field_level` hasn't been used yet?

src/cache/temporary_quantities.jl
```diff
@@ -13,6 +13,11 @@ function temporary_quantities(Y, atmos)
         ᶠtemp_scalar = Fields.Field(FT, face_space), # ᶠp, ᶠρK_E
         ᶜtemp_scalar = Fields.Field(FT, center_space), # ᶜ1
         ᶜtemp_scalar_2 = Fields.Field(FT, center_space), # ᶜtke_exch
+        ᶜtemp_scalar_3 = Fields.Field(FT, center_space), # ᶜtke_exch
+        temp_field_level = Fields.level(Fields.Field(FT, center_space), 1),
+        ᶠtemp_field_level = Fields.level(Fields.Field(FT, face_space), half),
+        temp_field_level_2 = Fields.level(Fields.Field(FT, center_space), 1),
+        temp_field_level_3 = Fields.level(Fields.Field(FT, center_space), 1),
```
szy21, line 20: It seems we don't need these two.

szy21 (commented):
Thanks Akshay! This mostly looks good. I left a few comments.

szy21:
I just thought a bit more about this. If we use this version of vertical diffusion with bulk surface fluxes (`DefaultExchangeCoefficient`), it is somewhat inconsistent and may cause problems. For example, for stable regime we will have zero vertical diffusion but still strong surface fluxes (as the drag coefficient is fixed). Not sure if it will become unstable, but it may have some effect on the large-scale circulation. We can either keep both options, or use monin obukhov whenever we use vertical diffusion (but we should keep a test for bulk surface fluxes).

src/prognostic_equations/vertical_diffusion_boundary_layer.jl
LenkaNovak: I agree the `C` should be consistent, but may it be still beneficial to keep this formulation for the initial testing? Also, in the case of wanting to have a continuous diffusivity between the value derived from `SurfaceFluxes.jl` and the interior, I'm getting an extra `\kappa` factor in eqn 19 of Frierson. Maybe it's easier to talk offline about this....

src/prognostic_equations/vertical_diffusion_boundary_layer.jl
```diff
@@ -17,37 +17,89 @@ import ClimaCore.Operators as Operators
...
-    p_pbl = FT(85000)
-    p_strato = FT(10000)
-    K_E = C_E * norm_v_a * z_a
-    return p > p_pbl ? K_E : K_E * exp(-((p_pbl - p) / p_strato)^2)
+function compute_boundary_layer_height(Ri_c::FT) where {FT}
+    # Needs a column search given colidx
+    return nothing
+end
+function compute_bulk_richardson_number(
+    θ_v::FT,
```
LenkaNovak, line 25: Is `θ_v` θ_v(z) and `θ_a` θ_v(z_a). Since `v` refers to virtual for both, I wonder if we could use `θ_z` or something for the first one, or we could add comments next to the arguments?

LenkaNovak (commented):
LGTM, @akshaysridhar. Just a few small comments!

src/prognostic_equations/vertical_diffusion_boundary_layer.jl
szy21: Sure, that's ok. Just that the dynamics will still not be the same as Frierson et al. and fms (unless it has been changed), as they use a simplified Monin-Obukhov which is physically consistent with this vertical diffusion. So I think it might be closer to their setup in some sense if we choose C to be the drag coefficient from our surface flux. It probably doesn't matter too much, so either way is fine.

src/prognostic_equations/vertical_diffusion_boundary_layer.jl
```diff
@@ -77,23 +129,87 @@ function vertical_diffusion_boundary_layer_tendency!(
...
-    (; C_E) = p.atmos.vert_diff
+    (; params) = p
+    thermo_params = CAP.thermodynamics_params(params)
+    (; ᶜts, ᶜp, ᶜspecific, sfc_conditions) = p.precomputed # assume ᶜts and ᶜp have been updated
+
+    κ = CAP.von_karman_const(params)
+    grav = CAP.grav(params)
+    
+    # TODO : Move to ClimaParameters
+    z₀ = FT(3.21e-5)
```
szy21, line 140: The default z₀ for Monin Obukhov is 1e-5, see [here](https://github.com/CliMA/ClimaAtmos.jl/blob/765bbda92963d7e0b663a075a9241ca3aa63c21a/src/surface_conditions/surface_setups.jl#L22). If it's not too much work it would be good to make it a parameter. Otherwise it might be good to make these two the same.

akshaysridhar:
Day 75 (simulation in progress) Spatial Averages
friersondiffusion

akshaysridhar:
This branch `as/frierson-refactor` supersedes this PR content following the changes in #2404. I'll rebase once the current long-runs are complete.

akshaysridhar:
Day 260 (Spatial Averages and Instantaneous Surface / ToA Contours):
friersondiffusion

src/cache/precomputed_quantities.jl
```diff
@@ -369,6 +463,82 @@ NVTX.@annotate function set_precomputed_quantities!(Y, p, t)
...
+        )
+
+        #### Detect 𝒽, boundary layer height per column
+        h_boundary_layer = Fields.level(dz_local, 1)
+        h_boundary_layer =
+            compute_boundary_layer_height!(h_boundary_layer, dz_local, Ri, Ri_c)
+
+        ## Exchange coefficients
+        @. C_E =
+            compute_diffusion_coefficient(Ri_a, Ri_c, ᶜΔz_surface ./ 2, z₀, κ)
```
szy21, line 529: rename it to `compute_exchange_coefficient` or `compute_drag_coefficient`?

akshaysridhar:
@sriharshakandala Can you please take a look at the `perf/flame.jl` file before we add this to the merge queue ?

PR #2351, opened 2023-11-09
remove some regression tests

szy21:
## Purpose 

Also comments out the SSP+zalesak job

## Content

----

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -133,10 +133,6 @@ steps:
         command: >
           julia --color=yes --project=examples examples/hybrid/driver.jl
           --config_file $CONFIG_PATH/single_column_nonorographic_gravity_wave.yml
-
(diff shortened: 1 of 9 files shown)


Sbozzolo:
What's the test `test_env` testing?

The gravity wave tests look like unit tests, can we move them out of buildkite and into the julia Test system?

szy21:
> What's the test `test_env` testing?
> 
I don't know about this test. Maybe @charleskawczynski knows?

charleskawczynski:
Sorry I'm late to this, `test_env` is the only buildkite job that launches julia with `julia --project=test`, which re-resolves the environment and tests ClimaAtmos with the latest versions (which is why that job is marked as `soft_fail: true`). It was my way to get a bit of a look-ahead to see what things coming down the pipeline were failing so that I could fix problems before people realized that they were coming.

charleskawczynski:
(cc @szy21, @Sbozzolo, @sriharshakandala)

PR #2358, opened 2023-11-13
Update instantaneous zenith angle function

sriharshakandala:
## Purpose 
Update instantaneous zenith angle function to use fixed orbital parameters.

----


sriharshakandala:
Please see https://buildkite.com/clima/climaatmos-target-gpu-simulations/builds/141#018bc98f-d764-46f6-9fd3-7edae437f3a9 for the GPU pipeline

src/callbacks/callbacks.jl
```diff
@@ -102,8 +102,12 @@ NVTX.@annotate function rrtmgp_model_callback!(integrator)
         max_zenith_angle = FT(π) / 2 - eps(FT)
         irradiance = FT(CAP.tot_solar_irrad(params))
         au = FT(CAP.astro_unit(params))
-        (; orbital_data) = p.radiation
-
+        date0 = DateTime("2000-01-01T11:58:56.816")
```
szy21, line 105: Could we use CLIMAParameter for this? (If not in this PR, could you open an issue so that we remember to change this?)

szy21 (approved):

src/callbacks/callbacks.jl
sriharshakandala: Sure.

sriharshakandala:
These seems to be  internal `Float32` to `Float64` conversions in the RRTMGP interface code! This needs to be fixed!

regression_tests/ref_counter.jl
```diff
@@ -1 +1 @@
-137
+1
```
akshaysridhar, line 1:
```suggestion
138
```

akshaysridhar: I'd need to double check the regression test infrastructure - not sure if simply updating the `ref_counter` is enough to define a self-reference or if it retains the "last known build" (in this case 137) somehow.

akshaysridhar: It seems pasting the output from the regression tests is sufficient (hence `1` will be a new, and sensible self-reference.

PR #2359, opened 2023-11-13
Remove parallel remapping in ci

szy21:
## Content

----


charleskawczynski:
Should we remove the script, too? Or are people still planning on using it?

szy21:
> Should we remove the script, too? Or are people still planning on using it?

Oh, right, I forgot there is a separate script. @akshaysridhar Could I remove it as well?

szy21:
We talked offline and I'm going to remove the script as well.

PR #2367, opened 2023-11-14
Improve `buoyancy_gradients` design

charleskawczynski:
This PR does two things:
 - Renames  `EnvBuoyGrad` to `EnvBuoyGradVars` and defines `EnvBuoyGrad` as a singleton for dispatching. This is a bit simpler of a pattern being that, in the existing design, `EnvBuoyGrad` contains all of the variables for all `AbstractEnvBuoyGradClosure` specializations. This does mean that we need to pass an extra variable around, but that one is solely used for dispatch. Being that we currently have only one concrete subtype of `AbstractEnvBuoyGradClosure`, we could just do away with the extra parameter and duplicate structs, but I thought that adding the additional parameter wasn't too painful so that we can easily extend / dispatch.
 - Modifies the outer constructor of `EnvBuoyGradVars` to take a thermo state. This should hopefully improve the memory loads, since the thermo function calls can all occur in registers after the thermo state is loaded. Alternatively, we could make `EnvBuoyGradVars` simply hold the thermo state ( or just bypass that altogether ), but then those thermo call would need to be sprinkled in `buoyancy_gradients` and `buoyancy_gradient_chain_rule`, and it's not clear to me that that is worth the additional compute, not to mention it may just make the code look a bit more complicated.

### DIFF
--- a/src/ClimaAtmos.jl
+++ b/src/ClimaAtmos.jl
@@ -1,6 +1,7 @@
 module ClimaAtmos
 
 using NVTX, Colors
+import Thermodynamics as TD
 
 include(joinpath("parameters", "Parameters.jl"))
 import .Parameters as CAP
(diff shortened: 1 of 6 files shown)


charleskawczynski:
Perhaps this will help with https://github.com/CliMA/Thermodynamics.jl/issues/157?

szy21 (approved):
The code change looks good to me, thanks!

PR #2376, opened 2023-11-21
update to Julia 1.9.4, add CUDA back into examples

simonbyrne:
## Purpose 
Try to fix longruns
- updates to Julia 1.9.4
- initiatalizes CUDA inside examples project

## Content

----


szy21 (approved):

szy21:
Trying longrun here: https://buildkite.com/clima/climaatmos-longruns/builds/1265

simonbyrne:
Okay, so i figured out the problem: if you don't specify a CUDA runtime version, CUDA_Runtime_jll adds a cache dependency on the runtime library:
https://github.com/JuliaBinaryWrappers/CUDA_Runtime_jll.jl/blob/d7e700c623a51b93ef2421233f1dbc97d8de5904/.pkg/platform_augmentation.jl#L62-L82
which then gets invalidated if you run it on a different node.

The solution is to specify the CUDA version in the preferences.

PR #2379, opened 2023-11-21
Add implicit diffusion solver

dennisYatunin:
## Purpose 

This PR adds an implicit solver for vertical diffusion, which will be extended to support diagnostic EDMF in a followup PR.

## Content

- Bump the compat entry for ClimaCore to `0.11.1`, which gives us access to the new `ApproximateBlockArrowheadIterativeSolve` algorithm, and update all of the `Manifest.toml` files.
- Add the new config option `implicit_diffusion`, which determines whether to include vertical diffusion in the implicit tendency (and whether to include its derivative in the Jacobian). The current default value is `false`, but we will change this to `true` after further testing.
- Replace the `BlockArrowheadSolve` with an `ApproximateBlockArrowheadIterativeSolve` when `implicit_diffusion` is `true`.
- Add a new example called `sphere_baroclinic_wave_rhoe_equilmoist_impvdiff` to the pipeline. This is a copy of `sphere_baroclinic_wave_rhoe_equilmoist_expvdiff`, but with `implicit_diffusion` set to `true` and `dt` increased to `400secs`. In order for the simulation to produce reasonable results, the value of `max_newton_iters_ode` needs to be increased to 2. This example currently takes a while to compile (almost half an hour on the cluster), and it prints a long compiler warning about type inference. It should be possible to fix that with the changes in CliMA/ClimaCore.jl#1506, which I'll wrap up after we have some preliminary results with diagnostic EDMF.
- Decrease `dt` for the `sphere_baroclinic_wave_rhoe_equilmoist_expvdiff` example to `20secs` because using `40secs` or `30secs` causes the simulation to crash when `t_end` is set to `6days`. Decrease `t_end` to `6hours` so that this example does not slow down CI.

----


src/prognostic_equations/implicit/implicit_tendency.jl
```diff
@@ -9,6 +9,16 @@ NVTX.@annotate function implicit_tendency!(Yₜ, Y, p, t)
     Yₜ .= zero(eltype(Yₜ))
     Fields.bycolumn(axes(Y.c)) do colidx
         implicit_vertical_advection_tendency!(Yₜ, Y, p, t, colidx)
+        if p.atmos.diff_mode == Implicit()
```
szy21, line 12: What's the difference between this and `p.atmos.diff_mode isa Implicit`?

szy21 (approved):

src/prognostic_equations/implicit/implicit_tendency.jl
dennisYatunin: None; since `Implicit` is a singleton type, the assembly code should be identical. I just thought `==` is more readable/common than `isa`.

PR #2393, opened 2023-11-28
Clean up Github Actions/buildkite

Sbozzolo:
This PR:

- ensures that Github action workflows are killed when a new commit is pushed to the same branch/PR (following
https://stackoverflow.com/a/67939898)

- removes the need for the test environment from builkite jobs. The test environment is used only for one job. When the depot is cold, instantiating the test environment takes several minutes (when we have a good depot it takes 1-2 minutes). Given that instantiating the environment has to run before all the tests, this leads to unnecessary latency.

- adds a new GitHub workflow 'lookahead'. The lookahead workflow is non-required to merge PRs and runs tests and one simulation (the one removed from the buildkite run) with julia 1.9 and julia 1.10 and the latest dependencies.

- update checkout action to v4 (should remove warnings about deprecated version of node)

- removed a codecov action (we run codecov as part of ci.yml)


.github/codecov.yml
```diff
@@ -1,11 +0,0 @@
-comment: false
```
charleskawczynski, line 1: Can we keep this file? The GH annotations in particular are a bit annoying

charleskawczynski (commented):
Overall looks fine to me. Can we keep the `codecov.yml`? It suppresses very noisy comments by GH.

.github/codecov.yml
Sbozzolo: Can you share an example of the annotations you mention? 

I thought this yaml file was not being used. What components uses it?

charleskawczynski: They look like this: https://user-images.githubusercontent.com/155044/181088096-36eb1d45-5b7d-4f4f-9721-185effbb9115.png

Sbozzolo: Do you have one from our CI? 

How is this file being used?

Sbozzolo: I followed BechmarkTools (https://github.com/JuliaCI/BenchmarkTools.jl/blob/master/.codecov.yml) and moved the `codecov.yml` file to the root of the project (and prepended a `.`).

Sbozzolo: Just kidding, the [documentation](https://docs.codecov.com/docs/codecov-yaml) says that it has to be in `.git`.

PR #2400, opened 2023-12-01
Add precipitation to Jacobian

trontrytel:
Add precipitation terminal velocity terms to the jacobian


szy21 (approved):

src/prognostic_equations/implicit/implicit_solver.jl
```diff
@@ -247,6 +247,8 @@ function Wfact!(A, Y, p, dtγ, t)
             p.precomputed.ᶠu³,
             p.precomputed.ᶜK,
             p.precomputed.ᶜp,
+            p.precomputed.ᶜwᵣ,
+            p.precomputed.ᶜwₛ,
```
szy21, line 251: We won't have these if we don't use one moment microphysics.

PR #2401, opened 2023-12-01
Add cloud fraction diagnostics

trontrytel:
This PR adds cloud fraction diagnostics based on diagnosed covariances.


src/cache/precomputed_quantities.jl
```diff
@@ -112,11 +112,22 @@ function precomputed_quantities(Y, atmos)
...
+    cloud_diagnostics_quantities =
+        atmos.cloud_diagnostics isa DiagnosticCovariance ?
+        (;
+            ᶜqₜ′qₜ′ = similar(Y.c, FT),
+            ᶜθ′θ′ = similar(Y.c, FT),
+            ᶜθ′qₜ′ = similar(Y.c, FT),
+            cloud_fraction = similar(Y.c, FT)
+            ᶜqₜ_grad = (Y.f, Geometry.WVector{FT}),
+            ᶜθ_grad = (Y.f, Geometry.WVector{FT}),
+        ) : (;)
```
szy21, line 124: We probably only need `cloud_fraction`. It seems we can use temporary fields for all other variables, or just inline them in the `get_cloud_fraction` function.

src/cache/precomputed_quantities.jl
```diff
@@ -54,6 +54,7 @@ function precomputed_quantities(Y, atmos)
         ᶜh_tot = similar(Y.c, FT),
         sfc_conditions = Fields.Field(SCT, Spaces.level(axes(Y.f), half)),
     )
+    cloud_diagnostics = (; cloud_fraction = similar(Y.c, FT),)
```
szy21, line 57:
```suggestion
    cloud_diagnostics = (; ᶜcloud_fraction = similar(Y.c, FT),)
```
(not very important, but it might be good to specify that it is at cell center)

szy21 (approved):
I didn't look at the quad_loop, but the rest looks good, thanks! I may look at it if I have more time, but feel free to merge it.

Sbozzolo:
Note that since you are changing the type/struct of the cache, this will be a breaking change.

trontrytel:
> Note that since you are changing the type/struct of the cache, this will be a breaking change.

It's because I'm renaming `env_thermo_quad` int `SG_quad`?

Sbozzolo:
> > Note that since you are changing the type/struct of the cache, this will be a breaking change.
> 
> It's because I'm renaming `env_thermo_quad` int `SG_quad`?

You are also changing its type to be a subtype of `AbstractSGSamplingType`.

I don't think anyone downstream uses this, but if the cache is supposed to be a public interface, this will change it.

src/cache/cloud_fraction.jl
```diff
@@ -0,0 +1,159 @@
...
+)
+    FT = eltype(qt_mean)
+
+    @assert SG_quad.quadrature_type isa GaussianQuad
+
+    sqrt2 = FT(sqrt(2))
+
+    # Epsilon defined per typical variable fluctuation
+    eps_q = FT(eps(FT)) * max(FT(eps(FT)), qt_mean)
+    eps_θ = FT(eps(FT))
```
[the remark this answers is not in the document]
akshaysridhar: Is there a reason for the above float instruction ? 
```
julia> eps(FT) === FT(eps(FT))
true
```

src/cache/cloud_fraction.jl
```diff
@@ -0,0 +1,159 @@
+import StaticArrays as SA
+import ClimaCore.RecursiveApply: rzero, ⊞, ⊠
+
+# TODO: write a test with scalars that are linear with z
```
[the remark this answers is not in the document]
szy21: No, this is for a physical test of the covariance function. We will add it when tackling "add unit tests to ClimaAtmos".

src/cache/cloud_fraction.jl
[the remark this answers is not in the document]
trontrytel: Great! Thank you. We were going overboard here...

src/cache/cloud_fraction.jl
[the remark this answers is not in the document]
trontrytel: I added a quick sanity check for the cloud fraction in the single column precipitation test. I'm basically checking that it's between 0 and 1 and not a NaN.

src/cache/cloud_fraction.jl
```diff
@@ -0,0 +1,159 @@
...
+    @assert SG_quad.quadrature_type isa GaussianQuad
+
+    sqrt2 = FT(sqrt(2))
+
+    # Epsilon defined per typical variable fluctuation
+    eps_q = FT(eps(FT)) * max(FT(eps(FT)), qt_mean)
+    eps_θ = FT(eps(FT))
+
+    # limit σ_q to prevent negative q_tot_hat
+    σ_q_lim = -qt_mean / (sqrt2 * SG_quad.a[1])
```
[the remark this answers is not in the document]
trontrytel: Good point. I'll go over it and unify the notation. Since now it's become an internal diagnostics routine, could I stick with short names and just go with `q` and `θ` everywhere inside the quadrature loops? I'll make sure to document what those are

trontrytel:
> > > Note that since you are changing the type/struct of the cache, this will be a breaking change.
> > 
> > 
> > It's because I'm renaming `env_thermo_quad` int `SG_quad`?
> 
> You are also changing its type to be a subtype of `AbstractSGSamplingType`.
> 
> I don't think anyone downstream uses this, but if the cache is supposed to be a public interface, this will change it.

I'll consider if I stand behind my name changes enough to do the release, or revert them

Sbozzolo:
> > > > Note that since you are changing the type/struct of the cache, this will be a breaking change.
> > > 
> > > 
> > > It's because I'm renaming `env_thermo_quad` int `SG_quad`?
> > 
> > 
> > You are also changing its type to be a subtype of `AbstractSGSamplingType`.
> > I don't think anyone downstream uses this, but if the cache is supposed to be a public interface, this will change it.
> 
> I'll consider if I stand behind my name changes enough to do the release, or revert them

You shouldn't feel constrained by this. Also, in the past people changed the cache all the time.

src/cache/cloud_fraction.jl
```diff
@@ -0,0 +1,159 @@
...
+    return 2 * coeff * ᶜmixing_length^2 * dot(ᶜ∇Φ, ᶜ∇Ψ)
+end
+
+"""
+   Compute the grid scale cloud fraction based on sub-grid scale properties
+"""
+function set_cloud_fraction!(Y, p, ::DryModel)
+    FT = eltype(Y)
+    (; ᶜcloud_fraction) = p.precomputed
+    @. ᶜcloud_fraction = FT(0)
```
charleskawczynski, line 19:
```suggestion
    @. p.precomputed.ᶜcloud_fraction = 0
```
I don't think type casting is needed for `Int`s, since it should fall back on Julia's type promotion. Also, `ᶜcloud_fraction` already has an element type, which cannot change.

src/diagnostics/core_diagnostics.jl
```diff
@@ -210,6 +210,22 @@ add_diagnostic_variable!(
     end,
 )
 
+###
+# Cloud fraction (3d)
+###
+add_diagnostic_variable!(
+    short_name = "cf",
```
szy21, line 217:
```suggestion
    short_name = "cl",
```
I know it's not intuitive! lol

src/diagnostics/core_diagnostics.jl
```diff
@@ -210,6 +210,22 @@ add_diagnostic_variable!(
...
+    long_name = "Cloud fraction",
+    units = "%",
+    compute! = (out, state, cache, time) -> begin
+        FT = eltype(cache.precomputed.ᶜcloud_fraction)
+        if isnothing(out)
+            return copy(cache.precomputed.ᶜcloud_fraction) .* FT(100)
```
szy21, line 223: You probably don't need FT for Int. @charleskawczynski ?

src/diagnostics/core_diagnostics.jl
trontrytel: oh no! where can I file a complaint about it?

szy21: me lol, or cmip panel

trontrytel:
For computing the Smagorinsky length scale I changed from

```
        l_smag =
            c_smag *
            ᶜdz *
            max(0, 1 - N_eff^2 / ᶜPr / (2 * ᶜstrain_rate_norm))^(1 / 4)
```
to
```
N_eff > FT(0) && N_eff < sqrt(2 * Pr * ϵ_st) ?
           c_smag * dz * (1 - N_eff^2 / Pr / 2 / ϵ_st)^(1 / 4) : c_smag * dz 
```
Otherwise the `max` was giving me zeros almost all the time and was making the whole length scale equal to zero. Does that look reasonable @tapios , @szy21 ?

Additionally I had to change the `EnvBuoyGrad` struct to pass in also the liquid and ice cloud condensate for the `NonEquilibrium` moisture option. I think I can do it more elegantly, but maybe in the next PR. 

The changes in allocations are small. So if the above two issues are fine, I would bump the version, bump the allocations and merge it in.

szy21:
> For computing the Smagorinsky length scale I changed from
> 
> ```
>         l_smag =
>             c_smag *
>             ᶜdz *
>             max(0, 1 - N_eff^2 / ᶜPr / (2 * ᶜstrain_rate_norm))^(1 / 4)
> ```
> 
> to
> 
> ```
> N_eff > FT(0) && N_eff < sqrt(2 * Pr * ϵ_st) ?
>            c_smag * dz * (1 - N_eff^2 / Pr / 2 / ϵ_st)^(1 / 4) : c_smag * dz 
> ```
> 
> Otherwise the `max` was giving me zeros almost all the time and was making the whole length scale equal to zero. Does that look reasonable @tapios , @szy21 ?
> 
I think it makes sense that mixing length goes to zero when buoyancy gradient is larger than strain rate, in which case it will reduce to one quadrature point (or we can do many quadrature points with the same value, as the variance is zero). It is common for mixing length to be zero during the spin-up, as we start with zero velocity and the shear is zero. I don't fully understand why this term is still mostly zero after e.g. 6 hours of bomex, which is why I haven't use this as a lower limit for mixing length. But I think we can figure out that later.

So, in short, I think it would be better to keep the check for `N_eff > FT(0)` but remove `N_eff < sqrt(2 * Pr * ϵ_st)` for now, and make sure the cloud diagnostics work with zero mixing length.

trontrytel:
> So, in short, I think it would be better to keep the check for `N_eff > FT(0)` but remove `N_eff < sqrt(2 * Pr * ϵ_st)` for now, and make sure the cloud diagnostics work with zero mixing length.

It seems weird to have this logic:

`N_eff < 0   =>   l_smag = c_smag * dz` 

`0 < N_eff <  sqrt(2 * Pr * ϵ_st)  =>   l_smag = 0` 

`N_eff > sqrt(2 * Pr * ϵ_st) => c_smag * dz * (1 - N_eff^2 / Pr / 2 / ϵ_st)^(1 / 4)`

The couple of cases I tried worked with zero (co)variances, just produced no cloud fraction. Thats why I was looking for ways to make the estimated (co)variances larger.

tapios:
> > So, in short, I think it would be better to keep the check for `N_eff > FT(0)` but remove `N_eff < sqrt(2 * Pr * ϵ_st)` for now, and make sure the cloud diagnostics work with zero mixing length.
> 
> It seems weird to have this logic. I would at least want to make things smooth between the different limits
> 
> `N_eff < 0 => l_smag = c_smag * dz`
> 
> `0 < N_eff < sqrt(2 * Pr * ϵ_st) => l_smag = 0`
> 
> `N_eff > sqrt(2 * Pr * ϵ_st) => c_smag * dz * (1 - N_eff^2 / Pr / 2 / ϵ_st)^(1 / 4)`
> 
> The couple of cases I tried worked with zero (co)variances, just produced no cloud fraction. Thats why I was looking for ways to make the estimated (co)variances larger. But maybe you are right, that the CI does not provide meaningful conditions to really test it

The original logic (truncating c_smag * dz by a stratification factor) is fine for vertical mixing. It's possible we need to think about fluctuations generated by horizontal SGS motions (i.e., similar logic with horizontal gradients and a horizontal mixing length).

When you do not get clouds at all, that means the mixing length is also zero near the top of the boundary layer?

szy21:
> 
> It seems weird to have this logic. I would at least want to make things smooth between the different limits
> 
> `N_eff < 0 => l_smag = c_smag * dz`
> 
> `0 < N_eff < sqrt(2 * Pr * ϵ_st) => l_smag = 0`
> 
> `N_eff > sqrt(2 * Pr * ϵ_st) => c_smag * dz * (1 - N_eff^2 / Pr / 2 / strain_rate)^(1 / 4)`
> 
I think the logic is actually:
`N_eff < 0 => l_smag = c_smag * dz` 
`0 < N_eff < sqrt(2 * Pr * ϵ_st) => c_smag * dz * (1 - N_eff^2 / Pr / 2 / strain_rate)^(1 / 4)`
`N_eff > sqrt(2 * Pr * ϵ_st) =>  l_smag = 0`
(i.e. switch the second and third lines). Isn't this smooth?

trontrytel:
> I think the logic is actually: `N_eff < 0 => l_smag = c_smag * dz` `0 < N_eff < sqrt(2 * Pr * ϵ_st) => c_smag * dz * (1 - N_eff^2 / Pr / 2 / strain_rate)^(1 / 4)` `N_eff > sqrt(2 * Pr * ϵ_st) => l_smag = 0` (i.e. switch the second and third lines). Isn't this smooth?

You are right. This makes more sense!

trontrytel:
> When you do not get clouds at all, that means the mixing length is also zero near the top of the boundary layer?

I will add mixing length to the output tomorrow and see how it looks in the spherical plots from longer simulations. 

Below are some quick plots after 2 days of `sphere_held_suarez_rhoe_equilmoist_topography_dcmip` from our CI. Thanks to @Sbozzolo new diagnostics and analysis utilities.

test_ta

test_cl

test_hur

tapios:
Looks pretty reasonable! When you plot relative humidity, could you do it in some such way that RH > 1 on the grid scale is clearly recognizable? I'd like to see where you actually have grid-scale supersaturation.

trontrytel:
Here is the Smagorinsky mixing length for the same simulation
test_mixlgm

src/cache/cloud_fraction.jl
```diff
@@ -0,0 +1,233 @@
...
+    Compute the Smagorinsky length scale from
+    - c_smag coefficient
+    - N_eff - buoyancy frequency = sqrt(max(ᶜlinear_buoygrad, 0))
+    - dz - vertical grid scale
+    - Pr - Prandtl number
+    - ϵ_st - strain rate norm
+"""
+function smagorinsky_length_scale(c_smag, N_eff, dz, Pr, ϵ_st)
+    FT = eltype(c_smag)
+    #return N_eff > FT(0) && N_eff < sqrt(2 * Pr * ϵ_st) ?
```
szy21, line 23:
```suggestion
```

src/cache/cloud_fraction.jl
```diff
@@ -0,0 +1,233 @@
...
+    return N_eff > FT(0) ?
+           c_smag * dz * max(0, (1 - N_eff^2 / Pr / 2 / ϵ_st))^(1 / 4) :
```
szy21, line 25:
```suggestion
           c_smag * dz * max(0, (1 - N_eff^2 / Pr / 2 / max(ϵ_st, eps(FT)))^(1 / 4) :
```
Just to be safe. You don't have an issue right now as you are calling it during the diagnostics, when ϵ_st is probably never zero. But if we call it at the first time step (which we may do in the end), it will cause problems as we start with zero velocities.

szy21 (approved):
Thanks!

PR #2402, opened 2023-12-01
Add implict solver for diagnostic EDMF

dennisYatunin:
## Purpose 
This PR extends #2379 to diagnostic EDMF.

## Content
- Move the `edmfx_sgs_diffusive_flux_tendency!` into the implicit tendency when `implicit_diffusion` is `true`.
- Put `ᶜK_u` and `ᶜK_h` into the precomputed quantities when `vert_diff` is `true`.
- Modify the diffusion of `ᶜuₕ` in `vertical_diffusion_boundary_layer_tendency!` to use `2 * ᶠstrain_rate` so that it is consistent with EDMF.
- Add a new `diagnostic_edmfx_dycoms_rf01_box_implicit` example, with a `dt` of `40secs`. This currently looks a bit wrong, and it crashes when `dt` is `50secs` or larger (whereas the explicit version still runs with a `dt` of `100secs`).

----


szy21:
Thanks for the assert which shows a few mistakes in the current config files:) Fixing them here: https://github.com/CliMA/ClimaAtmos.jl/pull/2403

szy21 (approved):

PR #2408, opened 2023-12-04
Increase sampling rate for allocations to 1

Sbozzolo:
This provides a full picture of what is happening. It is not much more expensive.

### DIFF
--- a/perf/flame.jl
+++ b/perf/flame.jl
@@ -34,10 +34,36 @@ ProfileCanvas.html_file(joinpath(output_dir, "flame.html"), results)
 ##### Allocation tests
 #####
 
+allocs_limit = Dict()
+allocs_limit["flame_perf_target"] = 4656
+allocs_limit["flame_perf_target_tracers"] = 204288
+allocs_limit["flame_perf_target_edmfx"] = 253440
+allocs_limit["flame_perf_diagnostics"] = 3016328
+allocs_limit["flame_perf_target_diagnostic_edmfx"] = 893504
+allocs_limit["flame_sphere_baroclinic_wave_rhoe_equilmoist_expvdiff"] =
+    67443909648
+allocs_limit["flame_perf_target_threaded"] = 5857808
+allocs_limit["flame_perf_target_callbacks"] = 46407936
+allocs_limit["flame_perf_gw"] = 4868951088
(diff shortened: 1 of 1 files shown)


Sbozzolo:
https://buildkite.com/clima/climaatmos-ci/builds/15237 for timing information (to see the additional computational cost)

Sbozzolo:
> https://buildkite.com/clima/climaatmos-ci/builds/15237 for timing information (to see the additional computational cost)
> 
> For two jobs the cost is not negligible:
> 
> * barowave jfnk
> * gravity wave

I could reduce the sampling specifically for those two jobs.

sriharshakandala:
Would this be memory intensive?

Sbozzolo:
> Would this be memory intensive?

I am finding that this is prohibitive for jobs that allocate GBs (e.g., gravity waves), so I am trying with scaling the sampling rate with the expected amount of allocations.

perf/flame.jl
```diff
@@ -34,10 +34,36 @@ ProfileCanvas.html_file(joinpath(output_dir, "flame.html"), results)
...
+# expensive there is too many of them. Here, we set the default sample rate to
+# 1, but lower it to a smaller value when we expect the job to produce lots of
+# allocations. Empirically, we find that on the Caltech cluster the limit is 10
+# M of allocation. We scale the sampling rate linearly with that.
+sampling_rate = 1
+
+if haskey(job_id, allocs_limit)
+    max_allocs_for_full_sampling = 10e6
+    sampling_rate = min(1, max_allocs_for_full_sampling / allocs_limit[job_id])
+end
```
charleskawczynski, line 61: Can we just do something simpler here? How about
```suggestion
if haskey(job_id, allocs_limit) && allocs_limit[job_id] > 10e6
    sampling_rate = 0.1
end
```

Also, I think `min` should be `max` here? (`sampling_rate` is between 0 and 1).

Sbozzolo: If `max_allocs_for_full_sampling = 10e6` and `allocs_limit[job_id] = 1`, you'd get a sampling rate of `10e6`, so you want the min I think.

But I agree, it is much simpler to just have 1 or 0.1, so let me do that.

charleskawczynski (approved):

PR #2411, opened 2023-12-05
Add back get_integrator

Sbozzolo:
`get_integrator` got removed in #2179. This PR adds it back (with a deprecation notice). This PR also bumps ClimaAtmos to 0.17.2 to avoid breaking packages that use `get_integrator`.


juliasloan25 (approved):
thank you!

PR #2414, opened 2023-12-05
Improve diagnostics inference

charleskawczynski:
This PR makes the `variable` property of `ScheduledDiagnosticIterations` concretely typed, which may help with inference / allocations.


perf/flame.jl
```diff
@@ -17,7 +17,8 @@ simulation = CA.get_simulation(config)
 
 import SciMLBase
 SciMLBase.step!(integrator) # compile first
-CA.call_all_callbacks!(integrator) # compile callbacks
+CA.call_all_callbacks!(integrator) # compile callbacks for out = ...
```
Sbozzolo, line 20: I don't think you need to do this: the case with = should have been already called once when you get the integrator (https://github.com/CliMA/ClimaAtmos.jl/blob/2812637362192d774d14004a8a84e677dd21fd9a/src/solver/type_getters.jl#L891-L923).

charleskawczynski: Ah, ok, I didn't realize we did that. Looks like that wasn't improving things anyway. I'm still not sure why there are more allocations when we provide more type info.

Sbozzolo:
Even if the allocations increased, I think that it is a good idea to have everything concrete, so I'd still go ahead with this PR. What do you think?

charleskawczynski:
> Even if the allocations increased, I think that it is a good idea to have everything concrete, so I'd still go ahead with this PR. What do you think?

Yeah, I suppose so, I'll rebase and update the numbers

PR #2424, opened 2023-12-09
add entrainment and detrainment limiter to diagnostic edmf

### LINKED ISSUE #2407: Rethink the limiter on entrainment and detrainment for diagnostic EDMF
It should somehow depend on dz

szy21:
## Purpose 

Limits diagnostic edmf entrainment and detrainment to w / dz. Also refactors the limiter on entrainment and detrainment.

Right now I am limiting both entrainment and detrainment to w / dz for diagnostic edmf. I am not sure if it makes sense for detrainment, as this makes it small when w goes to zero near the cloud top, but I think we want detrainment to be large there. I'll need to think more about it. For entrainment, I can't use this upper limit at the first level, as w is zero and limiting it will result in zero entrainment, and zero area fraction throughout the column. So it's a bit ugly:( Another way is to calculate w at the next level first then use that for the limiter, but this can potentially result in different entrainment for w and tracers, which I would like to avoid unless we have to. Other than these issues this PR is ready for review.

Some time ago I increased the entrainment to make the timestep more limited by mass flux. The simulations didn't fail but the results look pretty noisy. I reduced dt here to make the results look slightly smoother. We can revisit this after we treat mass flux implicitly.

Closes #2407 

## Content

----

### DIFF
--- a/config/model_configs/diagnostic_edmfx_aquaplanet.yml
+++ b/config/model_configs/diagnostic_edmfx_aquaplanet.yml
@@ -12,7 +12,7 @@ edmfx_sgs_diffusive_flux: true
 moist: equil 
 precip_model: 0M
 override_τ_precip: false
-dt: 10secs 
+dt: 50secs 
 t_end: 1hours 
 dt_save_to_disk: 600secs
 toml: [toml/diagnostic_edmfx_trmm_box.toml]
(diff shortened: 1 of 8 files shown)


trontrytel:
I like the idea of splitting the entr/detr limiter function from the actual entr/detr formulae themselves. Annoying with the special treatment of the first level. We could come up with some additional entrainment term that is only acting close to the surface and treat it separately. But not sure if thats less annoying

trontrytel (approved):

szy21:
I'll need to think a bit more about this. Entrainment and detrainment are also used in the prognostic TKE equation, so besides some w / dz limit, I think we probably need some dt limit as well.

PR #2428, opened 2023-12-12
Add callback that estimates walltime

charleskawczynski:
This PR adds a callback that logs some helpful runtime information:

```
┌ Info: Progress
│   simulation_time = "21 hours, 30 minutes"
│   n_steps_completed = 129
│   wall_time_per_step = "101 milliseconds, 824 microseconds"
│   wall_time_total = "2 minutes, 26 seconds"
│   wall_time_remaining = 133.4913111176602
│   wall_time_spent = "13 seconds, 135 milliseconds"
│   percent_complete = "9.0%"
│   sypd = 16.144
│   date_now = 2023-12-14T10:42:19.127
└   estimated_finish_date = 2023-12-14T10:44:32.618
```

The callback uses a doubling backoff (we log on step 2, 4, 8, 16, 32, ...) to provide frequent estimates near the beginning of simulations, but increasingly infrequent as time goes on to avoid bloating the log (I don't want people to have to scroll through a zillion info statements in the longruns to get to the end of the log).

### DIFF
--- a/perf/flame.jl
+++ b/perf/flame.jl
@@ -17,6 +17,7 @@ simulation = CA.get_simulation(config)
 
 import SciMLBase
 SciMLBase.step!(integrator) # compile first
+SciMLBase.step!(integrator) # compile print_walltime_estimate, which skips the first step to avoid timing compilation
 CA.call_all_callbacks!(integrator) # compile callbacks
 import Profile, ProfileCanvas
 output_dir = job_id
@@ -36,17 +37,17 @@ ProfileCanvas.html_file(joinpath(output_dir, "flame.html"), results)
 #####
 
 allocs_limit = Dict()
-allocs_limit["flame_perf_target"] = 147_520
-allocs_limit["flame_perf_target_tracers"] = 179_776
+allocs_limit["flame_perf_target"] = 148_256
+allocs_limit["flame_perf_target_tracers"] = 180_512
 allocs_limit["flame_perf_target_edmfx"] = 7_005_552
 allocs_limit["flame_perf_diagnostics"] = 25_356_928
-allocs_limit["flame_perf_target_diagnostic_edmfx"] = 1_309_968
+allocs_limit["flame_perf_target_diagnostic_edmfx"] = 1_311_040
 allocs_limit["flame_sphere_baroclinic_wave_rhoe_equilmoist_expvdiff"] =
     4_018_252_656
 allocs_limit["flame_perf_target_threaded"] = 1_276_864
 allocs_limit["flame_perf_target_callbacks"] = 37_277_112
-allocs_limit["flame_perf_gw"] = 3_226_428_736
-allocs_limit["flame_perf_target_prognostic_edmfx_aquaplanet"] = 1_257_712
+allocs_limit["flame_perf_gw"] = 3_226_429_472
+allocs_limit["flame_perf_target_prognostic_edmfx_aquaplanet"] = 1_258_848
 
 # Ideally, we would like to track all the allocations, but this becomes too
 # expensive there is too many of them. Here, we set the default sample rate to
--- a/src/cache/cache.jl
+++ b/src/cache/cache.jl
@@ -1,5 +1,7 @@
 struct AtmosCache{
     FT <: AbstractFloat,
+    FTE,
+    WTE,
     SD,
     AM,
     NUM,
@@ -27,6 +29,12 @@ struct AtmosCache{
 }
     """Timestep of the simulation (in seconds). This is also used by callbacks and tendencies"""
     dt::FT
+
+    """End time of the simulation (in seconds). This used by callbacks"""
+    t_end::FTE
+
+    """Walltime estimate"""
+    walltime_estimate::WTE
 
     """Start date (used for insolation)."""
     start_date::SD
@@ -93,7 +101,7 @@ end
 
 # The model also depends on f_plane_coriolis_frequency(params)
 # This is a constant Coriolis frequency that is only used if space is flat
-function build_cache(Y, atmos, params, surface_setup, dt, start_date)
+function build_cache(Y, atmos, params, surface_setup, dt, t_end, start_date)
     FT = eltype(params)
 
     ᶜcoord = Fields.local_geometry_field(Y.c).coordinates
@@ -184,6 +192,8 @@ function build_cache(Y, atmos, params, surface_setup, dt, start_date)
 
     args = (
         dt,
+        t_end,
+        WallTimeEstimate(),
         start_date,
         atmos,
         numerics,
--- a/src/callbacks/callbacks.jl
+++ b/src/callbacks/callbacks.jl
@@ -434,6 +434,94 @@ function save_restart_func(integrator, output_dir)
     return nothing
 end
 
+Base.@kwdef mutable struct WallTimeEstimate
+    """Number of calls to the callback"""
+    n_calls::Int = 0
+    """Int indicating next time the callback will print to the log"""
+    n_next::Int = 1
+    """Wall time of previous call to update `WallTimeEstimate`"""
+    t_wall_last::Float64 = -1
+    """Sum of elapsed walltime over calls to `step!`"""
+    ∑Δt_wall::Float64 = 0
+    """Fixed increment to increase n_next by after 5% completion"""
+    n_fixed_increment::Float64 = -1
+end
+import Dates
+function print_walltime_estimate(integrator)
+    (; walltime_estimate, dt, t_end) = integrator.p
+    wte = walltime_estimate
+
+    # Notes on `ready_to_report`
+    #   - The very first call (when `n_calls == 0`), there's no elapsed
+    #     times to report (and this is called during initialization,
+    #     before `step!` has been called).
+    #   - The second call (`n_calls == 1`) is after `step!` is called
+    #     for the first time, but we don't want to report this since it
+    #     includes compilation time.
+    #   - Calls after that (`n_calls > 1`) exclude compilation and provide
+    #     the best wall time estimates
+
+    ready_to_report = wte.n_calls > 1
+    if ready_to_report
+        # We need to account for skipping cost of `Δt_wall` when `n_calls == 1`:
+        factor = wte.n_calls == 2 ? 2 : 1
+        Δt_wall = factor * (time() - wte.t_wall_last)
+    else
+        wte.n_calls == 1 && @info "Progress: Completed first step"
+        Δt_wall = Float64(0)
+        wte.n_next = wte.n_calls + 1
+    end
+    wte.∑Δt_wall += Δt_wall
+    wte.t_wall_last = time()
+
+    if wte.n_calls == wte.n_next && ready_to_report
+        t = integrator.t
+        n_steps_total = ceil(Int, t_end / dt)
+        n_steps = ceil(Int, t / dt)
+        wall_time_ave_per_step = wte.∑Δt_wall / n_steps
+        wall_time_ave_per_step_str = time_and_units_str(wall_time_ave_per_step)
+        percent_complete = round(t / t_end * 100; digits = 1)
+        n_steps_remaining = n_steps_total - n_steps
+        wall_time_remaining = wall_time_ave_per_step * n_steps_remaining
+        wall_time_remaining_str = time_and_units_str(wall_time_remaining)
+        wall_time_total =
+            time_and_units_str(wall_time_ave_per_step * n_steps_total)
+        wall_time_spent = time_and_units_str(wte.∑Δt_wall)
+        simulation_time = time_and_units_str(Float64(t))
+        sypd = round(
+            simulated_years_per_day(
+                EfficiencyStats((zero(t), t), wte.∑Δt_wall),
+            );
+            digits = 3,
+        )
+        estimated_finish_date =
+            Dates.now() + compound_period(wall_time_remaining, Dates.Second)
+        @info "Progress" simulation_time = simulation_time n_steps_completed =
+            n_steps wall_time_per_step = wall_time_ave_per_step_str wall_time_total =
+            wall_time_total wall_time_remaining = wall_time_remaining_str wall_time_spent =
+            wall_time_spent percent_complete = "$percent_complete%" sypd = sypd date_now =
+            Dates.now() estimated_finish_date = estimated_finish_date
+
+        # the first fixed increment is equivalent to
+        # doubling (which puts us at 10%), so we check
+        # if we're below 5%.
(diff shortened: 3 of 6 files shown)


Sbozzolo:
Thanks, this is very useful! I will review it tomorrow morning.

A very quick comment is that I find `total` a little misleading: at the beginning, I thought I meant "total elapsed time since the beginning"

charleskawczynski:
Ah, `time_and_units` looks wrong.. It'd be nice if we can do this more elegantly. Looking at DateTIme now..

Sbozzolo:
> Ah, `time_and_units` looks wrong.. It'd be nice if we can do this more elegantly. Looking at DateTIme now..

Let me know when you want me to review the PR :)

simonbyrne:
I find the output kind of hard to parse: could you put the values next to the name? Or even on multiple lines (the macros do this automatically if you provide extra arguments
```
julia> @info "Wall time" var"per step"=100 total=200 remaining=300
┌ Info: Wall time
│   per step = 100
│   total = 200
└   remaining = 300
```

simonbyrne:
> Ah, `time_and_units` looks wrong.. It'd be nice if we can do this more elegantly. Looking at DateTIme now..

[`canonicalize`](https://docs.julialang.org/en/v1/stdlib/Dates/#Dates.canonicalize) is helpful here:
```julia
julia> using Dates

julia> canonicalize(Second(10_000))
2 hours, 46 minutes, 40 seconds
```

charleskawczynski:
> [`canonicalize`](https://docs.julialang.org/en/v1/stdlib/Dates/#Dates.canonicalize) is helpful here:
> 
> ```julia
> julia> using Dates
> 
> julia> canonicalize(Second(10_000))
> 2 hours, 46 minutes, 40 seconds
> ```

Yeah, but for now (https://github.com/JuliaLang/julia/pull/52519 😄), there's no way to convert from `Real`.

Sbozzolo:
> > Let me know when you want me to review the PR :)
> 
> I think I have a nice solution, but maybe you can review the rest of the PR in the mean time (just ignore `time_and_units` for now). @Sbozzolo

One general design comment before I look at the details is that maybe we can be a little bit more ambitious that just tracking the walltime. You added a `TWE` object to the cache, maybe we can turn this into a more general instrumentation object for ClimaAtmos runs. For example, the object might also track memory usage, and maybe optionally print more detailed/frequent stats to file. At this stage, it would just track walltime, but this would provide a natural place where to put extra instrumentation.

What do you think?

(For this PR, this would probably just amounts to renaming the struct)

(And I also agree with Simon that the output is hard to read.)

charleskawczynski:
Let's also add time spent

simonbyrne:
I think it would be good to have:
- time elapsed (how long so far)
- sypd estimate (can compute from wall time / time step)

simonbyrne:
ideally we would also want to exclude the first time step.

charleskawczynski:
Alright, I've
 - added time elapsed / spent
 - Renamed `n_steps` to `n_calls`, which is a more accurate name
 - Avoided `n_steps = 1` (since the cb is called during init, which messed up the timings)
 - Added a sypd estimate

perf/flame.jl
```diff
@@ -17,6 +17,7 @@ simulation = CA.get_simulation(config)
 
 import SciMLBase
 SciMLBase.step!(integrator) # compile first
+SciMLBase.step!(integrator) # compile print_walltime_estimate, which skips the first step to avoid
```
Sbozzolo, line 20: It looks like that the comment has been truncated midsentence?

src/cache/cache.jl
```diff
@@ -28,6 +30,12 @@ struct AtmosCache{
     """Timestep of the simulation (in seconds). This is also used by callbacks and tendencies"""
     dt::FT
 
+    """Timestep of the simulation (in seconds). This is also used by callbacks and tendencies"""
```
Sbozzolo, line 33: Is this used in tendencies?

src/callbacks/callbacks.jl
```diff
@@ -434,6 +434,64 @@ function save_restart_func(integrator, output_dir)
     return nothing
 end
 
+Base.@kwdef mutable struct WallTimeEstimate
+    n_calls::Int = 0
+    n_next::Int = 1
+    t_last::Float64 = -1
+    ∑Δt_wall::Float64 = 0
```
Sbozzolo, line 441: Can you please add docstrings to the various attributes describing what they are?

Also, maybe we want change `t_last` to `t_wall_last` to disambiguate it with the actual simulation time

src/callbacks/callbacks.jl
```diff
@@ -434,6 +434,64 @@ function save_restart_func(integrator, output_dir)
...
+end
+import Dates
+function print_walltime_estimate(integrator)
+    (; walltime_estimate, dt, t_end) = integrator.p
+    wte = walltime_estimate
+    # 1st call initializes `t_last` (during integrator init)
+    # 2nd call computes `Δt` for the first time (but also includes compilation), so we skip
+    cond = wte.n_calls > 1
+    if cond
```
Sbozzolo, line 450: Can you please rename `cond` to something more descriptive? Maybe `is_first_invocation` and flip the conditionals?

src/callbacks/callbacks.jl
```diff
@@ -434,6 +434,64 @@ function save_restart_func(integrator, output_dir)
...
+        # We need to account for skipping cost of `Δt` when `n_calls == 1`:
+        factor = wte.n_calls == 2 ? 2 : 1
+        Δt = factor * (time() - wte.t_last)
```
Sbozzolo, line 453: I think it would be useful to append `_wall` or something like that to the times and deltas that are walltimes.

It gets a little confusion to keep track of which is which below.

src/callbacks/callbacks.jl
```diff
@@ -434,6 +434,64 @@ function save_restart_func(integrator, output_dir)
...
+        wte.n_calls == 1 && @info "Progress: Completed first step"
+        Δt = Float64(0)
+        wte.n_next = wte.n_calls + 1
+    end
+    wte.∑Δt_wall += Δt
+    wte.t_last = time()
+
+    if wte.n_calls == wte.n_next && cond
+        t = integrator.t
+        n_steps_total = ceil(Int, t_end / dt)
```
Sbozzolo, line 464: I think that this might not be correct when we restart a simulation from a checkpoint.

src/callbacks/callbacks.jl
```diff
@@ -434,6 +434,64 @@ function save_restart_func(integrator, output_dir)
...
+        n_steps = ceil(Int, t / dt)
+        t_wall_ave_per_step = wte.∑Δt_wall / n_steps
+        twps = t_wall_ave_per_step
+        _twps = time_and_units(t_wall_ave_per_step)
+        percent_complete = round(t / t_end * 100; digits = 1)
+        n_steps_remaining = n_steps_total - n_steps
+        erwt = twps * n_steps_remaining
+        _erwt = time_and_units(erwt)
+        etwt = time_and_units(twps * n_steps_total)
```
Sbozzolo, line 473: There are some transformations of units and types here, but it unclear from the variable names which units are being used. Instead of just prepending `_`, you could add be more explicit.

I'd also suggest adding a comment to explain that `e*wt` stands for estimated * remaining walltime.

src/callbacks/callbacks.jl
```diff
@@ -434,6 +434,64 @@ function save_restart_func(integrator, output_dir)
...
+            );
+            digits = 3,
+        )
+        estimated_finish_date =
+            Dates.now() + compound_period(erwt, Dates.Second)
+        @info "Progress" simulation_time = t_sim n_steps_completed = n_steps wall_time_per_step =
+            _twps wall_time_total = etwt wall_time_remaining = erwt wall_time_spent =
+            spent percent_complete = "$percent_complete%" sypd = sypd date_now =
+            Dates.now() estimated_finish_date = estimated_finish_date
+        wte.n_next *= 2 # doubling factor (to reduce log noise)
```
Sbozzolo, line 488: I think we should have a cap (say, every 10 %), so that we still get periodic logs after 50 % of the simulation.

src/utils/utilities.jl
```diff
@@ -210,6 +210,48 @@ function prettytime(t)
     return @sprintf("%.3f %s", value, units)
 end
 
+import Dates
+using Dates
```
Sbozzolo, line 214: Do we need both?

src/utils/utilities.jl
```diff
@@ -210,6 +210,48 @@ function prettytime(t)
...
+    for i in eachindex(periods)
+        T, n = periods[i]
+        N = Int64(1)
+        for j in (i - 1):-1:firstindex(periods) # less-precise periods
+            Tc, nc = periods[j]
+            N *= nc
+            @eval time_per_time(::Type{$T}, ::Type{$Tc}) = $N
+        end
+    end
+end
```
Sbozzolo, line 227: Can you please add a comment to explain this?

Sbozzolo (commented):
Thank you! I have only minor cosmetic comments.

src/cache/cache.jl
charleskawczynski: Whoops, forgot to edit that, I’ll update

perf/flame.jl
charleskawczynski: Yeah, will fi

src/utils/utilities.jl
charleskawczynski: I'll ditch the `using` and just qualify things

src/callbacks/callbacks.jl
charleskawczynski: It's actually the _third_ invocation (`n_calls` is zero to start).

 - The very first call (when `n_calls == 0`), there's no elapsed times to report (and this is called during initialization, before `step!` has been called).
 - The second call (`n_calls == 1`) is after `step!` is called for the first time, but we don't want to report this since it includes compilation time.
 - Calls after that (`n_calls > 1`) exclude compilation and provide the best wall time estimates

Maybe we should include this list/description above in the docs?

Regarding the variable name, how about `ready_to_report = wte.n_calls > 1`? (that's what I'll do tentatively, let me know if you prefer something else)

src/callbacks/callbacks.jl
charleskawczynski: Yeah, it's not correct for restarted simulations. I'd like to extend support for this in a followup PR-- I think the logic will be a bit more complicated (I've done this before in another project, but in Fortran). IIRC, the reason it's not as simple is because we're accumulating `dt_wall` to provide an estimate that kind of converges over time (callbacks and variation in timings can make the results noisy, which is a downside of not accumulating the estimates). So, if we wanted to keep the accumulation version, we'd need to save this to disc and read it in for restarted simulations.

src/callbacks/callbacks.jl
charleskawczynski: `time_and_units` takes in floats and returns strings, I'll document it. I'll also make the names verbose for clarity

charleskawczynski: Please reopen if it still seems unclear

src/callbacks/callbacks.jl
charleskawczynski: Please unresolve if you think it still needs another look

src/utils/utilities.jl
charleskawczynski: Sure, I'll add some docs (I took this from Dates, but it wasn't really documented there). Ultimately, this is just defining:

```julia
time_per_time(::Type{Dates.Second}, ::Type{Dates.Minute}) = 60
time_per_time(::Type{Dates.Second}, ::Type{Dates.Hour}) = 3600
time_per_time(::Type{Dates.Second}, ::Type{Dates.Day}) = 86400
time_per_time(::Type{Dates.Second}, ::Type{Dates.Week}) = 604800
time_per_time(::Type{Dates.Millisecond}, ::Type{Dates.Second}) = 1000
time_per_time(::Type{Dates.Millisecond}, ::Type{Dates.Minute}) = 60000
time_per_time(::Type{Dates.Millisecond}, ::Type{Dates.Hour}) = 3600000
time_per_time(::Type{Dates.Millisecond}, ::Type{Dates.Day}) = 86400000
time_per_time(::Type{Dates.Millisecond}, ::Type{Dates.Week}) = 604800000
time_per_time(::Type{Dates.Day}, ::Type{Dates.Week}) = 7
time_per_time(::Type{Dates.Hour}, ::Type{Dates.Day}) = 24
time_per_time(::Type{Dates.Hour}, ::Type{Dates.Week}) = 168
time_per_time(::Type{Dates.Minute}, ::Type{Dates.Hour}) = 60
time_per_time(::Type{Dates.Minute}, ::Type{Dates.Day}) = 1440
time_per_time(::Type{Dates.Minute}, ::Type{Dates.Week}) = 10080
time_per_time(::Type{Dates.Nanosecond}, ::Type{Dates.Microsecond}) = 1000
time_per_time(::Type{Dates.Nanosecond}, ::Type{Dates.Millisecond}) = 1000000
time_per_time(::Type{Dates.Nanosecond}, ::Type{Dates.Second}) = 1000000000
time_per_time(::Type{Dates.Nanosecond}, ::Type{Dates.Minute}) = 60000000000
time_per_time(::Type{Dates.Nanosecond}, ::Type{Dates.Hour}) = 3600000000000
time_per_time(::Type{Dates.Nanosecond}, ::Type{Dates.Day}) = 86400000000000
time_per_time(::Type{Dates.Nanosecond}, ::Type{Dates.Week}) = 604800000000000
time_per_time(::Type{Dates.Microsecond}, ::Type{Dates.Millisecond}) = 1000
time_per_time(::Type{Dates.Microsecond}, ::Type{Dates.Second}) = 1000000
time_per_time(::Type{Dates.Microsecond}, ::Type{Dates.Minute}) = 60000000
time_per_time(::Type{Dates.Microsecond}, ::Type{Dates.Hour}) = 3600000000
time_per_time(::Type{Dates.Microsecond}, ::Type{Dates.Day}) = 86400000000
time_per_time(::Type{Dates.Microsecond}, ::Type{Dates.Week}) = 604800000000
```

src/callbacks/callbacks.jl
charleskawczynski: Yeah, I guess that's not a bad idea. It should only add another, at most, 10 more printed logs.

src/callbacks/callbacks.jl
charleskawczynski: For now, I'll just add a warning if a simulation is restarted.

Sbozzolo (approved):
Looks good to me! :)

PR #2434, opened 2023-12-15
Make the 1M scheme more stable

trontrytel:
This should make it possible to run the `sphere_aquaplanet_rhoe_equilmoist_allsky_gw_res` CI case for 24 hours. Here is how the model looks at the end of that day:

test__time_86400 0

I'm also testing the longruns CI job `longrun_aquaplanet_rhoe_equilmoist_nz63_0M_55km_rs35km_clearsky_tvinsolation`.


szy21 (approved):

trontrytel:
The longruns case failed. I'll keep looking into it.

Issue #2439, opened 2023-12-15
Improve likelihood that broadcasted structs are gpu compatible


charleskawczynski:
xref: https://github.com/CliMA/ClimaAtmos.jl/issues/2438#issuecomment-1858183144

charleskawczynski:
I think I did this in ClimaTimesteppers, and maybe this would be useful here: https://github.com/CliMA/ClimaTimeSteppers.jl/blob/main/test/integrator_utils.jl

charleskawczynski:
Is suppose #2440 closes this. We can reopen if it continues to be a problem.

PR #2440, opened 2023-12-15
Assert atmos to not have UnionAll types

charleskawczynski:
Let's see if this can help with https://github.com/CliMA/ClimaAtmos.jl/issues/2439. I'll need to fix the resulting errors.


Sbozzolo:
Would it be possible to have this as unit test instead?

charleskawczynski:
> Would it be possible to have this as unit test instead?

It's possible, but then we would need to construct all the different instances that we make in CI to perform the tests (assuming we want it strictly enforced).

PR #2444, opened 2023-12-18
Save final state when simulation crashes

Sbozzolo:
With this commit, the state of the simulation at the last timestep is saved whenever a simulation crashes (e.g., saturation adjustment).

### DIFF
--- a/src/solver/solve.jl
+++ b/src/solver/solve.jl
@@ -67,6 +67,8 @@ function solve_atmos!(simulation)
             return AtmosSolveResults(sol, :success, walltime)
         end
     catch ret_code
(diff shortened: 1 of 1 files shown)


szy21 (approved):
Thanks!

szy21:
Do you think it would be useful to save the cache when the simulation crashes? Or is it too much?

Sbozzolo:
> Do you think it would be useful to save the cache when the simulation crashes? Or is it too much?

I can add that

Sbozzolo:
> Member

Actually, it is not easy, because the cache has a mix of types and we don't have a HDF5 that can handle it easily.

PR #2456, opened 2023-12-27
add mixing length to precomputed quantities

szy21:
## Purpose 

This PR moves mixing length to the default precomputed quantities (with or without EDMF) and uses EDMF mixing length to calculate cloud fraction. This is the main part of this PR, but I made some other changes:

- Move the grid-mean mixing length closure to a new file `closure.jl`. I think this makes `precomputed_quantities.jl` more clean. To keep the PR small I only moved the grid-mean mixing length closure, but I will move some other functions to the new file in the future.
- Make `lmix` a default diagnostic, with or without EDMF.
- I finally removed the limiter `if l < 0 set l = l_max` in the edmf mixing length closure. I don't understand where it comes from, and the limiter on dissipation seems to suffice. This limiter is causing problems as the simplified covariance is proportional to `l^2`. With this limiter, `l = l_max` when TKE is zero, resulting in very large covariance and quadrature points with negative theta_li. If we use the full covariance (eq. 50 on overleaf) this problem should be less severe as both TKE and diffusivity go to zero. In any case, I think we should remove this filter to avoid future problems. I would like to move on with this as we can at least have some cloud fraction diagnostics with EDMF. We can revisit it if there are issues with the longrun. @trontrytel @tapios What do you think?

## Content

----


Project.toml
```diff
@@ -68,7 +68,7 @@ IntervalSets = "0.5, 0.6, 0.7"
 Krylov = "0.9"
 LinearAlgebra = "1"
 Logging = "1"
-NCDatasets = "0.11, 0.12, 0.13"
```
szy21, line 71: There are some compatibility issues with NCDatasets. See #2399

src/prognostic_equations/closures.jl
```diff
@@ -0,0 +1,94 @@
...
+    )
+
+    ᶠu = p.scratch.ᶠtemp_C123
+    @. ᶠu = C123(ᶠinterp(Y.c.uₕ)) + C123(ᶠu³)
+    ᶜstrain_rate = p.scratch.ᶜtemp_UVWxUVW
+    compute_strain_rate_center!(ᶜstrain_rate, ᶠu)
+
+    turbconv_params = CAP.turbconv_params(params)
+    ᶜprandtl_nvec = p.scratch.ᶜtemp_scalar_2
+    # TODO: Fix the broadcasting error in turbulent_prandtl_number for XZPoint
```
szy21, line 75: There seem to be some broadcasting issues with a 1D surface field and a 2D XZ plane field. This is a temporary hack. I will check with Simon.

src/ClimaAtmos.jl
```diff
@@ -75,6 +75,7 @@ include(
     ),
 )
 include(joinpath("prognostic_equations", "hyperdiffusion.jl"))
+include(joinpath("prognostic_equations", "closures.jl"))
```
tapios, line 78: Can we give it a more specific name? This is about SGS closures, but right now just for mixing lengths, right? So why not `mixing_length_closure` or something like this?

src/prognostic_equations/closures.jl
```diff
@@ -0,0 +1,94 @@
...
+
+"""
+    Compute the Smagorinsky length scale from
+    - c_smag coefficient
+    - N_eff - buoyancy frequency = sqrt(max(ᶜlinear_buoygrad, 0))
+    - dz - vertical grid scale
+    - Pr - Prandtl number
+    - ϵ_st - strain rate norm
+"""
+function smagorinsky_length_scale(c_smag, N_eff, dz, Pr, ϵ_st)
```
tapios, line 16: If we want to be picky (and maybe we do), a better name would be `smagorinsky_lilly_length`.

src/prognostic_equations/edmfx_closures.jl
```diff
@@ -218,10 +218,13 @@ function mixing_length(
     # kz scale (surface layer)
     if obukhov_length < 0.0 #unstable
         l_W =
-            vkc * (ᶜz - z_sfc) / (sqrt(sfc_tke / ustar / ustar) * c_m) *
+            vkc * (ᶜz - z_sfc) /
+            max(sqrt(sfc_tke / ustar / ustar) * c_m, eps(FT)) *
             min((1 - 100 * (ᶜz - z_sfc) / obukhov_length)^FT(0.2), 1 / vkc)
```
tapios, line 223: What is the 'min' factor here? I do not understand where it comes from. At least we need documentation (e.g., is the exponent 0.2 universal or some ad hoc factor)?

tapios (approved):
Thank you, @szy21! Your reasoning here makes sense to me, and the code looks good to me.

src/ClimaAtmos.jl
szy21: Yes, this is about SGS closures without EDMF. Right now it only has mixing length, but I was thinking about moving the buoyancy gradient, prandtl number, dycore vertical diffusivity, and maybe cloud fraction into this file as well. So basically it will contain all the closures that does not require EDMF variables. (EDMF closures, such as mixing length or non-hydrostatic pressure drag, are in a separate file `edmfx_closures.jl`.) Any suggestion on names?

src/prognostic_equations/edmfx_closures.jl
szy21: This is copied from TurbulenceConvection: https://github.com/CliMA/TurbulenceConvection.jl/blob/b613a8f7c9d6336ac361f4c76edfeb6dbfb51361/src/closures/mixing_length.jl#L20. I didn't find details in Ignacio's paper, except that it says an "empirical stability function" after eq. 34. Maybe we should use the one consistent with SurfaceFluxes?

src/ClimaAtmos.jl
tapios: How about `gm_sgs_closures`? It's not perfectly specific, but maybe a little bit clearer?

src/prognostic_equations/edmfx_closures.jl
tapios: Ok. It looks like Businger functions. It would be best to get these from SurfaceFluxes.

szy21: Sounds good, I don't want to change the calculation of mixing length in this PR, but I'll change this next.

szy21:
This PR makes all the dycore simulations a factor of two slower. I think it might be from the buoyancy gradient. I need to look into it. If I can't find where the problem is I will temporarily setthe  grid-mean mixing length to ~ dz, or use a callback. @trontrytel This may suggest that the slowdown you saw before is not from the quadrature, but from the mixing length calculation?

szy21:
It is indeed mostly from the buoyancy gradient calculation. Not sure if I can figure out why it is slow soon. Moving it to callback for now: https://github.com/CliMA/ClimaAtmos.jl/pull/2466

PR #2464, opened 2023-12-30
remove the hack for prandtl number of XZ plane

szy21:
## Purpose 

To be merged after https://github.com/CliMA/ClimaCore.jl/pull/1584

## Content

----


simonbyrne:
You can also use the new length scale stuff as well.

szy21:
why ci, why 😭

PR #2480, opened 2024-01-04
Update to Julia 1.10

Sbozzolo:


charleskawczynski (approved):
Can we merge this?

Sbozzolo:
No, even with small sampling rate, ProfileCanvas is very slow and some performance jobs take much longer.

charleskawczynski:
> No, even with small sampling rate, ProfileCanvas is very slow and some performance jobs take much longer.

Buildkite finished in 50 min, that's about how long it's currently taking. What's taking longer exactly?

Sbozzolo:
This job took 41 minutes: https://buildkite.com/clima/climaatmos-ci/builds/15741#018ceb9b-eaa6-420f-82d5-fb88e4b77054

We could reduce the sampling rate even more, but I personally think that when the sampling rate is too small the flame becomes almost uninofrmative (you can never assume that what's being displayed is accurate or not).

charleskawczynski:
> This job took 41 minutes: https://buildkite.com/clima/climaatmos-ci/builds/15741#018ceb9b-eaa6-420f-82d5-fb88e4b77054
> 
> We could reduce the sampling rate even more, but I personally think that when the sampling rate is too small the flame becomes almost uninofrmative (you can never assume that what's being displayed is accurate or not).

Ahhhh, that makes sense. There is an extraordinarily large number of allocations in that job (due to Krylov.jl). It's not clear that we will fix this due to the direction that our timestepper/numerics work is going (cc @dennisYatunin ). For now, let's lower the limit for that job and merge this PR.

dennisYatunin:
We could also remove that job altogether, if that would help. We might want to keep support for Krylov.jl, but only as a debugging/testing tool for the implicit solver. Since we do not intend to use this functionality in actual simulations, we do not need to test it for performance.

Sbozzolo:
Now it's the newly added diffiuson case to time out

https://buildkite.com/clima/climaatmos-ci/builds/15949#018d1960-74b8-4653-9f5d-8f9852cdd972

I'll look into reducing sampling rate for this too

PR #2481, opened 2024-01-04
Add graceful exit feature

### LINKED ISSUE #2475: Add graceful exit watchdog file

charleskawczynski:
This PR adds a feature/struct, `GracefulExit`, that allows us to edit the file `$output_dir/graceful_exit.dat` to gracefully exit a running simulation. This will allow us to inspect the integrator at any point (that we're able to stop it).

Ideally, we could restart the simulating with another function call, ~but I think we can leave that for a separate PR~. Actually, I think calling `SciMLBase.solve!` (or maybe a function in ClimaAtmos) should be sufficient.

I'm open to improving names.

Closes #2475

### DIFF
--- a/src/solver/solve.jl
+++ b/src/solver/solve.jl
@@ -6,6 +6,23 @@ struct EfficiencyStats{TS <: Tuple, WT}
     walltime::WT
 end
 
+mutable struct GracefulExit
+    path::String
+    function GracefulExit(; path)
+        mkpath(path)
+        file = joinpath(path, "graceful_exit.dat")
+        open(file, "w") do io
+            println(io, "#      To gracefully exit job associated")
+            println(io, "#      with the output path:")
+            println(io, "#           $path")
+            println(io, "#      please change the following number from")
+            println(io, "#      0 (continue running) to 1 (gracefully exit)")
+            println(io, "0")
+        end
+        return new(path)
+    end
+end
+
 simulated_years_per_day(es::EfficiencyStats) =
     simulated_years(es) / walltime_in_days(es)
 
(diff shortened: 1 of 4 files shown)


src/solver/solve.jl
```diff
@@ -6,23 +6,78 @@ struct EfficiencyStats{TS <: Tuple, WT}
...
+            exit_loop = parse(Int, line)
+        catch
+            # Let's re-set the file, in
+            # case it was incorrectly edited.
+            GracefulExit(; path = ge.path)
+        end
+        break
+    end
+    return exit_loop == 1
+end
```
Sbozzolo, line 67: Since this function is executed at the end of each step, I would suggest keeping it as simple as possible. 

Also, writing ascii files on a networked file system from different processes might lead to problems. 

I would recommend for this function to just read the first character, if it is not 0, kill the simulation gracefully. 

We should also consider the performance of opening and reading one file at each timestep on slow filesystems.

src/solver/solve.jl
```diff
@@ -6,23 +6,78 @@ struct EfficiencyStats{TS <: Tuple, WT}
     walltime::WT
 end
 
+mutable struct GracefulExit
```
Sbozzolo, line 9: Why do we need a mutable struct?

Sbozzolo (commented):
Can we implement this as a callback instead?

As in:

https://docs.sciml.ai/DiffEqDocs/stable/features/callback_functions/#Example-2:-Terminating-an-Integration

src/solver/solve.jl
charleskawczynski: Ah, that’s an artifact of my first implementation. I’ll make it immutable

charleskawczynski:
> Can we implement this as a callback instead?
> 
> As in:
> 
> https://docs.sciml.ai/DiffEqDocs/stable/features/callback_functions/#Example-2:-Terminating-an-Integration

I had that at first, but then this run mode had to be a part of the atmos cache, I thought that this might be preferable. I’m fine moving it though.

src/solver/solve.jl
charleskawczynski: That’s what I previously had but anticipated that people would want this somewhat documented, hence the comments

Sbozzolo:
> > Can we implement this as a callback instead?
> > As in:
> > https://docs.sciml.ai/DiffEqDocs/stable/features/callback_functions/#Example-2:-Terminating-an-Integration
> 
> I had that at first, but then this run mode had to be a part of the atmos cache, I thought that this might be preferable. I’m fine moving it though.

I agree that it should not be in the cache if possible. But, do we need to introduce an additional struct to achieve the behavior we want to have?

The simplest implementation for this feature might be a callback 
```julia
output_dir = simulation.output_dir
function condition(u, t, integrator)
    open(joinpath(output_dir, "graceful_exit.dat"),"r") do file
        while !eof(file)
            code = read(file, UInt8)
            return code != 0
        end
    end
end
affect!(integrator) = terminate!(integrator)
cb = DiscreteCallback(condition, affect!)
```
And we add this callback to the integrator.

I haven't thought about this too much, so maybe I am missing something.

charleskawczynski:
> > > Can we implement this as a callback instead?
...

Yes, we can use `terminate!` along with the callback. `terminate!` empties `tstops`, which is checked in the loop condition. This may not play nicely with restarting, but we can see how people like it. Also, `tstops` is a mutable array, so we could always push `t_end` back in afterwards.

simonbyrne:
See my comments on #2475: I'd like a little more clarity around what it aims to do, and how it will work. In particular, for MPI jobs there is no guarantee that that all processes will have the same view of the file system.

Sbozzolo:
Closed in favor of https://github.com/CliMA/ClimaAtmos.jl/pull/2511 ?

PR #2487, opened 2024-01-05
Fix diagnostics for longruns

Sbozzolo:
Some reductions and periods were incorrect.


szy21 (approved):
It's actually better to have instantaneous output rather than time average for longrun baroclinic waves (same as ci baroclinic waves), and plot day 10 diagnostics. Would it be easy to change? If not don't worry, I can change it later.

Sbozzolo:
> It's actually better to have instantaneous output rather than time average for longrun baroclinic waves (same as ci baroclinic waves), and plot day 10 diagnostics. Would it be easy to change? If not don't worry, I can change it later.

Yes, very easy to change.

Sbozzolo:
Folded in #2485 to reduce random CI failures.

PR #2489, opened 2024-01-05
Add option to evaluate the diagnostics on the vertical levels

Sbozzolo:
This PR introduces a new option `netcdf_output_at_levels`. When this configuration is set to true, the output diagnostics for the NetCDF files are evaluated on the vertical levels (as opposed to a collection of prescribed vertical points). This option is incompatible with `netcdf_interpolate_z_over_msl` (which is used to interpret `z` as elevation from mean sea level as opposed to elevation from the surface).

This feature is implemented using the very same infrastructure used for the general case with vertical interpolation but passing as target points the grid points. As a result, the vertical interpolation still happens and floating point arithmetic leads to evaluation that might not precisely on the grid centers, but within 0.01 % of it (as empirically measured). I chose this approach because it is much simpler to implement than not doing the interpolation at all. 

As a bonus, this PR adds the option to have a list of `short_name` in the `diagnostic` section of the YAML file, as in
```
diagnostics:
  - short_name: [pfull, wa, ra, rv]
    period: 1days
```
It also fixes https://github.com/CliMA/ClimaAtmos.jl/issues/2505.

### DIFF
--- a/src/diagnostics/netcdf_writer.jl
+++ b/src/diagnostics/netcdf_writer.jl
@@ -108,7 +108,8 @@ end
 """
     add_space_coordinates_maybe!(nc::NCDatasets.NCDataset,
                                  space::Spaces.AbstractSpace,
-                                 num_points)
+                                 num_points;
+                                 names)
 
 Add dimensions relevant to the `space` to the given `nc` NetCDF file. The range is
 automatically determined and the number of points is set with `num_points`, which has to be
@@ -117,6 +118,8 @@ a cubed sphere, 2 for a surface, 1 for a column.
 
 The function returns an array with the names of the relevant dimensions. (We want arrays
 because we want to preserve the order to match the one in num_points).
+
+In some cases, the names are adjustable passing the keyword `names`.
 """
 function add_space_coordinates_maybe! end
 
@@ -155,23 +158,24 @@ end
 function add_space_coordinates_maybe!(
     nc::NCDatasets.NCDataset,
     space::Spaces.FiniteDifferenceSpace,
-    num_points_z,
+    num_points_z;
+    names = ("z",),
 )
-    name = "z"
+    name, _... = names
     zpts = target_coordinates(space, num_points_z)
-    add_dimension_maybe!(nc, "z", zpts, units = "m")
+    add_dimension_maybe!(nc, name, zpts, units = "m", axis = "Z")
     return [name]
 end
 
 add_space_coordinates_maybe!(
     nc::NCDatasets.NCDataset,
     space::Spaces.AbstractSpectralElementSpace,
-    num_points,
+    num_points;
 ) = add_space_coordinates_maybe!(
     nc,
     space,
     num_points,
-    Meshes.domain(space.topology),
+    Meshes.domain(space.topology);
 )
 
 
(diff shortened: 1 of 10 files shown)


src/diagnostics/netcdf_writer.jl
```diff
@@ -497,13 +536,30 @@ function NetCDFWriter(;
...
 
+    if disable_vertical_interpolation
+        # It is a little tricky to override the number of vertical points because we don't
+        # know if the vertical direction is the 2nd (as in a plane) or 3rd index (as in a
+        # box or sphere). To set this value, we check if we are on a plane or not
+
+        # TODO: Get the number of dimensions directly from the space
+        num_horiz_dimensions =
+            Spaces.horizontal_space(space) isa Spaces.SpectralElementSpace1D ?
+            1 : 2
```
Sbozzolo, line 554: @simonbyrne is there a better way to do this?

src/diagnostics/netcdf_writer.jl
```diff
@@ -534,21 +590,39 @@ function write_field!(
...
+        # When we disable vertical_interpolation, we override the vertical points with
+        # the reference values for the vertical space.
+        if writer.disable_vertical_interpolation && !is_horizontal_space
+            # We need Array(parent()) because we want an array of values, not a DataLayout
+            # of Points
+            vpts = Array(
+                parent(
+                    space.grid.vertical_grid.center_local_geometry.coordinates,
+                ),
+            )
```
Sbozzolo, line 628: @simonbyrne is there a better way to do this?

I need a simple array of numbers

src/diagnostics/netcdf_writer.jl
```diff
@@ -138,40 +141,48 @@ function target_coordinates(
     S <:
     Union{Spaces.CenterFiniteDifferenceSpace, Spaces.FaceFiniteDifferenceSpace},
 }
-    # Logarithmically spaced with base e
+    # Exponentially spaced with base e
+    #
+    # We mimic something that looks like pressure levels
+    #
+    # P ~ exp(-z/H)
```
szy21, line 148:
```suggestion
    # p ~ p₀ exp(-z/H)
```

src/diagnostics/netcdf_writer.jl
```diff
@@ -138,40 +141,48 @@ function target_coordinates(
...
     FT = Spaces.undertype(space)
     vert_domain = space.topology.mesh.domain
     z_min, z_max = FT(vert_domain.coord_min.z), FT(vert_domain.coord_max.z)
-    # We are creating a log-space here, so we have to ensure that z_min > 0.
-    # If z_min = 0, we set the the first point to be at altitude of 100
-    log_z_min = z_min ≈ 0 ? FT(2) : log(z_min)
-    log_z_max = log(z_max)
-    return collect(exp.(range(log_z_min, log_z_max, num_points_z)))
+    # We add a floor of 100 meters to z_min to avoid having to deal with the singular value z = 0
+    z_min = z_min + 100
```
szy21, line 158: How about `z_min = max(z_min, 30)` or `z_min = max(z_min, H_EARTH / 300)` in case we have non-zero `z_min` in the future? And I chose ~30 because for our target resolution we have three levels below 100m. I was thinking of using H_EARTH in case we ever change it, but that's up to you.

szy21 (commented):
Thank you! Just some minor comments for now. I'll check the buildkite artifacts later.

src/diagnostics/netcdf_writer.jl
Sbozzolo: Fixed, thanks

src/diagnostics/netcdf_writer.jl
Sbozzolo: Fixed, thanks

szy21 (approved):
The artifacts look good, thanks!

src/diagnostics/netcdf_writer.jl
simonbyrne: Not easily. We do define `coordinate_type` on some objects, but not grids or spaces.

src/diagnostics/netcdf_writer.jl
simonbyrne: This works.

We probably should define an accessor function for getting the "reference" vertical grid.

src/diagnostics/netcdf_writer.jl
Sbozzolo: Okay, thanks!

src/diagnostics/netcdf_writer.jl
Sbozzolo: I agree.

PR #2503, opened 2024-01-09
store the hyperdiffusion coefficient pre-length scaling

simonbyrne:
## Purpose 
@szy21 @charleskawczynski This should avoid the need to require the length scale at configuration time.

cf https://github.com/CliMA/ClimaCore.jl/pull/1588#issuecomment-1883953591

## Content

----

### DIFF
--- a/config/longrun_configs/longrun_aquaplanet_rhoe_equil_55km_nz63_clearsky_0M_earth.yml
+++ b/config/longrun_configs/longrun_aquaplanet_rhoe_equil_55km_nz63_clearsky_0M_earth.yml
@@ -10,8 +10,8 @@ z_max: 55000.0
 rayleigh_sponge: true
 topography: "Earth"
 topo_smoothing: true
-kappa_4_vorticity: 4.0
-kappa_4_scalar: 4.0
+vorticity_hyperdiffusion_coefficient: 4.0
+scalar_hyperdiffusion_coefficient: 4.0
 moist: "equil"
 precip_model: "0M"
 surface_setup: "DefaultMoninObukhov"
--- a/src/prognostic_equations/hyperdiffusion.jl
+++ b/src/prognostic_equations/hyperdiffusion.jl
@@ -58,7 +58,15 @@ NVTX.@annotate function hyperdiffusion_tendency!(Yₜ, Y, p, t)
     (; hyperdiff, turbconv_model) = p.atmos
     isnothing(hyperdiff) && return nothing
 
-    (; κ₄_vorticity, κ₄_scalar, divergence_damping_factor) = hyperdiff
+    (; ν₄_vorticity_coeff, ν₄_scalar_coeff, divergence_damping_factor) =
+        hyperdiff
+
+    h_space = Spaces.horizontal_space(axes(Y.c))
+    h_length_scale = Spaces.node_horizontal_length_scale(h_space) # mean nodal distance
+
+    ν₄_scalar = ν₄_scalar_coeff * h_length_scale^3
+    ν₄_vorticity = ν₄_vorticity_coeff * h_length_scale^3
+
     n = n_mass_flux_subdomains(turbconv_model)
     diffuse_tke = use_prognostic_tke(turbconv_model)
     ᶜJ = Fields.local_geometry_field(Y.c).J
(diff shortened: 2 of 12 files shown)


src/solver/model_getters.jl
```diff
@@ -36,28 +36,26 @@ end
 
 function get_hyperdiffusion_model(parsed_args, spaces, ::Type{FT}) where {FT}
     hyperdiff_name = parsed_args["hyperdiff"]
-    h_space = Spaces.horizontal_space(spaces.center_space)
-    h_length_scale = Spaces.node_horizontal_length_scale(h_space) # mean nodal distance
-    κ₄_vorticity = FT(parsed_args["kappa_4_vorticity"]) * h_length_scale^3
-    κ₄_scalar = FT(parsed_args["kappa_4_scalar"]) * h_length_scale^3
+    ν₄_vorticity = FT(parsed_args["kappa_4_vorticity"])
+    ν₄_scalar = FT(parsed_args["kappa_4_scalar"])
```
simonbyrne, line 40: may want to rename these arguments?

szy21:
This will break the `CAM_SE`, as it is setting the actual hyperdiffusivity `κ₄` instead of the coefficient `ν₄`.

In general, this change means we will do the calculation every time we call hyperdiffusion tendency, which is not necessary. But I guess it is cheap enough?

src/solver/model_getters.jl
```diff
@@ -36,28 +36,26 @@ end
...
     elseif hyperdiff_name in ("CAM_SE",)
         # To match hyperviscosity coefficients in:
         #    https://agupubs.onlinelibrary.wiley.com/doi/epdf/10.1029/2017MS001257
         #    for equation A18 and A19
-        Ne = parsed_args["h_elem"]
-        κ₄_vorticity = FT(0.15 * (30 / Ne * 1.1 * 10^5)^3) # ν_vort
-        κ₄_scalar = FT(0.75 * (30 / Ne * 1.1 * 10^5)^3) # ν_q
+        # Need to scale by (1.1e5 / (sqrt(4 * pi / 6) * 6.371e6 / (3*30)) )^3  ≈ 1.238
+        ν₄_vorticity = FT(0.150 * 1.238) # ν_vort
+        ν₄_scalar = FT(0.751  * 1.238) # ν_q
```
simonbyrne, line 54: @szy21 this should be equivalent

simonbyrne:
> In general, this change means we will do the calculation every time we call hyperdiffusion tendency, which is not necessary. But I guess it is cheap enough?

Yes, it should be pretty cheap. I wouldn't worry about it.

szy21:
Sounds good. What’s the advantage of this over the current implementation though?

simonbyrne:
It simplifies some of the logic, and you don't need https://github.com/CliMA/ClimaCore.jl/pull/1588

szy21:
We don't need the climacore PR now either. That PR is just trying to make it more convenient to get the length scale I guess. With this change we don't need space information for `get_atmos`. I guess that's the advantage?

src/solver/model_getters.jl
szy21: how about `kappa_4_vorticity_coeff` and `kappa_4_scalar_coeff`, and `κ₄_vorticity_coeff` and `κ₄_scalar_coeff` instead of `ν₄`? I'm also ok with renaming all `κ₄` to `ν₄` (it may be even better).

szy21 (approved):

config/longrun_configs/longrun_aquaplanet_rhoe_equil_55km_nz63_clearsky_0M_earth.yml
```diff
@@ -10,8 +10,8 @@ z_max: 55000.0
 rayleigh_sponge: true
 topography: "Earth"
 topo_smoothing: true
-kappa_4_vorticity: 4.0
-kappa_4_scalar: 4.0
+vector_hyperdiffusion_coefficient: 4.0
+scalar_hyperdiffusion_coefficient: 4.0
```
simonbyrne, line 14: @szy21 how about these as names? they seem a bit more descriptive

szy21: This looks better I think. But maybe `vorticity_hyperdiffusion_coefficient` instead of `vector_hyperdiffusion_coefficient` because the grad div term in momentum hyperdiffusion is also a vector but uses vorticity_hyperdiffusion_coefficient * divergence damping factor.

In the code I think it would be more clear to name them something like `ν₄_vorticity_coeff' and `ν₄_vorticity`, instead of `ν₄` being the coefficient and `κ₄` being the actual value.

simonbyrne:
@szy21 how about this?

szy21:
Looks great, thank you!

src/solver/model_getters.jl
```diff
@@ -36,32 +36,31 @@ end
 
 function get_hyperdiffusion_model(parsed_args, spaces, ::Type{FT}) where {FT}
```
szy21, line 37: Ah, I forgot to comment on this. We can remove `spaces` now?

szy21: And we can remove `spaces` in get_atmos too.

src/prognostic_equations/hyperdiffusion.jl
```diff
@@ -134,39 +142,41 @@ NVTX.@annotate function hyperdiffusion_tendency!(Yₜ, Y, p, t)
...
     end
 end
 
 NVTX.@annotate function tracer_hyperdiffusion_tendency!(Yₜ, Y, p, t)
     (; hyperdiff, turbconv_model) = p.atmos
     isnothing(hyperdiff) && return nothing
 
-    (; κ₄_scalar) = hyperdiff
+    (; ν₄_scalar_coeff) = hyperdiff
+    h_length_scale = Spaces.node_horizontal_length_scale(h_space) # mean nodal distance
```
szy21, line 178:
```suggestion
    h_space = Spaces.horizontal_space(axes(Y.c))
    h_length_scale = Spaces.node_horizontal_length_scale(h_space) # mean nodal distance
```

PR #2506, opened 2024-01-11
Turn off save to disk for GPU aquaplanet DYAMOND simulation

sriharshakandala:
## Purpose 
Turn off save to disk for GPU aquaplanet DYAMOND simulation

----


Sbozzolo:
What's the purpose of this change?

sriharshakandala:
> What's the purpose of this change?

The GPU pipeline is primarily for looking at the performance benchmarks. We would want to turnoff file writes for measurements! We can test IO in the general CI pipeline and/or longrun cpu/gpu pipeline!

sriharshakandala:
We can have a separate GPU pipeline focussed on IO, if we would like to tune IO performance!

Sbozzolo:
Implemented in #2616

PR #2511, opened 2024-01-11
Add graceful exit feature as a callback

charleskawczynski:
Supersedes #2481.


simonbyrne:
Can you add a comment or two on how this works? i.e. what should the file be named, how it should work? MPI limitations?

src/callbacks/callbacks.jl
```diff
@@ -301,3 +301,25 @@ function gc_func(integrator)
...
+    if isfile(file)
+        open(file, "r") do io
+            while !eof(io)
+                try
+                    code = parse(Int, read(io, Char))
+                    return code != 0
+                catch
+                    open(io -> print(io, 0), file, "w")
+                    return false
+                end
```
simonbyrne, line 317: what does this do?

charleskawczynski: If the file is somehow corrupted, this "resets" the file to a "continue running" state.

simonbyrne: why not just make it "if the file exists?". Then you can just do
```
touch outdir/exit
```

simonbyrne: (you would also want to remove the file if it exists at the start, i.e. if re-running the simulation)

Sbozzolo: I think that reading a number from a file will open up more control options in the future. E.g., we can assign different actions to different numbers.

simonbyrne: Sure, I don't have strong opinions on this. I was only pointing out its very easy to create an empty file.

src/callbacks/callbacks.jl
```diff
@@ -301,3 +301,35 @@ function gc_func(integrator)
...
+
+"""
+    maybe_graceful_exit(integrator)
+
+This callback is called after every timestep
+to allow users to gracefully exit a running
+simulation. To do so, users can navigate to
+and open `$output_dir/graceful_exit.dat`, change
+the file contents from 0 to 1, and the running
+simulation will gracefully exit with the integrator.
```
simonbyrne, line 313:
```suggestion
This callback is called after every timestep
to allow users to gracefully exit a running
simulation. To do so, users can navigate to
and open `$output_dir/graceful_exit.dat`, change
the file contents from 0 to 1, and the running
simulation will gracefully exit with the integrator.

!!! note
    This may not be reliable for MPI jobs.
```

src/callbacks/callbacks.jl
charleskawczynski: Ok, I'll leave it as it is.

PR #2512, opened 2024-01-11
Fix job-id for consistency with existing longruns

akshaysridhar:
Fix job id to match plot identifier.
Current main results in 
```Warning: No plot found for Val{:longrun_aquaplanet_rhoe_equil_gray_55km_nz63_0M}()```


szy21 (commented):
Actually could you change the job name to `longrun_aquaplanet_rhoe_equil_55km_nz63_gray_0M` instead? I want to make the name consistent.

szy21 (approved):
Thank you!

PR #2524, opened 2024-01-13
add option for 0 or 1 cloud fraction

### LINKED ISSUE #2054: Use cloud fraction and in-cloud lwp and iwp in radiation
We should use partial cloud fraction for radiation calculation. Note that the lwp and iwp used as input for RRTMGP [here](https://github.com/CliMA/ClimaAtmos.jl/blob/367849d8d8d6fcda88b8bab3ef86afbf1e5ee4d5/src/callbacks/callbacks.jl#L200-L203) should be in-cloud instead of grid-mean. That is, we should divide grid-mean values by cloud fraction.

szy21:
## Purpose 

Adds a cloud model which specifies whether the cloud properties are calculated from grid-scale or quadratures. Currently this is only used for cloud fraction, but we can use it for liquid water and ice too. Also changes the radiation input to cloud fraction and in-cloud lwp and iwp.

Closes #2054 

## Content

----

### DIFF
--- a/src/callbacks/callbacks.jl
+++ b/src/callbacks/callbacks.jl
@@ -38,7 +38,7 @@ end
 function cloud_fraction_model_callback!(integrator)
     Y = integrator.u
     p = integrator.p
-    set_cloud_fraction!(Y, p, p.atmos.moisture_model)
+    set_cloud_fraction!(Y, p, p.atmos.moisture_model, p.atmos.cloud_model)
 end
 
 NVTX.@annotate function rrtmgp_model_callback!(integrator)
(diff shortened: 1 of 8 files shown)


src/callbacks/callbacks.jl
```diff
@@ -166,10 +166,14 @@ NVTX.@annotate function rrtmgp_model_callback!(integrator)
         )
         # multiply by 1000 to convert from kg/m^2 to g/m^2
         @. ᶜlwp =
-            1000 * Y.c.ρ * TD.liquid_specific_humidity(thermo_params, ᶜts) * ᶜΔz
+            1000 *
```
tapios, line 169: Are we doing this unit conversion because RRTMG needs it in g/m^2? If so, it's better to replace 1000 by `kg_to_g_conversion` or something like it. Otherwise, sticking with SI units is better.

tapios (approved):
Looks good from what I can tell.

src/callbacks/callbacks.jl
szy21: Yes RRTMG needs g/m^2, I'll replace it.

PR #2523, opened 2024-01-13
modify diagnostics in config files

szy21:
## Purpose 

Uses the recently added group feature. Adds cloud fraction to default diagnostics. Moves diagnostic edmf aquaplanet to GPU pipeline.

## Content

----
- [ ] I have read and checked the items on the review checklist.

PR #2525, opened 2024-01-13
fix diagnostics in some model configs

szy21:
## Purpose 

Fixes some typos in diagnostics in configs introduced in #2523. Also removes `post_process` as we don't use it anymore. 

## Content

----

### DIFF
--- a/config/default_configs/default_config.yml
+++ b/config/default_configs/default_config.yml
@@ -123,9 +123,6 @@ reference_job_id:
   help: |-
     Identifier of job to use as the "reference" solution in the quicklook plot; the current job's results get compared to the results of the quicklook job on the main branch
   value: ~
-post_process:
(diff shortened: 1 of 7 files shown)


config/default_configs/default_config.yml
```diff
@@ -123,9 +123,6 @@ reference_job_id:
   help: |-
     Identifier of job to use as the "reference" solution in the quicklook plot; the current job's results get compared to the results of the quicklook job on the main branch
   value: ~
-post_process:
```
szy21, line 126: @Sbozzolo just want to double check we don't use this anymore.

Sbozzolo: I didn't even know it was a thing. Yes, I don't think it is used anywhere now.

Sbozzolo (approved):

PR #2531, opened 2024-01-16
modify some configs

szy21:
## Purpose 

- Adds implicit diffusion config for `default_perf`, `longrun_aquaplanet_amip` and `diagnostic_edmfx_aquaplanet`
- Removes `Float32` in some longrun configs as that is the default.

## Content

----

### DIFF
--- a/config/perf_configs/jet_n_failures.yml
+++ b/config/perf_configs/jet_n_failures.yml
@@ -1,1 +1,2 @@
 job_id: "jet_n_failures"
+implicit_diffusion: false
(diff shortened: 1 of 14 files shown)


config/perf_configs/jet_n_failures.yml
```diff
@@ -1 +1,2 @@
 job_id: "jet_n_failures"
+implicit_diffusion: false
```
charleskawczynski, line 2: @szy21, why was this added to this jet job?

szy21: Oh, if it's not clear, `implicit_diffusion` has been always false in this jet job. But I set it to true in the default perf config in this PR, so this line is to set it to false again.

PR #2533, opened 2024-01-17
Move vertical advection of TKE to the implicit tendency

dennisYatunin:
## Purpose 

This PR moves the vertical advection of TKE from `explicit_vertical_advection_tendency!` to `implicit_vertical_advection_tendency!`. After this is merged in, we will also move the dissipation of TKE to the implicit tendency.

----


szy21:
We should modify the Jacobian as well :)

dennisYatunin:
We've decided to keep the advection of TKE explicit, since making it implicit would require computing derivatives of the diagnostic EDMF integrals with respect to grid-mean quantities. I'm closing this PR.

PR #2536, opened 2024-01-18
Move dissipation to the implicit tendency

### LINKED ISSUE #2498: Make TKE dissipation implicit

dennisYatunin:
## Purpose 

This PR moves the dissipation of TKE from `edmfx_tke_tendency!` to `edmfx_sgs_diffusive_flux_tendency!`, and it updates the Jacobian appropriately. It also simplifies the `dt`-based correction to dissipation.

Closes #2498 

----


szy21:
This breaks prognostic EDMF as TKE dissipation is moved out of `edmfx_tke_tendency!`, which is applied to both prognostic and diagnostic EDMF, but only added to `edmfx_sgs_diffusive_flux!` for diagnostic EDMF. I think it would be better for now to just have separate tendencies for TKE dissipation for prognostic and diagnostic EDMF. We can work on this together tomorrow.

src/prognostic_equations/edmfx_sgs_flux.jl
```diff
@@ -230,7 +233,14 @@ function edmfx_sgs_diffusive_flux_tendency!(
                 bottom = Operators.SetValue(ρatke_flux[colidx]),
             )
             @. Yₜ.c.sgs⁰.ρatke[colidx] -=
-                ᶜdivᵥ_ρatke(-(ᶠρaK_u[colidx] * ᶠgradᵥ(ᶜtke⁰[colidx])))
+                ᶜdivᵥ_ρatke(-(ᶠρaK_u[colidx] * ᶠgradᵥ(ᶜtke⁰[colidx]))) +
+                dissipation(
```
szy21, line 237: maybe better to name it something like `tke_dissipation`

src/prognostic_equations/edmfx_sgs_flux.jl
```diff
@@ -342,3 +362,6 @@ function edmfx_sgs_diffusive_flux_tendency!(
 
     return nothing
 end
+
+dissipation(ρatke⁰, tke⁰, l, c_d, dt) =
+    tke⁰ >= 0 ? c_d * ρatke⁰ * sqrt(tke⁰) / max(l, 1) : ρatke⁰ / dt
```
szy21, line 367: Just note that this is assuming ρatke = ρa * tke, which is not necessarily true because of `divided_by_ρa`, but I think it's ok.

szy21 (approved):

szy21:
Feel free to merge it. I couldn't find which equations in the documentation correspond to the code, but we can look at it together tomorrow:)

PR #2537, opened 2024-01-18
tag v0.19.1

juliasloan25:
## Purpose 
use ClimaCore 0.11.8 (bug fixes)
tag ClimaAtmos v0.19.1 (includes update to julia v1.10 needed by coupler)

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "ClimaAtmos"
 uuid = "b2c96348-7fb7-4fe0-8da9-78d88439e717"
 authors = ["Climate Modeling Alliance"]
-version = "0.19.0"
+version = "0.19.1"
 
 [deps]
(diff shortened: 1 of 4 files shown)


szy21 (approved):
I saw that when we bumped ClimaAtmos version before we didn't change ClimaAtmos version in Manifest.toml. Is that something we should do? Or is there a difference between minor and patch releases?

juliasloan25:
> I saw that when we bumped ClimaAtmos version before we didn't change ClimaAtmos version in Manifest.toml. Is that something we should do? Or is there a difference between minor and patch releases?

Hmm, when I make a release I always update the checked-in Manifests. To me it seems like a good idea to make sure we're using the right version of the package in subdirectories. I don't think there's a difference with this between minor and patch releases, and I see different people use different approaches. I can ask on slack if we have a preferred convention

PR #2538, opened 2024-01-18
Use SetDivergence boundary conditions for precipitation

dennisYatunin:
## Purpose 

This PR replaces the `NullBoundaryCondition()` at the bottom of the `DivergenceF2C` operator for precipitation with a `SetDivergence(0)` boundary condition. This should slightly change the results for simulations with 1-moment microphysics.

This also slightly simplifies the definition of the implicit tendency for precipitation, and it fixes an issue with how ClimaCore v0.11.9 was added to the manifest files.

----

### DIFF
--- a/config/model_configs/sphere_aquaplanet_rhoe_equilmoist_allsky_gw_res.yml
+++ b/config/model_configs/sphere_aquaplanet_rhoe_equilmoist_allsky_gw_res.yml
@@ -4,7 +4,9 @@ dz_bottom: 300.0
 dt: "400secs"
 t_end: "24hours"
 dt_save_state_to_disk: "24hours"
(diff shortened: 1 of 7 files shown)


trontrytel (approved):
Thank you!

dennisYatunin:
Looks like this makes the one simulation with 1-moment microphysics crash. I'm reducing its `t_end` from `24hours` to `22hours`.

dennisYatunin:
Instead of reducing `t_end` for that job, I've modified it to use implicit Frierson diffusion. I've also reduced the `t_end` of the moist baroclinic wave with implicit diffusion from 4 days to 6 days in order to prevent it from slowing down CI.

PR #2543, opened 2024-01-18
Add spectra in postprocessing pipeline

Sbozzolo:
The number of points was changed because ClimaCoreSpectra seem to require one dimension to be twice as much as to the other

### DIFF
--- a/post_processing/ci_plots.jl
+++ b/post_processing/ci_plots.jl
@@ -4,6 +4,8 @@ import ClimaAnalysis
 import ClimaAnalysis: Visualize as viz
 import ClimaAnalysis: SimDir, slice_time, slice
 
+import ClimaCoreSpectra: power_spectrum_2d
+
 using Poppler_jll: pdfunite
 import Base.Filesystem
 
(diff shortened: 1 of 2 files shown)


Sbozzolo:
@akshaysridhar, would you like to add the spectra you are interested in producing to `ci_plots.jl`?

To do that, just add a function call like
```julia
make_spectra_generic(
    simulation_path,
    vars;
    slicing_kwargs = ClimaAnalysis.Utils.kwargs(x = 0.0, time = LAST_SNAP),
)
```
to the relevant `make_plots` function.

akshaysridhar:
@Sbozzolo The array size appears to be an assumption within the ClimaCore library - e.g. https://github.com/CliMA/ClimaCore.jl/blob/c83abb94cebc82046a392ba423cedd3681d2a93a/lib/ClimaCoreSpectra/test/gcm_visual_test.jl#L25. Otherwise this LGTM. Thanks! cc @LenkaNovak

Sbozzolo:
Do you want the PR merge as it is, or would you like to add a commit to this PR that computes with your favorite spectra?

akshaysridhar:
> Do you want the PR merge as it is, or would you like to add a commit to this PR that computes with your favorite spectra?

Merge as is I think; we can then keep updating the post-proc functions (same as choice of current default diagnostic outputs) as we add spectra for tke, and other variables.

Issue #2530, opened 2024-01-16
Improve the performance of buoyancy_gradients


szy21:
#2456 makes all the dycore simulations slower by a factor of two. When looking at it more closely, it seems most of the slowdown is from [buoyancy_gradients](https://github.com/CliMA/ClimaAtmos.jl/blob/dcbe507bf0962aae88590423d688f91986b14942/src/prognostic_equations/buoyancy_gradients.jl#L13). As a temporary solution, the mixing length and thus buoyancy gradients calculation is moved to a callback in #2466 without EDMF.  With EDMF, this is still called at each timestep. It would be good if we can improve the performance of buoyancy_gradients. cc @charleskawczynski  
```[tasklist]
### Tasks
- [x] Reduce number of times that local geometry is explicitly passed as an argument (https://github.com/CliMA/ClimaAtmos.jl/pull/2367, https://github.com/CliMA/ClimaAtmos.jl/pull/2544)
- [x] Analyze the performance of `cloud_fraction_model_callback!`
- [x] Add cloud diagnostics to flame graph (#2668)
- [x] Hoist out ᶜgradᵥ(ᶠinterp(ϕ)) for both broadcast expressions, and see where that gets us. Done in #2673
- [x] Reduce size of structs to reduce memory throughput https://github.com/CliMA/ClimaAtmos.jl/pull/2680.
- [x] Understand and report why performance was so poor for these kernels (perhaps open issues in ClimaCore). Answer: I think the quad loop is simply compute-intensive (9 SA calls is a lot), and there are no metric terms, so this kernel is relatively simple on the ClimaCore side. For the buoyancy gradients, we could see some improvement by using shared memory (right now, we double the number of loads/stores in the FD operators for 2-point stencils).
- [ ] (optional) Apply RootSolvers/Thermodynamics optimizations (e.g., CliMA/RootSolvers.jl#51) to reduce the arithmetic intensity of those kernels (they are very expensive)
- [ ] Add GPU job that exercises `cloud_fraction_model_callback!`: https://github.com/CliMA/ClimaAtmos.jl/pull/2684
- [x] (optional) Inline all thermo functions to allow for the compiler to perform Common Subexpression Elimination (CSE). On the CPU, it's required that we first fix https://github.com/CliMA/ClimaCore.jl/issues/1602 so that `PhasePartition` is inlined. We already always inline on the GPU, so this will have no impact there.
```

charleskawczynski:
[...]

PR #2544, opened 2024-01-18
Try improving buoyancy gradient kernel performance

charleskawczynski:
This PR attempts to further improve the performance of the buoyancy gradient kernel.


charleskawczynski:
Should help with #2530.

PR #2550, opened 2024-01-19
Introduce z_reference in NetCDF for topo runs

Sbozzolo:
Now, diagnostics for topography runs defines a new 1D dimension (`z_reference`) and provides a 3D `z` variable.


Sbozzolo:
@szy21 do you want to give this brach a try?

It should be fully working in all cases.

szy21:
It works great, thanks!

szy21 (approved):

PR #2558, opened 2024-01-21
Use more ClimaAnalysis in post of precipitation test

Sbozzolo:
This PR reworks the post processing step for `single_column_precipititation_test`. Previously, the function had to compute the name for the labels and hardcoded the variables that were plotted. Now, a new variable can be added simply by adding it to the `vars` array and no processing is needed for labels/names.

This PR also demonstrates how to plot two variables on the same axis. CC: @nefrathenrici


trontrytel (approved):
Thank you!

PR #2559, opened 2024-01-22
Add 0M microphysics to prognostic edmf

trontrytel:
This PR adds 0M microphysics to prognostic edmf


szy21:
The code change looks good, but I need to spend some time deriving the terms for the advective equation. The source terms in the overleaf eq. 18c and 18.d look complicated and different from the code. Maybe we can do it together sometime next week.

trontrytel:
> The code change looks good, but I need to spend some time deriving the terms for the advective equation. The source terms in the overleaf eq. 18c and 18.d look complicated and different from the code. Maybe we can do it together sometime next week.

Sounds good. And yes, it definitely needs more thought. I wanted to keep it as a draft and go over the equations together.

szy21 (approved):

PR #2562, opened 2024-01-22
Update breaking parameter changes 

nefrathenrici:
Content
- Update compat for insolation, cloudmicrophys, thermo, and surfacefluxes
- Update `create_parameter_set` for the CLIMAparameters changes
- Remove all `alias` fields from the toml files, rename `entr_tau` and `detr_tau` in toml files
- Change `updraft_number` to be configured in `default_edmf_config.yml` instead of retrieved from CLIMAParameters
- Add checks for unused parameters (in case of typos or other errors) and logs all used parameters
- Does not update RRTMGP or fully update Thermodynamics. RRTMGP update will be split off into another PR.
- Update lookahead GHA to dev before instantiating


sriharshakandala:
In addition to above, we should no longer need to convert to `Float64` for running `RRTMGP.jl`. We should use consistent Floating point precision throughout. Can we also add this change in this PR?

Project.toml
```diff
@@ -46,20 +46,20 @@ ArgParse = "1"
 ArtifactWrappers = "0.2"
 Artifacts = "1"
 AtmosphericProfilesLibrary = "0.1"
-CLIMAParameters = "0.7.25"
+CLIMAParameters = "0.8.1"
+CPUSummary = "0.2"
```
charleskawczynski, line 50: We shouldn't need to add this as a dependency?

charleskawczynski: Oh, I see that it just got moved

src/parameterized_tendencies/radiation/radiation.jl
```diff
@@ -185,19 +185,19 @@ function radiation_model_cache(
...
-            solar_zenith_angle = weighted_irradiance = NaN # initialized in callback
+            cos_zenith = weighted_irradiance = NaN # initialized in callback
         end
 
         radiation_model = RRTMGPI.RRTMGPModel(
             rrtmgp_params,
             data_loader,
             # TODO: pass FT through so that it can be controlled at the top-level
-            Float64,
+            FT,
```
charleskawczynski, line 200: Let's maybe apply this change in a separate PR, since this will impact our Float32 jobs

charleskawczynski:
This looks great, for the multi-gpu job, this error looks a bit odd. Any thoughts, @simonbyrne?

src/solver/type_getters.jl
```diff
@@ -778,10 +778,16 @@ function get_simulation(config::AtmosConfig)
...
-    end
+    # Check that all set parameters have been used
+    param_filepath =
+        joinpath(sim_info.output_dir, "$(sim_info.job_id)_parameters.toml")
+    CP.log_parameter_information(
+        config.toml_dict,
+        param_filepath,
+        strict = true,
+    )
+    !config.parsed_args["log_params"] && rm(param_filepath)
```
charleskawczynski, line 789:
```suggestion
    if ClimaComms.iamoot(config.comms_ctx)
      config.parsed_args["log_params"] || rm(param_filepath)
    end
```
Maybe this is the issue? @simonbyrne pointed out that there are currently multiple jobs with the same job id, though.

nefrathenrici: If there are multiple jobs with the same ID this is definitely the problem.

nefrathenrici: Is it ok to rename the job at `config/gpu_configs/gpu_hs_rhoe_equil_55km_nz63_0M_4process.yml`? 

I can also add a test for unique job IDs so that this doesn't happen again.

charleskawczynski: Yeah, I think that's fine. I thought I had added such a test on builkite, actually, but I don't see it. Ah, maybe I'm thinking of `Ensure mse tables are reset when necessary`. Adding such a test would be great.

test/config.jl
```diff
@@ -27,3 +27,32 @@ include(joinpath("..", "perf", "common.jl"))
...
+            data = YAML.load_file(filepath)
+            if haskey(data, "job_id")
+                job_id_dict[filepath] = data["job_id"]
+            end
+        end
+    end
+    return job_id_dict
+end
+
+file_to_job_id = extract_job_ids("config")
```
charleskawczynski, line 45: I kind of expected that, somewhere, we'd use / test with `Base.allunique`

charleskawczynski: Can this be simplified?

charleskawczynski: This PR is overall a huge upgrade, so I'm going to enable merge now, but maybe we can clean this up later.

nefrathenrici: I will need to update RRTMGP and Climaparameters anyway, I can do it then.

PR #2565, opened 2024-01-23
Change default netcdf interpolation resolution

szy21:
## Content

----


szy21:
@Sbozzolo This resolution makes more sense for our current setup. Are you ok with changing it? Or is it better if we have a separate branch for the single column? Or should we make it a configuration?

Sbozzolo:
No objections.

szy21:
> No objections. Single columns should be fixed by making them actual single columns instead of 3D spaces.

Thanks! And I guess it makes sense that this increase diagnostic allocation by a factor of 4: https://buildkite.com/clima/climaatmos-ci/builds/16084#018d3403-ae99-4a5a-9a86-26eff086730a?

PR #2567, opened 2024-01-23
Track precipitation energy in conservation check

### LINKED ISSUE #2458: Investigate conservation issues in aquaplanet runs
In #2399 , we found that the target aquaplanet case exhibits growing water and energy conservation errors (by introducing `longrun_aquaplanet_rhoe_equilmoist_0M_clearsky_tvinsolation_slabocean` in longruns). We need to understand and correct this leakage and track it in our Buildkite runs, which only track the moist baroclinic wave (`sphere_baroclinic_wave_rhoe_equilmoist_conservation_ft64`).

## Testing numerical integrals
- `sum(3d_field)` v `sum(column_integral_definite!(3d_field))`
- within floating point precision, using this code:
```
using ClimaCore: Operators, Fields, Geometry, Spaces
using ClimaCore.Utilities: half

ᶜ3d = integrator.p.precipitation.ᶜ3d_rain
col_integrated_rain = zeros(axes(Fields.level(Geometry.WVector.(integrator.u.f.u₃), half)))

Fields.bycolumn(axes(ᶜ3d)) do colidx
    Operators.column_integral_definite!(
            col_integrated_rain[colidx],
            ᶜ3d[colidx],)
end

sum_from_col = sum(col_integrated_rain ./ Fields.dz_field(axes(col_integrated_rain)) .* 2)
sum_direct = sum(ᶜ3d)
```

### Float64
- problem: using turbulent fluxes causes severe nonconservation (with mostSF is O(e-5) and without SF O(e-16) )
see `slabocen_test_frierdiff_implicit_ft64_0turbflux` vs 
`slabocen_test_frierdiff_implicit_ft64` [build](https://buildkite.com/clima/climaatmos-ci/builds/16030)
- NB: 
- zero drag error: e-6
- zero lhr, evap: e-16
- zero lhr, evap, init q_tot: e-6
- zero lhr, evap, init q_tot, zero precip: e-16
- zero lhr, evap, init q_tot, ᶜS_ρq_tot[colidx] * e_tot_0M_precipitation_sources_helper(...): e-9

### Other checks
- FT32 (with mostSF is O(e-5) and without SF O(e-7) )
- vert_diffusion v frierson: similar error (e-16 for non precipitating setups)
- moist hyperdiffusion is conservative
- MOST: no substantial error difference between mostSF and bulkSF O

# Summary
- the energy conservation issues only occur in moist setups, namely (and exclusively) where precipitation is nonzero
- the linked PR includes accounting for energy change due to precipitation in the conservation check. This improves the error by 3 orders of magnitude. 
- further possible improvements include writing a more robust conservation checker, independent of the tendency code (e.g., as done in ClimaCoupler.ConservationChecker). Since this includes further development and performance assessment, we leave this for another issue (#2568). 

## Linked PR Components
- [x] tracks precipitation term in conservation_check
- [x] converts precipitation to kg / m2 / s (consistent with evaporation)
- [x] add more diagnostics to the slab ocean experiments
- [x] adds a case with the slab ocean and conservation checks to CI
- [x] modifies the conservation check tolerances to be more stringent

## QA
- [x] CI passes 
- [x] low resol. longrun passes
- [x] AMIP resol. slabocean longrun passes ([build](https://buildkite.com/clima/climaatmos-longruns/builds/1407#018d5560-9d6c-41e0-aa1f-74b8833031f1))

LenkaNovak:
## Purpose 

closes #2458 

----


examples/hybrid/driver.jl
```diff
@@ -154,8 +135,15 @@ if config.parsed_args["check_conservation"]
...
+
+    energy_net =
+        abs(
+            energy_atmos_change + energy_surface_change -
+            energy_radiation_input,
+        ) / energy_total
+    @info "    Net energy change: $energy_net"
     @test (energy_atmos_change + energy_surface_change) / energy_total ≈
-          energy_radiation_input / energy_total atol = 5 * sqrt(eps(FT))
+          energy_radiation_input / energy_total atol = sqrt(eps(FT))
```
szy21, line 146: Is this equal to `@test (energy_net / energy_total) ≈ 0 atol = sqrt(eps(FT))`? This might be more clear especially as you already have energy_net.

examples/hybrid/driver.jl
```diff
@@ -227,3 +219,27 @@ if config.parsed_args["check_precipitation"]
         @test maximum(sol.prob.p.precomputed.ᶜcloud_fraction[colidx]) <= FT(1)
     end
 end
+
+# Visualize the solution
```
szy21, line 223: I wonder if it is better to move it back, as we may still want to visualize a simulation even if the conservation check fails. Not a big issue though, so up to you. The artifact failure should be fixed soon.

post_processing/ci_plots.jl
```diff
@@ -475,6 +475,59 @@ function make_plots(
     )
 end
 
+function make_plots(
```
szy21, line 478: Just a note that @Sbozzolo and I will fix the longrun output frequency soon, so this will not be needed. We can merge this for now.

src/cache/cache.jl
```diff
@@ -142,6 +146,14 @@ function build_cache(Y, atmos, params, surface_setup, sim_info)
     net_energy_flux_toa = [Geometry.WVector(FT(0))]
     net_energy_flux_sfc = [Geometry.WVector(FT(0))]
 
+    conservation_check =
+        atmos.surface_model isa PrognosticSurfaceTemperature ?
+        (;
+            col_integrated_precip_energy_tendency = zeros(
+                axes(Fields.level(Geometry.WVector.(Y.f.u₃), half)),
+            )
+        ) : (; col_integrated_precip_energy_tendency = (;))
```
szy21, line 155: It might be good to always have this cache. We will need it for atmosphere only energy check anyway.

src/parameterized_tendencies/microphysics/precipitation.jl
```diff
@@ -163,6 +160,17 @@ function precipitation_tendency!(
                     ᶜts[colidx],
                     ᶜΦ[colidx],
                 )
+            if p.atmos.surface_model isa PrognosticSurfaceTemperature
```
szy21, line 163: Same as before, it might be good to always calculate this.

src/prognostic_equations/surface_temp.jl
```diff
@@ -37,25 +36,37 @@ function surface_temp_tendency!(Yₜ, Y, p, t, slab::PrognosticSurfaceTemperatur
...
+        earth_area = CA.horizontal_integral_at_boundary(zeros(axes(pet)) .+ 1)
+        if p.atmos.precip_model isa Microphysics0Moment
+            # correction distributed over all surface average
+            @. Yₜ.sfc.T -=
+                sum(pet) / (ρ_ocean * cp_ocean * depth_ocean) / earth_area
+        end
+
+        # @. Yₜ.sfc.T -=
+        #     p.conservation_check.col_integrated_precip_energy_tendency /
+        #     (ρ_ocean * cp_ocean * depth_ocean)
```
szy21, line 56: What would happen if you do this? They should be equivalent? If this works I would prefer this, as it is cleaner.

src/parameterized_tendencies/microphysics/precipitation.jl
```diff
@@ -27,9 +27,9 @@ precipitation_tendency!(Yₜ, Y, p, t, colidx, ::NoPrecipitation) = nothing
 function precipitation_cache(Y, precip_model::Microphysics0Moment)
     FT = Spaces.undertype(axes(Y.c))
     return (;
-        ᶜS_ρq_tot = similar(Y.c, FT),
-        ᶜ3d_rain = similar(Y.c, FT),
-        ᶜ3d_snow = similar(Y.c, FT),
+        ᶜS_ρq_tot = zeros(axes(Y.c)),
+        ᶜ3d_rain = zeros(axes(Y.c)),
+        ᶜ3d_snow = zeros(axes(Y.c)),
```
szy21, line 32: Is it causing problems if these are initialized using `similar`? For some reason I don't fully understand, I think it is preferred to not have prescribed initial values for the cache, if not needed.

config/model_configs/aquaplanet_rhoe_equil_clearsky_tvinsol_0M_slabocean.yml
```diff
@@ -0,0 +1,21 @@
+FLOAT_TYPE: "Float32"
+dt_save_state_to_disk: "10days"
+t_end: "10days"
+moist: "equil"
+precip_model: "0M"
+surface_setup: "DefaultMoninObukhov"
+vert_diff: "FriersonDiffusion"
+implicit_diffusion: true
+max_newton_iters_ode: 2
```
trontrytel, line 9: Should we be using this one? `approximate_linear_solve_iters: 2`

src/parameterized_tendencies/microphysics/precipitation.jl
trontrytel: Does it mean that if we are running with `! PrognosticSurfaceTemperature` we don't expect energy conservation?

trontrytel (approved):
Thank you!

config/longrun_configs/longrun_bw_rhoe_equil_highres_topography_earth.yml
```diff
@@ -12,6 +12,7 @@ precip_model: "0M"
 topography: "Earth"
 job_id: "longrun_bw_rhoe_equil_highres_topography_earth"
 moist: "equil"
+netcdf_interpolate_z_over_msl: true
```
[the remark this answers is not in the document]
LenkaNovak: Kind of. I was trying to figure out why [this](https://buildkite.com/clima/climaatmos-longruns/builds/1407#018d55cb-4e92-45ea-9f85-004e1781c2a2) crashes. It's not related to this PR so, unless there is a quick fix, I can open an issue on it and delete this line for now.

examples/hybrid/driver.jl
LenkaNovak: Ahh, good point! The plotting kept failing for all sorts of reasons, which is why I moved it. I could add `skip=true` to `@test` for now? Hopefully the conservation checking will be revamped soon anyway with a more sophisticated solution.

src/cache/cache.jl
LenkaNovak: But wouldn't this add unnecessary variables to dry runs? Or do we not want to run ClimaAtmos dry anymore? 🤔

src/prognostic_equations/surface_temp.jl
LenkaNovak: It was less stable. With `0M` the energy contribution seems to abrupt in some edge cases. I'd imagine `1M` won't have that problem, which is why I left this here. That said, I can try to run the current long runs for this being uncommented, and if they pass I can change it back.

src/cache/cache.jl
szy21: Ah, right. I meant having this whenever precip is not `nothing`. (i.e. not dependent on `PrognosticSurfaceTemperature`)

examples/hybrid/driver.jl
szy21: hmm, maybe let's keep it as it is then. I'm afraid that people will merge PRs without paying attention to whether this check fails if we set skip=true.

src/prognostic_equations/surface_temp.jl
szy21: Maybe just increase depth_ocean to see if that helps?

config/model_configs/aquaplanet_rhoe_equil_clearsky_tvinsol_0M_slabocean.yml
LenkaNovak: If you say so... some of the other long runs still use `max_newton_iters_ode: 2`. @szy21 , please can you advise?

szy21: Yes, we should use `approximate_linear_solve_iters: 2`. If there are still long runs using `max_newton_iters_ode: 2` we should fix them:)

LenkaNovak: All of these configs, or just long runs? 
Screen Shot 2024-01-29 at 4 51 00 PM

LenkaNovak: (except the default?)

src/parameterized_tendencies/microphysics/precipitation.jl
LenkaNovak: You could do it. The fluxes are just not being tracked / collected here. We can add it to the next PR on conservation.

config/model_configs/aquaplanet_rhoe_equil_clearsky_tvinsol_0M_slabocean.yml
szy21: Just longruns and `flame_perf_target_frierson`, but we can remove the ones that have 1 as that is the default. Don't worry about it for this PR. I can change it.

src/parameterized_tendencies/microphysics/precipitation.jl
LenkaNovak: Let me try similar with the final version...

src/prognostic_equations/surface_temp.jl
LenkaNovak: (although that's cheating) :-P

src/cache/cache.jl
LenkaNovak: @szy21, it seems that by generalizing this as suggested above, we include some jobs that are also checked by flame graphs. These checks fail because this check adds substantially more allocations (more than expected- see [build](https://buildkite.com/clima/climaatmos-ci/builds/16296)). Would you prefer that we 1) only make this applicable to the slab ocean example for now and try to deal with the allocation issue as part of the next PR; or 2) increase the allowable allocation limit?

src/prognostic_equations/surface_temp.jl
LenkaNovak: It works! :)

src/cache/cache.jl
szy21: I think this [line](https://github.com/CliMA/ClimaAtmos.jl/blob/a079b1ddfeea16c7e3f349adc70f4673528d0955/src/parameterized_tendencies/microphysics/precipitation.jl#L114) is allocating. Could you try caching `col_precip_energy` as what is done for e.g. `col_integrated_rain` and see if it helps with allocation? It seems you can use `col_precip_energy` directly instead of `col_integrated_precip_energy_tendency` in the surface temperature equation, and that shouldn't allocate.

LenkaNovak: This normally shouldn't allocate. I tried it anyway, and it doesn't make that much difference. :-(

LenkaNovak: see [this build](https://buildkite.com/clima/climaatmos-ci/builds/16299#018d58e1-ff70-4885-923c-8dca1a99f563)

LenkaNovak: It's probably to do with [this](https://github.com/CliMA/ClimaAtmos.jl/blob/b4ef211c25426e71233912cc493af8293ec81b0c/src/parameterized_tendencies/microphysics/precipitation.jl#L162C1-L169C14).

szy21: Oh! It might be you need to have a temporary field for
```
@. ᶜS_ρq_tot[colidx] * e_tot_0M_precipitation_sources_helper(
                    thermo_params,
                    ᶜts[colidx],
                    ᶜΦ[colidx],
                )
```
Otherwise it may allocate a new field for the integral? I can give it a try.

szy21: This works locally: https://github.com/CliMA/ClimaAtmos.jl/blob/f5559a40a3cc1189fc03160816073adb77ddbdb7/src/parameterized_tendencies/microphysics/precipitation.jl#L155-L167

LenkaNovak: yep, [here's](https://buildkite.com/clima/climaatmos-ci/builds/16300) the test.

szy21: Yeah, just use the temporary field as above should fix it.

szy21: (You can probably use ᶜtemp_scalar instead of ᶜtemp_scalar_3. I didn't check the entire file when trying it.)

LenkaNovak: Ahh thank you. Sorry, GH didn't refresh, so I hadn't seen your suggestion above. Just running it now.

LenkaNovak: 🎉

PR #2571, opened 2024-01-25
Add auto-debug feature

charleskawczynski:
This PR adds an "auto-debug" feature. Enabling this feature via the `run_mode` toml argument will run the simulation with some modifications:

 - A call to `Random.seed!` is called before each and every step, to ensure reproducibility.
 - When a simulation crashes, it is automatically "replays" the crash, starting from N (a given parameter) `dt_save_state_to_disk`'s ago but with a higher temporal diagnostic frequency.
 - This replay occurs inside a try-catch block and, once caught, we search for the most problematic columns (those with Inf / NaNs) and plot them for all state variables.
 - We then gracefully exit with the integrator.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -406,6 +406,12 @@ steps:
           --config_file $CONFIG_PATH/test_io.yml
         artifact_paths: "test_io/*"
 
+      - label: ":computer: Test auto-debug"
+        command: >
+          julia --color=yes --project=examples examples/hybrid/driver.jl
+          --config_file $CONFIG_PATH/test_auto_debug.yml
+        artifact_paths: "test_auto_debug/*"
+
       - label: ":computer: MPI io test"
         command: >
           srun julia --color=yes --project=examples examples/hybrid/driver.jl
--- a/config/default_configs/default_config.yml
+++ b/config/default_configs/default_config.yml
@@ -262,3 +262,9 @@ netcdf_output_at_levels:
 warn_allocations_diagnostics:
   help: "When true, a dry-run for all the diagnostics is performed to check whether the functions allocate additional memory (which reduces performances)"
   value: false
+run_mode:
+  help: "[Productionrun,AutoDebugRun]"
+  value: "ProductionRun"
+simulate_crash:
+  help: "Bool used in AutoDebugRun to simulate a crashed simulation (for testing)"
+  value: false
--- a/config/model_configs/test_auto_debug.yml
+++ b/config/model_configs/test_auto_debug.yml
@@ -1,0 +1,6 @@
+dt: "1secs"
+t_end: "100secs"
+auto_debug: true
+simulate_crash: true
+job_id: "test_auto_debug"
+dt_save_state_to_disk: "10secs"
--- a/examples/hybrid/driver.jl
+++ b/examples/hybrid/driver.jl
@@ -13,6 +13,11 @@ if !(@isdefined config)
 end
 simulation = CA.get_simulation(config)
 (; integrator) = simulation
+if CA.get_run_mode(integrator) isa CA.AutoDebugRun
+    include(joinpath(pkgdir(CA), "post_processing", "auto_debug_plots.jl"))
+    CA.get_run_mode(integrator).auto_plot =
+        integrator -> auto_debug_plots(integrator)
+end
 sol_res = CA.solve_atmos!(simulation)
 
 (; atmos, params) = integrator.p
--- a/post_processing/auto_debug_plots.jl
+++ b/post_processing/auto_debug_plots.jl
@@ -1,0 +1,154 @@
+import ClimaAtmos as CA
+import ClimaComms
+import ClimaCore.InputOutput
+import ClimaCore.Fields
+import ClimaCore.Spaces
+import CairoMakie as MK
+import REPL.TerminalMenus as TM
+function auto_debug_plots(integrator)
+    debugger = CA.get_run_mode(integrator)
+    comms_ctx = ClimaComms.context(integrator.u.c)
+    (; dt, t) = integrator
+    auto_debug_plots(debugger, comms_ctx, dt, t)
+end
+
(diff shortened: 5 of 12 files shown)


Sbozzolo:
It seems that several EDMF runs take more than twice as long to complete on buildkite. Can you please look into this?

Sbozzolo:
Can the run mode be in the AtmosSimulation instead of the Integrator?

szy21:
You can estimate that by comparing this two builds: https://buildkite.com/clima/climaatmos-ci/builds/16102#018d37f3-c07a-4db2-95b1-0f2dc601aace and https://buildkite.com/clima/climaatmos-ci/builds/16118#018d392a-0d31-4d57-857e-281e2e461452. init_diagnostics is four times slower after increasing the interpolation resolution by a factor of 4 (which makes sense?)

szy21:
Changing the frequency of output is not that efficient. I think we need a way to specify interpolation resolution. Anyway, I think you can ignore the increase in time to solution of EDMF jobs for this PR.

charleskawczynski:
> Changing the frequency of output is not that efficient. I think we need a way to specify interpolation resolution. Anyway, I think you can ignore the increase in time to solution of EDMF jobs for this PR.

This only happens when using the autodebug feature, everything should be unchanged when it's off

szy21:
> > Changing the frequency of output is not that efficient. I think we need a way to specify interpolation resolution. Anyway, I think you can ignore the increase in time to solution of EDMF jobs for this PR.
> 
> This only happens when using the autodebug feature, everything should be unchanged when it's off

Right, I meant the EDMF jobs were taking more than twice before this PR, so has nothing to do with this PR.

charleskawczynski:
> Right, I meant the EDMF jobs were taking more than twice before this PR, so has nothing to do with this PR.

Ah, ok, I see. I figured

charleskawczynski:
> Can the run mode be in the AtmosSimulation instead of the Integrator?

What's the benefit of doing this? It would require changing lots of function signatures, which I'd rather not hash out. We can change it later if there's a need, but I'd rather move forward with this since it's working.

Sbozzolo:
> > Can the run mode be in the AtmosSimulation instead of the Integrator?
> 
> What's the benefit of doing this? It would require changing lots of function signatures, which I'd rather not hash out. We can change it later if there's a need, but I'd rather move forward with this since it's working.

Here's some of my thoughts:
- This PR adds a new way to run the integrator. The integrator can and should be able to exist independently of the `run_mode` instead of of being entangled. 
- Experience has shown that the cache can affect compilation time in a very significant way (e.g., adding/not-adding `do_dss`), especially considering the black-box that is SciML. I really would like to keep the cache as simple as possible. `RunMode` could live in the `AtmosSimulation`. `solve_atmos!` already takes an `AtmosSimulation`, so there's already a natural place where to put the run_mode logic and you can dispatch at the level of `timed_solve!`
- `export_state` should not be in `call_every_dt`. The integrator is mutable, `run_mode` can add a new callback to output the state more often when replaying without affecting the other callbacks. This feels more natural if you take the perspective that `run_mode` is controlling and instrumenting the integrator as opposed to being part of it. 

What are the drawbacks of the run mode outside the integrator?

charleskawczynski:
> Here's some of my thoughts:
> 
> * This PR adds a new way to run the integrator. The integrator can and should be able to exist independently of the `run_mode` instead of of being entangled.

I agree that this would be ideal.

> * Experience has shown that the cache can affect compilation time in a very significant way (e.g., adding/not-adding `do_dss`), especially considering the black-box that is SciML. I really would like to keep the cache as simple as possible. `RunMode` could live in the `AtmosSimulation`. `solve_atmos!` already takes an `AtmosSimulation`, so there's already a natural place where to put the run_mode logic and you can dispatch at the level of `timed_solve!`

We should make profile-guided decisions when it comes to performance. The performance issues you mentioned were mostly fixed by making spaces mutable. I agree that, ideally, we would keep the cache as simple as possible, and I'm not thrilled about adding it, but callbacks (which we must inject changes into for this feature) do not have access to `AtmosSimulation`, and I'd prefer that we not further complicate the timestepper.

> * `export_state` should not be in `call_every_dt`. The integrator is mutable, `run_mode` can add a new callback to output the state more often when replaying without affecting the other callbacks. This feels more natural if you take the perspective that `run_mode` is controlling and instrumenting the integrator as opposed to being part of it.

Again, the callbacks do not have access to `AtmosSimulation`, so I don't see how this can be done..

> What are the drawbacks of the run mode outside the integrator?

We would need to change the timestepper interface. I'm open to doing this in the long run, but I'd like to first see what we need before making those changes, and it's easier to first iterate in ClimaAtmos. Once we have something converged, we can think about how to make this feature available by the timestepper itself.

Sbozzolo:
> Again, the callbacks do not have access to AtmosSimulation, so I don't see how this can be done..

Why do they need access to `AtmosSimulation`? 

If you only want to access and modify the `AutoDebugRun` object in callback, you can pack it into a closure. This feature is only for debugging and has no performance requirements[*], we can definetely afford the downsides of closures for a much cleaner design.

[*] The only requirement is that this feature does not significantly impact performance/latency for production runs.

charleskawczynski:
> If you only want to access and modify the `AutoDebugRun` object in callback, you can pack it into a closure.

We could do that, but one issue with that design is that we'll need to search through all the callbacks at runtime to get `AutoDebugRun`. Also, that search cannot be statically elided, like it can be in the case of my proposed design.

I'm fine with making this change, but I can't guarantee that it won't have a performance impact when its turned on.

In case there are other issues with that approach, I'll leave this PR and open a separate one where it's handled via AtmosSimulation and callbacks.

charleskawczynski:
This has gotten pretty stale, but I do still like the idea of this. I'll convert this into a draft PR for reference, since we'll likely need to refresh this off main.

charleskawczynski:
This has gotten quite stale. Closing.

PR #2573, opened 2024-01-25
remove get_state

szy21:
## Purpose 

Removes `get_state` as this function is not being used.

## Content

----


Sbozzolo (approved):
Thank you!

PR #2574, opened 2024-01-25
Update EDMF plots

nefrathenrici:
Refactors `make_plots_generic` and customizes the EDMF line plots for readability.

Content

`make_plots_generic` 
- creates a grid of plots (default only 1 column) and can take in a custom plotting function.
- takes a `summary_files` kwarg to prepend initial pages to the final PDF.

EDMF plots
- plot gridmean/updraft on the same axis for vertical profiles
- make plots on a 4x2 grid
- Remove y log scale
- Add more variables

### DIFF
--- a/post_processing/ci_plots.jl
+++ b/post_processing/ci_plots.jl
@@ -26,7 +26,7 @@ function common_dirname(files::Vector{T}) where {T <: AbstractString}
     return joinpath(split_files[1][1:last_common_dir]...)
 end
 
-function make_plots(sim, simulation_path)
+function make_plots(sim, _)
     @warn "No plot found for $sim"
 end
 
@@ -59,41 +59,91 @@ YLOGSCALE = Dict(
     ),
 )
 
+long_name(var) = var.attributes["long_name"]
+short_name(var) = var.attributes["short_name"]
+
+"""
+    parse_var_attributes(var)
+
+Takes in an OutputVar and parses some of its attributes into a short, informative string.
+Used to generate unique titles when the same var is being plotted for several times/locations.
+This could be extended to parse more attributes.
+
+For example, the sample attributes:
+attributes = Dict(
+    "units" => "%",
+    "short_name" => "cl",
+    "slice_y" => "0.0",
+    "long_name" => "Cloud fraction, Instantaneous x = 0.0 m y = 0.0 m",
+    "slice_y_units" => "m",
+    "slice_x_units" => "m",
+    "comments" => "",
+    "slice_x" => "0.0",
+)
+will be parsed into "cl, x = 0.0, y = 0.0"
+"""
+function parse_var_attributes(var)
+    MISSING_STR = "MISSING_ATTRIBUTE"
+    attr = var.attributes
+    name = replace(short_name(var), "up" => "")
+
+    attributes = ["slice_lat", "slice_lon", "slice_x", "slice_y", "slice_time"]
+    info = [
+        replace(key, "slice_" => "") * " = " * get(attr, key, MISSING_STR)
+        for key in attributes
+    ]
+    # Filter out missing entries
+    info = filter(x -> !occursin(MISSING_STR, x), [name, info...])
+
+    return join(info, ", ")
+end
+
 function make_plots_generic(
     output_path,
     vars,
     args...;
+    plot_fn = nothing,
     output_name = "summary",
+    summary_files = String[],
+    MAX_NUM_COLS = 1,
+    MAX_NUM_ROWS = 4,
     kwargs...,
 )
-    MAX_PLOTS_PER_PAGE = 4
+    # Default plotting function needs access to kwargs
+    if isnothing(plot_fn)
+        plot_fn =
+            (grid_loc, var) -> viz.plot!(grid_loc, var, args...; kwargs...)
+    end
+
+    MAX_PLOTS_PER_PAGE = MAX_NUM_ROWS * MAX_NUM_COLS
     vars_left_to_plot = length(vars)
 
-    # Define fig, and p_loc, used below. (Needed for scope)
-    fig = CairoMakie.Figure(
-        resolution = (900, 300 * min(vars_left_to_plot, MAX_PLOTS_PER_PAGE)),
-    )
-    p_loc = [1, 1]
-    page = 1
-    summary_files = String[]
-
-    for (var_num, var) in enumerate(vars)
-        # Create a new page if this is the first plot
-        if mod(var_num, MAX_PLOTS_PER_PAGE) == 1
-            fig = CairoMakie.Figure(
-                resolution = (
-                    900,
-                    300 * min(vars_left_to_plot, MAX_PLOTS_PER_PAGE),
-                ),
-            )
-            p_loc = [1, 1]
+    # Define fig, grid, and grid_pos, used below. (Needed for scope)
+    makefig() = CairoMakie.Figure(resolution = (900, 300 * MAX_NUM_ROWS))
+    gridlayout() =
+        map(1:MAX_PLOTS_PER_PAGE) do i
+            row = mod(div(i - 1, MAX_NUM_COLS), MAX_NUM_ROWS) + 1
+            col = mod(i - 1, MAX_NUM_COLS) + 1
+            return fig[row, col] = CairoMakie.GridLayout()
         end
-        viz.plot!(fig, var, args...; p_loc, kwargs...)
 
-        p_loc[1] += 1
+    fig = makefig()
+    grid = gridlayout()
+    page = 1
+    grid_pos = 1
+
+    for var in vars
+        if grid_pos > MAX_PLOTS_PER_PAGE
+            fig = makefig()
+            grid = gridlayout()
+            grid_pos = 1
+        end
+
+        plot_fn(grid[grid_pos], var)
+        grid_pos += 1
 
         # Flush current page
-        if p_loc[1] > min(MAX_PLOTS_PER_PAGE, vars_left_to_plot)
+        if grid_pos > min(MAX_PLOTS_PER_PAGE, vars_left_to_plot)
             file_path = joinpath(output_path, "$(output_name)_$page.pdf")
             CairoMakie.save(file_path, fig)
             push!(summary_files, file_path)
@@ -110,6 +160,7 @@ function make_plots_generic(
 
     # Cleanup
     Filesystem.rm.(summary_files, force = true)
+    return output_file
 end
 
 """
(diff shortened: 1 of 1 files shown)


post_processing/ci_plots.jl
```diff
@@ -542,19 +543,112 @@ EDMFBoxPlots = Union{
...
+function parse_long_name(var)
+    name = long_name(var)
+    for str in ["east ", "west ", "north ", "south "]
+        name = replace(name, "degrees_" * str => "")
+    end
+    var_name, rest = split(name, ",")
+    reduction, rest = split(rest, "(s) ")
+    coord, time = split(rest, " time = ")
+    return "$(short_name(var)), $coord, t = $time"
+end
```
Sbozzolo, line 590: What is your goal here?

post_processing/ci_plots.jl
```diff
@@ -542,19 +543,112 @@ EDMFBoxPlots = Union{
...
+        # Flush current page
+        if var_num % MAX_PLOTS_PER_PAGE == 0 ||
+           var_num == length(keys(grouped_vars))
+            file_path = joinpath(output_path, "edmf_vars_z_temp_$page.pdf")
+            CairoMakie.save(file_path, fig)
+            push!(summary_files, file_path)
+            page += 1
+        end
+    end
+    return summary_files
```
Sbozzolo, line 651: A lot of the logic here is for page handling. I think it would be nice if they were abstracted away and added to `make_plots_generic`. Then `make_plot_generic` could take a plotting function.

post_processing/ci_plots.jl
nefrathenrici: Printing the full long name is much too long for a grid of plots, so this keeps the necessary info and makes it shorter

Sbozzolo: What do you want to keep? This is function is not very robust. We try to add some extra information in the attributes of OutputVar.

nefrathenrici: Yes, this function is pretty ugly. It would be nice to have a field for coordinate and timestamp when it is applicable.

post_processing/ci_plots.jl
```diff
@@ -542,19 +543,112 @@ EDMFBoxPlots = Union{
...
+        "haup",
+        "waup",
+        "tke",
+        "arup",
+        "hus",
+        "husup",
+        "hur",
+        "hurup",
+        "cl",
+        "clw",
```
szy21, line 563:
```suggestion
        "clw",
        "clwup",
```

post_processing/ci_plots.jl
```diff
@@ -542,19 +543,112 @@ EDMFBoxPlots = Union{
...
+        "cli",
```
szy21, line 564:
```suggestion
        "cli",
        "cliup"
```

post_processing/ci_plots.jl
```diff
@@ -542,19 +543,112 @@ EDMFBoxPlots = Union{
...
+    var_name, rest = split(name, ",")
+    reduction, rest = split(rest, "(s) ")
+    coord, time = split(rest, " time = ")
+    return "$(short_name(var)), $coord, t = $time"
+end
+
+function make_plots_edmf_vars_z(output_path, vars_z)
+    # Group gridmean and updraft variables
+    get_longname_suffix(v) =
+        startswith(long_name(v), "Updraft") ? long_name(v)[9:end] : long_name(v)
```
szy21, line 595: This seems a bit risky as the name can change. If we have to do something like this, it might be better to use short names to group variables, as they are more unlikely to change.

szy21 (commented):
The plots look great, thanks! Just notice that it is still one contour plot per row - is this intended?

post_processing/ci_plots.jl
nefrathenrici: Gabriele and I discussed adding timestamp and coordinate as diagnostic variable attributes. I will just use the short name for now!

nefrathenrici:
> The plots look great, thanks! Just notice that it is still one contour plot per row - is this intended?

@szy21 There seem to be issues with gridding the contour plots. I believe it has to do with the color bar, but need to look into it properly.
Screenshot 2024-01-26 at 10 27 40 AM

szy21:
Ah, ok, then what you have now is good. Thanks!

post_processing/ci_plots.jl
```diff
@@ -539,22 +557,95 @@ EDMFBoxPlots = Union{
...
+Groups pairs of updraft and gridmean EDMF OutputVars into a Dict.
+This is fairly brittle - it just checks that the long name starts with "Updraft"
+and matches on the rest of the long name.
+"""
+function group_edmf_vars(vars)
+    get_longname_suffix(v) =
+        startswith(long_name(v), "Updraft") ? long_name(v)[9:end] : long_name(v)
+    grouped_vars = OrderedDict{String, Vector{ClimaAnalysis.OutputVar}}()
+    for var in sort(vars, by = get_longname_suffix)
+        suffix = get_longname_suffix(var)
```
Sbozzolo, line 594:
```suggestion
suffix = replace(long_name(v), "Updraft " => ""
```

post_processing/ci_plots.jl
```diff
@@ -539,22 +557,95 @@ EDMFBoxPlots = Union{
...
+    length(var_group) > 1 && Makie.axislegend(ax)
+end
+
+"""
+    group_edmf_vars(vars)
+
+Groups pairs of updraft and gridmean EDMF OutputVars into a Dict.
+This is fairly brittle - it just checks that the long name starts with "Updraft"
+and matches on the rest of the long name.
+"""
```
Sbozzolo, line 588: Can you make the pairings from the short names?

post_processing/ci_plots.jl
```diff
@@ -539,22 +557,95 @@ EDMFBoxPlots = Union{
     Val{:prognostic_edmfx_trmm_column},
 }
 
+"""
+    plot_edmf_vert_profile(fig, p_loc, var_group)
+
+Helper function for `make_plots_generic`. Takes a list of variables and plots
+them on the same axis.
+"""
+function plot_edmf_vert_profile(fig, p_loc, var_group)
```
Sbozzolo, line 566: This function is mutating, so it should have a `!`

post_processing/ci_plots.jl
```diff
@@ -59,41 +60,72 @@ YLOGSCALE = Dict(
     ),
 )
 
+long_name(var) = var.attributes["long_name"]
+short_name(var) = var.attributes["short_name"]
+
+function var_info(var)
```
Sbozzolo, line 66: This function makes assumptions:
- when `slice_lat` is available, also `slice_lon` is available
- if `slice_lat` is not available, `slice_x` and `slice_y` are both available.

Sbozzolo: Can you also add a docstring and give an example of what this function does (in the docstring)?

post_processing/ci_plots.jl
```diff
@@ -590,22 +623,112 @@ EDMFBoxPlots = Union{
...
+        CairoMakie.lines!(ax, var.data, z, label = short_name(var))
+    end
+    length(var_group) > 1 && Makie.axislegend(ax)
+end
+
+
+"""
+    plot_custom_title!(grid_loc, var)
+
+Helper function for `make_plots_generic`. Plots a variable with a custom title.
```
Sbozzolo, line 652: This is not just "a custom title", it is a specific title. Can you add it to the docstring?

post_processing/ci_plots.jl
```diff
@@ -590,22 +623,112 @@ EDMFBoxPlots = Union{
...
+    for var in sort(vars, by = var_info)
+        suffix = var_info(var)
+        if haskey(grouped_vars, suffix)
+            push!(grouped_vars[suffix], var)
+        else
+            grouped_vars[suffix] = [var]
+        end
+    end
+    return grouped_vars
+end
```
Sbozzolo, line 677: I find this function hard to understand.

Why does the dictionary have to be ordered?
Why do the variables have to be sorted, and why sorted by `var_info`?

What I understand is that you are matching some variables with their `up` version. The trouble comes from the fact that there's also slicing information. What if you do the matching before slicing? See below

post_processing/ci_plots.jl
```diff
@@ -590,22 +623,112 @@ EDMFBoxPlots = Union{
...
+        "clwup",
+        "cli",
+        "cliup",
+    ]
     reduction = "inst"
     period = "30m"
+
     vars = [
         get(simdir; short_name, reduction, period) for short_name in short_names
     ]
```
Sbozzolo, line 709: Pseudo code
```suggestion
    vars = [
        get(simdir; short_name, reduction, period) for short_name in short_names
    ]
    grouped_vars = Any[]
    for index, short_name in enumerate(short_names)
        # skip updraft variables
        short_name[end-2:end] == "up" && continue

        up_index = find(short_name * "up" in short_names)
        if isnothing(up_index)
           add_tup = tuple(vars[index])
        else
           add_tup = tuple(vars[index], vars[up_index])
         end  
            push!(grouped_vars, add_tup)
    end
```
`grouped_vars` is a vector of 1/2-tuples of `OutputVars` that you can slice.

post_processing/ci_plots.jl
```diff
@@ -4,6 +4,7 @@ import ClimaAnalysis
 import ClimaAnalysis: Visualize as viz
 import ClimaAnalysis: SimDir, slice_time, slice
 import ClimaAnalysis.Utils: kwargs as ca_kwargs
+import OrderedCollections: OrderedDict
```
Sbozzolo, line 7: You don't need this anymore

post_processing/ci_plots.jl
```diff
@@ -627,19 +786,41 @@ EDMFSpherePlots =
...
+        (
+            slice(
+                get(simdir; short_name, reduction, period),
+                lon = 0.0,
+                lat = lat,
+            ) for short_name in var_names
+        ) for lat in latitudes, var_names in short_name_tuples
+    ]
+    var_groups_z = [
+        (collect(slice(v, time = LAST_SNAP) for v in group)...,) for
```
Sbozzolo, line 806: DOes this work?
```suggestion
        ([slice(v, time = LAST_SNAP) for v in group]) for
```

post_processing/ci_plots.jl
```diff
@@ -606,18 +657,126 @@ EDMFBoxPlots = Union{
...
+        up_name = base_name * "up"
+
+        if base_name in short_names_to_be_processed &&
+           up_name in short_names_to_be_processed
+            tuple_to_be_added = (base_name, up_name)
+        else
+            tuple_to_be_added = (name,)
+        end
+
+        [pop!(short_names_to_be_processed, n) for n in tuple_to_be_added]
```
Sbozzolo, line 719:
```suggestion
        foreach(n -> pop!(short_names_to_be_processed, n), tuples_to_be_added)
```
This avoids creating a new vector only to enjoy the side effect

Sbozzolo: And maybe we want to use `delete!` instead of `pop!`

post_processing/ci_plots.jl
```diff
@@ -606,18 +657,126 @@ EDMFBoxPlots = Union{
...
+
+        base_name = replace(name, "up" => "")
+        up_name = base_name * "up"
+
+        if base_name in short_names_to_be_processed &&
+           up_name in short_names_to_be_processed
+            tuple_to_be_added = (base_name, up_name)
+        else
+            tuple_to_be_added = (name,)
+        end
```
Sbozzolo, line 717: Can you add back some of the comments I left in the review? I think they are helpful to understand this code.

Sbozzolo (commented):
This looks great. Only cosmetic comments are left.

post_processing/ci_plots.jl
nefrathenrici: Yes! That's a little nicer.

Sbozzolo (approved):
Thank you! This looks really nice and clean. I am very happy on how it turned out! 

Just squash your commits and feel free to merge.

Issue #2344, opened 2023-11-08
Estimate maximum allowable timestep for moist baro wave


charleskawczynski:
We need to estimate the maximum allowable timestep for:
 - moist baro wave
 - horizontal limiters (apply to advection + hyperdiffusion)
 - vertical FCT
 - SSP time stepper
 - target spatial resolution:
   - `nh_poly = 3` (`Nq = 4`)
   - `h_elem = 30`
   - `z_elem = 63`
 - target time: 100days

Is the target resolution (which I grabbed from the dyamond job) okay? Also, how long should this be stable for to declare success? I've put 100 days for now, which will take ~1.5 hours to run assuming we can simulate at 10 SYPD, so this will need to be a long run. Based on the latest dyamond results (https://github.com/CliMA/ClimaAtmos.jl/issues/2314#issuecomment-1801136086), we're currently at 0.468 SYPD. So, at the moment, we should be able to finish 100 days in ~13 hours on the A100. Hopefully this will dramatically improve once we figure out what's wrong.

This issue is connected to the Update Limiters milestone. We can't reference ClimaCore milestones in ClimaAtmos, but I'd like to run these simulation in ClimaAtmos CI.

szy21:
The high domain top in dyamond and other aquaplanet simulation is not realistic for baroclinic wave initial conditions, so that may cause some problems. Maybe you can keep z_max at 30000 but use 63 layers. If you want a faster turnaround I think 30 days would be good enough to check whether the simulation is stable. Having said that, it should be faster than dyamond, as there is no radiation.

szy21:
Also, just note that the maximum timestep of moist held suarez and aquaplanet is in general 2 times smaller than that of moist baroclinic wave. See the current longrun pipeline (using ARS, central difference, and no limiter) for example.

charleskawczynski:
@cmbengue, @dennisYatunin, @tapios, we've estimated the maximum timestep in #2542, can we close this?

tapios:
Where are the results? We need the matrix of maximum allowable timestep and time to solution for ARS vs SSP, FCT vs CD, and possibly maximum number of Newton iteration. It needs to include a stable configuration with FCT.

charleskawczynski:
These are the current results:
```
dts tested ∈ [50, ..., 300]
                     | stable  |   unstable
ARS_nolim_CD         | 150     |   160
SSP_nolim_FCT        |  none   |   all
[...]

PR #2577, opened 2024-01-26
Add ARS no limiter FCT to longrun pipeline

charleskawczynski:
xref: https://github.com/CliMA/ClimaAtmos.jl/issues/2344#issuecomment-1911089538.


szy21 (commented):
Let's add the toml file following this line: https://github.com/CliMA/ClimaAtmos.jl/blob/b370eb872966793a8de7af0ea1da57d1b3cfd7b1/config/longrun_configs/longrun_hs_rhoe_equil_55km_nz63_0M.yml#L15

szy21:
Also, it seems you are not using GPU? If so, maybe just put it in ./buildkite/longruns/pipeline.yml?

szy21 (approved):
Feel free to merge after making the changes.

szy21:
Oh, one more thing. If you want plots, add the job_id to ci_plots.jl.

charleskawczynski:
> Also, it seems you are not using GPU? If so, maybe just put it in ./buildkite/longruns/pipeline.yml?

Right, do you think it should be a CPU or GPU job?

szy21:
> > Let's add the toml file following this line:
> > https://github.com/CliMA/ClimaAtmos.jl/blob/b370eb872966793a8de7af0ea1da57d1b3cfd7b1/config/longrun_configs/longrun_hs_rhoe_equil_55km_nz63_0M.yml#L15
> 
> What is this doing again?

This is adding the correct parameters for the rayleigh sponge.

szy21:
> > Also, it seems you are not using GPU? If so, maybe just put it in ./buildkite/longruns/pipeline.yml?
> 
> Right, do you think it should be a CPU or GPU job?

I was thinking using central cluster for this job (which is what longrun/pipeline.yml uses), as it may still change. We have been using CPUs for that pipeline. But if we can use both central CPU and GPU in the same pipeline, I think GPU is good too.
Basically I just want to keep longrun_gpu/pipeline.yml short as it is using the eight clima GPUs, and there are already seven jobs (I may move some out later).

charleskawczynski:
> I was thinking using central cluster for this job (which is what longrun/pipeline.yml uses), as it may still change. We have been using CPUs for that pipeline. But if we can use both central CPU and GPU in the same pipeline, I think GPU is good too. Basically I just want to keep longrun_gpu/pipeline.yml short as it is using the eight clima GPUs, and there are already seven jobs (I may move some out later).

I'll leave it as CPU for now, but we can change it later if it seems needed.

szy21:
> > I was thinking using central cluster for this job (which is what longrun/pipeline.yml uses), as it may still change. We have been using CPUs for that pipeline. But if we can use both central CPU and GPU in the same pipeline, I think GPU is good too. Basically I just want to keep longrun_gpu/pipeline.yml short as it is using the eight clima GPUs, and there are already seven jobs (I may move some out later).
> 
> I'll leave it as CPU for now, but we can change it later if it seems needed.

Sounds good. It seems you are using only one CPU. It will be very slow. Is that intended?

charleskawczynski:
> Sounds good. It seems you are using only one CPU. It will be very slow. Is that intended?

I didn't know that, I thought all of the other jobs in `longrun/pipeline.yml` were also using CPU. I'm happy to switch it to a gpu job.

szy21:
> > Sounds good. It seems you are using only one CPU. It will be very slow. Is that intended?
> 
> I didn't know that, I thought all of the other jobs in `longrun/pipeline.yml` were also using CPU. I'm happy to switch it to a gpu job.

Yes, but they are using multiple cpus. I think you need to specify slurm_ntasks and add srun before julia?

szy21:
Should we close this? We are using SSPKnoth for most simulations now.

charleskawczynski:
> Should we close this? We are using SSPKnoth for most simulations now.

Fine by me.

PR #2579, opened 2024-01-26
add TOA and surface radiation diagnostics

### LINKED ISSUE #2578: Add surface and TOA radiative fluxes to diagnostics

szy21:
## Purpose 

Closes #2578 
Also cleans up diagnostics for slab ocean experiments

## Content

----

### DIFF
--- a/src/diagnostics/radiation_diagnostics.jl
+++ b/src/diagnostics/radiation_diagnostics.jl
@@ -40,7 +40,94 @@ add_diagnostic_variable!(
 )
 
 ###
-# Upwelling shortwave radiation (3d)
+# TOA downwelling shortwave radiation (2d)
+###
+compute_rsdt!(out, state, cache, time) =
+    compute_rsdt!(out, state, cache, time, cache.atmos.radiation_mode)
+compute_rsdt!(_, _, _, _, radiation_mode::T) where {T} =
+    error_diagnostic_variable("rsdt", radiation_mode)
+
+function compute_rsdt!(
+    out,
+    state,
+    cache,
+    time,
+    radiation_mode::T,
+) where {T <: RRTMGPI.AbstractRRTMGPMode}
+    nlevels = Spaces.nlevels(axes(state.c))
+    if isnothing(out)
+        return Fields.level(
+            RRTMGPI.array2field(
+                cache.radiation.radiation_model.face_sw_flux_dn,
+                axes(state.f),
+            ),
+            nlevels + half,
+        )
+    else
+        out .= Fields.level(
+            RRTMGPI.array2field(
+                cache.radiation.radiation_model.face_sw_flux_dn,
+                axes(state.f),
+            ),
+            nlevels + half,
+        )
+    end
+end
+
+add_diagnostic_variable!(
+    short_name = "rsdt",
+    long_name = "TOA Incident Shortwave Radiation",
+    standard_name = "toa_incoming_shortwave_flux",
+    units = "W m^-2",
+    comments = "Downward shortwave radiation at the top of the atmosphere",
+    compute! = compute_rsdt!,
+)
+
+###
+# Surface downwelling shortwave radiation (2d)
+###
+compute_rsds!(out, state, cache, time) =
+    compute_rsds!(out, state, cache, time, cache.atmos.radiation_mode)
+compute_rsds!(_, _, _, _, radiation_mode::T) where {T} =
+    error_diagnostic_variable("rsds", radiation_mode)
+
+function compute_rsds!(
+    out,
+    state,
+    cache,
+    time,
+    radiation_mode::T,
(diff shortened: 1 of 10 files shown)


src/diagnostics/radiation_diagnostics.jl
```diff
@@ -39,6 +39,93 @@ add_diagnostic_variable!(
...
+        )
+    else
+        out .= Fields.level(
+            RRTMGPI.array2field(
+                cache.radiation.radiation_model.face_sw_flux_dn,
+                axes(state.f),
+            ),
+            nlevels + half,
+        )
+    end
```
szy21, line 74: @Sbozzolo @charleskawczynski Any idea why this doesn't seem to work on GPU? https://buildkite.com/clima/climaatmos-ci/builds/16180#018d44bc-b6a9-443b-8969-0c6befdbab2d/163-260

charleskawczynski: Hmm, for some reason it's doing scalar indexing. I can checkout the branch to see what's going on

charleskawczynski, line 74:
```suggestion
    face_sw_flux_dn_field = RRTMGPI.array2field(cache.radiation.radiation_model.face_sw_flux_dn, axes(state.f))
    lev = Fields.level(face_sw_flux_dn_field, nlevels + half)
    if isnothing(out)
        return lev
    else
        out .= lev
    end
```
Can we break down the expression like this?

szy21: Yes. Why would this work though?

szy21: Ah, I don't think it works: https://buildkite.com/clima/climaatmos-ci/builds/16245#018d537a-779d-4403-ac51-0bfea09917d1/163-260

Sbozzolo:
The issue is with saving the value on the accumulator. `copy` doesn't seem to work on this type:
```
view(reshape(view(::CuArray{Float64, 2, CUDA.Mem.DeviceBuffer}, 1:11, :), 11, 4, 4, 1, 216), 11, :, :, :, :)
```
It might be because there's two nested views.

Sbozzolo:
Minumum breaking example:
```julia
using CUDA
A = CuArray(reshape(1:16, 4, 4))
copy(view(reshape(view(A, 1:3, :), 3, 14), 3, :, :))
```

src/solver/type_getters.jl
```diff
@@ -958,8 +958,14 @@ function get_simulation(config::AtmosConfig)
                     )
                 else
                     # Add to the accumulator
+
+                    # We use similar + .= instead of copy because of a bug in CUDA.jl 5.2.0
```
charleskawczynski, line 962: Can you please provide a link to the issue?

Sbozzolo: There's no issue yet, I asked on slack #gpu, if someone confirms that this is a problem, I will open the issue.

src/diagnostics/radiation_diagnostics.jl
charleskawczynski: It might not, but it could help narrow down the issue.

Sbozzolo: The problem is described below. It is the `copy` function with called with a particular combination of `view/reshape/view`. This only occurs in this case because of the different DataLayout.

Sbozzolo:
Ops, I miscopied! The `14` should be a `4`
```
julia> copy(view(reshape(view(A, 1:3, :), 4, 3), 3, :))
┌ Warning: Performing scalar indexing on task Task (runnable) @0x00007f5e41d44010.
│ Invocation of getindex resulted in scalar indexing of a GPU array.
│ This is typically caused by calling an iterating implementation of a method.
│ Such implementations *do not* execute on the GPU, but very slowly on the CPU,
│ and therefore are only permitted from the REPL for prototyping purposes.
│ If you did intend to index this array, annotate the caller with @allowscalar.
└ @ GPUArraysCore ~/.julia/packages/GPUArraysCore/uOYfN/src/GPUArraysCore.jl:106
3-element CuArray{Int64, 1, CUDA.Mem.DeviceBuffer}:
  3
  9
 14
```

src/solver/type_getters.jl
Sbozzolo: I was pointed out to this https://github.com/JuliaGPU/Adapt.jl/issues/21 with the following comments:
> because of how Julia works, it's tricky to define GPU-accelerated functions for nested array wrappers, so this is triggering fallback functionality from Base
> this would need a change in Base, or a more radical type system change, to make this easier. alternatively, relying more on unified memory would make the fallbacks work better, which is probably what we're going to do instead.

Sbozzolo (approved):

PR #2580, opened 2024-01-26
Make tracer vert adv always explicit

### LINKED ISSUE #2510: Update IMEX splitting of vertical advection
Our current plan is to have the following IMEX splitting of vertical advection:
- Implicit
  - Central vertical advection of active scalars at grid-mean velocity
  - Vertical advection of precipitation at its terminal velocity
  - Central vertical advection of grid-mean values at subgrid-scale velocities (EDMF only)
  - Vertical advection of subgrid-scale values at subgrid-scale velocities (prognostic EDMF only)
- Explicit
  - Vertical advection of horizontal velocity and passive scalars at grid-mean velocity
  - Upwinding corrections to vertical advection of active scalars at grid-mean velocity
  - (Maybe) updwinding corrections to advection of grid-mean values at subgrid-scale velocities (EDMF only)

In order to make our code for the dycore (ignoring EDMF for now) match this plan, we need the following two changes:
- [x]  Split vertical advection at grid-mean velocity into an implicit central component and explicit upwinding corrections
  - Remove the `density_upwinding` flag, and replace it with `Val(:none)` (it does not make sense to upwind density, and what we have now effectively amounts to upwinding the value `1` everywhere)
  - Remove the `energy_upwinding` and `tracer_upwinding` flags from `implicit_tendency!` and `update_implicit_equation_jacobian!`, and replace them with `Val(:none)`
  - Add some new terms to `remaining_tendency!`
    - `vertical_transport!(<energy>, ..., energy_upwinding) - vertical_transport!(<energy>, ..., Val(:none))`
    - `vertical_transport!(<tracer>, ..., tracer_upwinding) - vertical_transport!(<tracer>, ..., Val(:none))`
- [ ] Make the vertical advection of passive tracers at grid-mean velocity explicit
  - Remove all calls to `vertical_transport!` that involve passive tracers (scalars other than `ρ`, `ρe_tot`, and `ρq_tot`) and the grid-mean velocity `ᶠu³` from `implicit_tendency!`, and remove the corresponding Jacobian blocks from `update_implicit_equation_jacobian!`
  - Replace the upwinding correction terms for passive tracers in `remaining_tendency!` with the full vertical advection terms:
    - `vertical_transport!(<tracer>, ..., tracer_upwinding)`

The first change is expected to improve stability for all cases that use upwinding or flux-corrected transport. The second change could lead to a minor decrease in stability, but we do not expect this to happen.

charleskawczynski:
Closes #2510. @dennisYatunin, I think you meant to include `q_liq`/`q_ice` to be non-passive tracers, as I believe they would play a role in the dynamics for non-equilibrium cases, so I left those where they are. I didn't see any other tracers that needed to be moved in `update_implicit_equation_jacobian!`, but maybe @dennisYatunin can double check my work.

### DIFF
--- a/src/prognostic_equations/advection.jl
+++ b/src/prognostic_equations/advection.jl
@@ -67,14 +67,14 @@ NVTX.@annotate function horizontal_tracer_advection_tendency!(Yₜ, Y, p, t)
 end
 
 NVTX.@annotate function explicit_vertical_advection_tendency!(Yₜ, Y, p, t)
-    (; turbconv_model) = p.atmos
+    (; moisture_model, turbconv_model) = p.atmos
     n = n_prognostic_mass_flux_subdomains(turbconv_model)
     advect_tke = use_prognostic_tke(turbconv_model)
     point_type = eltype(Fields.coordinate_field(Y.c))
     (; dt) = p
     ᶜJ = Fields.local_geometry_field(Y.c).J
     (; ᶜf) = p.core
-    (; ᶜu, ᶠu³, ᶜK) = p.precomputed
+    (; ᶜh_tot, ᶜu, ᶠu³, ᶜK) = p.precomputed
     (; edmfx_upwinding) = n > 0 || advect_tke ? p.atmos.numerics : all_nothing
     (; ᶜuʲs, ᶜKʲs) = n > 0 ? p.precomputed : all_nothing
     (; ᶠu³⁰) = advect_tke ? p.precomputed : all_nothing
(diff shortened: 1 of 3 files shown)


dennisYatunin:
Although `q_liq` and `q_ice` play a role in the dynamics, their contribution to the thermodynamic state is negligible because they are several orders of magnitude smaller than `q_tot`. So, we are ignoring their contribution to the pressure gradient, which means that they do not play a role in the creation of fast waves. This means that we can treat them explicitly.

The changes I'm currently seeing move the advection tendencies for all tracers, including `q_tot`. The advection of `q_tot` must remain in the implicit tendency, as it plays a role in the creation of fast waves.

Also, the Jacobian needs to be updated. All blocks that correspond to the advection of passive tracers (this currently includes `q_liq`, `q_ice`, `q_rai`, and `q_sno`) should be removed. For example, the `available_tracer_names` in the `advection_blocks` should be replaced with `ρq_tot_if_available` in the `ImplicitEquationJacobian` constructor.

charleskawczynski:
Alright, @dennisYatunin, I've updated based on your comments. How does this look?

dennisYatunin:
Thanks, @charleskawczynski! I've fixed the remaining issues in this PR, and everything is now working as expected. However, the water budget conservation check is failing for the 1-moment microphysics test in CI. This is unfortunate, but it makes sense that we've introduced some additional error by making the advection of rain and snow explicit. @szy21 and @trontrytel, we should discuss this result next week.

PR #2585, opened 2024-01-26
gpu target runs: pass --cpus-per-task to srun

simonbyrne:
## Purpose 

Apparently the `--cpus-per-task` arguments in `sbatch` are _not_ passed to `srun`. This seems to avoid catastrophic fall offs in scaling when the machine is busy.

run before change: https://buildkite.com/clima/climaatmos-target-gpu-simulations/builds/186
- moist HS 4xA100: sypd: 1.4577044490289248

run after change: https://buildkite.com/clima/climaatmos-target-gpu-simulations/builds/187
- moist HS 4xA100: sypd: 4.996379365801879

## Content

----


simonbyrne:
Looking at the profiles, there is a large (400ms) pause at the start, then the performance picks up to around 6.4 SYPD.

sriharshakandala (approved):

PR #2593, opened 2024-01-28
decrease longrun output frequency

szy21:
## Purpose 

After increasing the default netcdf interpolation resolution, some of  the longrun output file size exceed the maximum buildkite artifact size, see e.g. this [error](https://buildkite.com/clima/climaatmos-gpulongruns/builds/33#018d4981-f0fd-42b8-82a2-9c8bc8847b36/531-557). Not a big problem, but we don't need such frequent output for most jobs anyway, so this PR reduces the longrun output frequency. Also makes some other minor changes of some longrun configs.

## Content

----


Sbozzolo:
We can merge this, but we can also look at more permanent solutions:
- tar into multiple smaller archives, and
- a more meaningful set of defaults for longruns.

szy21:
superseded by #2605

Issue #2597, opened 2024-01-29
Allow statically eliding many JET failures in Krylov/KrylovKit.jl


charleskawczynski:
I just tried to run the JET script on the PR where we upgrade some dependencies. The printing is so horribly bad that I can only fit 3 lines on the terminal (only VI seems to be able to handle opening the log) and it's therefore very difficult to diagnose any issues.

I think the best course of action is:
 - Fix the printing
 - Reduce noise of other issues
 - Reduce extraneous issues
 - Add a JET test (if CI permits) to prevent future regressions. I think we used to have tests, but they were very expensive. We should revisit adding them back in now that compile times are better.
 - Diagnose the actual inference failure to upgrade our dependencies.

The log can be downloaded from [this build](https://buildkite.com/clima/climaatmos-ci/builds/16243#018d4c52-812d-4b9e-a7f1-9a64278a1763), and here is a snippet of a fraction of a line:

```
[...]

PR #2598, opened 2024-01-29
Do not add orchestrate_diagnostic if not needed

Sbozzolo:
And make sure that the progress callback is the last one called.

See https://github.com/CliMA/ClimaAtmos.jl/issues/2597

### DIFF
--- a/src/solver/type_getters.jl
+++ b/src/solver/type_getters.jl
@@ -501,8 +501,10 @@ function get_callbacks(parsed_args, sim_info, atmos, params, comms_ctx)
 
     dt_cf = FT(time_to_seconds(parsed_args["dt_cloud_fraction"]))
 
-    callbacks =
-        (callbacks..., call_every_dt(cloud_fraction_model_callback!, dt_cf))
+    if dt_cf > 0
+        callbacks =
+            (callbacks..., call_every_dt(cloud_fraction_model_callback!, dt_cf))
+    end
 
     if parsed_args["log_progress"] && !sim_info.restart
         @info "Progress logging enabled."
--- a/src/solver/type_getters.jl
+++ b/src/solver/type_getters.jl
@@ -830,63 +830,79 @@ function get_simulation(config::AtmosConfig)
     end
     @info "initializing diagnostics: $s"
 
-    length(diagnostics) > 0 && @info "Computing diagnostics:"
+    diagnostic_callbacks = tuple()
 
-    for writer in writers
-        writer_str = nameof(typeof(writer))
-        diags_with_writer =
-            filter((x) -> getproperty(x, :output_writer) == writer, diagnostics)
-        diags_outputs = [
-            getproperty(diag, :output_short_name) for diag in diags_with_writer
+    if length(diagnostics) > 0
+        @info "Computing diagnostics:"
+
+        for writer in writers
+            writer_str = nameof(typeof(writer))
+            diags_with_writer = filter(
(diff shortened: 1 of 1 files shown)


Sbozzolo:
@charleskawczynski CI is green if you need this :)

src/solver/type_getters.jl
```diff
@@ -511,6 +500,18 @@ function get_callbacks(parsed_args, sim_info, atmos, params, comms_ctx)
...
+        @info "Progress logging enabled."
+        callbacks = (
+            callbacks...,
+            call_every_n_steps(
+                (integrator) -> print_walltime_estimate(integrator);
+                skip_first = true,
+            ),
+        )
+    end
+
```
charleskawczynski, line 514: Can we make `cloud_fraction_model_callback` below optional too?

charleskawczynski: Ah, I'll follow up with that later.

src/solver/type_getters.jl
```diff
@@ -865,28 +866,35 @@ function get_simulation(config::AtmosConfig)
     end
     @info "Prepared diagnostic callbacks: $s"
 
-    # It would be nice to just pass the callbacks to the integrator. However, this leads to
-    # a significant increase in compile time for reasons that are not known. For this
-    # reason, we only add one callback to the integrator, and this function takes care of
-    # executing the other callbacks. This single function is orchestrate_diagnostics
+    diagnostic_callbacks = tuple()
```
charleskawczynski, line 869: @Sbozzolo, can we also skip the work of `diagnostics_iterations` above and below? Since those will not be relevant in the case of absent diagnostics?

charleskawczynski: Also, get we remove the `try-catch` with `write_field!` in it and instead just call the callback. This will be simpler, and it will have a better chance of accurately precompiling the code we want.

charleskawczynski: Same with the compute (just call the callback)

charleskawczynski: Also, the block inside `if config.parsed_args["warn_allocations_diagnostics"]` doesn't really make sense, the contents will always allocate on the first call.

Sbozzolo: > @Sbozzolo, can we also skip the work of `diagnostics_iterations` above and below? Since those will not be relevant in the case of absent diagnostics?

I changed to skip the work above. The work below is automatically skipped when there's 0 diagnostics.

> Also, get we remove the try-catch with write_field! in it and instead just call the callback. This will be simpler, and it will have a better chance of accurately precompiling the code we want.

The reason there's a try catach is to provide more context. Diagnostics are aribtrary functions. They also meant to be extended by users. This block tries to catch possible problems and point the user to the specific problem instead of having (just) a complex stacktrace. 

> Also, the block inside if config.parsed_args["warn_allocations_diagnostics"] doesn't really make sense, the contents will always allocate on the first call.

That's not a first call. The first call occurs in the try-catch.

src/solver/type_getters.jl
Sbozzolo: I added a check for the `dt`. When negative, the callback is not added.

src/solver/type_getters.jl
charleskawczynski: `diagnostics_iterations` won't exist, and you'll get an error when there's zero diagnostics.

PR #2599, opened 2024-01-29
Allow unlogged parameters in TOML files

nefrathenrici:
CloudMicrophysics parameter constructors do not log parameters that they use, triggering an error when logging parameter information in `src/solver/type_getters.jl`. 

This PR disables strict logging, allow unlogged parameter overrides with a warning.

To reproduce the bug, create a TOML file:
```
[rain_autoconversion_timescale]
value = 360
type = "float"
```
Then run:
```
import ClimaAtmos as CA
config = CA.AtmosConfig(Dict("toml" =>["toml/test.toml"], "precip_model" => "1M"))
```

### DIFF
--- a/src/solver/type_getters.jl
+++ b/src/solver/type_getters.jl
@@ -773,12 +773,13 @@ function get_simulation(config::AtmosConfig)
     end
 
     # Check that all set parameters have been used
+    # Strict logging temporarily disabled for compatibility with CloudMicrophysics overrides
     param_filepath =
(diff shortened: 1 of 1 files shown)


trontrytel:
Thank you. Hopefully we can switch it back on soon!

trontrytel (approved):

nefrathenrici:
> Thank you. Hopefully we can switch it back on soon!

At least this still warns, even if no one reads warnings!

PR #2600, opened 2024-01-29
Make number of NetCDF points YAML controllable

Sbozzolo:
It would be useful to dispatch over the grid type (with https://github.com/CliMA/ClimaCore.jl/pull/1567) to have more reasonable defaults.


szy21 (approved):
Should we add an example in ci? Maybe one of the sphere or edmf jobs?

Sbozzolo:
> Should we add an example in ci? Maybe one of the sphere or edmf jobs?

I added one example for a single column where I also show that we cannot specify 1 point (it has to be at least 2). It has to be at least 2 because of how ranges work in Julia. They hare inclusive of the extremes and ill-defined when you only have one point. I could add some logic to handle this edge case, but I don't think it's worth doing it for the time being.

szy21:
Looks great, thanks!

PR #2605, opened 2024-01-29
Use t_end to determine output frequency

Sbozzolo:
The logic is as follows:

If `t_end < 1 day` take hourly means,
if `t_end < 1 year` take daily means,
If `t_end >= 1 year` take monthly means.


szy21:
This will still cause issues for the current longrun, as they are less than a year with daily output. I will get back to you with a better logic later this afternoon.

szy21:
How about:
If t_end < 1 day take hourly means,
if t_end < 30 day take daily means,
if t_end < 120 day take 10 day means,
If t_end >= 120 day take monthly (30 day) means.

Is this too complicated?

Sbozzolo:
> How about: If t_end < 1 day take hourly means, if t_end < 30 day take daily means, if t_end < 120 day take 10 day means, If t_end >= 120 day take monthly (30 day) means.
> 
> Is this too complicated?

No, easy peasy

src/diagnostics/default_diagnostics.jl
```diff
@@ -69,130 +79,54 @@ function average_pre_output_hook!(accum, counter)
...
-"""
-    hourly_average(short_names...; output_writer)
-
-
-Return a `ScheduledDiagnostics` that computes the hourly average for the given variable.
-"""
-hourly_average(short_names; output_writer) =
-    hourly_averages(short_names; output_writer)[1]
+function frequency_averages(t_end::Real)
+    if t_end > 365 * 86400
```
szy21, line 97: Just note that we have a [parameter](https://github.com/CliMA/CLIMAParameters.jl/blob/39d2b9173f60d7ea63395166f22036b156b1d3f5/src/parameters.toml#L1571) `day`. It seems we don't have access to parameters here though, so this is fine I think.

szy21, line 97:
```suggestion
    if t_end >= 120 * 86400
```

src/diagnostics/default_diagnostics.jl
```diff
@@ -69,130 +79,54 @@ function average_pre_output_hook!(accum, counter)
...
+        return monthly_averages
+    elseif t_end > 120 * 86400
```
szy21, line 99:
```suggestion
    elseif t_end >= 30 * 86400
```

src/diagnostics/default_diagnostics.jl
```diff
@@ -69,130 +79,54 @@ function average_pre_output_hook!(accum, counter)
...
+        return tendaily_averages
+    elseif t_end > 86400
```
szy21, line 101:
```suggestion
    elseif t_end >= 86400
```
(Just me being picky and following the logic [here](https://github.com/CliMA/ClimaAtmos.jl/pull/2605#issuecomment-1915792577) lol)

szy21: Actually I don't mind too much. Just the first one `if t_end > 365 * 86400` I think it would be good to have 360 or whatever divisible by 30 .

Sbozzolo: > (Just me being picky and following the logic https://github.com/CliMA/ClimaAtmos.jl/pull/2605#issuecomment-1915792577 lol)

I read your comment and ended up implementing something compeltely different :/

szy21 (approved):
Thank you!

PR #2609, opened 2024-01-30
Assorted small improvements

Sbozzolo:
I was trying to debug the issue with a test taking forever to run on ubuntu on GitHub and I ended up cleaning a few small things. I identified the problem with the interpolating the diagnostics. I don't still don't know why this takes so long on GitHub: it is approximately 1000x slower than everywhere else. I verified that the package versions are identical (between a "good" run and a "bad" one). In any case, we don't need any diagnostic for the specific test, so I disabled them. 

The main real change in this PR is to reduce the number of syncs that are forced for NetCDF files. `sync` (= flush) was introduced because when running on a GPU, the NetCDF file will be buffered until the end of the run. This was problematic for simulations that are interrupted for any reason (all the data is lost). 

Now, `sync` is called once per cycle, ie, once all the diagnostics are called, accumulated, and reduced.

Working (ie, fast) build: https://github.com/CliMA/ClimaAtmos.jl/actions/runs/7759147548/job/21162571999

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,6 +18,7 @@ jobs:
   test:
     name: ci ${{ matrix.version }} - ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
+    timeout-minutes: 30
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -15,6 +15,7 @@ concurrency:
 jobs:
   docbuild:
     runs-on: ubuntu-latest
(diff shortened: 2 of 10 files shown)


charleskawczynski:
I think we should enable a way to statically elide the diagnostics (#2598). I like a handful of changes in this PR (using ClimaCore API for getproperty calls, GHA timeout times), but the diagnostic pieces are pretty different, and I think need more though, since the flame graph is showing a pretty big perf hit.

Can you split the PR up? I'm happy to get non-controversial parts in asap.

charleskawczynski (approved):

PR #2610, opened 2024-01-30
Update Thermodynamics only

charleskawczynski:
This is a peel off from #2581.


szy21:
The MSE change is as expected, so you can just increase the ref counter. The allocation increases are still there though.

charleskawczynski:
Superseded

PR #2619, opened 2024-02-01
Extract nc files from tar for reproducibility tests

charleskawczynski:
This PR fixes CI:
 - We converted the NC files to a tarball (I'm not sure why, but this is fine)
 - And the remaining NC files were cleaned up / deleted. So, our reproducibility logic could not find them.
 - This PR attempts to extract the NC files from the tarball before comparing against main.

cc @Sbozzolo, @LenkaNovak


Sbozzolo:
The reason they were put in a tarball is because of buildkite artifacts. It is not possible to order the artifacts, and when you have 80 artifacts finding the summary pdf became an arduous task.

Sbozzolo (approved):

charleskawczynski:
> The reason they were put in a tarball is because of buildkite artifacts. It is not possible to order the artifacts, and when you have 80 artifacts finding the summary pdf became an arduous task.

That's what I figured it might be, sounds good.

PR #2623, opened 2024-02-02
Flush diagnostics once per cycle

Sbozzolo:


src/solver/type_getters.jl
```diff
@@ -879,6 +879,12 @@ function get_simulation(config::AtmosConfig)
     end
     @info "Prepared diagnostic callbacks: $s"
 
+    # One "cycle" is defined as the collection of iterations needed to go through all the
+    # (diagnostic) callbacks
```
charleskawczynski, line 883: cycle, defined in `solve!` is a slightly different definition:

```
Run `step!` the least common multiple times
for all callbacks. i.e., all callbacks will
have been called at least `n_cycles` times.
```
In other words, if callback A and B are called every 3 and 4 steps, then a cycle is 11 steps. (maybe it'd be helpful to include that example in the docs). Would it be better to use a cycle in the way it's currently defined?

Sbozzolo: Yes, my comment is wrong, but that's what I meant.

Sbozzolo:
@charleskawczynski 

I reorderd the callbacks here, but this is not needed for the I/O change. Should I keep or remove this commit?

charleskawczynski:
> @charleskawczynski
> 
> I reorderd the callbacks here, but this is not needed for the I/O change. Should I keep or remove this commit?

Either way is fine, I think we should probably rewrite a bit of `get_simulation`. All of the diagnostics should really be grabbed in the callbacks, along with the others. Having large functions with closures and many variables is the sort of thing that increases compile times. I plan on following up with this in or after #2598.

Sbozzolo:
> > @charleskawczynski
> > I reorderd the callbacks here, but this is not needed for the I/O change. Should I keep or remove this commit?
> 
> Either way is fine, I think we should probably rewrite a bit of `get_simulation`. All of the diagnostics should really be grabbed in the callbacks, along with the others. Having large functions with closures and many variables is the sort of thing that increases compile times. I plan on following up with this in or after #2598.

The diagnostics are handled separately because they have to be called separately after the integrator is initialized. But, I agree, that there's room for improvement.

Sbozzolo:
> > > @charleskawczynski
> > > I reorderd the callbacks here, but this is not needed for the I/O change. Should I keep or remove this commit?
> > 
> > 
> > Either way is fine, I think we should probably rewrite a bit of `get_simulation`. All of the diagnostics should really be grabbed in the callbacks, along with the others. Having large functions with closures and many variables is the sort of thing that increases compile times. I plan on following up with this in or after #2598.
> 
> The diagnostics are handled separately because they have to be called separately after the integrator is initialized. But, I agree, that there's room for improvement.

Also because the timings on builkite seem to indicate that this change impacts performance quite significantly (not in the direction I was hoping for)

charleskawczynski:
> The diagnostics are handled separately because they have to be called separately after the integrator is initialized. But, I agree, that there's room for improvement.

Sure, but we can still rewrite `get_simulation` and call those specific callbacks separately after the integrator is initialized 🙂

PR #2626, opened 2024-02-02
Update precipitation tendency to fix boundary condition

dennisYatunin:
## Purpose 

This PR changes the advection of precipitation at terminal velocity to use the `RightBiasC2F()` operator instead of the `vertical_transport!` function. This fixes the instability from the bottom boundary condition that we have been seeing in simulations with one-moment microphysics.

This also removes `precip_upwinding` from the `config`, since we will always use the first-order downward biasing operator `RightBiasC2F()`.

----


trontrytel (approved):
Thank you! I confirm that it makes the precipitating case run for 10 days again. Lets merge it in

PR #2628, opened 2024-02-05
Add support for comparison plots

Sbozzolo:
With this PR, `make_plots` can now also produce a PDF report that compares multiple outputs. 
Suppose you ran the simulation `aquaplanet` three times and the output is in `output/run1`, `output/run2`, `output/run3`. 
With `make_plots(Val(:aquaplent), "output/run1")` the usual PDF report is generated. But with With `make_plots(Val(:aquaplent), ["output/run1", "output/run2", "output/run3")`  the same report is generated but with the output from the three simulations side-by-side.

For example, for two simulations, the output would look something like:
image
(In this case the two columns are identical because I didn't have two datasets to compare, so I compared twice the same data).

For the time being, no check is performed that the two folders have the same output variables.

### DIFF
--- a/post_processing/ci_plots.jl
+++ b/post_processing/ci_plots.jl
@@ -1,3 +1,43 @@
+# ci_plots.jl:
+#
+# Automatic produce a PDF report of a given job_id.
+#
+# FAQ: How to add a new job?
+#
+# To add a new job, you need to define a new method for the `make_plots` function. The
+# `make_plots` has to take two arguments: a `Val(:job)`, and a list of output paths. In most
+# cases, `make_plots` will work with just one output path, but it still has to accept a list
+# so that the function can be used to compare different outputs with the same report.
+# Support for comparison is for the most part automated as long as `map_comparison` and
+# `make_plot_generic` are used.
+#
+# Consider for example
+# ```julia
+#     function make_plots(
+#         ::Val{:box_hydrostatic_balance_rhoe},
+#         output_paths::Vector{<:AbstractString},
+#     )
+#         simdirs = SimDir.(output_paths)
+#         short_names, reduction = ["wa", "ua"], "average"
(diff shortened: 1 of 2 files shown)


post_processing/ci_plots.jl
```diff
@@ -98,17 +154,66 @@ function parse_var_attributes(var)
...
     output_name = "summary",
     summary_files = String[],
     MAX_NUM_COLS = 1,
-    MAX_NUM_ROWS = 4,
+    MAX_NUM_ROWS = min(4, length(vars)),
     kwargs...,
 )
+
+    # We treat a Vector of 1 element as a normal string
+    length(output_path) == 1 && (output_path = output_path[1])
```
charleskawczynski, line 207:
```suggestion
    if output_path isa Vector && length(output_path) == 1
        output_path = output_path[1]
    end
```

charleskawczynski: Actually, the logic is a bit complicated because `output_path` can be multiple things. I would suggest either:
 - Writing this exclusively for a vector `output_path`, or precomputing `vec_len_is_1 = output_path isa Vector && length(output_path) == 1` and using this for the special casing logic (and index into `output_path` on the fly)

charleskawczynski: Maybe there's not too many edge cases. Just watch out for/fix when `length(output_path::String) == 1`, resulting in a `Char`.

Sbozzolo: I tried cleaning up. I agree that it is not the best, but hopefully now it is more clear what's happening.

Sbozzolo: @charleskawczynski any outstanding issues with this PR?

charleskawczynski: No, I think it looks good. I just added it to the merge queue

PR #2634, opened 2024-02-06
Make RRTMGP inherit float type

charleskawczynski:
This PR makes RRTMGP inherit the given float type, so that RRTMGP can run with Float32.


charleskawczynski:
Superseded by #2704.

PR #2636, opened 2024-02-06
Add 1-moment microphysics coupled to grid-mean state to longruns

trontrytel:
This PR adds the 1-moment microphysics to the longruns. The simulation can run for about 24 days before crashing. I'll be looking into it in the upcoming days. But maybe this is good enough to also start testing on the GPU and/or with topography?

There are some negative numbers in the simulations and sometimes they are too big for comfort. I'll be looking into that.

The actual heavy lifting regarding the boundary conditions for precipitation was done here: https://github.com/CliMA/ClimaAtmos.jl/pull/2626

On top of that here I'm:
- decreasing the hyperdiffusion for precipitation
- limiting the precipitation tendencies individually by tracer/dt. This does not guarantee the positivity, but it's at least some approximation of what could be needed.
- allowing the negative precipitation tracers to be advected

### DIFF
--- a/docs/src/equations.md
+++ b/docs/src/equations.md
@@ -375,3 +375,17 @@ It is assummed that some fraction ``\alpha`` of snow is melted during the proces
 ```math
 \frac{d}{dt} \rho e = \rho \mathcal{S}_{acc} ((1+\alpha) I_{liq} - \alpha I_{ice} + \Phi)
 ```
+
+### Stability and positivity
+
+All source terms are individually limited such that they don't exceed the
+  available tracer specific humidity.
+```math
+\mathcal{S}_{x \rightarrow y} = min(\mathcal{S}_{x \rightarrow y}, \frac{q_{x}}{dt})
+```
+This will not ensure positivity because the sum of all source terms,
+  combined with the advection tendency,
+  could still drive the solution to negative numbers.
+It should however help mitigate some of the problems.
+The source terms functions treat negative specific humidities as zeros,
+  so the simulations should be stable even with small negative numbers.
--- a/src/parameterized_tendencies/microphysics/precipitation.jl
+++ b/src/parameterized_tendencies/microphysics/precipitation.jl
@@ -211,6 +211,7 @@ function compute_precipitation_cache!(Y, p, colidx, ::Microphysics1Moment, _)
     α(ts) = TD.Parameters.cv_l(thp) / Lf(ts) * (Tₐ(ts) - cmp.ps.T_freeze)
 
     qₚ(ρqₚ, ρ) = max(FT(0), ρqₚ / ρ)
+    limit(q, dt) = q / dt
 
     # zero out the source terms
     @. ᶜSqₜᵖ[colidx] = FT(0)
@@ -223,7 +224,7 @@ function compute_precipitation_cache!(Y, p, colidx, ::Microphysics1Moment, _)
 
     # rain autoconversion: q_liq -> q_rain
     @. ᶜSᵖ[colidx] = min(
-        qₗ(ᶜts[colidx]) / dt,
+        limit(qₗ(ᶜts[colidx]), dt),
         CM1.conv_q_liq_to_q_rai(
             cmp.pr.acnv1M,
             qₗ(ᶜts[colidx]),
@@ -236,7 +237,7 @@ function compute_precipitation_cache!(Y, p, colidx, ::Microphysics1Moment, _)
 
     # snow autoconversion assuming no supersaturation: q_ice -> q_snow
     @. ᶜSᵖ[colidx] = min(
-        qᵢ(ᶜts[colidx]) / dt,
+        limit(qᵢ(ᶜts[colidx]), dt),
         CM1.conv_q_ice_to_q_sno_no_supersat(
             cmp.ps.acnv1M,
             qᵢ(ᶜts[colidx]),
@@ -249,7 +250,7 @@ function compute_precipitation_cache!(Y, p, colidx, ::Microphysics1Moment, _)
 
     # accretion: q_liq + q_rain -> q_rain
     @. ᶜSᵖ[colidx] = min(
-        qₗ(ᶜts[colidx]) / dt,
+        limit(qₗ(ᶜts[colidx]), dt),
         CM1.accretion(
             cmp.cl,
             cmp.pr,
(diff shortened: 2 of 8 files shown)


src/parameterized_tendencies/microphysics/precipitation.jl
```diff
@@ -211,6 +211,7 @@ function compute_precipitation_cache!(Y, p, colidx, ::Microphysics1Moment, _)
     α(ts) = TD.Parameters.cv_l(thp) / Lf(ts) * (Tₐ(ts) - cmp.ps.T_freeze)
 
     qₚ(ρqₚ, ρ) = max(FT(0), ρqₚ / ρ)
+    limit(q, dt) = q / dt
```
akshaysridhar, line 214: Could you please add a comment about the limiter application in the docs section under `Microphysics source terms` so we can track it against future changes ? (Presumably when we have better approximations the intent is to relax or update these limiters?)

toml/longrun_1M.toml
```diff
@@ -0,0 +1,15 @@
+[zd_rayleigh]
+value = 35000.0
+type = "float"
+
+[alpha_rayleigh_uh]
+value = 0.0
+type = "float"
+
+#[rain_autoconversion_timescale]
```
akshaysridhar, line 9: Is this commented block intentional? It seems the default values are 1000s, 100s, for the `rain`, `snow` vars respectively.

toml/longrun_1M.toml
```diff
@@ -0,0 +1,15 @@
+[zd_rayleigh]
+value = 35000.0
+type = "float"
+
+[alpha_rayleigh_uh]
+value = 0.0
+type = "float"
```
akshaysridhar, line 7:
```suggestion
```

If I understand this correctly the default value has been updated here, we shouldn't need this block : https://github.com/CliMA/CLIMAParameters.jl/blob/110de02b2c1d12673b130ae94295e1628bdf7f4e/src/parameters.toml#L1745

akshaysridhar (changes requested):
Thanks @trontrytel. The changes seem reasonable based on the PR description; 
(1) Can you please add a comment about the tendency limiter in the ClimaAtmos microphysics docs section? 
(2) There is a commented set of parameters in the toml file - if you can confirm this is intentional we can close that issue.

toml/longrun_1M.toml
trontrytel: Yes, but I can delete it for now. I don't know yet what would be best/most stable for the Atmos runs. I just wanted to remember what are the parameter names to test them later.

docs/src/equations.md
```diff
@@ -375,3 +375,17 @@ It is assummed that some fraction ``\alpha`` of snow is melted during the proces
 ```math
 \frac{d}{dt} \rho e = \rho \mathcal{S}_{acc} ((1+\alpha) I_{liq} - \alpha I_{ice} + \Phi)
 ```
+
```
akshaysridhar, line 378: Thanks!

akshaysridhar (approved):

szy21:
You will need to add the job to ci_plots if you want to see the beautiful plots. Also you can add it the longrun jobs documentation if you want - I don't care too much about it:)

akshaysridhar:
> You will need to add the job to ci_plots if you want to see the beautiful plots. Also you can add it the longrun jobs documentation if you want - I don't care too much about it:)

True, I'm happy for this to be a separate PR (since this doesn't use default diagnostics)

trontrytel:
> You will need to add the job to ci_plots if you want to see the beautiful plots. Also you can add it the longrun jobs documentation if you want - I don't care too much about it:)

o.O I do want plots and ci_docs! I'll add it Thank you

trontrytel:
I'll add to the CI docs, but will do the plotting as a separate PR. It would be good to plot precipitation by default, if we are running with the 1-moment scheme

docs/src/equations.md
```diff
@@ -375,3 +375,17 @@ It is assummed that some fraction ``\alpha`` of snow is melted during the proces
...
+Each source terms are individually limited such that they don't exceed the
+  available tracer specific humidity.
+```math
+\mathcal{S}_{x \rightarrow y} = min(\mathcal{S}_{x \rightarrow y}, \frac{q_{x}}{dt})
+```
+This will not ensure positivity because the sum of all source terms,
+  combined with the advection tendency,
+  could still drive the solution to negative numbers.
+It should however help mitigate some of the problems.
+The source terms functions treat negative specific humidities as zeros,
```
LenkaNovak, line 390: Does this mean we're not conserving water in this case? 🤔

docs/src/equations.md
```diff
@@ -375,3 +375,17 @@ It is assummed that some fraction ``\alpha`` of snow is melted during the proces
 ```math
 \frac{d}{dt} \rho e = \rho \mathcal{S}_{acc} ((1+\alpha) I_{liq} - \alpha I_{ice} + \Phi)
 ```
+
+### Stability and positivity
+
+Each source terms are individually limited such that they don't exceed the
```
LenkaNovak, line 381: Each -> All

src/parameterized_tendencies/microphysics/precipitation.jl
LenkaNovak, line 214: I know this is only rearranging what we had before, so I presume the answer is no, but shouldn't `dt` here be the RK stage `dt`, rather than the full `dt`?

LenkaNovak (approved):
Thank you, @trontrytel. I only had a few clarification questions.

docs/src/equations.md
trontrytel: I don't think so. The negative numbers are only treated as zeros when computing the tendencies for each process. We don't actually overwrite anything in the state vector in Atmos.

For example if I have small negative rain specific humidity in one of my grid cells, I will still advect it around and keep the negative value in `Y` unchanged. But I will not compute the evaporation or collision tendency for that grid cell. This may lead to errors in the microphysics representation, but not to the conservation budget for mass/water etc.

src/parameterized_tendencies/microphysics/precipitation.jl
trontrytel: It should be RK stage `dt` but the limiting should be done on the full tendency and not individual components of it. 

The problem with that is that if I were to limit the full tendency I would have to know how it affects the individual components. Because I would have to know how to adjust the corresponding source terms to density, energy, etc. That's complicated, because for example some sink terms to rain lead to changes in density/energy of the working fluid, but others do not.

So to keep it simple for now, we instead apply the limiter on each tendency component and use the `dt` and not RK `dt`. It's not accurate, but the goal is to keep large negative numbers from developing, not necessarily to always guarantee positivity. We were even discussing trying to limit by more than just `dt` and testing how that would affect the stability vs accuracy.

docs/src/equations.md
LenkaNovak: That makes sense!

src/parameterized_tendencies/microphysics/precipitation.jl
LenkaNovak: Ahh ok, thank you for explaining! As long as it's an intentional choice ;-)

PR #2637, opened 2024-02-07
Add start date to NetCDF files

Sbozzolo:
Until we have actual dates in our simulation, we need to save the start date to compare with observations/other models.


trontrytel (approved):
Makes sense to me!

PR #2643, opened 2024-02-07
Move diagnostics preparation into callbacks

charleskawczynski:
Supersedes #2598.

### DIFF
--- a/src/solver/type_getters.jl
+++ b/src/solver/type_getters.jl
@@ -886,6 +886,7 @@ function get_simulation(config::AtmosConfig)
                         t_start,
                         output_dir,
                     )
+                    NCDatasets.sync(diag.output_writer.open_files[output_dir])
                 else
                     # Add to the accumulator
 
--- a/perf/benchmark_netcdf_io.jl
+++ b/perf/benchmark_netcdf_io.jl
@@ -66,6 +66,7 @@ for device in keys(timings)
         integrator.t,
         simulation.output_dir,
     )
+    NCDatasets.sync(netcdf_writer.open_files[simulation.output_dir])
     # Now, profile
     @info "Profiling ($device_name)"
     prof = Profile.@profile CAD.save_diagnostic_to_disk!(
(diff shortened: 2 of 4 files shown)


src/solver/type_getters.jl
```diff
@@ -825,84 +608,22 @@ function get_simulation(config::AtmosConfig)
...
-    # The generic constructor for SciMLBase.CallbackSet has to split callbacks into discrete
-    # and continuous. This is not hard, but can introduce significant latency. However, all
-    # the callbacks in ClimaAtmos are discrete_callbacks, so we directly pass this
-    # information to the constructor
-    continuous_callbacks = tuple()
-    discrete_callbacks = (callback..., diagnostic_callbacks)
-
-    s = @timed_str begin
-        all_callbacks =
-            SciMLBase.CallbackSet(continuous_callbacks, discrete_callbacks)
```
Sbozzolo, line 905: Can you leave this explanation?

src/callbacks/callbacks.jl
```diff
@@ -347,3 +347,271 @@ function reset_graceful_exit(output_dir)
...
+
+    s = @timed_str begin
+        diagnostics, writers = get_diagnostics(parsed_args, atmos, spaces)
+    end
+    @info "initializing diagnostics: $s"
+
+    if length(diagnostics) > 0
+        @info "Computing diagnostics:"
+    else
+        return (), ()
```
Sbozzolo, line 509: Can please you document the expected return objects/types for this function?

charleskawczynski: I actually had to fix this because it failed in the empty case, I switched to NamedTuples, so it should also be clearer.

Sbozzolo (approved):

charleskawczynski:
Something is breaking, and I must have missed something. My computer did crashed in-between making changes, so maybe not everything was saved. There's not actually that many changes, so I may just re-apply them to be safe.

Sbozzolo:
> Something is breaking, and I must have missed something. My computer did crashed in-between making changes, so maybe not everything was saved. There's not actually that many changes, so I may just re-apply them to be safe.

I will look into this.

charleskawczynski:
Accidentally force pushed to this, it was pretty stale anyway, closing.

PR #2644, opened 2024-02-07
Add function barriers to nc write_field

charleskawczynski:
This PR adds a function barrier to `write_field!`, plus `@inbounds`, in attempt to improve the performance.


charleskawczynski:
This didn't have the impact I was hoping, but we should look into the index ordering, to make sure that data is being load/stored contiguously

PR #2645, opened 2024-02-07
Try CTS branch with nvtx annotations and ranges

charleskawczynski:
This PR tries out the CTS branch with annotations and ranges.

I'll take a look at the NVTX output, and the memory footprint of some jobs, @Sbozzolo, since these NVTX range calls wrap around the callbacks, do you know which job is diagnostics heavy? I'd like to look at the job that's most likely to stress NVTX-related issues

### DIFF
--- a/docs/Manifest.toml
+++ b/docs/Manifest.toml
@@ -2,7 +2,7 @@
 
 julia_version = "1.10.0"
 manifest_format = "2.0"
-project_hash = "1b7b7b12f210f5da35bb4f69ad563a64237fde09"
+project_hash = "e87302db215ea786d20636f8bd0f2a0f02bd5a64"
 
 [[deps.ADTypes]]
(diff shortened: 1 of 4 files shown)


Sbozzolo:
Among the regular jobs, EDMF jobs add lots of diagnostics that are run and output frequently. Prognostic EDMF in particular.

charleskawczynski:
For a prognostic EDMF job:
 - Main: `[ Info: Memory currently used (after solve!) by the process (RSS): 6.70 GiB`
 - This branch: `[ Info: Memory currently used (after solve!) by the process (RSS): 6.73 GiB`
so, this isn't too bad. Maybe we can work with the one `@range` call.

charleskawczynski:
The allocations don't look too bad, and I can see the new ranges, so I'm going to merge the CTS PR.

PR #2646, opened 2024-02-07
Add IO benchmark job

Sbozzolo:
Adds a GPU job with lots of diagnostics that always runs on `clima` and produces a flamegraph for 1 simulated day.

### DIFF
--- a/.buildkite/gpu_pipeline/pipeline.yml
+++ b/.buildkite/gpu_pipeline/pipeline.yml
@@ -105,6 +105,22 @@ steps:
   - group: "DYAMOND GPU strong scaling"
     steps:
 
+      - label: "gpu_aquaplanet_dyamond with diagnostics - strong scaling - 1 GPU"
+        command:
+          - mkdir -p gpu_aquaplanet_dyamond_ss_diag_1process
+          - >
+            srun --cpu-bind=threads --cpus-per-task=4
+            julia --threads=3 --color=yes --project=examples examples/hybrid/driver.jl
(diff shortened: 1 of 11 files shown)


.buildkite/gpu_pipeline/pipeline.yml
```diff
@@ -122,3 +123,11 @@ steps:
           slurm_gpus_per_task: 1
           slurm_cpus_per_task: 4
           slurm_ntasks: 4
+
+      - label: ":fire: Target EDMF Aquaplanet 1 thread"
+        key: "flame_longrun_aquaplanet_rhoe_equil_55km_nz63_clearsky_diagedmf_0M"
+        command:
+          - julia --color=yes --project=examples perf/longer_flame.jl config/longrun_configs/longrun_aquaplanet_rhoe_equil_55km_nz63_allsky_diagedmf_0M.yml
```
charleskawczynski, line 130:
```suggestion
          - julia --color=yes --project=perf perf/longer_flame.jl config/longrun_configs/longrun_aquaplanet_rhoe_equil_55km_nz63_allsky_diagedmf_0M.yml
```

ProfileCanvas is in the perf env, if you want to just use that instead

.buildkite/gpu_pipeline/pipeline.yml
```diff
@@ -24,6 +24,7 @@ steps:
     command:
       - echo "--- Instantiate examples"
       - julia --project=examples -e 'using Pkg; Pkg.instantiate(;verbose=true)'
+      - julia --project=examples -e 'using Pkg; Pkg.add("ProfileCanvas")'  # For the flamegraph
```
charleskawczynski, line 27:
```suggestion
```

PR #2647, opened 2024-02-08
remove default instantaneous edmf diagnostics

szy21:
## Purpose 

Removes default instantaneous edmf diagnostics and specifies it in the yaml file when needed. I think this is better, as it doesn't rely on users to remember to set output_default_diagnostics to false when running long simulations with edmf. Also specifies `netcdf_interpolation_num_points` for edmf box cases (which make them run much faster in ci) and removes an unused yaml file.

## Content

----


szy21:
> I think we should fix the root issue (I've started working on this). If these diagnostics are useful, let's keep them.

They are useful for the ci, but not for the longrun, at least for now. We should still fix them, but we will turn them off for the longrun anyway. Depending how we set the default, users need to either remember adding new diagnostics when running short edmf jobs for debugging, or set default to false and specify their own diagnostics when running long edmf jobs. I think using time-average diagnostics as default and specifying instantaneous diagnostics when needed is better.

Sbozzolo:
> > I think we should fix the root issue (I've started working on this). If these diagnostics are useful, let's keep them.
> 
> They are useful for the ci, but not for the longrun, at least for now. We should still fix them, but we will turn them off for the longrun anyway. Depending how we set the default, users need to either remember adding new diagnostics when running short edmf jobs for debugging, or set default to false and specify their own diagnostics when running long edmf jobs. I think using time-average diagnostics as default and specifying instantaneous diagnostics when needed is better.

Okay!

Sbozzolo (approved):

PR #2649, opened 2024-02-08
Move diagnostics preparation into callbacks (round 2)

charleskawczynski:
This is a followup to #2648, supersedes #2643.


charleskawczynski:
This is pretty stale and most of this has already been applied in peel-off PRs.

PR #2652, opened 2024-02-08
Try ClimaCore avoid steprange branch

charleskawczynski:
This PR tries out the ClimaCore avoid steprange branch, which will hopefully reduce the register pressure in a few places.

I looked at a recent report, and I see `Registers Per Thread: 44` in `field_solve_kernel!` which, IIUC, is being used in the implicit jacobian advection block:

```julia
    advection_blocks = (
        (
    ...
        (@name(f.u₃), @name(f.u₃)) => similar(Y.f, TridiagonalRow_C3xACT3),
    )
```
So, I think this shouldn't be able to give a false negative. 🤞🏻 

@dennisYatunin, do you know off hand if this is being used in the dry baroclinic wave job?

### DIFF
--- a/docs/Manifest.toml
+++ b/docs/Manifest.toml
@@ -2,7 +2,7 @@
 
 julia_version = "1.10.0"
 manifest_format = "2.0"
-project_hash = "1b7b7b12f210f5da35bb4f69ad563a64237fde09"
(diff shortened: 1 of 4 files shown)


charleskawczynski:
I didn’t see any impact on threads, but I’m still not100% sure that this exercised the right code paths

PR #2661, opened 2024-02-08
Save parameters and config to output

nefrathenrici:
This PR saves the parameters and parsed_args to the output dir.


charleskawczynski:
One benefit of saving the yank file to the output folder, and not in the reproducibility subfolder, is that we could open/read those files from buildkite’s artifacts. Might that be better?

also, what is in the restart file?

charleskawczynski (approved):

PR #2664, opened 2024-02-09
Profile cloud diagnostics

charleskawczynski:
This PR adds profiling for the cloud diagnostics (which calls `buoyancy_gradients`), in attempt to diagnose #2530.


szy21:
@charleskawczynski Should we close this?

PR #2666, opened 2024-02-12
Zs/ft radiation

szy21:
## Content

----


szy21:
@charleskawczynski I think this might fix the radiation problem, but I am hard coding something and it has broadcasting errors. If you want you can try to fix the broadcasting error to see if all jobs pass. Otherwise we can just change RRTMGP (which is what we want in the end).

PR #2667, opened 2024-02-12
Cover entire globe with NetCDF

Sbozzolo:
This PR changes the default range of latitudes from -80, 80 to -90, 90, so that we have data all over the glob.e


LenkaNovak (approved):
It could be good to add more tick marks to the summary plots, so you can see the latitudinal range better. Otherwise LGTM. Thank you!

PR #2672, opened 2024-02-13
Try to fix lr hydrostatic balance func for gpus

charleskawczynski:
I think this might be all we need-- it won't be efficient, but I don't think it needs to be since it happens at initialization

### DIFF
--- a/src/utils/discrete_hydrostatic_balance.jl
+++ b/src/utils/discrete_hydrostatic_balance.jl
@@ -53,6 +53,8 @@ function set_discrete_hydrostatic_balanced_state!(Y, p)
     ) # broadcasting doesn't seem to work with kwargs, so define interim ls()
 end
 
+u₃_component(u::Geometry.AxisTensor) = u.u₃
+
 """
     set_discrete_hydrostatic_balanced_pressure!(ᶜp, ᶠgradᵥ_ᶜp, ᶜρ, ᶠgradᵥ_ᶜΦ, p1, colidx)
 Construct discrete hydrostatic balanced pressure `ᶜp` from density `ᶜρ`,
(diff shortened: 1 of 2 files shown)


src/utils/discrete_hydrostatic_balance.jl
```diff
@@ -80,8 +80,15 @@ function set_discrete_hydrostatic_balanced_pressure!(
     @. ᶠgradᵥ_ᶜp[colidx] = -(ᶠgradᵥ_ᶜΦ[colidx] * ᶠinterp(ᶜρ[colidx]))
     ᶜp_data = Fields.field_values(ᶜp[colidx])
     ᶠgradᵥ_ᶜp_data = Fields.field_values(ᶠgradᵥ_ᶜp[colidx])
-    ᶜp_data[1] = p1
+    ᶜp_data_lev₋₁ = Spaces.level(p1, 1-1)
     for i in 2:Spaces.nlevels(axes(ᶜp))
-        ᶜp_data[i] = ᶜp_data[i - 1] + ᶠgradᵥ_ᶜp_data[i].u₃
+        ᶜp_data_lev₋₁ = Spaces.level(p1, i-1)
```
szy21, line 85:
```suggestion
        ᶜp_data_lev₋₁ = Spaces.level(ᶜp_data, i-1)
```

szy21 (approved):
Thanks! Here are some suggestions based on my understanding. Is this what you meant?

charleskawczynski:
> Is this what you meant?

Yes, thanks for the suggestions. I think that branch was somewhat unfinished. I think this should work (although it won't be very performant). It's only done during initialization, so I think it should be fine. We could also split the CPU/GPU implementations so that the CPU doesn't suffer from this awkward pattern.

szy21:
I think this is fine. As you mentioned it's only used during the initialization. Also it's only used in this one job.

trontrytel (approved):

PR #2673, opened 2024-02-13
Hoist and precompute common variables

charleskawczynski:
This PR hoists and precomputes `ᶜgradᵥ_θ_virt,ᶜgradᵥ_q_tot,ᶜgradᵥ_θ_liq_ice`, to simplify some very expensive kernels. A step towards #2530.

Let's see how the flame graph updates with this PR.

The flame graphs show an improvement with the kernel but, more interestingly, the gradient-interpolations are pretty expensive. Since this PR hoists those computations, this PR should be more performant since those computations are repeated in the main branch. We should probably look into the performance of the gradient-interpolations, but we can also merge this in case we can't improve it as much as we hoped.

### DIFF
--- a/src/cache/precomputed_quantities.jl
+++ b/src/cache/precomputed_quantities.jl
@@ -41,6 +41,8 @@ function precomputed_quantities(Y, atmos)
     @assert isnothing(atmos.turbconv_model) || isnothing(atmos.vert_diff)
     TST = thermo_state_type(atmos.moisture_model, FT)
     SCT = SurfaceConditions.surface_conditions_type(atmos, FT)
+    cspace = axes(Y.c)
+    fspace = axes(Y.f)
     n = n_mass_flux_subdomains(atmos.turbconv_model)
     gs_quantities = (;
         ᶜspecific = specific_gs.(Y.c),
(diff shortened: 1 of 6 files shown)


src/cache/precomputed_quantities.jl
```diff
@@ -83,6 +85,17 @@ function precomputed_quantities(Y, atmos)
             ᶜS_q_totʲs = similar(Y.c, NTuple{n, FT}),
             ᶜS_q_tot⁰ = similar(Y.c, FT),
         ) : (;)
+    sgs_quantities =
+        (
+            atmos.turbconv_model isa AbstractEDMF ||
+            isnothing(atmos.turbconv_model)
+        ) ?
```
szy21, line 92: We don't need this check?

src/cache/precomputed_quantities.jl
```diff
@@ -447,6 +461,15 @@ NVTX.@annotate function set_precomputed_quantities!(Y, p, t)
     @. ᶜts = ts_gs(thermo_args..., ᶜspecific, ᶜK, ᶜΦ, Y.c.ρ)
     @. ᶜp = TD.air_pressure(thermo_params, ᶜts)
 
+    if turbconv_model isa AbstractEDMF || isnothing(turbconv_model)
```
szy21, line 464: Same here, we don't need this check?

szy21 (approved):
Thanks!

src/cache/precomputed_quantities.jl
charleskawczynski: Now that look at it, I think you’re right. How is that so? We always call this? With every model?

szy21: Yes, we need this for mixing length diagnostics and cloud fraction. I guess we don't have to call it for the dry model, but if we want mixing length diagnostics we still need it.

charleskawczynski: And right now it only happens in the callbacks?

szy21: Yes

szy21: Actually, to be more accurate, it only happens in the callbacks when `isnothing(atmos.turbconv_model)` is true. If `atmos.turbconv_model isa AbstractEDMF` is true, it happens at every stage.

charleskawczynski: Ok, so we can always put it in the cache, but we can filter when it's computed (that explains the perf hit in the flame graphs)

PR #2674, opened 2024-02-13
Try comparison feature

charleskawczynski:
This PR tries out the plot comparison feature.

### DIFF
--- a/examples/hybrid/driver.jl
+++ b/examples/hybrid/driver.jl
@@ -230,8 +230,41 @@ end
 
 # Visualize the solution
 if ClimaComms.iamroot(config.comms_ctx)
+    include(
+        joinpath(pkgdir(CA), "regression_tests", "self_reference_or_path.jl"),
+    )
     @info "Plotting"
-    make_plots(Val(Symbol(reference_job_id)), simulation.output_dir)
+    path = self_reference_or_path() # __build__ path (not job path)
+    if path == :self_reference
+        make_plots(Val(Symbol(reference_job_id)), simulation.output_dir)
+    else
+
+        main_job_path = joinpath(path, reference_job_id)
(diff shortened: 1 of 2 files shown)


charleskawczynski:
@Sbozzolo, can you take a look at this?

charleskawczynski:
> Does `path` contain the NetCDF output?

Ah, it’s possible that we need to unpack them from the tar. I’ll try that first thing tomorrow

Sbozzolo:
> > Does `path` contain the NetCDF output?
> 
> Ah, it’s possible that we need to unpack them from the tar. I’ll try that first thing tomorrow

Yes, they would need to be unpacked.

charleskawczynski:
@Sbozzolo, can we merge this?

Sbozzolo:
I think we still have to fix the margins, right?

charleskawczynski:
> I think we still have to fix the margins, right?

Isn't that on the ClimaAnalysis side?

Sbozzolo:
> > I think we still have to fix the margins, right?
> 
> Isn't that on the ClimaAnalysis side?

I don't think so, I think this will be in the `make_plots_generic` function in `ci_plots`. ClimaAnalysis plots in a given GridLayout, but it is up to main routine to properly arrange the GridLayouts. 

I will look into this.

Sbozzolo:
I fixed the margin, more or less. I guess the folder path is determining the width.

I don't like this because it makes comparison harder.

charleskawczynski:
> I fixed the margin, more or less. I guess the folder path is determining the width.
> 
> I don't like this because it makes comparison harder.

Interesting, looks cool!

charleskawczynski:
Also, can we move the main branch to the left side? I like reading from left to right, and seeing where we were to where we are going 🙂

Sbozzolo (approved):

Issue #2677, opened 2024-02-14
Regression function-based surface albedo parameterization (broadband)


LenkaNovak:
Following Jin et al., 2011, as required in issue https://github.com/CliMA/ClimaCoupler.jl/issues/577

# Components
- [x] add functions
- [x] implement unit test
- [x] compare with Jin et al 2011 
- [x] integrate in driver
- [x] test for other types of radiation
- [x] docs
- [x] clean up

# Notes
- long runs are non-deterministic (expected due to the random seed), but in some cases they can lead to failing conservation errors (e.g. compare [this build](https://buildkite.com/clima/climaatmos-longruns/builds/1431#018dafe7-66a0-4bef-8901-1c0a357f9d9b) v  [this build](https://buildkite.com/clima/climaatmos-longruns/builds/1432#018db066-a41e-40bc-83e0-27adea0abe3d)

# QA
## Long runs
- [x] high res slab ocean aquaplanet with `RegressionFunctionAlbedo` passes [build](https://buildkite.com/clima/climaatmos-longruns/builds/1432#018db066-a418-45ff-847b-3502df639135)

### Direct
<img width="537" alt="Screen Shot 2024-02-14 at 4 52 50 PM" src="https://github.com/CliMA/ClimaAtmos.jl/assets/38083334/9ce450d8-cb4f-4931-a152-ea479421c2e3">

w=0 (note for latitude of 90deg this doesn't quite converge to 1 due to the accuracy of the coefficients provided in the paper, but this is currently sufficient for our purposes):
![Screen Shot 2024-02-14 at 4 54 34 PM](https://github.com/CliMA/ClimaAtmos.jl/assets/38083334/27daf30b-0067-4253-864d-78a8de183a74)

### Diffuse

<img width="426" alt="Screen Shot 2024-02-14 at 4 52 43 PM" src="https://github.com/CliMA/ClimaAtmos.jl/assets/38083334/3e842394-9e13-42d7-930e-0c1597d48f05">

![Screen Shot 2024-02-14 at 4 54 45 PM](https://github.com/CliMA/ClimaAtmos.jl/assets/38083334/53c3c2c9-74b3-4422-8fc2-d8b7656bb0ec)

### slabplanet after 20d
direct:
[...]

PR #2678, opened 2024-02-14
Regression function-based surface albedo parameterization (broadband)

LenkaNovak:
closes https://github.com/CliMA/ClimaAtmos.jl/issues/2677

----


perf/flame.jl
```diff
@@ -45,7 +45,7 @@ allocs_limit["flame_perf_diagnostics"] = 108_877_816
 allocs_limit["flame_perf_target_diagnostic_edmfx"] = 401_944
 allocs_limit["flame_sphere_baroclinic_wave_rhoe_equilmoist_expvdiff"] =
     4_018_252_656
-allocs_limit["flame_perf_target_frierson"] = 4_015_545_328
+allocs_limit["flame_perf_target_frierson"] = 4_015_546_097
```
LenkaNovak, line 48: Is this permissible?

src/cache/surface_albedo.jl
```diff
@@ -0,0 +1,195 @@
+# we are ignoring the volume reflectance for now
+abstract type SurfaceAlbedoModel end
+
+"""
+    Base.@kwdef struct ConstantAlbedo{FT} <: SurfaceAlbedoModel
+
+A constant surface albedo model. The default value is 0.38. It is used purely for idealized experiments.
+"""
+Base.@kwdef struct ConstantAlbedo{FT} <: SurfaceAlbedoModel
+    α::FT = 0.38
```
LenkaNovak, line 10: Following the current default.

docs/bibliography.bib
```diff
@@ -219,3 +219,12 @@ @article{Rauber2007
...
 }
+
+@article{Jin2011,
+  title={A new parameterization of spectral and broadband ocean surface albedo},
+  author={Zhonghai Jin, Yanli Qiao, Yingjian Wang, Yonghua Fang, and Weining Yi,},
+  journal={Opt. Express},
+  volume={19},
+  pages={26429--26443},
+  year={2011},
+}
```
Sbozzolo, line 230: Can you please add a new line here?

perf/flame.jl
Sbozzolo: It is an inference issue. It's fine.

src/cache/surface_albedo.jl
```diff
@@ -0,0 +1,195 @@
...
+    Base.@kwdef struct RegressionFunctionAlbedo{FT} <: SurfaceAlbedoModel
+
+A regression function-based surface albedo model of Jin et al. (2011, referred to as J11).
+
+
+"""
+Base.@kwdef struct RegressionFunctionAlbedo{FT} <: SurfaceAlbedoModel
+    n::FT = 1.34                                                  # relative refractive index of water and air (n = n_w/n_a) TODO: f(wavelength) for a spectrally dependent scheme
+    n0::FT = 1.34                                                 # refractive index of water for visible light
+    p::Vector{FT} = [
```
Sbozzolo, line 23: This will not go on a GPU. You need to make it an SVector

src/cache/surface_albedo.jl
```diff
@@ -0,0 +1,195 @@
...
+        -8.5778,
+        4.071,
+        -7.6446,
+        0.1643,
+        -7.8409,
+        -3.5639,
+        -2.3588,
+        10.0538,
+    ]                                                             # regression coefficients of J11 for eq. 4
+    wave_slope::Function = u -> sqrt(FT(0.003) + FT(0.00512) * u) # mean wave slope distribution width (Cox-Munk model, J11 Eq. 2)
```
Sbozzolo, line 36: This will also not go on a GPU, you need to make it a parameterized type `F <: Function`

src/cache/surface_albedo.jl
```diff
@@ -0,0 +1,195 @@
...
+    q_clear::Vector{FT} = [-0.1482, -0.012, 0.1608, -0.02244]     # regression coefficients of J11 for eq. 5a
+    q_cloud::Vector{FT} = [-0.1479, -0.1502, -0.016]              # regression coefficients of J11 for eq. 5b
```
Sbozzolo, line 38: These should also be SVectors

src/cache/surface_albedo.jl
```diff
@@ -0,0 +1,195 @@
...
+            :,
+        )
+    diffuse_sw_surface_albedo .=
+        ones(axes(direct_sw_surface_albedo)) .* reshape(
+            surface_albedo_diffuse(
+                α_model,
+            ).(λ, μ, RRTMGPI.field2array(wind_speed)),
+            1,
+            :,
+        )
```
Sbozzolo, line 102: There's a quite a bit of Field/Array gymnastics here (also using private methods such as `Geometry._norm`). 

There ought to be a cleaner way to write this, maybe @charleskawczynski can help?

src/cache/surface_albedo.jl
```diff
@@ -0,0 +1,195 @@
...
+                (
+                    (n^2 * cosθ - sqrt(n^2 - sin(cosθ)^2)) /
+                    (n^2 * cosθ + sqrt(n^2 - sin(cosθ)^2))
+                )^2
+            rf_s(n, cosθ) =
+                (
+                    (cosθ - sqrt(n^2 - sin(cosθ)^2)) /
+                    (cosθ + sqrt(n^2 - sin(cosθ)^2))
+                )^2
+            rf(n, cosθ) = 0.5 * (rf_p(n, cosθ) + rf_s(n, cosθ))
```
Sbozzolo, line 134:
```suggestion
            rf(n, cosθ) =  (rf_p(n, cosθ) + rf_s(n, cosθ)) /2
```
`0.5` is Float64, `/2` will preserve the type at no runtime cost.

src/cache/surface_albedo.jl
```diff
@@ -0,0 +1,195 @@
...
+            n = α_model.n
+            σ = α_model.wave_slope(u)
+
+            q_clear = α_model.q_clear
+            α_clear =
+                q_clear[1] +
+                q_clear[2] * σ +
+                q_clear[3] * n +
+                q_clear[4] * n * σ
+
```
Sbozzolo, line 185: Can you please remove one empty line?

src/parameterized_tendencies/radiation/radiation.jl
```diff
@@ -209,8 +209,8 @@ function radiation_model_cache(
             center_temperature = NaN, # initialized in callback
             surface_temperature = NaN, # initialized in callback
             surface_emissivity = 1,
-            direct_sw_surface_albedo = 0.38,
-            diffuse_sw_surface_albedo = 0.38,
+            direct_sw_surface_albedo = NaN, # initialized in callback
+            diffuse_sw_surface_albedo = NaN, # initialized in callback
```
Sbozzolo, line 213: Note that `NaN` is Float64, so maybe we want to be more careful with types here.

src/solver/model_getters.jl
```diff
@@ -97,6 +97,20 @@ function get_surface_model(parsed_args)
...
+        ConstantAlbedo{FT}()
+    elseif albedo_name in ("true", true, "RegressionFunctionAlbedo")
+        isnothing(parsed_args["rad"]) && error(
+            "Radiation model not specified, so cannot use RegressionFunctionAlbedo",
+        )
+        RegressionFunctionAlbedo{FT}()
+    else
+        error("Uncaught surface albedo model `$albedo_name`.")
+    end
+end
```
Sbozzolo, line 112: I don't think we should support all these options. It becomes messy and unmanagable very quickly. What would true/false be when we add a third option?

I'd suggest calling the option `albedo_model` and sticking with only `ConstantAlbedo` or `RegressionFunctionAlbedo`.

test/surface_albedo.jl
```diff
@@ -0,0 +1,201 @@
...
+            atol = eps(FT),
+        ),
+    )
+    @test all(
+        isapprox.(
+            p.radiation.radiation_model.diffuse_sw_surface_albedo,
+            0.38;
+            atol = eps(FT),
+        ),
+    )
```
Sbozzolo, line 34: Shouldn't this be exactly 0.38?

test/surface_albedo.jl
```diff
@@ -0,0 +1,201 @@
...
+        0.02277521129128255,
+        0.025376772832975958,
+        0.03181634436201161,
+        0.04633667894977999,
+        0.0765734586254003,
+        0.1349348882833189,
+        0.2313406520807579,
+        0.32091070958565293,
+    ]
+
```
Sbozzolo, line 93: How were these points obtained exactly?

test/surface_albedo.jl
```diff
@@ -0,0 +1,201 @@
...
+            #     θ_list * 180 / π,
+            #     α_dir_list,
+            #     label = "n = $n, wind = $wind",
+            #     xlabel = "θ (deg)",
+            #     ylabel = "α_dir",
+            #     yaxis = :log,
+            #     yticks = ([0.001, 0.01, 0.1, 1], ["0.001", "0.01", "0.1", "1"]),
+            #     ylims = (0.001, 1),
+            #     legend = :topleft,
+            # )
```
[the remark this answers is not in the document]
Sbozzolo: (Same with other comments.)

test/surface_albedo.jl
```diff
@@ -0,0 +1,201 @@
...
+    end
+    # png(plot, "diffuse_albedo_fig4.png")
+
+    @test all(
+        isapprox.(
+            [α_diff_list_all[1], α_diff_list_all[2], α_diff_list_all[3]],
+            [visually_confirmed_1, visually_confirmed_2, visually_confirmed_3];
+            atol = sqrt(eps(FT)),
+        ),
+    )
```
Sbozzolo, line 199: Instead of creating lists of lists and testing afterwards, can you put the test in the for loop?

Sbozzolo (commented):
It looks pretty good overall!

I left minor comments on GPU compatibility. There's some field operations that I feel could be made cleaner. 

As for the tests, I'd like to understand what "visually confirmed" means and if 16 significant digits are meaningful or not.

config/longrun_configs/longrun_aquaplanet_rhoe_equil_55km_nz63_clearsky_tvinsol_0M_slabocean.yml
```diff
@@ -16,6 +16,7 @@ approximate_linear_solve_iters: 2
 rad: "clearsky"
 idealized_insolation: false
 dt_rad: "1hours"
+variable_surface_albedo: true
```
akshaysridhar, line 19: `surface_albedo` = "SomeAlbedoType" (?) Makes the experiment clearer I think

src/cache/surface_albedo.jl
akshaysridhar: Is the default referred to in the paper? If not, can we please include the source? (Looking over Jin et. al now, will review the docs as I do this)

docs/src/surface_albedo.md
```diff
@@ -0,0 +1,40 @@
...
+Currently, there are two methods available to specify ocean albedo:
+
+### 1) `ConstantAlbedo`:
+The default value is 0.38 and is used purely for idealized experiments.
+
+### 2) `RegressionFunctionAlbedo`:
+This is an empirically-derived parameterization of the direct and diffuse surface albedo (note, reflectivity of the inner ocean is ignored) of [Jin2011](@cite). The direct reflection is calculated using the Fresnel reflection at the air-sea interface due to the difference in refractive index between air and water. Note that the current implementation is using the broadband representation (relative refractive index, $n$, is independent of the wavelength). Its formulation is:
+
+```math
+α_{dir}(λ, μ, u) = rf(n, μ) - rf(n, μ) / rf(n0, μ) * f(μ, σ)
```
akshaysridhar, line 14:
```suggestion
α_{dir}(λ, μ, u) = rf(n, μ) - rf(n, μ) / rf(n_{0}, μ) * f(μ, σ)
```

docs/src/surface_albedo.md
```diff
@@ -0,0 +1,40 @@
...
+```
+
+where:
+- λ is the wavelength
+- μ is the cosine of the solar zenith angle
+- u is the wind speed
+- σ is the mean wave slope distribution width
```
akshaysridhar, line 21: `σ` follows equation (2) in Jin et al. right (Cox and Munk model) ? Issue 577 in coupler refers to this, but we might as well be verbose in the docs here.

docs/src/surface_albedo.md
```diff
@@ -0,0 +1,40 @@
...
+- α(λ, μ, u) is the surface albedo
+- rf(n, μ) is the Fresnel reflectance
```
akshaysridhar, line 23:
```suggestion
- r_{f}(n, μ) is the Fresnel reflectance
```

docs/src/surface_albedo.md
```diff
@@ -0,0 +1,40 @@
...
+- n is the relative refractive index of water and air (n = n_w/n_a)
+- n0 is the refractive index of water for visible light
```
akshaysridhar, line 25: Should we list the parameter values here for clarity (especially where they aren't part of ClimaParameters) - these are clear from the code but not in the docs right now

docs/src/surface_albedo.md
```diff
@@ -0,0 +1,40 @@
...
+- f(μ, σ) is the regression function
```
akshaysridhar, line 26: Same with the regression coefficients following eq (4) in the paper.

docs/src/surface_albedo.md
```diff
@@ -0,0 +1,40 @@
...
+
+Diffuse reflection is calculated depending on atmospheric conditions. For clear sky, it is calculated using nearly isotropic Rayleigh scattering, and for cloudy sky, albedo is adjusted for the presence of clouds. In the current implementation we assume clear skies. The formulations are:
+
+```math
+α_{diff}(λ, μ) = -0.1482 - 0.012σ + 0.1608n - 0.022nσ
```
akshaysridhar, line 31:
```suggestion
α_{diff}(λ, μ) = -0.1482 - 0.012σ + 0.1608n - 0.0244nσ
```
following equation (5a) in Jin et al ? -

akshaysridhar: Could we also include the plot from the PR which reproduces Fig 2. in the paper ? Thanks!

docs/src/surface_albedo.md
```diff
@@ -0,0 +1,40 @@
...
+```
+
+for clear sky, and
+
+```math
+α_{diff}(λ, μ) = -0.1479 - 0.1502n - 0.016nσ
+```
```
akshaysridhar, line 38: following eq (5b) in Jin et al.

src/cache/surface_albedo.jl
```diff
@@ -0,0 +1,195 @@
...
+        Fields.field_values(Fields.level(Fields.local_geometry_field(Y.c), 1))
+
+    wind_speed = p.scratch.ᶠtemp_field_level
+    wind_speed_values = Fields.field_values(wind_speed)
+    @. wind_speed_values =
+        Geometry._norm.(
+            CT12.(u_int, int_local_geometry_values),
+            int_local_geometry_values,
+        )
+    # TODO: Is there a neater way of defining wind speed? E.g., do we want to use SF.jl windspeed? including gustiness?  Or do we want to extrapolate to the surface using the MOST profile? It would be good to expose SF.surface_conditions in that case.
```
akshaysridhar, line 84: Currently the windspeed is an internal function in the SF package; There is additionally a `recover_profile` method if you wish to use a known L_MO to compute windspeed in the surface layer.

src/cache/surface_albedo.jl
```diff
@@ -0,0 +1,195 @@
...
+            return rf(n, cosθ) - rf(n, cosθ) / rf(n0, cosθ) * f(cosθ, σ)
+        end
+    return α_dir
+end
+
+"""
+    surface_albedo_diffuse!(α_model::RegressionFunctionAlbedo)
+
+Set the surface albedo using the Jin et al. (2011) empirical formulation.
+For now we assume that the cloud fraction is 0.0.
```
akshaysridhar, line 166: Can we add a `!!! note` block in the docs listing any such assumptions?

akshaysridhar (commented):
Thanks @LenkaNovak ; I've left some comments here; happy to discuss / help make changes and review again.

docs/src/surface_albedo.md
akshaysridhar: Wait, there was some issue with this right? (discussed during the Thursday meeting)

src/cache/surface_albedo.jl
LenkaNovak: Ahh, it's not following the paper. It's following the original ClimaAtmos default value for the contant albedo. Maybe it was 0.38 because it is approximately the average of the geometric and bond albedos of Earth? Tbh I would expect the bond albedo, but since the constant albedo is only used for idealized simulations, maybe it doesn't matter too much either way. I kept the same value as on main, so as to confirm that the regression tests are still passing.

LenkaNovak: @szy21 , do you know why we chose 0.38?

src/solver/model_getters.jl
LenkaNovak: Good point! 👍

test/surface_albedo.jl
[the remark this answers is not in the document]
LenkaNovak: I was hoping to leaving it here, in case someone wants to replot the plots that compare this with the paper plots. I can uncomment it, but in the other repos, we're trying not to depend on Plots in `/test`

test/surface_albedo.jl
LenkaNovak: Sure :)

test/surface_albedo.jl
LenkaNovak: Oh, actually, I may need to keep it within the eps precision, since `extrema = (0.3799999952316284, 0.3799999952316284)`

test/surface_albedo.jl
LenkaNovak: From inspecting the output using the plots, deciding it looks close enough to the Jin paper's plots and then printing those values as a reference. So this function should always produce results that reproduce the paper. 🤞 I agree that it could be clarified though. I can add a comment and rename the variables.

docs/src/surface_albedo.md
LenkaNovak: No this was me! Great eye! 👁️

test/surface_albedo.jl
LenkaNovak: Truncated with better docs, as discussed. Thank you for the suggestions!

src/cache/surface_albedo.jl
LenkaNovak: Sounds good. I opened an issue for this: https://github.com/CliMA/ClimaAtmos.jl/issues/2693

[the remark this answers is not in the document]
LenkaNovak: Yes, discussed offline and below

src/cache/surface_albedo.jl
szy21: It's from [this](https://journals.ametsoc.org/view/journals/clim/21/15/2007jcli2065.1.xml) paper, to achieve a relatively reasonable global mean energy balance. We should make it a parameter in the future, but this doesn't matter much.

LenkaNovak: Makes sense, thanks!

src/cache/surface_albedo.jl
LenkaNovak: I know. I've opened an [issue](https://github.com/CliMA/ClimaAtmos.jl/issues/2693) to expose a wrapper for the wind speed in particular, so we can make use of SF.jl, but we should probably address this more generally.

src/cache/surface_albedo.jl
LenkaNovak: good point!

src/parameterized_tendencies/radiation/radiation.jl
LenkaNovak: I've followed how we initialize the rest of the cache here, but that is a great point, and we should look into it. I opened an issue here: https://github.com/CliMA/ClimaAtmos.jl/issues/2695. This is maybe something that could be relevant to the float type instabilities @sriharshakandala , @juliasloan25  and @akshaysridhar  have been dealing with. 🤔

src/cache/surface_albedo.jl
szy21: I also think we should be able to simplify this. @dennisYatunin may be able to help too.

src/cache/surface_albedo.jl
```diff
@@ -0,0 +1,204 @@
...
+            n = α_model.n
+
+            # refractive index of water for visible light
+            n0 = α_model.n0
+
+            # mean wave slope distribution width
+            σ = α_model.wave_slope(u)
+
+            # Fresnel reflectance (assuming equal contribution of the p-polorized and s-polarized components, and using the perfect dielectric medium)
+            sin(cosθ) = sqrt(1 - cosθ^2)
```
szy21, line 131:
```suggestion
            sinθ(cosθ) = sqrt(1 - cosθ^2)
```
I think this is more consistent with the doc and less confusing, as sin(cosθ) makes me think it's the sin of cosθ when I look at it first time. But it's up to you.

src/cache/surface_albedo.jl
```diff
@@ -0,0 +1,204 @@
...
+Set the surface albedo using the Jin et al. (2011) empirical formulation.
+
+!!! note
+    For now we assume that the cloud fraction is 0.0.
+"""
+function surface_albedo_diffuse(
+    α_model::RegressionFunctionAlbedo{FT},
+) where {FT}
+    α_diff =
+        (λ, θ, u) -> begin
```
szy21, line 182:
```suggestion
        (λ, cosθ, u) -> begin
```
It doesn't matter here, but just to be consistent with how you call it.

src/cache/surface_albedo.jl
```diff
@@ -0,0 +1,204 @@
...
+            could_fraction = 0.0 # TODO: connect this to the EDMF
```
szy21, line 183:
```suggestion
            could_fraction = 0 # TODO: connect this to the EDMF
```
0.0 will be Float64? I'm not sure, @Sbozzolo?

test/surface_albedo.jl
```diff
@@ -0,0 +1,217 @@
...
+            0.014,
+            0.023,
+            0.044,
+            0.093,
+            0.193,
+            0.321,
+        ],
+    ]
+
+    n_list = FT.([1.34, 1.2])
```
szy21, line 136: If the values correspond to n=1.34, should we only test n=1.34 here?

szy21 (approved):
Thank you for doing this and for the documentation! I only have some minor comments.

src/cache/surface_albedo.jl
```diff
@@ -0,0 +1,204 @@
...
+    int_local_geometry_values =
+        Fields.field_values(Fields.level(Fields.local_geometry_field(Y.c), 1))
+
+    wind_speed = p.scratch.ᶠtemp_field_level
+    wind_speed_values = Fields.field_values(wind_speed)
+    @. wind_speed_values =
+        Geometry._norm.(
+            CT12.(u_int, int_local_geometry_values),
+            int_local_geometry_values,
+        )
```
szy21, line 91: Would `norm(Fields.level(Y.c.uₕ, 1))` work? like [here](https://github.com/CliMA/ClimaAtmos.jl/blob/b9e5d1af9c909af361b14b948311a47688798f59/src/cache/precomputed_quantities.jl#L556)

szy21:
Would you like to add a job using this to make sure it works fine? It will lead to an unrealistic energy balance for aquaplanet, but should be ok for testing.

docs/src/surface_albedo.md
```diff
@@ -0,0 +1,80 @@
...
+α_{diff}(λ, μ, u) = -0.1479 + 0.1502n - 0.016nσ(u)
+```
+
+for cloudy sky.
+
+
+## Comparison with Jin et al. (2011)
+
+- direct albedo (compare with Fig. 2)
+![](assets/direct_albedo_fig2.png)
```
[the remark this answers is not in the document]
LenkaNovak: ahh .gitignore!

src/cache/surface_albedo.jl
LenkaNovak: No that makes sense, thanks!

src/cache/surface_albedo.jl
LenkaNovak: yes, this is safer, thanks!

test/surface_albedo.jl
LenkaNovak: This is testing the values used in the paper.

src/cache/surface_albedo.jl
LenkaNovak: Ahh, that's the norm I was looking for! :D Still, I think we'll want to extrapolate to the surface as mentioned above, but this should do for now.

test/surface_albedo.jl
szy21: Oh, I see! I was misled by the above comments "# these values were visually confirmed by comparing `α_dir_list` generated using this script with Figure 2 from Jin et al 2011 for n = 1.34". Should it be for "n=1.34 and n=1.20"?

LenkaNovak:
> Would you like to add a job using this to make sure it works fine? It will lead to an unrealistic energy balance for aquaplanet, but should be ok for testing.

I've already applied it to the slab ocean long run and it runs fine. If you'd like to add it to the regular CI too, we can do that too, but I'm not sure it's necessary, since the unit test checks for the correct values and for initialization in the integrator. Up to you though. :)

config/longrun_configs/longrun_aquaplanet_rhoe_equil_55km_nz63_clearsky_tvinsol_0M_slabocean.yml
```diff
@@ -16,6 +16,7 @@ approximate_linear_solve_iters: 2
 rad: "clearsky"
 idealized_insolation: false
 dt_rad: "1hours"
+albedo_model: "ConstantAlbedo"
```
szy21, line 19:
```suggestion
albedo_model: "RegressionFunctionAlbedo"
```
(Do you mean this?)

szy21:
> > Would you like to add a job using this to make sure it works fine? It will lead to an unrealistic energy balance for aquaplanet, but should be ok for testing.
> 
> I've already applied it to the slab ocean long run and it runs fine. If you'd like to add it to the regular CI too, we can do that too, but I'm not sure it's necessary, since the unit test checks for the correct values and for initialization in the integrator. Up to you though. :)

Longrun is good enough I think!

test/surface_albedo.jl
szy21: Do you get exactly equal if you add FT before 0.38?

szy21: I tried this:
```
julia> a = Float32(0.38)
0.38f0

julia> a == 0.38
false

julia> a == Float32(0.38)
true
```

LenkaNovak: Ahh yeah, well spotted. This works too!

charleskawczynski (commented):
@LenkaNovak, can we remove the .png files, and instead generate them in the docs? This way, we don't check them into the main branch of the repo (which can slow down download/git-clone times).

LenkaNovak:
> @LenkaNovak, can we remove the .png files, and instead generate them in the docs? This way, we don't check them into the main branch of the repo (which can slow down download/git-clone times).

Sure, that's a neater solution. But this will add a Plots dependency to `docs/Project.jl`. Is that ok?

charleskawczynski:
> But this will add a Plots dependency to docs/Project.jl. Is that ok?

Yeah, I think that's fine. Actually, adding Makie would be better if that can work

Sbozzolo:
Yes, there's only a handful of places where Plots is used in ClimaAtmos. It would be got to not add more and use Makie instead.

PR #2679, opened 2024-02-14
Hoist unit basis vector, avoid identity conversion

charleskawczynski:


charleskawczynski:
This isn’t a notable improvement

PR #2684, opened 2024-02-14
Make cloud diagnostics job run on gpu

charleskawczynski:


szy21:
We have a cloud diagnostics job on gpu now.

PR #2688, opened 2024-02-15
Update dependencies

nefrathenrici:
**Purpose**: Update to minor releases for ClimaParameters, Thermodynamics, RRTMGP, and CloudMicrophysics 

This PR:
- widens the SurfaceFluxes compat from 0.9.1 to 0.9
- updates interface changes for RRTMGP and CloudMicrophysics.
- enables strict parameter logging - all overridden parameters need to be used to prevent silent failures.


charleskawczynski (approved):

test/Project.toml
```diff
@@ -72,7 +72,6 @@ ClimaCorePlots = "0.2"
 ClimaCoreSpectra = "0.1"
 ClimaCoreTempestRemap = "0.3"
 ClimaCoreVTK = "0.7"
-CloudMicrophysics = "0.15.2"
```
charleskawczynski, line 75: Should we add `0.15` back in?

nefrathenrici: Sure, but 0.15 uses ClimaParameters 0.8 so I don't think it would get used

PR #2690, opened 2024-02-15
Adapt interpolation object for GPU runs. 

akshaysridhar:


src/solver/type_getters.jl
```diff
@@ -135,17 +136,21 @@ function get_spaces(parsed_args, params, comms_ctx)
         warp_function = nothing
     elseif topography == "Earth"
         data_path = joinpath(topo_elev_dataset_path(), "ETOPO1_coarse.nc")
+        array_type = ClimaComms.array_type(ClimaComms.device())
```
charleskawczynski, line 139: Why are we getting the device this way? Shouldn't we already have it somewhere?

akshaysridhar: Something like `ctx = ClimaComms.context(axes(Y.c))` to then get `ctx.device` perhaps based on this ```ClimaComms.context(device=device())``` in ClimaComms perhaps ? ; I'll check if we have this stored somewhere already in`model` or `type` getters.

akshaysridhar: @charleskawczynski Fixed (device is accessible in ctx) and is available at this level without calling the ClimaComms `device` function

akshaysridhar:
@charleskawczynski If the compat entry update + re-built Manifests look reasonable I can merge this.

Project.toml
```diff
@@ -42,6 +43,7 @@ Thermodynamics = "b60c26fb-14c3-4610-9d3e-2d17fe7ff00c"
 YAML = "ddb6d928-2868-570f-bddf-ab3f9cf99eb6"
 
 [compat]
+Adapt = "3.7"
```
Sbozzolo, line 46: Why this version?

Would this work?
```suggestion
Adapt = "3, 4"
```

charleskawczynski: This won't do anything yet because RRTMGP still restricts us

charleskawczynski (approved):

Project.toml
akshaysridhar: Yeah my intention was to pin it to a known working version, here, `3.7` (given the issues we seemed to be having elsewhere with CUDA / Adapt upgrades)

perf/Project.toml
```diff
@@ -1,4 +1,5 @@
 [deps]
+Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
```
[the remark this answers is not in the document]
Sbozzolo: But this is the `perf` env. `Adapt` is never used directly here.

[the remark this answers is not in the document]
Sbozzolo: Ops, got it.

PR #2700, opened 2024-02-20
Try ClimaCore specialize on fieldvector

charleskawczynski:
This PR tries out https://github.com/CliMA/ClimaCore.jl/pull/1615

### DIFF
--- a/docs/Manifest.toml
+++ b/docs/Manifest.toml
@@ -2,7 +2,7 @@
 
 julia_version = "1.10.0"
 manifest_format = "2.0"
-project_hash = "1b7b7b12f210f5da35bb4f69ad563a64237fde09"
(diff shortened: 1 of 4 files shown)


charleskawczynski:
Nice, in the gpu moist held suarez:

 - Main: `[ Info: Memory currently used (after solve!) by the process (RSS): 5.65 GiB`
 - This branch: `[ Info: Memory currently used (after solve!) by the process (RSS): 4.55 GiB`

The runtime/compilation time doesn't seem largely impacted, so that's good.

cc @Sbozzolo, @dennisYatunin

Sbozzolo:
Looks great! The only job that seems to take noticeably longer is the perf aquaplanet. Let's keep an eye on that, maybe it's just a fluctuation.

PR #2713, opened 2024-02-23
overwrite duplicate diagnostics

### LINKED ISSUE #2709: re-running add_diagnostic_variable! produces error (in coupler)
In Atmos's diagnostics, trying to add a diagnostic that's already been added results in an error (in [this line](https://github.com/CliMA/ClimaAtmos.jl/blob/main/src/diagnostics/diagnostic.jl#L155-L156)). When debugging in the coupler, we often re-run our driver from the same julia session (using Revise for changes). However, we've recently added some calls to `add_diagnostic_variable!` in our driver that are causing our runs to fail when we re-run in a session that's already added the diagnostics

(this isn't really a bug, but it is a little annoying and has an easy fix :) )

### Proposed solution
Do nothing if a variable is already in the diagnostics, instead of failing with an error

### Example error message
(in ClimaCoupler.jl experiments/AMIP, after running `include("coupler_driver.jl")` in the same julia session)

> julia> include("coupler_driver.jl")
> ERROR: LoadError: diagnostic mse already defined
> Stacktrace:
>  [1] error(s::String)
>    @ Base ./error.jl:35
>  [2] #add_diagnostic_variable!#3
>    @ ClimaAtmos.Diagnostics ~/.julia/packages/ClimaAtmos/9h5KM/src/diagnostics/diagnostic.jl:155 [inlined]
>  [3] top-level scope
>    @ ~/ClimaCoupler.jl/experiments/AMIP/components/atmosphere/climaatmos_extra_diags.jl:16
>  [4] include(fname::String)
>    @ Base.MainInclude ./client.jl:489
>  [5] top-level scope
>    @ ~/ClimaCoupler.jl/experiments/AMIP/components/atmosphere/climaatmos_init.jl:17
>  [6] include(fname::String)
>    @ Base.MainInclude ./client.jl:489
>  [7] top-level scope
>    @ ~/ClimaCoupler.jl/experiments/AMIP/coupler_driver.jl:126
>  [8] include(fname::String)
>    @ Base.MainInclude ./client.jl:489
>  [9] top-level scope
>    @ REPL[1]:1
> in expression starting at /Users/jsloan/ClimaCoupler.jl/experiments/AMIP/components/atmosphere/climaatmos_extra_diags.jl:16
> in expression starting at /Users/jsloan/ClimaCoupler.jl/experiments/AMIP/components/atmosphere/climaatmos_init.jl:17
> in expression starting at /Users/jsloan/ClimaCoupler.jl/experiments/AMIP/coupler_driver.jl:126

juliasloan25:
## Purpose 
Instead of raising an error when a user tries to add a diagnostic that already exists, this PR warns the user and overwrites the existing diagnostic with the new entry.

closes #2709

### DIFF
--- a/src/diagnostics/diagnostic.jl
+++ b/src/diagnostics/diagnostic.jl
@@ -152,8 +152,14 @@ function add_diagnostic_variable!(;
     comments = "",
     compute!,
 )
-    haskey(ALL_DIAGNOSTICS, short_name) &&
-        error("diagnostic $short_name already defined")
+    haskey(ALL_DIAGNOSTICS, short_name) && @warn(
+        "overwriting diagnostic `$short_name` entry containing fields\n" *
+        "$(map(
+            field -> "$(getfield(ALL_DIAGNOSTICS[short_name], field))",
+            filter(field -> field != :compute!, fieldnames(DiagnosticVariable)),
+        ))"
+    )
+
 
     ALL_DIAGNOSTICS[short_name] = DiagnosticVariable(;
         short_name,


Sbozzolo:
Maybe print out more information on what was there before overwriting?

juliasloan25:
> Maybe print out more information on what was there before overwriting?

Do you want me to also show what it's getting overwritten with?

Sbozzolo:
> > Maybe print out more information on what was there before overwriting?
> 
> Do you want me to also show what it's getting overwritten with?

No, we assume that users will know what they called.

juliasloan25:
Okay! I updated the warning message so it will print something like 
```
┌ Warning: overwriting diagnostic `mse` entry containing fields
│ [mse, Moist static energy, moist_static_energy, J/kg, Moist static energy]
└ @ ClimaAtmos.Diagnostics ~/.julia/packages/ClimaAtmos/gIAbX/src/diagnostics/diagnostic.jl:155
```

src/diagnostics/diagnostic.jl
```diff
@@ -152,8 +152,15 @@ function add_diagnostic_variable!(;
...
-    haskey(ALL_DIAGNOSTICS, short_name) &&
-        error("diagnostic $short_name already defined")
+    haskey(ALL_DIAGNOSTICS, short_name) && @warn(
+        "overwriting diagnostic `$short_name` entry containing fields\n" *
+        "[$(ALL_DIAGNOSTICS[short_name].short_name), " *
+        "$(ALL_DIAGNOSTICS[short_name].long_name), " *
+        "$(ALL_DIAGNOSTICS[short_name].standard_name), " *
+        "$(ALL_DIAGNOSTICS[short_name].units), " *
+        "$(ALL_DIAGNOSTICS[short_name].comments)]"
+    )
```
Sbozzolo, line 162: There ought to be a cleaner way to write this. You could use `fieldnames`.

src/diagnostics/diagnostic.jl
```diff
@@ -152,8 +152,14 @@ function add_diagnostic_variable!(;
     comments = "",
     compute!,
 )
-    haskey(ALL_DIAGNOSTICS, short_name) &&
-        error("diagnostic $short_name already defined")
+    haskey(ALL_DIAGNOSTICS, short_name) && @warn(
+        "overwriting diagnostic `$short_name` entry containing fields\n" *
+        "$(map(
+            fieldname -> "$(getfield(ALL_DIAGNOSTICS[short_name], fieldname))",
+            fieldnames(DiagnosticVariable)[1:(end - 1)],
```
Sbozzolo, line 159: I'd say that we can leave the function (so that we remove the [1:(end-1)]. It will probably print the function name, which is not too bad.

juliasloan25: I changed it to use filter to exclude the function, and now the output looks like this from the coupler:
```
┌ Warning: overwriting diagnostic `mse` entry containing fields
│ ("mse", "Moist static energy", "moist_static_energy", "J/kg", "Moist static energy")
└ @ ClimaAtmos.Diagnostics ~/.julia/packages/ClimaAtmos/DeXjA/src/diagnostics/diagnostic.jl:155
```

Sbozzolo (approved):

PR #2716, opened 2024-02-23
add a simple edmf test case

### LINKED ISSUE #2706: Add a simple edmf test case with analytical solution

szy21:
## Purpose 

Adds a simple plume test case for edmf. This includes adding a flag for turning off grid-scale tendencies. It would be useful for testing the numerics for edmf.

Closes #2706 

## Content

----

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -544,6 +544,14 @@ steps:
         agents:
           slurm_mem: 20GB
 
+      - label: ":genie: Prognostic EDMFX simple plume test in a column"
+        command: >
+          julia --color=yes --project=examples examples/hybrid/driver.jl
+          --config_file $CONFIG_PATH/prognostic_edmfx_simpleplume_column.yml
+        artifact_paths: "prognostic_edmfx_simpleplume_column/*"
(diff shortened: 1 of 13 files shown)


trontrytel (approved):
LGTM! 

I wish there was a way to add those simple tests like the `advection_test` we did before and now the simple plume, without having to add more options to the `AtmosModel`. This pattern has the risk of becoming very cluttered if we continue to need to switch off different things. Maybe if we need to add another one like this we could try to have some sort of umbrella `debugging` flag and have sub-options there.

szy21:
> LGTM!
> 
> I wish there was a way to add those simple tests like the `advection_test` we did before and now the simple plume, without having to add more options to the `AtmosModel`. This pattern has the risk of becoming very cluttered if we continue to need to switch off different things. Maybe if we need to add another one like this we could try to have some sort of umbrella `debugging` flag and have sub-options there.

Yes, I thought about that. I didn't combine them in the end because `advection_test` turns off edmf tendencies, while `gs_tendency` only controls grid-scale tendencies. But maybe we can have a matrix that controls both tendencies.

PR #2722, opened 2024-02-26
Try ClimaCore diag specialization

charleskawczynski:
This PR tries out a branch of ClimaCore that attempts to specialize on "diagonal" broadcast expressions with fields. 

xref: https://github.com/CliMA/ClimaCore.jl/issues/1128, https://github.com/CliMA/ClimaCore.jl/issues/1403.


charleskawczynski:
We're no longer pursuing this this (focusing on MultiBroadcastFusion optimization instead).

PR #2725, opened 2024-02-27
Set Insolation compat to 0.9.1

nefrathenrici:
0.9.2 is incompatible with Atmos since it uses the new ClimaParams instead of CLIMAParameters. 
Until I get all of the other dependencies updated and use ClimaParams, we need to narrow the compat.


nefrathenrici:
Superseded by #2703

PR #2727, opened 2024-02-27
remove some gpu jobs from ci

szy21:
## Purpose 

Removes some gpu jobs:

- `gpu_held_suarez_rhoe_hightop` and the comparison as we have a comparison for baroclinic wave. Actually, the comparison has been failing, see the broken test in the most recent [build](https://buildkite.com/clima/climaatmos-ci/builds/16952#018de942-a672-4d31-88f7-b1d4186fe1ea/165-233). Is it useful to keep the comparison?
- `gpu_implicit_barowave_wrt_h_elem`: it's the same as `gpu_implicit_barowave` except for `dt_save_state_to_disk`

These jobs don't seem very useful to me. @charleskawczynski @sriharshakandala What do you think? As a side note, the resolutions for the GPU performance jobs are `z_elem: 25 h_elem: 12`. Where do we get the resolution from?

## Content

----

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -638,29 +638,6 @@ steps:
           - "sphere_baroclinic_wave_rhoe"
           - "gpu_baroclinic_wave_rhoe"
 
-      - label: "GPU: HS (ρe) hightop"
-        key: "gpu_held_suarez_rhoe_hightop"
-        command: >
-          julia --color=yes --project=examples examples/hybrid/driver.jl
-          --config_file $PERF_CONFIG_PATH/gpu_held_suarez_rhoe_hightop.yml
-        artifact_paths: "gpu_held_suarez_rhoe_hightop/*"
-        agents:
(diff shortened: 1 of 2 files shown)


charleskawczynski:
`gpu_implicit_barowave_wrt_h_elem` is what @bloops has been using for analyzing performance. I think we would like to keep that running

szy21:
Ah ok, I'll keep that. How about the comparison with CPUs? All the comparisons seem broken (which is fine I think, as we shouldn't expect the same results).

charleskawczynski:
> How about the comparison with CPUs? All the comparisons seem broken (which is fine I think, as we shouldn't expect the same results).

I guess that's fine 🤷🏻

szy21:
I guess the relative error can be useful. I'll keep one BW comparison and remove the HS one.

charleskawczynski (approved):

sriharshakandala (approved):

PR #2730, opened 2024-02-28
Add an interface for debugging the implicit solver Jacobian

dennisYatunin:
## Purpose 

This PR splits the `ImplicitEquationJacobian` into an `ApproxJacobian`, an `ExactJacobian`, and a `DebugJacobian`. The first option is the Jacobian approximation we've been using until now, while the second option is an exact Jacobian computed using `ForwardDiff` and inverted using column-wise LU decomposition. The third option can act like either of the other two, and it checks the approximation against the exact computation whenever the exact Jacobian is updated.

This PR also gets rid of some outdated code related to `OrdinaryDiffEq` in order to simplify the interface design.

@coderabbitai ignore

## To-do

Add some working examples to CI.

## Content

----


Project.toml
```diff
@@ -41,6 +43,7 @@ ArgParse = "1"
 ArtifactWrappers = "0.2"
 Artifacts = "1"
 AtmosphericProfilesLibrary = "0.1.7"
+BlockArrays = "1"
```
Sbozzolo, line 46: Can we make the new dependencies `weakdeps` and implement them as extensions?

Sbozzolo:
If would be nice to also add some user documentation that explains what is the point of this feature, when people should use it, and so on.

Project.toml
dennisYatunin: The BlockArrays dependency will be removed once some things are moved to ClimaCore. The ForwardDiff dependency cannot be an extension; it is required for autodiff, which we will be using for EDMF simulations for the foreseeable future.

Sbozzolo: Being used for the foreseeable future and being and extensions are not conflicting. 

You can also make it an extension that is automatically loaded all the time. This will give us a path to make it a normal extension in the future

dennisYatunin: I am extremely confused by what you are trying to achieve here. The point of this PR is to add two flags to our configuration: `use_exact_jacobian` and `debug_approximate_jacobian`. Setting either of these flags to `true` means that `ForwardDiff.jl` needs to be loaded. These flags will be used extensively in the development of implicit EDMF (and all other aspects of the implicit solver) moving forward.

As far as I am aware, the optional loading of packages based on YAML configurations is not the point of using extensions. Moreover, `ForwardDiff.jl` is already a dependency of `ClimaCore.jl`, `CloudMicrophysics.jl`, and `RootSolvers.jl`, so making it an extension will have no impact whatsoever on compilation time. So, I do not see any justification for making this an extension.

Sbozzolo: What I am trying to achieve is isolating dependencies and providing pathways to making them optional or removing them. As you think of a target ClimaAtmos, is this feature a core feature, or is this an optional one? If it not a core feature, the target ClimaAtmos should not have a hard-dependency on `ForwardDiff`.

The fact that `ForwardDiff` is already used by three of our packages does not mean that we should not try to keep our footprint as small as reasonably possible or that we won't remove the dependency in the future from those three packages. 

In this case, turning it into an extension is trivial: `ForwardDiff` is only used in two functions, so you can just move the two functions into a file in `ext/` and put functions stubs in their place.

That said, I agree that `ForwardDiff` is currently used elsewhere (and this is unlikely to change any time soon) and turning into an extension might too much at this point. That said, I think we should have a mindset where optional functionalities should add no costs to packages. In this case, the cost is an extra direct dependency. I would like us to be mindful of this because never thinking about this is how we got environments with 500 packages that take 30 minutes to instantiate.

dennisYatunin: Sorry if I wasn't clear earlier. Moving forward, this will be a core feature of ClimaAtmos. Similarly, the ability to solve a nonlinear equation using autodiff is a core feature of RootSolvers. I think this means both packages will continue to have ForwardDiff as a dependency.

src/cache/temporary_quantities.jl
```diff
@@ -68,5 +69,6 @@ function temporary_quantities(Y, atmos)
         ᶜdiffusion_h_matrix = similar(Y.c, TridiagonalMatrixRow{FT}),
         ᶜdiffusion_u_matrix = similar(Y.c, TridiagonalMatrixRow{FT}),
         ᶠtridiagonal_matrix_c3 = similar(Y.f, TridiagonalMatrixRow{C3{FT}}),
+        temp_matrix = Array{FT}(undef, num_maxtrix_rows, num_maxtrix_rows),
```
charleskawczynski, line 72: I don't see any other references to this variable

src/cache/precomputed_quantities.jl
```diff
@@ -454,15 +468,19 @@ function instead of recomputing the value yourself. Otherwise, it will be
 difficult to ensure that the duplicated computations are consistent.
 """
 NVTX.@annotate function set_precomputed_quantities!(Y, p, t)
+    set_implicit_precomputed_quantities!(Y, p, t)
+    set_explicit_precomputed_quantities!(Y, p, t)
+end
+function set_implicit_precomputed_quantities!(Y, p, t)
```
charleskawczynski, line 474: Can we split this off into a separate PR that is non-behavior changing?

src/diagnostics/core_diagnostics.jl
```diff
@@ -1163,3 +1163,154 @@ add_diagnostic_variable!(
     comments = "Mass of water vapor per mass of air",
     compute! = compute_husv!,
 )
+
+###
+# Implicit solver
+###
+function compute_ejac1!(out, state, cache, time)
```
charleskawczynski, line 1170: Please add docs for this.

src/diagnostics/core_diagnostics.jl
```diff
@@ -1163,3 +1163,154 @@ add_diagnostic_variable!(
...
+add_diagnostic_variable!(
+    short_name = "ejac1",
+    long_name = "Exact Jacobian matrix of first column",
+    standard_name = "exact_jacobian",
+    units = "",
+    comments = "Exact Jacobian matrix of tendency in first column",
+    compute! = compute_ejac1!,
+)
+
+tensor_axes_tuple(type) =
```
charleskawczynski, line 1192:
```suggestion
tensor_axes_tuple(::Type{type}) where {type} =
```

src/diagnostics/core_diagnostics.jl
```diff
@@ -1163,3 +1163,154 @@ add_diagnostic_variable!(
...
+    else
+        sub_names = fieldnames(typeof(value))
+        sub_values =
+            MatrixFields.unrolled_map(Base.Fix1(getfield, value), sub_names)
+        nonempty_sub_values =
+            MatrixFields.unrolled_filter(x -> sizeof(x) > 0, sub_values)
+        @assert length(nonempty_sub_values) == 1
+        primitive_value_at_index(nonempty_sub_values[1], (row_axes, col_axes))
+    end
+@static if hasfield(Method, :recursion_relation)
```
charleskawczynski, line 1228:
```suggestion

@static if hasfield(Method, :recursion_relation)
```
So that it's clear that this is separate from the above function.

src/cache/precomputed_quantities.jl
```diff
@@ -137,9 +136,9 @@ function precomputed_quantities(Y, atmos)
             ᶜdetrʲs = similar(Y.c, NTuple{n, FT}),
             ᶜturb_entrʲs = similar(Y.c, NTuple{n, FT}),
             ᶠnh_pressure₃ʲs = similar(Y.f, NTuple{n, C3{FT}}),
-            ᶜgradᵥ_θ_virt⁰ = Fields.Field(C3{FT}, cspace),
-            ᶜgradᵥ_q_tot⁰ = Fields.Field(C3{FT}, cspace),
-            ᶜgradᵥ_θ_liq_ice⁰ = Fields.Field(C3{FT}, cspace),
+            ᶜgradᵥ_θ_virt⁰ = similar(Y.c, C3{FT}),
+            ᶜgradᵥ_q_tot⁰ = similar(Y.c, C3{FT}),
+            ᶜgradᵥ_θ_liq_ice⁰ = similar(Y.c, C3{FT}),
```
charleskawczynski, line 141: Any particular reason that we're using `similar` here over `Field`? I kind of prefer `Field` over `similar`, since those constructors don't actually need the field values in `Y.c`. (I was thinking that we could refactor in the other direction).

src/prognostic_equations/gm_sgs_closures.jl
```diff
@@ -19,7 +19,7 @@ function smagorinsky_lilly_length(c_smag, N_eff, dz, Pr, ϵ_st)
     return N_eff > FT(0) ?
            c_smag *
            dz *
-           max(0, 1 - N_eff^2 / Pr / 2 / max(ϵ_st, eps(FT)))^(1 / 4) :
+           max(FT(0), 1 - N_eff^2 / Pr / 2 / max(ϵ_st, eps(FT)))^(1 / FT(4)) :
```
charleskawczynski, line 22: `max` does type promotions, so this is actually not needed.

src/prognostic_equations/implicit/approx_jacobian.jl
```diff
@@ -0,0 +1,879 @@
...
+    sgs_advection_flag = if isnothing(alg.sgs_advection_flag)
+        atmos.sgs_adv_mode == Implicit() ? UseDerivative() : IgnoreDerivative()
+    else
+        alg.sgs_advection_flag
+    end
+    sgs_mass_flux_flag = if isnothing(alg.sgs_mass_flux_flag)
+        atmos.sgs_adv_mode == Implicit() ? UseDerivative() : IgnoreDerivative()
+    else
+        alg.sgs_mass_flux_flag
+    end
```
charleskawczynski, line 66:
```suggestion
    treat_∂(::Nothing, mode::Bool) = mode ? UseDerivative() : IgnoreDerivative()
    treat_∂(::Nothing, ::Implicit) = UseDerivative()
    treat_∂(::Nothing, ::Explicit) = IgnoreDerivative()
    treat_∂(flag, _) = flag

    topography_flag = treat_∂(alg.topography_flag, has_topography(axes(Y.c)))
    diffusion_flag = treat_∂(alg.diffusion_flag, atmos.diff_mode)
    sgs_advection_flag = treat_∂(alg.sgs_advection_flag, atmos.sgs_adv_mode)
    sgs_mass_flux_flag = treat_∂(alg.sgs_mass_flux_flag, atmos.sgs_adv_mode)
```
We could write a couple helpers here, if you think it's worth it.

src/cache/precomputed_quantities.jl
dennisYatunin: I'm ambivalent about using `Field` or `similar`. I was just enforcing consistency with this change, but we can hold off on that until a future PR.

src/prognostic_equations/gm_sgs_closures.jl
dennisYatunin: Yes, the `FT(0)` is not needed here, and I've removed it. However, the `FT(4)` is needed, since `1 / 4` is always a `Float64`.

src/callbacks/get_callbacks.jl
```diff
@@ -175,20 +175,64 @@ function get_diagnostics(parsed_args, atmos_model, Y, p, sim_info, output_dir)
...
             )...,
             diagnostics...,
         ]
     end
-    diagnostics = collect(diagnostics)
+    if parsed_args["output_default_diagnostics"] && (
+        parsed_args["use_exact_jacobian"] ||
+        parsed_args["debug_approximate_jacobian"]
+    )
+        t_span_seconds = dt isa ITime ? seconds(t_span) : t_span
```
Sbozzolo, line 197: You can write 
```suggestion
        t_span_seconds = float(t_span)
```
this will work for both Float and ITime

post_processing/jacobian_plots.jl
```diff
@@ -0,0 +1,357 @@
...
+        Yₜ = data_at_t["Yₜ"][]
+
+        safe_inv(x) = iszero(x) ? zero(x) : inv(x)
+        rescaled_∂Yₜ_∂Y = safe_inv.(Yₜ) .* ∂Yₜ_∂Y .* Yₜ'
+
+        # Take the transpose of each matrix so that its rows are plotted along
+        # the y-axis and its columns are plotted along the x-axis.
+        (parse(Float64, t_string), title, ∂Yₜ_∂Y', rescaled_∂Yₜ_∂Y', Yₜ)
+    end
+    Base.close(reader)
```
charleskawczynski, line 66: I'd recommend using the `do`-syntax, as that will do a `try-catch-finally`, and close the file in the case that something in here breaks, preventing memory leaks.

post_processing/jacobian_plots.jl
```diff
@@ -0,0 +1,357 @@
...
+    return time_series
+end
+
+function plot_jacobian(output_path, time_series, file_name, Yₜ_end, dt)
```
charleskawczynski, line 70: This function looks nearly atmos-agnostic. Should it be a ClimaCore feature?

dennisYatunin: This will be refactored to directly plot a FieldMatrix and added to ClimaCore by Teja. The current array plotting is very clunky and will be removed at the earliest opportunity.

.buildkite/pipeline.yml
```diff
@@ -31,6 +31,10 @@ steps:
 
       - echo "--- Instantiate .buildkite"
       - "julia --project=.buildkite -e 'using Pkg; Pkg.instantiate(;verbose=true); Pkg.precompile(;strict=true); using CUDA; CUDA.precompile_runtime(); Pkg.status()'"
+      - "julia --project=.buildkite -e 'using Pkg; Pkg.add(Pkg.PackageSpec(;name=\"ClimaCore\", rev=\"tr/fieldmatrix-functions-for-jacobian\"))'"
+      - "julia --project=.buildkite -e 'using Pkg; Pkg.add(Pkg.PackageSpec(;name=\"ClimaComms\", rev=\"main\"))'"
+      - "julia --project=.buildkite -e 'using Pkg; Pkg.add(Pkg.PackageSpec(;name=\"Thermodynamics\", rev=\"dy/more_autodiff_fixes\"))'"
```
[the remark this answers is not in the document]
charleskawczynski: Yep, the branch has been merged, I used the `main` branch.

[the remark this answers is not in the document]
dennisYatunin: Just rebased; it refers to the main branch now.

charleskawczynski:
Local notes (to run some slow experiments)

```julia
# dy/jacobian_debugging

julia +1.11 --project=.buildkite -e 'using Pkg; Pkg.add(Pkg.PackageSpec(;name="ClimaCore", rev="tr/fieldmatrix-functions-for-jacobian"))'
julia +1.11 --project=.buildkite -e 'using Pkg; Pkg.add(Pkg.PackageSpec(;name="ClimaComms", rev="main"))'
julia +1.11 --project=.buildkite -e 'using Pkg; Pkg.add(Pkg.PackageSpec(;name="Thermodynamics", rev="main"))'
julia +1.11 --project=.buildkite -e 'using Pkg; Pkg.add(Pkg.PackageSpec(;name="CloudMicrophysics", rev="main"))'

julia +1.11 --project=.buildkite .buildkite/ci_driver.jl --config_file config/model_configs/prognostic_edmfx_aquaplanet.yml --job_id prognostic_edmfx_aquaplanet

nsys profile --trace=nvtx,cuda --output=prognostic_edmfx_aquaplanet.nsys-rep julia +1.11 --project=.buildkite .buildkite/ci_driver.jl --config_file config/model_configs/prognostic_edmfx_aquaplanet.yml --job_id prognostic_edmfx_aquaplanet
scp -r clima:/home/charliek/CliMA/ClimaAtmos.jl/prognostic_edmfx_aquaplanet.nsys-rep ./

nsys profile --trace=nvtx,cuda --output=plane_schar_mountain_float32_test.nsys-rep julia +1.11 --project=.buildkite .buildkite/ci_driver.jl --config_file config/model_configs/plane_schar_mountain_float32_test.yml --job_id plane_schar_mountain_float32_test
scp -r clima:/home/charliek/CliMA/ClimaAtmos.jl/plane_schar_mountain_float32_test.nsys-rep ./
```

src/prognostic_equations/implicit/auto_dense_jacobian.jl
```diff
@@ -0,0 +1,204 @@
...
+    FT_dual =
+        ForwardDiff.Dual{AutoDenseJacobian, FT, exact_jacobian_batch_size()}
+    Y_dual = replace_parent_type(Y, FT_dual)
+    Yₜ_dual = similar(Y_dual)
+    precomputed_dual =
+        replace_parent_type(implicit_precomputed_quantities(Y, atmos), FT_dual)
+    scratch_dual = replace_parent_type(temporary_quantities(Y, atmos), FT_dual)
+
+    n_columns = Fields.ncolumns(Y.c)
+    n_εs = length(first_column(Y))
```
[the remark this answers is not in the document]
dennisYatunin: The length of a FieldVector is the number of scalar values it contains.

PR #2737, opened 2024-02-28
Remove types from TOML overrides

nefrathenrici:
Types are no longer needed if the parameter exists in the default file, simplifying the syntax slightly


szy21 (approved):
Thanks!

PR #2740, opened 2024-02-29
Tests for ClimaAtmos utilities

akshaysridhar:
This PR adds unit tests for ClimaAtmos `utilities` functions.


sriharshakandala (approved):

sriharshakandala:
We can move some of the boiler plate code out as we add more tests, but we can keep this for now!

PR #2751, opened 2024-03-04
Update scaling pipeline, remove stray line from CI pipeline.yml

nefrathenrici:
The CPU scaling pipeline has broken with the move to the new partition. 
This PR moves to the new partition, runs only on icelake nodes and has core counts set properly for icelake. 

CPU Scaling efficiency is affected dramatically, going from ~70% on 128 cores to ~90 on 112 cores!

Passing scaling [build](https://buildkite.com/clima/climaatmos-scaling/builds/577)

### DIFF
--- a/.buildkite/scaling/pipeline.sh
+++ b/.buildkite/scaling/pipeline.sh
@@ -3,7 +3,8 @@ set -euo pipefail
 
 FT="Float32"
 resolutions=("low" "mid" "high")
-process_counts=("1 2 4 8 16 32" "1 2 4 8 16 32 64" "1 2 4 8 16 32 64 128")
+# Process counts for icelake nodes
+process_counts=("1 2 4 7 14 28" "1 2 4 7 14 28 56" "1 2 4 7 14 28 56 112")
 max_procs_per_node=16 # limit this artificially for profiling
 profiling=disable
 exclusive=true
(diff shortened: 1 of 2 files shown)


Sbozzolo:
The GPU pipeline runs on a place without climacommon.

Have you tested the scaling pipeline with this PR?

nefrathenrici:
> The GPU pipeline runs on a place without climacommon.
> 
> Have you tested the scaling pipeline with this PR?

Tests are failing [here](https://buildkite.com/clima/climaatmos-scaling/builds/564#018e0aaf-2ef3-41fa-aa0a-ead9960c6e11). Do you know why it is unable to find `climacommon`? 
Setting `export MODULEPATH=/groups/esm/modules:$MODULEPATH` doesn't work.

Sbozzolo:
The version number seems to be mispelled

nefrathenrici:
Seems to be working now - see [build here](https://buildkite.com/clima/climaatmos-scaling/builds/566#018e0acc-d13b-42f6-98e5-951d848c3702)

Sbozzolo:
I think that we should force the node type to be the same, otherwise this pipeline might run on nodes with different architectures, giving results that are not really meaningful.

nefrathenrici:
> I think that we should force the node type to be the same, otherwise this pipeline might run on nodes with different architectures, giving results that are not really meaningful.

Is there a way to do this besides specifying a single target architecture?

Sbozzolo:
Specifying a single target architecture does not ensure that the job is run
on that architecture.

You can use slurm to specify nodes based on available features, which are
used to distinguish the different node types on our cluster.
For the scaling pipeline, we might want to stick with icelake or epyc.

The relevant keyword is slurm_constraint: icelake (or epyc)

.buildkite/scaling/pipeline.sh
```diff
@@ -64,22 +64,16 @@ done
 # set up environment and agents
 cat << 'EOM'
 agents:
-  queue: central
-  modules: julia/1.10.0 cuda/12.2 ucx/1.14.1_cuda-12.2 openmpi/4.1.5_cuda-12.2 nsight-systems/2023.3.1
+  queue: new-central
+  modules: climacommon/2024_02_27
```
Sbozzolo, line 68: You can just add it here so that it holds for all your jobs in this pipeline.
```
slurm_constraint = 'icelake'
```

.buildkite/scaling/pipeline.sh
```diff
@@ -64,22 +64,17 @@ done
...
+  slurm_constraint: epyc
```
Sbozzolo, line 69: Please, add a comment to explain why we are setting a constraint

(And is there a specific reason you picked eypc over icelake?)

nefrathenrici: I tried epyc because I tried to restrict the `JULIA_CPU_TARGET` to just icelake and I got the following error:
```julia
ERROR: Your CPU does not support the CX16 instruction, which is required by this version of Julia! 
This is often due to running inside of a virtualized environment. 
Please read https://docs.julialang.org/en/v1/devdocs/sysimg/ for more.
```
I have set it back to icelake with the full cpu target.

Sbozzolo: I think epyc is better for this particular test. Eypc nodes have 64 cores, and your scaling tests are powers of 2. icelake nodes have 56 cores.

I was asking if you picked epyc for any (other) particular reason.

nefrathenrici: Ah, that is good to know, I will switch it back to epyc. I think I am misunderstanding how `JULIA_CPU_TARGET` works. 
Why do we need to supply targets besides epyc if the scaling pipeline is only running on epyc nodes?

Sbozzolo: You shouldn't have to. I think you should even be able to remove the JULIA_CPU_TARGET line.

nefrathenrici: There are only 28 epyc nodes and 462 icelake nodes and the scaling build has been waiting for 1 hour for epyc nodes, I think it makes sense to just use icelake even if the scaling tests are power of 2.

Sbozzolo: In this case, I think we should change the number of cores we are testing over to make sure that we are properly saturating the node (in fact, typical scaling tests are for number of nodes, not for number of cores).

nefrathenrici:
After discussion with @charleskawczynski, we will disable the weekly builds for now, since we aren't actually using the results.

The changes in this PR are still useful for constraining to icelake nodes and setting proper node counts, so I will merge it in

PR #2758, opened 2024-03-05
Fuse dynamic and tracer tendency dss calls

charleskawczynski:
This PR fuses the calls to `weighted_dss!` for the dynamic and tracer tendencies, eliminating 4 DSS calls per timestep.


charleskawczynski:
Looks like this is passing! 🎉

Issue #2721, opened 2024-02-26
Job has non-deterministic results due to RRTMGP


nefrathenrici:
Job ID: `sphere_aquaplanet_rhoe_equilmoist_allsky_gw_raw_zonallyasymmetric`
This is likely caused by cloud-radiation interactions in the simulation added after #2704 .

szy21:
I did some test by looking at the norm of the solution. This may not be as accurate as MSE, but it shows:
FT32 simulation with FT32 radiation: non-determinisitic
FT64 simulation with FT64 radiation: non-deterministic
FT32 simulation with FT64 radiation (which is the same as what we have before): deterministic

@dennisYatunin suggested that maybe radiation is always not deterministic, but in the FT32 simulation with FT64 radiation case, we lose the last few non-deterministic digits when converting from FT64 to FT32. That makes sense to me. I don't think the issue is related to the PR.

szy21:
I think this may have been fixed by the recent fix in RRTMGP. I run the simulation a couple of times on ci and the norm of the resolution is the same. I can add back the regression test. @charleskawczynski is there an easy way to make sure it is reproducible before adding the regression test?

charleskawczynski:
I think opting into the regression test is the simplest way

szy21:
Ok, I’ll add it in and keep an eye on the builds.

szy21:
The regression test has been added back in #2798, and the most recent two builds passed, so I think we can close this issue.

PR #2764, opened 2024-03-07
move cloud fraction callback before rrtmgp

### LINKED ISSUE #837: Non-deterministic results in all-sky long run
These two builds:

 - https://buildkite.com/clima/climaatmos-longruns/builds/402#01832def-6aa0-452e-b220-a637e972191e
 - https://buildkite.com/clima/climaatmos-longruns/builds/403#01833083-97ea-42e6-86ad-7af6f607b480

occurred one day apart and the only PRs between were #834 and #835. So, it looks like our results for this job are non-deterministic, and we should fix this. Potential culprits are:
 - Use of `rand` without (or before) `Random.seed`
 - (threading) race condition in RRTMGP, ClimaCore, or some nested combination

szy21:
## Purpose 

This is at least one of the reasons that the all-sky radiation job is non-deterministic. Part of #2721. Also increases jet nfailures. 

## Content

----

### DIFF
--- a/perf/jet_test_nfailures.jl
+++ b/perf/jet_test_nfailures.jl
@@ -38,7 +38,7 @@ using Test
     # inference. By increasing this counter, we acknowledge that
     # we have introduced an inference failure. We hope to drive
     # this number down to 0.
-    n_allowed_failures = 42
+    n_allowed_failures = 112
     @show n
     @test n ≤ n_allowed_failures
     if n < n_allowed_failures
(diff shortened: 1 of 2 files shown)


szy21:
@charleskawczynski I think this should fix the non-deterministic issue, and I checked locally that the job is reproducible by looking at the norm of the solution. But is there an easy way to make sure the regression test works before merging it?

charleskawczynski (approved):

charleskawczynski:
Good catch! I'm not sure why this would result in new JET failures, though

trontrytel (approved):
Thank you for catching that!

charleskawczynski:
Feel free to increase the number of allowed JET failures, so that we can merge this @szy21!

PR #2766, opened 2024-03-07
Do not move output outside of buildkite climaatmos-ci

Sbozzolo:
We only want to move things around on the main climaatmos-ci pipeline.


Sbozzolo:
It passes:

https://buildkite.com/clima/climaatmos-target-gpu-simulations/builds/217

charleskawczynski (approved):

PR #2771, opened 2024-03-08
Interface tests: ClimaAtmos spaces, callbacks and parameterized tendency construction

akshaysridhar:
Adds tests for ClimaAtmos space construction, rrtmgp interfaces, callbacks, topography (and some parameterized tendencies).


sriharshakandala:
Can we add a quick summary for each of the tests!

test/callbacks.jl
```diff
@@ -0,0 +1,72 @@
...
+    # atmoscallbacks from discrete callbacks
+    @test cb_3.f!() == π
+    atmos_cbs = CA.atmos_callbacks(cb_set)
+    # test against expected callback outcomes
+    tspan = [0, test_tend]
+    @test CA.n_expected_calls(cb_set, test_dt, tspan)[2] == test_tend
+    @test CA.n_steps_per_cycle_per_cb(cb_set, test_dt) ==
+          [test_nsteps; test_dt; 3]
+end
+
```
Sbozzolo, line 71: This test is not testing that the callback are really working and doing what they are intended to (in terms of frequency). 

For example, I am pretty sure that the `call_every_dt` is currently wrong because of floating point accuracy 
(the check on >= will lead to unstable results). As a result, the callback might be scheduled one step before or after it should.

It might also be good to think more about the role of the AtmosCallbacks in ClimaAtmos. I think they are an opportunity to refactor, and maybe move to a different package so that they can be shared across repos.

test/test_helpers.jl
```diff
@@ -0,0 +1,151 @@
+
```
Sbozzolo, line 1:
```suggestion
```

test/test_helpers.jl
```diff
@@ -0,0 +1,151 @@
...
+       UVW(Geometry.VVector(v)) +
+       UVW(Geometry.WVector(w))
+    ᶠu = @. UVW(Geometry.UVector(ᶠu)) +
+       UVW(Geometry.VVector(ᶠv)) +
+       UVW(Geometry.WVector(ᶠw))
+    # Get covariant components
+    uₕ = @. Geometry.Covariant12Vector(ᶜu)
+    uᵥ = @. Geometry.Covariant3Vector(ᶠu)
+    return uₕ, uᵥ
+end
```
Sbozzolo, line 151: Can you document the purpose of this?

test/topography.jl
```diff
@@ -0,0 +1,60 @@
...
+    coords = Fields.coordinate_field(bubble_space)
+    return (;
+        bubble_space = bubble_space,
+        no_bubble_space = no_bubble_space,
+        lat = lat,
+        long = long,
+        coords = coords,
+        FT = FT,
+    )
+end
```
Sbozzolo, line 52: Isn't this available in test_utils?

test/solver/model_getters.jl
```diff
@@ -0,0 +1,43 @@
...
+
+@testset "Model config" begin
+    config = CA.AtmosConfig(Dict("config" => "box"))
+    parsed_args = config.parsed_args
+    @test CA.get_model_config(parsed_args) isa CA.BoxModel
+
+    config = CA.AtmosConfig(Dict("config" => "plane"))
+    parsed_args = config.parsed_args
+    @test CA.get_model_config(parsed_args) isa CA.PlaneModel
+end
```
Sbozzolo, line 11: Model config is another place where we can refactor: if we use Grids, we immediately know if we are on a Plane or a Box

test/parameterized_tendencies/sponge/rayleigh_sponge.jl
```diff
@@ -0,0 +1,37 @@
...
+    "Rayleigh-sponge functions"
+    ### Boilerplate default integrator objects
+    config = CA.AtmosConfig(Dict("initial_condition" => "DryBaroclinicWave"))
+    (; Y) = generate_test_simulation(config)
+    zmax = maximum(CC.Fields.coordinate_field(Y.f).z)
+    z = CC.Fields.coordinate_field(Y.c).z
+    Y.c.uₕ.components.data.:1 .= ones(axes(Y.c))
+    Y.c.uₕ.components.data.:2 .= ones(axes(Y.c))
+    FT = eltype(Y)
+    ᶜYₜ = Y .* FT(0)
```
Sbozzolo, line 19:
```suggestion
    ᶜYₜ = zero(Y)
```
might work

Sbozzolo (commented):
Can you also add job that runs the tests on GPU in buildkite/

test/callbacks.jl
akshaysridhar: Agreed that this is an opportunity to refactor the callback code and reassess testing scope when this is done - this PR intends to set up patterns for future additions to the test suite. I will follow up with more comprehensive tests over the next few weeks.

test/test_helpers.jl
akshaysridhar: Please see updated block

test/topography.jl
akshaysridhar: Yes, fixed.

PR #2772, opened 2024-03-08
Reduce allocations in NetCDF writer

Sbozzolo:
Every time `write_field` is called for a NetCDF writer, the code would allocate small vectors containing the desired target dimension. Once the file is written the first time, such vectors were only used to ensure that the target file contains consistent information. This PR moves this check one at higher level, removing the need to allocate the vectors.

The function that is allocating is `target_coordinates`, that now is called only when writing the file the first time.

This PR also fixed an unrelated import error.

### DIFF
--- a/src/diagnostics/netcdf_writer.jl
+++ b/src/diagnostics/netcdf_writer.jl
@@ -566,11 +566,11 @@ function NetCDFWriter(;
         interpolate(remapper, Fields.coordinate_field(space).z)
 
     return NetCDFWriter{typeof(num_points), typeof(interpolated_physical_z)}(
-        Dict(),
+        Dict{String, Remapper}(),
         num_points,
         compression_level,
         interpolated_physical_z,
-        Dict(),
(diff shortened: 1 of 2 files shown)


src/diagnostics/netcdf_writer.jl
```diff
@@ -19,41 +19,49 @@
...
+    for (k, v) in kwargs
+        dim.attrib[String(k)] = v
+    end
+
+    dim[:] = points
+
+    return nothing
+end
+
+function dimension_already_exists(
```
charleskawczynski, line 51: Maybe just `dimension_exists`?

charleskawczynski (approved):

src/diagnostics/netcdf_writer.jl
Sbozzolo: Good suggestion!

Sbozzolo:
(This barely changed the allocations, as most of them are in NCDataset.)

PR #2773, opened 2024-03-08
Add job to benchmark IO

Sbozzolo:
This PR:
- Updates to ClimaCore 0.13.1 to use `interpolate!`. Now, the output diagnostics array is allocated only once, and reused.
- Adds a IO performance job. This job benchmarks the IO portion of the NetCDF writer and it compares the performance of our NetCDF writer with just using NCDatasets. Currently, calling sync makes our writer much slower. This will be subject to future investigations.
- Adds a NEWS.md

### DIFF
--- a/perf/benchmark_netcdf_io.jl
+++ b/perf/benchmark_netcdf_io.jl
@@ -1,0 +1,120 @@
+import ClimaAtmos as CA
+import ClimaAtmos.Diagnostics as CAD
+import ClimaComms
+import Profile, ProfileCanvas
+import NCDatasets
+import Base: rm
+using BenchmarkTools
+
+# This script runs our NetCDF writer and compares its performance with
+# NCDatasets. It also produces a flamegraph for IO.
+
+# Number of target interpolation point along each dimension
+const NUM = 100
+
+timings = Dict{ClimaComms.AbstractDevice, Any}(
+    ClimaComms.CPUSingleThreaded() => missing,
+)
+# If a GPU is available, runs on CPU and GPU
+if ClimaComms.device() isa ClimaComms.CUDADevice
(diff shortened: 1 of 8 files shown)


perf/benchmark_netcdf_io.jl
```diff
@@ -0,0 +1,119 @@
...
+    outarray = Array(netcdf_writer.remappers["rhoa"]._interpolated_values)
+    v[1, :, :, :] = outarray
+
+    function add_nc(nc)
+        v = nc["rhoa"]
+        temporal_size, spatial_size... = size(v)
+        time_index = temporal_size + 1
+        nc["time"][time_index] = 10.0 * time_index
+        v[time_index, :, :, :] = outarray
+    end
```
charleskawczynski, line 107:
```suggestion
    function add_nc(nc, outarray)
        v = nc["rhoa"]
        temporal_size, spatial_size... = size(v)
        time_index = temporal_size + 1
        nc["time"][time_index] = 10.0 * time_index
        v[time_index, :, :, :] = outarray
    end
```
Removing the closure over `outarray` allows us to hoist this definition out of the loop. Also, more importantly, you'll want to do that for the benchmark so that it doesn't have to reach outside of the scope of the function.

Sbozzolo: Good point!

charleskawczynski:
Can we increase the number of calls in the `Flame graph: perf target (IO)` job? The flame graph doesn't have any granularity

Sbozzolo:
I can call it more times, but it will lead to the same result: nc_sync domaintes everything. I think the flame will be more useful once I remove nc_sync

What I am looking at is mostly in the stdout:
image

charleskawczynski (approved):
Sounds good, looks good to me.

PR #2779, opened 2024-03-11
add mass flux divergence to detrainment

szy21:
## Purpose 

Adds a term for mass flux divergence in detrainment. This is not implemented for diagnostic edmf as calculating the mass flux divergence at level k requires mass flux at level k, which we don't have when solving edmf variable at level k. We can get around it by using levels k-1 and k-2 to calculate the divergence, but it's a bit more complicated and we are not focusing on improving diagnostic EDMF now.

I keep both `vert_div` and `massflux_vert_div` term, to have no behavior change for diagnostic edmf. We can reevaluate later if `vert_div` is still needed. I also remove `GeneralizedHarmonicsEntrainment` and `GeneralizedHarmonicsDetrainment`, as it's unlikely we will use them.
 
part of #2705 

## Content

----


post_processing/ci_plots.jl
```diff
@@ -954,6 +954,7 @@ EDMFBoxPlots = Union{
     Val{:prognostic_edmfx_gabls_box},
     Val{:prognostic_edmfx_bomex_fixtke_box},
     Val{:prognostic_edmfx_bomex_box},
+    Val{:prognostic_edmfx_bomex_column},
```
szy21, line 957: remove this

szy21:
I will update ClimaParams for other environment as well.

trontrytel (approved):

post_processing/ci_plots.jl
trontrytel: do you still want to remove this?

szy21: yes I will remove this

PR #2786, opened 2024-03-13
Add YAML Linter GHA

nefrathenrici:
This PR adds a basic YAML linter to keep our config files clean. 

Two open questions
- Should this action lint all YAML files, including `pipeline.yml`?
- Should this action be part of the existing Julia formatter action?

If people like this, I will format the YAML files nicely in this PR as well.

### DIFF
--- a/.github/workflows/yaml_format.yml
+++ b/.github/workflows/yaml_format.yml
@@ -1,0 +1,18 @@
+name: YAML Lint
+on: [push]  # yamllint disable-line rule:truthy
+jobs:
+  config:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: yaml-lint
+        uses: ibiqlik/action-yamllint@v3
+        with:
+          file_or_dir: config
+          strict: true
+          # Ignore warning 
+          config_data: |
+            extends: default
+            rules:
+              document-start: disable
+              line-length: disable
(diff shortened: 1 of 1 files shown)


charleskawczynski (approved):
Looks fine to me, maybe someone else can review this, too?

Sbozzolo:
Can you post what the diff/output would look like?

nefrathenrici:
Here is some [sample output](https://github.com/CliMA/ClimaAtmos.jl/actions/runs/8348529195/job/22850655134?pr=2786) - it is not as nice as the Julia formatter because it just lints instead of formatting. 

I am looking into [yamlfmt](https://github.com/google/yamlfmt) for actual formatting

nefrathenrici:
@Sbozzolo I have gotten it to run the diff, so the output will be like this:
```
--- a/config/gpu_configs/gpu_aquaplanet_diagedmf.yml
+++ b/config/gpu_configs/gpu_aquaplanet_diagedmf.yml
@@ -7,23 +7,23 @@ z_max: 55000.0
 z_elem: 63
 dz_bottom: 30.0
 dz_top: 3000.0
-moist: equil 
-surface_setup: DefaultMoninObukhov 
+moist: equil
+surface_setup: DefaultMoninObukhov
 rad: allskywithclear
 idealized_insolation: false
 dt_rad: 1hours
 dt_cloud_fraction: 1hours
-turbconv: diagnostic_edmfx 
+turbconv: diagnostic_edmfx
 implicit_diffusion: true
 ```

Sbozzolo:
If 90 % of what this code does is removing trailing spaces, I think we should set up something that removes trailing spaces from every file. We have tons of them.

PR #2788, opened 2024-03-13
change y axis to linear scale in ci plots

szy21:
## Purpose 

Because this is preferred by Tapio:)

## Content

----

### DIFF
--- a/post_processing/ci_plots.jl
+++ b/post_processing/ci_plots.jl
@@ -27,7 +27,7 @@
 #             vars,
 #             y = 0.0,
 #             time = LAST_SNAP,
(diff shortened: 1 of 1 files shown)


szy21:
Actually I like the previous scale for some cases. I’ll keep both options.

Sbozzolo:
If you want linear scale, I think you can just remove the argument. That's the default.

szy21:
> If you want linear scale, I think you can just remove the argument. That's the default.

Yeah, I kept the argument just to have more ticks. I don't care too much about it though, so I can remove it if you prefer.

Sbozzolo:
No, it's fine.

PR #2789, opened 2024-03-14
Revert "Add nsys reports back into CI jobs"

szy21:
Reverts CliMA/ClimaAtmos.jl#2782

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -641,9 +641,9 @@ steps:
       - label: "GPU: GPU dry baroclinic wave"
         key: "target_gpu_implicit_baroclinic_wave"
         command:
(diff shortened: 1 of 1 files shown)


Sbozzolo (approved):
Jobs occasionally timeout with nsys.

charleskawczynski:
> Jobs occasionally timeout with nsys.

Which build timed out? Are we sure that this was an issue with nsight systems?

Sbozzolo:
This is an example I could find: https://buildkite.com/clima/climaatmos-ci/builds/16795#_ 

The commit was just changing the plots in a way that does not affect GPU jobs at all (https://github.com/CliMA/ClimaAtmos.jl/commit/7761daa9cff161543aa9f5f08a29c2d21eefa627)

We can find more examples if we look at old builds

PR #2795, opened 2024-03-16
use parameters for surface albedo

### LINKED ISSUE #2757: Add refractive index to parameters
The refractive index (n) in the regression-based albedo can be calibrated to match observations. It would be good to make it a parameter.

szy21:
## Purpose 

Use clima parameters for ocean surface albedo and water refractive index. Closes #2757 

## To-do

- [x] update ClimaParams in all environments

## Content

----

### DIFF
--- a/src/parameters/Parameters.jl
+++ b/src/parameters/Parameters.jl
@@ -74,6 +74,9 @@ Base.@kwdef struct ClimaAtmosParameters{FT, TP, RP, IP, MPP, WP, SFP, TCP} <:
     zd_viscous::FT
     zd_rayleigh::FT
     kappa_2_sponge::FT
+    # Radiation
+    idealized_ocean_albedo::FT
+    water_refractive_index::FT
 end
 
 Base.eltype(::ClimaAtmosParameters{FT}) where {FT} = FT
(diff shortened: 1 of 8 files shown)


src/parameters/Parameters.jl
```diff
@@ -74,6 +74,9 @@ Base.@kwdef struct ClimaAtmosParameters{FT, TP, RP, IP, MPP, WP, SFP, TCP} <:
     zd_viscous::FT
     zd_rayleigh::FT
     kappa_2_sponge::FT
+    # Radiation
+    ocean_surface_albedo::FT
```
LenkaNovak, line 78: It could be more clear to call this "idealized_ocean_albedo".

LenkaNovak (approved):
LGTM, thank you, just a minor comment on the naming.

src/parameters/Parameters.jl
szy21: Sounds good, I will add this to the next ClimaParams breaking release.

src/parameters/create_parameters.jl
```diff
@@ -111,6 +111,8 @@ function create_parameter_set(config::AtmosConfig)
         :drag_layer_vertical_extent => :σ_b,
         :kappa_2_sponge => :kappa_2_sponge,
         :held_suarez_minimum_temperature => :T_min_hs,
+        :ocean_surface_albedo => :ocean_surface_albedo,
```
nefrathenrici, line 114:
```suggestion
        :ocean_surface_albedo => :idealized_ocean_albedo,
```
Given @LenkaNovak's comment, we can rename this within ClimaAtmos without a new release.

szy21: Thanks! I will change this.

PR #2800, opened 2024-03-18
Try cuda always_inline

charleskawczynski:
This PR tries out always using `always_inline` for all CUDA kernel launches, since this can reduce function call overhead.


charleskawczynski:
Superseded by #2833.

Issue #2783, opened 2024-03-12
`surface_conditions_type` is unnecessarily large


charleskawczynski:
We should reduce the number of fields in `surface_conditions_type`, and instead compute its fields on the fly. The kernel that returns this is more expensive than calling saturation adjustment on the A100 (because there are so many memory loads).

PR #2805, opened 2024-03-18
Simplify surface function

charleskawczynski:
#2800 revealed that we're not able to always inline. The first culprit we run into is `surface_state_to_conditions`, which is quite complicated. This PR breaks off the temperature computation into a separate function, and passes less data to the function.

This function call (to `surface_state_to_conditions`) is more expensive than the call to saturation adjustment, which involves a non-linear solve. So, there is clearly a performance issue.

Related issue: #2783.


charleskawczynski:
Is this breaking for the coupler? cc @LenkaNovak, @juliasloan25. If so, I'll update the news.

Sbozzolo (approved):

juliasloan25:
> Is this breaking for the coupler? cc @LenkaNovak, @juliasloan25. If so, I'll update the news.

no it's not, thank you for checking though!

LenkaNovak (approved):
The restructuring LGTM (after resolving the current parsing errors) and shouldn't affect the coupler.

PR #2808, opened 2024-03-18
fix docstring

trontrytel:


akshaysridhar:
Thanks :) Caught while running tests against the 1M scheme

akshaysridhar (approved):

trontrytel:
> Thanks :) Caught while running tests against the 1M scheme

Thanks for catching that!

PR #2809, opened 2024-03-19
Refactor surface

charleskawczynski:
This PR refactors the surface `surface_state_to_conditions` by:

 - Making `surface_state_to_conditions` take the wrapped surface setup, and a local `surface_state` is then made inside.
 - Made `sfc_prognostic_temp` a non-optional argument and removed the splatting (sometimes this is difficult for the compiler to infer).

There's several benefits to the first bullet:
 - Expressions directly inside broadcast expressions are treated differently than if they were called inside, and sometimes this will result in allocations / runtime calls (https://github.com/CliMA/ClimaCore.jl/issues/1649).
 - The compiler may be able to better optimize this because it is explicitly clear that `surface_state`'s can be locally defined, and a full field of `surface_state`'s is not necessary.
 - We also explicitly pass the local geometry and interior z fewer times to the function

According to my local tests, this PR allows us to use `always_inline` for this kernel.

### DIFF
--- a/src/surface_conditions/surface_conditions.jl
+++ b/src/surface_conditions/surface_conditions.jl
@@ -27,14 +27,9 @@ function update_surface_conditions!(Y, p, t)
     wrapped_sfc_setup = sfc_setup_wrapper(sfc_setup)
     sfc_temp_var =
         p.atmos.surface_model isa PrognosticSurfaceTemperature ?
-        (; sfc_prognostic_temp = Fields.field_values(Y.sfc.T)) : (;)
+        Fields.field_values(Y.sfc.T) : nothing
     @. sfc_conditions_values = surface_state_to_conditions(
-        surface_state(
-            wrapped_sfc_setup,
-            sfc_local_geometry_values,
-            int_z_values,
-            t,
-        ),
+        wrapped_sfc_setup,
         sfc_local_geometry_values,
         int_ts_values,
         projected_vector_data(CT1, int_u_values, int_local_geometry_values),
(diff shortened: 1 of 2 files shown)


src/surface_conditions/surface_conditions.jl
```diff
@@ -27,14 +27,9 @@ function update_surface_conditions!(Y, p, t)
     wrapped_sfc_setup = sfc_setup_wrapper(sfc_setup)
     sfc_temp_var =
         p.atmos.surface_model isa PrognosticSurfaceTemperature ?
-        (; sfc_prognostic_temp = Fields.field_values(Y.sfc.T)) : (;)
+        Fields.field_values(Y.sfc.T) : nothing
```
szy21, line 30: Would this make it harder if we want to add more prognostic variables to `Y.sfc` in the future?

szy21 (approved):
Thanks!

src/surface_conditions/surface_conditions.jl
charleskawczynski: I don't think so, we could also just pass `Fields.field_values(Y.sfc)`, which will give us all of them. I mostly wanted to just get rid of the splatting

charleskawczynski:
Nice, this also appears to fix 7 JET failures in `update_surface_conditions!` for the GPU job.

The additional allocations very well could just be from https://github.com/JuliaGPU/Adapt.jl/issues/79. So, I'm going to bump the allocation limit and merge.

PR #2811, opened 2024-03-19
remove alias and type in toml files

szy21:
## Purpose 

Also removes `parameter_tests.toml` as it doesn't seem to be used anywhere.

## Content

----


nefrathenrici (approved):
Thank you!

Issue #2365, opened 2023-11-14
Make prognostic EDMF run on GPU


szy21:

szy21:
Currently failing: https://buildkite.com/clima/climaatmos-ci/builds/14915#018bd089-0cc4-4358-91fc-b01a72906e7b

szy21:
The most recent failure: https://buildkite.com/clima/climaatmos-ci/builds/17646#018e5944-daae-493a-bb13-e0744e0a254c. cc @charleskawczynski

charleskawczynski:
Accidentally closed by #2366.

szy21:
Fixed by #2819

PR #2814, opened 2024-03-20
Fix an inference failure in gpu prognostic edmf

charleskawczynski:
This PR fixes an inference failure when calling `mapreduce` inside broadcast expressions. Generally speaking, we need to provide `init` so that `mapreduce` can infer the result (due to the edge case of an empty collection). In one other case, however, we're hitting a compiler heuristic, and we need to define a custom `unrolled_dot_product`, so that the compiler can unroll / infer the result. 

relevant issue: #2365.

I'm going to merge this, and I'll followup on the next issue when I have a chance. Thanks again @szy21 for adding that job!

### DIFF
--- a/perf/flame.jl
+++ b/perf/flame.jl
@@ -50,7 +50,7 @@ allocs_limit["flame_perf_target_threaded"] = 1_276_864
 allocs_limit["flame_perf_target_callbacks"] = 398_984
 allocs_limit["flame_perf_gw"] = 3_268_961_856
 allocs_limit["flame_perf_target_prognostic_edmfx_aquaplanet"] = 299_616
-allocs_limit["flame_gpu_implicit_barowave_moist"] = 381_968
+allocs_limit["flame_gpu_implicit_barowave_moist"] = 658_664
 # Ideally, we would like to track all the allocations, but this becomes too
 # expensive there is too many of them. Here, we set the default sample rate to
 # 1, but lower it to a smaller value when we expect the job to produce lots of
--- a/src/cache/diagnostic_edmf_precomputed_quantities.jl
+++ b/src/cache/diagnostic_edmf_precomputed_quantities.jl
@@ -63,7 +63,8 @@ function set_diagnostic_edmfx_env_quantities_level!(
     turbconv_model,
 )
     @. u³⁰_halflevel = divide_by_ρa(
-        ρ_level * u³_halflevel - mapreduce(*, +, ρaʲs_level, u³ʲs_halflevel),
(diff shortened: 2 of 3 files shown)


dennisYatunin:
Does adding `init` to the `mapreduce` calls help with inference, or is it just about the unrolling?

szy21:
Thanks! I'm going to clean up the updraft top thing before the end of this week. Hopefully, we will just remove it.

charleskawczynski:
> Does adding `init` to the `mapreduce` calls help with inference, or is it just about the unrolling?

Maybe? I meant to check

charleskawczynski:
`mapreduce` with `init` works! (and it works when using `always_inline`), so this should be a good improvement! 🎉

charleskawczynski:
Going to merge, @dennisYatunin feel free to bikeshed/update the name if you'd like!

charleskawczynski:
Oh, wait, no, it's not working: https://buildkite.com/clima/climaatmos-ci/builds/17697#018e5e5e-ef2e-491c-bf48-40ed96006e76/196-260. Hm, that's odd. I'll have to take a look tomorrow. I still have the previous version stashed somewhere.

charleskawczynski:
Ah, I didn't see that we're using those functions in 

```julia
function set_sgs_ᶠu₃!(w_function, ᶠu₃, Y, turbconv_model)
    ρaʲs(sgsʲs) = map(sgsʲ -> sgsʲ.ρa, sgsʲs)
    u₃ʲs(sgsʲs) = map(sgsʲ -> sgsʲ.u₃, sgsʲs)
    @. ᶠu₃ = w_function(
        ᶠinterp(ρaʲs(Y.c.sgsʲs)),
        u₃ʲs(Y.f.sgsʲs),
        ᶠinterp(Y.c.ρ),
        Y.f.u₃,
        turbconv_model,
    )
    return nothing
end
```
I don't have that sort of example in my local reproducer, I'll update it and see what's going on. We may just need to revert to the simple recursion pattern. These tuples won't be large anyway, so it's probably not a bad use-case.

charleskawczynski:
Ok, the compromise is that we can use `mapreduce` with `init` (I've defined `mapreduce_with_init` since the `init` is somewhat complicated) for some cases, I got rid of `unrolled_mapsum`, but we still need `unrolled_dot_product`. This seems to be hitting a compiler heuristic.

charleskawczynski:
Hm, gpu allocations are still pretty noisy. I'm going to manually merge this

Issue #2818, opened 2024-03-21
Fix allocations in pronostic edmf implicit solver


szy21:
In #2607, we found that running prognostic edmf implicit increases allocation by several orders of magnitude. This is likely from the complicated Jacobian structure (removing some blocks decreases allocation).

charleskawczynski:
Some of these might actually be fixed by #2814.

PR #2823, opened 2024-03-22
fix some allocation in prognostic edmf implicit solver

szy21:
## Purpose 

This brings down the difference in allocation between implicit and explicit prognostic edmf from several orders of magnitude to 10. I'm not sure where the factor 10 is, and we can look at it further. Turns back on `implicit_diffusion` in the performance job config. Part of #2818 

## Content

----

### DIFF
--- a/src/prognostic_equations/implicit/implicit_solver.jl
+++ b/src/prognostic_equations/implicit/implicit_solver.jl
@@ -418,6 +418,9 @@ NVTX.@annotate function Wfact!(A, Y, p, dtγ, t)
         p.scratch.ᶜadvection_matrix,
         p.scratch.ᶜdiffusion_h_matrix,
         p.scratch.ᶜdiffusion_u_matrix,
+        p.scratch.ᶠbidiagonal_matrix_ct3,
+        p.scratch.ᶠbidiagonal_matrix_ct3_2,
+        p.scratch.ᶠtridiagonal_matrix_c3,
         p.dt,
         p.params,
         p.atmos,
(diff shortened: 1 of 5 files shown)


szy21:
hmm the allocation still increases by several orders of magnitude in the buildkite job. I'll take a look tomorrow.

szy21:
Now the allocation is back to 10x increase. It also fixes some gpu errors, but there is one remaining. I'll look at it in the next PR.

src/prognostic_equations/implicit/implicit_solver.jl
```diff
@@ -835,34 +851,30 @@ function update_implicit_equation_jacobian!(A, Y, p, dtγ, colidx)
...
+            tmp = ᶠtridiagonal_matrix_c3
+            @. tmp[colidx] =
+                -(ᶠgradᵥ_matrix()) ⋅ convert(
+                    BidiagonalMatrixRow{FT},
+                    ifelse(
+                        ᶜu₃ʲ[colidx].components.data.:1 > 0,
+                        convert(BidiagonalMatrixRow{FT}, ᶜleft_bias_matrix()),
+                        convert(BidiagonalMatrixRow{FT}, ᶜright_bias_matrix()),
+                    ),
+                )
```
charleskawczynski, line 865: Why is the convert needed here?

dennisYatunin: Otherwise the `ifelse` isn't type-stable (it can return either a `BandMatrixRow` on the `-half` diagonal or the `+half` diagonal, depending on the velocity component).

dennisYatunin: Oh, sorry, I thought you were asking about the inner `convert`. The outer `convert` shouldn't be needed, but I guess it reduced the allocations?

szy21: Ah, we were trying to see if this helps with GPU. I'll remove it.

charleskawczynski (commented):
So, in summary, this adds implicit diffusion, which drastically increases allocations, and then the code changes reduce the allocations, but the result is that it's still 10x the current allocations? If I've got that right, then feel free to merge!

Also, you may want to rebase after the inlining has landed!

charleskawczynski (approved):

szy21:
> So, in summary, this adds implicit diffusion, which drastically increases allocations, and then the code changes reduce the allocations, but the result is that it's still 10x the current allocations? If I've got that right, then feel free to merge!

Yes, it's hard to say how much the increase was before, as the job kept failing. After I fixed something it was 10000x, so it was more than that. And now after this PR it is 10x.

> Also, you may want to rebase after the inlining has landed!
Sounds good!

charleskawczynski:
I think some of these issues will be upgraded in julia 1.11

Issue #2827, opened 2024-03-22
Diagnostics callback is slow


charleskawczynski:
I ran the gpu dyamond case on the A100, and saw that the diagnostics callback (callback 5) is slow:

<img width="975" alt="Screen Shot 2024-03-22 at 9 54 08 AM" src="https://github.com/CliMA/ClimaAtmos.jl/assets/1880641/1cd91d6c-a10f-48fe-9555-bf9234962736">

Roughly:
 - a full cycle is about 23 seconds
 - rrtmgp kernels: 2.6 seconds
 - remapper / IO: 4.6 seconds
 - steps between callback 5: 16 seconds

One thing is clear: the gpu is not busy during the remapper / IO segment.

Sbozzolo:
![image](https://github.com/CliMA/ClimaAtmos.jl/assets/9167485/89c76f7e-9148-4263-beee-cd348aef873f)

Compute for diagnostics is less than 1 %. All the time is spent in NCDataset.

PR #2828, opened 2024-03-22
Group NCDatasets.sync calls

charleskawczynski:
In light of #2827, this PR hoists the `NCDatasets.sync` to a separate callback so that we can (eventually) and use `Threads.@spawn` to leverage threads (if we have them).

I've also named/annotated `compute_callback` and `output_callback`, so that they show up in the NVTX reports.

### DIFF
--- a/src/diagnostics/diagnostic.jl
+++ b/src/diagnostics/diagnostic.jl
@@ -661,6 +661,38 @@ function accumulate!(diag_accumulator, diag_storage, reduction_time_func)
     return nothing
 end
 
+any_sync_needed(step, diagnostics) =
+    any(diag -> sync_needed(step, diag), diagnostics)
+
+# sync_needed(step, diag) = diag.cbf.n > 0 && step % diag.cbf.n == 0
+sync_needed(step, diag) = diag.output_every > 0 && step % diag.output_every == 0
+
+import NVTX
+NVTX.@annotate function sync_nc_datasets(diagnostics, output_dir, step, context)
+    any(diag -> diag.output_writer isa NetCDFWriter, diagnostics) || return
+    any_sync_needed(step, diagnostics) || return
+    ClimaComms.iamroot(context) || return
+    tasks = map(diagnostics) do diag
+        writer = diag.output_writer
+        if writer isa NetCDFWriter && sync_needed(step, diag)
+            output_path = outpath_name(output_dir, diag)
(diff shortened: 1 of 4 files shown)


Sbozzolo:
I think we should also avoid calling sync every time. Instead, let's flush the output once per diagnostic cycle.

charleskawczynski:
> I think we should also avoid calling sync for every diagnostic. Instead, let's flush the output once per diagnostic cycle.
> 
> As in #2623

Yeah, I agree, but first I'd like to see how this performs, since it's functionally equivalent. Updating it to be called every diagnostics lcd cycle is a much simpler change than this is turning out to be.

Sbozzolo:
> > I think we should also avoid calling sync for every diagnostic. Instead, let's flush the output once per diagnostic cycle.
> > As in #2623
> 
> Yeah, I agree, but first I'd like to see how this performs, since it's functionally equivalent. Updating it to be called every diagnostics lcd cycle is a much simpler change than this is turning out to be.

Do you mean functionally equivalent to main?

charleskawczynski:
> Do you mean functionally equivalent to main?

Yeah. It's technically different in terms of order of compute / io, but it should at least be calling `sync` the same number of times.

src/diagnostics/diagnostic.jl
```diff
@@ -661,6 +661,37 @@ function accumulate!(diag_accumulator, diag_storage, reduction_time_func)
...
+
+import NVTX
+NVTX.@annotate function sync_nc_datasets(step, diagnostics, output_dir)
+    any(diag -> diag.output_writer isa NetCDFWriter, diagnostics) || return
+    any_sync_needed(step, diagnostics) || return
+    tasks = map(diagnostics) do diag
+        writer = diag.output_writer
+        if writer isa NetCDFWriter && sync_needed(step, diag)
+            output_path = outpath_name(output_dir, diag)
+            # TODO: find out why Threads.@spawn segfaults in multithreaded IO test
```
Sbozzolo, line 678: Something we might want to keep in mind is this:

https://juliaparallel.org/MPI.jl/stable/reference/environment/#MPI.ThreadLevel

Sbozzolo: (It should not be a problem at the moment because sync does not contain an MPI call.)

charleskawczynski: Yeah, I thought that only the root process does this, but it's a good point.

PR #2829, opened 2024-03-22
Refactor get_diagnostics

charleskawczynski:
This PR refactors `get_diagnostics`:

 - remove `return` statements from `map` (it's a bit confusing because `return` doesn't return from the function)
 - Moves the `writers` `@info` statement  into`get_diagnostics`


charleskawczynski:
This is a baby-step from #2643.

Issue #2831, opened 2024-03-22
GPU allocations are unstable


Sbozzolo:
The perf job that tracks gpu allocation can change significantly across jobs.

For example, it passes here: https://buildkite.com/clima/climaatmos-ci/builds/17742
It fails here: https://buildkite.com/clima/climaatmos-ci/builds/17747

charleskawczynski:
The flakiness of that job is probably dominated by `orchestrate_diagnostics`. It's probably not the leading allocator for real runs, but it's significant for that particular flame graph job.

<img width="1991" alt="Screen Shot 2024-03-22 at 3 04 11 PM" src="https://github.com/CliMA/ClimaAtmos.jl/assets/1880641/4b2fe1b6-db51-49ba-9c34-0f82d8c16d15">

PR #2834, opened 2024-03-22
Add function barrier to orchestrate diagnostics

charleskawczynski:
This PR adds a function barrier to orchestrate diagnostics, in hopes to fix some of the diagnostics allocations (#2831.)

I actually kind of like this better anyway, since it's more clear what variables we're using in closures.

### DIFF
--- a/perf/flame.jl
+++ b/perf/flame.jl
@@ -38,19 +38,19 @@ ProfileCanvas.html_file(joinpath(output_dir, "flame.html"), results)
 #####
 
 allocs_limit = Dict()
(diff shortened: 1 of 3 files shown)


charleskawczynski:
It looks like this had a pretty big impact.

charleskawczynski:
Before:

Screen Shot 2024-03-22 at 4 53 20 PM

After:

Screen Shot 2024-03-22 at 4 53 56 PM

(this change is from the first commit)

Sbozzolo:
Indeed, this is much nicer!

PR #2837, opened 2024-03-23
Fix benchmark timing for netcdf_io

charleskawczynski:
The existing io benchmark is not using `CUDA.@sync` inside `@benchmark`, which means that the timings do not include kernel execution time (see the CUDA docs [here](https://cuda.juliagpu.org/dev/development/profiling/#Time-measurements)).

We could alternatively use https://github.com/CliMA/ClimaComms.jl/pull/73 once it lands to avoid some code duplication.

### DIFF
--- a/perf/benchmark_netcdf_io.jl
+++ b/perf/benchmark_netcdf_io.jl
@@ -85,15 +85,16 @@ for device in keys(timings)
     @info "Flame saved in $flame_path"
 
     @info "Benchmarking our NetCDF writer (only IO) ($device_name)"
-    timings[device] = @benchmark CAD.save_diagnostic_to_disk!(
-        $netcdf_writer,
-        $field,
-        $rhoa_diag,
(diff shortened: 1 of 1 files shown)


Sbozzolo:
Thanks for pointing this out!

I am fine with using directly @cuda_sync if you prefer.

charleskawczynski:
I've also removed the loop over timings (it's still cpu/gpu compatible), since we never dynamically change the device.

Sbozzolo:
> I've also removed the loop over timings (it's still cpu/gpu compatible), since we never dynamically change the device.

The loop over timings is to change the device so that we benchmark both CPU and GPU (when on a GPU node)

charleskawczynski:
Yes, I noticed that and reverted that change.

Sbozzolo:
Moved to ClimaDiagnostics

PR #2847, opened 2024-03-28
Separate precipitation sources and sinks

trontrytel:
This PR:
- separates precipitation sources and sinks for the 1-moment microphysics. This is in preparation for coupling with the SGS model. We will need an SGS version of the precipitation sources. The precipitation sinks remain unchanged for w/wo SGS model.
- moves the precipitation sources and sinks into the `microphysics_warppers` file
- introduces the `compute_precipitation_cache` function for the 1-moment scheme. Hopefully this is where we will dispatch between w/wo SGS model
- reuses the `q_tot_precipitation_sources` for the 0-moment microphysics.

This should not introduce any changes in the behavior.

### DIFF
--- a/src/parameterized_tendencies/microphysics/microphysics_wrappers.jl
+++ b/src/parameterized_tendencies/microphysics/microphysics_wrappers.jl
@@ -3,64 +3,280 @@
 import Thermodynamics as TD
 import CloudMicrophysics.Microphysics0M as CM0
 
+# define some aliases and functions to make the code more readable
+const Iₗ = TD.internal_energy_liquid
+const Iᵢ = TD.internal_energy_ice
+const Lf = TD.latent_heat_fusion
+const Tₐ = TD.air_temperature
+const PP = TD.PhasePartition
+const qᵥ = TD.vapor_specific_humidity
+qₗ(thp, ts) = TD.PhasePartition(thp, ts).liq
+qᵢ(thp, ts) = TD.PhasePartition(thp, ts).ice
+
+# helper function to safely get precipitation from state
+function qₚ(ρqₚ::FT, ρ::FT) where {FT}
+    return max(FT(0), ρqₚ / ρ)
+end
+
+# helper function to limit the tendency
+function limit(q::FT, dt::FT) where {FT}
+    return q / dt / 5
+end
+
 """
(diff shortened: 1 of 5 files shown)


src/parameterized_tendencies/microphysics/microphysics_wrappers.jl
```diff
@@ -3,22 +3,42 @@
 import Thermodynamics as TD
 import CloudMicrophysics.Microphysics0M as CM0
 
+# define some aliases and functions to make the code more readable
+const Iₗ = TD.internal_energy_liquid
+const Iᵢ = TD.internal_energy_ice
+const qₜ = TD.PhasePartition
```
szy21, line 9: I think q_t is a bit confusing, but I don't have a better name. q_pt may be better, but we are using q_pt in some other places in the code so I'm not sure if that's less confusing.

szy21 (approved):
Looks good, thanks! Just one comment on the names:)

src/parameterized_tendencies/microphysics/microphysics_wrappers.jl
trontrytel: How about `PP = TD.PhasePartition` ?

src/parameterized_tendencies/microphysics/precipitation.jl
```diff
@@ -40,16 +40,19 @@ function precipitation_cache(Y, precip_model::Microphysics0Moment)
 end
 
 function compute_precipitation_cache!(Y, p, colidx, ::Microphysics0Moment, _)
-    (; params) = p
+    (; params, dt) = p
     (; ᶜts) = p.precomputed
     (; ᶜS_ρq_tot) = p.precipitation
     cm_params = CAP.microphysics_params(params)
     thermo_params = CAP.thermodynamics_params(params)
-    #TODO missing limiting by q_tot/Δt
```
trontrytel, line 48: It seems that I lied and there are some small changes in the mse for the 0-moment scheme. I think they are coming from the fact that now I'm using `q_tot_precipitation_sources` function which does limit the tendency by `q_tot / dt`. I think this is better than before, so I'll bump the ref counter

PR #2854, opened 2024-03-28
Ensure that the CI plots are in the CI folder

Sbozzolo:
`ci_plots` writes the files in the first folder, so it is currently writing to the our reference folder as opposed to the ci one. This swaps the order, ensuring that we write the correct artifact path.

I also did some import cleaning in the meantime.


trontrytel (approved):
Thank you!

PR #2858, opened 2024-03-29
Add 1-moment microphysics to prognostic EDMF

trontrytel:
This PR adds 1-moment microphysics scheme to prognostic EDMF.

Precipitation (rain and snow) is only computed as a grid mean quantity (i.e. we don't have updraft and environment precipitation variables). The sources to precipitation are computed as a sum of contributions from updrafts and the environment. The precipitation sinks (evaporation, melting, deposition/sublimation) are only computed on the grid mean. 

My main priority was to only have one definition of `compute_precipitation_sources!` and `compute_precipitation_sinks!`. This is where I loop over all microphysics processes and I don't want to have multiple versions of it. That function returns 4 outputs (source terms to total water, rain, snow and energy) and I didn't know how to write it outside of a `colidx` loop.

Things I don't like about the code right now:
- the additional `colidx` loop
- all the additional variables I'm adding to the cache
- `compute_precipitation_sinks!` and `compute_precipitation_sources!` behave differently. The `sources` one zeroes out the cached `S` before adding to it. The `sinks` does not. This is because when used without EDMF, the `sinks` are done after the `sources` and add to the same cached values. This is clunky, but I didn't have a better idea and I did not want to add another 4 variables to the cache

Also, can I re-use my tmp scalar between different updraft j-s and the environment?

```
    tmp1 = p.scratch.ᶜtemp_scalar
    n = number of updrafts

    do colidx
        # Sources from the updrafts
        for j in 1:n
            foo!(
                tmp1[colidx],
               all_the_sgs_j_variables_and_dollars[colidx]
            )
        end
```

### DIFF
--- a/config/model_configs/prognostic_edmfx_trmm_column_0M.yml
+++ b/config/model_configs/prognostic_edmfx_trmm_column_0M.yml
@@ -1,4 +1,4 @@
-job_id: prognostic_edmfx_trmm_column
+job_id: prognostic_edmfx_trmm_column_0M
 initial_condition: TRMM_LBA
 rad: TRMM_LBA
 surface_setup: TRMM_LBA
@@ -13,7 +13,7 @@ edmfx_velocity_relaxation: true
 prognostic_tke: true
 moist: equil
 apply_limiter: false
-precip_model: "1M"
+precip_model: "0M"
 config: column
 z_max: 16400
 x_elem: 2
@@ -30,9 +30,5 @@ netcdf_interpolation_num_points: [8, 8, 82]
 diagnostics:
   - short_name: [ts, ta, thetaa, ha, pfull, rhoa, ua, va, wa, hur, hus, cl, clw, cli, hussfc, evspsbl]
     period: 10mins
-  - short_name: [arup, waup, taup, thetaaup, haup, husup, hurup, clwup, cliup, waen, taen, thetaaen, haen, husen, huren, clwen, clien, tke]
-    period: 10mins
-  - short_name: [entr, detr, lmix, bgrad, strain, edt, evu]
-    period: 10mins
-  - short_name: [husra, hussn]
+  - short_name: [arup, waup, taup, thetaaup, haup, husup, hurup, clwup, cliup, waen, taen, thetaaen, haen, husen, huren, clwen, clien, tke, lmix]
     period: 10mins
--- a/src/parameterized_tendencies/microphysics/precipitation.jl
+++ b/src/parameterized_tendencies/microphysics/precipitation.jl
@@ -18,7 +18,7 @@ precipitation_cache(Y, atmos::AtmosModel) =
 #####
 
 precipitation_cache(Y, precip_model::NoPrecipitation) = (;)
-precipitation_tendency!(Yₜ, Y, p, t, colidx, ::NoPrecipitation) = nothing
+precipitation_tendency!(Yₜ, Y, p, t, colidx, ::NoPrecipitation, _) = nothing
 
 #####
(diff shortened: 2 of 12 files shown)


src/parameterized_tendencies/microphysics/precipitation.jl
```diff
@@ -268,3 +309,46 @@ function precipitation_tendency!(
...
+    (; ᶜSeₜᵖ⁰, ᶜSqₜᵖ⁰, ᶜSqᵣᵖ⁰, ᶜSqₛᵖ⁰, ᶜρa⁰) = p.precomputed
+    # Source terms from EDMFX updrafts
+    (; ᶜSeₜᵖʲs, ᶜSqₜᵖʲs, ᶜSqᵣᵖʲs, ᶜSqₛᵖʲs) = p.precomputed
+    # Grid mean precipitation sinks
+    (; ᶜSqₜᵖ, ᶜSqᵣᵖ, ᶜSqₛᵖ, ᶜSeₜᵖ) = p.precipitation
+
+    compute_precipitation_cache!(Y, p, colidx, precip_model, turbconv_model)
+
+    # Update from environment precipitation sources
+    @. Yₜ.c.ρ[colidx] += ᶜρa⁰[colidx] * ᶜSqₜᵖ⁰[colidx]
```
trontrytel, line 332: I'll wire the environment and grid mean contributions in one line

config/model_configs/prognostic_edmfx_trmm_column_0M.yml
```diff
@@ -0,0 +1,34 @@
+job_id: prognostic_edmfx_trmm_column
```
szy21, line 1:
```suggestion
job_id: prognostic_edmfx_trmm_column_0M
```

src/parameterized_tendencies/microphysics/precipitation.jl
```diff
@@ -218,6 +212,7 @@ function compute_precipitation_cache!(Y, p, colidx, ::Microphysics1Moment, _)
         ᶜSqₛᵖ[colidx],
         ᶜSeₜᵖ[colidx],
         Y.c.ρ[colidx],
+        Y.c.ρ[colidx],
```
szy21, line 215: I find it a bit confusing to pass in two `Y.c.ρ` here. It seems one of them is to get specific quantities of q_rai and q_snow? How about using q_rai and q_snow as input to the function?

szy21 (commented):
Thanks! Just some minor comments for now. I will go through the rest later today or over the weekend.

szy21 (approved):
Just a thought. Looking at the code, another way to calculate sources for one moment microphysics is to dispatch on `compute_precipitation_sources!` by the sgs model, rather than saving the precipitation sources in edmf precomputed quantities. We can keep this for now and see what would be better as we add more options.

trontrytel:
> Just a thought. Looking at the code, another way to calculate sources for one moment microphysics is to dispatch on `compute_precipitation_sources!` by the sgs model, rather than saving the precipitation sources in edmf precomputed quantities. We can keep this for now and see what would be better as we add more options.

Sounds good. I agree, there should be a more elegant way to switch between 0M/1M w/wo EDMF(diagnostic and prognostic).  And also w/wo quadratures in the future. I think with the promised advent of fusion I could write the whole 1-moment microphysics as a point-wise function that returns a tuple of numbers. Instead of having to use it inside `colidx` call. 

But it is easier for me to write a better solution once we have something working.

trontrytel:
I'll also add some flame graph job for the 1M

PR #2871, opened 2024-04-03
Cleanup in 0M diagnostic EDMF

trontrytel:
This PR:
-  gets rid of the `ᶜS_e_totʲs_helper` from the 0-moment diagnostic edmf. I think with that I can have the same cache structure for the prognostic and diagnostic edmf for both 0M and 1M options
- adds `ᶜS_ρe_tot` to the 0-moment precipitation cache to simplify the tendency evaluations
- makes diagnostic 0M environmental precip a dispatch choice
- what I do in the diagnostic 0M updraft precipitation is a little hacky, but it avoids allocating for when we are not running with 0M scheme

### DIFF
--- a/src/cache/precomputed_quantities.jl
+++ b/src/cache/precomputed_quantities.jl
@@ -58,10 +58,7 @@ function precomputed_quantities(Y, atmos)
     cloud_diagnostics = (; ᶜcloud_fraction = similar(Y.c, FT),)
     precipitation_sgs_quantities =
         atmos.precip_model isa Microphysics0Moment ?
-        (;
-            ᶜS_q_totʲs = similar(Y.c, NTuple{n, FT}),
-            ᶜS_q_tot⁰ = similar(Y.c, FT),
-        ) :
+        (; ᶜSqₜᵖʲs = similar(Y.c, NTuple{n, FT}), ᶜSqₜᵖ⁰ = similar(Y.c, FT)) :
         atmos.precip_model isa Microphysics1Moment ?
         (;
             ᶜSeₜᵖʲs = similar(Y.c, NTuple{n, FT}),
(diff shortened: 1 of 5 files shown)


src/cache/precomputed_quantities.jl
```diff
@@ -122,9 +122,6 @@ function precomputed_quantities(Y, atmos)
             ᶜentrʲs = similar(Y.c, NTuple{n, FT}),
             ᶜdetrʲs = similar(Y.c, NTuple{n, FT}),
             ᶠnh_pressure³ʲs = similar(Y.f, NTuple{n, CT3{FT}}),
-            ᶜS_q_totʲs = similar(Y.c, NTuple{n, FT}),
-            ᶜS_q_tot⁰ = similar(Y.c, FT),
```
szy21, line 126: You still need these two?

szy21 (approved):
Looks good, thanks!

src/cache/precomputed_quantities.jl
trontrytel: https://github.com/CliMA/ClimaAtmos.jl/blob/40e6a2fa0a77aaa74d4ad666eb5f0d97711fef80/src/cache/precomputed_quantities.jl#L134

I am adding them here. But only for the 0M scheme. I guess diagnostic edmf doesn't really do dispatch...

szy21: Ah, I see. It should be easy to add dispatch if needed.

trontrytel: I might have to add another loop over levels for the precipitation dispatch. But maybe that would be ok? I'll think about it

src/parameterized_tendencies/microphysics/precipitation.jl
```diff
@@ -43,6 +44,7 @@ function compute_precipitation_cache!(Y, p, colidx, ::Microphysics0Moment, _)
     (; params, dt) = p
     (; ᶜts) = p.precomputed
     (; ᶜS_ρq_tot) = p.precipitation
```
szy21, line 46:
```suggestion
    (; ᶜS_ρq_tot, ᶜS_ρe_tot) = p.precipitation
```

szy21 (approved):
Thank you!

PR #2846, opened 2024-03-27
move get_cloud_fraction to precomputed quantities

szy21:
## Content

----
- [ ] I have read and checked the items on the review checklist.

PR #2874, opened 2024-04-04
Add cloud diagnostics per stage option

### LINKED ISSUE #2873: Improve performance of `set_cloud_fraction!`
`set_cloud_fraction!` is slow ([xref](https://github.com/CliMA/ClimaAtmos.jl/issues/2868#issuecomment-2036181342)) , and we need to improve its performance.

charleskawczynski:
Supersedes #2846. I figure we can merge this so that we can analyze the performance, and then strip out the `call_cloud_diagnostics_per_stage` option once we've closed #2873.


src/cache/cloud_fraction.jl
```diff
@@ -33,6 +33,17 @@ NVTX.@annotate function set_cloud_fraction!(
     (; ᶜts, ᶜmixing_length, ᶜcloud_fraction) = p.precomputed
     thermo_params = CAP.thermodynamics_params(params)
     if isnothing(turbconv_model)
+        if p.atmos.call_cloud_diagnostics_per_stage isa
```
szy21, line 36: We need to calculate these gradients as well if set_cloud_fraction is in the callback I think.

src/cache/cloud_fraction.jl
```diff
@@ -50,6 +61,17 @@ NVTX.@annotate function set_cloud_fraction!(
     (; ᶜts, ᶜp, ᶜmixing_length, ᶜcloud_fraction) = p.precomputed
     (; turbconv_model) = p.atmos
     if isnothing(turbconv_model)
+        if p.atmos.call_cloud_diagnostics_per_stage isa
```
szy21, line 64: same as before

szy21 (approved):
Thank you! I just have one comment.

PR #2877, opened 2024-04-05
Remove duplicate code in scaling plots, add debug prints

charleskawczynski:
This PR removes some duplicate code in the scaling plots, and adds some debug prints

### DIFF
--- a/post_processing/plot_gpu_scaling_utils.jl
+++ b/post_processing/plot_gpu_scaling_utils.jl
@@ -1,0 +1,37 @@
+using JLD2
+function get_jld2data(output_dir, job_id, s)
+    FT = Float64
+    nprocs_clima_atmos = Int[]
+    ncols_per_process = Int[]
+    walltime_clima_atmos = FT[]
+    found = false
+    for foldername in readdir(output_dir)
+        if occursin(job_id, foldername) && occursin(s, foldername)
+            nprocs_string = split(split(foldername, s)[end], "process")[1]
+            file = joinpath(
+                output_dir,
+                foldername,
+                "output_active",
+                "scaling_data_$(nprocs_string)_processes.jld2",
+            )
+            if !isfile(file)
+                @show readdir(output_dir)
+                @show readdir(dirname(file))
+            end
+            dict = load(file)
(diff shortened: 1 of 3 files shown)


post_processing/plot_gpu_scaling_utils.jl
```diff
@@ -0,0 +1,36 @@
...
+    walltime_clima_atmos = FT[]
+    found = false
+    for foldername in readdir(output_dir)
+        if occursin(job_id, foldername) && occursin(s, foldername)
+            nprocs_string = split(split(foldername, s)[end], "process")[1]
+            file = joinpath(
+                output_dir,
+                foldername,
+                "scaling_data_$(nprocs_string)_processes.jld2",
+            )
```
nefrathenrici, line 15:
```suggestion
            file = joinpath(
                output_dir,
                foldername,
                "output_active",
                "scaling_data_$(nprocs_string)_processes.jld2",
            )
```

I think this change will fix the new path from the OutputPathGenerator, see [here](https://buildkite.com/clima/climaatmos-target-gpu-simulations/builds/258#018ebf56-f2bc-47f5-b611-a38ee78ac817).

nefrathenrici (approved):
Thanks! The `@show` calls should be removed and the output path needs to be updated, but this is cleaner than before.

post_processing/plot_gpu_scaling_utils.jl
charleskawczynski: Perfect, thanks for launching that

charleskawczynski:
> Thanks! The @show calls should be removed and the output path needs to be updated, but this is cleaner than before.

I was actually thinking of leaving the `@show` calls, for debugging when this sort of thing happens.

Also, I want to leave the 

```julia
    @show nprocs_clima_atmos
    @show ncols_per_process
    @show walltime_clima_atmos
```
calls temporarily-- I'd like to also create a table in this job, but we can do that in a separate PR.

charleskawczynski:
Since CI is effectively hanging due to central, and this PR doesn't touch anything that is run in CI (only in our gpu target pipeline), I'm going to manually merge this so that we can test it out.

PR #2878, opened 2024-04-05
Widen gpu allocations buffer

charleskawczynski:
This PR widens the gpu allocations buffer, so that it fails less frequently.

### DIFF
--- a/perf/flame.jl
+++ b/perf/flame.jl
@@ -88,8 +88,10 @@ using Test
 # Threaded/gpu allocations are not deterministic, so let's add a buffer
 # TODO: remove buffer, and threaded tests, when
 #       threaded/unthreaded functions are unified
-buffer = if any(x -> occursin(x, job_id), ("threaded", "gpu"))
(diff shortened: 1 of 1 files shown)


Sbozzolo:
The job still failed. :(

Also, having a buffer that is 5 times larger than the target one seems to defy the purpose of this job.

charleskawczynski:
> Also, having a buffer that is 5 times larger than the target one seems to defy the purpose of this job.

Agreed, but the flakiness is a drag on CI. I think I'd prefer to remove the lower limit, and just add an info statement, like we used to have.

Sbozzolo:
If this test is barely indicative, let us remove it or have it soft fail. We can revisit it with julia 1.11.

charleskawczynski:
> If this test is barely indicative, let us remove it or have it soft fail. We can revisit it with julia 1.11.

The problem with soft fail is that it can then fail for other reasons, and CI will still pass. I'd prefer we just relax the limits, so that CI is no longer flakey. I'll bump it 10x, and remove the lower limit, for the gpu job.

PR #2882, opened 2024-04-06
update pi groups in entrainment and detrainment

szy21:
## Purpose 

Updates to the new pi groups in entrainment and detrainment closure and adds a bomex case that uses the pi group closure. It doesn't break although the results don't look good. I would like to merge it in first as it would be easier to test it further.

## Content

----

### DIFF
--- a/toml/prognostic_edmfx_bomex_explicit.toml
+++ b/toml/prognostic_edmfx_bomex_explicit.toml
@@ -1,0 +1,62 @@
+[C_E]
+value = 0.0176
+
+[EDMF_surface_area]
+value = 0.1
+
+[EDMF_min_area]
+value = 1.0e-5
+
+[EDMF_max_area]
+value = 0.5
+
+[entr_inv_tau]
+value = 0.0002
+
+[entr_coeff]
+value = 0.3
+
+[min_area_limiter_scale]
+value = 0
+
+[min_area_limiter_power]
+value = 10
+
+[detr_inv_tau]
+value = 0
+
+[detr_coeff]
+value = 0
+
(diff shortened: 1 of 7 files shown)


toml/prognostic_edmfx_bomex_explicit.toml
```diff
@@ -0,0 +1,62 @@
...
+value = 0
+
+[detr_massflux_vertdiv_coeff]
+value = 1
+
+[max_area_limiter_scale]
+value = 0
+
+[pressure_normalmode_drag_coeff]
+value = 20.0
```
szy21, line 44: This value is different from Costa's namelist because of differences in the nonhydrostatic pressure drag (using scale height in climaatmos but updraft top in tc).

trontrytel (approved):

toml/prognostic_edmfx_bomex_explicit.toml
```diff
@@ -0,0 +1,62 @@
+[C_E]
+value = 0.0176
+
+[EDMF_surface_area]
+value = 0.1
+
+[EDMF_min_area]
+value = 1.0e-5
```
costachris, line 8: I used min_area = 1e-10 but don't think CA is clipping area anyway (?)

costachris (approved):
Perfect

toml/prognostic_edmfx_bomex_explicit.toml
szy21: Yeah, the min_area here is different from the one in TC. This is used to regulate divide_by_rhoa. It shouldn't matter.

PR #2885, opened 2024-04-08
Remove old code (coupler compatibility)

akshaysridhar:
Remove unused lines. 
cc @LenkaNovak @szy21


szy21:
This was needed to run diagnostic edmf in the coupler. Do we not need it anymore?

akshaysridhar:
> This was needed to run diagnostic edmf in the coupler. Do we not need it anymore?

Yeah I see that now - I wanted to test this by removing the old block (and it's clear from the `ci` outputs that we still need it).  I'll start an issue to track that and close this PR.

PR #2886, opened 2024-04-08
Try using reformulated timestepping scheme

dennisYatunin:
## Purpose 

This PR switches ClimaAtmos to the reformulated timestepping scheme introduced in CliMA/ClimaTimeSteppers.jl#264.

----


szy21:
I think we don't need this anymore? Feel free to reopen if it's not the case.

PR #2892, opened 2024-04-09
Rm code dup, rm some prints, tabulate scaling

charleskawczynski:
This PR
 - Removes some `@show` calls that I didn't intend to leave
 - Tabulates some of the scaling results
 - Removes some code duplication in the scaling results


charleskawczynski:
> The tables are great for exact numbers!
Agreed-- very copy-paste friendly, too.

> The only change is the [artifact](https://github.com/CliMA/ClimaAtmos.jl/blob/a7b463a6421d09b2a56b7676993efc3a9326276a/.buildkite/gpu_pipeline/pipeline.yml#L164) [paths](https://github.com/CliMA/ClimaAtmos.jl/blob/a7b463a6421d09b2a56b7676993efc3a9326276a/.buildkite/gpu_pipeline/pipeline.yml#L228) for the plotting jobs, they should not have `output_active` in the path since these jobs don't use `OutputPathGenerator`

Ah, great point. Thank you!

PR #2895, opened 2024-04-11
Integrate calibration pipeline into atmos

nefrathenrici:
This PR adds the framework for reproducible calibration experiments, starting with one example experiment.

`calibration/README.md` provides an overview. Most of the actual code is in [ClimaCalibrate](https://github.com/CliMA/ClimaCalibrate.jl)

## Content 

`calibration/model_interface.jl`: Model interface hooks for CalibrateAtmos. This will be used in every atmos calibration. 
`test/calibration_interface.jl`: Basic tests for the model interface file.
`calibration/experiments`: Folder containing subfolders of calibration experiments. Currently only contains `sphere_held_suarez_rhoe_equilmoist`

sphere_held_suarez_rhoe_equilmoist/
- `experiment_config.yml`: Stores filepaths for calibration artifacts and other configuration data
- `model_config.yml`: Atmos yaml config for the calibration
- `pipeline.sbatch`: Script to run calibration experiment on central
- `observation_map.jl`: Code to process model output into observation space
- `postprocessing.jl`: Plotting script
- `prior.toml`: Prior distribution file


test/calibration_interface.jl
```diff
@@ -0,0 +1,30 @@
...
+
+@testset "Atmos Configuration" begin
+    @test parsed_args["moist"] == "equil"
+    @test parsed_args["toml"] == [file_path]
+    @test parsed_args["output_dir"] == member_path
+    @test parsed_args["restart_file"] ==
+          "/groups/esm/ClimaArtifacts/artifacts/atmos_held_suarez_obs/day200.0.hdf5"
+end
+
+rm(file_path)
```
Sbozzolo, line 30: Use temporary folders instead of relying on creating a new folder and removing it. The folder will not be cleaned in case of failing tests.

test/calibration_interface.jl
```diff
@@ -0,0 +1,30 @@
...
+
+atmos_config = ClimaCalibrate.set_up_forward_model(1, 1, experiment_dir)
+(; parsed_args) = atmos_config
+
+@testset "Atmos Configuration" begin
+    @test parsed_args["moist"] == "equil"
+    @test parsed_args["toml"] == [file_path]
+    @test parsed_args["output_dir"] == member_path
+    @test parsed_args["restart_file"] ==
+          "/groups/esm/ClimaArtifacts/artifacts/atmos_held_suarez_obs/day200.0.hdf5"
```
Sbozzolo, line 27: You should use Artifacts or ClimaArtifacts instead of hardcoding the path.

This test has to be reproducible even outside of the Caltech cluster

Sbozzolo:
I am not too familiar with the calibration pipeline, so it will take me a little bit of time to understand and properly review this PR.

calibration/experiments/sphere_held_suarez_rhoe_equilmoist/Project.toml
```diff
@@ -0,0 +1,15 @@
...
+ClimaUtilities = "b3f4f4ca-9299-4f7f-bd9b-81e1242a7513"
+Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
+EnsembleKalmanProcesses = "aa8a2aa5-91d8-4396-bcef-d4f2ec43552d"
+JLD2 = "033835bb-8acc-5ee8-8aae-3f567f8a3819"
+NetCDF = "30363a11-5582-574a-97bb-aa9a979735b9"
+TOML = "fa267f1f-6049-4f14-aa54-33bafae1ed76"
+YAML = "ddb6d928-2868-570f-bddf-ab3f9cf99eb6"
+
+[compat]
+ClimaAtmos = "=0.24.0"
```
Sbozzolo, line 15: If ClimaAtmos is pinned to a specific version, why do we need to have a job in buildkite?

calibration/experiments/sphere_held_suarez_rhoe_equilmoist/Project.toml
```diff
@@ -0,0 +1,15 @@
+[deps]
+CairoMakie = "13f3f980-e62b-5c42-98c6-ff1f3baf88f0"
+ClimaAnalysis = "29b5916a-a76c-4e73-9657-3c8fd22e65e6"
+ClimaAtmos = "b2c96348-7fb7-4fe0-8da9-78d88439e717"
+ClimaCalibrate = "4347a170-ebd6-470c-89d3-5c705c0cacc2"
+ClimaUtilities = "b3f4f4ca-9299-4f7f-bd9b-81e1242a7513"
+Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
+EnsembleKalmanProcesses = "aa8a2aa5-91d8-4396-bcef-d4f2ec43552d"
+JLD2 = "033835bb-8acc-5ee8-8aae-3f567f8a3819"
+NetCDF = "30363a11-5582-574a-97bb-aa9a979735b9"
```
Sbozzolo, line 10: Is NetCDF needed?

calibration/experiments/sphere_held_suarez_rhoe_equilmoist/observation_map.jl
```diff
@@ -0,0 +1,35 @@
...
+    for m in 1:ensemble_size
+        member_path =
+            TOMLInterface.path_to_ensemble_member(output_dir, iteration, m)
+        simdir = SimDir(joinpath(member_path, "output_active"))
+        try
+            G_ensemble[:, m] .= process_member_data(simdir)
+        catch err
+            @info "Error during observation map for ensemble member $m" err
+            G_ensemble[:, m] .= NaN
+        end
```
Sbozzolo, line 19: What are you trying to `try-catch` here?

calibration/experiments/sphere_held_suarez_rhoe_equilmoist/observation_map.jl
```diff
@@ -0,0 +1,35 @@
+import EnsembleKalmanProcesses: TOMLInterface
+import ClimaCalibrate: observation_map, ExperimentConfig
+using ClimaAnalysis
+
+const config = ExperimentConfig(@__DIR__)
+function observation_map(iteration)
+    (; ensemble_size, output_dir) = config
+    dims = 1
```
Sbozzolo, line 8: Maybe `dims` can have a more descrpitive name. `dims` of what?

calibration/experiments/sphere_held_suarez_rhoe_equilmoist/postprocessing.jl
```diff
@@ -0,0 +1,109 @@
...
+    spread_vec = Float64[]
+    for ensemble in u_vec
+        ensemble_error = 0
+        ensemble_spread = 0
+        ensemble_mean = mean(ensemble)
+        for i in ensemble
+            ensemble_error += abs(i - theta_star)^2
+            ensemble_spread += abs(i - ensemble_mean)^2
+
+        end
```
Sbozzolo, line 30:
```suggestion
            ensemble_error += abs(i - theta_star)^2
            ensemble_spread += abs(i - ensemble_mean)^2
        end
```

calibration/experiments/sphere_held_suarez_rhoe_equilmoist/postprocessing.jl
```diff
@@ -0,0 +1,109 @@
...
+    theta_star = 65.0,
+    output = joinpath("output", "sphere_held_suarez_rhoe_equilmoist"),
+)
+    u_vec = EKP.get_u(eki)
+
+    error_vec = Float64[]
+    spread_vec = Float64[]
+    for ensemble in u_vec
+        ensemble_error = 0
+        ensemble_spread = 0
```
Sbozzolo, line 24:
```suggestion
        ensemble_error = 0.0
        ensemble_spread = 0.0
```

calibration/model_interface.jl
```diff
@@ -0,0 +1,55 @@
...
+using ClimaUtilities.ClimaArtifacts
+
+import ClimaCalibrate:
+    set_up_forward_model, run_forward_model, path_to_ensemble_member
+
+"""
+    set_up_forward_model(member, iteration, experiment_dir::AbstractString)
+
+Returns an AtmosConfig object for the given member and iteration.
+If given an experiment id string, it will load the config from the corresponding YAML file.
```
Sbozzolo, line 12: Is this implemented?

calibration/model_interface.jl
```diff
@@ -0,0 +1,55 @@
...
+This assumes that the config dictionary has `output_dir` and `restart_file` keys.
+"""
+function set_up_forward_model(member, iteration, experiment_dir::AbstractString)
+    config_dict = YAML.load_file(joinpath(experiment_dir, "model_config.yml"))
+    # If relative path, append experiment dir
+    if haskey(config_dict, "restart_file") &&
+       !isabspath(config_dict["restart_file"])
+        config_dict["restart_file"] =
+            joinpath(experiment_dir, config_dict["restart_file"])
+    end
```
Sbozzolo, line 23: Shouldn't this be handled internally by Atmos?

Sbozzolo:
What is the ultimate intent of this PR? Is it to test that things work? To teach people how to do calibration? Other reasons?

calibration/experiments/sphere_held_suarez_rhoe_equilmoist/Project.toml
charleskawczynski: Yeah, we don't want to pin it if we're checking in the manifest. Alternatively, if we don't check in the manifest, then we could leave it pinned, but that's less reliably reproducible.

calibration/experiments/sphere_held_suarez_rhoe_equilmoist/observation_map.jl
```diff
@@ -0,0 +1,32 @@
...
+        simdir = SimDir(joinpath(member_path, "output_active"))
+        G_ensemble[:, m] .= process_member_data(simdir)
+    end
+    return G_ensemble
+end
+
+const meters = 1.0
+const days = 86400.0
+# Cut off first 120 day to get equilibrium, take second level slice
+function process_member_data(simdir)
```
charleskawczynski, line 22: What is the type of `simdir` here?

calibration/experiments/sphere_held_suarez_rhoe_equilmoist/observation_map.jl
```diff
@@ -0,0 +1,32 @@
...
+    isempty(simdir.vars) && return  NaN
+
+    ta = get(simdir; short_name = "ta", reduction = "average", period = "60d")
+    zonal_avg_temp_observations =
+        slice(average_lat(average_lon(ta)), z = 242meters)
+    sample_observation = Vector{Float64}(undef, 1)
+    sample_observation .=
+        slice(zonal_avg_temp_observations, time = 240days).data
```
charleskawczynski, line 30:
```suggestion
    sample_observation = slice(zonal_avg_temp_observations, time = 240days).data
```
Does this not work? Where is `slice` coming from? Can we qualify these methods?

calibration/experiments/sphere_held_suarez_rhoe_equilmoist/postprocessing.jl
```diff
@@ -0,0 +1,106 @@
...
+        ensemble_error /= length(ensemble)
+        ensemble_spread /= length(ensemble)
+
+        push!(error_vec, ensemble_error)
+        push!(spread_vec, ensemble_spread)
+    end
+
+    phi_vec = transform_unconstrained_to_constrained(prior, u_vec)
+    u_series = [getindex.(u_vec, i) for i in 1:10]
+    phi_series = [getindex.(phi_vec, i) for i in 1:10]
```
charleskawczynski, line 36:
```suggestion
    u_series = map(i->getindex.(u_vec, i), 1:10)
    phi_series = map(i->getindex.(phi_vec, i), 1:10)
```
Do we need the `.` in `getindex.(u_vec, i)`?

calibration/experiments/sphere_held_suarez_rhoe_equilmoist/postprocessing.jl
```diff
@@ -0,0 +1,106 @@
...
+import Statistics: mean
+import CairoMakie: Makie
+import ClimaCalibrate
+
+function convergence_plot(
+    eki,
+    prior;
+    theta_star = 65.0,
+    output = joinpath("output", "sphere_held_suarez_rhoe_equilmoist"),
+)
```
charleskawczynski, line 14: Can we add type parameters here so we know what types are coming in?

test/calibration_interface.jl
```diff
@@ -0,0 +1,31 @@
...
+
+member_path = joinpath(
+    "output",
+    "sphere_held_suarez_rhoe_equilmoist",
+    "iteration_001",
+    "member_001",
+)
+file_path = joinpath(member_path, "parameters.toml")
+mkpath(dirname(file_path))
+touch(file_path)
```
charleskawczynski, line 18: Please use `mktempdir` instead, and remove the `rm(file_path)` below

calibration/experiments/sphere_held_suarez_rhoe_equilmoist/Project.toml
nefrathenrici: Since this is experiment a perfect model calibration, the observations will need to be newly generated when the ClimaAtmos version gets updated. I'll keep the pin and remove the manifest.

nefrathenrici: For the buildkite test, I thought we should test the `model_interface.jl` is working as intended since there will no doubt be interface changes for Atmos in the future

calibration/experiments/sphere_held_suarez_rhoe_equilmoist/observation_map.jl
nefrathenrici: This works. `slice` and other methods come from ClimaAnalysis, I will qualify them.

calibration/experiments/sphere_held_suarez_rhoe_equilmoist/postprocessing.jl
nefrathenrici: The broadcast is needed, but this snippet is confusing. It is "transposing" the vector of vectors `u_vec`.

Something like this is clearer: `eachcol(reduce(vcat, u_vec))`

calibration/experiments/sphere_held_suarez_rhoe_equilmoist/observation_map.jl
nefrathenrici: It is a `ClimaAnalysis.SimDir`

nefrathenrici:
> What is the ultimate intent of this PR? Is it to test that things work? To teach people how to do calibration? Other reasons?

The ultimate intent is to provide the framework for future Atmos calibrations, such as #3075. The secondary goal is to provide an example of a documented, reproducible calibration example via `sphere_held_suarez_rhoe_equilmoist`

test/calibration_interface.jl
```diff
@@ -0,0 +1,40 @@
+#= Tests for ensuring that the calibration/model interface sets AtmosConfig correctly.
```
charleskawczynski, line 1: How long does this take to run? Should it be included in `test/runtests.jl` or on buildkite?

nefrathenrici: This is quick, the only issue is that ClimaCalibrate is unregistered so it needs to be added via `import Pkg; Pkg.add(url="https://github.com/CliMA/ClimaCalibrate.jl.git")` in the test instantiation.

charleskawczynski: Ok, in that case maybe we can add it as soft-fail, until ClimaCalibrate is registered.

charleskawczynski (approved):
Overall looks good to me. I figure that this will likely go through a few iterations, but this looks like a good first iteration.

Sbozzolo:
> > What is the ultimate intent of this PR? Is it to test that things work? To teach people how to do calibration? Other reasons?
> 
> The ultimate intent is to provide the framework for future Atmos calibrations, such as #3075. The secondary goal is to provide an example of a documented, reproducible calibration example via `sphere_held_suarez_rhoe_equilmoist`

Then maybe this could be turned into a curated tutorial + the minimal tests required to ensure that pipeline can be run in the future?

This code, as it is, can exclusively run on the Caltech cluster and can only be meanfigully run with a specific version of ClimaAtmos. Also, many components of it cannot be easily tested, and the files themeselves will likely never be used. So, at the end of the value, the main value is showing how a simple calibration can be set up. In this case, we can double down on that aspect and turn this into a full tutorial.

If the intent if for these files to grow into the actual calibration pipeline, maybe we could make them a Julia subpackage (as our ClimaCore libs) and have proper tests for this package so that we can ensure we can build upon the code and expand it.

nefrathenrici:
> Then maybe this could be turned into a curated tutorial + the minimal tests required to ensure that pipeline can be run in the future?
> 
> This code, as it is, can exclusively run on the Caltech cluster and can only be meanfigully run with a specific version of ClimaAtmos. Also, many components of it cannot be easily tested, and the files themeselves will likely never be used. So, at the end of the value, the main value is showing how a simple calibration can be set up. In this case, we can double down on that aspect and turn this into a full tutorial.

I agree that this will probably not be run in a meaningful way again, but I don't see what is wrong with including the details/artifacts of the experiment in this PR to reproduce it if needed. To show how to set up a calibration, I have written a [tutorial](https://clima.github.io/ClimaCalibrate.jl/dev/atmos_setup_guide/) for ClimaAtmos specifically, that can be moved into the ClimaAtmos docs and cleaned up a bit.

> If the intent if for these files to grow into the actual calibration pipeline, maybe we could make them a Julia subpackage (as our ClimaCore libs) and have proper tests for this package so that we can ensure we can build upon the code and expand it.

My goal is to grow this into multiple experiments that run using the `model_interface.jl` with expanded tests. This will eventually include the more complex/realistic experiments. I want to add an e2e test for a quicker calibration (<1hour) that gets run on multiple backends. What are the advantages of turning this into a subpackage? I am certainly open to it, but don't understand why it is needed.

Sbozzolo:
My question is essentially on what needs to be maintained/continuously tested.

The two options I see are:
- this is a tutorial/example: it's mostly frozen in time, with a specific version of Atmos and ClimaCalibrate, it teaches people how to run stuff but it is not really meant to be run and we don't need to be too concerned with keeping having bleeding edge (as long as the content is still relevant and meanigful). This is perfectly valid and valuable.
- this is the beginning of a module that will be eventually contain the full calibration stack. In this case, we want to run it as much as possible and ensure that it can keep growing safely. This module needs ClimaAtmos as a dependency (and not the other way around), so it feels natural to be a subpackage, where it be tested separately and independently. This is also perfectly valid and valuable.

charleskawczynski:
> this is the beginning of a module that will be eventually contain the full calibration stack.

> My question is essentially on what needs to be maintained/continuously tested.
> 
> The two options I see are:
> 
> * this is a tutorial/example: it's mostly frozen in time, with a specific version of Atmos and ClimaCalibrate, it teaches people how to run stuff but it is not really meant to be run and we don't need to be too concerned with keeping having bleeding edge (as long as the content is still relevant and meanigful). This is perfectly valid and valuable.
> * this is the beginning of a module that will be eventually contain the full calibration stack. In this case, we want to run it as much as possible and ensure that it can keep growing safely. This module needs ClimaAtmos as a dependency (and not the other way around), so it feels natural to be a subpackage, where it be tested separately and independently. This is also perfectly valid and valuable.

I'm all about breaking things up into small packages, but what exactly will go into this package? Most of the content in this PR is ClimaAtmos-specific, and there already exists `ClimaCalibration`. That's the modular part of the calibration stack AFAIU.

Sbozzolo:
> > this is the beginning of a module that will be eventually contain the full calibration stack.
> 
> > My question is essentially on what needs to be maintained/continuously tested.
> > The two options I see are:
> > 
> > * this is a tutorial/example: it's mostly frozen in time, with a specific version of Atmos and ClimaCalibrate, it teaches people how to run stuff but it is not really meant to be run and we don't need to be too concerned with keeping having bleeding edge (as long as the content is still relevant and meanigful). This is perfectly valid and valuable.
> > * this is the beginning of a module that will be eventually contain the full calibration stack. In this case, we want to run it as much as possible and ensure that it can keep growing safely. This module needs ClimaAtmos as a dependency (and not the other way around), so it feels natural to be a subpackage, where it be tested separately and independently. This is also perfectly valid and valuable.
> 
> I'm all about breaking things up into small packages, but what exactly will go into this package? Most of the content in this PR is ClimaAtmos-specific, and there already exists `ClimaCalibration`. That's the modular part of the calibration stack AFAIU.

It's not a very satisfying answer, but I'd say: everything that needs to be maintained and updated (including the list dependencies). We don't necessarily need a subpackage, but I think it'd be good to identify what is the "example" part of this PR and what is the "framework for calibration" component and ensure that this second part is tested and maintained and the first one is perfectly reproducible (ie, with frozen versions). 

(Also ClimaAtmos should not depend on the `atmos_held_suarez_obs` artifact.)

charleskawczynski:
I’d argue that it may be too early to know what parts of this could be abstracted out.

that said, I agree pretty strongly about the artifacts. @nefrathenrici, can we split this artifact out?

nefrathenrici:
> It's not a very satisfying answer, but I'd say: everything that needs to be maintained and updated (including the list dependencies). We don't necessarily need a subpackage, but I think it'd be good to identify what is the "example" part of this PR and what is the "framework for calibration" component and ensure that this second part is tested and maintained and the first one is perfectly reproducible (ie, with frozen versions).

The "framework for calibration" component is essentially just `calibration/model_interface.jl` and ClimaCalibrate. ClimaCalibrate is already reasonably well tested, but within atmos the `model_interface.jl` needs an end-to-end test that runs a simple calibration and checks the output as part of the buildkite tests. I am not sure what other tests are required, for example if we start running calibrations on the Clima or NCAR GPUs, but I would like to discuss this further.

The "example" is everything in `calibration/experiments/sphere_held_suarez_rhoe_equilmoist`. The ClimaCalibrate docs already contain a [setup guide](https://clima.github.io/ClimaCalibrate.jl/dev/atmos_setup_guide/) for atmos calibrations, using this calibration as an example. It needs a bit of updating, but I could move it here.

> that said, I agree pretty strongly about the artifacts. @nefrathenrici, can we split this artifact out?

Yes, I can remove it and ensure that the DEPOT_PATH is set for the caltech cluster for the example calibration.

Sbozzolo:
> The "example" is everything in `calibration/experiments/sphere_held_suarez_rhoe_equilmoist`. The ClimaCalibrate docs already contain a [setup guide](https://clima.github.io/ClimaCalibrate.jl/dev/atmos_setup_guide/) for atmos calibrations, using this calibration as an example. It needs a bit of updating, but I could move it here.

I just want to make sure that the intended purpose of this new code is clear, along with its implied maitenance burden. For example, as we release ClimaAtmos 0.25, is this something that we need to update? Or, in other words, as we develop Atmos, how concerned should we be with this particular folder?

Here are possible scenarios:
1. the code added is for reproducibility purposes. Dependencies are frozen and nothing has to be maintained. The readme seems to imply that this is the case, but then the model_interface should stay frozen (if the model_interfaced diverged from ClimaAtmos 0.24, we'd lose reproducibility).  
2. the code added is for educational purposes. In this case, I feel that the better place where to place most of this code is alongside the relevant tutorial, where the code can be run with an updating version of ClimaCalibrate but a fixed version of Atmos.
3. the code added will grow in the full calibration pipeline. In this case, we have to keep everything updated and working.

Maybe what we want is turning `model_interface.jl` into an extension for ClimaAtmos (with ClimaCalibrate), or a subpackage. In this way, a version of ClimaAtmos (or a specific version of the subpackage) implies a version of model_interface so that we can ensure reproducibility.

> > that said, I agree pretty strongly about the artifacts. @nefrathenrici, can we split this artifact out?
> 
> Yes, I can remove it and ensure that the DEPOT_PATH is set for the caltech cluster for the example calibration.

This should be done using the Artifacts system. You should add the artifact to the Artifacts.toml of your leaf Julia project and not to the one for atmos itself.

Project.toml
```diff
@@ -52,7 +52,7 @@ AtmosphericProfilesLibrary = "0.1"
 ClimaComms = "0.6.2"
 ClimaCore = "0.14.6"
 ClimaDiagnostics = "0.2"
-ClimaParams = "0.10.4"
+ClimaParams = "0.10.8"
```
nefrathenrici, line 55: This is needed for the most recent version of RRTMGP

.github/workflows/ci.yml
```diff
@@ -18,7 +18,7 @@ jobs:
   test:
     name: ci ${{ matrix.version }} - ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
-    timeout-minutes: 30
+    timeout-minutes: 45
```
nefrathenrici, line 21: The new test in the GHA CI takes ~15 minutes, and the CI takes ~35 minutes in total

Sbozzolo: Maybe we can split this off to a second action so that it can run in parallel?

Project.toml
Sbozzolo: If this is the case, it should be RRTMGP that defines this compat, not ClimaAtmos

calibration/experiments/sphere_held_suarez_rhoe_equilmoist/Manifest.toml
```diff
@@ -0,0 +1,2875 @@
...
+uuid = "b2c96348-7fb7-4fe0-8da9-78d88439e717"
+version = "0.24.0"
+
+[[deps.ClimaCalibrate]]
+deps = ["Distributions", "EnsembleKalmanProcesses", "JLD2", "Random", "TOML", "YAML"]
+git-tree-sha1 = "b281e4e1e6713cf5f26088837107842a64cbc987"
+repo-rev = "main"
+repo-url = "https://github.com/CliMA/ClimaCalibrate.jl.git"
+uuid = "4347a170-ebd6-470c-89d3-5c705c0cacc2"
+version = "0.1.0"
```
Sbozzolo, line 358: Can you use the released version?

calibration/experiments/sphere_held_suarez_rhoe_equilmoist/model_config.yml
```diff
@@ -0,0 +1,15 @@
+dt: 500secs
+t_end: 560days
+ode_algo: ARS343
+moist: equil 
+forcing: held_suarez
+precip_model: 0M
+job_id: sphere_held_suarez_rhoe_equilmoist
+output_dir: output/sphere_held_suarez_rhoe_equilmoist
+restart_file: /groups/esm/ClimaArtifacts/artifacts/atmos_held_suarez_obs/day200.0.hdf5
```
Sbozzolo, line 9: This will be a problem everywhere but the caltech cluster

Sbozzolo: Maybe add instructions on how to fill this field properly if one wants to reproduce this

calibration/experiments/sphere_held_suarez_rhoe_equilmoist/postprocessing.jl
```diff
@@ -0,0 +1,105 @@
...
+        for i in ensemble
+            ensemble_error += abs(i - theta_star)^2
+            ensemble_spread += abs(i - ensemble_mean)^2
+        end
+        ensemble_error /= length(ensemble)
+        ensemble_spread /= length(ensemble)
+
+        push!(error_vec, ensemble_error)
+        push!(spread_vec, ensemble_spread)
+    end
```
Sbozzolo, line 31:
```suggestion
error_vec = Float64[]
spread_vec = Float64[]
for ensemble in u_vec
    ensemble_mean = mean(ensemble)
    ensemble_error = mean((abs.(ensemble .- theta_star)).^2)
    ensemble_spread = mean((abs.(ensemble .- ensemble_mean)).^2)

    push!(error_vec, ensemble_error)
    push!(spread_vec, ensemble_spread)
end
```
Right?

calibration/model_interface.jl
```diff
@@ -0,0 +1,65 @@
...
+    set_up_forward_model(member, iteration, config_dict::AbstractDict)
+
+Return an AtmosConfig object for the given member and iteration.
+
+Turns off default diagnostics and sets the TOML parameter file to the member's path.
+This assumes that the  config dictionary has an `output_dir` key.
+"""
+function set_up_forward_model(member, iteration, ::ExperimentConfig)
+    # Assume experiment_dir is project dir
+    experiment_dir = dirname(Base.active_project())
```
Sbozzolo, line 24: There is no reason to assume that the experiment_dir is the same as the active project dir. Actually, in many case it is preferable for the two not to be the same (e.g., you are on a cluster with limited quota in $HOME, so you want to run on a scratch space)

calibration/test/Manifest.toml
Sbozzolo, line 1: We should not check this Manifest in. Tests should be with the most recent versions of the dependencies

calibration/test/e2e_test.jl
```diff
@@ -0,0 +1,169 @@
...
+    # Spread should be heavily decreased as particles have converged
+    @test last(spread) / first(spread) < 0.1
+    # Parameter should be close to true value
+    @test mean(last(params)) ≈ astronomical_unit rtol = 0.02
+end
+
+# Caltech HPC backend
+backend = CAL.get_backend()
+@info "Running Calibration E2E test on $backend"
+if backend <: CAL.SlurmBackend
```
Sbozzolo, line 94:
```suggestion
if backend isa CAL.SlurmBackend
```

calibration/test/e2e_test.jl
```diff
@@ -0,0 +1,169 @@
...
+    end
+end
+
+# Debug plots
+function scatter_plot(julia_eki::EKP.EnsembleKalmanProcess)
+    f = CairoMakie.Figure(resolution = (800, 600))
+    ax = CairoMakie.Axis(
+        f[1, 1],
+        ylabel = "Absolute Parameter Value",
+        xlabel = "TOA Radiative SW Flux 30day average",
```
Sbozzolo, line 130: I think that this label is incorrect

calibration/experiments/sphere_held_suarez_rhoe_equilmoist/postprocessing.jl
charleskawczynski: Better still:
```julia
    meanabsdiff²(x, e) = sum(i-> abs(i - x)^2, e)/length(e)
    error_vec = map(ensemble -> meanabsdiff²(theta_star, ensemble), u_vec)
    error_vec = map(ensemble -> meanabsdiff²(mean(ensemble), ensemble), u_vec)
```
?

Project.toml
nefrathenrici: My mistake, the compat should change in Atmos. ClimaParams v0.10.8 is required to run radiation within Atmos, RRTMGP can run with older versions of CP.

calibration/test/e2e_test.jl
nefrathenrici: I believe `<:` is required because the backend is the struct itself (`CaltechHPC`), not an instance of the struct (`CaltechHPCBackend()`).

charleskawczynski: Probably better to use `backend <: CAL.SlurmBackend` if `backend` is a type, to not assume how `backend` is constructed.

calibration/test/e2e_test.jl
nefrathenrici: Could you clarify this? It looks correct to me

Sbozzolo: Yes, you are right, I thought this was something else.

.github/workflows/calibration_test.yml
```diff
@@ -0,0 +1,22 @@
...
+  pull_request:
+  merge_group:
+
+jobs:
+  test:
+    name: ClimaCalibrate E2E Test
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      - uses: julia-actions/cache@v2
```
Sbozzolo, line 14: If you add `julia-actions/cache`, you need to add the permission to read/write the cache. 

Also, it's a good idea to add a concurrency group so that new runs kill previous ones.

```
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: true

# Needed to allow julia-actions/cache to delete old caches that it has created
permissions:
  actions: write
  contents: read
```

.github/workflows/ci.yml
```diff
@@ -38,6 +38,10 @@ jobs:
           version: ${{ matrix.version }}
           arch: ${{ matrix.arch }}
       - uses: julia-actions/julia-buildpkg@v1
+      - name: Calibration E2E Test
+        run: |
+          julia --project=calibration/test -e 'using Pkg; Pkg.instantiate(;verbose=true)'
+          julia --project=calibration/test calibration/test/e2e_test.jl
```
Sbozzolo, line 44: Is this still needed now that you moved to a separate workflow?

calibration/experiments/sphere_held_suarez_rhoe_equilmoist/model_interface.jl
```diff
@@ -0,0 +1,54 @@
...
+    run_forward_model(atmos_config::CA.AtmosConfig)
+
+Run the atmosphere model with the given an AtmosConfig object.
+Currently only has basic error handling.
+"""
+function run_forward_model(atmos_config::CA.AtmosConfig)
+    simulation = CA.get_simulation(atmos_config)
+    sol_res = CA.solve_atmos!(simulation)
+    if sol_res.ret_code == :simulation_crashed
+        !isnothing(sol_res.sol) && sol_res.sol .= eltype(sol_res.sol)(NaN)
```
Sbozzolo, line 49:
```suggestion
        !isnothing(sol_res.sol) && sol_res.sol .= NaN
```
I think that this should work because the type cast is implicit when using `.`

calibration/experiments/sphere_held_suarez_rhoe_equilmoist/pipeline.sbatch
```diff
@@ -0,0 +1,33 @@
...
+# Load data and configurations
+n_iterations = 2
+ensemble_size = 10
+observations = JLD2.load_object(joinpath(artifact_path, "obs_mean.jld2"))
+noise = JLD2.load_object(joinpath(artifact_path, "obs_noise_cov.jld2"))
+prior = get_prior(joinpath(experiment_dir, "prior.toml"))
+output_dir = joinpath("output", "sphere_held_suarez_rhoe_equilmoist")
+
+experiment_config = ExperimentConfig(; n_iterations, ensemble_size, 
+                                    observations, noise, output_dir, prior)
```
Sbozzolo, line 25:
```suggestion
experiment_config = ExperimentConfig(; n_iterations, ensemble_size, 
                                       observations, noise, output_dir, prior)
```

calibration/test/Manifest.toml
Sbozzolo, line 1: I think we should not include this Manifest file.

Maintaining Manifest files is a chore, it introduces a lot of noises in the commits/git history, and does not bring strong benefits when the code on GitHub actions (as opposed to buildkite, where we instantiate once and run lots of jobs in parallel). I don't think we need this test to be reproducible, what I think we need more is for this test to identify problems a early as possible)

calibration/test/Project.toml
```diff
@@ -0,0 +1,12 @@
...
+CairoMakie = "13f3f980-e62b-5c42-98c6-ff1f3baf88f0"
+ClimaAnalysis = "29b5916a-a76c-4e73-9657-3c8fd22e65e6"
+ClimaAtmos = "b2c96348-7fb7-4fe0-8da9-78d88439e717"
+ClimaCalibrate = "4347a170-ebd6-470c-89d3-5c705c0cacc2"
+ClimaComms = "3a4d1b5c-c61d-41fd-a00a-5873ba7a1b0d"
+ClimaUtilities = "b3f4f4ca-9299-4f7f-bd9b-81e1242a7513"
+Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
+EnsembleKalmanProcesses = "aa8a2aa5-91d8-4396-bcef-d4f2ec43552d"
+JLD2 = "033835bb-8acc-5ee8-8aae-3f567f8a3819"
+MPI = "da04e1cc-30fd-572f-bb4f-1f8673147195"
```
Sbozzolo, line 11: Is `MPI` needed?

calibration/test/Manifest.toml
nefrathenrici: I agree, we don't need to closely track all of the package versions for this project.

calibration/test/Project.toml
nefrathenrici: Ah, MPI is needed for `ClimaComms.@import_required_backends` 

see [here](https://buildkite.com/clima/climaatmos-ci/builds/19611#01907adc-f092-4e93-b3ad-8004f862856e)

Sbozzolo (approved):

calibration/experiments/sphere_held_suarez_rhoe_equilmoist/prior.toml
```diff
@@ -0,0 +1,5 @@
+["equator_pole_temperature_gradient_wet"]
+prior = "Parameterized(Normal(4.779568,0.31223328))"
+constraint = "[bounded_below(0)]"
+# EKP.ParameterDistributions.constrained_gaussian("name", 125, 40, 0, Inf)
```
odunbar, line 4: quick note - we have the funtionality to create a `constrained_gaussian` directly in EKP in v1.1.5 I thought so why is this still done with in the unconstrained format?

nefrathenrici: I think I created this prior.toml before adding the `constrained_gaussian` parser. I'll update it now, thanks!

PR #2898, opened 2024-04-11
make edmf boundary condition dependent on surface area fraction

szy21:
## Purpose 

Following TurbulentConvection. This will be needed once we move to prognostic surface area fraction. Note this will change MSE of diagnostic edmf as percentile_bounds_mean_norm(1-0.1, 1) is not exactly 1.75, which is what we currently use for the coefficient.

## Content

----


trontrytel (approved):
lgtm!

szy21:
@charleskawczynski Could you take a look at why invalidations failed for this PR?

charleskawczynski (approved):
The change in invalidations is mysterious. Don't worry about it for this PR.

PR #2907, opened 2024-04-15
Change cp to cv in large_scale_advection.jl

szy21:
## Purpose 

I think this is more correct. It doesn't seem to have too much effect on the result.

## Content

----

### DIFF
--- a/src/prognostic_equations/forcing/large_scale_advection.jl
+++ b/src/prognostic_equations/forcing/large_scale_advection.jl
@@ -45,7 +45,7 @@ function large_scale_advection_tendency!(
     #     (hv = cv_v * (ᶜT - T_0) + Lv_0 - R_v * T_0)
     @. Yₜ.c.ρe_tot[colidx] +=
         Y.c.ρ[colidx] * (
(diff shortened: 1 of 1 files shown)


src/prognostic_equations/forcing/large_scale_advection.jl
```diff
@@ -45,7 +45,7 @@ function large_scale_advection_tendency!(
     #     (hv = cv_v * (ᶜT - T_0) + Lv_0 - R_v * T_0)
     @. Yₜ.c.ρe_tot[colidx] +=
         Y.c.ρ[colidx] * (
-            TD.cp_m(thermo_params, ᶜts) * ᶜdTdt_hadv +
+            TD.cv_m(thermo_params, ᶜts) * ᶜdTdt_hadv +
```
szy21, line 48: @charleskawczynski Do you remember if there is any reason why it is cp_m here? If not I think it is probably a typo.

szy21:
The results barely change so I'll go ahead and merge this in.

src/prognostic_equations/forcing/large_scale_advection.jl
charleskawczynski: I don’t recall

PR #2908, opened 2024-04-16
set netcdf_output_at_levels default to true

szy21:
## Purpose 

Even though the z coordinate in the netcdf output is not correct, I think this is more useful for debugging as we can see the data on the raw levels. I kept one with `netcdf_output_at_levels: false` in ci to make sure it works.

## Content

----


akshaysridhar:
I think this is perhaps fine for the CI plots since the design intent is to use `ClimaAnalysis` to generate the correct warped surface levels as a post-processing step. cc @Sbozzolo

Sbozzolo (approved):

Sbozzolo:
Tests passing, feel free to merge

Issue #2911, opened 2024-04-17
Fix broken NVTX reports


charleskawczynski:
We need to fix the broken NVTX reports, both on central and on clima.

charleskawczynski:
This is needed to fix #2530.

charleskawczynski:
Error messages are:
```
qdstrm

Generating '/tmp/slurm-40915165/nsys-report-4d1c.qdstrm'
[1/1] [====27%                     ] report.nsys-rep
Importer error status: Importation failed.
Import Failed with unexpected exception: /dvs/p4/build/sw/devtools/Agora/Rel/QuadD_Main/QuadD/Host/QdstrmImporter/main.cpp(34): Throw in function {anonymous}::Importer::Importer(const boost::filesystem::path&, const boost::filesystem::path&)
Dynamic exception type: boost::wrapexcept<QuadDCommon::RuntimeException>
std::exception::what: RuntimeException
[QuadDCommon::tag_message*] = Status: AnalysisFailed
Error {
  Type: RuntimeError
  SubError {
    Type: InvalidArgument
    Props {
      Items {
        Type: OriginalExceptionClass
        Value: "N5boost10wrapexceptIN11QuadDCommon24InvalidArgumentExceptionEEE"
      }
      Items {
        Type: OriginalFile
        Value: "/dvs/p4/build/sw/devtools/Agora/Rel/QuadD_Main/QuadD/Host/Analysis/Modules/EventCollection.cpp"
      }
      Items {
        Type: OriginalLine
        Value: "1055"
      }
      Items {
        Type: OriginalFunction
        Value: "void QuadDAnalysis::EventCollection::CheckOrder(QuadDAnalysis::EventCollectionHelper::EventContainer&, const QuadDAnalysis::ConstEvent&) const"
      }
      Items {
        Type: ErrorText
        Value: "Wrong event order has been detected when adding events to the collection:\nnew event ={ StartNs=403098042813 StopNs=403129613160 GlobalId=349883374385042 Event={ TraceProcessEvent=[{ Correlation=139850 EventClass=1 TextId=920 ReturnValue=0 },] } Type=48 }\nlast event ={ StartNs=448052547615 StopNs=448084509068 GlobalId=349883374385042 Event={ TraceProcessEvent=[{ Correlation=209574 EventClass=1 TextId=920 ReturnValue=0 },] } Type=48 }"
      }
    }
  }
}
Generated:
[...]

PR #2912, opened 2024-04-17
Add nsys reports back to CI

charleskawczynski:
We need this to identify the performance issues (#2910), and find out why nsight systems is sometimes breaking (#2911).


szy21 (approved):

Sbozzolo:
We can try again and see if it works now.

Sbozzolo (approved):

Issue #2920, opened 2024-04-18
Add capabilities to read and interpolate from LES data


szy21:
This will be used in GCM driven single column cases.

szy21:
Closed by #2922

PR #2922, opened 2024-04-19
Add logic for reading/interpolating GCM forcings from file and storing in cache

costachris:
## Purpose 
Add logic for reading GCM forcings from file,  interpolating to CA vertical grid, and storing in cache. Part of #2920

## Content

----

### DIFF
--- a/config/default_configs/default_config.yml
+++ b/config/default_configs/default_config.yml
@@ -232,6 +232,9 @@ ls_adv:
 external_forcing:
   help: "External forcing for single column experiments [`nothing` (default), `GCM`]"
   value: ~
+external_forcing_file:
+  help: "External forcing file containing large-scale forcings, initial conditions, and boundary conditions [`nothing` (default), `path/to/file`]"
+  value: ~
 fps:
   help: "Frames per second for animations"
   value: 5
--- a/src/prognostic_equations/forcing/external_forcing.jl
+++ b/src/prognostic_equations/forcing/external_forcing.jl
@@ -5,12 +5,24 @@
 import Thermodynamics as TD
 import ClimaCore.Spaces as Spaces
 import ClimaCore.Fields as Fields
+import NCDatasets as NC
+import StatsBase
+import Dierckx
+
+function interp_vertical_prof(x, xp, fp)
+    spl = Dierckx.Spline1D(xp, fp; k = 1)
+    return spl(vec(x))
+end
+
+mean_nc_data(data, group, var; imin = 100) =
+    StatsBase.mean(data.group[group][var][:, :][:, imin:end], dims = 2)[:]
+init_nc_data(data, group, var) = data.group[group][var][:, :][:, 1]
 
 external_forcing_cache(Y, atmos::AtmosModel) =
     external_forcing_cache(Y, atmos.external_forcing)
 
-external_forcing_cache(Y, ::Nothing) = (;)
-function external_forcing_cache(Y, ::GCMForcing)
+external_forcing_cache(Y, external_forcing::Nothing) = (;)
+function external_forcing_cache(Y, external_forcing::GCMForcing)
     FT = Spaces.undertype(axes(Y.c))
     ᶜdTdt_fluc = similar(Y.c, FT)
     ᶜdqtdt_fluc = similar(Y.c, FT)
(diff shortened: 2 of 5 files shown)


config/default_configs/default_config.yml
```diff
@@ -232,6 +232,9 @@ ls_adv:
 external_forcing:
   help: "External forcing for single column experiments [`nothing` (default), `GCM`]"
   value: ~
+external_forcing_file:
+  help: "External forcing file containing large-scale forcings, initial conditions, and boundary conditions [`nothing` (default), `GCM`]"
```
haakon-e, line 236:
```suggestion
  help: "External forcing file containing large-scale forcings, initial conditions, and boundary conditions [`nothing` (default), `path/to/file`]"
```
something like this?

src/prognostic_equations/forcing/external_forcing.jl
```diff
@@ -5,12 +5,75 @@
...
+        if haskey(ds, key)
+            return ds[key]
+        else
+            for group_option in ["profiles", "reference", "timeseries"]
+                haskey(ds.group, group_option) || continue
+                if haskey(ds.group[group_option], key)
+                    return ds.group[group_option][key]
+                end
+            end
+        end
```
haakon-e, line 30: is this second branch ever hit?
If the file format is consistent, and we (for now) assume time-constant forcing, can this be simplified?

src/prognostic_equations/forcing/external_forcing.jl
```diff
@@ -5,12 +5,75 @@
...
+    end
+    return nothing
```
haakon-e, line 32: maybe we should throw an error if variable is not found?

src/prognostic_equations/forcing/external_forcing.jl
```diff
@@ -5,12 +5,75 @@
...
+end
+
+
+function pyinterp(x, xp, fp)
+    spl = Dierckx.Spline1D(xp, fp; k = 1)
+    return spl(vec(x))
+end
```
haakon-e, line 39: could we either rename this function (don't understand why it's called `pyinterp`, or just write out the spline in `getvar` and `get_nudgevar` below?

src/prognostic_equations/forcing/external_forcing.jl
```diff
@@ -5,12 +5,75 @@
...
+    pyinterp(zc_gcm, zc_les, init_nc_data(ds, "profiles", var))
+
+"""
+Calculated height dependent relaxation timescale.
+"""
+function compute_gcm_driven_τ(
+    z::FT,
+    τᵣ::FT = 24.0 * 3600.0,
+    zᵢ::FT = 3000.0,
+    zᵣ::FT = 3500.0,
```
haakon-e, line 58: add some docs on what these things are?

src/prognostic_equations/forcing/external_forcing.jl
```diff
@@ -25,21 +88,36 @@ function external_forcing_cache(Y, ::GCMForcing)
...
-    @. ᶜτ_scalar = 24hr
+    external_forcing_file = external_forcing.external_forcing_file
+
+    NC.Dataset(external_forcing_file, "r") do ds
+        Fields.bycolumn(axes(Y.c)) do colidx
+            zc_les = Array(get_nc_data(ds, "z_half"))
+            zc_gcm = Fields.coordinate_field(Y.c).z[colidx]
+
+            parent(ᶜdTdt_fluc[colidx]) .=
+                getvar("dtdt_fluc", ds, zc_gcm, zc_les)
```
haakon-e, line 99: maybe can write a wrapper function for all these?
```Julia
function getvar!(var, varname)
    parent(var[colidx]) .= getvar(varname, ds, zc_gcm, zc_les)
end
```

src/solver/types.jl
```diff
@@ -115,7 +115,9 @@ struct LargeScaleAdvection{PT, PQ}
     prof_dqtdt::PQ # Set large-scale drying
 end
 # maybe need to <: AbstractForcing
-struct GCMForcing end
+struct GCMForcing{S}
```
haakon-e, line 118:
```suggestion
struct GCMForcing{S <: String}
```
should this subtype string?

haakon-e (approved):
Looks good! Left some comments/suggestions/questions

src/prognostic_equations/forcing/external_forcing.jl
szy21: I think we almost always hit the second branch because pycles saves variables in groups.

src/prognostic_equations/forcing/external_forcing.jl
```diff
@@ -5,12 +5,75 @@
...
+    τᵣ::FT = 24.0 * 3600.0,
+    zᵢ::FT = 3000.0,
+    zᵣ::FT = 3500.0,
+) where {FT <: Real}
+    # returns height-dependent relaxation timescale from eqn. 9 in `Shen et al. 2021`
+    if z < zᵢ
+        return FT(0)
+    elseif zᵢ <= z <= zᵣ
+        cos_arg = pi * ((z - zᵢ) / (zᵣ - zᵢ))
+        # return (FT(0.5) / τ) * (1 - cos(cos_arg))
```
szy21, line 65:
```suggestion
```

src/prognostic_equations/forcing/external_forcing.jl
```diff
@@ -5,12 +5,75 @@
...
+getvar(var, ds, zc_gcm, zc_les) =
+    pyinterp(zc_gcm, zc_les, mean_nc_data(ds, "profiles", var))
+# nudge to initial LES profiles of u, v, qt, and T (equivalent to GCM initial condition)
+get_nudgevar(var, ds, zc_gcm, zc_les) =
+    pyinterp(zc_gcm, zc_les, init_nc_data(ds, "profiles", var))
+
+"""
+Calculated height dependent relaxation timescale.
+"""
+function compute_gcm_driven_τ(
```
szy21, line 54: Is this function used anywhere? If not, I think it's better to have it in a separate PR that sets nudging timescale.

szy21 (commented):
Thanks! We should change the current gcm driven job in ci to use the external forcing file. We can talk tomorrow at the EDMF meeting about if / how we should use artifact.

src/prognostic_equations/forcing/external_forcing.jl
costachris: it will be used to compute the τ, but isn't called yet. So I can save it for a separate PR

src/prognostic_equations/forcing/external_forcing.jl
costachris: Will remove this function from this PR and add in separate one.

src/prognostic_equations/forcing/external_forcing.jl
costachris: ^ +1 . Also I added a line to throw an error if it does not.

costachris: function removed :)

src/solver/types.jl
costachris: strings aren't explicitly typed in other structs in `types.jl`, so I'll just follow that convention for now

PR #2928, opened 2024-04-19
Add spectrum of total kinetic energy in baroclinc wave

Sbozzolo:
Add total kinetic energy to the diagnostics and adds a function to compute its spectrum in the `ci_plots`. The function uses `ClimaCoreSpectra.power_spectrum_2d`, which returns the power spectrum along the two wave numbers, and integrates over the azimuthal wave number ($m$).

### DIFF
--- a/post_processing/ci_plots.jl
+++ b/post_processing/ci_plots.jl
@@ -280,13 +280,57 @@ function make_plots_generic(
 end
 
 """
-    make_spectra_generic
+    compute_spectrum(var)
 
-Use ClimaCoreSpectra to compute and plot spectra for the given `vars`.
-
-Extra arguments are passed to `ClimaAnalysis.slice`
-
+Compute the spectrum associated to the given variable. Returns a ClimaAnalysis.OutputVar.
 """
+function compute_spectrum(var; mass_weight = nothing)
+    # power_spectrum_2d seems to work only when the two dimensions have precisely one
+    # twice as many points as the other
+    dim1, dim2, dim3 = var.index2dim[1:3]
+
+    len1 = length(var.dims[dim1])
+    len2 = length(var.dims[dim2])
+
+    len1 == 2len2 || error("Cannot take this spectrum ($len1 != 2 $len2)")
+
+    (dim1 == "lon" || dim1 == "long") ||
+        error("First dimension has to be longitude (found $dim1)")
+    dim2 == "lat" || error("Second dimension has to be latitude (found $dim2)")
+    dim3 == "z" || error("Third dimension has to be altitude (found $dim3)")
+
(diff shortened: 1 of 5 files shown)


post_processing/ci_plots.jl
```diff
@@ -280,13 +280,58 @@ function make_plots_generic(
...
+
+    len1 == 2len2 || error("Cannot take this spectrum ($len1 != 2 $len2)")
+
+    (dim1 == "lon" || dim1 == "long") ||
+        error("First dimension has to be longitude (found $dim1)")
+    dim2 == "lat" || error("Second dimension has to be latitude (found $dim2)")
+    dim3 == "z" || error("Third dimension has to be altitude (found $dim3)")
+
+    FT = eltype(var.data)
+    mass_weight = ones(FT, length(var.dims[dim3]))
```
akshaysridhar, line 303: We're hardcoding weights here; could this be passed as an argument to this function with this default value instead?

post_processing/ci_plots.jl
```diff
@@ -280,13 +280,58 @@ function make_plots_generic(
...
+
+    attributes = Dict(
+        "short_name" => "log_spectrum_" * var.attributes["short_name"],
+        "long_name" => "Spectrum of " * var.attributes["long_name"],
+        "units" => "",
+    )
+    path = nothing
+
+    return ClimaAnalysis.OutputVar(
+        attributes,
```
akshaysridhar, line 327: I like that this is unified through ClimaAnalysis now; thanks!

post_processing/ci_plots.jl
```diff
@@ -597,6 +650,14 @@ function make_plots(
     vars = map_comparison(simdirs, short_names) do simdir, short_name
         return get(simdir; short_name, reduction)
     end
+    vars_spectra =
+        map_comparison(simdirs, short_names_spectra) do simdir, short_name
+            compute_spectrum(
+                slice(get(simdir; short_name, reduction), time = LAST_SNAP),
```
szy21, line 656:
```suggestion
                slice(get(simdir; short_name, reduction), time = 10days),
```
(We want spectra at day 10 instead of the last time step, would this work?)

post_processing/ci_plots.jl
```diff
@@ -645,6 +706,13 @@ function make_plots(
...
```
szy21, line 712:
```suggestion
                slice(get(simdir; short_name, reduction), time = 10days),
```

szy21 (approved):
Thank you!

post_processing/ci_plots.jl
Sbozzolo: Yes, it will work

PR #2931, opened 2024-04-20
Add julia 1.11-beta to lookahead and add dependabot

Sbozzolo:
See if we can catch some problems earlier than later.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "weekly"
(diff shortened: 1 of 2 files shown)


Sbozzolo:
https://github.com/CliMA/ClimaAtmos.jl/actions/runs/8765896794/job/24057894026?pr=2931

AtmosConfig cannot be inferred/compiled in Julia 1.11-beta1

charleskawczynski (commented):
Looks great, I'd like to see the error, but I'm running into cuda compilation issues.

Maybe we can remove CUDA as a direct dependency first, as none of our GHA runners will run on gpus.

Sbozzolo:
When I tried on my laptop, compilation would just hang with no errors.

Hopefully things will improve in the next betas

charleskawczynski:
I’m happy to merge this with soft-fail after rebasing

charleskawczynski (approved):

PR #2937, opened 2024-04-22
Deep atmosphere init for baroclinic wave problem

akshaysridhar:
Update BaroclinicWave problem initialization with [deep atmosphere equations](https://rmets.onlinelibrary.wiley.com/doi/epdf/10.1002/qj.2241)


szy21:
I think it would be better to keep the option for shallow atmosphere. Could you make the initial condition dependent on deep or shallow atmosphere?

PR #2938, opened 2024-04-22
Test longruns without dss ops in mesh-warp gradient

akshaysridhar:
Tracks test branch in ClimaCore with no-dss on meshwarp horizontal gradient terms


szy21:
Is this still relevant?

Issue #2950, opened 2024-04-24
Improve prognostic implicit edmf performance


charleskawczynski:
This issue is for tracking the performance of the prognostic implicit edmf performance.
```[tasklist]
### Tasks
- [ ] Make a reproducer for the kernels discussed here: https://github.com/CliMA/ClimaAtmos.jl/pull/2951#issuecomment-2077315044
- [ ] Implement shared memory for FD kernels, see if this
- [ ] Collect launch statistics/benchmarks based on hard-coded threads/blocks vs `CUDA.launch_configuration`
- [ ] Implement and apply broadcast fusion for similar pointwise expressions
- [ ] Implement and apply broadcast fusion for similar FD expressions
```

charleskawczynski:
I looked at the nvtx report from #2947, and this is what I'm seeing:

Here is a full `step!` (labeled by `ClimaTimeSteppers` on the left), for the big picture view:

<img width="1961" alt="Screen Shot 2024-04-24 at 9 25 40 AM" src="https://github.com/CliMA/ClimaAtmos.jl/assets/1880641/4be757bf-3106-485d-a058-a906775008fb">

The 3 "large" cpu calls to `remaining_tendency!` / `hyperdiffusion_tendency!` are the ones that follow the implicit solve. They only appear large because _many_ kernels are launched during the implicit solve, which pile up in the gpu queue, and there's a `CUDA.synchronize` call in `hyperdiffusion_tendency!`'s DSS call.

Let's zoom in to see who is responsible for these kernel launches. I'll click on the cpu functions (labeled `ClimaAtmos`), and the kernels (`99.9% Kernels (named by NVTX)` above will highlight:

`dss!`:

<img width="1972" alt="Screen Shot 2024-04-24 at 9 40 36 AM" src="https://github.com/CliMA/ClimaAtmos.jl/assets/1880641/ef751f09-a6b0-40b5-9f1b-fea3e107b6f9">

`set_precomputed_quantities!`

<img width="1969" alt="Screen Shot 2024-04-24 at 9 43 46 AM" src="https://github.com/CliMA/ClimaAtmos.jl/assets/1880641/e7584ebd-9c20-47bb-8048-adb62bceb738">

`wfact!`:

[...]

PR #2951, opened 2024-04-24
Flatten buoyancy_gradients call

charleskawczynski:
A step towards #2530.


charleskawczynski:
I looked at the nsight report in https://buildkite.com/clima/climaatmos-target-gpu-simulations/builds/272, and `buoyancy_gradients` + the 3 hoisted `ᶜgradᵥ-ᶠinterp` calls are now 3.9 ms, so this PR cuts the time in half (xref: https://github.com/CliMA/ClimaAtmos.jl/issues/2950#issuecomment-2077307860).

What I suspect is the culprit for this slowdown, is both/either not using shared memory, or register spills, due to complexity.

We should probably make our FD use shared memory, but this is at least an improvement, so I'll merge.

Issue #2901, opened 2024-04-12
Implement GCM driven SCM


szy21:
Work in progress...

Step 1: Add all cache and tendency for GCM driven SCM, use hard coded values and make sure the case runs

- [x] #2916
- [x] horizontal advection
- [x] vertical fluctuation
- [x] nudging
- [x] #2917
- [x] #2918
- [x] radiation
- [x] add to buildkite

Step 2: Read and interpolate forcing from the LES file
- [x] #2920
- [x] #2952

see also https://github.com/CliMA/TurbulenceConvection.jl/issues/505

costachris:
Original TC PR:
https://github.com/CliMA/TurbulenceConvection.jl/commit/da2ae4e45838b4413fba50c267a21554db4641e7

szy21:
This is completed I think.

PR #2956, opened 2024-04-25
Initial conditions for GCM driven SCM 

### LINKED ISSUE #2916: Add initial condition for GCM driven SCM

haakon-e:
ref: #2901

closes #2916

## Content

----

### DIFF
--- a/src/initial_conditions/initial_conditions.jl
+++ b/src/initial_conditions/initial_conditions.jl
@@ -1124,3 +1124,66 @@ function (initial_condition::PrecipitatingColumn)(params)
     end
     return local_state
 end
+
+"""
+    GCMDriven
+
+The `InitialCondition` from a provided GCM forcing file.
+"""
+Base.@kwdef struct GCMDriven <: InitialCondition
+    external_forcing_file::String = error("No forcing file provided")
+end
+
+function (initial_condition::GCMDriven)(params)
+    (; external_forcing_file) = initial_condition
+    thermo_params = CAP.thermodynamics_params(params)
+
+    # Read forcing file
+    z_gcm = read_gcm_z(external_forcing_file)
+    θ, u, v, q_tot, ρ₀ =
+        map(read_gcm_initial_conditions(external_forcing_file)) do value
+            Dierckx.Spline1D(z_gcm, value; k = 1)
+        end
+
+    function local_state(local_geometry)
+        (; z) = local_geometry.coordinates
+        FT = typeof(z)
+        return LocalState(;
+            params,
+            geometry = local_geometry,
+            thermo_state = ts = TD.PhaseEquil_ρθq(
+                thermo_params,
+                FT(ρ₀(z)),
+                FT(θ(z)),
+                FT(q_tot(z)),
+            ),
+            velocity = Geometry.UVVector(FT(u(z)), FT(v(z))),
+            turbconv_state = EDMFState(; tke = FT(0)),
+        )
+    end
+    return local_state
+end
+
+function read_gcm_z(external_forcing_file)
+    NC.NCDataset(external_forcing_file) do ds
+        read_gcm_driven_reference_profile(Float64, ds, "reference", "z")
+    end
+end
+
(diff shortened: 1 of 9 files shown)


src/initial_conditions/initial_conditions.jl
```diff
@@ -1124,3 +1124,94 @@ function (initial_condition::PrecipitatingColumn)(params)
...
+    group,
+    varname,
+    FT,
+    dimnames::NTuple{N, String},
+) where {N}
+    grp = ds.group[group]
+    dimids = NC.nc_inq_dimid.(grp.ncid, dimnames)
+    varid = NC.nc_inq_varid(grp.ncid, varname)
+    NC.Variable{FT, N, typeof(grp)}(grp, varid, dimids)
+end
```
haakon-e, line 1217: I think this should probably move somewhere else (now or later).
suggestions welcome

haakon-e: I can add some documentation as needed, but basically it ensures we can read from NetCDF files in a completely type-stable way (thanks Dennis for help!)

szy21:
I wonder if we could combine some of the functions for reading files and interpolation with those in this file: https://github.com/CliMA/ClimaAtmos.jl/blob/main/src/prognostic_equations/forcing/external_forcing.jl? We can move the functions to utils if that is better.

src/initial_conditions/initial_conditions.jl
```diff
@@ -1124,3 +1124,94 @@ function (initial_condition::PrecipitatingColumn)(params)
...
+    ds,
+    group,
+    varname,
+    dimnames::NTuple{N, String},
+) where {N}
+    grp = ds.group[group]
+    dimids = NC.nc_inq_dimid.(grp.ncid, dimnames)
+    varid = NC.nc_inq_varid(grp.ncid, varname)
+    NC.Variable{FT, N, typeof(grp)}(grp, varid, dimids)
+end
```
haakon-e, line 1217: cc: @Sbozzolo function for fetching variable type stably.

szy21 (approved):
Looks good, thanks!

src/utils/read_gcm_driven_scm_data.jl
```diff
@@ -0,0 +1,48 @@
...
+_gcm_driven_nz(ds) = ds.group["reference"].dim["z"]
+
+_gcm_driven_variable_zt(FT, ds, group, varname) =
+    _gcm_driven_variable(FT, ds, group, varname, ("z", "t"))
+
+_gcm_driven_variable_z(FT, ds, group, varname) =
+    _gcm_driven_variable(FT, ds, group, varname, ("z",))
+
+
+function _gcm_driven_variable(
```
costachris, line 28: I think these 8 helper functions should be consolidated into just a few.  The get initial profile operation, for example, could be 1 line. The use case for adding ie. `_gcm_driven_variable` functions isn't clear to me, given we know the group name and they're specified elsewhere in the code. The group could maybe be a kwarg elsewhere with a default to "profiles".

src/initial_conditions/initial_conditions.jl
```diff
@@ -1124,3 +1124,66 @@ function (initial_condition::PrecipitatingColumn)(params)
...
+        read_gcm_driven_reference_profile(Float64, ds, "reference", "z")
+    end
+end
+
+function read_gcm_initial_conditions(external_forcing_file)
+    NC.NCDataset(external_forcing_file) do ds
+        (
+            # Note: `Float64` is type of the read GCM data, different from `FT` used elsewhere
+            read_gcm_driven_initial_profile(
+                Float64,
```
costachris, line 1178: If `Float64` is the type of the read GCM data, could it just be inferred from the ds arg in `read_gcm_driven_..._profile`? 
If not, could you just have it as a default kwarg or defined within `read_gcm_driven_...._profile`, so that it can be changed in one place if need be.

costachris (changes requested):
Requested a couple of clarifications/suggestions

PR #2958, opened 2024-04-25
surface fluxes for gcm driven scm

### LINKED ISSUE #2918: Add surface flux in GCM driven SCM

haakon-e:
ref #2901 

closes #2918 

## Content

----


src/surface_conditions/surface_setups.jl
szy21, line 1: I only looked at this file for now and it looks good. I don't think we need to pass in q_vap to SurfaceState (although with fluxes=HeatFluxes it shouldn't do anything). And we can double check with Costa on the momentum surface flux.

costachris: This seems correct. In TC the `ρu_flux` and `ρv_flux` are computed internally by `SF.surface_conditions`. The same happens here (`MoninObukhov` -> `SurfaceState` -> `set_surface_conditions!` -> `atmos_surface_conditions` -> `SF.surface_conditions`) as far as I can tell.

PR #2960, opened 2024-04-25
Fix kinetic energy conservation with topography

dennisYatunin:
## Purpose 

This PR introduces some changes that should allow us to conserve kinetic energy in the presence of topography.

## To-do

- Add a test for exact conservation of kinetic energy in the presence of topography.

## Content

- Replace `ᶠwinterp(ᶜJ, ᶜρ)` in the vertical transport equation with `ᶠinterp(ᶜJ * ᶜρ) / ᶠJ`, since `ᶠdiv` internally multiplies its argument by `ᶠJ`.
- Replace `dot(uₕ, CTh(uₕ)) + 2 * dot(ᶜinterp(uᵥ), CT3(uₕ))` in the kinetic energy equation with `dot(uₕ, CTh(uₕ) + CTh(ᶜinterp(uᵥ))) + dot(ᶜinterp(uᵥ), CT3(uₕ))`, which makes it so that `∫ ᶜρₜ * ᶜK * dz = -∫ ᶜρ * ᶜKₜ * dz`.
- Update the relevant Jacobian terms.

----
- [ ] I have read and checked the items on the review checklist.

PR #2961, opened 2024-04-26
Add test for conservation of kinetic energy with topography

dennisYatunin:
## Purpose 

This PR adds a test for exact conservation of kinetic energy in the presence of topography. The test currently passes with the changes from #2960, and it would be helpful to see whether it passes without those changes.

## To-do

Update the relevant Jacobian terms, and add a test that uses implicit timestepping.

----


szy21:
Do you need to change something in the Jacobian as well?

dennisYatunin:
Yes, I need to turn off gravity and pressure gradients in the vertical velocity tendency derivatives

PR #2962, opened 2024-04-26
Output CF, q_liq and q_ice from quadratures and use it in radiation

### LINKED ISSUE #2862: Use quadrature for cloud water diagnostics

trontrytel:
closes #2862


szy21 (approved):
Thanks! Could you change the diagnostics for clw and cli as well?

PR #2968, opened 2024-04-30
test RRTMGP fix

juliasloan25:
## Purpose 
Use CommonDataModel@0.3.6 (which introduces an error via RRTMGP), and the branch RRTMGP#js/ft-broadcast to test fix in https://github.com/CliMA/RRTMGP.jl/pull/484


juliasloan25:
no longer needed

PR #2970, opened 2024-05-01
Test updated dss ops

akshaysridhar:
Tracks ClimaCore branch with updated dss operations (3rd component is not a scalar)


akshaysridhar:
Prelim Results with double mountain (Hughes & Jablonowski (2023)) with updated dss ordering and 3 component update in ClimaCore.jl 
z_sfc
u
w
rv
hus

PR #2983, opened 2024-05-06
Fuse some similar pointwise broadcast expressions

charleskawczynski:
This PR uses `@fused_direct` to fuse some broadcast expressions


charleskawczynski:
We're going to put this on hold until we fix https://github.com/CliMA/ClimaCore.jl/issues/1910.

charleskawczynski:
Hm, not sure why this closed when I force-pushed.. I'll reopen a new one.

PR #2995, opened 2024-05-10
Add job_id to gpu and LR pipelines

charleskawczynski:
This PR is a followup to #2994. [xref](https://github.com/CliMA/ClimaAtmos.jl/pull/2994#issuecomment-2103456215). Thanks for catching that @szy21!


szy21 (approved):
Thank you!

PR #3006, opened 2024-05-13
Remove some use of bycolumn

charleskawczynski:
This PR attempts to remove the pretty widespread use of `bycolumn`, in light of it being an issue: https://github.com/CliMA/ClimaCore.jl/issues/1609

I left the gravity wave alone, since that needs a lot of other things fixed.


szy21 (approved):
The code change looks good to me.

PR #3008, opened 2024-05-13
Add benchmarks, remove `default_perf_config_file`

charleskawczynski:
This PR:
 - Adds some benchmarks to the gpu target pipeline
 - Removes `default_perf_config_file`, and correspondingly updates the perf scripts and inlines the defaults into the remaining perf config files (to preserve behavior). Next, we can unify those yaml files.


charleskawczynski:
I added the breaking label, as the coupler uses the ClimaAtmos yaml configs. [xref](https://github.com/CliMA/ClimaCoupler.jl/issues/800#issuecomment-2120372375)

PR #3015, opened 2024-05-15
add FT wrapper for scalars

haakon-e:
## Purpose 

Add macro `@convert_scalars` to conveniently wrap scalars (e.g. integers, floats) with the provided `FT`, e.g.:

```julia
FT = Float32
@convert_scalars FT 1 + 2.0 + 3.0f0 * π + sin(pi)
# expands to: FT(1) + FT(2.0) + FT(3.0f0) * FT(π) + sin(FT(pi))
```

### DIFF
--- a/src/utils/utilities.jl
+++ b/src/utils/utilities.jl
@@ -404,3 +404,42 @@ function gaussian_smooth(arr::AbstractArray, sigma::Int = 1)
 
     return smoothed_arr
 end
+
+
+"""
+    macro convert_scalars(FT_expr, expr)
+
+The `convert_scalars` macro is used to convert scalar values in an expression to a specified type.
+
+## Arguments
+- `FT_expr`: A function or expression that specifies the type to which the scalar values should be converted.
+- `expr`: The expression in which scalar values need to be converted.
+
+## Returns
+The macro returns the modified expression with scalar values converted to the specified type.
+
+## Example
+```julia
+FT = Float32
+@convert_scalars FT 1 + 2.0 + 3.0f0 * π + sin(pi)
+# expands to: FT(1) + FT(2.0) + FT(3.0f0) * FT(π) + sin(FT(pi))
+```
+"""
+macro convert_scalars(FT_expr, expr)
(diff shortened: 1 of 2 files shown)


charleskawczynski (commented):
Where is this intended to be used?

haakon-e:
I wanted this for some functions I'll be writing that provide forcing / boundary conditions for the LES. They will look similar to (but mine have more terms needing to be wrapped in `FT`) e.g. 

https://github.com/CliMA/ClimaAtmos.jl/blob/0ee96c094a43c2f989a5925f106780d98edbc6a6/src/initial_conditions/initial_conditions.jl#L646-L650

which (in the case of the example above), would become something like

```julia
edmfx_q_tot(::Type{FT}) where {FT} = @convert_scalars FT z ->
    z < 0.7e4 ? 1e-3 * exp(-(z - 4e3)^2 / 2 / 1e3^2) : 0

edmfx_q_tot(::Type{FT}) where {FT} = @convert_scalars FT z ->
    z < 0.7e4 ? 1e-3 * exp(-(z - 4e3)^2 / 2 / 1e3^2) : 0
```

src/utils/utilities.jl
```diff
@@ -404,3 +404,42 @@ function gaussian_smooth(arr::AbstractArray, sigma::Int = 1)
...
+FT = Float32
+@convert_scalars FT 1 + 2.0 + 3.0f0 * π + sin(pi)
+# expands to: FT(1) + FT(2.0) + FT(3.0f0) * FT(π) + sin(FT(pi))
+```
+"""
+macro convert_scalars(FT_expr, expr)
+
+    # Helper function to process expressions
+    function process_expr(expr, FT_expr)
+        if isa(expr, Number) || expr in (:π, :pi)
```
dennisYatunin, line 432: I'd recommend using something like `names(MathConstants)[2:end]` instead of manually listing out `:π`, `:pi`, etc.

szy21:
What's the status of this PR?

PR #3017, opened 2024-05-15
Make updraft explicit in prognostic edmf single column cases

szy21:
## Purpose 

Also unifies some toml files and modifies edmf plots.

## Content

----


charleskawczynski:
The failure is some sort of issue with the benchmark, I'll manually merge and fix this.

PR #3025, opened 2024-05-17
Forcing for ISDAC

haakon-e:
## Purpose 

See https://github.com/CliMA/AtmosphericProfilesLibrary.jl/pull/41 for forcing

## Content

----

### DIFF
--- a/config/model_configs/les_isdac_box.yml
+++ b/config/model_configs/les_isdac_box.yml
@@ -1,0 +1,30 @@
+job_id: "les_isdac_box"
+initial_condition: "ISDAC" 
+subsidence: "ISDAC" 
+surface_setup: "ISDAC"
+external_forcing: "ISDAC"
+rad: "ISDAC"
+moist: "equil"
+config: "box" 
+hyperdiff: "false" 
+implicit_diffusion: false
+ode_algo: "SSP33ShuOsher"
+precip_model: "1M"
+smagorinsky_lilly: true
+c_smag: 0.20
+x_max: 3.2e3
+y_max: 3.2e3 
+z_max: 2e3
+x_elem: 4
+y_elem: 4
+z_elem: 10
+z_stretch: false
+dt: "0.05secs" 
+t_end: "10mins" 
+dt_save_state_to_disk: "1mins"
+# restart_file: "restart/isdac/day0.0.hdf5"
+netcdf_interpolation_num_points: [10, 10, 20]
+diagnostics:
+  - short_name: [ts, ta, thetaa, ha, pfull, rhoa, ua, va, wa, hur, hus, cl, clw, cli, hussfc, evspsbl]
+    reduction: average
+    period: 1mins
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -172,6 +172,19 @@ steps:
         agents:
           slurm_mem: 20GB
 
+      - label: ":genie: LES ISDAC in a box"
+        command: >
+          julia --color=yes --project=examples examples/hybrid/driver.jl
+          --config_file $CONFIG_PATH/les_isdac_box.yml
+          --job_id les_isdac_box
+        artifact_paths: "les_isdac_box/*"
+        env:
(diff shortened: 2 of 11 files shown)


config/model_configs/les_isdac_box.yml
```diff
@@ -0,0 +1,32 @@
+# device: CPUSingleThreaded
+job_id: "les_isdac_box"
+initial_condition: "ISDAC"
```
akshaysridhar, line 3: The current assembly and yaml file information is reasonable, but let's start a separate issue to define `cases` that collect all relevant forcing terms for a given configuration.

config/model_configs/les_isdac_box.yml
```diff
@@ -0,0 +1,32 @@
+# device: CPUSingleThreaded
+job_id: "les_isdac_box"
+initial_condition: "ISDAC" 
+subsidence: "ISDAC" 
+surface_setup: "ISDAC"
+external_forcing: "ISDAC"
+rad: "ISDAC"
+moist: "equil"
+config: "box" 
+hyperdiff: "true"
```
akshaysridhar, line 10: We can also set this to `false` and remove the `coefficient` values below to reduce the `yaml` file size.

config/model_configs/les_isdac_box.yml
```diff
@@ -0,0 +1,32 @@
...
+z_elem: 200
+z_stretch: false
+dt: "0.1secs" 
+t_end: "6hours" 
+dt_save_state_to_disk: "10mins"
+# restart_file: "restart/isdac/day0.0.hdf5"
+netcdf_interpolation_num_points: [64, 64, 200]
+diagnostics:
+  - short_name: [ts, ta, thetaa, ha, pfull, rhoa, ua, va, wa, hur, hus, cl, clw, cli, hussfc, evspsbl]
+    period: 5secs
```
akshaysridhar, line 32: Let's define `reduction: average` and an appropriate `period` for this operation here. `5sec` instantaneous output might be too much here (reasonable for prototyping/ debugging I suppose).

akshaysridhar (commented):
I'll take a closer look at the AtmosphericProfiles shortly and comment here once that is done, but this LGTM overall.

config/model_configs/les_isdac_box.yml
```diff
@@ -0,0 +1,32 @@
+# device: CPUSingleThreaded
+job_id: "les_isdac_box"
+initial_condition: "ISDAC" 
+subsidence: "ISDAC"
```
szy21, line 4: Does ISDAC have subsidence? I didn't see it in this PR. If not I think it would be better to remove this line.

config/model_configs/les_isdac_box.yml
szy21, line 1: If you want to run it in ci and make plots, you need to add this job to .buildkite/pipeline.yml and post_processing/ci_plots.jl. The t_end you have is probably too long for ci.

config/model_configs/les_isdac_box.yml
haakon-e: Here it is:

https://github.com/CliMA/ClimaAtmos.jl/pull/3025/files#diff-ab240b0d863b064ad1bd28f7844952c3809af57b3424ed31794a4835f6349e47R319-R320

haakon-e:
@akshaysridhar could you have a look at the config file and see if there's something I should change to make the test pass? I'm probably just misconfiguring resolution / time step / other knobs relative to a robust LES setup.

szy21:
Where is `smagorinsky_lilly` used? Maybe it is unstable because there is no diffusion.

jty53:
When I ran this, the simulation crashed in the first few iterations until I reduced the timestep to satisfy the cfl condition (thanks @akshaysridhar).

jty53 (commented):
`device: CUDA` is invalid. Use `device: CUDADevice`

jty53 (changes requested):
`GCMForcing{DType}([...]) ` is in the wrong elseif block. Moved up into appropriate section for GCMForcing.

akshaysridhar (approved):

PR #3031, opened 2024-05-20
Add 3 timestep restart unit test

### LINKED ISSUE #2604: Make sure restart works correctly
We should make sure the simulation results are identical when running from a restart file and when running from a cold start (freshly). It would be good if we can add a test for this in ci. cc @charleskawczynski 

Some thoughts from @Sbozzolo :

- How are the new callback initialized? For example, if you have a callback called every N/dt, does it pick up from the old ones has left or is a new counter?
- We have to test MPI/GPU cases as well
- What happens when you change settings across restarts?
- Accmulated diagnostics will not work across restarts
- Progress log will probably be off across restarts
- Where are restart/output files saved? If we use the ActiveLink style and keep restart_dir = .../output_active, we will need to be careful because the base dir associated to output_active changes with the new simulation.

akshaysridhar:
Tests for restart and custom output directory generation. Assumes `output_active` path will be generated using `ClimaUtilities.OutputPathGenerator`. Addresses #2604 (not sufficient to close this issue).

### DIFF
--- a/test/restart.jl
+++ b/test/restart.jl
@@ -1,0 +1,137 @@
+import ClimaAtmos as CA
+import SurfaceFluxes as SF
+import ClimaAtmos.Parameters as CAP
+import ClimaCore as CC
+import Thermodynamics as TD
+import CloudMicrophysics as CM
+import SciMLBase
+import ClimaCore.Spaces
+import ClimaComms
+using Test
+
+@testset begin
+    "Test restarts across configuration combinations"
+    ### Test Description
+    # Generate a simulation with some complexity of 
+    # config arguments. Some config combinations are
+    # incompatible so we do not sweep over all possible
+    # iterations.
+
+    # Modify the timestep to 1-second increments. 
+    # Save simulation state at each timestep, 
+    # and generate a restart file at 0secs, 2secs simulation time.
+    # Verify objects read in using ClimaCore.InputOutput functions 
+    # are identical (i.e. restarts result
+    # in the same simulation states as if one were to advance
+    # the timestepper uninterrupted.)
+
+    # TODO: Restart and diagnostic behaviour needs to be 
+    # clearly defined when config files have different 
+    # settings (or when tendency computations conflict with
+    # dt or t_end parsed args)
+
+    for configuration in ["sphere", "column"]
+        for moisture in ["equil"]
+            for turb_conv in ["diagnostic_edmfx", "prognostic_edmfx"]
+                for precip in ["0M", "1M"]
+
+                    mktempdir() do output_loc
+                        test_dict = Dict(
+                            "check_nan_every" => 3,
+                            "log_progress" => false,
+                            "moist" => moisture,
+                            "precip_model" => precip,
+                            "config" => configuration,
+                            "turbconv" => turb_conv,
+                            "perturb_initstate" => false,
+                            "dt" => "1secs",
+                            "t_end" => "3secs",
+                            "dt_save_state_to_disk" => "1secs",
+                            "enable_diagnostics" => false,
+                        )
+
+                        ### Boilerplate default integrator objects
+                        config = CA.AtmosConfig(
+                            merge(
+                                test_dict,
+                                Dict(
+                                    "output_dir" => joinpath(
+                                        output_loc,
+                                        "gen_test_output",
+                                    ),
+                                ),
+                            ),
+                        )
+
+                        @info config.parsed_args["output_dir"]
+                        simulation = CA.get_simulation(config)
+                        Y₀ = simulation.integrator.u
+                        CA.solve_atmos!(simulation)
+
+                        restart_dir = simulation.output_dir
+                        @test isfile(joinpath(restart_dir), "day0.2.hdf5")
+                        restart₁ = joinpath(restart_dir, "day0.2.hdf5")
+
+                        @info "Restart #1 from: " restart₁
+                        config₁ = CA.AtmosConfig(
+                            merge(
+                                test_dict,
+                                Dict(
+                                    "restart_file" => restart₁,
+                                    "output_dir" => joinpath(
+                                        output_loc,
+                                        "restart_test_output",
+                                    ),
+                                ),
+                            ),
+                        )
+                        simulation_test₁ = CA.get_simulation(config₁)
+                        @info "Advancing restarted simulation"
+                        CA.solve_atmos!(simulation_test₁)
+                        @info "Restarted simulation complete"
+
+                        # TODO: Extend for MPI and GPU checks as part of standard runtests.jl
+                        file0 = joinpath(
+                            output_loc,
+                            "gen_test_output/output_active/day0.0.hdf5",
+                        )
+                        f0 = CC.InputOutput.HDF5Reader(
+                            file0,
+                            ClimaComms.SingletonCommsContext(),
+                        )
+                        f0Y = CC.InputOutput.read_field(f0, "Y")
+
+                        file1 = joinpath(
+                            output_loc,
+                            "gen_test_output/output_active/day0.3.hdf5",
+                        )
+                        file2 = joinpath(
+                            output_loc,
+                            "restart_test_output/output_active/day0.3.hdf5",
+                        )
+                        f1 = CC.InputOutput.HDF5Reader(
+                            file1,
+                            ClimaComms.SingletonCommsContext(),
+                        )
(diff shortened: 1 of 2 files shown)


test/runtests.jl
```diff
@@ -19,7 +19,9 @@ using Test
 @safetestset "Sponge interface tests" begin @time include("parameterized_tendencies/sponge/rayleigh_sponge.jl") end
 @safetestset "Precipitation interface tests" begin @time include("parameterized_tendencies/microphysics/precipitation.jl") end
 @safetestset "Model getters" begin @time include("solver/model_getters.jl") end
-@safetestset "Topography tests" begin @time include("topography.jl") end
+@safetestset "Topography tests" begin @time include("topography.jl")
```
juliasloan25, line 22: missing an `end` here

szy21:
Could we add radiation callback and EDMF too? Maybe we can have two cases, a simple one and a more complex one?

Sbozzolo:
I think "Make sure restarts work correctly" is a little more nuanced than this.

- How are the new callback initialized? For example, if you have a callback called every N/dt, does it pick up from the old ones has left or is a new counter? 
- We have to test MPI/GPU cases as well
- What happens when you change settings across restarts? 
- Accmulated diagnostics will not work across restarts 
- Progress log will probably be off across restarts
- Where are restart/output files saved? If we use the ActiveLink style and keep `restart_dir = .../output_active`, we will need to be careful because the base dir associated to `output_active` changes with the new simulation.

These are some of the first things that come to mind but I am sure that there are much more. Restarts are not easy to handle properly!

szy21:
> I think "Make sure restarts work correctly" is a little more nuanced than this.
> 
> * How are the new callback initialized? For example, if you have a callback called every N/dt, does it pick up from the old ones has left or is a new counter?
> * We have to test MPI/GPU cases as well
> * What happens when you change settings across restarts?
> * Accmulated diagnostics will not work across restarts
> * Progress log will probably be off across restarts
> * Where are restart/output files saved? If we use the ActiveLink style and keep `restart_dir = .../output_active`, we will need to be careful because the base dir associated to `output_active` changes with the new simulation.
> 
> These are some of the first things that come to mind but I am sure that there are much more. Restarts are not easy to handle properly!

Right, I don't think this PR should close the issue. I moved your comments to the issue so we can discuss how to address them.

akshaysridhar:
I'm testing all reasonable combinations of this `simulation` offline, there is a tradeoff w.r.t testing time (assuming that the user sets identical configurations for now). I've updated  the description to remove closing keyword cc @szy21 (my understanding of this was that the primary focus here is to ensure that simulation objects have consistent properties and spaces) - I'll add some smaller issues to capture the details which are more closely related to callbacks and progress logging but maintain this PR as a check of "state" accuracy across MPI/GPU combinations. There are other interface related issues when automating this due to default config args and "unused" config options for different combinations of restarted simulations - I think we generally need to update our error handling and hidden default settings w.r.t simulation configurations cc @Sbozzolo .

szy21:
I think this is a good start. Let's address MPI and GPU in this PR if that's easy. We can address other issues in separate PRs. For callbacks and diagnostics, I think for now we can force the simulation time to be multiple of the N or dt of callback and accumulated diagnostics (or throw a warning if it's not). I don't think we need to worry too much about what happens when changing settings across restarts, as it's very unlikely that one will change the settings during one simulation, so there is no truth to compare with.

@akshaysridhar @Sbozzolo What do you think? We can move the discussion to the issue.

test/restart.jl
```diff
@@ -0,0 +1,103 @@
...
+        ),
+    )
+    simulation_test₁ = CA.get_simulation(config₁);
+    SciMLBase.step!(simulation_test₁.integrator);
+    
+    Y = simulation.integrator.u
+    Y₀ = simulation_test₀.integrator.u
+    Y₁ = simulation_test₁.integrator.u
+
+    rm("./test_restart", recursive=true)
```
Sbozzolo, line 96: If the test fails, this will not be executed and some stray files will be left on the disk.

You should use temporary directories instead.

Sbozzolo:
Yes, this is a perfectly good test to have, but should not close the issue.

test/restart.jl
```diff
@@ -0,0 +1,102 @@
...
+                    FT = eltype(Y)
+
+                    @info "Check simulation states"
+                    @test Spaces.axes(Y) == Spaces.axes(Y₁)
+                    @test extrema(parent(Y₁.c.ρ) .- parent(Y.c.ρ)) ==
+                          (FT(0), FT(0))
+                    @test extrema(parent(Y₁.c.ρe_tot) .- parent(Y.c.ρe_tot)) ==
+                          (FT(0), FT(0))
+                    @test extrema(parent(Y₁.c.ρq_tot) .- parent(Y.c.ρq_tot)) ==
+                          (FT(0), FT(0))
```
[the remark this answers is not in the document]
akshaysridhar: Now uses `all` to check fieldvectors

akshaysridhar:
> This is a good start, but I think we should keep in mind that this is still a relatively weak test with respect to fully supporting restarts. We should go ahead with this, but we should also think about restarts more in general and what we want to support

We may need to look at the HDF5Reader / Writer features: on spheres, we get exact matches for scalar tracers and not for the vectors (e.g O(1e-5) error in `Y.f.u\_3` upon restarting.

akshaysridhar:
> Is the error relative or absolute?
> 
> Can you coming up a simple example where you set up a space, set up a vector on the space, save it to HDF5 and read it back and show that they are different?

Absolute error : In the ubuntu ci build we have 
```
[ Info: Check file-read from checkpoint data
maximum(abs.(parent(f1Y.c.uₕ) .- parent(f2Y.c.uₕ))) = 2.9802322f-8
maximum(abs.(parent(f1Y.f.u₃) .- parent(f2Y.f.u₃))) = 3.0517578f-5
```

test/restart.jl
```diff
@@ -0,0 +1,102 @@
...
+                                "dt_save_state_to_disk" => "2secs",
+                            ),
+                        ),
+                    )
+
+                    @info config.parsed_args["output_dir"]
+                    simulation = CA.get_simulation(config)
+                    CA.solve_atmos!(simulation)
+
+                    restart_dir = simulation.output_dir
```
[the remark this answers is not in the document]
charleskawczynski: I think we probably want `restart_dir = dirname(parsed_args["restart_file"])` here. This is nice in that it is declarative. I would hope that this is compatible with `OutputPathGenerator`, or that we can somehow make it compatible with `OutputPathGenerator`.

test/restart.jl
```diff
@@ -0,0 +1,144 @@
...
+                                ),
+                            ),
+                        )
+                        simulation_test₁ = CA.get_simulation(config₁)
+                        @info "Advancing restarted simulation"
+                        CA.solve_atmos!(simulation_test₁)
+                        @info "Restarted simulation complete"
+
+                        # TODO: Extend for MPI and GPU checks as part of standard runtests.jl
+                        # TODO: Fix output paths once `OutputPathGenerator` is restored.
```
Sbozzolo, line 94:
```suggestion
```

test/restart.jl
```diff
@@ -0,0 +1,144 @@
...
+                        ) == FT(0)
+                        @test maximum(
+                            abs.(parent(f1Y.c.ρe_tot) .- parent(f2Y.c.ρe_tot)),
+                        ) == FT(0)
+                        @test maximum(
+                            abs.(parent(f1Y.f.u₃) .- parent(f2Y.f.u₃)),
+                        ) <= sqrt(eps(FT))
+                        @test maximum(
+                            abs.(parent(f1Y.c.uₕ) .- parent(f2Y.c.uₕ)),
+                        ) <= sqrt(eps(FT))
```
Sbozzolo, line 138: This might be condensed to
```suggestion
                        @test maximum(
                            abs.(parent(f1Y.c) .- parent(f2Y.c)),
                        ) <= sqrt(eps(FT))
                        @test maximum(
                            abs.(parent(f1Y.f) .- parent(f2Y.f)),
                        ) <= sqrt(eps(FT))                        
```

akshaysridhar: Indeed, this was the prior setup - I was curious to check scalar vs vector differences (this needs to be looked at within the ClimaCore HDF5Reader / Writer setups)

Sbozzolo: Ah, got it!

szy21:
What's the status of this PR?

PR #3032, opened 2024-05-20
Adds an option to periodically check for NaNs in the state

Sbozzolo:
Not quite like #1935, but almost functionally equivalent when you set the period to 1.

(I cleaned up some unrelated config in the meantime)

Closes #1935

### DIFF
--- a/src/callbacks/get_callbacks.jl
+++ b/src/callbacks/get_callbacks.jl
@@ -166,6 +166,17 @@ function get_callbacks(config, sim_info, atmos, params, Y, p, t_start)
             ),
         )
     end
+    check_nan_every = parsed_args["check_nan_every"]
+    if check_nan_every > 0
+        @info "Checking NaNs in the state every $(check_nan_every) steps"
+        callbacks = (
+            callbacks...,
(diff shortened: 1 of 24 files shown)


src/callbacks/get_callbacks.jl
```diff
@@ -166,6 +166,14 @@ function get_callbacks(config, sim_info, atmos, params, Y, p, t_start)
             ),
         )
     end
+    check_nan_every = parsed_args["check_nan_every"]
+    if check_nan_every > 0
+        @info "Checking NaNs in the state every $(check_nan_every) steps"
+        callbacks = (
+            callbacks...,
+            call_every_n_steps((integrator) -> check_nans(integrator)),
```
szy21, line 174:
```suggestion
            call_every_n_steps((integrator) -> check_nans(integrator), n = check_nan_every),
```

szy21 (approved):
Thanks!

src/callbacks/get_callbacks.jl
Sbozzolo: Thank you!

PR #3038, opened 2024-05-22
Do not exercise diagnostics when disabled

charleskawczynski:
This PR allows users to fully disable diagnostics, so that it is not exercised, when disabled. I find myself having to do this locally quite often, so I thought that it might be a good thing to merge.

### DIFF
--- a/config/default_configs/default_config.yml
+++ b/config/default_configs/default_config.yml
@@ -259,6 +259,9 @@ override_τ_precip:
 output_default_diagnostics:
   help: "Output the default diagnostics associated to the selected atmospheric model"
   value: true
+enable_diagnostics:
+  help: "Set to false to fully disable the diagnostics"
+  value: true
 netcdf_output_at_levels:
   help: "Do not perform any vertical interpolation in the output NetCDF files. Instead, interpolate horizontally and output the level. "
   value: true
--- a/src/solver/type_getters.jl
+++ b/src/solver/type_getters.jl
@@ -663,17 +663,21 @@ function get_simulation(config::AtmosConfig)
     @info "get_callbacks: $s"
 
     # Initialize diagnostics
-    s = @timed_str begin
-        scheduled_diagnostics, writers = get_diagnostics(
-            config.parsed_args,
-            atmos,
-            Y,
-            p,
-            t_start,
-            sim_info.dt,
-        )
+    if config.parsed_args["enable_diagnostics"]
+        s = @timed_str begin
+            scheduled_diagnostics, writers = get_diagnostics(
+                config.parsed_args,
+                atmos,
+                Y,
+                p,
+                t_start,
+                sim_info.dt,
+            )
+        end
+        @info "initializing diagnostics: $s"
+    else
+        writers = nothing
     end
-    @info "initializing diagnostics: $s"
 
     continuous_callbacks = tuple()
     discrete_callbacks = callback
@@ -705,13 +709,15 @@ function get_simulation(config::AtmosConfig)
     end
     @info "init integrator: $s"
 
-    s = @timed_str begin
-        integrator = ClimaDiagnostics.IntegratorWithDiagnostics(
-            integrator,
-            scheduled_diagnostics,
-        )
+    if config.parsed_args["enable_diagnostics"]
+        s = @timed_str begin
+            integrator = ClimaDiagnostics.IntegratorWithDiagnostics(
+                integrator,
+                scheduled_diagnostics,
+            )
+        end
+        @info "Added diagnostics: $s"
     end
-    @info "Added diagnostics: $s"
 
     reset_graceful_exit(output_dir)
 
--- a/src/solver/solve.jl
+++ b/src/solver/solve.jl
@@ -103,7 +103,7 @@ function solve_atmos!(simulation)
         maxrss_str = prettymemory(maxrss())
         @info "Memory currently used (after solve!) by the process (RSS): $maxrss_str"
 
-        foreach(close, output_writers)
+        isnothing(output_writers) || foreach(close, output_writers)
     end
 end
 


szy21:
If the user defines customized diagnostics in the yaml file (e.g. [here](https://github.com/CliMA/ClimaAtmos.jl/blob/3fbb1bccebc8090418eff1f17b0d7f5fc80c77d7/config/model_configs/prognostic_edmfx_bomex_column.yml#L29)) but set `output_default_diagnostics` to false, the model still needs to output diagnostics. We should check that as well?

charleskawczynski:
> If the user defines customized diagnostics in the yaml file (e.g. [here](https://github.com/CliMA/ClimaAtmos.jl/blob/3fbb1bccebc8090418eff1f17b0d7f5fc80c77d7/config/model_configs/prognostic_edmfx_bomex_column.yml#L29)) but set `output_default_diagnostics` to false, the model still needs to output diagnostics. We should check that as well?

There's no documentation/defaults on `diagnostics`. Perhaps we can/should add a separate flag, `enable_diagnostics = true`, which can be used to perform this check. I think it would be convenient to have a single switch

src/solver/type_getters.jl
```diff
@@ -705,13 +709,15 @@ function get_simulation(config::AtmosConfig)
     end
     @info "init integrator: $s"
 
-    s = @timed_str begin
-        integrator = ClimaDiagnostics.IntegratorWithDiagnostics(
-            integrator,
-            scheduled_diagnostics,
-        )
+    if config.parsed_args["output_default_diagnostics"]
```
Sbozzolo:
```suggestion
    if !isempty(scheduled_diagnostics)
```

src/solver/type_getters.jl
```diff
@@ -663,17 +663,21 @@ function get_simulation(config::AtmosConfig)
...
-    s = @timed_str begin
-        scheduled_diagnostics, writers = get_diagnostics(
-            config.parsed_args,
-            atmos,
-            Y,
-            p,
-            t_start,
-            sim_info.dt,
-        )
+    if config.parsed_args["output_default_diagnostics"]
```
Sbozzolo:
```suggestion
if config.parsed_args["output_default_diagnostics"] || !isempty(config.parsed_args["diagnostics"])
```

Sbozzolo: (Should be something like this, I haven't tried)

config/default_configs/default_config.yml
```diff
@@ -259,6 +259,9 @@ override_τ_precip:
 output_default_diagnostics:
   help: "Output the default diagnostics associated to the selected atmospheric model"
   value: true
+enable_diagnostics:
+  help: "Set to false to fully disable the diagnostics"
+  value: true
```
Sbozzolo, line 264:
```suggestion
```

src/solver/type_getters.jl
```diff
@@ -663,17 +663,21 @@ function get_simulation(config::AtmosConfig)
...
-    s = @timed_str begin
-        scheduled_diagnostics, writers = get_diagnostics(
-            config.parsed_args,
-            atmos,
-            Y,
-            p,
-            t_start,
-            sim_info.dt,
-        )
+    if config.parsed_args["enable_diagnostics"]
```
Sbozzolo, line 666:
```suggestion
    if config.parsed_args["enable_diagnostics"] || !isempty(config.parsed_args["diagnostics"]
```

src/solver/type_getters.jl
```diff
@@ -663,17 +663,21 @@ function get_simulation(config::AtmosConfig)
...
+                atmos,
+                Y,
+                p,
+                t_start,
+                sim_info.dt,
+            )
+        end
+        @info "initializing diagnostics: $s"
+    else
+        writers = nothing
```
Sbozzolo, line 679:
```suggestion
        writers, scheduled_diagnostics = nothing, nothing
```

src/solver/type_getters.jl
```diff
@@ -705,13 +709,15 @@ function get_simulation(config::AtmosConfig)
     end
     @info "init integrator: $s"
 
-    s = @timed_str begin
-        integrator = ClimaDiagnostics.IntegratorWithDiagnostics(
-            integrator,
-            scheduled_diagnostics,
-        )
+    if config.parsed_args["enable_diagnostics"]
```
Sbozzolo, line 712:
```suggestion
    if !isnothing(scheduled_diagnostics)
```

Sbozzolo (commented):
Maybe the more important question is, why do you find yourself disabling this piece of code? What needs to be changed so that you don't have to do that?

Sbozzolo:
> > If the user defines customized diagnostics in the yaml file (e.g. [here](https://github.com/CliMA/ClimaAtmos.jl/blob/3fbb1bccebc8090418eff1f17b0d7f5fc80c77d7/config/model_configs/prognostic_edmfx_bomex_column.yml#L29)) but set `output_default_diagnostics` to false, the model still needs to output diagnostics. We should check that as well?
> 
> There's no documentation/defaults on `diagnostics`. Perhaps we can/should add a separate flag, `enable_diagnostics = true`, which can be used to perform this check. I think it would be convenient to have a single switch

There is documentation. It's here: https://clima.github.io/ClimaAtmos.jl/dev/diagnostics/

And I don't think we should add another redundant switch that effectively just disables other yaml configurations: it makes handling interaction between configurations harder to predict and maintain.

src/solver/type_getters.jl
charleskawczynski: I don't want to exercise this code, that is the point of this PR.

Sbozzolo: I don't understand. If you add the suggestion above, this code is not exercised when there are no diagnostics

Sbozzolo: So, if you want to not exercise this code, you can just comment out the diagnostics from your configuration

charleskawczynski:
> And I don't think we should add another redundant switch that effectively just disables other yaml configurations: it makes handling interaction between configurations harder to predict and maintain.

It's not redundant in that it ensures that our diagnostics are not exercised, the other flags do not do this. Think of this as a developer flag

Sbozzolo:
> > Maybe the more important question is, why do you find yourself disabling this piece of code?
> 
> Because the diagnostics is still brittle. I already gave feedback on the diagnostics and `IntegratorWithDiagnostics` design, and you insisted to keep it. All I'm asking for is a way to disable it now, because it's frequently causing problems for me during development. IIRC, the last error I saw was an assumption about the orography type. @bloops recently ran into an issue when we were trying to run ClimaAtmos on the cloud. I'm happy to open issues when they arise, but these snags are slowing development for me and others.

I am happy to have this PR merged, but please, report these issues and tag me so that we can fix them. I didn't know you or @bloops had problems.

> It's not redundant in that it ensures that our diagnostics are not exercised, the other flags do not do this. Think of this as a developer flag

I think that my suggestion is equivalent in terms of code not being exercised but does not add an extra configuration that needs to be maintained.

Sbozzolo (approved):
> Thats fair, but it does require users to specify both flags to disable diagnostics (and the `diagnostics` flag does not seem to have a default), and those may or may not live in two different files for a particular configuration. I'm happy to (eventually) remove this flag, but I don't see it as adding major overhead, and I think that the name and docs are clear.

It's not a major overhead, but I would like to try resisting configuration creep, especially when other flags are very close in scope. We have already more than 100 and it is so easy to end up with configurations that are misconfigured and find unexpected results.


The default value for `diagnostics` is its absence. `diagnostics` is not a flag, but a nested structure, and it cannot be described in the same way we use for the other flags in `default_config.yml`. We can probably add it by having it parse as `nothing` by default, and handle that case specially.

PR #3039, opened 2024-05-22
Upgrade dependencies

charleskawczynski:
This PR upgrades dependencies, including ClimaCore, which has some performance improvements.


charleskawczynski:
This passed before, I'm going to do a manual merge and combine this with #3041, and remove the `lim!` benchmarking, as this seems to be hanging.

PR #3040, opened 2024-05-22
Bump compat for cloudmicrophysics

Sbozzolo:
Nothing has changed on the atmos side

Closes #3037


trontrytel (approved):

charleskawczynski:
The manifests aren't changed in this PR, so this PR only changes the versions used in the test environment cc @trontrytel, @Sbozzolo

Sbozzolo:
> The manifests aren't changed in this PR, so this PR only changes the versions used in the test environment cc @trontrytel, @Sbozzolo

Yes, it's fine. Nothing has changed for ClimaAtmos, so we don't need to use the most recent version, but we can start supporting the newer ones.

PR #3047, opened 2024-05-24
Remove CPUSummary from extras

Sbozzolo:
It doesn't seem that this is used anywhere.


charleskawczynski (approved):
LGTM, but I think that this came in due to https://github.com/JuliaLang/Pkg.jl/issues/3014, and it may just pop back up

PR #3050, opened 2024-05-24
Add precip heating terms

trontrytel:
Add precipitation heating term from Raymond 2013 (doi:10.1002/jame.20050). We already take the freezing and melting into account. So we are just adding the temperature gradient term

Screenshot 2024-06-04 at 3 17 58 PM

### DIFF
--- a/src/parameterized_tendencies/microphysics/microphysics_wrappers.jl
+++ b/src/parameterized_tendencies/microphysics/microphysics_wrappers.jl
@@ -16,6 +16,8 @@ const qᵥ = TD.vapor_specific_humidity
 qₜ(thp, ts) = TD.PhasePartition(thp, ts).tot
 qₗ(thp, ts) = TD.PhasePartition(thp, ts).liq
 qᵢ(thp, ts) = TD.PhasePartition(thp, ts).ice
+cᵥₗ(thp) = TD.Parameters.cv_l(thp)
+cᵥᵢ(thp) = TD.Parameters.cv_i(thp)
 
 # helper function to limit the tendency
 function limit(q::FT, dt::FT, n::Int) where {FT}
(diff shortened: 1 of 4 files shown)


src/parameterized_tendencies/microphysics/microphysics_wrappers.jl
```diff
@@ -261,6 +263,44 @@ function compute_precipitation_sources!(
...
+)
+    # TODO - at some point we want to switch to assuming that precipitation
+    # is at wet bulb temperature
+
+    # compute full temperature gradient
+    @. ᶜ∇T = CT123(ᶜgradᵥ(ᶠinterp(Tₐ(thp, ᶜts))))
+    @. ᶜ∇T += CT123(gradₕ(Tₐ(thp, ᶜts)))
+    # dot product with effective velocity of precipitation
+    # (times q and specific heat)
+    @. ᶜSeₜᵖ -= dot(ᶜ∇T, (ᶜu - C123(Geometry.WVector(ᶜwᵣ)))) * cᵥₗ(thp) * ᶜqᵣ
```
szy21, line 301: The docs say using c_p here. Do we know which one we should use?

trontrytel: `p` in `c_p` there was standing for precipitating species not constant pressure. Sorry thats confusing. I thought we said it should be under constant volume in all the places other than radiative tendnecies?

szy21: Ah ok. Well, I'm not sure if I will ever understand when to use which...but that sounds good to me:)

szy21 (approved):
Looks good to me, thanks!

PR #3055, opened 2024-05-28
tag v0.24.1

AlexisRenchon:
## Purpose 
Tag bi-weekly release


juliasloan25:
It looks like this has the same CUDA error that we saw in land so it'll need to be pinned here too

AlexisRenchon:
@juliasloan25 I will merge when you approve!

juliasloan25 (approved):
Looks good! Similar to land, it might be good to open an issue about pinning the CUDA version here

PR #3070, opened 2024-06-05
Test 1M EDMF (diag and prog) on GPU

trontrytel:
This PR makes EDMFX (prognostic and diagnostic) work with 1M microphysics on the GPU. Thank you @charleskawczynski !

@szy21 - I wasn't sure if you would like to also keep the 0M edmf aquaplanet runs in the CI?


szy21 (approved):
No we don't have to keep the 1M one. Why does this change make it work on GPU?

trontrytel:
You mean you want to keep the 0M in the GPU CI? Not the 1M? I think it would be good to have 1M tests on the GPU. Maybe we can add both then?

For diagnostic edmf we had to flatten the detrainment function signature to pass less things in. For prognostic edmf it's a bug in ClimaCore. Once it's fixed I should be able to use the normal fill (i.e. `@.  blah = FT(0)` instead of `@. blah = FT(0) * blah2`)

charleskawczynski (approved):
LGTM. Also, it's possible that this additional `entrainment` method may not even be needed after https://github.com/CliMA/ClimaCore.jl/pull/1782 🤞🏻. That said, I think it's a good idea to merge this before that

szy21:
> You mean you want to keep the 0M in the GPU CI? Not the 1M? I think it would be good to have 1M tests on the GPU. Maybe we can add both then?
> 
> For diagnostic edmf we had to flatten the detrainment function signature to pass less things in. For prognostic edmf it's a bug in ClimaCore. Once it's fixed I should be able to use the normal fill (i.e. `@. blah = FT(0)` instead of `@. blah = FT(0) * blah2`)

Sorry that was a typo (I was replying while explaining edmf equations to others lol). I meant we don't need to keep the 0M one.

src/parameterized_tendencies/microphysics/microphysics_wrappers.jl
```diff
@@ -160,11 +160,13 @@ function compute_precipitation_sources!(
...
     thp,
 )
-    FT = eltype(Sqₜᵖ)
-    @. Sqₜᵖ = FT(0)
-    @. Sqᵣᵖ = FT(0)
-    @. Sqₛᵖ = FT(0)
-    @. Seₜᵖ = FT(0)
+    FT = eltype(thp)
+    # @. Sqₜᵖ = FT(0) should work after fixing
+    # https://github.com/CliMA/ClimaCore.jl/issues/1786
```
akshaysridhar, line 165: Thanks @trontrytel. Can you please update the news file as well to mark this change?

PR #3071, opened 2024-06-06
Remove Distributions

### LINKED ISSUE #3046: Remove Distributions.jl
PR https://github.com/CliMA/ClimaAtmos.jl/pull/2898 adds Distributions.jl to the dependency tree of ClimaAtmos. From Distributions.jl, it seems that we only need to compute quantiles and comulative distribution functions. Both of them are not too difficult to compute analytically. 

Given that the need for Distributions.jl is very narrow and the required functions can be easily coded up, we should drop this unnecessary dependency.

Sbozzolo:
```julia
    xp_low = Distributions.quantile(Distributions.Normal(), low_percentile)
    xp_high = Distributions.quantile(Distributions.Normal(), high_percentile)
    return (gauss_int(xp_high) - gauss_int(xp_low)) / max(
        Distributions.cdf(Distributions.Normal(), xp_high) -
        Distributions.cdf(Distributions.Normal(), xp_low),
        eps(FT),
    )
```

Part of this block is doing nothing.

`Distributions.quantile(percentile)` computes the `x` so that the CDF is `percentile`. Then, `Distributions.cdf(x)` computes the `percentile` corresponding to `x`. The two functions are one the inverse of the other when the same distribution is used.

So, all we need to remove distributions is implement the inverse error function. I used one of the appromation listed on Wikipedia:
image

Closes #3046


szy21 (approved):
Thanks! The code change looks good to me. I didn't check the math but I trust you and wiki:) Could you check what is the result for `percentile_bounds_mean_norm(0.9, 1)`, and maybe add a test? (I can add the test later if you are busy with other things.)

Sbozzolo:
> Thanks! The code change looks good to me. I didn't check the math but I trust you and wiki:) Could you check what is the result for `percentile_bounds_mean_norm(0.9, 1)`, and maybe add a test? (I can add the test later if you are busy with other things.)

I added a test, I just forgot to commit it!

PR #3079, opened 2024-06-11
Update deps

charleskawczynski:


szy21:
Is the change from `1.10.4` to `1.10.3` for julia intentional?

charleskawczynski:
Ah, no, I just haven’t updated locally yet. I’ll do that today

PR #3084, opened 2024-06-12
Add precipitation surface fluxes for 1M microphysics

trontrytel:
This PR adds the surface precipitation fluxes for the 1M microphysics scheme. I'm not super sure if this will be correct with topography, but I tried and had support from @dennisYatunin 

It also renames the 0M precipitation fluxes from `col_integrated_rain` (and snow) to `surface_rain_flux` (and snow). It is a breaking release because of this. I will follow up with a PR to the Coupler.

I added surface precipitation plots whenever I saw surface evaporation plots in the CI.

### DIFF
--- a/config/model_configs/single_column_precipitation_test.yml
+++ b/config/model_configs/single_column_precipitation_test.yml
@@ -18,3 +18,5 @@ toml: [toml/single_column_precipitation_test.toml]
 diagnostics:
   - short_name: [hus, clw, cli, husra, hussn, ta, wa]
     period: 500secs
+  - short_name: [pr, ]
+    period: 10secs
--- a/src/diagnostics/core_diagnostics.jl
+++ b/src/diagnostics/core_diagnostics.jl
@@ -599,21 +599,30 @@ add_diagnostic_variable!(
 
 ###
 # Precipitation (2d)
-# TODO: add precipitation flux for the 1-moment microphysics
 ###
 compute_pr!(out, state, cache, time) =
     compute_pr!(out, state, cache, time, cache.atmos.precip_model)
 compute_pr!(_, _, _, _, precip_model::T) where {T} =
     error_diagnostic_variable("pr", precip_model)
 
-function compute_pr!(out, state, cache, time, precip_model::Microphysics0Moment)
+function compute_pr!(
+    out,
+    state,
+    cache,
+    time,
+    precip_model::Union{
+        NoPrecipitation,
+        Microphysics0Moment,
+        Microphysics1Moment,
+    },
+)
     if isnothing(out)
-        return cache.precipitation.col_integrated_rain .+
-               cache.precipitation.col_integrated_snow
+        return cache.precipitation.surface_rain_flux .+
+               cache.precipitation.surface_snow_flux
     else
         out .=
-            cache.precipitation.col_integrated_rain .+
-            cache.precipitation.col_integrated_snow
+            cache.precipitation.surface_rain_flux .+
+            cache.precipitation.surface_snow_flux
     end
 end
 
--- a/src/parameterized_tendencies/microphysics/precipitation.jl
+++ b/src/parameterized_tendencies/microphysics/precipitation.jl
@@ -17,7 +17,13 @@ precipitation_cache(Y, atmos::AtmosModel) =
 ##### No Precipitation
 #####
 
-precipitation_cache(Y, precip_model::NoPrecipitation) = (;)
+function precipitation_cache(Y, precip_model::NoPrecipitation)
+    FT = Spaces.undertype(axes(Y.c))
+    return (;
+        surface_rain_flux = zeros(axes(Fields.level(Y.f, half))),
+        surface_snow_flux = zeros(axes(Fields.level(Y.f, half))),
+    )
+end
 precipitation_tendency!(Yₜ, Y, p, t, ::NoPrecipitation, _) = nothing
 
 #####
@@ -31,12 +37,8 @@ function precipitation_cache(Y, precip_model::Microphysics0Moment)
         ᶜS_ρe_tot = similar(Y.c, FT),
         ᶜ3d_rain = similar(Y.c, FT),
         ᶜ3d_snow = similar(Y.c, FT),
-        col_integrated_rain = zeros(
-            axes(Fields.level(Geometry.WVector.(Y.f.u₃), half)),
-        ),
-        col_integrated_snow = zeros(
-            axes(Fields.level(Geometry.WVector.(Y.f.u₃), half)),
-        ),
+        surface_rain_flux = zeros(axes(Fields.level(Y.f, half))),
+        surface_snow_flux = zeros(axes(Fields.level(Y.f, half))),
     )
 end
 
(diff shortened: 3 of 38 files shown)


szy21 (approved):
Thanks! I may do the version update in a separate PR, especially that it is a breaking release (and there will probably be a version update tomorrow or Monday). But up to you.

src/parameterized_tendencies/microphysics/precipitation.jl
```diff
@@ -283,6 +297,31 @@ function compute_precipitation_cache!(
...
+
+    (; ᶠtemp_scalar) = p.scratch
+    slg = Fields.level(Fields.local_geometry_field(ᶠtemp_scalar), Fields.half)
+
+    # Constant extrapolation: - put values from bottom cell center to bottom cell face
+    ˢρ = Fields.Field(Fields.field_values(Fields.level(Y.c.ρ, 1)), axes(slg))
+    ˢqᵣ = Fields.Field(Fields.field_values(Fields.level(ᶜqᵣ, 1)), axes(slg))
+    ˢqₛ = Fields.Field(Fields.field_values(Fields.level(ᶜqₛ, 1)), axes(slg))
+    ˢwᵣ = Fields.Field(Fields.field_values(Fields.level(ᶜwᵣ, 1)), axes(slg))
+    ˢwₛ = Fields.Field(Fields.field_values(Fields.level(ᶜwₛ, 1)), axes(slg))
```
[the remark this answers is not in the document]
dennisYatunin: That would cost an extra kernel launch, since we'd still need to do a projection after extrapolating.

src/parameterized_tendencies/microphysics/precipitation.jl
```diff
@@ -283,6 +297,31 @@ function compute_precipitation_cache!(
...
+
+    # Project the flux to C3 vector and convert to physical units.
+    # (This should work with topography)
+    @. surface_rain_flux = projected_vector_data(C3, ˢρ * ˢqᵣ * ˢwᵣ, slg)
```
dennisYatunin, line 321: I think we might actually want CT3 instead of C3 here, since CT3 is the one that's orthogonal to topography (and therefore the one that's orthogonal to wind velocity at the surface).

szy21: I think C3 is orthogonal to topography?

szy21: Also since we are trying to make it correct with topography...should it be `ˢwᵣ+ w` where w is the physical vertical velocity at the surface (which is not zero when there is topography)?

dennisYatunin: If we use CT3, then we don't need to add any extra component from grid mean velocity because u³ is zero at the surface.

szy21: CT3 is along the grid, so vertical (no orthogonal)?

dennisYatunin: C3 is orthogonal to wind velocity at the surface in the Euclidean sense (it makes a 90 degree angle), but CT3 is orthogonal in the non-Euclidian sense (the dot product of the unit vector along the CT3 axis with the wind velocity vector is 0 at the surface, based on how we've defined our metric tensor).

dennisYatunin: Yes, sorry, I think orthogonality and u³ being zero are distinct concepts that I got confused. C3 should be orthogonal to CT12, which runs along the surface, in the sense that their dot products are 0.

config/model_configs/single_column_precipitation_test.yml
```diff
@@ -18,3 +18,5 @@ toml: [toml/single_column_precipitation_test.toml]
 diagnostics:
   - short_name: [hus, clw, cli, husra, hussn, ta, wa]
     period: 500secs
+  - short_name: [pr, ]
+    period: 10secs
```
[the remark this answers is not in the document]
trontrytel: Yes. I do a very low frequency output for the profiles. It would result in 3 data points for the time series.

src/diagnostics/core_diagnostics.jl
```diff
@@ -606,14 +606,27 @@ compute_pr!(out, state, cache, time) =
...
 
-function compute_pr!(out, state, cache, time, precip_model::Microphysics0Moment)
+function compute_pr!(out, state, cache, time, precip_model::NoPrecipitation)
     if isnothing(out)
-        return cache.precipitation.col_integrated_rain .+
-               cache.precipitation.col_integrated_snow
+        return zeros(Spaces.horizontal_space(axes(state.c.ρ)))
+    else
+        out .= zeros(Spaces.horizontal_space(axes(state.c.ρ)))
+    end
```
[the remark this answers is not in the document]
trontrytel: I would rather define it here, so that I can easily add it to all the simulations in the CI. I.e. my logic was that I want to have a precipitation flux in the plots in the CI for all the simulations. Hopefully soon they will all be running with at least 0M microphysics, but I think now still the default is no precipitation

src/parameterized_tendencies/microphysics/precipitation.jl
trontrytel: So I should stick with orthogonal to the topography, aka `C3`? And then add `w`. But then, shouldn't I add the whole velocity vector and then do the `C3` projection? Why only add `w`?

trontrytel:
> Thanks! I may do the version update in a separate PR, especially that it is a breaking release (and there will probably be a version update tomorrow or Monday). But up to you.

Sure! Yes, it would be great to do the version update after this is merged in. So that I can also update the Coupler and run the simulations with land there

src/diagnostics/core_diagnostics.jl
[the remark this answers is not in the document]
szy21: If you want precipitation flux with NoPrecipitation, I think it would be better to define it in the cache and return it for the diagnostics, like what you have for 0M and 1M, rather than having a dispatch on the diagnostics. This way we can avoid the case where someone (accidentally) adds precipitation for NoPrecipitation, but the diagnostics show zero.

src/parameterized_tendencies/microphysics/precipitation.jl
dennisYatunin: CT3 might be what we want here because it guarantees that the precipitation flux will be zero when the precipitation velocity is 0. If we use C3, we will always have an additional flux from grid mean velocity (regardless of whether we use w or the full velocity vector) in the presence of topography, which feels wrong to me. 

Sorry about using the word orthogonal; I don't think it's the main thing we need to focus on in this non-Euclidian setting.

szy21: Oh, wait, is `wᵣ` positive or negative?

szy21: precipitation_flux is positive upward, in 0M, I think. So there should be a minus sign here if wᵣ is positive downward.

trontrytel: Yes, good catch! 

Lets talk tomorrow and figure out w/wo `w` and w/wo `T`

src/diagnostics/core_diagnostics.jl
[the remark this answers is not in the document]
trontrytel: Sounds good. I'll add that

szy21:
Just a heads up, I noticed that the 1M run in this week's build breaks after ~5 days: https://buildkite.com/clima/climaatmos-gpulongruns/builds/206#01901a45-511a-47d8-a869-666664bc3d9b. It seems we don't have a regression test for microphysics, but this seems to be the only PR that is related. Maybe double check the change in the tendency?

PR #3090, opened 2024-06-13
limit surface albedo between 0 and 1

szy21:
## Purpose 

This shouldn't be needed, but just to be safe.

## Content

----


src/cache/surface_albedo.jl
```diff
@@ -115,51 +115,61 @@ Calculate the direct surface albedo using the regression function of Jin et al.
...
-                    p[5] * σ +
-                    p[6] * σ * cosθ
-                ) * exp(
-                    p[7] +
-                    p[8] * cosθ +
-                    p[9] * cosθ^2 +
-                    p[10] * σ +
-                    p[11] * σ * cosθ,
+            if cosθ <= 0
+                return 0
```
Sbozzolo, line 119:
```suggestion
                return zero(FT)
```

src/cache/surface_albedo.jl
```diff
@@ -115,51 +115,61 @@ Calculate the direct surface albedo using the regression function of Jin et al.
...
+                        p[11] * σ * cosθ,
+                    )
+
+                # return the albedo (J11, eq. 1)
+                return min(
+                    1,
+                    max(
+                        0,
+                        rf(n, cosθ) - rf(n, cosθ) / rf(n0, cosθ) * f(cosθ, σ),
+                    ),
```
Sbozzolo, line 170:
```suggestion
                return min(
                    one(FT),
                    max(
                        zero(FT),
                        rf(n, cosθ) - rf(n, cosθ) / rf(n0, cosθ) * f(cosθ, σ),
                    ),
```

src/cache/surface_albedo.jl
```diff
@@ -177,25 +187,36 @@ function surface_albedo_diffuse(
...
-                q_clear[4] * n * σ
-
-            # cloudy sky (J11, eq. 5b)
-            q_cloud = α_model.q_cloud
-            α_cloud = q_cloud[1] + q_cloud[2] * n + q_cloud[3] * n * σ
-
-            # return the albedo
-            return could_fraction * α_cloud + (1 - could_fraction) * α_clear
+            if cosθ <= 0
+                return 0
```
Sbozzolo, line 191:
```suggestion
                return zero(FT)
```

src/cache/surface_albedo.jl
```diff
@@ -177,25 +187,36 @@ function surface_albedo_diffuse(
...
+                α_cloud = q_cloud[1] + q_cloud[2] * n + q_cloud[3] * n * σ
+
+                # return the albedo
+                return min(
+                    1,
+                    max(
+                        0,
+                        could_fraction * α_cloud +
+                        (1 - could_fraction) * α_clear,
+                    ),
```
Sbozzolo, line 217:
```suggestion
                    one(FT),
                    max(
                        zero(FT),
                        could_fraction * α_cloud +
                        (1 - could_fraction) * α_clear,
                    ),
```

src/cache/surface_albedo.jl
szy21: Thanks! Is zero(FT) the same as FT(0)?

Sbozzolo: No, zero(FT) is slightly better, probably because it is inferred and compiled

szy21:
I also change the values in the test when cos(zenith_angle) is smaller or equal than 0.

PR #3099, opened 2024-06-18
remove filter for radiative fluxes

szy21:
## Purpose 

I think it's not ideal to have different values in the diagnostics and in the variables that are used by the model, so I decided to remove this filter. Alternatively, I can add a filter to set the SW fluxes to zero, but I think it would be good to reveal the problem and see if modifying rrtmgp can fix it. If this continues to be a problem, I can use my own branch with the filter for the vesri run.

## Content

----

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,7 +3,8 @@ ClimaAtmos.jl Release Notes
 
 Main
 -------
-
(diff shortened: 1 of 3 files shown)


Sbozzolo (approved):

Sbozzolo:
I agree that this should be handled upstream

PR #3101, opened 2024-06-18
Sync NetCDF file at every write

Sbozzolo:
This change essentially disables the internal buffer in the NetCDF library. With this change, every time something has to be written to disk, it is immediately written (as opposed to buffering writes).

This ensures that datasets are always written and no data is lost, but comes to a performance cost.

This was already the default behavior for GPU runs, so this change only affects CPU runs
Closes https://github.com/CliMA/ClimaCoupler.jl/issues/858


szy21 (approved):
Thanks!

PR #3122, opened 2024-06-21
Remove `bycol` ops in vertical diffusion function

akshaysridhar:
Removes one of the pending `bycolumn` operations from the vertical diffusion Frierson function. `colidx` is still used in some tendency calculations (gravity wave, external forcing)

### DIFF
--- a/src/cache/precomputed_quantities.jl
+++ b/src/cache/precomputed_quantities.jl
@@ -358,7 +358,6 @@ function eddy_diffusivity_coefficient(
     end
 end
 
-### Frierson (2006) diffusion
 function compute_boundary_layer_height!(
     h_boundary_layer,
     f_b::FT,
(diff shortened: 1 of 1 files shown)


akshaysridhar:
I'll run a `longrun` sweep, but since this is a test of particular interest, here are the results for the `longrun_aquaplanet_clearsky_1M` job_id, run to the default longrun simulation end time which is 120 days. 
[summary.pdf](https://github.com/user-attachments/files/15930828/summary.pdf)
[summary_2D.pdf](https://github.com/user-attachments/files/15930827/summary_2D.pdf)

cc @trontrytel

trontrytel:
I think those look reasonable!

config/longrun_configs/longrun_aquaplanet_clearsky_1M.yml
```diff
@@ -8,6 +8,7 @@ z_max: 55000.0
 rayleigh_sponge: true
 moist: "equil"
 precip_model: "1M"
+hyperdiffusion: "CAM_SE"
```
szy21, line 11: This should not be needed as CAM_SE is the default.

szy21 (approved):

akshaysridhar:
@szy21 I added a version update change as well to this PR so we can release this before the longruns get started.

akshaysridhar:
(version update removed)

szy21:
The plots in summary look strange to me. Specific humidity seems uniform, and there seem to be no rain, or cloud ice, or liquid water. They don't seem to be consistent with the plots in summary_2D. Could you double check?

PR #3123, opened 2024-06-21
Move conservation check to src

Sbozzolo:
So that it can be used without the driver

### DIFF
--- a/examples/hybrid/driver.jl
+++ b/examples/hybrid/driver.jl
@@ -116,55 +116,24 @@ end
 
 # Conservation checks
 if config.parsed_args["check_conservation"]
-    @info "Checking conservation"
     FT = Spaces.undertype(axes(sol.u[end].c.ρ))
+    @info "Checking conservation"
+    (; energy_conservation, mass_conservation, water_conservation) =
+        CA.check_conservation(sol)
 
-    # energy
-    energy_total = sum(sol.u[end].c.ρe_tot)
-    energy_atmos_change = sum(sol.u[end].c.ρe_tot) - sum(sol.u[1].c.ρe_tot)
+    @info "    Net energy change / total energy: $energy_conservation"
+    @info "    Net mass change / total mass: $mass_conservation"
+
     sfc = p.atmos.surface_model
-    if sfc isa CA.PrognosticSurfaceTemperature
-        sfc_cρh = sfc.ρ_ocean * sfc.cp_ocean * sfc.depth_ocean
-        energy_total +=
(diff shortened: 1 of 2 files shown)


src/solver/solve.jl
```diff
@@ -196,3 +196,69 @@ function precompile_callbacks(integrator)
...
+    else
+        energy_surface_change = -p.net_energy_flux_sfc[][]
+    end
+    energy_radiation_input = -p.net_energy_flux_toa[][]
+
+    energy_net =
+        abs(
+            energy_atmos_change + energy_surface_change -
+            energy_radiation_input,
+        ) / energy_total
```
szy21, line 242:
```suggestion
    energy_conservation =
        abs(
            energy_atmos_change + energy_surface_change -
            energy_radiation_input,
        ) / energy_total
```
(I think there is something wrong with the conservation test before? We are dividing it by energy_total here, but then divide it by energy_total again for energy_conservation.)

src/solver/solve.jl
```diff
@@ -196,3 +196,69 @@ function precompile_callbacks(integrator)
...
+        water_atmos_change = sum(sol.u[end].c.ρq_tot) - sum(sol.u[1].c.ρq_tot)
+        water_surface_change = horizontal_integral_at_boundary(
+            sol.u[end].sfc.water .- sol.u[1].sfc.water,
+        )
+
+        water_conservation =
+            abs(water_atmos_change + water_surface_change) / water_total
+    end
+
+    energy_conservation = energy_net / energy_total
```
szy21, line 261:
```suggestion
```

examples/hybrid/driver.jl
```diff
@@ -116,55 +116,23 @@ end
 
 # Conservation checks
 if config.parsed_args["check_conservation"]
-    @info "Checking conservation"
     FT = Spaces.undertype(axes(sol.u[end].c.ρ))
+    @info "Checking conservation"
+    (; energy_conservation, mass_conservation, water_conservation) =
+        CA.check_conservation(sol)
+
+    @info "    Net energy change / total energy: $energy_conservation"
```
szy21, line 124:
```suggestion
    @info "    Net energy change / total energy: $energy_conservation"
    @info "    Net mass change / total mass: $mass_conservation"
```

szy21 (approved):
Thanks!

src/solver/solve.jl
Sbozzolo: Yes, I think you are right

PR #3124, opened 2024-06-21
limit temperature for radiation

szy21:
## Purpose 

Limit the temperature input to RRTMGP so that they fall into the lookup table range

## To-do

- [ ] Use ClimaParams

## Content

----


Sbozzolo:
This should be handled by RRMTPGP

szy21:
I will keep the limit here for now, but I opened an issue in rrtmgp: https://github.com/CliMA/RRTMGP.jl/issues/509.

PR #3126, opened 2024-06-22
Add cloud diagnostics for diagnostic edmf

trontrytel:
There is some repetition between the quadrature and diagnostic edmf optiion. I could maybe try to combine them. Tho on the other hand dispatching is probably the more elegant way to go

### DIFF
--- a/src/cache/cloud_fraction.jl
+++ b/src/cache/cloud_fraction.jl
@@ -66,7 +66,7 @@ NVTX.@annotate function set_cloud_fraction!(
 
     FT = eltype(params)
     thermo_params = CAP.thermodynamics_params(params)
-    (; ᶜts, ᶜp, ᶜmixing_length, cloud_diagnostics_tuple) = p.precomputed
+    (; ᶜts, ᶜmixing_length, cloud_diagnostics_tuple) = p.precomputed
     (; turbconv_model) = p.atmos
     if isnothing(turbconv_model)
         if p.atmos.call_cloud_diagnostics_per_stage isa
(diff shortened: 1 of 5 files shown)


src/cache/cloud_fraction.jl
```diff
@@ -94,6 +94,57 @@ NVTX.@annotate function set_cloud_fraction!(
...
+        ᶜmixing_length,
+        thermo_params,
+    )
+    # updrafts
+    n = n_mass_flux_subdomains(p.atmos.turbconv_model)
+
+    for j in 1:n
+        @. cloud_diagnostics_tuple += NamedTuple{(:cf, :q_liq, :q_ice)}(
+            tuple(
+                ifelse(TD.has_condensate(thermo_params, ᶜtsʲs.:($$j)), 1, 0),
```
szy21, line 139:
```suggestion
                draft_area(ᶜρaʲs.:($$j), ᶜρʲs.:($$j)) * ifelse(TD.has_condensate(thermo_params, ᶜtsʲs.:($$j)), 1, 0),
```

src/cache/cloud_fraction.jl
```diff
@@ -94,6 +94,57 @@ NVTX.@annotate function set_cloud_fraction!(
...
+    if p.atmos.call_cloud_diagnostics_per_stage isa CallCloudDiagnosticsPerStage
+        (; ᶜgradᵥ_θ_virt, ᶜgradᵥ_q_tot, ᶜgradᵥ_θ_liq_ice) = p.precomputed
+        thermo_params = CAP.thermodynamics_params(p.params)
+        @. ᶜgradᵥ_θ_virt =
+            ᶜgradᵥ(ᶠinterp(TD.virtual_pottemp(thermo_params, ᶜts)))
+        @. ᶜgradᵥ_q_tot =
+            ᶜgradᵥ(ᶠinterp(TD.total_specific_humidity(thermo_params, ᶜts)))
+        @. ᶜgradᵥ_θ_liq_ice =
+            ᶜgradᵥ(ᶠinterp(TD.liquid_ice_pottemp(thermo_params, ᶜts)))
+    end
```
szy21, line 120: This block should not be needed, as for diagnostic edmf these quantities are already calculated at each stage [here](https://github.com/CliMA/ClimaAtmos.jl/blob/bfabc07266a1a561103d671819115c81cebfac21/src/cache/precomputed_quantities.jl#L484). It might be good to add an assert somewhere so that we only use cloud_model "diagnostic_edmf" when turbconv is "diagnostic_edmfx". (We should unify the name at some point...)

szy21 (approved):
Thank you!

src/cache/cloud_fraction.jl
trontrytel: Agreed. I don't think this should even be an option for the user. Instead it should be always this way when running with diagnostic edmf. It's just that I wasn't sure how well it will be working in the model and wanted to have it as a temporary option.

PR #3130, opened 2024-06-24
Add dycoms RF02 to CI

trontrytel:
## Content

----


szy21 (approved):
Should we just change the name for rad to DYCOMS (similar to what we have for subsidence) to make it more clear?

PR #3131, opened 2024-06-24
Update to RRTMGP v0.17.0

nefrathenrici:
The only change in RRTMGP has been the addition of aerosol optics, requiring an additional field.


szy21:
I think maybe we should make `center_relative_humidity` keyword arguments for non gray radiation. I will give it a try.

PR #3133, opened 2024-06-24
And check conservation of water also without PrognosticSurfaceTemperature

Sbozzolo:
And check conservation of water also without PrognosticSurfaceTemperature

### DIFF
--- a/src/solver/solve.jl
+++ b/src/solver/solve.jl
@@ -241,26 +241,20 @@ function check_conservation(sol)
             energy_radiation_input,
         ) / energy_total
 
-    mass_conservation =
-        (sum(sol.u[end].c.ρ) - sum(sol.u[1].c.ρ)) / sum(sol.u[1].c.ρ)
-
-    water_conservation = zero(Spaces.undertype(axes(sol.u[end].c.ρ)))
+    water_surface_change = zero(Spaces.undertype(axes(sol.u[end].c.ρ)))
+    water_total = sum(sol.u[end].c.ρq_tot)
+    water_atmos_change = sum(sol.u[end].c.ρq_tot) - sum(sol.u[1].c.ρq_tot)
 
     if sfc isa PrognosticSurfaceTemperature
-        # water
-        water_total = sum(sol.u[end].c.ρq_tot)
-        water_atmos_change = sum(sol.u[end].c.ρq_tot) - sum(sol.u[1].c.ρq_tot)
         water_surface_change = horizontal_integral_at_boundary(
             sol.u[end].sfc.water .- sol.u[1].sfc.water,
         )
-
-        mass_conservation =
-            (sum(sol.u[end].c.ρ) - sum(sol.u[1].c.ρ) + water_surface_change) /
-            sum(sol.u[1].c.ρ)
-
-        water_conservation =
-            abs(water_atmos_change + water_surface_change) / water_total
     end
+    mass_conservation =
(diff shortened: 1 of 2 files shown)


src/solver/solve.jl
```diff
@@ -242,21 +242,19 @@ function check_conservation(sol)
         ) / energy_total
 
     mass_conservation =
-        (sum(sol.u[end].c.ρ) - sum(sol.u[1].c.ρ)) / sum(sol.u[1].c.ρ)
+        (sum(sol.u[end].c.ρ) - sum(sol.u[1].c.ρ)) / sum(sol.u[begin].c.ρ)
```
szy21, line 245:
```suggestion
        (sum(sol.u[end].c.ρ) - sum(sol.u[begin].c.ρ)) / sum(sol.u[begin].c.ρ)
```
Also @dennisYatunin prefers 1 over begin. I don't have a preference as long as we are consistent:)

src/solver/solve.jl
```diff
@@ -242,21 +242,19 @@ function check_conservation(sol)
...
 
-    water_conservation = zero(Spaces.undertype(axes(sol.u[end].c.ρ)))
+    water_surface_change = zero(Spaces.undertype(axes(sol.u[end].c.ρ)))
+    water_total = sum(sol.u[begin].c.ρq_tot)
+    water_atmos_change = sum(sol.u[end].c.ρq_tot) - sum(sol.u[1].c.ρq_tot)
```
szy21, line 249:
```suggestion
    water_atmos_change = sum(sol.u[end].c.ρq_tot) - sum(sol.u[begin].c.ρq_tot)
```

szy21:
The issue with using t=0 is that `water_total` will be zero for most of the cases (where the initial condition is dry).

Sbozzolo:
> The issue with using t=0 is that `water_total` will be zero for most of the cases (where the initial condition is dry).

Okay, it makes sense.

Sbozzolo:
> The issue with using t=0 is that `water_total` will be zero for most of the cases (where the initial condition is dry).

I rolled back to checking with `end` (even if I think we should not do that), and singled out the change to check water conservation also without PrognosticSurfaceTemperature.

szy21 (approved):
Looks good. Just keep in mind that when sfc is not Prognostic, water conservation will not hold when there is source or sink of moisture (evaporation or precipitation).

src/solver/solve.jl
```diff
@@ -241,26 +241,20 @@ function check_conservation(sol)
             energy_radiation_input,
         ) / energy_total
 
-    mass_conservation =
-        (sum(sol.u[end].c.ρ) - sum(sol.u[1].c.ρ)) / sum(sol.u[1].c.ρ)
-
-    water_conservation = zero(Spaces.undertype(axes(sol.u[end].c.ρ)))
+    water_surface_change = zero(Spaces.undertype(axes(sol.u[end].c.ρ)))
+    water_total = sum(sol.u[1].c.ρq_tot)
```
szy21, line 245:
```suggestion
    water_total = sum(sol.u[end].c.ρq_tot)
```
This should fix the error here: https://buildkite.com/clima/climaatmos-ci/builds/19579#01907544-653d-482c-ae12-4c964b6778c9/200-485

szy21:
Actually I want to output water conservation check without PrognosticSurfaceTemperature as well. I'll push to this PR.

PR #3134, opened 2024-06-24
add RCEMIPII sphere SST

szy21:
## Purpose 

co-authored-by: @cmschmitt519 

## Content

----


src/surface_conditions/surface_conditions.jl
```diff
@@ -334,6 +334,13 @@ function surface_state_to_conditions(
     )
 end
 
+function surface_temperature(::RCEMIPIISphereSST, coordinates)
+    (; lat) = coordinates
+    FT = eltype(lat)
+    T = FT(300) + FT(1.25) / 2 * cosd(360 * lat / 54)
+    return T
+end
```
Sbozzolo, line 342: Can you please add references so that one can find where these numbers come from?

szy21: Yes, we will also make some of them parameters in a future PR.

PR #3147, opened 2024-06-26
add aerosol radiation

szy21:
## Purpose 

Add aerosols to radiation. For now we assume there is only sulfate aerosol.

## Content

----

### DIFF
--- a/src/callbacks/callbacks.jl
+++ b/src/callbacks/callbacks.jl
@@ -61,6 +61,7 @@ NVTX.@annotate function rrtmgp_model_callback!(integrator)
     (; params) = p
     (; idealized_insolation, idealized_h2o, idealized_clouds) = p.radiation
     (; ᶠradiation_flux, radiation_model) = p.radiation
+    (; radiation_mode) = p.atmos
 
     # If we have prescribed aerosols, we need to update them
     for (key, tv) in pairs(p.tracers.prescribed_aerosol_timevaryinginputs)
@@ -85,7 +86,7 @@ NVTX.@annotate function rrtmgp_model_callback!(integrator)
     @. ᶜT =
         min(max(TD.air_temperature(thermo_params, ᶜts), FT(T_min)), FT(T_max))
 
-    if !(radiation_model.radiation_mode isa RRTMGPI.GrayRadiation)
+    if !(radiation_mode isa RRTMGPI.GrayRadiation)
         ᶜrh = Fields.array2field(
             radiation_model.center_relative_humidity,
             axes(Y.c),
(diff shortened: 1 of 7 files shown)


src/callbacks/callbacks.jl
```diff
@@ -157,6 +158,17 @@ NVTX.@annotate function rrtmgp_model_callback!(integrator)
...
 
+    if !(radiation_mode isa RRTMGPI.GrayRadiation)
+        if radiation_mode.aerosol_radiation
+            ᶜΔz = Fields.local_geometry_field(Y.c).∂x∂ξ.components.data.:9
+            ᶜaero_conc = Fields.array2field(
+                radiation_model.center_aerosol_column_mass_density,
+                axes(Y.c),
+            )
+            @. ᶜaero_conc = p.tracers.prescribed_aerosol_fields.:SO4 * ᶜΔz
+        end
```
nefrathenrici, line 169: Will this fail if SO4 is not added in the config? We could add a check and error if the aerosol configuration is not properly set.

nefrathenrici (approved):
Thank you! I just left one comment.

src/callbacks/callbacks.jl
szy21: Yes, I was thinking about what is the best way to do that. Maybe I will add an assert for aerosol_radiation and prescribed_aerosols. I will make some changes shortly and let you take a look again.

szy21:
@nefrathenrici I added an assert when building radiation cache, so that we only need to check it once at the beginning of the simulation. I need to pass in aerosol_names to radiation cache for this, but this is the best I can think of. What do you think?

src/parameterized_tendencies/radiation/radiation.jl
```diff
@@ -32,14 +32,21 @@ function radiation_model_cache(
...
 )
     context = ClimaComms.context(axes(Y.c))
     device = context.device
     (; idealized_h2o, idealized_insolation, idealized_clouds) = radiation_mode
+    if !(radiation_mode isa RRTMGPI.GrayRadiation)
+        (; aerosol_radiation) = radiation_mode
+        if aerosol_radiation
+            @assert "SO4" in aerosol_names
+        end
+    end
```
nefrathenrici, line 49:
```suggestion
    (; aerosol_radiation) = radiation_mode
    if !(radiation_mode isa RRTMGPI.GrayRadiation) &&
       aerosol_radiation &&
       !("SO4" in aerosol_names)
        error("Aerosol radiation is turned on but SO4 is not provided")
    end
```
I think a more informative error would be helpful, here is what the snippet could look like

szy21: Thanks. I made the change except for moving (; aerosol_radiation) = radiation_mode in the if block, because GrayRadiation mode doesn't contain aerosol_radiation as an argument.

PR #3156, opened 2024-06-28
add lookup table for aerosols

### LINKED ISSUE #3154: add aerosol lookup table to RRTMGP interface
#3147 adds aerosol state to the RRTMGPInterface, but didn't add the aerosol look up table, so the aerosols still don't affect radiation.

szy21:
## Purpose 

Closes #3154 

## Content

----


nefrathenrici (approved):
Thank you! Will we always want to use the Merra lookup table?

szy21:
> Thank you! Will we always want to use the Merra lookup table?

I think that's the only one we have now.

PR #3157, opened 2024-06-28
Added RCEMIPII plane sst

cmschmitt519:
## Purpose 

Add RCEMIPII plane sst distribution

## Content

Previously, all box models with prescribed surface temp were automatically set to 300K. I kept this logic except for the RCEMIPII case

----

### DIFF
--- a/config/model_configs/rcemipii_box_diagnostic_edmfx.yml
+++ b/config/model_configs/rcemipii_box_diagnostic_edmfx.yml
@@ -1,10 +1,11 @@
 surface_setup: DefaultMoninObukhov
 surface_temperature: RCEMIPII
+insolation: rcemipii
+config: box
 rad: allskywithclear
 turbconv: diagnostic_edmfx
 implicit_diffusion: true
 approximate_linear_solve_iters: 2
-insolation: rcemipii
 prognostic_tke: true
 edmfx_upwinding: first_order
 edmfx_entr_model: "Generalized"
@@ -15,8 +16,15 @@ edmfx_sgs_diffusive_flux: true
 moist: equil
 precip_model: 0M
 override_τ_precip: false
-dt: 100secs
-t_end: 12hours
+dt: 30secs
+t_end: 3600secs
 dt_save_state_to_disk: 12hours
 toml: [toml/diagnostic_edmfx_0M.toml]
+netcdf_interpolation_num_points: [8, 8, 60]
+netcdf_output_at_levels: true
+diagnostics:
+  - short_name: [ts, ta, thetaa, ha, pfull, rhoa, ua, va, wa, hur, hus, cl, clw, cli, hussfc, evspsbl, pr]
+    period: 5mins
+  - short_name: [arup, waup, taup, thetaaup, haup, husup, hurup, clwup, cliup, waen, tke, lmix]
+    period: 5mins
 ode_algo: ARS343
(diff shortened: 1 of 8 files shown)


src/solver/model_getters.jl
```diff
@@ -36,14 +36,20 @@ end
...
+        "RCEMIPIIPlane",
+    )
     return if surface_temperature == "ZonallyAsymmetric"
         ZonallyAsymmetricSST()
     elseif surface_temperature == "ZonallySymmetric"
         ZonallySymmetricSST()
     elseif surface_temperature == "RCEMIPIISphere"
         RCEMIPIISphereSST()
+    elseif surface_temperature == "RCEMIPIIPlane"
+        RCEMIPIIPlaneSST()
```
szy21, line 52: I think it would be better to call it RCEMIPIIBoxSST, as it is for a box setup.

config/model_configs/rcemipii_box_diagnostic_edmfx.yml
szy21, line 1: You need to add this to .buildkite/pipeline.yml to run it in ci. Also add it to ci_plots.jl if you want to see plots in ci.

szy21 (commented):
This looks good. But I think it would be better if we make the surface_temperature function dispatch on both sfc_temperature and model_config. This way we can have only one RCEMIPII surface temperature type, but work for the sphere or box. It also help avoid the case that e.g. some users use the sphere SST for a box setup. If it is too complicated, we can get this in and refactor it in a separate PR.

src/surface_conditions/surface_conditions.jl
```diff
@@ -170,9 +170,14 @@ function surface_state_to_conditions(
...
-            coordinates isa Geometry.LatLongPoint
+            coordinates isa Geometry.LatLongPoint ||
+            atmos.sfc_temperature isa RCEMIPIISST
         )
-            surface_temperature(atmos.sfc_temperature, coordinates)
+            surface_temperature(
+                atmos.sfc_temperature,
+                atmos.model_config,
+                coordinates,
+            )
```
Sbozzolo, line 180: `atmos.model_config` carries redundant information (we can find if we are on a Box or on a Sphere from the `Space`) and I would like to eventually get rid of it.

Given this, and the fact that the chain of conditionals in this function started being very complex, could you try rewriting this function so that it is easier to understand what it does and it does not use `model_config`?

Sbozzolo: `eltype` of `coordiantes` will be `LatLongPoint` for a sphere and `XYPoint` for a box.

szy21: Oh that's a good point. We can dispatch on coordinates then.

akshaysridhar (approved):
Pending buildkite check; but code changes LGTM. Thanks @cmschmitt519.

src/surface_conditions/surface_conditions.jl
```diff
@@ -335,21 +329,51 @@ function surface_state_to_conditions(
...
+    (; x) = coordinates
+    FT = eltype(x)
+    T = FT(300) + FT(1.25) / 2 * cos(2 * FT(pi) * x / 6000)
+    return T
+end
+
+#For non-RCEMIPII box models with prescribed surface temp, assume that the latitude is 0.
+function surface_temperature(
+    ::Union{ZonallySymmetricSST, ZonallyAsymmetricSST},
+    coordinates::Union{Geometry.XYPoint, Geometry.XYZPoint},
```
akshaysridhar, line 356: This also needs to work with `XZPoint` to support the "2D" extruded configuration.

PR #3159, opened 2024-06-29
move surface tendency after precipitation tendency

### LINKED ISSUE #3113: Water is not conserved when precipitation is present
The simplest case to reproduce this is possibly enabling `PrognosticSurfaceTemperature` is the `single_column_precipitation_test` case. Steps to reproduce:
```julia
import ClimaAtmos as CA
import SciMLBase
config = CA.AtmosConfig("configs/single_column_precipitation_test.yml")
config.parsed_args["prognostic_surface"] = "PrognosticSurfaceTemperature"
simulation = CA.get_simulation(config)
initial_water = sum(simulation.integrator.u.c.ρq_tot)
SciMLBase.step!(simulation.integrator)
water_sfc = CA.horizontal_integral_at_boundary(simulation.integrator.u.sfc.water)
water_tot = sum(simulation.integrator.u.c.ρq_tot)

println((water_sfc + water_tot - initial_water)/ water_tot)
```
This should be round-off, but it is not.

We verified that water is well conserved in absence of precipitation.

Investigation started by 

> See these two builds with ClimaAtmos v0.26.0:
> ARS: https://buildkite.com/clima/climaatmos-gpulongruns/builds/245#01903269-239f-4966-8cc9-f7d6dc2c2570/155-574
> net water change: 5.4187683e-5
> SSP: https://buildkite.com/clima/climaatmos-gpulongruns/builds/244#019031bf-c8a3-4be3-861e-6eb54f165278/155-576
> net water change: 0.23737198 (!)

> @dennisYatunin @Sbozzolo Could you take a look if the test is wrong, or if there is something with SSP?

szy21:
## Purpose 

Move surface tendency after precipitation tendency so that precipitation used in atmos and surface tendencies are in sync. Fixes water conservation issue with SSP. After this change, SSP conserves water but ARS doesn't (2% error in 100 days): https://buildkite.com/clima/climaatmos-gpulongruns/builds/264#_. We should look into ARS if we want to use it.

I don't like that we need to have a specific order of tendencies for the test to work, but this is a quick fix and refactoring precipitation needs more work. cc @trontrytel . We should check what the coupler is doing.

Also changes energy conservation to use the initial condition as the reference. Only water conservation uses the final state as the reference, because the initial condition can be zero.

Closes #3113 

## Content

----

### DIFF
--- a/src/prognostic_equations/remaining_tendency.jl
+++ b/src/prognostic_equations/remaining_tendency.jl
@@ -24,7 +24,6 @@ end
 
 NVTX.@annotate function additional_tendency!(Yₜ, Y, p, t)
     viscous_sponge_tendency!(Yₜ, Y, p, t, p.atmos.viscous_sponge)
-    surface_temp_tendency!(Yₜ, Y, p, t, p.atmos.surface_model)
 
     # Vertical tendencies
     rayleigh_sponge_tendency!(Yₜ, Y, p, t, p.atmos.rayleigh_sponge)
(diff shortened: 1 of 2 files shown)


szy21:
cc @akshaysridhar

Sbozzolo (approved):

PR #3164, opened 2024-07-01
GCM-driven SCM modifications 

costachris:
Change LES default imin index, expose surface area parameter in diagnostic EDMF, set top cell forcing tendency to zero.

### DIFF
--- a/src/initial_conditions/initial_conditions.jl
+++ b/src/initial_conditions/initial_conditions.jl
@@ -1177,7 +1177,7 @@ end
 function gcm_initial_conditions(external_forcing_file)
     NC.NCDataset(external_forcing_file) do ds
         (  # TODO: Cast to CuVector for GPU compatibility
-            gcm_driven_profile(ds, "thetali_mean")[:, 1],  # 1 is initial time index
+            gcm_driven_profile(ds, "thetali_mean")[:, 1], # 1 is initial time index
             gcm_driven_profile(ds, "u_mean")[:, 1],
             gcm_driven_profile(ds, "v_mean")[:, 1],
             gcm_driven_profile(ds, "qt_mean")[:, 1],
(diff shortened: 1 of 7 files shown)


src/initial_conditions/initial_conditions.jl
```diff
@@ -1177,7 +1177,7 @@ end
 function gcm_initial_conditions(external_forcing_file)
     NC.NCDataset(external_forcing_file) do ds
         (  # TODO: Cast to CuVector for GPU compatibility
-            gcm_driven_profile(ds, "thetali_mean")[:, 1],  # 1 is initial time index
+            gcm_driven_profile(ds, "theta_mean")[:, 1],
```
szy21, line 1180: I think `thetali_mean` is more correct here, because you are using `PhaseEquil_ρθq` to construct the initial condition thermo state, where θ is liquid ice potential temperature.

src/prognostic_equations/forcing/external_forcing.jl
```diff
@@ -182,5 +181,11 @@ function external_forcing_tendency!(Yₜ, Y, p, t, ::GCMForcing)
     )
     # <-- subsidence
 
+    ρe_tot_top = Fields.level(Yₜ.c.ρe_tot, Spaces.nlevels(axes(Y.c)))
```
szy21, line 184: Could you add a comment here on why this is needed?

szy21 (approved):
Thanks!

src/prognostic_equations/forcing/external_forcing.jl
```diff
@@ -181,6 +180,12 @@ function external_forcing_tendency!(Yₜ, Y, p, t, ::GCMForcing)
         Val{:first_order}(),
     )
     # <-- subsidence
+    # needed to address top boundary condition for forcings. Otherwise upper portion of domain is anomalously warm
```
szy21, line 183:
```suggestion
    # needed to address top boundary condition for forcings. Otherwise upper portion of domain is anomalously cold
```
(it's anomalously warm with a different boundary condition, which I didn't merge.)

szy21 (approved):

PR #3167, opened 2024-07-02
rename radiation_model to rrtmgp_model

szy21:
## Purpose 

Removes radiation_mode from radiation_model as it is the same as radiation_mode in atmos. This does require an additional radiation_mode argument in some functions, but I still think it is better not to have redundant information. @Sbozzolo @dennisYatunin What do you think?

Edit: From an offline discussion we are renaming radiation_model to rrtmgp_model.

## Content

----

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,8 @@ ClimaAtmos.jl Release Notes
 
 Main
 -------
+- ![][badge-💥breaking] Change "radiation_model" in the radiation cache to "rrtmgp_model".
+  PR [#3167](https://github.com/CliMA/ClimaAtmos.jl/pull/3167)
 - ![][badge-💥breaking] Change the "idealized_insolation" argument to "insolation", 
   and add RCEMIP insolation. PR [#3150](https://github.com/CliMA/ClimaAtmos.jl/pull/3150)
 
--- a/src/cache/surface_albedo.jl
+++ b/src/cache/surface_albedo.jl
@@ -56,7 +56,7 @@ Set the surface albedo to a constant value.
 function set_surface_albedo!(Y, p, t, α_model::ConstantAlbedo{FT}) where {FT}
 
     (; direct_sw_surface_albedo, diffuse_sw_surface_albedo) =
-        p.radiation.radiation_model
+        p.radiation.rrtmgp_model
 
     @. direct_sw_surface_albedo = α_model.α
     @. diffuse_sw_surface_albedo = α_model.α
@@ -75,7 +75,7 @@ function set_surface_albedo!(
 ) where {FT}
 
     (; direct_sw_surface_albedo, diffuse_sw_surface_albedo, cos_zenith) =
-        p.radiation.radiation_model
+        p.radiation.rrtmgp_model
 
     λ = FT(0) # spectral wavelength (not used for now)
     μ = cos_zenith
--- a/src/callbacks/callbacks.jl
+++ b/src/callbacks/callbacks.jl
@@ -60,7 +60,7 @@ NVTX.@annotate function rrtmgp_model_callback!(integrator)
     (; ᶜts, cloud_diagnostics_tuple, sfc_conditions) = p.precomputed
     (; params) = p
     (; idealized_h2o, idealized_clouds) = p.radiation
-    (; ᶠradiation_flux, radiation_model) = p.radiation
+    (; ᶠradiation_flux, rrtmgp_model) = p.radiation
     (; radiation_mode) = p.atmos
 
     # If we have prescribed aerosols, we need to update them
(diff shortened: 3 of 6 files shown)


Sbozzolo:
Can you dispatch over the type of `radiation_model` instead of reintroducing `radiation_mode`? (Or viceversa, it's too confusing)

szy21:
> Can you dispatch over the type of `radiation_model` instead of reintroducing `radiation_mode`? (Or viceversa, it's too confusing)

`radiation_model` and `radiation_mode` are two different things. radiation_model only has one type which is RRTMGPModel, and it contains information needed for radiation calculation. while `radiation_mode` is the actual radiation type. `radiation_model` only exists when `radiation_mode` is `AbstractRRTMGPMode`. Would it be less confusing if I rename `radiation_model` to something like `rrtmgp_model`?

Sbozzolo:
`radiation_model and radiation_mode are two different things. radiation_model only has one type which is RRTMGPModel, and it contains information needed for radiation calculation. `

Yes, but a `radiation_model` is created with a given `radiation_mode` (and I assume it only makes sense with that `radiation_mode`?), so you either can infer your `radiation_mode` from the `radiation_model`, or you end up allowing inconsistentencies

szy21:
> `radiation_model and radiation_mode are two different things. radiation_model only has one type which is RRTMGPModel, and it contains information needed for radiation calculation. `
> 
> Yes, but a `radiation_model` is created with a given `radiation_mode` (and I assume it only makes sense with that `radiation_mode`?), so you either can infer your `radiation_mode` from the `radiation_model`, or you end up allowing inconsistentencies

`radiation_model` is created [here](https://github.com/CliMA/ClimaAtmos.jl/blob/c8a9db1afa4ddb5e041033cd63d9dbdc29ddc3e9/src/parameterized_tendencies/radiation/radiation.jl#L212), and has different keyword arguments given different types of `radiation_mode`. And currently we have `radiation_mode` in `radiation_model`, which is what I am trying to remove. I don't know if there is a easy way to infer `radiation_mode` from `radiation_model`?

Sbozzolo (approved):

src/parameterized_tendencies/radiation/RRTMGPInterface.jl
```diff
@@ -521,6 +521,16 @@ function RRTMGPModel(
                 end
                 lookups = (; lookups..., lookup_lw_cld)
             end
+            if radiation_mode.aerosol_radiation
+                local lookup_lw_aero
+                data_loader(joinpath("rrtmgp-aerosols-merra-lw.nc")) do ds
+                    lookup_lw_aero =
+                        RRTMGP.LookUpTables.LookUpAerosolMerra(ds, FT, DA)
+                end
```
charleskawczynski, line 529:
```suggestion
                lookup_lw_aero = data_loader(joinpath("rrtmgp-aerosols-merra-lw.nc")) do ds
                    RRTMGP.LookUpTables.LookUpAerosolMerra(ds, FT, DA)
                end
```

src/parameterized_tendencies/radiation/RRTMGPInterface.jl
```diff
@@ -582,6 +592,17 @@ function RRTMGPModel(
                 end
                 lookups = (; lookups..., lookup_sw_cld)
             end
+
+            if radiation_mode.aerosol_radiation
+                local lookup_sw_aero
+                data_loader(joinpath("rrtmgp-aerosols-merra-sw.nc")) do ds
+                    lookup_sw_aero =
+                        RRTMGP.LookUpTables.LookUpAerosolMerra(ds, FT, DA)
+                end
```
charleskawczynski, line 601:
```suggestion
                lookup_sw_aero = data_loader(joinpath("rrtmgp-aerosols-merra-sw.nc")) do ds
                      RRTMGP.LookUpTables.LookUpAerosolMerra(ds, FT, DA)
                end
```

src/parameterized_tendencies/radiation/RRTMGPInterface.jl
szy21: Thanks! I was wondering why we need a local variable here.

szy21: Oh, I don't know why it shows up in this PR. Maybe I need to rebase. This was added in #3156, following the pattern of lookup_lw_cld [here](https://github.com/CliMA/ClimaAtmos.jl/blob/e1aa08ce7a70cc96f507cd623aa25d80ee45b3fe/src/parameterized_tendencies/radiation/RRTMGPInterface.jl#L515). I will change it. Should I make similar changes to the cloud lookup tables?

szy21: When I made this change (and similar changes to cloud lookup tables) I got NaN when running jobs interactively, and it was non-deterministic. I'll keep what we have for now as this PR is just renaming things (this change is not there anymore after rebasing), but I will open a new PR on removing the local variables.

szy21: Ah, the non-deterministic result is from something else (about aerosols). I will fix that first, then change the local variables.

PR #3172, opened 2024-07-03
Update to CloudMicrophysics v0.22

trontrytel:


szy21 (approved):
Thanks! We should update the Manifest in docs/ and perf/ as well.

szy21:
It doesn't have to be in this PR, but do you want to add a regression test for 1M microphysics in case of future changes?

PR #3173, opened 2024-07-03
update aerosol in the boundary layer for radiation

### LINKED ISSUE #3168: sphere_aquaplanet_rhoe_equilmoist_allsky_gw_res is nondeterministic
See the latest build of main branch here: https://buildkite.com/clima/climaatmos-ci/builds/19580#01907559-2d36-4711-856d-c68813e7b7e9. It fails immediately but works when rerunning. We saw something similar in the coupler that 1M jobs may be nondeterministic. @trontrytel Any idea if there could be something not initialized correctly for 1M?

szy21:
## Purpose 

Update aerosol state in the isothermal boundary layer for radiation

## Content

----

### DIFF
--- a/src/parameterized_tendencies/radiation/RRTMGPInterface.jl
+++ b/src/parameterized_tendencies/radiation/RRTMGPInterface.jl
@@ -1088,6 +1088,7 @@ function update_boundary_layer!(model)
     @views as.t_lev[end, :] .= as.t_lev[end - 1, :]
     update_boundary_layer_vmr!(model.radiation_mode, as)
     update_boundary_layer_cloud!(model.radiation_mode, as)
+    update_boundary_layer_aerosol!(model.radiation_mode, as)
 end
 update_boundary_layer_vmr!(::GrayRadiation, as) = nothing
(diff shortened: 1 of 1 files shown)


src/parameterized_tendencies/radiation/RRTMGPInterface.jl
```diff
@@ -1118,6 +1119,27 @@ function update_boundary_layer_cloud!(cloud_state)
     @views cloud_state.cld_frac[end, :] .= cloud_state.cld_frac[end - 1, :]
 end
 
+update_boundary_layer_aerosol!(::GrayRadiation, _) = nothing
+update_boundary_layer_aerosol!(
+    ::Union{
+        ClearSkyRadiation,
+        AllSkyRadiation,
+        AllSkyRadiationWithClearSkyDiagnostics,
+    },
```
dennisYatunin, line 1128: Since this is everything that's not `GrayRadiation`, you don't need to specify this `Union` type here. The method above will catch `GrayRadiation`, and this can just catch everything else.

dennisYatunin (approved):

PR #3176, opened 2024-07-05
Update dependencies

charleskawczynski:
This PR updates the dependencies


Sbozzolo:
I think `calibration/experiments/sphere_held_suarez_rhoe_equilmoist/Manifest.toml` should not be updated. This calibration is  specific to this set of packages and it is not tested when packages are updated.

PR #3181, opened 2024-07-08
Parameterize RCEMIPII sst distribution

cmschmitt519:
## Purpose 

Add parameters for SST distribution
Modify Earth rotation and ocean albedo for RCEMIP
Keep short run configurations in model configs

## Content

----

### DIFF
--- a/src/parameters/Parameters.jl
+++ b/src/parameters/Parameters.jl
@@ -12,8 +12,12 @@ const ACAP = AbstractClimaAtmosParameters
 abstract type AbstractTurbulenceConvectionParameters end
 const ATCP = AbstractTurbulenceConvectionParameters
 
+abstract type AbstractSurfaceTemperatureParameters end
+const ASTP = AbstractSurfaceTemperatureParameters
+
 Base.broadcastable(param_set::ACAP) = tuple(param_set)
 Base.broadcastable(param_set::ATCP) = tuple(param_set)
+Base.broadcastable(param_set::ASTP) = tuple(param_set)
 
 Base.@kwdef struct TurbulenceConvectionParameters{FT} <: ATCP
     surface_area::FT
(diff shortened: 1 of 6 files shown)


src/parameters/Parameters.jl
```diff
@@ -44,6 +48,13 @@ Base.@kwdef struct TurbulenceConvectionParameters{FT} <: ATCP
     max_area_limiter_power::FT
 end
 
+Base.@kwdef struct SurfaceTemperatureParameters{FT} <: ASTP
+    SST_mean::FT
+    SST_delta::FT
+    SST_wavelength::FT
+    SST_wavelength_latitude::FT
+end
+
```
szy21, line 57: @nefrathenrici Is it better to group parameters in a struct like this or add them to ClimaAtmosParameters?

cmschmitt519: I grouped them like this so they would be broadcasted correctly into surface_state_to_conditions. ClimaAtmosParameters never gets passed into this function, where surface_temperature is called.

szy21: That makes sense, thanks!

src/surface_conditions/surface_conditions.jl
```diff
@@ -332,28 +340,33 @@ end
...
 function surface_temperature(
     ::RCEMIPIISST,
     coordinates::Union{Geometry.XZPoint, Geometry.XYZPoint},
+    surface_temp_params,
 )
     (; x) = coordinates
+    (; SST_mean, SST_delta, SST_wavelength) = surface_temp_params
     FT = eltype(x)
-    T = FT(300) + FT(1.25) / 2 * cos(2 * FT(pi) * x / 6000)
+    T = SST_mean + SST_delta / 2 * cos(2 * FT(pi) * x / SST_wavelength)
```
szy21, line 361:
```suggestion
    T = SST_mean - SST_delta / 2 * cos(2 * FT(pi) * x / SST_wavelength)
```
(It's a minus sign in Eq. 1 in Wing et al. 2023?)

toml/rcemipii_diagnostic_edmfx_0M.toml
```diff
@@ -0,0 +1,32 @@
...
+value = 0
+
+[angular_velocity_planet_rotation]
+value = 0
+
+[ocean_surface_albedo]
+value = 0.07
+
+[mean_sea_level_pressure]
+value = 101480
```
szy21, line 32: Could you add an empty line at the end of this file?

szy21 (approved):
Thank you!

PR #3187, opened 2024-07-10
add aerosol diagnostics

szy21:
## Content

----

### DIFF
--- a/src/diagnostics/tracer_diagnostics.jl
+++ b/src/diagnostics/tracer_diagnostics.jl
@@ -1,0 +1,98 @@
+# This file is included in Diagnostics.jl
+
+# Tracers
+
+function compute_aerosol!(out, state, cache, time, aerosol_name)
+    :prescribed_aerosols_field in propertynames(cache.tracers) ||
+        error("Aerosols do not exist in the model")
+    aerosol_name in propertynames(cache.tracers.prescribed_aerosols_field) ||
+        error("$aerosol_name does not exist in the model")
+    if isnothing(out)
+        return copy(
+            getproperty(cache.tracers.prescribed_aerosols_field, aerosol_name),
+        )
+    else
+        out .=
+            getproperty(cache.tracers.prescribed_aerosols_field, aerosol_name)
+    end
+end
+
+###
+# Dust concentration (3d)
+###
+add_diagnostic_variable!(
+    short_name = "mmrdust",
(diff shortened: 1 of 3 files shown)


src/diagnostics/tracer_diagnostics.jl
szy21, line 1: @Sbozzolo Could you take a look at if this pattern looks ok before I add all the aerosols?

src/diagnostics/tracer_diagnostics.jl
```diff
@@ -0,0 +1,38 @@
+# This file is included in Diagnostics.jl
+
+# Tracers
+
+###
+# Downwelling shortwave radiation (3d)
+###
```
Sbozzolo, line 7: Plase, update the description

src/diagnostics/tracer_diagnostics.jl
Sbozzolo: The structure of aerosol is very consistent, so we should just target all the aerosol in one go. Let's not write a compute function for each aerosol, but one that can work generically. The might might be what you already have but also take the aerosol name, then, in `compute!` you return something like `compute! = (out, u, p, t) -> compute_aerosol(out, u, p, t, :SO4)`

src/diagnostics/tracer_diagnostics.jl
```diff
@@ -0,0 +1,38 @@
...
+    out,
+    state,
+    cache,
+    time,
+    tracers,
+)
+    if !(:prescribed_aerosols_field in propertynames(tracers))
+        error("Aerosols do not exist in the model")
+    elseif !(:SO4 in propertynames(tracers.prescribed_aerosols_field))
+        error("SO4 does not exist in the model")
```
Sbozzolo, line 21:
```suggestion
    :prescribed_aerosols_field in propertynames(tracers) || error("Aerosols do not exist in the model")
    :SO4 in propertynames(tracers.prescribed_aerosols_field) || error("SO4 does not exist in the model")
```

Sbozzolo: Instead of a single if-else, you can have a first step where you check consisntency, and then the rest of the function assumes that you are in a good place

szy21:
@Sbozzolo Would you like to take a look again?

Sbozzolo (approved):

PR #3201, opened 2024-07-17
clean up longrun

szy21:
## Purpose 

There is a requirement that we should focus on global runs and use consistent numerics configs and parameters. This PR cleans up the longruns to help us focus on the most important simulations. Specific changes are:

- Change radiation to "allskywithclear"
- Remove custom smoothing_order config
- Remove the SLEVE coordinate. We should decide which coordinate we want to use and only maintain that.
- Change t_end to 360 days, or whatever we can get within 1 day wallclock time 

In general, the atmos longruns should support our target AMIP runs listed here: https://github.com/CliMA/ClimaCoupler.jl/issues/900. If more simulations in the hierarchy are needed, we can talk about whether we should include them in the longrun.

Note that the stability of the simulations after these changes hasn't been tested.

cc @akshaysridhar @trontrytel 

## Content

----


docs/src/longruns.md
```diff
@@ -13,6 +18,11 @@ longrun_bw_rhoe_equil_highres
 Moist baroclinic wave. Test the moist dycore initial value problem.
 ```
 ```
+longrun_bw_rhoe_equil_veryhighres
```
akshaysridhar, line 21: Can we please change the name from `veryhighres` to `<xx>km` or something similar?

PR #3204, opened 2024-07-17
Fix precomputed quantities for Rosenbrock timestepper

dennisYatunin:
## Purpose 

This PR tries out CliMA/ClimaTimeSteppers.jl#305 in ClimaAtmos.jl

----


Project.toml
```diff
@@ -52,7 +52,7 @@ ClimaComms = "0.6.2"
 ClimaCore = "0.14.6"
 ClimaDiagnostics = "0.2"
 ClimaParams = "0.10.8"
-ClimaTimeSteppers = "0.7.32"
```
szy21, line 55: Do we need to keep 0.7.32?

szy21 (approved):
We should update the Manifest in docs/ too

PR #3229, opened 2024-08-12
Use Rayleigh sponge and interactive surface fluxes for gcm-driven EDMF setup

costachris:
Modify default gcm-driven EDMF setup to use Rayleigh sponge and interactive surface fluxes.

### DIFF
--- a/config/model_configs/prognostic_edmfx_gcmdriven_column.yml
+++ b/config/model_configs/prognostic_edmfx_gcmdriven_column.yml
@@ -8,6 +8,7 @@ implicit_diffusion: true
 implicit_sgs_advection: false
 approximate_linear_solve_iters: 2
 edmfx_upwinding: first_order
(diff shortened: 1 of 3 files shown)


szy21 (approved):

Julians42 (commented):
A little late to the game but looks good! One question, probably for @szy21, is that we set the rayleigh sponge height to start at 20km but searching through tomls it seems like most others set it to 30km or even 35km. Wondering if we should be following the same? I only checked the 20km to make sure it was stable and then moved on. Wondering if that might at all influence things?...

szy21:
We set it to 35 km for the 55 km domain top. For the single-column case that has a 40 km domain top, around 20 km is fine. We can fine-tune it later.

Julians42:
Ah ok that makes sense/seems reasonable. Thanks for explaining!

PR #3230, opened 2024-08-12
Use sgs quadrature fields for cloud diagnostics

costachris:
Use subdomain values (with quadratures in the environment) to compute cloud properties for prognostic EDMF.

### DIFF
--- a/src/cache/cloud_fraction.jl
+++ b/src/cache/cloud_fraction.jl
@@ -68,6 +68,7 @@ NVTX.@annotate function set_cloud_fraction!(
     thermo_params = CAP.thermodynamics_params(params)
     (; ᶜts, ᶜmixing_length, cloud_diagnostics_tuple) = p.precomputed
     (; turbconv_model) = p.atmos
+
     if isnothing(turbconv_model)
         if p.atmos.call_cloud_diagnostics_per_stage isa
            CallCloudDiagnosticsPerStage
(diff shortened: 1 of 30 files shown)


src/cache/cloud_fraction.jl
```diff
@@ -140,6 +155,56 @@ NVTX.@annotate function set_cloud_fraction!(
             ),
         )
     end
+
+end
+
+NVTX.@annotate function set_cloud_fraction!(
```
szy21, line 161: Does it not work if you combine this with the function for diagnostic edmf?

costachris: The issue is that the existing diagnostic EDMF `set_cloud_fraction!` function uses precomputed quantities (ie. `ᶜρaʲs` for updrafts) which aren't computed for prognostic edmf. To have a combined function I needed if/else statements for basically every operation in the function to handle these, so ended up just dispatching with 2 functions.

szy21: Ah, right, I forgot about that. Something like this might work if you want to give it a try: https://github.com/CliMA/ClimaAtmos.jl/blob/66ad251dc5c9e7550ce0a2bf943e9717d0b20ab8/src/prognostic_equations/edmfx_tke.jl#L26

szy21 (approved):
Just a suggestion on trying to unify the two function, but that's up to you. Otherwise it looks good to me, thanks!

trontrytel (approved):
Thank you for adding this! I agree that it would be nice to unify prognostic vs diagnostic. But since we cache different things for the two schemes, it might be cleaner to just leave it like it is and rely on dispatching

szy21:
Could you change all the prognostic edmf case in the ci to `quadrature_sgs` as well?

PR #3233, opened 2024-08-14
update dependencies

szy21:
## Content

----


charleskawczynski (approved):

szy21:
The most recent ClimaCore patch release actually breaks some ClimaAtmos tests. Fixing it now.

charleskawczynski:
> The most recent ClimaCore patch release actually breaks some ClimaAtmos tests. Fixing it now.

Ah, bummer, thanks for fixing this!

akshaysridhar:
There are a few places in `initial_conditions` where the `integral` ops are being used - are there any interface changes from this upgrade to ClimaCore 0.14.11 ?

PR #3235, opened 2024-08-15
Add support for monthly calendar diagnostics

Sbozzolo:
Using the new schedules defined in https://github.com/CliMA/ClimaDiagnostics.jl/pull/67

Floating point errors can lead to some accumulated variables having one additional timestep.

Use in the yaml with `1months`

### DIFF
--- a/src/diagnostics/default_diagnostics.jl
+++ b/src/diagnostics/default_diagnostics.jl
@@ -5,11 +5,18 @@
 # level interfaces, add them here. Feel free to include extra files.
 
 """
-    default_diagnostics(model, t_start, t_end; output_writer)
+    default_diagnostics(model, t_start, t_end, reference_date; output_writer)
 
 Return a list of `ScheduledDiagnostic`s associated with the given `model` that use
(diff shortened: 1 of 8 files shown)


src/diagnostics/default_diagnostics.jl
```diff
@@ -5,10 +5,10 @@
 # level interfaces, add them here. Feel free to include extra files.
 
 """
-    default_diagnostics(model, t_start, t_end; output_writer)
+    default_diagnostics(model, t_start, t_end, reference_date; output_writer)
 
 Return a list of `ScheduledDiagnostic`s associated with the given `model` that use
-`output_write` to write to disk. `t_end` is the expected simulation end time and it is used
+`output_write` to write to disk. `t_end, reference_date` is the expected simulation end time and it is used
```
szy21, line 11: `reference_date` is the start date of the simulation?

szy21 (approved):
Thank you!

src/diagnostics/default_diagnostics.jl
Sbozzolo: Yes, at the end of the day.

PR #3240, opened 2024-08-15
Remove direct dependency on Colors

Sbozzolo:
Hopefully, `ClimaAtmos` does not directly require `Colors` to run


charleskawczynski:
I we do, or NVTX needs it for annotations?

charleskawczynski (approved):

PR #3242, opened 2024-08-16
Debug CI times

charleskawczynski:


charleskawczynski:
Looks like the compile times are from ClimaCore. I'll try bisecting

cc @szy21, @dennisYatunin

charleskawczynski:
I'm going to use this snippet from PkgDevTools:

```julia
function with_precompile_set(fn; precompile=false)
    # ENV does not always have JULIA_PKG_PRECOMPILE_AUTO
    # precompile_value = ENV["JULIA_PKG_PRECOMPILE_AUTO"]
    x = "JULIA_PKG_PRECOMPILE_AUTO"
    precompile₀ = haskey(ENV, x) ? ENV[x] : nothing
    if !precompile
        @info "Temporarily setting ENV[$x] = 0"
        ENV[x] = 0
    end
    fn()
    if !precompile
        if isnothing(precompile₀); pop!(ENV, x)
        else; ENV[x] = precompile₀
        end
    end
    return precompile₀
end

using Pkg
with_precompile_set() do
	Pkg.add(PackageSpec(;name="ClimaCore", rev="ef82fe7cc8bae7fc638655c8ada6ed8fc5f627bc"))
end
```

charleskawczynski:
Bummer, it does seem to be from adding `Nh` as a static parameter. I'm not really sure why.

szy21:
Thanks for looking into this!

charleskawczynski:
I think that the result of this investigation lead to 3 findings:

 - Static Nh in ClimaCore increased compile times (to be temporarily reverted)
 - https://github.com/JuliaLang/julia/issues/55807
 - And compile times in diagnostics: https://github.com/CliMA/ClimaDiagnostics.jl/issues/85 https://github.com/CliMA/ClimaDiagnostics.jl/issues/82 (with improvement https://github.com/CliMA/ClimaDiagnostics.jl/pull/86)

PR #3244, opened 2024-08-16
Add minimum exchange coeff + exponentially decaying diffusion to Frierson functions

### LINKED ISSUE #3228: Add a lower limit for diffusivity in Frierson diffusion
This will help stabilize the simulation

akshaysridhar:
Closes #3228 
[summary_2D.pdf](https://github.com/user-attachments/files/16640943/summary_2D.pdf)
[summary.pdf](https://github.com/user-attachments/files/16640944/summary.pdf)

`edt` instantaneous at upper levels shows ~ 0 diffusivity. 
Screenshot 2024-08-16 at 12 37 55 PM
Screenshot 2024-08-16 at 12 42 03 PM

### DIFF
--- a/src/cache/precomputed_quantities.jl
+++ b/src/cache/precomputed_quantities.jl
@@ -336,7 +336,7 @@ function eddy_diffusivity_coefficient(
     κ::FT,
 ) where {FT}
     # Equations (17), (18)
-    if z < f_b * h
+    if z <= f_b * h
         K_b =
             compute_surface_layer_diffusivity(z, z₀, κ, C_E, Ri, Ri_a, Ri_c, uₐ)
         return K_b
@@ -360,14 +360,13 @@ end
 
 function compute_boundary_layer_height!(
     h_boundary_layer,
-    f_b::FT,
     dz,
     Ri_local,
     Ri_c::FT,
     Ri_a,
 ) where {FT}
     nlevels = Spaces.nlevels(Spaces.axes(Ri_local))
-    for level in 1:nlevels
+    for level in 1:(nlevels - 1)
         h_boundary_layer .=
             ifelse.(
                 Fields.Field(
@@ -375,7 +374,7 @@ function compute_boundary_layer_height!(
                     axes(h_boundary_layer),
                 ) .< Ri_c,
                 Fields.Field(
-                    Fields.field_values(Fields.level(dz, level)),
+                    Fields.field_values(Fields.level(dz, level + 1)),
                     axes(h_boundary_layer),
                 ),
                 h_boundary_layer,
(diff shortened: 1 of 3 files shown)


src/cache/precomputed_quantities.jl
```diff
@@ -390,17 +390,24 @@ function compute_bulk_richardson_number(
     grav,
     z::FT,
 ) where {FT}
-    # TODO Gustiness from params
-    return (grav * z) * (θ_v - θ_v_a) / (θ_v_a * (norm_ua)^2 + FT(10))
+    # TODO Gustiness from ClimaParams
+    return (grav * z) * (θ_v - θ_v_a) / (θ_v_a * (max((norm_ua)^2, FT(10))))
```
szy21, line 394:
```suggestion
    return (grav * z) * (θ_v - θ_v_a) / (θ_v_a * (max((norm_ua)^2, FT(1))))
```

src/cache/precomputed_quantities.jl
```diff
@@ -549,11 +556,11 @@ NVTX.@annotate function set_precomputed_quantities!(Y, p, t)
         grav = CAP.grav(params)
         FT = Spaces.undertype(axes(ᶜK_h))
         z₀ = FT(1e-5)
-        Ri_c = FT(1.0)
+        Ri_c = FT(2.0)
```
szy21, line 559:
```suggestion
        Ri_c = FT(1.0)
```

src/cache/precomputed_quantities.jl
```diff
@@ -354,13 +355,12 @@ function eddy_diffusivity_coefficient(
         K = K_b * (z / f_b / h) * (1 - (z - f_b * h) / (1 - f_b) / h)^2
         return K
     else
-        return FT(0)
+        return eddy_diffusivity_coefficient(C_E, uₐ, z, ᶜp)
```
szy21, line 358: Do we need this change? If not I think it would be better not to have it, as this may case a discontinuity in diffusivity.

src/cache/precomputed_quantities.jl
```diff
@@ -602,16 +609,21 @@ NVTX.@annotate function set_precomputed_quantities!(Y, p, t)
         h_boundary_layer = f_b .* Fields.level(ᶜz, Spaces.nlevels(axes(Y.c)))
```
szy21, line 609:
```suggestion
        h_boundary_layer = f_b .* Fields.level(ᶜz, 1)
```
Let's talk about this.

szy21 (approved):

src/cache/precomputed_quantities.jl
szy21: I'm leaving soon. But I think this makes more sense because otherwise h_boundary_layer will be the highest model layer if Ri is never smaller than Ri_c.

szy21:
I left some minor comments, but feel free to merge it if you don't agree with them. We can talk later. It would be good to try this with the longrun tonight.

akshaysridhar:
@szy21 @trontrytel With the latest [commit 1998e87](https://github.com/CliMA/ClimaAtmos.jl/pull/3244/commits/1998e87182d96b3d57adbd208b5e5073c637a295)  

We get the following results (see pdf summary plots below) for the 1M microphysics test 

[summary_2D.pdf](https://github.com/user-attachments/files/16677980/summary_2D.pdf)
[summary.pdf](https://github.com/user-attachments/files/16677982/summary.pdf)

@szy21 Ri_c is now the default value; however mean surface windspeed at the 15m level appear to be ~7m/s so empirically this O(10) gust norm is perhaps reasonable (we can revisit this). With h_boundary_layer set to level 1 prior to the column-wise check, our runs blow up during the spinup period within hours of simulation startup presumably due to specific columns where we don't meet the Ri <= Ri_c criterion.

szy21:
Sounds good. I still think h_boundary = 0.1 * domain_top is too high though, and may hide some problems. So the minimum diffusivity when z > h is insufficient to stabilize it. Could you try setting h_boundary to something like 500m and see if it helps?

src/cache/precomputed_quantities.jl
```diff
@@ -334,33 +334,23 @@ function eddy_diffusivity_coefficient(
...
-        )
         K = K_b * (z / f_b / h) * (1 - (z - f_b * h) / (1 - f_b) / h)^2
         return K
     else
-        return FT(0)
+        return eddy_diffusivity_coefficient(C_E, uₐ, z, ᶜp)
     end
 end
 
 function compute_boundary_layer_height!(
```
szy21, line 352: If possible, let's make the boundary layer height the first level where Ri exceeds Ri_c.

PR #3255, opened 2024-08-20
Use convert over assert in radiation

charleskawczynski:
This is a correction to #3248. This should fix things for both Float32 / Float64.


szy21 (approved):
Thanks! I'll add a test for Float64

PR #3256, opened 2024-08-20
Adds artifact and artifact functionality for GCM forcing file

### LINKED ISSUE #2952: Use artifact for the GCM forcing file
See https://github.com/CliMA/ClimaArtifacts

Julians42:
## Purpose 
Adds a ClimaArtifact, called `cfsite_gcm_forcing`, that allows single column model runs to use forcing from the GCM at all cfsites. In doing so, this PR also adds functionality to parse yml files for artifacts.

Closes #2952

## To-do

Update `ClimaArtifacts.jl` with documentation

## Content

The external forcing file can be called directly from the `.yml` file which will fetch the artifact. The artifact has already been added to HPC. For local runs, the file will be loaded lazily. To emphasize the roll of artifacts as folders we specify the external_forcing_file as the following:
```yml
external_forcing_file: artifact"cfsite_gcm_forcing"/HadGEM2-A_amip.2004-2008.07.nc
cfsite_number : "site23"
```
where the cfsite number was added previously. 

----


src/solver/types.jl
```diff
@@ -572,7 +573,7 @@ function AtmosConfig(
     comms_ctx = isnothing(comms_ctx) ? get_comms_context(config) : comms_ctx
 
     config = config_with_resolved_and_acquired_artifacts(config, comms_ctx)
-
+    print("After config_with_resolved_and_acquired_artifacts")
```
Sbozzolo, line 576: This was left here

src/solver/types.jl
```diff
@@ -608,33 +609,37 @@ In all the other cases, return the input unchanged.
...
-    context::ClimaComms.AbstractContext,
+    context::ClimaComms.AbstractCommsContext,
 )
     matched = match(r"artifact\"([a-zA-Z0-9_]+)\"(\/.*)?", input_str)
     if isnothing(matched)
         return input_str
     else
         artifact_name, other_path = matched
-        return joinpath(@clima_artifact(artifact_name, context), other_path)
+        println(@clima_artifact(string(artifact_name), context))
```
Sbozzolo, line 619: This was left here too

Sbozzolo:
Can you please add a note about this new capability in the docs (maybe in the `configs.md` file) and in the NEWS.md?

szy21 (approved):
Looks good to me after addressing the comments. Should we add the data to ClimaArtifact?

Julians42:
@szy21 I drafted a branch yesterday to add the data to ClimaArtifacts. I was hoping you could help by pointing me towards where the data comes from and how best to document it before PRing. i.e., what information would be helpful to have in the read me on [this branch](https://github.com/CliMA/ClimaArtifacts/tree/j/add_gcm_forcing_file/cfsite_gcm_forcing)? Happy to do the documentation if you send a paper link or some notes :)

szy21:
> @szy21 I drafted a branch yesterday to add the data to ClimaArtifacts. I was hoping you could help by pointing me towards where the data comes from and how best to document it before PRing. i.e., what information would be helpful to have in the read me on [this branch](https://github.com/CliMA/ClimaArtifacts/tree/j/add_gcm_forcing_file/cfsite_gcm_forcing)? Happy to do the documentation if you send a paper link or some notes :)

I think we can document that the data is derived from HadGEM2-A and contains large-scale forcing to drive single-column models for several cfSites in July. And we can add what variables are in the file (vertical profiles of horizontal advection tendency of temperature and humidity, subsidence, cloud properties, etc.) and the time-frequency of the data (6 hourly). I can take a look at the PR once it is ready.

costachris:
> It occurred to me: should this artifact be an option at all? Will people want to use different files for this GCM forcing? If the answer is "likely not', then we shouldn't make it configurable (until such need arises)

There may be use cases for other files - this forcing file is only for the HadGEM2-A model in the month of July (there are several other models and months available from Zhaoyi's library). My understanding is the artifacts were for CI and so that users (without access to the central cluster) have access to the forgings needed to run the GCM-driven setup. I think folks at Clima (with access to Caltech HPC, myself included) will need other files, but doubtful whether users w/o Caltech HPC access will need access to all those other forcing files.

Sbozzolo:
> > It occurred to me: should this artifact be an option at all? Will people want to use different files for this GCM forcing? If the answer is "likely not', then we shouldn't make it configurable (until such need arises)
> 
> There may be use cases for other files - this forcing file is only for the HadGEM2-A model in the month of July (there are several other models and months available from Zhaoyi's library). My understanding is the artifacts were for CI and so that users (without access to the central cluster) have access to the forcings needed to run the GCM-driven setup. I think folks at Clima (with access to Caltech HPC, myself included) will need other files, but doubtful whether users w/o Caltech HPC access will need access to all those other forcing files.

Artifacts provide a machine-independent way to specify locations of data so that you can run the code on computers that are not the Caltech cluster. This is very relevant to people at CliMA too: you want your code to run on `Derecho` or `GCP` or any other machine we use. As a general rule, there should be nothing in ClimaAtmos that only people at Caltech can use (besides CI) because we want our codes to be usable and our results reproducible. 

You answered my question: if you expect to be using other models and other months, then this should be a configurable argument.

NEWS.md
```diff
@@ -4,6 +4,11 @@ ClimaAtmos.jl Release Notes
 Main
 -------
 
+v0.27.4
+-------
+- Add artifact support for GCM-driven single column
+  PR [#3256](https://github.com/CliMA/ClimaAtmos.jl/pull/3256)
```
Sbozzolo, line 10: Let's make this entry more informative. 

This PR brings a new capability to ClimaAtmos, namely, being able to decode `artifact` from YAML. Let's inform readers of this.

docs/src/config.md
```diff
@@ -43,6 +43,14 @@ dt_save_state_to_disk: "10mins"
 toml: [toml/prognostic_edmfx.toml]
 ```
 
+Keys can also point to artifacts. As artifacts are folders, we specify both the artifact name,
```
Sbozzolo, line 46: Let's add a backlink to the ClimaArtifacts repo in this paragraph

src/solver/types.jl
```diff
@@ -593,6 +598,53 @@ function AtmosConfig(
...
+    _::ClimaComms.AbstractCommsContext,
+)
+    return input
+end
+
+"""
+    config_with_resolved_and_acquired_artifacts(input_str::AbstractString, context::ClimaComms.AbstractCommsContext)
+
+Substitutes strings of the form `artifact"name"/something/else` with the actual
+artifact path.
```
Sbozzolo, line 636:
```suggestion
Substitute strings of the form `artifact"name"/something/else` with the actual
artifact path.
```

NEWS.md
Julians42: Agreed!

Sbozzolo (approved):

Sbozzolo:
Thank you! Please, clean up your commits before merging.

PR #3261, opened 2024-08-23
Try ClimaCore with FieldArrays

charleskawczynski:
Testing


charleskawczynski:
No more need for this now that we have field-ending datalayouts.

PR #3267, opened 2024-08-26
A GPU compatible version for non-orographic gravity wave parameterization

Xiaoyang-Xie:
## Purpose 
Let Nogw parameterization work on GPU

## To-do
Speed it up (Now this version now runs 50% slower than the previous version of the code.)

## Content
Replace the for-end loop structure with unrolled functions

----


Sbozzolo:
This code is really complex, has little to no comments that explain what is happening, uses advanced features and introduces a dependency on UnrolledUtilities, a package that is not documented in such a way that people can understand what it does and when/how to use it without in-person guidance. I am worried that almost no one will be able to extend and maintain this piece of code.

@dennisYatunin can you help reducing complexity and improving clarity? 

If you really want to use `UnrolledUtilities`, can you first add documentation so that people can understand what the package does and how to use it? If you want to use `UnrolledUtilities` in ClimaAtmos, the target demographics for your documentation should be graduate students / research scientists.

config/model_configs/sphere_nonorographic_gravity_wave.yml
```diff
@@ -0,0 +1,6 @@
+dt_save_state_to_disk: "400secs"
+initial_condition: "IsothermalProfile"
+t_end: "1500secs"
+hyperdiff: false
```
szy21, line 4:
```suggestion
t_end: "1500secs"
```

src/parameterized_tendencies/gravity_wave_drag/non_orographic_gravity_wave.jl
```diff
@@ -1,11 +1,17 @@
...
 import ClimaCore.Fields as Fields
 import ClimaCore.Geometry as Geometry
-import ClimaCore.Operators as Operators
+import ClimaCore.Operators as Operator
+
+Operators.placeholder_space(current_space, parent_space) =
+    current_space isa Spaces.AbstractSpace &&
+    parent_space isa Spaces.AbstractSpace &&
+    Spaces.issubspace(current_space, parent_space) ?
+    Operators.PlaceholderSpace() : current_space
```
charleskawczynski, line 14: This overrides a method in ClimaCore. If this is needed, it needs to be put into ClimaCore.

src/parameterized_tendencies/gravity_wave_drag/non_orographic_gravity_wave.jl
```diff
@@ -726,24 +502,172 @@ function non_orographic_gravity_wave_forcing(
...
+end
+
+# calculate the intermittency factor eps -> assuming constant Δc.
 function calc_intermitency(ρ_source_level, source_ampl, nk, Bsum)
     return (source_ampl / ρ_source_level / nk) / Bsum
 end
 
 function gw_average!(wave_forcing, wave_forcing_m1)
-    L1 = Operators.LeftBiasedC2F(; bottom = Operators.SetValue(0.0))
+    L1 = Operators.LeftBiasedC2F(; bottom = Operators.SetValue(0.0f0))
```
charleskawczynski, line 667: Can we not hard-code Float32 here? (and below)

src/parameterized_tendencies/gravity_wave_drag/non_orographic_gravity_wave.jl
Xiaoyang-Xie: Yes, it is needed. Otherwise the code could not run on GPU. Dennis tells me he will open a pull request in ClimaCore soon.

src/parameterized_tendencies/gravity_wave_drag/non_orographic_gravity_wave.jl
```diff
@@ -726,24 +502,173 @@ function non_orographic_gravity_wave_forcing(
...
         @. uforcing = uforcing + u_waveforcing
         @. vforcing = vforcing + v_waveforcing
 
     end
     return nothing
 end
 
-# calculate the intermittency factor eps -> assuming constant Δc.
+# Using column_accumulate function, calculate the gravity wave forcing at each point.
+function waveforcing_column_accumulate!(
```
charleskawczynski, line 514: I'm not keen on nested anonymous functions like this. Can we somehow break this giant function into smaller ones?

Xiaoyang-Xie: I've refactored part of the wave source calculation into smaller function calls. Does this approach work for you? Breaking down the unrolled_reduce function, which is a big part of the whole function, isn’t easy because it directly uses many externally passed parameters. If we avoid using anonymous functions, we’ll have to keep passing those parameters in the reduce operation, which might make the function less elegant.

charleskawczynski: `wave_source` looks great, thanks for updating that. It would be nice if it's possible to do something similar with the block inside the `unrolled_reduce(Val(nc), (mask, FT1(0.0))) do (mask, fm), (n)` block, but I understand if it's not possible.

PR #3272, opened 2024-08-29
Use accessor for ClimaAnalysis units

Sbozzolo:
https://github.com/CliMA/ClimaAnalysis.jl/pull/70 is very _slightly_ breaking if units are accessed directly (the return value might no longer be a String, but can still be printed as a string). To be safe, I changed to use the getter function.

### DIFF
--- a/examples/Manifest.toml
+++ b/examples/Manifest.toml
@@ -2,7 +2,7 @@
 
 julia_version = "1.10.4"
 manifest_format = "2.0"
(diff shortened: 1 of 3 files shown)


szy21 (approved):
I think you need to update the Manifest as well?

Sbozzolo:
> I think you need to update the Manifest as well?

I thought we were already using 0.5.7 (it was released 2 weeks ago), but clearly no. Updating it now.

PR #3276, opened 2024-09-01
Fix compats, add downgrade action, remove unused dependencies

Sbozzolo:
The compats in ClimaAtmos were largely wrong. In this commit, I fix them. I also add a new action that checks that our tests pass with the oldest possible combination of packages that is advertised to be compatible with ClimaAtmos.

Since we required ClimaComms 0.6, I also removed all the checks on that version for the import_required_backends.

In the process, I also removed dependencies that were not required.


charleskawczynski (approved):
Nice!

PR #3286, opened 2024-09-10
Try nospecialize in build_cache

charleskawczynski:


charleskawczynski:
The main issue seems to be compiling `set_precompute_quantities!`, so I'm hoping that we just need to fix inference failures/invalidations.

PR #3289, opened 2024-09-11
Add function barrier to coriolis computation

charleskawczynski:
This PR adds a function barrier to the computation of the coriolis force, which I noticed captures a variable (`f`):

```julia
        f = CAP.f_plane_coriolis_frequency(params)
        coriolis_f_plane(coord) = Geometry.WVector(f)
        ᶜf³ = @. CT3(coriolis_f_plane(ᶜcoord))
        ᶠf¹² = nothing
```
Due to this, users cannot copy-paste this into the REPL. I've also refactored that code to use `map` with an assertion so that it can be copy-pasted into the REPL. I did not yet apply these changes to the deep atmosphere, because we need to somehow avoid capturing `global_geom`.

### DIFF
--- a/src/cache/cache.jl
+++ b/src/cache/cache.jl
@@ -131,30 +131,7 @@ function build_cache(
     ᶜΦ = grav .* ᶜcoord.z
     ᶠΦ = grav .* ᶠcoord.z
 
-    if eltype(ᶜcoord) <: Geometry.LatLongZPoint
-        Ω = CAP.Omega(params)
-        global_geom = Spaces.global_geometry(axes(ᶜcoord))
-        if global_geom isa Geometry.DeepSphericalGlobalGeometry
-            @info "using deep atmosphere"
-            coriolis_deep(coord::Geometry.LatLongZPoint) = Geometry.LocalVector(
-                Geometry.Cartesian123Vector(zero(Ω), zero(Ω), 2 * Ω),
-                global_geom,
-                coord,
-            )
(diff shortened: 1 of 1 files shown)


src/cache/cache.jl
```diff
@@ -266,3 +243,36 @@ function build_cache(
...
+            coriolis_deep(coord::Geometry.LatLongZPoint) = Geometry.LocalVector(
+                Geometry.Cartesian123Vector(zero(Ω), zero(Ω), 2 * Ω),
+                global_geom,
+                coord,
+            )
+            ᶜf³ = @. CT3(CT123(coriolis_deep(ᶜcoord)))
+            ᶠf¹² = @. CT12(CT123(coriolis_deep(ᶠcoord)))
+        else
+            wvec = map(ᶜcoord) do coord
+                Ω = CAP.Omega(params)::eltype(coord)
```
Sbozzolo, line 262: Can `Ω` be moved outside of the map?

charleskawczynski: No, that will turn this into a captured variable over `Ω`.

Sbozzolo: Right! Now that you say it makes perfect sense, but my first instinct would have been to move it outside because it is a constant.

charleskawczynski: Yeah, for inference the scope is the important piece, and ideally we do not reach for anything outside the anonymous function (`do coord ... end`) to determine the return type inside the anonymous function.

charleskawczynski:
I thought this merged and accidentally pushed to this branch.. Need to reapply..

PR #3306, opened 2024-09-16
Update dependencies

charleskawczynski:

### DIFF
--- a/docs/Manifest.toml
+++ b/docs/Manifest.toml
@@ -36,24 +36,28 @@ uuid = "1520ce14-60c1-5f80-bbc7-55ef81b5835c"
 version = "0.4.5"
 
 [[deps.Accessors]]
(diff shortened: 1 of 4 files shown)


Sbozzolo (approved):

szy21:
The thermodynamics update causes behavior changes. @trontrytel Should we check the longruns (dycore and aquaplanet) before merging this?

szy21:
A few jobs in ci failed; I think it's probably from Thermodynamics. Could we double check to make sure ClimaCore doesn't change anything?

charleskawczynski:
> A few jobs in ci failed; I think it's probably from Thermodynamics. Could we double check to make sure ClimaCore doesn't change anything?

Yeah, I'm pretty sure it's thermodynamics. ClimaCore is only fixing a few inference failures.

PR #3307, opened 2024-09-16
Automatically detect restart files and add tests for restarts

Sbozzolo:
This commit adds a new option, `detect_restart_file`. When `detect_restart_file` is true, ClimaAtmos looks at the previous `output_XXXX` output folder and tries to see if there's a restart file in it. If yes, it uses it. This behavior is overridden when a restart file is explicitely passed. This commit uses and re-implments the same logic implemented in ClimaUtilities.OutputPathGenerator. (We might want to move that logic to ClimaUtilities at some point.)

This PR also adds a test to ensure that restarted simulations are identical to non-restarted ones.

This is accomplished by running a simulation for three steps and comparing it to two other cases:
1. A simulation restarted from the checkpoint produced at step 3
2. A simulation restarted from the checkpoint produced at step 2 and solved

This is a variation of #3031
Closes #3031 

This is the first step in improving compatibility with restarts, and more work is needed. There is a little bit of commented out code to loop over different configurations. I am working on it now, but I would like to start getting this test merged in, as I add more complexity and fix things.

GPU tests are commented out because of ClimaCore bug,
MPI tests are commented out because of process hanging.

I will work on this next


src/callbacks/callbacks.jl
```diff
@@ -333,7 +333,13 @@ NVTX.@annotate function save_state_to_disk_func(integrator, output_dir)
     output_file = joinpath(output_dir, "day$day.$sec.hdf5")
     comms_ctx = ClimaComms.context(integrator.u.c)
     hdfwriter = InputOutput.HDF5Writer(output_file, comms_ctx)
-    InputOutput.HDF5.write_attribute(hdfwriter.file, "time", t) # TODO: a better way to write metadata
+    # TODO: a better way to write metadata
+    InputOutput.HDF5.write_attribute(hdfwriter.file, "time", t)
+    InputOutput.HDF5.write_attribute(
+        hdfwriter.file,
+        "atmos_model_hash",
+        hash(p.atmos),
```
charleskawczynski, line 341:
```suggestion
        hash(typeof(p.atmos)),
```
I think we might want to just make sure that the types are the same? Otherwise this might be too brittle (e.g., if `p.atmos` depends on `t`/`t_0`, is `t` the restart time?)

Sbozzolo: `p.atmos` is the `AtmosModel`, so, ideally, it should pretty much only contain types used for dispatch:
```
    model_config::MC = nothing
    moisture_model::MM = nothing
    precip_model::PM = nothing
    cloud_model::CM = nothing
    call_cloud_diagnostics_per_stage::CCDPS = nothing
    forcing_type::F = nothing
    subsidence::S = nothing
    radiation_mode::RM = nothing
    ls_adv::LA = nothing
    external_forcing::EXTFORCING = nothing
    edmf_coriolis::EC = nothing
    advection_test::AT = nothing
    tendency_model::TM = nothing
    edmfx_model::EDMFX = nothing
    turbconv_model::TCM = nothing
    non_orographic_gravity_wave::NOGW = nothing
    orographic_gravity_wave::OGW = nothing
    hyperdiff::HD = nothing
    vert_diff::VD = nothing
    diff_mode::DM = nothing
    sgs_adv_mode::SAM = nothing
    viscous_sponge::VS = nothing
    rayleigh_sponge::RS = nothing
    sfc_temperature::ST = nothing
    insolation::IN = nothing
    surface_model::SM = nothing
    surface_albedo::SA = nothing
    numerics::NUM = nothing
```

If these types contain extra information, I think we should check for consistency of that too.

charleskawczynski: > `p.atmos` is the `AtmosModel`, so, ideally, it should pretty much only contain types used for dispatch:

But it's not only used for dispatch. For example, a Rayleigh sponge is a property of the atmos model, which is computed like this:

```julia
function get_rayleigh_sponge_model(parsed_args, params, ::Type{FT}) where {FT}
    rs_name = parsed_args["rayleigh_sponge"]
    return if rs_name in ("false", false)
        nothing
    elseif rs_name in ("true", true, "RayleighSponge")
        zd = params.zd_rayleigh
        α_uₕ = params.alpha_rayleigh_uh
        α_w = params.alpha_rayleigh_w
        RayleighSponge{FT}(; zd, α_uₕ, α_w)
    else
        error("Uncaught rayleigh sponge model `$rs_name`.")
    end
end
```
That means that users cannot restart with a stronger sponge (perhaps to damp late instabilities). This is just one example, but I believe that there are more submodels to the atmos model that have values.

`typeof(p.atmos)` is arguably to strict still. For example, it prohibits restarting a simulation that started with Float32 and changes to Float64. We could further transform, or write some sort of custom hash, but generally I think we'll want something a bit more flexible.

> If these types contain extra information, I think we should check for consistency of that too.

Check for what sort of consistency?

src/solver/type_getters.jl
```diff
@@ -291,13 +291,20 @@ function get_spaces_restart(Y)
...
     @assert !isnothing(restart_file)
     reader = InputOutput.HDF5Reader(restart_file, comms_ctx)
     Y = InputOutput.read_field(reader, "Y")
     t_start = InputOutput.HDF5.read_attribute(reader.file, "time")
+    atmos_model_hash_in_restart =
+        InputOutput.HDF5.read_attribute(reader.file, "atmos_model_hash")
+    if atmos_model_hash_in_restart != atmos_model_hash
+        error(
+            "Restart file $(restart_file) was constructed with a different AtmosModel",
+        )
```
charleskawczynski, line 306: I'm wondering if we can do better exception handling here. This might be a fine crude start, but it might be nice to serialize the objects we're restarting, and then a check like this could print the differences so that users better understand why this might fail.

src/solver/type_getters.jl
```diff
@@ -502,9 +509,54 @@ function get_sim_info(config::AtmosConfig)
...
+
+    if parsed_args["detect_restart_file"]
+        lowercase(parsed_args["output_dir_style"]) == "activelink" ||
+            error("detect_restart_file works only with ActiveLink")
+        # output_dir will be something like ......./output_1234
+        name_rx = r"output_(\d\d\d\d)"
+        restart_file_rx = r"day\d+\.\w+\.hdf5"
+        counter_match = match(name_rx, output_dir)
+        if !isnothing(counter_match)
+            counter = parse(Int, counter_match[1])
```
charleskawczynski, line 522: Can we add exception handling here?

src/solver/type_getters.jl
```diff
@@ -502,9 +509,54 @@ function get_sim_info(config::AtmosConfig)
...
+            if counter != 0
+                # As implemented in ClimaUtilities.OutputPathGenerator
```
charleskawczynski, line 524: Where is this implemented in ClimaUtilities? If it's not reusable, could we make it reuseable to reduce code duplication?

src/solver/type_getters.jl
```diff
@@ -502,9 +509,54 @@ function get_sim_info(config::AtmosConfig)
...
+                previous_counter_str = lpad(previous_counter, 4, "0")
+                counter_str = lpad(counter, 4, "0")
+
+                previous_folder =
+                    replace(output_dir, counter_str => previous_counter_str)
+
+                isdir(previous_folder) ||
+                    error("Could not find a folder in $(previous_folder)")
+
+                possible_restart_files = filter(
```
charleskawczynski, line 535: Can we handle the case where `possible_restart_files` returns  `nothing`?

src/solver/type_getters.jl
```diff
@@ -502,9 +509,54 @@ function get_sim_info(config::AtmosConfig)
         style = allowed_dir_styles[lowercase(requested_style)],
     )
 
+    restart_file = nothing
+
+    if parsed_args["detect_restart_file"]
```
charleskawczynski, line 514: Can we put the contents of this `if`-block into a separate function?

test/restart.jl
```diff
@@ -0,0 +1,297 @@
...
+Keyword arguments
+=================
+
+- `name` is used to collect the name of the property while we go recursively
+  over all the properties. You can pass a base name.
+- `ignore` is a collection of `Symbol`s that identify properties that are
+  ignored when walking through the tree. This is useful for properties that
+  are known to be different (e.g., `output_dir`).
+"""
+function compare(one, two; name = "", ignore = Set([]))
```
charleskawczynski, line 47: Can we not use `one` as a variable name? It's already a julia method.

charleskawczynski: This is a pretty complicated function, can we somehow break it down?

src/callbacks/callbacks.jl
Sbozzolo: > That means that users cannot restart with a stronger sponge (perhaps to damp late instabilities). This is just one example, but I believe that there are more submodels to the atmos model that have values.

That's correct. At this point, the safest thing to do is to disallow this behavior because we have no guarantee that the content of `p.atmos` would not lead to configurations that are internally inconsistent. 

I agree we want more flexibility, but I think that should be done as a second step and more carefully: first, we allow checkpoints with restrictions, and then we can work on removing those restrictions.

> Check for what sort of consistency?

`Y` and `p` have to be valid `Y` and `p` for the given `AtmosModel`. A case where they would not be valid is if you changed gravitational acceleration.

src/solver/type_getters.jl
Sbozzolo: Yes, I agree. My approach is to start with a restrict, but verified-to-be-working case, and move towards more general and flexible restarts.

This line would also probably disallow running with different versions of ClimaAtmos, which is not ideal.

src/solver/type_getters.jl
Sbozzolo: What can go wrong?

`counter_match` has to have at least one item, otherwise it would be `nothing`. It has to be parsable as integer because it matched `(\d\d\d\d)`, a collection of four numbers.

src/solver/type_getters.jl
Sbozzolo: It's implemented in `ClimaUtilities.OutputPathGenerator`, but it is inside a function that does other things.

I can easily move it out and call it from ClimaAtmos to avoid duplication

config/default_configs/default_config.yml
```diff
@@ -283,6 +283,9 @@ restart_file:
 prescribe_ozone:
   help: "Prescribe time and spatially varying ozone from a file [`false` (default), `true`]"
   value: false
+detect_restart_file:
+  help: "When true, try finding a restart file and use it to restart the simulation. Only works with ActiveLink."
```
akshaysridhar, line 287: OutputPathGenerator.ActiveLinkStyle

src/callbacks/callbacks.jl
akshaysridhar: Physics component mismatches will be caught elsewhere upon generating the `simulation` object, e.g. `prognostic_tke` or `0M / 1M`, `equil / nonequil` (each of these options requiring different variables in the state `Y`); I agree that displaying mismatched elements in these error messages would be helpful! It seems reasonable to me to restrict this first pass to focus on "history critical" elements, i.e. restrict the possibility of updating tendencies / model parameters upon restart. Do `checkpoint-restart` and more flexible `data-as-init` use-cases need to be entirely separate (e.g. higher-res interpolations or timestepper swaps)?

config/default_configs/default_config.yml
Sbozzolo: `ActiveLink` is the name of the YAML option

src/solver/type_getters.jl
Sbozzolo: I realized I could make something more robust, so this is no longer needed/used.

src/solver/type_getters.jl
Sbozzolo: I realized I could make something more robust, so this is no longer needed/used.

src/solver/type_getters.jl
charleskawczynski: Sounds good, thank you!

src/callbacks/callbacks.jl
charleskawczynski: Ok, I'm fine with this so long it also doesn't prevent someone from manually restarting without this restriction, because that would be removing existing functionality--crude as it may be.

test/restart.jl
charleskawczynski: Also, just to note, ClimaCore does define `Fields.rcompare` (which is basically used in `==` for Fields).

src/callbacks/callbacks.jl
Sbozzolo: I can turn it into a warning

test/restart.jl
Sbozzolo: > Also, just to note, ClimaCore does define `Fields.rcompare` (which is basically used in `==` for Fields).

Can I use it for approximate equality?

Sbozzolo:
I cleaned up the test a little bit. In this, I ran into several issues with `Test.jl`. Most importantly, `Test` really wants to print out stacktraces, but this is not something we want. To go around this, I resorted to using a different type of testset: a test set with a let block. This behaves differently and does not add stacktraces, but also does not print out a nice summary.

Failing tests look like this:
```

Test Failed at /home/sbozzolo/repos/ClimaAtmos.jl/test/restart.jl:33
  Expression: maximum(error) <= 100 * eps(eltype(arr1))
   Evaluated: 0.00023873926f0 <= 1.1920929f-5
     Context: name = integrator.p.precomputed.ᶜgradᵥ_θ_liq_ice.components.data.1
              what = just reading data
              config = sphere equil 1M Earth

solve!:: 0.000002 seconds (1 allocation: 400 bytes)
Test Failed at /home/sbozzolo/repos/ClimaAtmos.jl/test/restart.jl:33
  Expression: maximum(error) <= 100 * eps(eltype(arr1))
   Evaluated: 0.00018313102f0 <= 1.1920929f-5
     Context: name = integrator.p.precomputed.ᶜlinear_buoygrad
              what = reading and simulating
              config = sphere equil 1M Earth

Test Failed at /home/sbozzolo/repos/ClimaAtmos.jl/test/restart.jl:33
  Expression: maximum(error) <= 100 * eps(eltype(arr1))
   Evaluated: 0.00018306637f0 <= 1.1920929f-5
     Context: name = integrator.p.precomputed.ᶜgradᵥ_θ_virt.components.data.1
              what = reading and simulating
              config = sphere equil 1M Earth

Test Failed at /home/sbozzolo/repos/ClimaAtmos.jl/test/restart.jl:33
  Expression: maximum(error) <= 100 * eps(eltype(arr1))
   Evaluated: 0.00018306637f0 <= 1.1920929f-5
     Context: name = integrator.p.precomputed.ᶜgradᵥ_θ_liq_ice.components.data.1
              what = reading and simulating
              config = sphere equil 1M Earth
```

test/restart.jl
charleskawczynski: Hm, I don't think so (off hand), but I think it's fewer lines of code if you want to use that pattern instead

Sbozzolo:
@charleskawczynski can you take a second look when you have time?

I also couldn't figure out how to use `rcompare` for this problem.

src/solver/type_getters.jl
```diff
@@ -477,6 +482,42 @@ thermo_state_type(::EquilMoistModel, ::Type{FT}) where {FT} = TD.PhaseEquil{FT}
 thermo_state_type(::NonEquilMoistModel, ::Type{FT}) where {FT} =
     TD.PhaseNonEquil{FT}
 
+function get_restart_file(config::AtmosConfig, base_output_dir)
```
charleskawczynski, line 485: I tried refactoring this a bit, and I think there may be a missing edge case?

```julia
to_named_tuple(dict::Dict) = (; (Symbol(k) => v for (k, v) in dict)...)
get_restart_file(config::AtmosConfig, base_output_dir) =
    get_restart_file(base_output_dir; to_named_tuple(config.parsed_args)...)

function get_restart_file(base_output_dir; detect_restart_file=nothing, output_dir_style=nothing, restart_file=nothing)
    isnothing(restart_file) || return restart_file
    lowercase(output_dir_style) == "activelink" ||
        error("detect_restart_file works only with ActiveLink")
    # output_dir will be something like ABC/DEF/output_1234
    name_rx = r"output_(\d\d\d\d)"
    restart_file_rx = r"day\d+\.\w+\.hdf5"
    existing_outputs =
        filter(x -> !isnothing(match(name_rx, x)), readdir(base_output_dir))
    if !isempty(existing_outputs)
        latest_output = first(sort(existing_outputs, rev = true))
        previous_folder = joinpath(base_output_dir, latest_output)
        possible_restart_files = filter(
            f -> occursin(restart_file_rx, f),
            readdir(previous_folder),
        )
        if !isempty(possible_restart_files)
            restart_file_name =
                last(CA.sort_files_by_time(possible_restart_files))
            return joinpath(previous_folder, restart_file_name)
        end
    else
        error("??")
    end
end
```

This function no longer depends on `parsed_args`

Sbozzolo: `restart_file` is set to `nothing` at the beginning of the function and changed only when `!isempty(existing_outputs)`. Having an empty existing outputs is allowed: it means that it is the first time we run this simulation.

Sbozzolo: This rewrite is string-parsing with extra steps and it is the only function in type getters to be structured like this. If we want to do this properly and really decouple the function from argument parsing, we should use the correct type for `output_dir_style`, a subtype of `ClimaUtilities.OutputPathGenerator.OutputPathGeneratorStyle`, instead of matching a string.

Sbozzolo: I changed to not use `parsed_args` and dispatch on the type of `output_dir_style`

test/restart.jl
```diff
@@ -0,0 +1,335 @@
...
+
+- `name` is used to collect the name of the property while we go recursively
+  over all the properties. You can pass a base name.
+- `ignore` is a collection of `Symbol`s that identify properties that are
+  ignored when walking through the tree. This is useful for properties that
+  are known to be different (e.g., `output_dir`).
+
+`:rc` is some CUDA/CuArray internal object that we don't care about
+"""
+function compare(v1, v2; name = "", ignore = Set([:rc]))
```
charleskawczynski, line 100: Let's chat offline about how we can improve this function, or somehow use `Fields.rcompare`/`Fields.@rprint_diff`, or write a refactored form of this function that satisfies the needs of this test.

charleskawczynski: Some of the concrete issues are:
 - It does too many things (comparison, performs different tests, has exclusions, tracks names)
 - The macro + closure over `FAILED` makes this function not portable
 - There's code duplication in leaf-level implementations
 - The comparisons are inconsistent (some with `==` and others with `_error`)

I think we can refactor this to keep most of the same functionality, but it'd be helpful to discuss this.

test/restart.jl
```diff
@@ -0,0 +1,335 @@
...
+    else
+        moistures = ["equil"]
+        precips = ["1M"]
+        topography = "NoWarp"
+        turbconv_models = ["diagnostic_edmfx"]
+    end
+
+    for turbconv_mode in turbconv_models
+        for moisture in moistures
+            for precip in precips
```
charleskawczynski, line 178: One problem with this loop is that lots of objects inside the loop will be type-unstable, but this is arguably a problem with how we configure the driver.

I don't know if we for sure need to make changes to this loop, but I did notice that it takes quite a while to run (I ctrl-C'd after a while on my computer).

It could be an issue with allocations inside `compare`, so perhaps it'll be fine after refactoring that.

Using `let` is a bit of a trade-off: using `let` prevents leaky tests, but it also means that we need to compile the same methods multiple times.

src/solver/type_getters.jl
```diff
@@ -477,6 +482,54 @@ thermo_state_type(::EquilMoistModel, ::Type{FT}) where {FT} = TD.PhaseEquil{FT}
...
+        error("detect_restart_file works only with ActiveLink")
+    return restart_file
+end
+
+function get_restart_file(
+    output_dir_style::OutputPathGenerator.ActiveLinkStyle,
+    base_output_dir;
+    restart_file = nothing,
+    detect_restart_file = false,
+)
```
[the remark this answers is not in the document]
Sbozzolo: I like this version! One minor thing is that `@assert isdir(base_output_dir)` is not correct: it is fine to not have a `base_output_dir` if this is the first time we run this simulation (it will be created). In that case, `restart_file` should be nothing.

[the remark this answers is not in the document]
Sbozzolo: This is the updated version:
```julia
function auto_detect_restart_file(
    output_dir_style::OutputPathGenerator.ActiveLinkStyle,
    base_output_dir
)
    # if base_output_dir does not exist, we return restart_file = nothing because there is
    # no restart file to be detected
    isdir(base_output_dir) || return nothing

    # output_dir will be something like ABC/DEF/output_1234
    name_rx = r"output_(\d\d\d\d)"
    restart_file_rx = r"day\d+\.\w+\.hdf5"
    restart_file = nothing

    existing_outputs =
        filter(x -> !isnothing(match(name_rx, x)), readdir(base_output_dir))

    isempty(existing_outputs) && return nothing

    latest_output = first(sort(existing_outputs, rev = true))
    previous_folder = joinpath(base_output_dir, latest_output)
    possible_restart_files = filter(
        f -> occursin(restart_file_rx, f),
        readdir(previous_folder),
    )
    if isempty(possible_restart_files)
        @warn "Detected folder $(previous_folder), but no restart file was found"
        return nothing
    end

    restart_file_name =
        last(CA.sort_files_by_time(possible_restart_files))
    restart_file = joinpath(previous_folder, restart_file_name)
    @assert isfile(restart_file) "Restart file does not exist"

    return restart_file
end
```

I also removed the other assert because it is also fine to start a new simulation if no restart files were found

test/restart.jl
```diff
@@ -74,6 +64,11 @@ function _error(
     return error
 end
 
+function rmap(val; ignore = Set([:rc]))
+    properties = filter(x -> !(x in ignore), propertynames(val))
+    properties = propertynames(val)
+end
```
charleskawczynski, line 70: @Sbozzolo, I think that this might be leading to the Revise issue?

test/restart.jl
```diff
@@ -0,0 +1,338 @@
...
+                pass,
+                getproperty(v1, p),
+                getproperty(v2, p);
+                name = "$(name).$(p)",
+                ignore,
+            )
+        end
+    end
+    return pass
+end
```
charleskawczynski, line 119:
```suggestion
function _compare(
    pass,
    v1::T,
    v2::T;
    name,
    ignore,
) where {T <: Fields.Field}
    return pass && _compare(pass, Fields.field_values(v1), Fields.field_values(v2); name, ignore)
end
function _compare(pass, v1::T, v2::T; name, ignore) where {T}
    # Recursive case
    for p in filter(x -> !(x in ignore), propertynames(v1))
        pass &= _compare(
            pass,
            getproperty(v1, p),
            getproperty(v2, p);
            name = "$(name).$(p)",
            ignore,
        )
    end
    return pass
end
```
Can't we break down this last leaf case so that the generic recursive case is independent of ClimaCore?

test/restart.jl
Sbozzolo: Weird. I removed it, we are not even using it anymore.

test/restart.jl
Sbozzolo: This doesn't walk into the properties of the Field when Fields have multiple properties (e.g. `u.c`). We want to display difference one variable at the time because that's what's the most informative.

Sbozzolo:
This is how the output looks like in case of failure:
```
config = sphere equil 0M NoWarp
solve!:: 0.554812 seconds (2.24 k allocations: 1.059 MiB, 1.74% compilation time)
    just reading data
integrator.p.precomputed.ᶜlinear_buoygrad error: 0.00020609275
integrator.p.precomputed.ᶜgradᵥ_θ_virt.components.data.1 error: 0.00023572937
integrator.p.precomputed.ᶜgradᵥ_θ_liq_ice.components.data.1 error: 0.00023572937
    reading and simulating
solve!:: 0.000005 seconds (1 allocation: 400 bytes)
integrator.p.precomputed.ᶜlinear_buoygrad error: 0.00018343895
integrator.p.precomputed.ᶜgradᵥ_θ_virt.components.data.1 error: 0.00018323408
integrator.p.precomputed.ᶜgradᵥ_θ_liq_ice.components.data.1 error: 0.00018323408
config = box equil 0M NoWarp
solve!:: 0.135827 seconds (2.60 k allocations: 493.500 KiB, 10.25% compilation time)
    just reading data
integrator.p.precomputed.ᶜlinear_buoygrad error: 0.00043861143
integrator.p.precomputed.ᶜgradᵥ_θ_virt.components.data.1 error: 0.0004185408
integrator.p.precomputed.ᶜgradᵥ_θ_liq_ice.components.data.1 error: 0.0004185408
    reading and simulating
solve!:: 0.000002 seconds (1 allocation: 400 bytes)
integrator.p.precomputed.ᶜlinear_buoygrad error: 0.00035853966
integrator.p.precomputed.ᶜgradᵥ_θ_virt.components.data.1 error: 0.0003376229
integrator.p.precomputed.ᶜgradᵥ_θ_liq_ice.components.data.1 error: 0.0003376229
config = column equil 0M NoWarp
┌ Warning: perturb_initstate flag is ignored for single column configuration
└ @ ClimaAtmos /central/scratch/esm/slurm-buildkite/climaatmos-ci/20797/climaatmos-ci/src/solver/type_getters.jl:193
┌ Warning: Bubble correction not compatible with single column configuration. It will be switched off.
└ @ ClimaAtmos /central/scratch/esm/slurm-buildkite/climaatmos-ci/20797/climaatmos-ci/src/solver/type_getters.jl:204
solve!:: 0.064094 seconds (2.45 k allocations: 343.078 KiB, 28.11% compilation time)
    just reading data
integrator.p.precomputed.ᶜlinear_buoygrad error: 0.0002045927
integrator.p.precomputed.ᶜgradᵥ_θ_virt.components.data.1 error: 0.0002344887
integrator.p.precomputed.ᶜgradᵥ_θ_liq_ice.components.data.1 error: 0.0002344887
    reading and simulating
solve!:: 0.000003 seconds (1 allocation: 400 bytes)
integrator.p.precomputed.ᶜlinear_buoygrad error: 9.4087634e-5
integrator.p.precomputed.ᶜgradᵥ_θ_virt.components.data.1 error: 0.0001073189
integrator.p.precomputed.ᶜgradᵥ_θ_liq_ice.components.data.1 error: 0.0001073189
```

test/restart.jl
```diff
@@ -0,0 +1,330 @@
...
+    # What we can safely print without filling STDOUT
+    return print_maybe(v1 == v2, "$name differs: $v1 vs $v2")
+end
+
+function _compare(v1::T, v2::T; name, ignore) where {T <: Number}
+    # We check with triple equal so that we also catch NaNs being equal
+    return print_maybe(v1 === v2, "$name differs: $v1 vs $v2")
+end
+
+function _compare(pass, v1::T, v2::T; name, ignore) where {T <: AbstractData}
```
charleskawczynski, line 135:
```suggestion
function _compare(pass, v1::T, v2::T; name, ignore) where {T <: AbstractData{<:Real}}
```

PR #3308, opened 2024-09-17
Add aerosols to job with regression test

Sbozzolo:
RRTMGP 0.19 probably changed behavior, but we didn't catch this because we don't test for aerosols.

### DIFF
--- a/config/model_configs/sphere_aquaplanet_rhoe_equilmoist_allsky_gw_raw_zonallyasymmetric.yml
+++ b/config/model_configs/sphere_aquaplanet_rhoe_equilmoist_allsky_gw_raw_zonallyasymmetric.yml
@@ -17,4 +17,6 @@ surface_temperature: "ZonallyAsymmetric"
 moist: "equil"
 albedo_model: "RegressionFunctionAlbedo"
 regression_test: true
+aerosol_radiation: true
+prescribed_aerosols: ["CB1", "CB2", "DST01", "DST02", "DST03", "DST04", "OC1", "OC2", "SO4"]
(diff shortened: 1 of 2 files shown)


config/model_configs/sphere_aquaplanet_rhoe_equilmoist_allsky_gw_raw_zonallyasymmetric.yml
```diff
@@ -17,4 +17,6 @@ surface_temperature: "ZonallyAsymmetric"
 moist: "equil"
 albedo_model: "RegressionFunctionAlbedo"
 regression_test: true
+aerosol_radiation: true
+prescribed_aerosols: ["CB1", "CB2", "DST01", "DST02", "DST03", "DST04", "OC1", "OC2", "SO4", "SOA", "SSLT01", "SSLT02", "SSLT03", "SSLT04"]
```
szy21, line 21:
```suggestion
prescribed_aerosols: ["CB1", "CB2", "DST01", "DST02", "DST03", "DST04", "OC1", "OC2", "SO4"]
```
(Let's remove sea salt as it is wrong now, as well as SOA which we don't use.)

szy21 (approved):

config/model_configs/sphere_aquaplanet_rhoe_equilmoist_allsky_gw_raw_zonallyasymmetric.yml
Sbozzolo: Ok!

PR #3313, opened 2024-09-18
Clarify inlining of unrolled_reduce for non-orographic gravity wave

dennisYatunin:
This PR replaces the call to `unrolled_reduce(op, Val(nc), init)` in the non-orographic gravity wave parametrization with a forcibly non-inlined call to `unrolled_reduce(op, StaticOneTo(nc), init)`.

The original solution was introduced in CliMA/UnrolledUtilities.jl#10 and CliMA/UnrolledUtilities.jl#12 in order to prevent compilation of 3 unit tests in CI from hanging (`nogw_test_3d.jl`, `nogw_test_mima.jl`, and `nogw_test_single_column.jl)`, but it does not have a direct equivalent in the standard library. The new version of this unrolled function call, which achieves the same goal of preventing compilation from hanging, is exactly equivalent to `reduce(op, Base.OneTo(nc); init)` (and also to `reduce(op, StaticOneTo(nc); init)`). The new version also makes it clear that inlining of the unrolled function call is what causes compilation to hang in this case.

**_Note 1_**: This does not constitute a general rule of thumb, as it is possible to construct similar examples where not inlining is what causes compilation to hang. For example, if `itr` is a `StaticBitVector` that contains 256 `Bools`, then the following code hangs without inlining of `unrolled_reduce`:
```
unrolled_reduce(((itr′, i)->unrolled_reduce(((itr′′, j)->unrolled_reduce(((itr′′′, k)->Base.setindex(itr′′′, !(itr′′′[min(i, j, k)]), k)), StaticOneTo(length(itr′′)); init = itr′′)), StaticOneTo(length(itr′)); init = itr′)), StaticOneTo(length(itr)); init = itr)
```
Inlining has been observed to improve compilation of several unit tests in UnrolledUtilities, whereas not inlining has only been observed to improve compilation of this one particular example in ClimaAtmos. So, UnrolledUtilities will continue to inline everything by default, and `@noinline` can be used on a case-by-case basis when compilation is too slow.

**_Note 2_**: Although the equivalent non-unrolled function call uses `init` as a keyword argument, and although `unrolled_reduce` supports using `init` as a keyword argument, this is not done here in order to ensure GPU compatibility. If the non-orographic gravity wave parameterization calls `unrolled_reduce(op, StaticOneTo(nc); init)`, the GPU compiler [claims](https://buildkite.com/clima/climaatmos-ci/builds/21144#019297b1-a2fc-400e-a860-d95edba1c2c1) that it cannot compile the resulting kernel because it contains a dynamic function call.

@Xiaoyang-Xie, the compilation mystery is finally resolved!


dennisYatunin:
Looks like CliMA/UnrolledUtilities.jl#13 didn't help with the compilation hanging issue. I will now try out CliMA/UnrolledUtilities.jl#15 instead.

dennisYatunin:
CliMA/UnrolledUtilities.jl#15 also didn't fix the issue. Now trying out CliMA/UnrolledUtilities.jl#16. We might need to combine these two PRs before we have a solution, but maybe this will work on its own.

PR #3317, opened 2024-09-19
Always use ClimaCommon modules

charleskawczynski:
Perhaps this is why we've been getting inconsistent timings on buildkite (and why some of our target gpu runs have been broken).

### DIFF
--- a/.buildkite/gpu_pipeline/pipeline.yml
+++ b/.buildkite/gpu_pipeline/pipeline.yml
@@ -1,7 +1,7 @@
 agents:
   queue: clima
   slurm_mem: 8G
-  modules: julia/1.10.0 cuda/julia-pref openmpi/4.1.5-mpitrampoline nsight-systems/2024.4.1
+  modules: climacommon/2024_05_27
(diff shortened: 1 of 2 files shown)


Sbozzolo:
`climacommon` is mainly a module on the Caltech cluster. On clima, it is just an alias for the `common` module, provided for convneinence and muscle-memory. The `common` module does not load nsight, so this would fail. When you load nsight, this is be perfectly equivalent to explicitely loading the modules.

The reason I didn't change them to `climacommon` is that it would require ensuring that every new release of climacommon also exists on `clima` (at least in name, even if they are aliases). Given that `clima` does modules a little differently, and is a more stable platform and doesn't see/need much updates, I didn't think it was necessary to add extra infrastructure/points of failure there. 

I would like to move `clima` to more consistent modules with the cluster, but that would require some work.

charleskawczynski:
Ok, makes sense. Synchronizing at the moment isn't too bad, but I hope will can somehow make synchronizing even smoother.

PR #3326, opened 2024-09-23
Do not add ᶜ∇²tke⁰ to cache if not needed

Sbozzolo:
For the diagnostic EDMF case, `ᶜ∇²tke⁰` is only used when `prognostic_tke` is true. This commit ensures that the variable is added to the cache only in that case. Prior to this commit, `ᶜ∇²tke⁰` was added unconditionally and would never be updated, so that it was always full with random numbers (being initialized with similar).

### DIFF
--- a/src/prognostic_equations/hyperdiffusion.jl
+++ b/src/prognostic_equations/hyperdiffusion.jl
@@ -34,14 +34,15 @@ function hyperdiffusion_cache(Y, hyperdiff::ClimaHyperdiffusion, turbconv_model)
     sgs_quantities =
         turbconv_model isa PrognosticEDMFX ?
         (;
-            ᶜ∇²tke⁰ = similar(Y.c, FT),
             ᶜ∇²uₕʲs = similar(Y.c, NTuple{n, C12{FT}}),
             ᶜ∇²uᵥʲs = similar(Y.c, NTuple{n, C3{FT}}),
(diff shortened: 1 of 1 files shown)


src/prognostic_equations/hyperdiffusion.jl
```diff
@@ -40,7 +40,7 @@ function hyperdiffusion_cache(Y, hyperdiff::ClimaHyperdiffusion, turbconv_model)
             ᶜ∇²mseʲs = similar(Y.c, NTuple{n, FT}),
             ᶜ∇²q_totʲs = similar(Y.c, NTuple{n, FT}),
         ) :
-        turbconv_model isa DiagnosticEDMFX ? (; ᶜ∇²tke⁰ = similar(Y.c, FT)) :
+        use_prognostic_tke(turbconv_model) ? (; ᶜ∇²tke⁰ = similar(Y.c, FT)) :
```
szy21, line 43: It should be the same for PrognosticEDMFX, where we only add ᶜ∇²tke⁰ to the cache if `use_prognostic_tke` (all other variables e.g. ᶜ∇²uₕʲs should always be in the cache though). Could you change that?

Sbozzolo: Changed

szy21 (approved):

PR #3336, opened 2024-09-24
Updates to gcm-driven single column calibrations

costachris:
Improve gcm-driven single column calibrations:

1.  modify priors
2. add entrainment pi group calibration
3.  improve plotting scripts
4.  add rmse metrics and plotting
5.  add eki metrics and plotting
6.  parallelize CA single column runs (cases) over cpu cores
7.  easier specification of normalization for LES/EDMF profiles (mean, std to transform vars -> N(0, 1)); option for taking log

A summary of the 

The observation vector for a single configuration is formed by concatenating profiles across calibration variables, where each geophysical variable is normalized to have approximately unit variance and zero mean. These variable-by-variable normalization factors are precomputed (`norm_factors_dict`) and applied to all observations. Following this operation, the spatiotemporal calibration window is applied and temporal means are computed to form the observation vector `y`. Because variables are normalized to have 0 mean and unit variance, a constant diagonal noise matrix is used (configurable as `const_noise`).

### Observation Map
1. **Time-mean**: Time-mean of profiles taken between [`y_t_start_sec`, `y_t_end_sec`] for `y` and [`g_t_start_sec`, `g_t_end_sec`] for `G`.
2. **Interpolation**: Case-specific (i.e., "shallow", "deep") interpolation to the calibration grid, defined with stretch-grid parameters in `z_cal_grid`.
3. **Normalization**: Variable-specific normalization using the mean and standard deviation defined in `norm_factors_by_var`. Optionally, take log of variables using `log_vars` before normalization.
4. **Concatenation**: Stack across cases in a batch, forming `y`, `G`.

#### cfSite 23
ensemble_plot_clw_i_11_cu

#### cfSite 17
ensemble_plot_clw_i_11

#### RMSE Plot
rmse_vs_iteration.pdf

### DIFF
--- a/calibration/experiments/gcm_driven_scm/prior_prognostic_pi_entr.toml
+++ b/calibration/experiments/gcm_driven_scm/prior_prognostic_pi_entr.toml
@@ -1,22 +1,15 @@
-[turb_entr_param_vec]
-prior = "VectorOfParameterized([Normal(-3.149, 0.554), Normal(8.228, 0.362)])"
-constraint = "repeat([bounded_below(0.0)], 2)"
+[entr_param_vec]
+prior = "VectorOfParameterized([Normal(0.0, 1.0), Normal(0.0, 1.0), Normal(0.0, 1.0), Normal(0.0, 1.0), Normal(0.0, 1.0), Normal(0.25, 0.15), Normal(0.0, 1.0), Normal(0.0, 1.0), Normal(0.0, 1.0), Normal(0.0, 1.0), Normal(0.0, 1.0), Normal(0.6, 0.3)])"
+constraint = "repeat([no_constraint()], 12)"
 type = "float"
 
 [entr_inv_tau]
-prior = "constrained_gaussian(entr_inv_tau, 0.0001, 0.0001, 0, Inf)"
+prior = "constrained_gaussian(entr_inv_tau, 1e-4, 6e-5, 0, Inf)"
 type = "float"
 
-[entr_coeff]
-prior = "constrained_gaussian(entr_coeff, 0.3, 0.2, 0, Inf)"
-type = "float"
-
-[detr_buoy_coeff]
-prior = "constrained_gaussian(detr_buoy_coeff, 0.12, 0.1, 0, Inf)"
-type = "float"
-
-[detr_vertdiv_coeff]
-prior = "constrained_gaussian(detr_vertdiv_coeff, 0.6, 0.4, 0, Inf)"
+[turb_entr_param_vec]
+prior = "VectorOfParameterized([Normal(-3.149, 0.554), Normal(8.228, 0.362)])"
+constraint = "repeat([bounded_below(0.0)], 2)"
 type = "float"
 
 [mixing_length_eddy_viscosity_coefficient]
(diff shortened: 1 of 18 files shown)


calibration/experiments/gcm_driven_scm/prior_prognostic_pi_entr.toml
```diff
@@ -0,0 +1,49 @@
+[entr_param_vec]
+prior = "VectorOfParameterized([Normal(0.0, 1.0), Normal(0.0, 1.0), Normal(0.0, 1.0), Normal(0.0, 1.0), Normal(0.0, 1.0), Normal(0.25, 0.15), Normal(0.0, 1.0), Normal(0.0, 1.0), Normal(0.0, 1.0), Normal(0.0, 1.0), Normal(0.0, 1.0), Normal(0.6, 0.3)])"
```
odunbar, line 2: Do you think this input format is sustainable? is there something you would have preferred to put in as input in an ideal world. We could add something into the EKP toml parsing if this is the case

calibration/experiments/gcm_driven_scm/run_calibration.jl
```diff
@@ -17,39 +15,44 @@ experiment_dir = dirname(Base.active_project())
...
-const model_config = experiment_config["model_config"]
-const z_max = experiment_config["z_max"]
-const eki_timestep = experiment_config["eki_timestep"]
-const const_noise = experiment_config["const_noise"]
-const prior =
-    CAL.get_prior(joinpath(experiment_dir, experiment_config["prior"]))
+
+# unpack experiment_config vars into scope 
+for (key, value) in experiment_config
+    @eval const $(Symbol(key)) = $value
```
odunbar, line 21: Usually this is not a great idea to do eval of something read from file, is this how it is typically done at the moment?

calibration/experiments/gcm_driven_scm/run_calibration.jl
```diff
@@ -79,23 +85,33 @@ end
 
 series_names = [ref_paths[i] for i in 1:length(ref_paths)]
 
-# minibatcher
+### define minibatcher
 rfs_minibatcher =
-    EKP.RandomFixedSizeMinibatcher(experiment_config["batch_size"])
+    EKP.FixedMinibatcher(collect(1:experiment_config["batch_size"]))
```
odunbar, line 90: So this is just creating 1 batch of `[1:batch_size]`. Is that intentional? Or did you want like 
`[[1:bs], [bs+1:2*bs], ... ]`

calibration/experiments/gcm_driven_scm/run_calibration.jl
```diff
@@ -79,23 +85,33 @@ end
...
-# ExperimentConfig is created from a YAML file within the experiment_dir
+###  EKI hyperparameters/settings
 @info "Initializing calibration" n_iterations ensemble_size output_dir
 CAL.initialize(
     ensemble_size,
     observations,
     prior,
     output_dir;
-    scheduler = EKP.DefaultScheduler(eki_timestep),
+    scheduler = EKP.DataMisfitController(on_terminate = "continue"),
```
odunbar, line 100: The PR message said "at default configuration" for the results displayed - but did it mean this?

odunbar (approved):
I've taken a look, and LGTM. 
- Could you add into the PR message about the additional normalization / data processing added for LES
- Could you check the minibatching? The setting of `FixedSizeMinibatcher[1:k]` will only (i think) create one minibatch. i.e. this is the same as calibrating over only cases `1:k` - was this the intent? or did you want something like `collect(1:k),collect(k+1:2k),...` If so this needs to be changed.

Otherwise my comments are really just small questions.

calibration/experiments/gcm_driven_scm/run_calibration.jl
costachris: It's just done in this file since a lot of the configuration variables are used. I had them listed out before like 
```
const n_iterations = experiment_config["n_iterations"]
const ensemble_size = experiment_config["ensemble_size"]
...
```
I can add that back if it's safer.

calibration/experiments/gcm_driven_scm/run_calibration.jl
costachris: I mean default for this setup (ie. running the calibration with `experiment_config.yml`).

calibration/experiments/gcm_driven_scm/prior_prognostic_pi_entr.toml
costachris: I think it can get unwieldy with i.e. hundreds of NN parameters. As is, I believe it allows for "repeat" logic to define the prior vector, which I could use to make it more concise. 

Otherwise I have a use case where I need to load pretrained NN weights as prior means, so I will probably need to load that from a file and set a constant spread. This may be a niche use case though.

PR #3337, opened 2024-09-24
Don't output diagnostics when simulation is less than 1h

Sbozzolo:
There's no point in adding hourly diagnostics when the simulation is less than 1h


juliasloan25 (approved):
Looks good, just one thought - this function taking in only `t_end` implies that `t_start = 0`. Maybe it should take in the simulation length instead

Sbozzolo:
> Looks good, just one thought - this function taking in only `t_end` implies that `t_start = 0`. Maybe it should take in the simulation length instead

Changed to use duration instead.

PR #3343, opened 2024-09-26
Fix restart with ozone

szy21:
WIP

## Content

----


Sbozzolo:
image

Sbozzolo (approved):

docs/src/aerosols.md
szy21, line 1: We can call this file ozone (if you only want ozone in this file) or tracers (if you want to have aerosols later). Ozone is not an aerosol:)

src/cache/tracer_cache.jl
```diff
@@ -1,40 +1,27 @@
...
+        axes(o3);
+        reference_date = start_date,
+        regridder_type = :InterpolationsRegridder,
+        method = LinearPeriodFillingInterpolation(Year(1)),
+    )
+    return (; o3, prescribed_o3_timevaryinginput)
+end
+
+function tracer_cache(Y, atmos, prescribed_aerosol_names, start_date)
+    if isempty(prescribed_aerosol_names)
```
szy21, line 20: Here it will return an empty namedtuple if prescribe_ozone is true but there is no prescribed_aerosol?

src/parameterized_tendencies/radiation/radiation.jl
```diff
@@ -10,10 +10,13 @@ import .Parameters as CAP
 import RRTMGP
 import .RRTMGPInterface as RRTMGPI
 
+import Dates: Year
+import ClimaUtilities.TimeVaryingInputs:
+    TimeVaryingInput, LinearPeriodFillingInterpolation
+
```
szy21, line 16: Do we need these?

src/solver/type_getters.jl
```diff
@@ -20,11 +20,17 @@ function get_atmos(config::AtmosConfig, params)
...
     precip_model = get_precipitation_model(parsed_args)
     cloud_model = get_cloud_model(parsed_args)
+    ozone = get_ozone(parsed_args)
     radiation_mode = get_radiation_mode(parsed_args, FT)
     forcing_type = get_forcing_type(parsed_args)
     call_cloud_diagnostics_per_stage =
         get_call_cloud_diagnostics_per_stage(parsed_args)
 
+    if isnothing(ozone) && radiation_mode isa RRTMGPI.AbstractRRTMGPMode
+        @warn "prescribe_ozone is set to nothing with an RRTMGP model. Resetting to IdealizedOzone. This behavior will stop being supported in some future release"
```
szy21, line 30: prescribe_ozone is default to false in default configs. Does this mean we want the user to specify true or false in the config? Why is it not a good idea to default it to IdealizedOzone if prescribe_ozone is not set (is this what you mean by prescribe_ozone is set to nothing)?

Sbozzolo: Discussed offline: in a nutshell, we shouldn't be adding stuff by default (this is also our pattern for most other types)

docs/src/aerosols.md
Sbozzolo: Renamed to tracers

PR #3344, opened 2024-09-26
Update dependencies

charleskawczynski:


Sbozzolo (approved):

charleskawczynski:
The JET failure is due to the fact that we're now unrolling the callback loop in ClimaTimeSteppers. On main, JET sees that loop and reports a single inference failure, and does not analyze inference inside each of those callbacks. With this PR, JET will now report inference on the contents of the callbacks. So, this will add a bit more friction to the adding type-unstable code in our callbacks.

PR #3346, opened 2024-09-30
Use ClimaReproducibilityTests.jl

charleskawczynski:
This PR replaces NCRegressionTests with ClimaReproducibilityTests.jl, which doesn't depend on NCDatasets. This interface is more transparent, has a path to lighter dependencies, and better matches the name of what we're doing.


test/Project.toml
```diff
@@ -9,6 +9,7 @@ ClimaAtmos = "b2c96348-7fb7-4fe0-8da9-78d88439e717"
 ClimaComms = "3a4d1b5c-c61d-41fd-a00a-5873ba7a1b0d"
 ClimaCore = "d414da3d-4745-48bb-8d80-42e94e092884"
 ClimaParams = "5c42b081-d73a-476f-9059-fd94b934656c"
+ClimaReproducibilityTests = "e0c89595-00ba-42a9-9f9b-061ef3dc23a1"
```
Sbozzolo, line 12: Is this used in the test environment?

charleskawczynski: Good question, I'll take a look.

charleskawczynski: I don't think so, I'll try removing it.

nefrathenrici (approved):
Thanks Charlie!

PR #3350, opened 2024-10-01
add a lower limit for pressure in radiation

szy21:
## Purpose 

When add_isothermal_boundary_layer is true, we add a layer in rrtmgp and set the pressure to half of the sum of top level pressure and rrtmgp minimum pressure [here](https://github.com/CliMA/ClimaAtmos.jl/blob/1c944547ce44e3a1d31a33eed820adc5bc26266c/src/parameterized_tendencies/radiation/RRTMGPInterface.jl#L1111). This PR limits the pressure to p_min multiplied by a small number to prevent pressure from decreasing or being constant with height at the top level.

This unfortunately not always works (e.g. it seems to be affected by sponge parameters). But I think overall it's still an improvement and we should add it.

## Content

----


Sbozzolo (approved):
Please, add more information on what this is accomplishing before merging.

szy21:
I updated the code comment and the PR message.

PR #3351, opened 2024-10-01
Reorganize restart tests

Sbozzolo:
This PR enables MPI tests for restarts and changes a little bit the structure of the restart test. I changed the order of things because I need to add more complex cases, but I don't want to add more nesting levels. Now, I can push more tests to the test dict directly.


charleskawczynski (approved):
Looks good!

Sbozzolo:
GPU + MPI also timed out a couple of times. I removed it and will work on it next.

PR #3356, opened 2024-10-03
Fix 3355

### LINKED ISSUE #3355: insolation_tuple is not correctly set for boxes/columns
https://github.com/CliMA/ClimaAtmos.jl/blob/aedb9984dd95a53a199f2c223d8b3f664f7d33bc/src/callbacks/callbacks.jl#L325

this update is not in-place, so `p.radiation.insolation_tuple` does not get update.

charleskawczynski:
Closes #3355.

### DIFF
--- a/src/callbacks/callbacks.jl
+++ b/src/callbacks/callbacks.jl
@@ -322,11 +322,13 @@ function set_insolation_variables!(Y, p, t, ::TimeVaryingInsolation)
         @. weighted_irradiance = irradiance * (au / last(insolation_tuple))^2
     else
         # assume that the latitude and longitude are both 0 for flat space
-        insolation_tuple = instantaneous_zenith_angle(d, δ, η_UTC, FT(0), FT(0))
-        rrtmgp_model.cos_zenith .=
-            cos(min(first(insolation_tuple), max_zenith_angle))
-        rrtmgp_model.weighted_irradiance .=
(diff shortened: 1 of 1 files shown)


szy21 (approved):

Sbozzolo:
I tried that, but it doesn't work. I'll paste the error tomorrow

charleskawczynski:
> I tried that, but it doesn't work. I'll paste the error tomorrow

It looks like it's the next line that's breaking, interesting.

charleskawczynski:
Ah, that's the only block that it's used, so this shouldn't actually change the behavior anywhere. Also, the existing form is technically a bit more efficient since `insolation_tuple` is not spatially varying in that case. We could micro-optimize the cache for this, but I think it's still better to be safe for now and still update `insolation_tuple` in the cache, in case it is used in the future (or in development branches).

Sbozzolo:
> Ah, that's the only block that it's used, so this shouldn't actually change the behavior anywhere. Also, the existing form is technically a bit more efficient since `insolation_tuple` is not spatially varying in that case. We could micro-optimize the cache for this, but I think it's still better to be safe for now and still update `insolation_tuple` in the cache, in case it is used in the future (or in development branches).

I agree, even if it is used only here, we should be consistent and keep it updated. It is counterintuitive that in some cases `p.radiation.insolation_tuple` is updated and valid, but in other cases is uninitialized.

PR #3359, opened 2024-10-03
Update insolation_tuple for boxes too

Sbozzolo:
While `insolation_tuple` seems to be currently used only in one function, it is stored in the cache and available everywhere. For spheres, `insolation_tuple` was updated in the radiation callback, but that was not the case for other configurations. This commit ensures that `insolation_tuple` is always updated. For non-spheres, we assume insolation at Greenwhich (lat = long = 0).

This commit also unifies a little bit the logic between the two cases (with and without sphere).

@szy21 could you double-check that my code rearrangement is valid? I think that the two branches of the if statement share most of the logic, but they are written slightly differently and I don't think the field2array should matter in this case

Closes #3356 #3355

### DIFF
--- a/src/callbacks/callbacks.jl
+++ b/src/callbacks/callbacks.jl
@@ -304,13 +304,13 @@ function set_insolation_variables!(Y, p, t, ::TimeVaryingInsolation)
             )
         )
     bottom_coords = Fields.coordinate_field(Spaces.level(Y.c, 1))
+    cos_zenith =
+        Fields.array2field(rrtmgp_model.cos_zenith, axes(bottom_coords))
+    weighted_irradiance = Fields.array2field(
+        rrtmgp_model.weighted_irradiance,
+        axes(bottom_coords),
+    )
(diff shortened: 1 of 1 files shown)


charleskawczynski:
Both this PR and #3356 are basically running into https://github.com/CliMA/ClimaCore.jl/issues/1718

Sbozzolo:
> Both this PR and #3356 are basically running into [CliMA/ClimaCore.jl#1718](https://github.com/CliMA/ClimaCore.jl/issues/1718)

This PR is passing

charleskawczynski:
> This PR is passing

Ah, I must have been looking at the wrong build, nice!

charleskawczynski:
Ah, on main, `rrtmgp_model.cos_zenith .= cos(min(first(insolation_tuple), max_zenith_angle))` is an Array-based broadcast expression. That was a bit confusing. Makes sense to me now.

charleskawczynski (approved):

szy21:
Yes, the code rearrangement makes sense to me, thanks!

PR #3364, opened 2024-10-04
Add deep convection to gcm-driven SCM setup

costachris:
Add relaxation timescale for deep convection in gcm-driven SCM setup


szy21:
The parameters change looks good to me. Gabriele can take a look at the code change.

Sbozzolo (commented):
Can you turn `forcing_type` into a julia type and use dispatch instead of matching strings?

Sbozzolo (commented):
Code looks good (altough, I think that it is more efficient to dispatch over the first argument). Could you please add some documentation, docstrings, and expand upon what the config `external_forcing_type` is?

Your help message for `external_forcing_type` is `"External forcing type", which doesn't teach much

szy21:
Actually, I think we should add these time scales to parameters instead of having a type. No need to change it now, but whoever uses it or changes it again for other types (e.g. reanalysis) should use parameters instead I think.

src/prognostic_equations/forcing/external_forcing.jl
```diff
@@ -136,8 +191,10 @@ function external_forcing_cache(Y, external_forcing::GCMForcing, params)
             set_insolation!(insolation)
             set_cos_zenith!(cos_zenith)
 
-            @. ᶜinv_τ_wind[colidx] = 1 / (6 * 3600)
-            @. ᶜinv_τ_scalar[colidx] = compute_gcm_driven_scalar_inv_τ(zc_gcm)
+            @. ᶜinv_τ_wind[colidx] =
+                compute_gcm_driven_momentum_inv_τ(zc_gcm; external_forcing_type)
+            @. ᶜinv_τ_scalar[colidx] =
```
costachris, line 196: @Sbozzolo I realized that using @. with a dispatched function seems to result in an error, since it's trying to broadcast over the type as well (which does not have a length, so I get `no method matching length`). Am I missing something? I also tried making it a kwarg instead to no avail

Sbozzolo: Add this line:
```
Base.broadcastable(x::AbstractGCMDrivenForcingType) = tuple(x)
```

PR #3365, opened 2024-10-07
Add copy to compute! functions for radiation

Sbozzolo:
The radiation diagnostics were returning references to objects, leading to conflicts when multiple diagnostics were added at the same time.

This commit ensures that all the radiation diagnostics return a new object.

https://github.com/CliMA/ClimaDiagnostics.jl/pull/88 fixes the problem direclty in ClimaDiagnostics, but I think it is still worth fixing in atmos, too, so that we can be compatible with previous versions of ClimaDiagnostics and more uniform with what we are doing with all the other diagnostics

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,12 @@ ClimaAtmos.jl Release Notes
 Main
 -------
 
+### ![][badge-🐛bugfix] Bug fixes
+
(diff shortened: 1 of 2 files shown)


szy21 (approved):
Thanks!

PR #3367, opened 2024-10-08
Manually fuse broadcast expressions in diag edmf

charleskawczynski:
This PR fuses a few more broadcast expressions in diagnostic edmf.


szy21 (approved):

charleskawczynski:
> The jobs that are failing are with precip_model: nothing. So maybe there is some issue with the empty field?

Yeah, I was afraid that might not work. I think there should be an easy enough work-around.

PR #3369, opened 2024-10-08
Update climacommon to 2024_12_16

Sbozzolo:
🤖 Beep boop. I am GabrieleBOT. 🤖

I received an update so that I can inform you directly of the changes (but feel free to check the [release notes](https://github.com/CliMA/ClimaModules/blob/main/NEWS.md)).

The most recent version of climacommon uses the newly released Julia 1.11.2.


charleskawczynski:
Among the GHA tests, it looks like the windows job gets the farthest, reaching (but not completing) `test_init_with_file.jl`. It could be that that is an artifact download issue (maybe even on windows?). Hm, but macOS got that far on the one before that 🤔. Ah, we have `timeout-minutes: 30` for CI, I'm fine bumping that to 1 hr. It's probably just taking longer due to slower compilation time.

perf/Manifest-v1.11.toml
```diff
@@ -0,0 +1,2883 @@
...
+
+    [deps.ClimaAnalysis.extensions]
+    GeoMakieExt = "GeoMakie"
+    MakieExt = "Makie"
+
+    [deps.ClimaAnalysis.weakdeps]
+    GeoMakie = "db073c08-6b98-4ee5-b6a4-5efafb3259c6"
+    Makie = "ee78f7c6-11fb-53f2-987a-cfe4a2b5a57a"
+
+[[deps.ClimaAtmos]]
```
charleskawczynski, line 352: This should be developed, and not fixed to a specific version, that's likely what's wrong with the perf job(s)

charleskawczynski:
On buildkite, it looks like some jobs are getting [precompiled twice](https://buildkite.com/clima/climaatmos-ci/builds/22180#01944688-0cec-416e-85ed-0dca6442ab58/182-188), ah this is likely due to using a fixed old version of atmos. Developing atmos in the perf env should fix this, I think.

charleskawczynski:
We can increase the ref counter to fix the reproducibility tests, could someone review the physics changes?

charleskawczynski:
Finally, the MPI job looks like it may be running out of memory, could we increase the mem request for `test restart MPI`?

charleskawczynski (commented):
Overall changes look fine, but it looks like we need to address some things that need fixed.

Sbozzolo:
I am finding that ClimaCore 0.14.12 is fine with Julia 1.10, but we need 0.14.19 for Julia 1.11 (https://github.com/CliMA/ClimaAtmos.jl/actions/runs/12694181365/job/35383420600). 

Unfortunately, I don't see any easy way to specify this constraint, so maybe the safest is just bumping the minimum to 0.14.19

charleskawczynski:
> I am finding that ClimaCore 0.14.12 is fine with Julia 1.10, but we need 0.14.19 for Julia 1.11 (https://github.com/CliMA/ClimaAtmos.jl/actions/runs/12694181365/job/35383420600).
> 
> Unfortunately, I don't see any easy way to specify this constraint, so maybe the safest is just bumping the minimum to 0.14.19

That makes sense. We kind of got bit by https://github.com/julialang/julia/issues/56295. The main branch is fine, but the high-level dispatch that effectively fixed this issue must have been merged in 0.14.19.

I don't know what we could have done to prevent this, unfortunately.

PR #3371, opened 2024-10-09
Fuse some broadcast expressions

charleskawczynski:
This PR uses `@fused_direct` to fuse some broadcast expressions. This is basically a rebase of #2983.


charleskawczynski:
MultiBroadcastFusion.jl is a nice idea, but I think it doesn't help us fix #3860 and the graph expansion problem, which is really increasing our bandwidth. I think the right path forward is one of two:

 - Use `LazyBroadcast`, `columnwise!`, and (not yet sketched out) `slabwise!` to compute everything (or nearly everything) on the fly with shared memory-backed states.
 - Ditch ClimaCore's operators / broadcasting and just go full kernel-based programming, like Oceananigans.jl

PR #3372, opened 2024-10-09
Update to pipelines

### LINKED ISSUE #2911: Fix broken NVTX reports

charleskawczynski:
This PR updates us to use climacommon `2024_10_08`, which should (hopefully) fix the nsight issues. Looks like this closes #2911.

I'm going to update a few things in the pipelines.

### DIFF
--- a/.buildkite/comparison/pipeline.sh
+++ b/.buildkite/comparison/pipeline.sh
@@ -13,7 +13,7 @@ profiling=disable
 # set up environment and agents
 cat << EOM
 env:
-  JULIA_VERSION: "1.8.5"
+  JULIA_VERSION: "1.10.5"
   MPICH_VERSION: "4.0.0"
   OPENMPI_VERSION: "4.1.1"
   CUDA_VERSION: "11.3"
@@ -82,7 +82,7 @@ cat << EOM
     - label: ":computer: MPI Held-Suarez $res resolution test(ρe_tot) - ($nprocs) process"
       key: $job_id
       command:
-        - module load cuda/11.3 nsight-systems/2023.3.1
+        - module load cuda/11.3 nsight-systems/2024.6.1
         - mpiexec $command
       artifact_paths:
         - "$job_id/scaling_data_${nprocs}_processes.jld2"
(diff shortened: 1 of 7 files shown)


szy21 (approved):
Do we need to worry about the increase in allowed failures?

charleskawczynski:
> Do we need to worry about the increase in allowed failures?

It's just one failure, so I don't think so. Actually, it's probably from going from Julia 1.10.0 to 1.10.5. I think it should be fine.

.buildkite/comparison/pipeline.sh
```diff
@@ -13,7 +13,7 @@ profiling=disable
 # set up environment and agents
 cat << EOM
 env:
```
Sbozzolo, line 15: This looks broken and will not run. For example, the module cuda/11.3 doesn't exist. We should remove this pipeline

Sbozzolo (approved):

sriharshakandala:
https://buildkite.com/clima/climaatmos-target-gpu-simulations/builds/349

.buildkite/comparison/pipeline.sh
charleskawczynski: Ok, we can fix the rest of what is broken later, for now, I'm just trying to keep versions in sync.

sriharshakandala:
I noticed we have recently added module `climacommon/2024_10_08`
This seems to be using `julia v1.10.5`. Is there a reason we are not using the latest release `v1.11.0` , which was released on Oct 7th?

sriharshakandala:
It looks like the diagnostic and progostic EDMF jobs are still failing with error code 139.
https://buildkite.com/clima/climaatmos-target-gpu-simulations/builds/349#_

charleskawczynski:
> It looks like the diagnostic and progostic EDMF jobs are still failing with error code 139. https://buildkite.com/clima/climaatmos-target-gpu-simulations/builds/349#_

Yeah, that's a different error though, the nsys reports are now being correctly generated. It looks like maybe they're just too big. I'll shorten the time of the simulations.

Sbozzolo:
I don't think the reports are being correctly generated: the jobs don't run to completion (the prognostic EDMF job doesn't even run one iteration)

charleskawczynski:
Yeah, I think that there's too much data collected before the simulation even starts, but the nsys reports are generated. I tried opening one but my computer ran out of memory and asked me to quite applications, nsight was taking up over 37 GB and had loaded 82% of the report.

PR #3376, opened 2024-10-10
Fix GPU target pipeline nsight reports

### LINKED ISSUE #3375: Some nsight jobs have error codes 139
See, for example, https://buildkite.com/clima/climaatmos-target-gpu-simulations/builds/349#019272fa-9aad-45fb-b6c0-375ad1481b51.

I think we can fix this by shortening the job time, or by narrowing the captured range.

charleskawczynski:
Closes #3375.


Sbozzolo:
There's three changes in this PR. Which one fixed the issue? Or all the three of them required?

charleskawczynski:
> There's three changes in this PR. Which one fixed the issue? Or all the three of them required?

Basically yes: the reports now work, but we're running out of memory. The memory requests fix those issues, but if you then try to download the files, they're huge. So I've also reduced the time, and added `--delay` to try and capture less of the trace (it should arguably be reduced further, but I'm at least now able to open files).

Sbozzolo (approved):
Thank you! 

I'd add some comments in the yml to explain why these flags are set (or a link to this issue)

charleskawczynski:
Not sure why some jobs broke on the build on main, but it previously passed: https://buildkite.com/clima/climaatmos-ci/builds/21046, so I think it's unrelated to this PR?

Sbozzolo:
Test restart MPI has been failing some times recently. I am not sure why. It's unrelated to this PR and we've seen it in other places too.

szy21:
But this GPU job also failed? https://buildkite.com/clima/climaatmos-ci/builds/21389#0192dde1-c0e6-4be4-a511-c8e1c1271ed5

PR #3378, opened 2024-10-10
Remove Adapt dependence, use ETOPO2022 dataset and regrid orography using SpaceVaryingInputs

akshaysridhar:


akshaysridhar:
32x_Biharmonic_Orography
256x_Biharmonic_Orography
Biharmonic_smoothing_0-256pass
Laplacian_0-256pass

Sbozzolo:
Please, also remove the `gaussian_smooth`, its tests, and the dependency on Distributions in `test`, if it that's no longer needed.

examples/topography_spectra.jl
Sbozzolo, line 1: Can you add some documentation to this example?

src/utils/AtmosArtifacts.jl
```diff
@@ -67,4 +67,15 @@ function aerosol_concentration_file_path(; context = nothing)
...
 
+"""
+    earth_orography_file_path(; context=nothing)
+
+Construct the file path for the 60arcsecond orography data NetCDF file.
+
+Downloads the 60arc-second dataset by default. 
+"""
+function earth_orography_file_path(; context = nothing)
+    return res_file_path("earth_orography_60arcseconds"; context)
```
Sbozzolo, line 78:
```suggestion
    full_name = "earth_orography_60arcseconds"
    return joinpath(@clima_artifact(full_name, context), "$(full_name).nc")
```

src/utils/common_spaces.jl
```diff
@@ -74,6 +74,38 @@ function make_horizontal_space(mesh, quad, comms_ctx, bubble)
     return space
 end
 
+function diffuse_surface_elevation_biharmonic!(
```
Sbozzolo, line 77: Can you add docstring to this function, including references to where the equations came from?

src/utils/common_spaces.jl
```diff
@@ -100,8 +132,39 @@ function make_hybrid_spaces(
...
+        @info "SpaceVaryingInputs: Remapping orography onto spectral space"
+        z_surface = SpaceVaryingInputs.SpaceVaryingInput(
+            AA.earth_orography_file_path(;
+                context = ClimaComms.context(h_space),
+            ),
+            "z",
+            h_space,
+        )
+        parent(z_surface) .=
+            ifelse.(parent(z_surface) .< FT(0), FT(0), parent(z_surface))
```
Sbozzolo, line 147: Maybe we should apply this to the input data directly? You can pass a similiar preprocess func to the `SpaceVaryingInput`

src/utils/common_spaces.jl
```diff
@@ -100,8 +132,39 @@ function make_hybrid_spaces(
...
+            hypsography = Hypsography.SLEVEAdaption(
+                Geometry.ZPoint.(z_surface),
+                FT(parsed_args["sleve_eta"]),
+                FT(parsed_args["sleve_s"]),
+            )
+        elseif parsed_args["mesh_warp_type"] == "Linear"
+            @info "Linear mesh warp"
+            hypsography =
+                Hypsography.LinearAdaption(Geometry.ZPoint.(z_surface))
+        end
```
Sbozzolo, line 167: What happens if `parsed_args["mesh_warp_type"] == "linear"`? Or some other cases that are not captured by this if-else?

src/solver/type_getters.jl
```diff
@@ -183,18 +164,27 @@ function get_spaces(parsed_args, params, comms_ctx)
...
             make_hybrid_spaces(
                 h_space,
                 z_max,
                 z_elem,
                 z_stretch;
-                parsed_args = parsed_args,
                 surface_warp = warp_function,
                 deep,
             )
+        elseif topography == "Earth"
```
szy21, line 178: We can first check topography here, then check warp_function, and can get rid of the `&& topography != "Earth"`? I think the logic is more clear this way.

src/utils/common_spaces.jl
```diff
@@ -100,8 +132,39 @@ function make_hybrid_spaces(
         z_mesh,
     )
     z_grid = Grids.FiniteDifferenceGrid(z_topology)
-    if isnothing(surface_warp)
+    if isnothing(surface_warp) && parsed_args["topography"] != "Earth"
         hypsography = Hypsography.Flat()
+    elseif isnothing(surface_warp) && parsed_args["topography"] == "Earth"
```
szy21, line 137: Same here, we can check topography first then surface_warp to make the logic more clear.

examples/topography_spectra.jl
szy21: If I understand correctly, this example is to create a space with different iterations of Laplacian smoothing and plot the spectra. Is there a way to plot the spectra from topography diagnostics from a simulation? I think the more general use case would be someone runs a global simulation and wants to see the spectra of that particular simulation, without generating the space again.

szy21 (commented):
Thanks! Here are some comments. And I agree that some documentation would be helpful.

examples/topography_spectra.jl
akshaysridhar: Yes; we've had such a plot in our CI pipelines before (for the DCMIP case) - `orog.nc` allows us to generate this plot as part of normal CI builds. I'll reintroduce this.

src/solver/type_getters.jl
akshaysridhar: Yes ; I'll go over this again and try to clean up some of these branching statements in `common_spaces` and `type_getters.`

src/utils/common_spaces.jl
akshaysridhar: Sure; I can add better docstrings; these functions are implementations of unsteady diffusion equations applied to the surface orography.

akshaysridhar: @Sbozzolo I'll move this function directly to ClimaCore (within the Hypsography module)

examples/Manifest.toml
```diff
@@ -1,8 +1,8 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.10.5"
+julia_version = "1.10.4"
```
szy21, line 3: Could we revert this change?

src/solver/type_getters.jl
```diff
@@ -183,17 +164,27 @@ function get_spaces(parsed_args, params, comms_ctx)
         else
             Meshes.Uniform()
         end
-        if warp_function == nothing
-            make_hybrid_spaces(h_space, z_max, z_elem, z_stretch; deep)
-        else
+        if topography == "Earth"
```
szy21, line 167:
```suggestion
        if warp_function == nothing
```

szy21: (It seems this and the following branches are returning the same thing. They both return `make_hybrid_spaces` with `surface_warp=nothing`.)

src/solver/type_getters.jl
```diff
@@ -183,17 +164,27 @@ function get_spaces(parsed_args, params, comms_ctx)
...
                 h_space,
                 z_max,
                 z_elem,
                 z_stretch;
+                surface_warp = nothing,
                 parsed_args = parsed_args,
+                deep,
+            )
+        elseif warp_function == nothing && topography != "Earth"
+            make_hybrid_spaces(h_space, z_max, z_elem, z_stretch; deep, parsed_args)
```
szy21, line 178:
```suggestion
```

src/utils/common_spaces.jl
```diff
@@ -100,7 +100,42 @@ function make_hybrid_spaces(
...
+            ),
+            "z",
+            h_space,
+        )
+        parent(z_surface) .=
+            ifelse.(parent(z_surface) .< FT(0), FT(0), parent(z_surface))
+        Δh_scale = Spaces.node_horizontal_length_scale(h_space)
+        Hypsography.diffuse_surface_elevation!(
+            z_surface;
+            κ = FT((Δh_scale)^2 / 100),
```
szy21, line 118: Could you add a comment on where this coefficient comes from?

src/solver/type_getters.jl
```diff
@@ -146,26 +146,7 @@ function get_spaces(parsed_args, params, comms_ctx)
...
-                linear_interpolation(
-                    (lon, lat),
-                    esmth,
-                    extrapolation_bc = (Periodic(), Flat()),
-                ),
-            )
-        end
-        @info "Generated interpolation stencil"
-        warp_function = generate_topography_warp(earth_spline)
+        warp_function = nothing
```
szy21, line 149: Is there a reason for setting warp_function to nothing for Earth topography, and using parsed_args["topography"] to set z_surface in `make_hybrid_spaces`? I find it a bit confusing that warp_function is nothing for Earth topography.

akshaysridhar: The previous iteration was an attempt to keep this a non-breaking change; I've updated the `make_hybrid_space` function to use `SpaceVaryingInputs` everywhere which simplifies some of these branches.

Sbozzolo:
@akshaysridhar could also remove the ETOPO1 artifact as part of this pull request? 

If it is still needed for tests, could you ensure that it is only downloaded in tests?

I think this is the only artifact left that ClimaCoupler has manually download, everything else is handled automatically.

juliasloan25:
Is this PR ready to merge?

akshaysridhar:
@szy21 There is a gravity wave function that uses the old artifact e.g. `gfdl_restart` vs `raw_topo` in `orography_gravity_wave` (without smoothing) - is it reasonable to update this here as well ? (I can separate that if needed)

szy21:
> @szy21 There is a gravity wave function that uses the old artifact e.g. `gfdl_restart` vs `raw_topo` in `orography_gravity_wave` (without smoothing) - is it reasonable to update this here as well ? (I can separate that if needed)

Feel free to update it if the test doesn't break and the result looks fine.

src/utils/common_spaces.jl
akshaysridhar: Updated, E3SM documentation guides this (otherwise it can be estimated from the grid resolutions and some nominal CFL)

akshaysridhar:
> @akshaysridhar could also remove the ETOPO1 artifact as part of this pull request?
> 
> If it is still needed for tests, could you ensure that it is only downloaded in tests?
> 
> I think this is the only artifact left that ClimaCoupler has manually download, everything else is handled automatically.

@Sbozzolo if the changes look reasonable here we can prepare to remove the Coupler download (as in the land-sea mask PR).

examples/topography_spectra.jl
```diff
@@ -0,0 +1,133 @@
...
+to apply `diffiter` instances of Laplacian diffusion passes 
+(which rely on the canonical unsteady diffusion equation).
+Returns a `Field` containing surface orography. 
+"""
+function generate_spaces(; h_elem = 60, diffiter = 32)
+    FT = Float32
+    param_dict = CP.create_toml_dict(FT)
+    params = CP.get_parameter_values(param_dict, ["planet_radius"])
+    cubed_sphere_mesh =
+        CA.cubed_sphere_mesh(; radius = params.planet_radius, h_elem)
```
Sbozzolo, line 32:
```suggestion
function generate_spaces(; h_elem = 60, diffiter = 32, planet_radius = 6.378e6)
    FT = Float32
    cubed_sphere_mesh =
        CA.cubed_sphere_mesh(; radius = FT(planet_radius), h_elem)
```
I think it is an overkill to have this function depending on `ClimaParams` when all we need is an Earth radius. This is also just an example.

examples/topography_spectra.jl
```diff
@@ -0,0 +1,133 @@
...
+    quad = Quadratures.GLL{4}()
+    comms_ctx = ClimaComms.SingletonCommsContext()
```
Sbozzolo, line 34: Why is this restriction?

examples/topography_spectra.jl
```diff
@@ -0,0 +1,133 @@
...
+    @assert h_space isa CC.Spaces.SpectralElementSpace2D
+    coords = CC.Fields.coordinate_field(h_space)
+    target_field = CC.Fields.zeros(h_space)
+    elev_from_file = SpaceVaryingInputs.SpaceVaryingInput(
+        AA.earth_orography_file_path(; context = comms_ctx),
+        "z",
+        h_space,
+    )
+    parent(elev_from_file) .=
+        ifelse.(parent(elev_from_file) .< FT(0), FT(0), parent(elev_from_file))
```
Sbozzolo, line 46: You can avoid using `parent` by using a mask like this:
```julia
function mask(x::FT) where {FT}
     return FT(x > 0)
end
```

examples/topography_spectra.jl
```diff
@@ -0,0 +1,133 @@
...
+    diffuse_surface_elevation!(
+        elev_from_file,
+        κ = FT((Δh_scale)^2 / 100),
+        dt = FT(1),
+        maxiter = diffiter,
+    )
+    parent(elev_from_file) .=
+        ifelse.(parent(elev_from_file) .< FT(0), FT(0), parent(elev_from_file))
```
Sbozzolo, line 54: Same as before

examples/topography_spectra.jl
```diff
@@ -0,0 +1,133 @@
...
+end
+
+"""
+    remap_to_array(var, hcoords)
+Given an Array of `hcoords`, this function remaps a ClimaCore
+Field `var` onto `hcoords`. 
+Returns an `Array` containing the values which make up `var`.
+"""
+function remap_to_array(var, hcoords)
+    remapper = Remapping.Remapper(CC.axes(var), hcoords)
```
Sbozzolo, line 65:
```suggestion
    remapper = Remapping.Remapper(axes(var), hcoords)
```

examples/topography_spectra.jl
```diff
@@ -0,0 +1,133 @@
...
+
+"""
+    generate_all_spectra(;h_elem=60)
+Uses the spectral calculator and space generation tools in this example file
+to generate a series of surface orography fields with different extents of 
+Laplacian smoothing. 
+Returns a `Plots.Figure`. 
+"""
+function generate_all_spectra(; h_elem = 60)
+    fig = Plots.plot()
```
Sbozzolo, line 108: Would it be possible to use Makie instead?

src/parameterized_tendencies/gravity_wave_drag/orographic_gravity_wave.jl
```diff
@@ -39,7 +39,7 @@ function orographic_gravity_wave_cache(Y, ogw::OrographicGravityWave)
     elseif ogw.topo_info == "raw_topo"
         # TODO: right now this option may easily crash
         # because we did not incorporate any smoothing when interpolate back to model grid
-        elevation_rll = joinpath(topo_elev_dataset_path(), "ETOPO1_coarse.nc")
+        elevation_rll = AA.earth_orography_file_path(; context=ClimaComms.context(Spaces.horizontal_space(axes(Y.c))))
```
Sbozzolo, line 42:
```suggestion
        elevation_rll = AA.earth_orography_file_path(; context=ClimaComms.context(Y.c)))
```
I think that this might be enough

src/solver/type_getters.jl
```diff
@@ -183,19 +149,14 @@ function get_spaces(parsed_args, params, comms_ctx)
...
-                deep,
-            )
-        end
+        make_hybrid_spaces(
+            h_space,
+            z_max,
+            z_elem,
+            z_stretch;
+            deep,
+            parsed_args = parsed_args,
```
Sbozzolo, line 158:
```suggestion
            parsed_args,
```

src/utils/common_spaces.jl
Sbozzolo: Same as above with regards to avoiding using `parent`

src/utils/common_spaces.jl
```diff
@@ -100,11 +98,66 @@ function make_hybrid_spaces(
...
+        # E3SM
+        # https://acme-climate.atlassian.net/wiki/spaces/DOC/pages/1456603764/V1+Topography+GLL+grids
+        Hypsography.diffuse_surface_elevation!(
+            z_surface;
+            κ = FT(28e7),
+            dt = FT(1),
+            maxiter = 128,
+        )
+        parent(z_surface) .=
+            ifelse.(parent(z_surface) .< FT(0), FT(0), parent(z_surface))
```
Sbozzolo, line 145: Here too

examples/topography_spectra.jl
akshaysridhar: Sure, this was intended as a quick demo of `params` usage in `CA` (I realise that the `ClimaParams` documentation also has this. I'll change this to just depend on the keyword arg.

examples/topography_spectra.jl
akshaysridhar: This example is light enough that it can be run on a single device, so MPI / generalized context was unnecessary. I'll generalize with `ClimaComms.context()`.

akshaysridhar:
@Sbozzolo Masks to avoid `parent`, `Makie`, and `context` have been fixed following your suggestions.

examples/topography_spectra.jl
```diff
@@ -0,0 +1,129 @@
...
+
+const AA = AtmosArtifacts
+
+
+function mask(x::FT) where {FT}
+    return x * FT(x > 0)
+end
+
+"""
+    generate_spaces(; h_elem=60, diffiter=32)
```
Sbozzolo, line 23:
```suggestion
    generate_spaces(; h_elem=60, diffiter=32, planet_radius = 6.378e6)
```

Sbozzolo (approved):

Sbozzolo:
Thank you!

akshaysridhar:
@Sbozzolo @szy21 `compute hmax` in the gravity wave parameterization is expensive with the new dataset (ci timeouts) so I added a simple downscaled array (Skip `nth` pt in data so array size is `3600x1800`).

examples/topography_spectra.jl
```diff
@@ -0,0 +1,129 @@
...
+For a given number of elements `h_elem`, this function generates a 
+horizontal spectral space (ClimaCore.SpectralElementSpace2D) and
+generates an interpolated (using SpaceVaryingInputs) orography field
+on this space. The `diffuse_surface_elevation` function is called 
+to apply `diffiter` instances of Laplacian diffusion passes 
+(which rely on the canonical unsteady diffusion equation).
```
tapios, line 29:
```suggestion
(which rely on the canonical diffusion equation).
```

examples/topography_spectra.jl
```diff
@@ -0,0 +1,129 @@
...
+    target_field = CC.Fields.zeros(h_space)
+    elev_from_file = SpaceVaryingInputs.SpaceVaryingInput(
+        AA.earth_orography_file_path(; context = comms_ctx),
+        "z",
+        h_space,
+    )
+    elev_from_file = @. mask(elev_from_file)
+    diffuse_surface_elevation!(
+        elev_from_file,
+        κ = FT(28e7 * (30 / h_elem)^2),
```
tapios, line 51:
```suggestion
        κ = diff_courant * Δh_scale^2 / dt,
```
Please avoid arbitrary numbers like 28e7 in code. I think what you want here is something like the above, where diff_courant is some parameter to ensure numerical stability (e.g., 0.1).

examples/topography_spectra.jl
```diff
@@ -0,0 +1,129 @@
...
+        dt = FT(1),
+        maxiter = diffiter,
```
tapios, line 53: As I mentioned on Slack, I think it's better to compute `maxiter` as `log(n) / diff_courant` where `n` is the fractional damping of the smallest resolved scale and would be a good input parameter (e.g., `n = 1000` or so).

examples/topography_spectra.jl
```diff
@@ -0,0 +1,129 @@
...
+"""
+function remap_to_array(test_var, hcoords)
+    remapper = Remapping.Remapper(axes(test_var), hcoords)
+    orog = Array(Remapping.interpolate(remapper, test_var))
+end
+
+"""
+    gen_spectra(test_var)
+Given a Field `test_var`, this function first calls `remap_to_array`
+to remap this spectral element field onto Gaussian long-lat
```
tapios, line 73:
```suggestion
to remap this spectral element field onto a latitude-longitude
```
(it's not Gaussian)

examples/topography_spectra.jl
```diff
@@ -0,0 +1,129 @@
...
+coordinates, and computes the spectra.
```
tapios, line 74:
```suggestion
grid, and computes the spectra.
```

examples/topography_spectra.jl
```diff
@@ -0,0 +1,129 @@
...
+Returns (`w_numbers`, `power_spectrum`) which contain the spherical
+wavenumbers and the spectral energy corresponding to this array
+of `w_numbers`. 
+"""
+function gen_spectra(test_var)
+    # Remap onto 1° resolution horizontal coordinates. 
+    longpts = range(-180.0, 180.0, 360)
+    latpts = range(-90.0, 90.0, 180)
+    hcoords =
```
tapios, line 83: The fixed resolution of this grid will create problems, e.g., for high-resolution topography. Please replace by a grid that adapts to the model resolution, e.g., with a grid spacing of 360*Δh_scale/(2*pi earth_radius).

src/utils/common_spaces.jl
```diff
@@ -100,11 +98,65 @@ function make_hybrid_spaces(
...
+        mask(x::FT) where {FT} = x * FT(x > 0)
+        z_surface = @. mask(z_surface)
+        # Coefficient for horizontal diffusion of surface orography
+        # is obtained from the V1, V2 Topography documentation from 
+        # E3SM
+        # https://acme-climate.atlassian.net/wiki/spaces/DOC/pages/1456603764/V1+Topography+GLL+grids
+        nelem = Meshes.n_elements_per_panel_direction(axes(z_surface))
+        Hypsography.diffuse_surface_elevation!(
+            z_surface;
+            κ = FT(28e7 * (30 / nelem)^2),
```
tapios, line 140: Replace analogous to the above formulation.

examples/topography_spectra.jl
akshaysridhar: Noted, this value is from the E3SM docs which I can make clearer - in general I'll replace this (via changes in `ClimaCore` to depend entirely on resolution and required attenuation factors) - (following the Slack discussion)

szy21:
Which one is the parameter / config that allows the user to choose damping factor? Or are you going to add it in the next PR?

akshaysridhar:
> Which one is the parameter / config that allows the user to choose damping factor? Or are you going to add it in the next PR?

In this instance, there is a parameter `n_attenuation` ; the entire smoothing method should, in my opinion, be in `ClimaCore` which requires an additional package release + update so I want to get the artifact changes with a simple implementation into `ClimaAtmos` through this PR first.

szy21:
> > Which one is the parameter / config that allows the user to choose damping factor? Or are you going to add it in the next PR?
> 
> In this instance, there is a parameter `n_attenuation` ; the entire smoothing method should, in my opinion, be in `ClimaCore` which requires an additional package release + update so I want to get the artifact changes with a simple implementation into `ClimaAtmos` through this PR first.

But `n_attenuation` is hard coded and not in config yet? I'm fine with getting this in, just to make sure.

akshaysridhar:
> > > Which one is the parameter / config that allows the user to choose damping factor? Or are you going to add it in the next PR?
> > 
> > 
> > In this instance, there is a parameter `n_attenuation` ; the entire smoothing method should, in my opinion, be in `ClimaCore` which requires an additional package release + update so I want to get the artifact changes with a simple implementation into `ClimaAtmos` through this PR first.
> 
> But `n_attenuation` is hard coded and not in config yet? I'm fine with getting this in, just to make sure.

Right, it is currently defined only within the `common_spaces` util

akshaysridhar:
> > > > Which one is the parameter / config that allows the user to choose damping factor? Or are you going to add it in the next PR?
> > > 
> > > 
> > > In this instance, there is a parameter `n_attenuation` ; the entire smoothing method should, in my opinion, be in `ClimaCore` which requires an additional package release + update so I want to get the artifact changes with a simple implementation into `ClimaAtmos` through this PR first.
> > 
> > 
> > But `n_attenuation` is hard coded and not in config yet? I'm fine with getting this in, just to make sure.
> 
> Right, it is currently defined only within the `common_spaces` util

@szy21: With the restart check updated in this commit, I've also added the damping factor to the config arg list.

test/restart.jl
```diff
@@ -69,6 +69,11 @@ function _error(
     arr2::AbstractArray;
     ABS_TOL = 100eps(eltype(arr1)),
 )
+    # There are some parameters, e.g. Obukhov length, for which Inf
+    # is a reasonable value (implying a stability parameter in the neutral boundary layer
+    # regime, for instance). We account for such instances with the `isfinite` function.
+    arr1 .*= isfinite.(arr1)
```
akshaysridhar, line 75: cc @Sbozzolo

Sbozzolo: Looks like a good solution to me!

PR #3379, opened 2024-10-10
Warn when dt_rad and dt_checkpoints are not compatible

Sbozzolo:
If we want our restarts to be reproducible, we have to worry about callbacks. Callbacks contain information about when they were last called, information that is not saved to checkpoint. This is particularly relevant for radiation. For example, if we save checkpoints every day, but call radiation every 15 h, we will have that restarting a simulation leads to a different evolution compared to not doing that. This is because radiation is called at different times:

Non-restarted simulation
t = 15h: radiation
t = 24h: checkpoint
t = 30h: radiation

Restarted simulation:
t = 15h: radiation
t = 24h: checkpoint
t = 39h: radiation

This commit adds a warning that a given simulation will not be reproducible when radiation and checkpoints frequencies are not commensurate.

To accomplish this, I added `periods_are_commensurate`, a functon that to check whether two `Dates.Periods` are compatible, meaning one is an integer multiple of the other. This is a little tricky to accomplish because some periods don't have a fixed length: for example, one Month can be 28, 29, 30, and 31 days.

In `periods_are_commensurate`, I handle all the cases that can be handled directly (e.g., periods with fixed length), and hardcoded some special, but relevant cases. The most important of these cases is when we have a Month.

Not all the cases are covered, but we can add them when/if they become relavant.

### DIFF
--- a/src/utils/utilities.jl
+++ b/src/utils/utilities.jl
@@ -403,3 +403,54 @@ function gaussian_smooth(arr::AbstractArray, sigma::Int = 1)
 
     return smoothed_arr
 end
+
+"""
+    isdivisible(dt_large::Dates.Period, dt_small::Dates.Period)
+
+Check if two periods are evenly divisible, i.e., if the larger period can be
+expressed as an integer multiple of the smaller period.
+
+In this, take into account the case when periods do not have fixed size, e.g.,
+one month is a variable number of days.
+
+# Examples
+```
+julia> isdivisible(Dates.Year(1), Dates.Month(1))
+true
+
+julia> isdivisible(Dates.Month(1), Dates.Day(1))
+true
+
+julia> isdivisible(Dates.Month(1), Dates.Week(1))
+false
+```
+
+## Notes
+
+Not all the combinations are fully implemented. If something is missing, please
+consider adding it.
+"""
+function isdivisible(dt_large::Dates.Period, dt_small::Dates.Period)
+    @warn "The combination $(typeof(dt_large)) and $(dt_small) was not covered. Please add a method to handle this case."
+    return false
+end
+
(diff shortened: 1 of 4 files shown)


szy21 (approved):
Thanks! It would be good to add a case for year too, as that will be commonly used in the future I think. We can use the same logic for month.

src/utils/utilities.jl
```diff
@@ -403,3 +400,62 @@ function gaussian_smooth(arr::AbstractArray, sigma::Int = 1)
 
     return smoothed_arr
 end
+
+"""
+    periods_are_commensurate(dt_large::Dates.Period, dt_small::Dates.Period)
```
charleskawczynski, line 405: Is commensurate the right word here? Aren't we basically looking to see if one is an integer multiple of the other?

Sbozzolo: You are right!

I started thinking about commensuare (two non-zero real numbers a and b are said to be commensurable if their ratio ⁠a/b⁠ is a rational number), but at the end I am not checking that, I am checking if they are integer multiple of the other

Sbozzolo:
> Looks good. Can/should we add similar logic to other periodic callbacks? This seems like a somewhat general constraint that we should try to satisfy

The other callbacks don't have to satisfy this propertly:
- log progress
- check NaN every N steps
- maybe gracefully exit
- call GC every N steps

Sbozzolo:
> Looks good. Can/should we add similar logic to other periodic callbacks? This seems like a somewhat general constraint that we should try to satisfy

But I will add this to the diagnostics next

charleskawczynski:
Theoretically, the other callbacks you listed (including diagnostics) should not impact reproducibility--assuming we don't sample random numbers in those sections. Radiation is one of the only callbacks that does impact the simulation by adding tendencies.

PR #3380, opened 2024-10-11
Try ClimaCore main

charleskawczynski:
This is testing to make sure that we don't have performance regressions in https://github.com/CliMA/ClimaCore.jl/pull/2034.


charleskawczynski:
Looks good.

PR #3382, opened 2024-10-14
Support reproducible restarts with RRTMGP.jl

sriharshakandala:
## Purpose 
Support reproducible restarts with RRTMGP.jl.
Seed random number generator with simulation time before calling `update_fluxes`.
The seeding can be turned on by setting `radiation_reset_rng_seed` to `true`. The default is set to `false`.

----


sriharshakandala:
The `end to end test` calibration test seems fail frequently, potentially due to problems with slurm daemon. This is unrelated to the current PR. @nefrathenrici is working on setting this test to `soft fail`

Sbozzolo (commented):
Note, this change will make all the call to `Random.seed!` in all our drivers ineffective. As a result:
- all the results should change
- the nightly AMIP will break (because the nightly AMIP relies on setting the seed to run an ensemble) (@juliasloan25 @szy21 @trontrytel)

I suggest we make this option false by default and turn it on for the jobs we care about. In the nightly AMIP, we could manually add some small initial perturbation, or add a controllable offset to the seed.

Can you also update the NEWS file to describe this change and its implications?

Ideally, we also would like this to be tested, but the tests for restarts have no clouds. I don't know if we have any easy way to add this.

More broadly, given that this PR is not changing MSEs, I think this means that we are not testing any regime with clouds in our reproducibility tests.

sriharshakandala:
@nefrathenrici : I am setting the `calibration_end_to_end_test` to soft fail in the CI pipeline here!

src/callbacks/callbacks.jl
```diff
@@ -246,7 +246,7 @@ NVTX.@annotate function rrtmgp_model_callback!(integrator)
 
     set_surface_albedo!(Y, p, t, p.atmos.surface_albedo)
 
-    RRTMGPI.update_fluxes!(rrtmgp_model)
+    RRTMGPI.update_fluxes!(rrtmgp_model, UInt32(t / integrator.p.dt))
```
trontrytel, line 249: So for nightly tiny AMIP runs we want this to be deterministic but different among the 4 runs? Would it be possible to make that also a function of slurm `job_id` or something that could be automatically different between the 4 runs?

szy21: I think for nightly amip we will just have the default `radiation_seed_rng: false`. We only set it to true when debugging a restarting run.

trontrytel: Ok, makes sense!

szy21: Oh, just saw @Sbozzolo's comments. Why would this change the behavior when using `radiation_seed_rng: false`?

Sbozzolo: > Oh, just saw @Sbozzolo's comments. Why would this change the behavior when using `radiation_seed_rng: false`?

It wouldn't, but I think it would be preferable to set this to true for our current AMIPs. This would allow us to directly use one of the restarts for debugging. When this is false, restarting a simulation leads to a different evolution.

szy21: Sounds good. The nightly amip now has one deterministic (unique_seed: false) and three random runs. I think most of the time we will use the deterministic run for debugging, so maybe we can set it to true for the deterministic run only. Adding a perturbation to the initial condition is a good idea though, as we will need it at some point.

Sbozzolo: >  I think most of the time we will use the deterministic run for debugging, so maybe we can set it to true for the deterministic run only.

Sounds good!

szy21 (commented):
Looks good to me after addressing Gabriele's comments, thanks!

szy21:
> More broadly, given that this PR is not changing MSEs, I think this means that we are not testing any regime with clouds in our reproducibility tests.

Yes, that's probably true. We will change the initial condition to have some moisture so it will take shorter time to make clouds.

sriharshakandala:
If there are no additional comments, I can go ahead and merge!

src/parameterized_tendencies/radiation/RRTMGPInterface.jl
```diff
@@ -25,12 +26,14 @@ struct AllSkyRadiation <: AbstractRRTMGPMode
     idealized_clouds::Bool
     add_isothermal_boundary_layer::Bool
     aerosol_radiation::Bool
+    reset_rng_seed::Bool
```
Sbozzolo, line 29: Can you please add a docstring/comment to this? Something like

```suggestion
    "reset_rng_seed is a boolean that controls whether the seed for the random number generator is reset every time the radiation callback is called. This is accomplished by ABC and it is useful for XYZ. Disable this for production runs."
    reset_rng_seed::Bool
```

Same below

src/solver/model_getters.jl
```diff
@@ -225,6 +225,8 @@ function get_radiation_mode(parsed_args, ::Type{FT}) where {FT}
     @assert add_isothermal_boundary_layer in (true, false)
     aerosol_radiation = parsed_args["aerosol_radiation"]
     @assert aerosol_radiation in (true, false)
+    reset_rng_seed = parsed_args["radiation_reset_rng_seed"]
+    @assert reset_rng_seed in (true, false)
```
Sbozzolo, line 229: Can you please add a warning when this option is turned on for radiation models that are not RRTGMP and allsky? Users should be notified that something that they explicitely asked is incompatible with their setup.

NEWS.md
```diff
@@ -4,6 +4,15 @@ ClimaAtmos.jl Release Notes
 Main
 -------
 
+### Reproducible restarts for simulations with clouds with RRTMGP
+
+- Reset the RNG seed before calling RRTGMP to a known value (the iteration number). 
+  When modeling cloud optics, RRTGMP uses a random number generator. Resetting 
+  the seed every time RRTGMP is called to a deterministic value ensures that the 
+  simulation is fully reproducible and can be restarted in a reproducible way.
+  Disable this option when running production runs.
```
Sbozzolo, line 13:
```suggestion
- Reset the RNG seed before calling RRTGMP to a known value (the iteration number). 
  When modeling cloud optics, RRTGMP uses a random number generator. Resetting 
  the seed every time RRTGMP is called to a deterministic value ensures that the 
  simulation is fully reproducible and can be restarted in a reproducible way.
  Disable this option when running production runs. 
  
  Note: Setting this option to `true` is behavior-changing. 
```

Sbozzolo (commented):
(I think two of my comments were not addressed, I left them again.)

Can you also please make the required change in ClimaCopuler?

src/solver/model_getters.jl
```diff
@@ -237,6 +239,9 @@ function get_radiation_mode(parsed_args, ::Type{FT}) where {FT}
         "TRMM_LBA",
         "ISDAC",
     )
+    if !(radiation_name in ("allsky", "allskywithclear")) && reset_rng_seed
+        @warn "reset_rng_seed does not have any effect with " * radiation_name "radiation option"
```
Sbozzolo, line 243: There is a missing `*` between `radiation_name` and `radiation_option`. 

I think the more Julian way would be to use string interpolation instead of concatenation:
```suggestion
        @warn "reset_rng_seed does not have any effect with rad = $(radiation_name)"
```

sriharshakandala: There is no actual difference here!

Sbozzolo (approved):
Look good, thank you!

(Please fix the bug with the string concatentation before merging)

Also, could you please take care of revelant changes in ClimaCoupler (to the nightly AMIP) as well?

.buildkite/pipeline.yml
```diff
@@ -566,6 +566,7 @@ steps:
       - label: "end to end test"
         command: julia --project=calibration/test calibration/test/e2e_test.jl
         artifact_paths: "calibration_end_to_end_test/*"
+        soft_fail: true
```
charleskawczynski, line 569: Is this being added due to the timeouts?

sriharshakandala: Yes. @nefrathenrici suggested there are some problems with slurm daemon and its best to set it to soft_fail for now!

nefrathenrici: Yes, IMSS have said the slurm database daemon is down intermittently because it is hosted on  the new storage system.
 ClimaCalibrate uses `sacct` to query jobs and has been getting this error:
```
sacct: error: slurm_persist_conn_open_without_init: failed to open persistent connection to host:head1:6819: Connection refused
sacct: error: Sending PersistInit msg: Connection refused
sacct: error: Problem talking to the database: Connection refused
```

the `build_history` script in buildkite also failed because of this https://buildkite.com/clima/climacoupler-ci/builds/4928#01929cd1-788d-40b6-bbcc-bf0972b32acf/202-210

charleskawczynski (approved):

juliasloan25:
Is this a breaking change (should I make today's release a minor release)?

Sbozzolo:
It's not breaking

PR #3390, opened 2024-10-18
tag v0.27.7

juliasloan25:
## Purpose 
weekly release
update dependencies in all checked-in manifests, and tag a patch release

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,21 +4,24 @@ ClimaAtmos.jl Release Notes
 Main
 -------
 
+v0.27.7
+-------
+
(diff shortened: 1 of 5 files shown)


szy21 (approved):
Let me make a new patch release of Thermodynamics first, which contains a bug fix. But maybe this doesn't need to wait?

juliasloan25:
> Let me make a new patch release of Thermodynamics first, which contains a bug fix. But maybe this doesn't need to wait?

It's okay to merge this first. After you make the patch release in Thermodynamics, we should update the Thermodynamics compat entries in ClimaAtmos and ClimaCoupler so that is the lower bound

PR #3396, opened 2024-10-23
Add check for accumulated diagnostics and restarts

### LINKED ISSUE #2604: Make sure restart works correctly

### LINKED ISSUE #3309: Add check that checkpoints and diagnostics output are synced up
As we move towards long simulations, checkpoints become more and more important. While adding support for diagnostics that work across restarts is possible, that's something that is not easy to implement (without imposing various restrictions).

As a first step, it would be good to at least ensure that diagnostics and checkpoitns are output at the same schedule (e.g., every month)

Sbozzolo:
This commit adds a warning when we can detect checkpointing and diagnostics being out of sync. ClimaAtmos now detects what frequencies are involved and print them out, e.g.
```
[ Info: Saving accumulated diagnostics to disk with frequency: 2 days, 1 day
```
Then, it checks that all these frequencies evenly divide the checkpointing frequency. If this does not happen, a warning is produced.

Closes #3309 
Closes #2604


src/callbacks/get_callbacks.jl
```diff
@@ -169,9 +189,28 @@ function get_diagnostics(parsed_args, atmos_model, Y, p, dt, t_start)
         @info "$writer_str: $diags_outputs"
     end
 
-    return diagnostics, writers
+    return diagnostics, writers, periods_reductions
 end
 
+function checkpoint_frequency_from_parsed_args(dt_save_state_to_disk)
```
charleskawczynski, line 195: Can we add a type annotation for `dt_save_state_to_disk`? This is a string, right?

src/utils/utilities.jl
```diff
@@ -454,3 +454,45 @@ function isdivisible(
...
+2 weeks
+
+julia> promote_period(Day(365))
+1 year
+
+julia> promote_period(Millisecond(1))
+1 millisecond
+```
+"""
+function promote_period(period::Dates.Period)
```
charleskawczynski, line 481: Should we maybe call this `round_period`? Promote kind of sounds like the value is not changing, but it may (even if it's just approximate)

Sbozzolo: Indeed, the value doesn't change. This "promotes" only when it is possible to use a single unit. For example, if you try promoting 86401 seconds, you will get back 86401 seconds (but if you promote 86400 seconds, you get 1 day). To me `round` implies that there might be loss of precision. 

I don't know what better word to use.

charleskawczynski: Yeah, `promote_period(Day(365)) => 1 year` seems like it's rounding

Sbozzolo: Ah, right, I removed that because indeed it was not possible without rounding! Thanks for catching that.

src/utils/utilities.jl
```diff
@@ -454,3 +454,49 @@ function isdivisible(
...
+    # Hard to do this with varying periods like Month/Year...
+    PeriodTypes = [
+        Dates.Week,
+        Dates.Day,
+        Dates.Hour,
+        Dates.Minute,
+        Dates.Second,
+        Dates.Millisecond,
+    ]
+    for PeriodType in PeriodTypes
```
charleskawczynski, line 496: Should we throw an error after this loop if none of the `ms % period_ms == 0` are true?

Sbozzolo: That's not possible: we convert everything to `ms`, so `Dates.Millisecond` will always satisfy the condition.

szy21 (approved):

charleskawczynski (approved):

PR #3397, opened 2024-10-23
remove unnecessary FT for type restriction

szy21:
## Purpose 

This is needed for autodiff to work.

## Content

----

### DIFF
--- a/src/cache/precomputed_quantities.jl
+++ b/src/cache/precomputed_quantities.jl
@@ -325,18 +325,7 @@ function eddy_diffusivity_coefficient(C_E, norm_v_a, z_a, p)
     K_E = C_E * norm_v_a * z_a
     return p > p_pbl ? K_E : K_E * exp(-((p_pbl - p) / p_strato)^2)
 end
-function eddy_diffusivity_coefficient(
-    z::FT,
-    z₀,
-    f_b::FT,
-    h::FT,
-    uₐ,
-    C_E::FT,
-    Ri::FT,
-    Ri_a::FT,
-    Ri_c::FT,
-    κ::FT,
-) where {FT}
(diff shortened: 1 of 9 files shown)


src/cache/precomputed_quantities.jl
szy21, line 1: @charleskawczynski @dennisYatunin Could you take a look at whether this is a preferred pattern? If so I will go ahead and change all the functions in the code.

charleskawczynski: This is probably fine. We could even remove `::FT` in `z::FT` (and `where {FT}`) and replace `FT(0)` with `zero(z)`. The main risk with this is that if we do accidentally introduce type instabilities, this function could majorly slow down-- and diagonal dispatch will protect us from happening. But if the syntax is too burdensome, then I'm fine with trying to remove it.

src/utils/utilities.jl
```diff
@@ -25,7 +25,7 @@ has_no_source_or_sink(parsed_args) = all((
 
 # we may be hitting a slow path:
 # https://stackoverflow.com/questions/14687665/very-slow-stdpow-for-bases-very-close-to-1
-fast_pow(x::FT, y::FT) where {FT <: AbstractFloat} = exp(y * log(x))
+fast_pow(x::FT, y) where {FT <: AbstractFloat} = exp(y * log(x))
```
charleskawczynski, line 28:
```suggestion
fast_pow(x, y) = exp(y * log(x))
```
We can remote `FT` entirely if it's not used

szy21:
@charleskawczynski Do you have more comments on this PR?

charleskawczynski (approved):
No, I think the changes are fine

PR #3398, opened 2024-10-24
Try using purely vertical spaces for column configurations

charleskawczynski:
While looking at calibrations, I realized that we seem to always generate `VIJFH` datalayouts, even for column cases. Our CPU and GPU code paths for each datalayout is split so that we can optimize kernels specifically for those datalayouts. So, we should probably allow these spaces to be flexible so that we get the most out of the true datalayout.

### DIFF
--- a/src/utils/common_spaces.jl
+++ b/src/utils/common_spaces.jl
@@ -57,29 +57,6 @@ function make_horizontal_space(
     return space
 end
 
-function make_column_spaces(
-    z_max,
-    z_elem,
-    z_stretch,
-    comms_ctx::ClimaComms.SingletonCommsContext,
-)
-    z_domain = Domains.IntervalDomain(
-        Geometry.ZPoint(zero(z_max)),
-        Geometry.ZPoint(z_max);
-        boundary_names = (:bottom, :top),
-    )
-    z_mesh = Meshes.IntervalMesh(z_domain, z_stretch; nelems = z_elem)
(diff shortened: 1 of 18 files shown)


Sbozzolo:
I also would like to see this. Note that this will require also fixing all the postprocessing (that assumed that columns were actually boxes) and fixing this: https://github.com/CliMA/ClimaCore.jl/issues/2032

src/utils/common_spaces.jl
```diff
@@ -56,6 +56,29 @@ function make_horizontal_space(
...
+    )
+    z_mesh = Meshes.IntervalMesh(z_domain, z_stretch; nelems = z_elem)
+    @info "z heights" z_mesh.faces
+    device = ClimaComms.device(comms_ctx)
+    z_topology = Topologies.IntervalTopology(
+        ClimaComms.SingletonCommsContext(device),
+        z_mesh,
+    )
+    cspace = Spaces.CenterFiniteDifferenceSpace(z_topology)
+    fspace = Spaces.CenterFiniteDifferenceSpace(cspace)
```
akshaysridhar, line 78:
```suggestion
    fspace = Spaces.FaceFiniteDifferenceSpace(cspace)
```

charleskawczynski:
I think we de-prioritized this effort because there are assumptions and annoyances around ClimaCore's Geometry module (vertical column `LocalGeometry`s do not have horizontal components, but they should).

PR #3399, opened 2024-10-24
Use single kernel for pointwise functions

sriharshakandala:
## Purpose 
Use single kernel for point-wise functions.

----


src/parameterized_tendencies/microphysics/microphysics_wrappers.jl
```diff
@@ -151,112 +152,193 @@ function compute_precipitation_sources!(
...
+        device,
+        dims,
+        compute_precipitation_sources_kernel!,
+        fvargs...,
+        args...,
+    )
+    return nothing
+end
+
+function compute_precipitation_sources_kernel!(
```
charleskawczynski, line 184: Hi @sriharshakandala, thanks for looking into this. But, what do we do when a user wants to add a new term in here that requires a derivative? If the answer is to split up the kernel, then this seems like a very inflexible solution compared to #3371, which should be able to achieve the same thing as what is done here. One of my bigger concerns, however, is that we cannot easily unit test and refactor `compute_precipitation_sources_kernel!`, as we move towards a more lazy representation that will allow us to fuse more kernels still.

sriharshakandala: This is only for pointwise functions. Spectral element derivatives are not pointwise functions!

sriharshakandala: I do not see any limitations with testing code with this approach.

charleskawczynski: > This is only for pointwise functions. Spectral element derivatives are not pointwise functions!

Yeah, but that means that if a user wants to try adding a gradient somewhere that they would need to break up this function, which seems like a lot of work just to add one term.

Also, this example has a collection of floats, how would that look for this example? https://github.com/CliMA/ClimaAtmos.jl/pull/3371/files#diff-471fb2001bc0d6f84924896fd8c99bd3d4edcc84bd5d79da9e9f35afef17f9cdR491-R497

tapios: In my view, the point of what Sriharsha is doing here is not so much establish a pattern for refactoring the code, but get a relatively simple example going that allows us to quantitatively assess the performance benefits of kernel fusion. So I'd suggest we table discussions of how to realize the benefits until we have a better sense what the benefits are.

charleskawczynski: Ok, yes, that makes sense. This would be a nice example to see the speed up.

PR #3400, opened 2024-10-25
update coupled set_surface_albedo

juliasloan25:
## Purpose 
Currently, ClimaCoupler.jl overwrites the `set_surface_albedo!` function defined in ClimaAtmos (see [here](https://github.com/CliMA/ClimaCoupler.jl/blob/b278a50316717050065951bb899e2a1da0f184d5/experiments/ClimaEarth/components/atmosphere/climaatmos.jl#L79-L94)). However, both functions have exactly the same type signature so this produces a warning. I'm moving the version we want to use into ClimaAtmos so we can remove it from ClimaCoupler.

## Content

----

### DIFF
--- a/src/cache/surface_albedo.jl
+++ b/src/cache/surface_albedo.jl
@@ -101,11 +101,28 @@ function set_surface_albedo!(
 end
 
 """
-    set_surface_albedo!(::CouplerAlbedo)
+    set_surface_albedo!(Y, p, t, ::CouplerAlbedo)
 
 Tell the ClimaAtmos to skip setting the surface albedo, as it is handled by the coupler.
+
+When running in a coupled simulation, set the surface albedo to 0.38 at the beginning of the simulation,
+so the initial callback initialization doesn't lead to NaNs in the radiation model.
+Subsequently, the surface albedo will be updated by the coupler.
 """
-set_surface_albedo!(Y, p, t, ::CouplerAlbedo) = nothing
+function set_surface_albedo!(Y, p, t, ::CouplerAlbedo)
+    if t == 0
+        FT = eltype(Y)
+        # set initial insolation initial conditions
+        !(p.atmos.insolation isa IdealizedInsolation) &&
+            set_insolation_variables!(Y, p, t, p.atmos.insolation)
+        # set surface albedo to 0.38
+        @warn "Setting surface albedo to 0.38 at the beginning of the simulation"
+        p.radiation.rrtmgp_model.direct_sw_surface_albedo .= FT(0.38)
+        p.radiation.rrtmgp_model.diffuse_sw_surface_albedo .= FT(0.38)
+    else
+        nothing
+    end
+end
 
 """
     surface_albedo_direct(α_model::RegressionFunctionAlbedo{FT}) where {FT}
--- a/test/surface_albedo.jl
+++ b/test/surface_albedo.jl
@@ -37,14 +37,19 @@ redirect_stderr(IOContext(stderr, :stacktrace_types_limited => Ref(false)))
     config.parsed_args["rad"] = "clearsky"
     config.parsed_args["albedo_model"] = "CouplerAlbedo"
     simulation = ClimaAtmos.get_simulation(config)
-    (; u, p, t) = simulation.integrator
+    (; u, p) = simulation.integrator
 
-    ClimaAtmos.set_surface_albedo!(u, p, t, p.atmos.surface_albedo)
+    ClimaAtmos.set_surface_albedo!(u, p, Float64(0), p.atmos.surface_albedo)
 
-    # test that the albedo is not initiated by atmos (includes NaNs)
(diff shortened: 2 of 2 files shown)


juliasloan25:
> [here](https://github.com/CliMA/ClimaAtmos.jl/blob/09f7afff5472fccd499822cfec6dd62986aa57e3/src/callbacks/callbacks.jl#L242)

It will set surface albedo there only at time `t=0` for coupled simulations. Otherwise it doesn't do anything. Maybe there's a better way to do this where we just initialize the value at 0.38?

szy21:
> > If we do this, will it set surface albedo to 0.38 [here](https://github.com/CliMA/ClimaAtmos.jl/blob/09f7afff5472fccd499822cfec6dd62986aa57e3/src/callbacks/callbacks.jl#L242) during radiation callback when running a coupled simulation?
> 
> It will set surface albedo to 0.38 there only at time `t=0` for coupled simulations. Otherwise it doesn't do anything. Maybe there's a better way to do this where we just initialize the value at 0.38?

Oh right I didn't realize there is a check for t=0. How about restarting simulations? Maybe we don't have to worry about it for now as I don't think the restart in the coupler is exactly correct anyway.

szy21 (approved):

juliasloan25:
> > > If we do this, will it set surface albedo to 0.38 [here](https://github.com/CliMA/ClimaAtmos.jl/blob/09f7afff5472fccd499822cfec6dd62986aa57e3/src/callbacks/callbacks.jl#L242) during radiation callback when running a coupled simulation?
> > 
> > 
> > It will set surface albedo to 0.38 there only at time `t=0` for coupled simulations. Otherwise it doesn't do anything. Maybe there's a better way to do this where we just initialize the value at 0.38?
> 
> Oh right I didn't realize there is a check for t=0. How about restarting simulations? Maybe we don't have to worry about it for now as I don't think the restart in the coupler is exactly correct anyway.

After t=0, the albedo gets updated in ClimaCoupler with an hourly callback of [water_albedo_from_atmosphere!](https://github.com/CliMA/ClimaCoupler.jl/blob/b278a50316717050065951bb899e2a1da0f184d5/src/FluxCalculator.jl#L420-L431). I think this would make restarts correct, unless `t` resets to 0 in a restart (I assumed it just resumes from whatever it was at when the previous simulation ended)

szy21:
> > > > If we do this, will it set surface albedo to 0.38 [here](https://github.com/CliMA/ClimaAtmos.jl/blob/09f7afff5472fccd499822cfec6dd62986aa57e3/src/callbacks/callbacks.jl#L242) during radiation callback when running a coupled simulation?
> > > 
> > > 
> > > It will set surface albedo to 0.38 there only at time `t=0` for coupled simulations. Otherwise it doesn't do anything. Maybe there's a better way to do this where we just initialize the value at 0.38?
> > 
> > 
> > Oh right I didn't realize there is a check for t=0. How about restarting simulations? Maybe we don't have to worry about it for now as I don't think the restart in the coupler is exactly correct anyway.
> 
> After t=0, the albedo gets updated in ClimaCoupler with an hourly callback of [water_albedo_from_atmosphere!](https://github.com/CliMA/ClimaCoupler.jl/blob/b278a50316717050065951bb899e2a1da0f184d5/src/FluxCalculator.jl#L420-L431). I think this would make restarts correct, unless `t` resets to 0 in a restart (I assumed it just resumes from whatever it was at when the previous simulation ended)

I think t is reset to 0 in a coupled restarting simulation, but I'm not sure.

juliasloan25:
> I think t is reset to 0 in a coupled restarting simulation, but I'm not sure.

I just took a look - it's controlled by a config variable `restart_t`, which has a default value of 0 but is overwritten in our one [restart test](https://github.com/CliMA/ClimaCoupler.jl/blob/b278a50316717050065951bb899e2a1da0f184d5/test/mpi_tests/local_checks.sh#L35) (which isn't the most reliable piece of code). We need to make sure that when we run a restart simulation, we set `restart_t` to the `t_end` of the simulation we're restarting from. 

Since this PR is using the same code that was previously in ClimaCoupler, just moving it to ClimaAtmos, I think it's okay to keep as-is (and, like you said, keep this in mind as we work out restarting coupled simulations)

szy21:
> > I think t is reset to 0 in a coupled restarting simulation, but I'm not sure.
> 
> I just took a look - it's controlled by a config variable `restart_t`, which has a default value of 0 but is overwritten in our one [restart test](https://github.com/CliMA/ClimaCoupler.jl/blob/b278a50316717050065951bb899e2a1da0f184d5/test/mpi_tests/local_checks.sh#L35) (which isn't the most reliable piece of code). We need to make sure that when we run a restart simulation, we set `restart_t` to the `t_end` of the simulation we're restarting from.
> 
> Since this PR is using the same code that was previously in ClimaCoupler, just moving it to ClimaAtmos, I think it's okay to keep as-is (and, like you said, keep this in mind as we work out restarting coupled simulations)

Sounds good, feel free to merge it.

PR #3402, opened 2024-10-25
Make the default config table searchable in docs

### LINKED ISSUE #3301: julia (`pretty_table`) output to docs is not searchable
Consider the [Default Configuration](https://clima.github.io/ClimaAtmos.jl/dev/config/) doc section, which enumerates arguments to the config file.

The text in this table in not discoverable from the "Search docs" page in the upper-left side of the webpage:
image

------
I came across this looking for information about the Rayleigh sponge, which has a config entry:

image

but there are no results in search:

image

imreddyTeja:
The content of the old config.md was moved into config_no_table.md with the include("config_table.jl") at the bottom removed.

Now, config_table.jl, creates the config.md by reading config_no_table.md and copying the lines into it. Then, it generates the pretty table and adds it to the end of the file.

This allows the table to be searchable, but it does not require users to manually update the docs when the default config yaml is modified.

## Purpose 
closes #3301 

## Content

----


Sbozzolo:
I think the file should not be checked in with git, since it is dynamically generated

Sbozzolo (approved):

PR #3403, opened 2024-10-26
Try hf datalayouts

charleskawczynski:
This upgrades to the main ClimaCore branch, and tries out the HF datalayouts.


charleskawczynski:
We can now try this without env changes

PR #3408, opened 2024-10-29
fix indivisible reduction periods warning

haakon-e:
As far as I can tell, #3396 implemented the warning for indivisible frequencies incorrectly.

e.g., if 
```julia
periods_reductions = Set([Dates.Second(3600)])  # every hour
checkpoint_frequency = Dates.Millisecond(1800000)  # every 30 minutes, i.e. the value of dt_save_state_to_disk
```
I currently get the warning: `Warning: Some accumulated diagnostics might not be evenly divisible by the checkpointing frequency (30 minutes)`

but, unless I am misunderstanding what's going on here, this should be perfectly fine.

------

This PR essentially reverses the order of the arguments to `CA.isdivisible` so that we now have:
```julia
if any(
    x -> !CA.isdivisible(x, checkpoint_frequency),
    periods_reductions,
)
    accum_str = join(CA.promote_period.(collect(periods_reductions)), ", ")
    checkpt_str = CA.promote_period(checkpoint_frequency)
    @warn "Diagnostics are accumulated at frequencies ($accum_str), which may not be evenly divisible by the checkpointing frequency (dt_save_state_to_disk = $checkpt_str)"
end
```

### DIFF
--- a/src/solver/type_getters.jl
+++ b/src/solver/type_getters.jl
@@ -777,7 +777,8 @@ function get_simulation(config::AtmosConfig)
                 x -> !CA.isdivisible(x, checkpoint_frequency),
                 periods_reductions,
             )
-                accum_str = join(CA.promote_period.(collect(periods_reductions)), ", ")
+                accum_str = 
+                    join(CA.promote_period.(collect(periods_reductions)), ", ")
                 checkpt_str = CA.promote_period(checkpoint_frequency)
                 @warn "Diagnostics are accumulated at frequencies ($accum_str), which may not be evenly divisible by the checkpointing frequency (dt_save_state_to_disk = $checkpt_str)"
             end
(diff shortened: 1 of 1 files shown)


Sbozzolo:
I think the warning is correct: if you have hourly diagnostics, but checkpoints every 30 minutes, you could restart the simulation from an invalid state if you restart from odd checkpoints.

haakon-e:
I see. So I did indeed misunderstand it then. I think I got caught up in the phrasing 

> Some accumulated diagnostics might not be evenly **divisible by** the checkpointing frequency

suggesting that we check if `(accumulated diagnostics) / (checkpointing frequency)` is an integer, but we actually do the opposite.

Sbozzolo:
I think you are right. Feel free to flip the phrasing in this PR.

haakon-e:
something like this?

PR #3410, opened 2024-10-29
Allow for flakiness in reproducibility tests

charleskawczynski:
This PR changes our reproducibility tests to allow for flakiness:

 - `self_reference_or_path` -> `latest_comparable_paths`, which now returns a list of paths
 - Users can now specify `--test_broken_report_flakiness true`, which will allow failures if the reproducibility test comes back flakey, but will _fail_ if it's successfully compared against 5 previous runs, so that users can remove the `test_broken_report_flakiness` keyword.

This way, we can more widely add these reproducibility tests, and get some measure as to which tests are reproducible, and by how much.

Co-authored by @Sbozzolo


reproducibility_tests/test_mse.jl
```diff
@@ -20,22 +23,53 @@ function get_params()
...
-    computed_mse_filename;
-    dicttype = OrderedCollections.OrderedDict,
-)
 
 import ClimaReproducibilityTests as CRT
 using Test
-for (var, reproducible) in CRT.test_mse(; computed_mse)
-    @test reproducible
+is_mse_file(x) = startswith(x, "computed_mse") && endswith(x, ".json")
+computed_mse_filenames = filter(is_mse_file, readdir(out_dir))
```
charleskawczynski, line 39: Do we want to limit the length of `computed_mse_filenames`?

reproducibility_tests/compute_mse.jl
```diff
@@ -61,86 +61,96 @@ function reproducibility_test(;
 )
     local ds_filename_reference
     reference_keys = map(k -> varname(k), collect(keys(reference_mse)))
+    paths = nothing # initialize for later handling
 
     if haskey(ENV, "BUILDKITE_COMMIT")
-        path = self_reference_or_path()
-        path == :self_reference && return reference_mse
-        ds_filename_reference = joinpath(path, ds_filename_computed)
+        paths = self_reference_or_paths()
```
charleskawczynski, line 67: Should we limit the number of paths here?

reproducibility_tests/test_mse.jl
charleskawczynski: I think that this is already limited by the number of paths returned.

reproducibility_tests/compute_mse.jl
charleskawczynski: Done

charleskawczynski:
> Can you please add/update documentation?

I did (in the readme), is there something that can be improved?

Sbozzolo:
> > Can you please add/update documentation?
> 
> I did (in the readme), is there something that can be improved?

Yes, the readme doesn't discuss this new feature you added. As of right now, users have way to understand what `test_broken_report_flakiness` does without reading the code, and developers have no place where they can learn about the logic behind this feature.

charleskawczynski:
One thing I realized is that, I think this new workflow requires that we move data on every merged PR, and not just when the reference counter is updated.

Sbozzolo:
> One thing I realized is that, I think this new workflow requires that we move data on every merged PR, and not just when the reference counter is updated.

Yes, I think that's what we discussed. We also discussed the alternative that all the experiments are moved when the reference counter is updated (so that we'd compare against the past N times the ref counter was incremented). 

The idea I liked was to have a folder structure that follows the ref counter with inside all the commits associated to that reference. Like:
```
187
├── 73hasd
├── fgsae7
└── sdf63a
188
....
```

charleskawczynski:
> > One thing I realized is that, I think this new workflow requires that we move data on every merged PR, and not just when the reference counter is updated.
> 
> Yes, I think that's what we discussed. We also discussed the alternative that all the experiments are moved when the reference counter is updated (so that we'd compare against the past N times the ref counter was incremented).
> 
> The idea I liked was to have a folder structure that follows the ref counter with inside all the commits associated to that reference. Like:
> 
> ```
> 187
> ├── 73hasd
> ├── fgsae7
> └── sdf63a
> 188
> ....
> ```

I agree that this is a good idea, but it's not strictly needed for allowing flakiness in the reproducibility tests, and doing this changes how we find files in this logic, so I opened [an issue](https://github.com/CliMA/ClimaAtmos.jl/issues/3421) to address this improvement.

reproducibility_tests/README.md
```diff
@@ -18,15 +18,22 @@ Our solution to dealing with failure modes is by providing users with two workfl
...
 
- - When a reference dataset does not exist, we still perform a reproducibility test so that we continuously exercise the testing infrastructure. However, we compare the solution dataset with itself (which we call a "self-reference"). Therefore, _all reproducibility tests for all jobs will pass_ (no matter what the results look like) when the reference counter is incremented. So, it is important to review the quality of the results when the reference counter is incremented.
+ - When a reference dataset does not exist, we still perform a reproducibility test so that we continuously exercise the testing infrastructure. However, we compare the solution dataset with itself. Therefore, _all reproducibility tests for all jobs will pass_ (no matter what the results look like) when the reference counter is incremented. So, it is important to review the quality of the results when the reference counter is incremented.
 
  - Every time the reference counter is incremented, data from that PR is saved onto Caltech's central cluster. And that solution's dataset is the new reference dataset that all future PRs are compared against (until the reference counter is incremented again).
```
Sbozzolo, line 28: Is this still accurate? (Are we saving data at each PR?)

reproducibility_tests/README.md
```diff
@@ -89,28 +96,29 @@ To think about tracking which dataset to compare against, it's helpful to consid
...
  3) Upon next CI run, before performing CI test,
-    we check if the counter indicates a self-reference by
-    checking if `reproducibility_tests/ref_counter.jl` in the PR
-    matches (e.g.,) `aRsVoY/ref_counter.jl` in the last
-    merged commit (on central). If yes, then it's a self
-    reference, if not, then we look-up the dataset based
-    on the counter.
+    we check if the counter indicates the existence of comparable
+    references by checking if `reproducibility_tests/ref_counter.jl`
+    in the PR matches (e.g.,) `aRsVoY/ref_counter.jl` in the last
```
Sbozzolo, line 120: This `(e.g.,)` is missing something

reproducibility_tests/latest_comparable_paths.jl
```diff
@@ -0,0 +1,110 @@
...
+function sorted_dataset_folder(; dir = pwd())
+    matching_paths = String[]
+    for file in readdir(dir)
+        !ispath(joinpath(dir, file)) && continue
+        push!(matching_paths, joinpath(dir, file))
+    end
+    isempty(matching_paths) && return ""
+    # sort by timestamp
+    sorted_paths =
+        sort(matching_paths; by = f -> Dates.unix2datetime(stat(f).mtime))
```
Sbozzolo, line 12:
```suggestion
        sort(matching_paths; by = f -> stat(f).mtime)
```
I don't think you need to convert to DateTime for this to work

reproducibility_tests/README.md
charleskawczynski: Ah, no, I missed that one in some recent commits. Updating...

charleskawczynski: Wait, that looks stale

reproducibility_tests/latest_comparable_paths.jl
```diff
@@ -0,0 +1,110 @@
+import Dates
+
+function sorted_dataset_folder(; dir = pwd())
+    matching_paths = String[]
+    for file in readdir(dir)
+        !ispath(joinpath(dir, file)) && continue
+        push!(matching_paths, joinpath(dir, file))
+    end
```
Sbozzolo, line 8:
```suggestion
    matching_paths = filter(isdir, readdir(dir; join=true))
```
I think that this is equivalent, if I understand the intent of `!ispath`

reproducibility_tests/README.md
charleskawczynski: Oh, I was looking at the wrong section. Yes, we move data every time the reference is incremented.

reproducibility_tests/latest_comparable_paths.jl
```diff
@@ -0,0 +1,110 @@
+import Dates
+
+function sorted_dataset_folder(; dir = pwd())
```
Sbozzolo, line 3:
```suggestion
""
    sorted_dataset_folder(; dir=pwd())

Return a the subdirectory paths within the given `dir` (defaults to the current working directory) sorted by modification time (oldest to newest). 

Return an empty vector if no subdirectories are found.
"""
function sorted_dataset_folder(; dir = pwd())
```

reproducibility_tests/latest_comparable_paths.jl
```diff
@@ -0,0 +1,110 @@
...
+        push!(matching_paths, joinpath(dir, file))
+    end
+    isempty(matching_paths) && return ""
+    # sort by timestamp
+    sorted_paths =
+        sort(matching_paths; by = f -> Dates.unix2datetime(stat(f).mtime))
+    return sorted_paths
+end
+
+function ref_counters_per_path(paths)
```
Sbozzolo, line 16:
```suggestion
""
    ref_counters_per_path(paths)

Read the `ref_counter.jl` file in each given path and parses the integer it contains.

Return a vector of integers, where each element corresponds to a path. 
If a path does not contain a `ref_counter.jl` file, the corresponding element is -1.

It assumes that `ref_counter.jl` contains the value as the first line of the file.
"""
function ref_counters_per_path(paths)
```

reproducibility_tests/latest_comparable_paths.jl
```diff
@@ -0,0 +1,110 @@
...
+
+    ref_counter_file_PR = joinpath(@__DIR__, "ref_counter.jl")
+    @assert isfile(ref_counter_file_PR)
+    ref_counter_PR = parse(Int, first(readlines(ref_counter_file_PR)))
+
+    ref_counters_main = ref_counters_per_path(sorted_paths)
+    i_comparable_references = findall(ref_counters_main) do ref_counter_main
+        ref_counter_main == ref_counter_PR
+    end
+    if i_comparable_references == nothing
```
Sbozzolo, line 63:
```suggestion
    if isnothing(i_comparable_references)
```
or
```suggestion
    if i_comparable_references === nothing
```

reproducibility_tests/latest_comparable_paths.jl
```diff
@@ -0,0 +1,110 @@
...
+        @warn "`ref_counter.jl` not found on main, assuming no comparable references"
+        @info "Please review output results before merging."
+        return String[]
+    end
+    @info "Found $(length(i_comparable_references)) comparable references:$i_comparable_references"
+    # Largest ref-counter reference path:
+    paths = map(i -> sorted_paths[i], i_comparable_references)
```
Sbozzolo, line 70:
```suggestion
    paths = sorted_paths[i_comparable_references]
```

reproducibility_tests/latest_comparable_paths.jl
```diff
@@ -0,0 +1,110 @@
...
+
+    # for p in paths
+    #     @info "Files in $p:" # for debugging
+    #     for file_on_main in readdir(p)
+    #         @info "   File:`$file_on_main`"
+    #     end
+    # end
+    @assert all(isfile, ref_counter_files_main)
+    ref_counters_main =
+        map(f -> parse(Int, first(readlines(f))), ref_counter_files_main)
```
Sbozzolo, line 88: I suggest we add a function `read_ref_coutner` because we do this operation often now.

reproducibility_tests/latest_comparable_paths.jl
charleskawczynski: I think `matching_paths = filter(ispath, readdir(dir; join=true))` is equivalent, `isdir` returns `false` for files:

```julia
help?> isdir
search: isdir isdirpath isdigit isdisjoint isdispatchtuple isxdigit isdefined @isdefined redirect_stdio redirect_stdin

  isdir(path) -> Bool

  Return true if path is a directory, false otherwise.

  Examples
  ≡≡≡≡≡≡≡≡

  julia> isdir(homedir())
  true
  
  julia> isdir("not/a/directory")
  false

  See also isfile and ispath.
```

reproducibility_tests/latest_comparable_paths.jl
charleskawczynski: I prefer to use `map` here.

reproducibility_tests/README.md
charleskawczynski: N references. We only compare against merged commits (i.e., on the main branch)

Sbozzolo: > N references. We only compare against merged results (i.e., on the main branch): only the head commit is moved to central, so only 1 folder should be moved per PR.

I am still confused.

N references, as the previous N times the ref counter was incremented?
Or the previous N PRs that were merged (ie, the latest N commits on main)?

reproducibility_tests/README.md
```diff
@@ -18,14 +18,21 @@ Our solution to dealing with failure modes is by providing users with two workfl
...
+ - When a PR passes CI on buildkite while in the github merge queue, or when a PR lands on the main branch, data from the HEAD commit of that PR is saved onto Caltech's central cluster. And that solution's dataset is the new reference dataset that all future PRs are compared against (until the reference counter is incremented again).
+
+## Allowing flaky tests
+
+Users can add the flag `test_broken_report_flakiness` to the `test_mse.jl` script: `julia --project=examples reproducibility_tests/test_mse.jl --test_broken_report_flakiness true`, which will have the following behavior:
+
+ - If the test is not reproducible (i.e., flaky) when compared against `N` appropriate references, then the test will pass and be reported as broken.
+ - If the test is reproducible when compared against `N` appropriate references, then the test will fail `@test_broken`, and users will be asked to fix the broken test. At which point you can remove the `--test_broken_report_flakiness true` flag from that particular job, reinforcing a strict reproducibility constraint.
```
Sbozzolo, line 35: Can you define `appropriate references`?

reproducibility_tests/README.md
```diff
@@ -72,7 +79,7 @@ Reprodicibility tests are performed at the end of `examples/hybrid/driver.jl`, a
  0) Run a simulation, with a particular `job_id`, to the final time.
```
Sbozzolo, line 79: Is there anything that needs to be added to this `Developer guide to reproducibility tests` due to the changes in this PR?

Sbozzolo (commented):
I think this looks good to me. Is it working?

reproducibility_tests/README.md
charleskawczynski: Yeah, sorry, I know it's a bit confusing. It's N references, not N times the ref counter was incremented, that's just an ever increasing number (and we garbage collect old results after a while). Maybe an example will help:

```
187 # PRs against ref counter 187 has 3 references
├── 73hasd
├── fgsae7
└── sdf63a
188 # PRs against ref counter 188 has 2 references
├── 73hasd
└── sdf63a
189 # PRs against ref counter 189 has 4 references
├── 73hasd
├── sd9g8y
├── uhw45l
└── sdf63a
```

reproducibility_tests/README.md
charleskawczynski: Yeah, it should just be "comparable references"

reproducibility_tests/README.md
charleskawczynski: No, the user interaction is effectively the same

reproducibility_tests/README.md
charleskawczynski: I'll update that

reproducibility_tests/README.md
Sbozzolo: Okay, perfect! This is very helpful and I think we should have in the README.

charleskawczynski: Ok, I'll add a bit about this, in anticipation of changing our folder structure.

charleskawczynski:
> I think this looks good to me. Is it working?

I think so? We can't know for sure until we just merge it and see (since code is executed only after we merge the PR 😅)

charleskawczynski:
Let's give this a shot! 🚀

PR #3413, opened 2024-10-30
Make MPI restart test more robust

Sbozzolo:
Sometimes the shared filesystem is slow and the folder is properly not synced across MPI processes. This commit adds an extra check to ensure that all the MPI processes see the tmpfolder before moving forward.


szy21 (approved):

test/restart.jl
```diff
@@ -384,6 +385,10 @@ if MANYTESTS
                             ClimaComms.iamroot(comms_ctx) ? mktempdir(pwd()) :
                             ""
                         output_loc = ClimaComms.bcast(comms_ctx, output_loc)
+                        # Sometimes the shared filesystem doesn't work properly
+                        # and the folder is not synced across MPI processes.
+                        # Let's add an additional check here.
+                        maybe_wait_filesystem(comms_ctx, output_loc)
```
charleskawczynski, line 391: What does `maybe_wait_filesystem` do exactly?

Sbozzolo: It checks that all the MPI processes see that the temporary folder was created. If they don't agree, it waits a tiny bit and check again

PR #3405, opened 2024-10-28
add option to prescribe clouds in radiation

szy21:
## Purpose 

Adds an option to prescribe cloud fields in radiation

Also removes nsys for the 4 gpu jobs as it keeps failing.

## Content

----
- [ ] I have read and checked the items on the review checklist.

PR #3414, opened 2024-10-30
Reduce nsight delay for 4 gpu job

charleskawczynski:
#3376 may have delayed past the simulation time, resulting in no generated reports: https://buildkite.com/clima/climaatmos-ci/builds/21389#0192dde1-c0e6-4be4-a511-c8e1c1271ed5

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -885,7 +885,7 @@ steps:
       - label: "GPU: GPU dry baroclinic wave - 4 gpus"
         key: "target_gpu_implicit_baroclinic_wave_4process"
         command:
-            nsys profile --delay 100 --trace=nvtx,cuda,mpi --output=target_gpu_implicit_baroclinic_wave_4process/output_active/report-%q{PMI_RANK}
+            nsys profile --delay 20 --trace=nvtx,cuda,mpi --output=target_gpu_implicit_baroclinic_wave_4process/output_active/report-%q{PMI_RANK}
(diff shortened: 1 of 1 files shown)


.buildkite/pipeline.yml
```diff
@@ -885,7 +885,7 @@ steps:
       - label: "GPU: GPU dry baroclinic wave - 4 gpus"
         key: "target_gpu_implicit_baroclinic_wave_4process"
         command:
-            nsys profile --delay 100 --trace=nvtx,cuda,mpi --output=target_gpu_implicit_baroclinic_wave_4process/output_active/report-%q{PMI_RANK}
+            nsys profile --delay 20 --trace=nvtx,cuda,mpi --output=target_gpu_implicit_baroclinic_wave_4process/output_active/report-%q{PMI_RANK}
```
Sbozzolo, line 888: This nsys command is not attached to any julia command, so the job fails.

Sbozzolo:
I think that this should work, but it doesn't and I am not sure why:
https://github.com/CliMA/ClimaAtmos.jl/pull/3405/commits/a25cd3369672a7abf3f3989c8ab3a40a9d9ae1d9

.buildkite/pipeline.yml
charleskawczynski: Ah, I thought this was a multi-line command

charleskawczynski:
Not sure what's going on, but maybe this was fixed.

Sbozzolo:
We commented out nsight for that job

charleskawczynski:
Ah, right

PR #3416, opened 2024-10-31
Update configs (default topo smoothing) + add docs

akshaysridhar:
Update default topo smoothing
Add docs demonstrating topo smoothing.


config/default_configs/default_config.yml
```diff
@@ -51,7 +51,7 @@ topo_smoothing:
   value: false
 topography_damping_factor:
   help: "Factor by which smallest resolved length-scale is to be damped"
-  value: 512
+  value: 5
```
szy21, line 54: This seems like a pretty large change. What value should we use in the AMIP simulation?

config/longrun_configs/longrun_moist_held_suarez_deepatmos.yml
szy21, line 1: I don't think held suarez needs implicit diffusion. I'm fine with adding it, but we should use the same configs in the deep and shallow atmos versions.

config/default_configs/default_config.yml
akshaysridhar: The default value in the AMIP configuration initially was set up to match the outcome from the working Gaussian configuration in the coarse AMIP config. Ideally I'd like to use no more than 10-100x attenuation at the small scales if it's shown to be robust across all our configs. (The first iteration of this was perhaps overly conservative). This outcome of this PR is tested against the 16 element longrun build result for now (up to 60days). (Plots of the surface elevation for h_elem=16 are below as part of the doc assets. )

szy21: Could you also try this in the nightly amip?

config/longrun_configs/longrun_moist_held_suarez_deepatmos.yml
akshaysridhar: I'll remove it in that case.

szy21:
I find it pretty hard to read the values from the plots with the colormap - is there a more quantitative colormap we can use? If it's hard then don't worry about it.

config/default_configs/default_config.yml
akshaysridhar: https://buildkite.com/clima/climacoupler-coarse-nightly-amip/builds/136 @szy21 the corresponding build is here

akshaysridhar:
> I find it pretty hard to read the values from the plots with the colormap - is there a more quantitative colormap we can use? If it's hard then don't worry about it.

Yes I can fix these up with cleaner plots with Makie.

docs/src/topography.md
```diff
@@ -0,0 +1,12 @@
+### Topography in ClimaAtmos
+
+Dataset source : https://www.ncei.noaa.gov/products/etopo-global-relief-model
+ClimaArtifact  : https://github.com/CliMA/ClimaArtifacts/tree/main/earth_orography 
+
+We currently use the ClimaUtilities `SpaceVaryingInput` tool to regrid (using linear interpolation) the ETOPO2022 ice-surface elevation dataset (see ClimaArtifacts) onto the required spectral element horizontal grid. The file `examples/topography_spectra.jl` provides simple tools to generate such regridded fields (and their spectra) on user-defined horizontal spaces. For existing ClimaAtmos simulation data, users may access the `orog.nc` dataset from the default diagnostic outputs to visualize or examine this data. As an example, we include plots of the generated topography on a cubed sphere with 16 elements per panel edge, and compare instances of `unsmoothed` and `smoothed` datasets to visualize the effect of our topography smoothing methods.
```
szy21, line 6:
```suggestion
We currently use the ClimaUtilities `SpaceVaryingInput` tool to regrid (using linear interpolation) the ETOPO2022 ice-surface elevation dataset (see ClimaArtifacts) onto the required spectral element horizontal grid. The file `examples/topography_spectra.jl` provides simple tools to generate such regridded fields (and their spectra) on user-defined horizontal spaces. For existing ClimaAtmos simulation data, users may access the `orog.nc` dataset from the default diagnostic outputs to visualize or examine this data. As an example, we include plots of the generated topography on a cubed sphere with 16 elements and 64 elements per panel edge, and compare instances of `unsmoothed` and `smoothed` datasets to visualize the effect of our topography smoothing methods. 
```

szy21 (approved):

PR #3420, opened 2024-11-02
Improve high-order horizontal derivatives

dennisYatunin:
## Purpose 

This PR fixes a potential issue with high-order horizontal derivatives at element boundaries that was identified while writing the dycore paper. Specifically, it replaces every "strong derivative" (an ill-defined notion for our hybrid CG-DG discretization) with a "weak derivative" (defined in terms of the weak formulation and GLL quadrature rule) and a call to DSS. Hopefully this will reduce numerical oscillations at element boundaries from hyperdiffusion and the viscous sponge layer.

## To-do
Find a specific example where this leads to a behavioral change.

----


dennisYatunin:
Following a conversation with Oswald, it looks like this is not the way we want to compute horizontal derivatives. Certain derivatives in the hyperdiffusion and sponge tendencies should remain strong, as they are now. I'm still unclear on the details, but hopefully we can hash them out in one of the paper appendices.

PR #3422, opened 2024-11-04
Use deep atmosphere equations by default

akshaysridhar:
Updates config arguments to set `deep_atmosphere: true` as default. Retains one experiment in shallow atmos configuration.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,6 +5,11 @@ Main
 -------
 ### Features
 
+### Update default configuration to use deep-atmosphere eqns, fix diagnostic bug
+PR [3422](https://github.com/CliMA/ClimaAtmos.jl/pull/3422)
+Updates the `default_config` to set `deep_atmosphere=true`, and updates the 
+`rv` relative vorticity diagnostic to store the curl of horizontal velocity.
+
 ### Allow different sizes of dust and sea salt for radiation
 
 Added functionality to allow five different size bins of dust and sea salt aerosols
@@ -17,6 +22,7 @@ PR [3555](https://github.com/CliMA/ClimaAtmos.jl/pull/3555)
 
 The option `FriersonDiffusion` is removed from `vert_diff` config. Use `DecayWithHeightDiffusion` instead.
 PR [3592](https://github.com/CliMA/ClimaAtmos.jl/pull/3592)
+
 
 v0.28.4
 -------
--- a/config/default_configs/default_config.yml
+++ b/config/default_configs/default_config.yml
@@ -278,8 +278,8 @@ log_progress:
   help: "Log simulation progress, including wall-time estimates"
   value: true
 deep_atmosphere:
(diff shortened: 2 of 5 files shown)


szy21 (approved):
Let's have one shallow atmosphere example in ci (you can choose anything, e.g.  baroclinic wave).

charleskawczynski:
I think we may want to write a few more unit tests in ClimaCore to show that this is working for our finite difference stencils. In particular, I think it may be helpful to show that our FD operators are conservative. Thoughts?

szy21:
> I think we may want to write a few more unit tests in ClimaCore to show that this is working for our finite difference stencils. In particular, I think it may be helpful to show that our FD operators are conservative. Thoughts?

What unit tests do we have now that is only for shallow atmosphere?

charleskawczynski:
> What unit tests do we have now that is only for shallow atmosphere?

I don't know that we have conservation tests in a deep atmosphere

szy21:
> > What unit tests do we have now that is only for shallow atmosphere?
> 
> I don't know that we have conservation tests in a deep atmosphere

We should add them if we don't have them. Could you point me to the tests for shallow atmosphere in ClimaCore?

charleskawczynski:
> Could you point me to the tests for shallow atmosphere in ClimaCore?

That's a good question, I don't actually see any at first glance. I only some conservation tests for horizontal limiters.

We should probably add some, I think we can spin off a test from one of the column heat/wave examples without too much work.

szy21:
I'm a bit concerned about the two orders of magnitude increase in conservation error in `sphere_baroclinic_wave_conservation_source`. This test is not exactly accurate, but I don't expect such a difference between shallow and deep atmosphere. Maybe check if `sum` is implemented correctly for deep atmosphere?

szy21:
Also this one `aquaplanet_rhoe_equil_clearsky_tvinsol_0M_slabocean_ft64` has several orders of magnitude increase in the conservation error. This test should conserve within floating point error (see the current main build).

akshaysridhar:
cc @dennisYatunin

akshaysridhar:
Baroclinic wave (w/ sources) 
CC Main (Following column integral update) 
```
[ Info:     Net energy change / total energy: 6.582717e-6
[ Info:     Net mass change / total mass: -5.46637e-7
[ Info:     Net water change / total water: 2.876292e-6
```
compare with previous CA outcome 
```
[ Info:     Net energy change / total energy: 6.2523477e-6
[ Info:     Net mass change / total mass: -4.5544166e-6
[ Info:     Net water change / total water: 0.0008104961
```
 cc: @szy21

szy21:
Great, thanks @akshaysridhar !

szy21:
what's the status of this PR?

akshaysridhar:
> what's the status of this PR?

It's ready to merge (I tried queueing it yesterday but we had pipeline issues so the buildkite check was terminated. I'll rebase and requeue this soon)

szy21:
could you fix the vorticity diagnostics in this PR too so we don't need to run ci twice?

PR #3425, opened 2024-11-07
Add more reproducibility tests

### LINKED ISSUE #3392: Add more reproducibility tests to the pipeline
We should add more repro tests to the CI pipeline, including gpu tests.

charleskawczynski:
This PR adds more reproducibility tests. Closes #3392.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -144,6 +144,10 @@ steps:
           julia --color=yes --project=examples examples/hybrid/driver.jl
           --config_file $CONFIG_PATH/single_column_hydrostatic_balance_ft64.yml
           --job_id single_column_hydrostatic_balance_ft64
+
+          julia --color=yes --project=examples reproducibility_tests/test_mse.jl
+          --job_id single_column_hydrostatic_balance_ft64
+          --out_dir single_column_hydrostatic_balance_ft64/output_active
         artifact_paths: "single_column_hydrostatic_balance_ft64/output_active/*"
 
   - group: "Box Examples"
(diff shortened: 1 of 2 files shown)


trontrytel:
How did we chose which tests are opted into the reproducibility suite? And do we duplicate any? (Meaning do we run them twice - once for reproducibility and once for whatever was in the CI

charleskawczynski:
> How did we chose which tests are opted into the reproducibility suite? And do we duplicate any? (Meaning do we run them twice - once for reproducibility and once for whatever was in the CI

They opt-in by running the second `test_mse.jl` script after the initial driver job is run. By default, we use strict reproducibility, but we can set the `test_broken_report_flakiness` flag to true if it turns out a job is not very reproducible.

we probably don’t care about reproducibility for all jobs, though, like the flame graphs or benchmarks, so I think it’s fine to leave it as an opt-in feature.

charleskawczynski:
I just added it to everything that looked like an experiment.

szy21:
We probably don't need reproducibility tests for all the jobs, but if it doesn't cost much I'm fine with adding more now. Does `reproducibility_test: true` in the config do anything now?

trontrytel (approved):
Thank you!

I like the flakiness flag ;)

charleskawczynski:
> We probably don't need reproducibility tests for all the jobs, but if it doesn't cost much I'm fine with adding more now. Does `reproducibility_test: true` in the config do anything now?

Yeah, I hope it’s not too expensive. I figure we can try it out. I think that flag no longer does anything, ever since we broke off the reproduceability test into a separate script. I’ll double check to confirm, and if it doesn’t do anything, then I’ll go and remove the flag.

charleskawczynski:
Actually, I'll remove the `reproducibility_test` flag in the next PR since we'll need to update `test_broken_report_flakiness` for any jobs that are flaky anyway.

charleskawczynski:
Oh, we currently use `reproducibility_test` to indicate that the results are exported to an NC file (so that the data is available when we run `reproducibility_tests/test_mse.jl`).

PR #3427, opened 2024-11-07
Make restart test more robust

Sbozzolo:
I am not sure if this would work


Sbozzolo:
I ran this three times and it didn't fail with "path not found"

Sbozzolo:
Done in other ways

PR #3430, opened 2024-11-09
Use env for repro tests, improve repro names

charleskawczynski:
This PR:
 - Removes the `reproducibility_test` flag in the yaml config, and we instead use an environment variable. I don't think experiments should explicitly need to worry about this, and there's no way for users to leverage this feature locally anyway, so an env flag seems more appropriate.
 - Improves some names to better reflect what some reproducibility test functions are doing.


Sbozzolo:
What's the issue with the 4 GPU job?

Sbozzolo (approved):

charleskawczynski:
> What's the issue with the 4 GPU job?

Not sure yet, still needs fixes..

charleskawczynski:
I incremented the reference counter because I think it's far easier to "reset" our comparable references than to support comparing data exported from HDF5 formats (which we now use to support reproducibility tests for GPU jobs) with NC files (which we previously used).

charleskawczynski:
If the restart job gets tripped up again, then I'll manually merge.

PR #3439, opened 2024-11-15
Add docs, unit tests, and refactor repro test infra

charleskawczynski:
This PR adds docs, unit tests, and refactors the reproducibility test infrastructure. In adding unit tests and documentation, I've realized that some of the terminology around the repro tests needs revised for certain terminology to be well defined.

In particular, if we consider the situation where commits are reverted, we can end up with multiple and diverging paths (multiple paths with the same reference counter, but made at different periods in development). For example, if our reference counter history looks like (top being most recent):

```julia
137
     /PR10 # is PR6 even valid to compare against, here?
136
     /PR9
     /PR10 # should this compare with PRs 4 and 5?
135  # Reverted
     /PR8 # should this compare with commits in PRs 1,2,3?
138
     /PR7
137
     /PR6
136
     /PR4
     /PR5
135
     /PR1
     /PR2
     /PR3
```
We could clean up older commits, but I think a safer option is to change the `latest_comparable_paths` to use the (newly) added function `compute_bins`. I originally used the term `bundle` (e.g., `compute_bundles`), which I'm almost leaning towards returning to. It's a longer name but sounds more...tidy.

This was previously not an issue when `latest_comparable_paths` was `latest_comparable_path`, because we always grabbed the most _recent_ matched reference to compare against, eliminating this potential issue.

I'll update the README once we land on a design that is coherent, but I think it might take a bit more refactoring to get there.

### DIFF
--- a/reproducibility_tests/reproducibility_utils.jl
+++ b/reproducibility_tests/reproducibility_utils.jl
@@ -1,3 +1,44 @@
+#=
+################################################################################
+Reproducibility Terminology.
+
+Consider the following set of reproducibility
+folders, prefixed by "reference counters", which
+allow users to compare against other reproducible
+states in that column.
+
+Note that reference counter changes can "rewind"
+(which may happen in the case of reverted commits).
+In such cases, we do consider the rewound state as
+an entirely new state, in order to fully preserve
+the history (to some depth).
+
+An important consequence of this requires precise
+terminology to avoid ambiguous descriptions.
+
+For example, "comparable references per reference counter"
+is not well defined, because the reference counter can
+be reverted. So, let's introduce the concept of a "bin",
+which can be defined as a collection of folders
+created in a period with the same reference counter.
+Folders created before and after that bin have a different
+reference counter. Also, `n_bins == n_reference_changes + 1`
+(modulo the edge case of when there are no bins)
+because, if the reference counter doesn't change, new results
+are put into the same bin.
+```
+comparable states
+         |                           ref counter changes ---->                         | oldest
+         |                                                                             |
+         |  bin 1      bin 2      bin 3      bin 4      bin 5      bin 6      bin 7    |
+         |                                                                             |
+         |  02_49f92   04_36ebe   05_beb8a   06_4d837   05_8c311   08_45875   10_bc1e0 |
+         |             04_d6e48              06_d6d73              08_1cc58            |
+         v             04_4c042                                                        v newest
+```
+################################################################################
+=#
+
 import Dates
 
 read_ref_counter(filename) = parse(Int, first(readlines(filename)))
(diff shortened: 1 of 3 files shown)


reproducibility_tests/reproducibility_utils.jl
```diff
@@ -118,57 +159,156 @@ function latest_comparable_paths(;
...
-    f = joinpath(path, "ref_counter.jl")
-    if !isfile(f)
-        return "ref_counter.jl does not exist"
-    else
-        ref_counter = parse(Int, first(readlines(f)))
-        return "ref_counter: $ref_counter"
+"""
+    invalid_reference_folders(; root_path)
+
+Returns all subfolders in `root_path`
```
Sbozzolo, line 165:
```suggestion
Return all subfolders in `root_path`
```

reproducibility_tests/reproducibility_utils.jl
```diff
@@ -118,57 +159,156 @@ function latest_comparable_paths(;
...
-        for i in 1:N
-            path = sorted_paths[i]
-            ref_counter_file = joinpath(path, "ref_counter.jl")
-            if !isfile(ref_counter_file)
-                push!(paths_to_delete, path)
+"""
+    compute_bins(root_path::String)
+    compute_bins(sorted_paths::Vector{String})
+
+Returns a vector of reproducibility bins.
```
Sbozzolo, line 182:
```suggestion
Return a vector of reproducibility bins.
```

reproducibility_tests/reproducibility_utils.jl
```diff
@@ -118,57 +159,156 @@ function latest_comparable_paths(;
...
-        for i in 1:length(paths_to_delete)
-            f = paths_to_delete[i]
-            @info "     (File, date): ($(f), $(Dates.unix2datetime(stat(f).mtime))). Reason: $(reason(f))"
-        end
-        for i in 1:length(paths_to_delete)
-            rm(paths_to_delete[i]; recursive = true, force = true)
+        push!(bins, paths_per_bin)
+    end
+    return bins
+end
```
Sbozzolo, line 232: I think the logic of this function hard to follow. Some of the the issues are: 
- there are two nested and interacting `while true`  (both of which start with an exit condition, meaning that we can just turn them into `while condition`)
- variable names are obscure
- the intent of the various instructions is not very clear to me

I tried writing it differently, but I am not sure if I am preserving the logic:
```suggestion
function compute_bins(sorted_paths::Vector{String})
    num_paths = length(sorted_paths)
    bins = Vector{String}[]
    path_index = 1
    while path_index <= num_paths
            paths_per_bin = String[]
            while path_index <= num_paths
                first_path = sorted_paths[path_index]
                first_ref_counter = read_ref_counter(joinpath(first_path, "ref_counter.jl"))
                push!(paths_per_bin, first_path)
                path_index += 1
                current_path = sorted_paths[path_index]
                current_ref_counter = read_ref_counter(joinpath(p, "ref_counter.jl"))
                if current_ref_counter == first_ref_counter
                    push!(paths_per_bin, current_path)
                    path_index += 1
                else
                    break
                end
        end
        push!(bins, paths_per_bin)
    end
    return bins
end
```
This version also doesn't make the intent very clear

reproducibility_tests/reproducibility_utils.jl
```diff
@@ -118,57 +159,156 @@ function latest_comparable_paths(;
...
+
+"""
+    get_reference_paths_to_delete(;
+        root_path,
+        keep_n_comparable_states = 5,
+        keep_n_bins_back = 7,
+    )
+
+Returns a list of folders to delete.
```
Sbozzolo, line 241:
```suggestion
Return a list of folders to delete.
```

reproducibility_tests/reproducibility_utils.jl
```diff
@@ -118,57 +159,156 @@ function latest_comparable_paths(;
...
+
+Our reference folders are saved, and can
+therefore build up significantly.
```
Sbozzolo, line 244:
```suggestion
Our reference folders are saved on a shared folder on the cluster, and can
therefore build up significantly and take a lot of storage.
```

Sbozzolo (approved):
Thanks! I think this is a big improvement in transparency on what the reproducibility tests are doing.

I just left one comment for a function whole logic I have hard time following

reproducibility_tests/reproducibility_utils.jl
charleskawczynski: So, one reason I left this generic is for re-useability / portability. I'd like to remove, or make as many things not actually depend on central (e.g., the unit tests can now exercise this locally, which doesn't have anything to do with central).

reproducibility_tests/reproducibility_utils.jl
charleskawczynski: Hm, I'm getting BoundsError with this version, also, needing to change `current_ref_counter = read_ref_counter(joinpath(p, "ref_counter.jl"))` to `current_ref_counter = read_ref_counter(joinpath(current_path, "ref_counter.jl"))`

charleskawczynski: I can try to improve the local variable names

charleskawczynski: I've updated the names and changed the `while` conditions

reproducibility_tests/reproducibility_utils.jl
charleskawczynski: I kept the `and take a lot of storage` piece.

Sbozzolo: Okay, I just wanted to tell readers why we care about removing things.

charleskawczynski: I think we can do that where it's called in `move_output.jl`

PR #3444, opened 2024-11-20
rename ConstantArea to SmoothArea detrainment and change the formula

szy21:
## Purpose 

This is technically breaking, but this detrainment is only used for testing EDMF in atmos so it shouldn't affect anything else.

## Content

----


szy21:
cc @Julians42. You can use the SmoothAreaDetrainment for the Soares case after rebasing.

PR #3445, opened 2024-11-20
Pass subset of ClimaAtmosParameters to detrainment

charleskawczynski:
This PR passes a subset of ClimaAtmosParameters to detrainment, as we are near hitting the limit of parameter memory space when running on the gpu.

### DIFF
--- a/src/prognostic_equations/edmfx_entr_detr.jl
+++ b/src/prognostic_equations/edmfx_entr_detr.jl
@@ -2,11 +2,14 @@
 ##### EDMF entrainment detrainment
 #####
 
+import Thermodynamics.Parameters as TDP
+
 """
    Return entrainment rate [1/s].
 
    Inputs (everything defined on cell centers):
-   - params set with model parameters
+   - `thermo_params` thermodynamics parameters
+   - `turbconv_params` turbulence convection parameters
    - ᶜz, z_sfc, ᶜp, ᶜρ, - grid-scale height, surface height, grid-scale pressure and density
    - buoy_flux_surface - buoyancy flux at the surface
    - ᶜaʲ, ᶜwʲ, ᶜRHʲ, ᶜbuoyʲ - updraft area, physical vertical velocity,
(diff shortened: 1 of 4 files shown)


szy21 (approved):
Thanks! We can just replace `turbconv_params` with `tc_params` in some of the functions?

src/prognostic_equations/edmfx_entr_detr.jl
```diff
@@ -100,8 +106,8 @@ function entrainment(
     ᶜtke⁰,
     ::GeneralizedEntrainment,
 )
-    FT = eltype(params)
-    turbconv_params = CAP.turbconv_params(params)
+    FT = eltype(thermo_params)
+    turbconv_params = turbconv_params
```
szy21, line 110:
```suggestion
```

src/prognostic_equations/edmfx_entr_detr.jl
```diff
@@ -271,8 +282,8 @@ function detrainment(
     ᶜtke⁰,
     ::GeneralizedDetrainment,
 )
-    FT = eltype(params)
-    turbconv_params = CAP.turbconv_params(params)
+    FT = eltype(thermo_params)
+    turbconv_params = turbconv_params
```
szy21, line 286:
```suggestion
```

szy21:
Hmm, somehow the reproducibility test failed. I don't think this should change MSE though?

src/prognostic_equations/edmfx_entr_detr.jl
charleskawczynski: Good catch, thanks!

charleskawczynski:
> Hmm, somehow the reproducibility test failed. I don't think this should change MSE though?

Yeah, that's strange. My best guess is that it ran on a different machine? Or there are still issues with the reproducibility test infrastructure that need worked out. I'm working on adding more unit tests now.

charleskawczynski:
Please let me know if this was sufficient to get your example working (cc @szy21). If not, I can do a followup with updating the actual broadcast expression, which I think could have a bigger impact since the local geometry is passed in.

PR #3446, opened 2024-11-21
Eliminate Rayleigh sponge cache

charleskawczynski:
This PR eliminates the cache for the Raleigh sponge.

As we discussed in the meeting today, it seems like doing this generally has several benefits:

 - fewer variables that we need to worry about keeping consistent
 - fewer reads/writes/kernel launches (better performance)

I'm actually not sure that I can think of downsides, at least for this particular example.

### DIFF
--- a/src/solver/types.jl
+++ b/src/solver/types.jl
@@ -126,10 +126,12 @@ Base.@kwdef struct ViscousSponge{FT} <: AbstractSponge
 end
 
 Base.@kwdef struct RayleighSponge{FT} <: AbstractSponge
+    zmax::FT
     zd::FT
     α_uₕ::FT
     α_w::FT
 end
+Base.Broadcast.broadcastable(x::RayleighSponge) = tuple(x)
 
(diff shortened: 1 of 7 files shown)


charleskawczynski:
Also, I noticed that it looks like `test/parameterized_tendencies/sponge/viscous_sponge.jl` is basically a duplicate of `test/parameterized_tendencies/sponge/rayleigh_sponge.jl`. We should fix that (I'll leave it for now since we likely just want to change some names).

src/solver/types.jl
```diff
@@ -126,10 +126,12 @@ Base.@kwdef struct ViscousSponge{FT} <: AbstractSponge
 end
 
 Base.@kwdef struct RayleighSponge{FT} <: AbstractSponge
+    zmax::FT
```
Sbozzolo, line 129: I don't think `zmax` should be in here. `zmax` is a property of the domain, one should be able to define a `RayleighSponge` independently on the computational grid used

PR #3447, opened 2024-11-21
Eliminate ViscousSponge cache, fix unit test

charleskawczynski:
This PR eliminates the ViscousSponge cache, and fixes the ViscousSponge unit test, which was not being exercised.

This is a followup to #3446.

### DIFF
--- a/src/solver/types.jl
+++ b/src/solver/types.jl
@@ -120,18 +120,17 @@ diffuse_momentum(::FriersonDiffusion{DM}) where {DM} = DM
 diffuse_momentum(::Nothing) = false
 
 abstract type AbstractSponge end
+Base.Broadcast.broadcastable(x::AbstractSponge) = tuple(x)
 Base.@kwdef struct ViscousSponge{FT} <: AbstractSponge
     zd::FT
     κ₂::FT
 end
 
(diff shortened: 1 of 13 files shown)


src/solver/types.jl
```diff
@@ -120,7 +120,9 @@ diffuse_momentum(::FriersonDiffusion{DM}) where {DM} = DM
 diffuse_momentum(::Nothing) = false
 
 abstract type AbstractSponge end
+Base.Broadcast.broadcastable(x::AbstractSponge) = tuple(x)
 Base.@kwdef struct ViscousSponge{FT} <: AbstractSponge
+    zmax::FT
```
szy21, line 125: I have the same question that @Sbozzolo has for the Rayleigh sponge: Is it possible to move this out of the struct?

charleskawczynski: Yes, I'll update this to do that

szy21 (approved):
Thanks!

charleskawczynski:
`Spaces.z_max` wasn't added until ClimaCore 0.14.18, so I need to update the compat. I could add a check, but the alternatives would be to dig into internals, which I think is not a good idea, or use `maximum`, which is not really performant.

PR #3450, opened 2024-11-21
test that entries in default config have "help" and "value"

haakon-e:
All entries in `default_config.yml` should have both "value" and "help" entries. If not set, the added test will fail. Tests are written in a way to make it easier for developers to see which keys are missing one of the entries. e.g.:
```julia
Check that entries to `default_config.yml` has `help` and `value` keys: Test Failed at ClimaAtmos.jl/test/config.jl:62
  Expression: isempty(missing_help)
   Evaluated: isempty(["forcing"])
```

### DIFF
--- a/test/config.jl
+++ b/test/config.jl
@@ -37,3 +37,17 @@ repeated_job_ids = filter(kv -> length(kv[2]) > 1, value_to_keys)
 file, io = mktemp()
 config_err = ErrorException("File $(CA.normrelpath(file)) is empty or missing.")
 @test_throws config_err CA.AtmosConfig(file)
+
+@testset "Check that entries in `default_config.yml` have `help` and `value` keys" begin
+    config = CA.load_yaml_file(CA.default_config_file)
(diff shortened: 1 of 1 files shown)


test/config.jl
```diff
@@ -37,3 +38,19 @@ repeated_job_ids = filter(kv -> length(kv[2]) > 1, value_to_keys)
 file, io = mktemp()
 config_err = ErrorException("File $(CA.normrelpath(file)) is empty or missing.")
 @test_throws config_err CA.AtmosConfig(file)
+
+@testset "Check that entries in `default_config.yml` have `help` and `value` keys" begin
+    default_config_file =
+        joinpath("..", "config", "default_configs", "default_config.yml")
```
Sbozzolo, line 44:
```suggestion
        joinpath(pkgdir(CA), "config", "default_configs", "default_config.yml")
```

Sbozzolo (approved):

PR #3459, opened 2024-11-26
fix: logging for preset comms device

haakon-e:
Correctly set `Logging` options if `comms_ctx` is already supplied to `AtmosConfig`.

### DIFF
--- a/src/solver/type_getters.jl
+++ b/src/solver/type_getters.jl
@@ -589,18 +589,7 @@ function get_comms_context(parsed_args)
     end
     comms_ctx = ClimaComms.context(device)
     ClimaComms.init(comms_ctx)
-    if ClimaComms.iamroot(comms_ctx)
-        Logging.global_logger(Logging.ConsoleLogger(stderr, Logging.Info))
-    else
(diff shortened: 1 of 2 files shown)


src/solver/type_getters.jl
```diff
@@ -613,6 +602,15 @@ function get_comms_context(parsed_args)
...
 end
 
+function set_global_logger(comms_ctx)
+    # Set logging to only display for the root process
+    if ClimaComms.iamroot(comms_ctx)
+        Logging.global_logger(Logging.ConsoleLogger(stderr, Logging.Info))
+    else
+        Logging.global_logger(Logging.NullLogger())
+    end
+end
```
Sbozzolo, line 612:
```suggestion
function silence_non_root_processes(comms_ctx)
    # Set logging to only display for the root process
    if ClimaComms.iamroot(comms_ctx)
        Logging.global_logger(Logging.ConsoleLogger(stderr, Logging.Info))
    else
        Logging.global_logger(Logging.NullLogger())
    end
end
```
Can you also add a dosctring?

Sbozzolo (approved):
Thank you!

PR #3462, opened 2024-12-03
Move SG_quad out of cache

Sbozzolo:
There is no reason for SG_quad to be in AtmosCache: it is constant and is firmly attached to a specific function.

### DIFF
--- a/src/solver/types.jl
+++ b/src/solver/types.jl
@@ -16,48 +16,71 @@ struct NoPrecipitation <: AbstractPrecipitationModel end
 struct Microphysics0Moment <: AbstractPrecipitationModel end
 struct Microphysics1Moment <: AbstractPrecipitationModel end
 
-# TODO: Add docstrings explaining this
-abstract type AbstractQuadratureType end
-struct LogNormalQuad <: AbstractQuadratureType end
-struct GaussianQuad <: AbstractQuadratureType end
+"""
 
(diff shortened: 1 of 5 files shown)


src/solver/types.jl
```diff
@@ -16,48 +16,74 @@ struct NoPrecipitation <: AbstractPrecipitationModel end
...
     end
 end
 quadrature_order(::SGSQuadrature{N}) where {N} = N
-quad_type(::SGSQuadrature{N}) where {N} = N #TODO - this seems wrong?
 
-# TODO: Add docstrings explaining this
+"""
+    AbstractCloudModel
+
+Defines how to compute the cloud fraction
```
Sbozzolo, line 59:
```suggestion
Defines how to compute the cloud fraction.
```

trontrytel:
Seems like all the buildkite errors are because of

```
ERROR: LoadError:
--
  |  
  | Pull request author:
  | It seems that a new dataset,
  |  
  | dataset file:`diagnostic_edmfx_aquaplanet/output_0000/prog_state.nc`,
  | was created, or the name of the dataset
  | has changed. Please increment the reference
  | counter in `reproducibility_tests/ref_counter.jl`.
  |  
  | If this is not the case, then please
  | open an issue with a link pointing to this
  | PR and build.

```
not sure why though. We should not be changing anything about the file names in the output here.

PR #3465, opened 2024-12-03
Remove output_dir from cache

Sbozzolo:
Output dir is constant and is only used for a callback and the diagnostics


Sbozzolo:
https://github.com/CliMA/ClimaCoupler.jl/pull/1106 fixes the ClimaCoupler downstream test

juliasloan25 (approved):

PR #3466, opened 2024-12-04
Remove start_date from cache

Sbozzolo:
`start_date` is constant and only used to initialize diagnostics and in the computation of insolation. Here, I remove it from the cache. For insolation, I moved it to the constructor. In the future, once we have the integer time, we will be able to remove it from there as well.

### DIFF
--- a/src/callbacks/callbacks.jl
+++ b/src/callbacks/callbacks.jl
@@ -297,16 +297,17 @@ function set_insolation_variables!(Y, p, t, ::IdealizedInsolation)
     rrtmgp_model.weighted_irradiance .= weighted_irradiance
 end
 
-function set_insolation_variables!(Y, p, t, ::TimeVaryingInsolation)
+function set_insolation_variables!(Y, p, t, tvi::TimeVaryingInsolation)
     FT = Spaces.undertype(axes(Y.c))
     params = p.params
     insolation_params = CAP.insolation_params(params)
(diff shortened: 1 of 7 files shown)


nefrathenrici (approved):
Thank you, the TODOs are helpful as well.

src/callbacks/callbacks.jl
```diff
@@ -297,16 +297,17 @@ function set_insolation_variables!(Y, p, t, ::IdealizedInsolation)
...
     params = p.params
     insolation_params = CAP.insolation_params(params)
     (; insolation_tuple, rrtmgp_model) = p.radiation
 
-    current_datetime = p.start_date + Dates.Second(round(Int, t)) # current time
+    current_datetime = tvi.start_date + Dates.Second(round(Int, t)) # current time
     max_zenith_angle = FT(π) / 2 - eps(FT)
     irradiance = FT(CAP.tot_solar_irrad(params))
     au = FT(CAP.astro_unit(params))
+    # TODO: Where does this date0 come from?
```
szy21, line 310: I think date0 may be t0 in this equation: https://clima.github.io/Insolation.jl/dev/ZenithAngleEquations/

szy21 (approved):

src/callbacks/callbacks.jl
Sbozzolo: It's a todo in the sense, we shouldn't hardcode a number like this. It should come from the package

PR #3467, opened 2024-12-04
rename override_τ_precip to override_precip_timescale

szy21:
## Purpose 

A bikeshedding PR, but I don't like unicode in the config as it could be hard for people who use e.g. vim. What do others think? 

I think the coupler uses override_τ_precip in a few places. I can make a following up PR in the coupler. Should I mark this as breaking?

## Content

----


Sbozzolo (approved):

nefrathenrici (approved):
Thank you!

PR #3470, opened 2024-12-05
Add dates to diagnostics

Sbozzolo:
Removing `p.start_date` meant that the diagnostics no longer have dates by default. Here, I add it back.

Failure here: https://buildkite.com/clima/climacoupler-coarse-nightly-amip/builds/170#0193948c-b940-4532-b497-c3ce4ec2991e


ph-kev (approved):
The format check is failing, but looks good otherwise.

PR #3472, opened 2024-12-06
lower conservation error threshold

szy21:
## Purpose 

The sources and sinks are integrated correctly when using `PrognosticSurfaceTempertature`, so I removed the branch that checks for sources and sinks. We should run conservation tests with `PrognosticSurfaceTemperature`. I have to increase the threshold as previously the simulations go to the branch with `has_no_source_or_sink`, which has a much larger tolerance, due to an error in the check of `vert_diff`.

## Content

----

### DIFF
--- a/examples/hybrid/driver.jl
+++ b/examples/hybrid/driver.jl
@@ -127,19 +127,9 @@ if config.parsed_args["check_conservation"]
     @info "    Net mass change / total mass: $mass_conservation"
     @info "    Net water change / total water: $water_conservation"
 
-    sfc = p.atmos.surface_model
-
-    if CA.has_no_source_or_sink(config.parsed_args)
-        @test energy_conservation ≈ 0 atol = 50 * eps(FT)
-        @test mass_conservation ≈ 0 atol = 50 * eps(FT)
-        @test water_conservation ≈ 0 atol = 50 * eps(FT)
-    else
-        @test energy_conservation ≈ 0 atol = sqrt(eps(FT))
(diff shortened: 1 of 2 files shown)


szy21:
The longrun conservation test will probably break after this change, but I think sqrt(eps) doesn't make sense. We can consider increasing the tolerance in the future, but I would like to understand the issue first.

examples/hybrid/driver.jl
```diff
@@ -127,19 +127,9 @@ if config.parsed_args["check_conservation"]
...
-    else
-        @test energy_conservation ≈ 0 atol = sqrt(eps(FT))
-        @test mass_conservation ≈ 0 atol = sqrt(eps(FT))
-        if sfc isa CA.PrognosticSurfaceTemperature
-            @test water_conservation ≈ 0 atol = sqrt(eps(FT))
-        end
-    end
+    @test energy_conservation ≈ 0 atol = 50 * eps(FT)
+    @test mass_conservation ≈ 0 atol = 50 * eps(FT)
+    @test water_conservation ≈ 0 atol = 50 * eps(FT)
```
akshaysridhar, line 132: Removing the case dependent tolerances here seems reasonable - the `Float64` energy conservation result in buildkite is worse than the moisture / mass tests, is it clear why that is?

szy21: The incorrect branching is due to the boolean and false string in vert_diff. I removed the `has_no_source_or_sink` function as it is not used anywhere now.

akshaysridhar (approved):

PR #3475, opened 2024-12-10
Add exponentially decaying with height diffusion option

trontrytel:
Add exponentially decaying with height diffusion `DecayWithHeightDiffusion`

The `DecayWithHeightDiffusion` works with the `Coupler` (at least for the one AMIP configuration that I have been testing), so I thought it would be good to switch to using it in `Atmos` by default. But lmk if you would prefer to keep more cases with `FriersonDiffusion` cc @szy21 , @akshaysridhar 

Also, I'm not sure if what I did in `src/parameters/create_parameters.jl` is the most elegant option. I didn't want to add `H` and `D₀` on top of `C_E` parameter because we will need either the first two or the `C_E` parameter. Never the three at the same time. cc @nefrathenrici 

Also, also, would it be ok to not define any fields for the vertical diffusion `structs` in `src/solver/model_getters.jl`? And instead just grab the parameters from the `params` `structs` when needed? Not sure what is the benefit of one option vs the other? cc @nefrathenrici


src/cache/precomputed_quantities.jl
```diff
@@ -333,8 +330,11 @@ ts_sgs(thermo_params, moisture_model, specific, K, Φ, p) = thermo_state(
     p,
 )
 
-function eddy_diffusivity_coefficient(C_E, norm_v_a, z_a, p)
-    p_pbl = 85000
+function eddy_diffusivity_coefficient(D₀, H, z_sfc, z)
+   return D₀ * exp(-(z - z_sfc) / H)
+end
+function eddy_diffusivity_coefficient(C_E, norm_v_a, z_a, p, p_sfc)
```
szy21, line 336: Let's keep this vertical diffusion unchanged so it's the same as the Held Suarez literature. We should not use it for any topography case.

src/cache/precomputed_quantities.jl
```diff
@@ -544,15 +544,31 @@ NVTX.@annotate function set_precomputed_quantities!(Y, p, t)
         )
     end
 
-    if vert_diff isa VerticalDiffusion
+    if vert_diff isa DecayWithHeightDiffusion
+        (; ᶜK_h) = p.precomputed
+
+        ᶜz = Fields.coordinate_field(Y.c).z
+
+        ᶠz_sfc = p.scratch.temp_data_face_level
```
szy21, line 552: I'm not sure if you need this line?

src/parameters/create_parameters.jl
```diff
@@ -135,6 +135,26 @@ function create_parameter_set(config::AtmosConfig)
...
+    vert_diff_model = parsed_args["vert_diff"]
+    name_map_vert_diff =
+        if vert_diff_model in ("true", true, "VerticalDiffusion", "FriersonDiffusion")
+            (; :C_E => :C_E,)
+        elseif vert_diff_model in ("DecayWithHeightDiffusion",)
+            (; :H_diffusion => :H, :D_0_diffusion => :D₀)
+        else
+            nothing
+        end
+    @info(name_map_vert_diff)
```
szy21, line 147:
```suggestion
```

config/model_configs/sphere_aquaplanet_rhoe_equilmoist_allsky_gw_raw_zonallyasymmetric.yml
```diff
@@ -9,7 +9,7 @@ dt: "400secs"
 surface_setup: "DefaultMoninObukhov"
 t_end: "15hours"
 non_orographic_gravity_wave: true
-vert_diff: true
+vert_diff: "DecayWithHeightDiffusion"
```
szy21, line 12:
```suggestion
vert_diff: "FriersonDiffusion"
```
(We should keep one ci test with FriersonDiffusion to maintain the code. It doesn't matter which one it is.)

szy21 (approved):
Looks good to me, thanks! I'm not sure about the parameter thing either.

szy21:
And could you update the NEWS.md?

nefrathenrici (approved):
This looks fine to me. It's better to split off the parameters than to add unused parameters to the ClimaAtmosParameter struct.

PR #3476, opened 2024-12-10
add mass and energy conservation diagnostics

szy21:
## Purpose 

Adds diagnostics that are useful for checking and plotting conservation, and functionality to write diagnostics to text files.

## Content

----

### DIFF
--- a/docs/Manifest.toml
+++ b/docs/Manifest.toml
@@ -342,10 +342,10 @@ version = "0.14.20"
     Krylov = "ba0b0d4f-ebba-5204-a429-3ac8c609bfb7"
 
 [[deps.ClimaDiagnostics]]
-deps = ["Accessors", "ClimaComms", "ClimaCore", "Dates", "NCDatasets", "SciMLBase"]
-git-tree-sha1 = "ace174fe5e4ae04c50a7835683baecd92e49c0d4"
+deps = ["Accessors", "ClimaComms", "ClimaCore", "Dates", "NCDatasets", "OrderedCollections", "SciMLBase"]
+git-tree-sha1 = "e9ac94af815dcae2a2ab24e54b53e76cca6258b7"
 uuid = "1ecacbb8-0713-4841-9a07-eb5aa8a2d53f"
-version = "0.2.10"
+version = "0.2.11"
 
 [[deps.ClimaParams]]
 deps = ["TOML"]
--- a/src/diagnostics/conservation_diagnostics.jl
+++ b/src/diagnostics/conservation_diagnostics.jl
@@ -1,0 +1,144 @@
+# This file is included in Diagnostics.jl
+
+# Diagnostics for computing mass and energy conservation
+# All the diagnostics in this file are scalars, we need to
+# turn them into vectors as the diagnostics need to be mutable
+# Related issue: https://github.com/CliMA/ClimaDiagnostics.jl/issues/100
+
+###
+# Total mass of the air (scalar)
+###
+add_diagnostic_variable!(
+    short_name = "massa",
+    long_name = "Total Mass of the Air",
+    units = "kg",
+    compute! = (out, state, cache, time) -> begin
+        if isnothing(out)
+            return [sum(state.c.ρ)]
+        else
+            out .= [sum(state.c.ρ)]
+        end
+    end,
+)
+
+###
+# Total energy of air (scalar)
+###
+add_diagnostic_variable!(
+    short_name = "energya",
+    long_name = "Total Energy of the Air",
+    units = "J",
+    compute! = (out, state, cache, time) -> begin
+        if isnothing(out)
+            return [sum(state.c.ρe_tot)]
+        else
+            out .= [sum(state.c.ρe_tot)]
+        end
+    end,
+)
+
+###
+# Total water of the air (scalar)
+###
+compute_watera!(out, state, cache, time) =
+    compute_watera!(out, state, cache, time, cache.atmos.moisture_model)
+compute_watera!(_, _, _, _, model::T) where {T} =
+    error_diagnostic_variable("watera", model)
+
+function compute_watera!(
+    out,
+    state,
+    cache,
+    time,
+    moisture_model::T,
+) where {T <: Union{EquilMoistModel, NonEquilMoistModel}}
+    if isnothing(out)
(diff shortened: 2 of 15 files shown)


docs/Manifest.toml
```diff
@@ -1,6 +1,6 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.10.7"
+julia_version = "1.10.5"
```
szy21, line 3: I can update Julia and change it back. But on clima cluster it's still 1.10.5 so if I update dependencies there it will be changed to 1.10.5. Not sure what is the best practice?

src/solver/solve.jl
```diff
@@ -243,6 +242,10 @@ function check_conservation(sol)
     end
     energy_radiation_input = -p.net_energy_flux_toa[][]
 
+    @info energy_atmos_change,
```
akshaysridhar, line 245: This block prints as follows: 
```
[ Info: Checking conservation
[ Info: (-3.5128077f17, -0.0f0, -0.0f0, -9.468598f22)
[ Info:     Net energy change / total energy: -3.7099553e-6
[ Info:     Net mass change / total mass: 1.2692552e-6
[ Info:     Net water change / total water: 4.2512484e-6
```
The 4-tuple values are not really clear from just the logging statement so maybe we can add string descriptions (or remove this info block entirely, since `driver.jl` prints the change after the check is complete). I checked that the `energy, mass` txt files are generated at the end of a baroclinic wave test simulation.

src/diagnostics/conservation_diagnostics.jl
```diff
@@ -0,0 +1,143 @@
...
+    if isnothing(out)
+        return [(sum(state.c.ρq_tot))]
+    else
+        out .= [(sum(state.c.ρq_tot))]
+    end
+end
+
+add_diagnostic_variable!(
+    short_name = "watera",
+    long_name = "Total Energy of the Air",
```
akshaysridhar, line 63: typo : Energy > Water

src/solver/solve.jl
szy21: Oh, I forgot to remove this. I added them to check the diagnostics are correct. I'll remove them.

src/diagnostics/conservation_diagnostics.jl
szy21: Thanks!

docs/Manifest.toml
Sbozzolo: Anything is fine as long as you don't change the minor version. I think it's nice to keep what was already here (not necessarily what's on the clusters) to reduce git diff noises

src/diagnostics/conservation_diagnostics.jl
```diff
@@ -0,0 +1,143 @@
+# This file is included in Diagnostics.jl
+
+# Diagnostics for computing mass and energy conservation
+# All the diagnostics in this file are scalars, we need to
+# turn them into vectors as the diagnostics need to be mutable
```
Sbozzolo, line 5: Can you open an issue in ClimaDiagnostics and link it here?

src/diagnostics/conservation_diagnostics.jl
```diff
@@ -0,0 +1,143 @@
...
+###
+# Total mass of the air (scalar)
+###
+add_diagnostic_variable!(
+    short_name = "massa",
+    long_name = "Total Mass of the Air",
+    units = "kg",
+    compute! = (out, state, cache, time) -> begin
+        if isnothing(out)
+            return [(sum(state.c.ρ))]
```
Sbozzolo, line 16:
```suggestion
            return [sum(state.c.ρ)]
```
Same with the other ones

src/solver/solve.jl
```diff
@@ -266,3 +265,36 @@ function check_conservation(sol)
 
     return (; energy_conservation, mass_conservation, water_conservation)
 end
+
+write_diagnostics_as_txt(simulation::AtmosSimulation) =
+    write_diagnostics_as_txt(
+        simulation.output_writers[1],
+        simulation.output_dir,
+    )
+write_diagnostics_as_txt(writer, output_dir) = nothing
```
Sbozzolo, line 274:
```suggestion
function write_diagnostics_as_txt(simulation::AtmosSimulation) 
    foreach(w -> write_diagnostics_as_txt(w, simulation.output_dir), filter(w -> w isa CAD.DictWriter, simulation.output_writers)
    return nothing
end

```

src/solver/solve.jl
```diff
@@ -266,3 +265,36 @@ function check_conservation(sol)
...
+    writer::ClimaDiagnostics.Writers.DictWriter,
+    output_dir,
+)
+    @info "Writing diagnostics to text files"
+    for diagnostic in keys(writer.dict)
+        first(values(writer[diagnostic])) isa Vector ||
+            "write_diagnostics_as_txt is not supported for diagnostics that are not vectors"
+        filename = joinpath(output_dir, diagnostic * ".txt")
+        times = collect(keys(writer[diagnostic]))
+        values_all = Base.getindex.(collect(values(writer[diagnostic])), 1)
```
Sbozzolo, line 292:
```suggestion
        values_all = getindex.(collect(values(writer[diagnostic])), 1)
```

src/solver/solve.jl
```diff
@@ -266,3 +265,36 @@ function check_conservation(sol)
...
+        simulation.output_writers[1],
+        simulation.output_dir,
+    )
+write_diagnostics_as_txt(writer, output_dir) = nothing
+
+"""
+    write_diagnostics_as_txt(writer, output_dir)
+
+Write diagnostics in DictWriter to text files. This function is
+added because currently we do not support writing scalars to netcdf.
```
Sbozzolo, line 280: Let's mention the specific diagnostics that are compatible with this

Sbozzolo: Specific type of diagnostics (ie, they have to be 1-element vectors). Let's also add a link here to the issue in ClimaDiagnostics

src/diagnostics/conservation_diagnostics.jl
szy21: Adde here: https://github.com/CliMA/ClimaDiagnostics.jl/issues/100

Sbozzolo (approved):

PR #3480, opened 2024-12-12
Remove model config

Sbozzolo:
The information is already available in the space


szy21 (approved):
Thanks! Maybe we can move is_column and is_sphere to utilities? Maybe other components will use them at some point.

PR #3485, opened 2024-12-17
Add new slope-limited transport operator 

akshaysridhar:
Adds new operator that uses constrained slope limiter methods, following Lin et al. (1994) Equations (1b,1c, 2, 5). 
Updates moist baroclinic wave problem to use this limiter configuration.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,6 +9,12 @@ v0.28.1
 -------
 ### Features
 
+### Add van Leer class operator
+
+Added a new vertical transport option `vanleer_limiter` (for tracer and energy variables)
+which uses methods described in Lin et al. (1994) to apply slope-limited upwinding. Adds 
+operator 
+
 ### Read initial conditions from NetCDF files
 
 Added functionality to allow initial conditions to be overwritten by
--- a/config/default_configs/default_config.yml
+++ b/config/default_configs/default_config.yml
@@ -206,11 +206,11 @@ viscous_sponge:
   help: "Viscous sponge [`true`, `false` (default)]"
   value: false
 tracer_upwinding:
-  help: "Tracer upwinding mode [`none` (default), `first_order` , `third_order`, `boris_book`, `zalesak`]"
-  value: none
+  help: "Tracer upwinding mode [`none` (default), `first_order` , `third_order`, `boris_book`, `zalesak`, `vanleer_limiter`]"
+  value: vanleer_limiter
 energy_upwinding:
-  help: "Energy upwinding mode [`none` (default), `first_order` , `third_order`, `boris_book`, `zalesak`]"
-  value: none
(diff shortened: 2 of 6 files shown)


examples/Manifest.toml
```diff
@@ -1,6 +1,6 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.10.7"
+julia_version = "1.10.4"
```
szy21, line 3: Let's keep this the same.

config/model_configs/sphere_baroclinic_wave_rhoe_equilmoist.yml
```diff
@@ -5,6 +5,8 @@ initial_condition: "MoistBaroclinicWave"
 dt: "450secs"
 t_end: "10days"
 moist: "equil"
+tracer_upwinding: vanleer_limiter
+energy_upwinding: vanleer_limiter
```
szy21, line 9: Let's make the vanleer_limiter the default.

charleskawczynski:
@dennisYatunin, could you please look at this when you have a chance?

charleskawczynski:
Also, since this is supposed to be set as the new default, should we put a new requirement that ClimaAtmos requires the latest ClimaCore? Conditionally requiring this seems a bit confusing, since the default would also need to be different, and that's set in the toml file. Thoughts, @Sbozzolo, @dennisYatunin?

Sbozzolo:
One simple way to preserve compatibility with prior versions could be to
change the default in the yaml file, and if we detect that the climacore
version is below 0.14.21 in type_getter we emit a warning saying that the
limiter is not available and disable it. This would prompt users to upgrade
their climacore version, while also allowing people to run with old
climacore versions if they need to.

szy21:
I can take a look tonight or tomorrow morning if Dennis hasn’t already.

dennisYatunin (approved):
Code changes look good to me! If @szy21 thinks the CI results are correct then this should be fine to merge in. I'm okay with bumping the required ClimaCore version since this has to be the new default, but Gabrielle's suggestion also works.

Have the longruns been checked against this branch yet? If not, we should probably do that before committing to the new default.

szy21:
For some reason, the summary plots are not there for the jobs that break reproducibility tests (and it seems those tests [break](https://buildkite.com/clima/climaatmos-ci/builds/22005) in the main branch too). I looked at some other tests and they look fine. Akshay tested this branch some time ago with the dycore longruns, so I think it would be ok.

szy21:
@charleskawczynski I'll bump the ref counter to get some plots and see if the reproducibility test error disappears.

szy21:
Alright, the ci results look good to me. The ClimaCoupler downstream tests break because of the ClimaCore version I think.

szy21 (approved):

charleskawczynski:
Rebasing...

PR #3492, opened 2024-12-23
Move aerosols to its own type

Sbozzolo:
Aerosols were handles by parsing directly the YAML arguments. This led to implicit assumptions, and required aerosols to be handled differently. For example, we had to pass the list `aerosol_names` to `build_cache` even when no aerosol was required.

This commit moves aerosol to its own type so that we can follow a more composable approach. It is still not fully possible to swap in a custom, user-defined type because the portion of code in RRTGMPInterface makes assumption on the names of the aerosols. I add error checking for that


charleskawczynski (commented):
Can we make aerosols model a property of the radiation model, instead of the atmos model?

Sbozzolo:
> Can we make aerosols model a property of the radiation model, instead of the atmos model?

I discussed this in the past with @szy21. We decided that it makes more sense to keep tracers separate because at some point in the future they are not going to be tied to the radiation model.

Sbozzolo:
I forgot that symbols are not isbits, so this is currently not GPU compatible. I will work on it.

PR #3493, opened 2024-12-23
Debug repro tests

charleskawczynski:
This PR wraps the reproducibility tests in a `@testset`, so that we can see the number of tests that pass / fail when running.


charleskawczynski:
I think the repro tests are a bit too quiet, I think we're supposed to see some tables printed.

PR #3509, opened 2025-01-03
Add `override_precipitation_timescale` option back in

nefrathenrici:
This PR reverts part of #3481, which removed the `override_precipitation_timescale` config argument. 
This was not caught by our regression tests, but the the PR build shows changed plots.

This PR also adds a small unit test to ensure that this option is being set correctly.

### DIFF
--- a/src/parameters/create_parameters.jl
+++ b/src/parameters/create_parameters.jl
@@ -7,15 +7,6 @@ import Thermodynamics.Parameters.ThermodynamicsParameters
 import CloudMicrophysics as CM
 import StaticArrays as SA
 
-function ClimaAtmosParameters(config::AtmosConfig)
-    (; toml_dict, parsed_args) = config
-    FT = CP.float_type(toml_dict)
-    return ClimaAtmosParameters(
-        toml_dict,
-        FT(CA.time_to_seconds(parsed_args["dt"])),
-    )
-end
-
 """
     ClimaAtmosParameters(FT, dt)
     ClimaAtmosParameters(toml_dict, dt)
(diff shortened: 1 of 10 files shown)


charleskawczynski:
Results for which job changed? Was it one that we had reproduceability tests for?

nefrathenrici:
I believe the regression tests were unaffected because the tests don't run for long enough to produce precipitation, but @szy21 noticed this issue and changed plots in the PR#3481 build

szy21:
> Results for which job changed? Was it one that we had reproduceability tests for?

I think it didn't show up in the reproducibility test as the job designed to catch this is too short to produce precipitation, so the PR didn't change the MSE of that job. There are some other jobs of which the results are changing, but we don't have reproducibility tests for them.

src/parameters/create_parameters.jl
```diff
@@ -10,10 +10,11 @@ import StaticArrays as SA
...
-    return ClimaAtmosParameters(
-        toml_dict,
-        FT(CA.time_to_seconds(parsed_args["dt"])),
-    )
+    override_dt =
+        parsed_args["override_precip_timescale"] ?
+        FT(CA.time_to_seconds(parsed_args["dt"])) : nothing
+
+    return ClimaAtmosParameters(toml_dict, override_dt)
 end
```
szy21, line 18: Should we move the function after the doc string?

szy21 (approved):
Thank you!

src/parameters/create_parameters.jl
nefrathenrici: yes, thank you!

PR #3516, opened 2025-01-07
set bubble to false

szy21:
## Purpose 

Our current implementation of the bubble correction is not internally consistent, as it modifies the volume element J without modifying the metric tensor g, so that we no longer satisfy the discrete identity J = sqrt(det(g)). Our current implementation also does not account for volume distortions due to topography, as it is only applied to the horizontal component of the volume element. In order to fix this (and be more similar to HOMME's implementation), we would need to apply the bubble correction to the full 3D metric Jacobian, and then compute J and g from the corrected Jacobian.

However, as Paul Ullrich pointed out, modifying g on interior nodal points without modifying it on boundary nodal points will have the consequence of changing the discrete curvature. I think this is basically the point of the bubble correction (trading accuracy in curvature for accuracy in volume integrals), but Tapio is not a fan of this tradeoff. Our current plan is to eliminate the bubble correction from ClimaCore altogether, and instead handle volume integral discrepancies during remapping, which will allow us to avoid changes in curvature terms.

We might end up revisiting the bubble correction later if this strategy does not work well, but we will still need to implement it in a completely different way to fix the current issues. So, there is no point in keeping the current implementation.

## Content

----

### DIFF
--- a/config/default_configs/default_config.yml
+++ b/config/default_configs/default_config.yml
@@ -150,7 +150,7 @@ smagorinsky_lilly:
   value: false
 bubble:
   help: "Enable bubble correction for more accurate surface areas"
-  value: true
+  value: false
 start_date:
   help: "Start date of the simulation"
   value: "20100101"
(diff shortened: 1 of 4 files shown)


charleskawczynski:
Fine by me. Why did we decide that?

szy21:
> Fine by me. Why did we decide that?

@dennisYatunin ?

PR #3517, opened 2025-01-10
Make single column example run reproducibility test

imreddyTeja:
## Purpose 
Create a reproducibility test for a single column run. This will be used to test the results of changing to a pure single column.

## Content
Adds reproducibility test to the "single column hydrostatic balance float64" job

----

### DIFF
--- a/reproducibility_tests/ref_counter.jl
+++ b/reproducibility_tests/ref_counter.jl
@@ -1,4 +1,4 @@
-196
+197
 
 # **README**
 #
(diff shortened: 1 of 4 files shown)


charleskawczynski (commented):
Looks good, I think we just need to
 - Add `single_column_hydrostatic_balance_ft64` to the `reproducibility_tests/reproducibility_test_job_ids.jl` list, and
 - increment the ref counter

reproducibility_tests/ref_counter.jl
```diff
@@ -20,6 +20,9 @@
 
 
 #=
+197
```
charleskawczynski, line 23: The uncommented number above must be incremented

charleskawczynski (approved):

PR #3521, opened 2025-01-14
WIP on updating velocities for dycore paper

trontrytel:
I'll start a longrun moist baro job to see when it actually crashes. But in the mean time, could you double check if things make sense? @szy21 @dennisYatunin

### DIFF
--- a/src/cache/precomputed_quantities.jl
+++ b/src/cache/precomputed_quantities.jl
@@ -48,6 +48,8 @@ function precomputed_quantities(Y, atmos)
         ᶜspecific = specific_gs.(Y.c),
         ᶜu = similar(Y.c, C123{FT}),
         ᶠu³ = similar(Y.f, CT3{FT}),
+        ᶜwₜqₜ = similar(Y.c, Geometry.WVector{FT}),
+        ᶜwₕhₜ = similar(Y.c, Geometry.WVector{FT}),
         ᶜK = similar(Y.c, FT),
         ᶜts = similar(Y.c, TST),
         ᶜp = similar(Y.c, FT),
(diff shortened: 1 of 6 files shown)


src/cache/precomputed_quantities.jl
```diff
@@ -517,9 +526,31 @@ NVTX.@annotate function set_precomputed_quantities!(Y, p, t)
     #     (; ᶜmixing_length) = p.precomputed
     #     compute_gm_mixing_length!(ᶜmixing_length, Y, p)
     # end
-
+    @. p.precomputed.ᶜwₜqₜ = Geometry.WVector(0)
+    @. p.precomputed.ᶜwₕhₜ = Geometry.WVector(0)
```
szy21, line 530: maybe unpack the ᶜwₜqₜ and ᶜwₕhₜ so we can get rid of several p.precomputed?

szy21: (And we can just set them both to zero for this PR for testing.)

szy21 (approved):
Looks good, thanks! I may take a closer look when I have some more time but you don't need to wait for me to merge it.

src/cache/precomputed_quantities.jl
trontrytel: Done! Set to zero and the CI test are running

szy21 (approved):
I don't know why it should cause behavior changes, but the code change looks good to me.

trontrytel:
Thank you for taking a look. We did add terms to the Jacobian that were not there before. I don't think they are important, but maybe different enough to show up on the mse? Or there is a bug somewhere because I can't seem to reproduce those changes when running on my local machine.

I bumped the ref counter so that I can merge it in and continue working on actually updating the RHS

szy21:
> Thank you for taking a look. We did add terms to the Jacobian that were not there before. I don't think they are important, but maybe different enough to show up on the mse? Or there is a bug somewhere because I can't seem to reproduce those changes when running on my local machine.
> 
I think all the terms we added to the Jacobian are zero now (or -I for diagonal), and adding zeros to Jacobian shouldn't change the result. But maybe I missed something. It's ok, let's merge it for now, and debug it if there is instabilities in the longrun.

PR #3522, opened 2025-01-16
Read CO2 from file

Sbozzolo:
Currently, when running AMIP, ClimaCoupler is responsible for updating the concentration of CO2 used for the radiation transport. This means that:
1. ClimaAtmos by itself cannot use observed CO2 concentrations
2. ClimaCoupler has to maintain specialized code just to update these values

The reason for why this was done is because originally only ClimaCoupler had the infrastructure to read in files. However, this is no longer the case and ClimaAtmos is already updating values read from files for aerosols and ozone.
This PR adds CO2 to the list.

### DIFF
--- a/src/callbacks/callbacks.jl
+++ b/src/callbacks/callbacks.jl
@@ -59,6 +59,13 @@ function update_o3!(p, t, ::PrescribedOzone)
     return nothing
 end
 
+update_co2!(_, _, _) = nothing
+function update_co2!(p, t, ::MaunaLoaCO2)
+    evaluate!(p.tracers.co2, p.tracers.prescribed_co2_timevaryinginput, t)
+    return nothing
+end
+
+
 NVTX.@annotate function rrtmgp_model_callback!(integrator)
     Y = integrator.u
     p = integrator.p
@@ -71,6 +78,7 @@ NVTX.@annotate function rrtmgp_model_callback!(integrator)
 
     # If we have prescribed ozone or aerosols, we need to update them
     update_o3!(p, t, p.atmos.ozone)
+    update_co2!(p, t, p.atmos.co2)
     if :prescribed_aerosols_field in propertynames(p.tracers)
         for (key, tv) in pairs(p.tracers.prescribed_aerosol_timevaryinginputs)
             field = getproperty(p.tracers.prescribed_aerosols_field, key)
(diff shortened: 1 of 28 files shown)


src/solver/type_getters.jl
```diff
@@ -57,6 +57,7 @@ function get_atmos(config::AtmosConfig, params)
     atmos = AtmosModel(;
         moisture_model,
         ozone,
+        co2 = PrescribedCO2(),
```
szy21, line 60: Does it mean we always prescribe CO2? Could we add an option to set co2 as a constant? You can either use `input_vmr("carbon_monoxide_GM")`, or hard code the value 397.547.

szy21 (approved):
Thanks!

src/callbacks/callbacks.jl
```diff
@@ -59,6 +59,13 @@ function update_o3!(p, t, ::PrescribedOzone)
     return nothing
 end
 
+update_co2!(_, _, _) = nothing
+function update_co2!(p, t, ::FixedCO2)
```
szy21, line 63:
```suggestion
function update_co2!(p, t, ::MuanaLoaCO2)
```

src/solver/model_getters.jl
```diff
@@ -308,6 +308,18 @@ function get_ozone(parsed_args)
     return parsed_args["prescribe_ozone"] ? PrescribedOzone() : IdealizedOzone()
 end
 
+function get_co2(parsed_args)
+    if isnothing(parsed_args["co2_model"])
+        return nothing
+    elseif lowercase(parsed_args["co2_mode"]) == "fixed"
```
szy21, line 314:
```suggestion
    elseif lowercase(parsed_args["co2_model"]) == "fixed"
```

src/solver/model_getters.jl
```diff
@@ -308,6 +308,18 @@ function get_ozone(parsed_args)
...
+        return FixedCO2()
+    elseif lowercase(parsed_args["co2_mode"]) == "maunaloa"
```
szy21, line 316:
```suggestion
    elseif lowercase(parsed_args["co2_model"]) == "maunaloa"
```

szy21 (approved):

PR #3523, opened 2025-01-16
Fix conversion from dt_rad to dt_rad_ms

ph-kev:
When converting from seconds to milliseconds, we need to multiply by 1_000.


Sbozzolo (approved):
Thank you!

PR #3525, opened 2025-01-17
Update ClimaTimeSteppers to latest version

dennisYatunin:
## Purpose 

This PR applies the DSS correction in CliMA/ClimaTimeSteppers.jl#348 to ClimaAtmos.

----


dennisYatunin:
Results seem to be unchanged, but maybe it would be good for @szy21 to take a look as well. I would expect a small improvement in conservation, but it could be too small to notice outside of the analytic mountain test.

calibration/experiments/sphere_held_suarez_rhoe_equilmoist/Manifest.toml
szy21, line 1: Why is this file deleted?

szy21 (commented):
Looks good, and it looks like there is some improvement in mass conservation.

calibration/experiments/sphere_held_suarez_rhoe_equilmoist/Manifest.toml
dennisYatunin: It's not deleted, it's just a large diff because it hasn't been updated since Julia 1.10.4.

szy21: Oh I see. Maybe @nefrathenrici can check if it is ok to update this.

nefrathenrici: This manifest should not be updated, it is used to reproduce the calibration experiment

dennisYatunin: Okay, I won't modify the two calibration manifests.

PR #3532, opened 2025-01-21
Integrate ClimaComms logging updates

nefrathenrici:
This PR tests out the latest logging features from ClimaComms, especially https://github.com/CliMA/ClimaComms.jl/pull/104.

All simulations will now have their output logs saved to disk in `output_active/output.log`
MPI simulations will have an additional folder `output_active/logs` storing the logs from each MPI process.


charleskawczynski (commented):
We'll need to upgrade the manifest, too. Otherwise we won't see the expected changes in CI.

PR #3536, opened 2025-01-22
Make precipitation part of the working fluid

trontrytel:


src/parameterized_tendencies/microphysics/precipitation.jl
```diff
@@ -213,15 +234,14 @@ function compute_precipitation_cache!(Y, p, ::Microphysics1Moment, _)
...
+        ᶜSqₗᵖ,
+        ᶜSqᵢᵖ,
         ᶜSqᵣᵖ,
         ᶜSqₛᵖ,
-        ᶜSeₜᵖ,
         Y.c.ρ,
-        ᶜqᵣ,
-        ᶜqₛ,
+        q_rai,
+        q_sno,
```
szy21, line 243: Just note that these could be negative.

szy21 (commented):
I only looked at changes to grid-mean now and they look good, thanks! Could you double check it still makes sense when we use TD.relative_humidity and TD.vol_vapor_mixing_ratio in the code? They should be mostly in radiation callback and diagnostics.

src/cache/cloud_fraction.jl
```diff
@@ -48,13 +56,24 @@ NVTX.@annotate function set_cloud_fraction!(
...
-        ),
-    )
+        )
+    else
+        @. cloud_diagnostics_tuple = make_named_tuple(
+            ifelse(
+                p.precomputed.ᶜspecific.q_liq + p.precomputed.ᶜspecific.q_ice > 0,
+                1,
+                0,
+            ),
```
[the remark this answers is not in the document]
trontrytel: We do. However if would return true for a case where hypothetically `q_liq=0` but `q_rai!=0`. I wanted to avoid that problem, so I do it by hand with the specific precomputed values

src/parameterized_tendencies/microphysics/precipitation.jl
trontrytel: That should be fine. I'll double check just in case but all microphysics functions return zero if one of the tracers is negative. And all the limiters inside ClimaAtmos precipitation are written anticipating tracers can be negative too. No trust!

Issue #2632, opened 2024-02-06
Performance roadmap


charleskawczynski:
This issue is a continuation of #635, but I'm excluding the items we've addressed, and adding more relevant updates.

## Key items to tackle

 - [ ] Improve RRTMGP performance (https://github.com/CliMA/RRTMGP.jl/issues/389 seems important)
 - [ ] GC pauses can happen at different places on different processes, which will hamper scaling efficiency (as processes will wait for other processes running GC): once we've reduced allocations we can disable the GC and trigger it manually intermittently. `GC.enable(false)` and `GC.gc()` to trigger manually
 - [ ] We still communicate whole elements instead of just boundaries: this will require a bit of work in ClimaCore, but we'd like to see hard numbers before going down this route. @sriharshakandala, have we resolved this bullet?
 - [ ] Use `*` over `/` if not being optimized **low impact, very little effort**
 - [ ] Measure performance of new diagnostics. We've added a callbacks flame graph, which shows time spent during diagnostics, but we should use more realistic parameters, like resolution and frequency of different diagnostics.
 - [x] Use sparse mat-mul operations for AxisTensor conversions. See PR https://github.com/CliMA/ClimaCore.jl/pull/853. We need to make sure that, on a similar note, that loading the local geometry isn't having a huge impact on our kernels. **Medium to high impact, large effort**
 - [ ] Add capability to fuse operations. Fusing operations would allow us to increase the arithmetic intensity of our kernels, enabling us to reach higher peak performance (as we are likely bandwidth bound). **Moderate to high impact, moderate to large effort**
 - [ ] Specialize on GPU performance of basic FieldVector broadcast expressions. I recently saw the timestepper portion of a Nsight systems report and realized each fieldvector broadcast expression is launching at least two kernels (cell center and cell face). Often these are just copy-to, or simple add/sub/mul/div against Real, and we can likely specialize to launch a single, perhaps more optimized kernel than what we're currently generating. I don't think we have any benchmarks for FieldVector BC expressions. This is actually reasonably low hanging fruit in that everything can happen in ClimaCore, and we may even be able to fix the performance issues we see in Krylov.jl. **Moderate impact, small to moderate effort**

## Performance notes / specs

 - I've excluded the items for: time stepper LU factorization, impact of loops over mesh components. I explained why in #635.

Nsight report for p100 Dyamond simulation:

[...]

PR #3540, opened 2025-01-24
Use LazyBroadcast for tendencies

charleskawczynski:
This PR adds the use of LazyBroadcast.jl to fuse the sponge operations. This is hopefully the beginning of a long road of fusing many operations, and moving the codebase into a more functional, and more unit-tested codebase.

### DIFF
--- a/reproducibility_tests/ref_counter.jl
+++ b/reproducibility_tests/ref_counter.jl
@@ -1,4 +1,4 @@
-208
+209
 
 # **README**
 #
@@ -20,6 +20,10 @@
 
 
 #=
+209
+- Floating point changes, with minor adjustments to how to sponge tendencies are
+  added.
+
 208
 - Update RRTMGP to v0.19.2, which changes cloud optics slightly
 
--- a/src/ClimaAtmos.jl
+++ b/src/ClimaAtmos.jl
@@ -4,6 +4,7 @@ using NVTX
 import LazyBroadcast
 import Thermodynamics as TD
 
+include("null_broadcasted.jl")
 include("compat.jl")
 include(joinpath("parameters", "Parameters.jl"))
 import .Parameters as CAP
@@ -112,6 +113,7 @@ include(
 )
 include(joinpath("parameterized_tendencies", "sponge", "rayleigh_sponge.jl"))
 include(joinpath("parameterized_tendencies", "sponge", "viscous_sponge.jl"))
+include(joinpath("parameterized_tendencies", "sponge", "sponge_tendencies.jl"))
 include(
     joinpath(
         "parameterized_tendencies",
--- a/src/null_broadcasted.jl
+++ b/src/null_broadcasted.jl
@@ -1,0 +1,100 @@
+# TODO: use https://github.com/CliMA/NullBroadcasts.jl when released
+
+"""
+    NullBroadcasted()
+
+A `Base.AbstractBroadcasted` that represents arithmetic object.
+
+An `NullBroadcasted()` can be added to, subtracted from, or multiplied by any
+value in a broadcast expression without incurring a runtime performance
+penalty.
+
+For example, the following rules hold when broadcasting instances of
+`NullBroadcasted`:
+
+```
+1 + NullBroadcasted() == 1
+NullBroadcasted() + 1 == 1
+1 - NullBroadcasted() == 1
+1 * NullBroadcasted() == NullBroadcasted()
+1 / NullBroadcasted() == NullBroadcasted()
+ ```
+"""
+struct NullBroadcasted <: Base.AbstractBroadcasted end
+Base.broadcastable(x::NullBroadcasted) = x
+
+struct NullBroadcastedStyle <: Base.BroadcastStyle end
+Base.BroadcastStyle(::Type{<:NullBroadcasted}) = NullBroadcasted()
+
+# Specialize on AbstractArrayStyle to avoid ambiguities with AbstractBroadcasted.
+Base.BroadcastStyle(::NullBroadcasted, ::Base.Broadcast.AbstractArrayStyle) =
+    NullBroadcasted()
+Base.BroadcastStyle(::Base.Broadcast.AbstractArrayStyle, ::NullBroadcasted) =
+    NullBroadcasted()
+
+# Add another method to avoid ambiguity between the previous two.
+Base.BroadcastStyle(::NullBroadcasted, ::NullBroadcasted) = NullBroadcasted()
+
+broadcasted_sum(args) =
+    if isempty(args)
+        NullBroadcasted()
+    elseif length(args) == 1
+        args[1]
+    else
+        Base.broadcasted(+, args...)
+    end
+Base.broadcasted(::NullBroadcasted, ::typeof(+), args...) =
+    broadcasted_sum(filter(arg -> !(arg isa NullBroadcasted), args))
+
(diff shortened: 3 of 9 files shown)


Sbozzolo:
I'd prefer we didn't introduce more technical concepts in science packages without proper documentation and written guidence on what they do and how to use them, especially if this becomes more and more common.

The goal should be that people developing this package should be able to understand and use this feature without input from experts.

Examples of questions: 
- Can I just call `@lazy` in front of any broadcasted expression and everything will work? 
- What benefits should I see?
- How does this enable kernel fusion? What's the difference with `ClimaCore.@fused`?

One way to check that you provided enough guidance is with user stories and asking people if they can use and understand this feature without your input.

Example of a user story:

I am a reserach scientist and I am debugging a piece of ClimaAtmos. I notice that the return objects of some the functions I call are some `Broadcast.Brocasted` objects that I don't know how to manipulate. This is not explained anywhere within the ClimaAtmos repo. I dig deeper and see that it might be connected to `@lazy`. I come to this conclusion because I am familiar with the other operations and `@lazy` is the only thing I have never seen. Following the imports, I end up in the `LazyBroadcast.jl` package. The only documentation in that pacakge is a readme that is very technical for me (e.g., I've never heard of "materializing"; it's not even explained in the Julia manual). The link to the Julia manual doesn't really help me: it mosty talks about broadcasting as in using the dot operator. The readme doesn't even mention the `@lazy` macro. It's unclear to me how the three points that are mentioned in the readme relate to my problem. The only examples provided are tests and don't explain what's happening. At this point, I need to ask for help, if I haven't done it already.


It's up to you to where you add documentation and guidance. For example, you might decide that LazyBrodcast is intended for a technical audience. That would be fine, but then you have to provide documentation within ClimaAtmos (or ClimaCore and links to the relevenat documentation in ClimaAtmos).


I think that this is very important. Many Clima packages, including ClimaAtmos and ClimaCore, require a lot of tribal knowledge to be used and developed. This is a big barrier to adoption, developlment, and suistanability in general. I think we should really strive to make sure that all our new contributions are propertly documented, while also catching up with the missing documentation.

charleskawczynski:
> I'd prefer we didn't introduce more technical concepts in science packages without proper documentation and written guidence on what they do and how to use them, especially if this becomes more and more common.

We can add documentation. We can also add better unit tests with this approach, which I've also updated.

> The goal should be that people developing this package should be able to understand and use this feature without input from experts.

I agree, but users will need to gain an idea of how to use things, and we can document how things work

> Examples of questions:
> 
> * Can I just call `@lazy` in front of any broadcasted expression and everything will work?

The new functional-based tendencies are dispatched in a binary way: one returns a broadcasted object, via `@lazy`, and the other returns a single point value that should be valid when the tendency is not to be added. Alternatively, we could use a design where we return a `Null` object, and filter the broadcast expressions. I'm open to either design.

> * What benefits should I see?

This has several benefits:
 - We can actually write unit tests, and more granular unit tests
 - We can [compose ClimaCore operators](https://github.com/CliMA/ClimaCore.jl/issues/2155), and there is no other way (that I know of)
 - We can fuse operations. I've spoken with several experts in the field about improving performance, and it is clear that large-scale fusion of kernels has the largest potential performance benefit. We don't know how much because it's difficult to measure, but the next optimization is at most 35%, based on [Amdahl's law](https://en.wikipedia.org/wiki/Amdahl%27s_law). I outlined the performance items in the performance [roadmap](https://github.com/CliMA/ClimaAtmos.jl/issues/2632), if you'd like to read more.
 - We might even be able to do lazy-caching for expensive quantities (I'm not yet sure what ingredients are needed for the compiler to perform common-sub-expression elimination, which may be useful for this approach).

> * How does this enable kernel fusion? What's the difference with `ClimaCore.@fused`?

This is "manual" fusion. We're manually collecting broadcasted objects, and putting them into a single `@.` call (which calls `materialize`, which performs the execution of the expression). I'm starting to think there is one big benefit of using LazyBroadcast over MultiBroadcastFusion: we can leverage dispatch and composition better with LazyBroadcast (again, see [this issue](https://github.com/CliMA/ClimaCore.jl/issues/2155)). Also, LazyBroadcast.jl's implementation is extremely simple, so it's easy for users to dig into the details if they want to.

> One way to check that you provided enough guidance is with user stories and asking people if they can use and understand this feature without your input.
>
> Example of a user story:
> 
> I am a reserach scientist and I am debugging a piece of ClimaAtmos. I notice that the return objects of some the functions I call are some `Broadcast.Brocasted` objects that I don't know how to manipulate. This is not explained anywhere within the ClimaAtmos repo. I dig deeper and see that it might be connected to `@lazy`. I come to this conclusion because I am familiar with the other operations and `@lazy` is the only thing I have never seen. Following the imports, I end up in the `LazyBroadcast.jl` package. The only documentation in that pacakge is a readme that is very technical for me (e.g., I've never heard of "materializing"; it's not even explained in the Julia manual). The link to the Julia manual doesn't really help me: it mosty talks about broadcasting as in using the dot operator. The readme doesn't even mention the `@lazy` macro. It's unclear to me how the three points that are mentioned in the readme relate to my problem. The only examples provided are tests and don't explain what's happening. At this point, I need to ask for help, if I haven't done it already.

I hear you. And I need help with this, as it's a big challenge. Please hear me: we should follow the most promising software design, and provide users with excellent documentation. I don't want to avoid the best path forward because we don't yet have documentation. I need your help writing it.

> It's up to you to where you add documentation and guidance. For example, you might decide that LazyBrodcast is intended for a technical audience. That would be fine, but then you have to provide documentation within ClimaAtmos (or ClimaCore and links to the relevenat documentation in ClimaAtmos).

Yes. Right now, I'm doing a sort-of [tracer-bullet](https://builtin.com/software-engineering-perspectives/what-are-tracer-bullets) PR, to see what this might look like.

> I think that this is very important. Many Clima packages, including ClimaAtmos and ClimaCore, require a lot of tribal knowledge to be used and developed. This is a big barrier to adoption, developlment, and suistanability in general. I think we should really strive to make sure that all our new contributions are propertly documented, while also catching up with the missing documentation.

Agreed.

Sbozzolo:
If this is the path we are taking, let's just take the time to document this now. This is part of the delivery of the feature. The questions I listed in the previous message were questions that users ought to be able to find the answer for somewhere, so your answers can be the starting point for this. Even if the final form of this will look different, the documentation written can be still adapted and reused.
If this is an experiment, let's not merge it into main until we figure things out. 

We will never recover from the missing documentation if we keep adding more missing documentation, especially when documentation is never a priority.

charleskawczynski:
Agreed, I think we should start by beefing up documentation at the bottom, and work our way up as things concretize. I just opened https://github.com/CliMA/LazyBroadcast.jl/pull/16.

charleskawczynski:
To avoid this PR from getting stale, I may add LazyBroadcast as a dependency in a separate PR, so that I don't need to continually rebase the manifest changes.

charleskawczynski:
I'm thinking that we can start by using LazyBroadcast, and not changing the order of tendency computations. This has two benefits:

 - We can microbenchmark fused vs unfused versions, in a separate and dedicated effort. I think that this is probably best since heuristics like register pressure can result in performance degradation-- it's not trivial to see that fusing two expressions will be faster or not.
 - We still benefit from being able to write unit tests.

I'll update the PR.

charleskawczynski:
I think we should address some edge cases first. In particular, I don't want any performance penalty in the case that the functional tendencies are absent.

PR #3545, opened 2025-01-27
Try modified version of ClimaTimeSteppers 0.8.0

dennisYatunin:
## Purpose 

This PR tries out a slight modification of ClimaTimeSteppers version 0.8.0, where the implicit tendency approximation is computed after DSS and the enforcement of the velocity boundary condition. Both the current version of ClimaTimeSteppers and this modified version seem equally valid to me, but maybe one of them will be more stable.

----


dennisYatunin:
Testing GPU longruns here: https://buildkite.com/clima/climaatmos-gpulongruns/builds/481

dennisYatunin:
This change leads to a slight stability improvement, and it now available in ClimaTimeSteppers v0.8.1.

PR #3546, opened 2025-01-27
Remove .dev

Sbozzolo:
This commit removes the `.dev` folder in order to simplify the repository. Issues with `.dev`:
1. It's undocumented, one has to know it's there
2. It brings in an additional julia environment that needs to be maintained
3. It sees no development and does not benefit from upstream developlment (e.g., `PkgDevTools` provides a superior way to update Manifests)
4. It is not aligned with what the rest of the julia community does

In this commit, I remove `.dev`, add deprecation notes so that users are informed about the change, and add documentation


charleskawczynski (commented):
Looks good to me, so long the formatter preserves our current rules

Sbozzolo:
> Looks good to me, so long the formatter preserves our current rules

I changed one of the rules. The current rules include `remove_extra_newlines = false`, and I removed it. Since the new format job passed, it means that this rule was already satisfied.

charleskawczynski:
There was a green light on this, despite the latest build failing (due to a depot issue), so I'm going to manually merge.

PR #3547, opened 2025-01-27
Consolidate manifests

Sbozzolo:
Manifests.toml are a burden to maintain and add a lot of git diff noise. The one benefit of Manifests.toml is that they help with tracking version dependencies.

However:
- For `docs`, the precise version of dependencies is not important, there is pretty much no version-dependent code being executed there
- `perf` and `examples` can be consolidated into a single `.buildkite` environment. This environment only has to have one manifest checked, the one for which buildkite is run (and not one for each version of Julia).
- The `perf` environment can be removed all together. It is a more niche environment: if reproducing a specific result is required, developers can use the `.buildkite` Manifest. If not, developers can also consider keeping the relevant tools (mainly JET) in the base environment.

This is the same pattern we have in other repos such as ClimaCore.

(Builds off #3546)

Closes #3548


charleskawczynski (approved):
I'm sorry about the merge conflict, I thought I had merged this before #3551, but I was mistaken.

This looks good to me, feel free to merge after rebasing, and thanks for updating this!

charleskawczynski:
I went ahead and rebased!

Sbozzolo:
Includes #3548

config/perf_configs/bm_perf_target.yml
```diff
@@ -1,16 +1,17 @@
 moist: "equil"
 apply_limiter: true
-z_elem: 25
+z_elem: 11
```
charleskawczynski, line 3: Why is this changing?

szy21: I think it's because this job failed: https://buildkite.com/clima/climaatmos-ci/builds/22496#0194af81-d116-4343-92f6-9b753e5763aa, which is likely related to the climatimestepper update.

You changed the wrong config file though I think?

charleskawczynski: Ah, ok. I think it might be better to just increase the mem request

charleskawczynski: Using low resolution is not really meaningful for performance monitoring jobs, so I think it'd be better to remove it instead of lowering the resolution.

szy21: That's fine with me. I prefer to make it soft fail though so we will remember to come back to fix this.

Sbozzolo: > Ah, ok. I think it might be better to just increase the mem request

The memory is GPU memory, so we would need to request a different type of GPU, which would increase the queue time significantly.

Sbozzolo: I changed it to a soft fail

charleskawczynski: Sounds good. I think there is something wrong, though. The fieldvector is being allocated here, and it's definitely not that large, so I don't know why we're OOMing. Maybe the GPU is not freeing memory as it should be.

szy21:
It passed everything in the merge queue, but somehow the format check didn't start.

Sbozzolo:
> It passed everything in the merge queue, but somehow the format check didn't start.

I have an idea of what it might be. Let me start by merging this manually

juliasloan25:
@Sbozzolo why is there a Manifest in the .buildkite/ directory but no Project?

charleskawczynski:
> @Sbozzolo why is there a Manifest in the .buildkite/ directory but no Project?

It was probably deleted by accident. We do need one becuase:
 - We'll need compats
 - users running `instantiate`s will add `.buildkite/Project.toml`

I'll open a PR.

PR #3548, opened 2025-01-27
tag v0.28.3, up deps

juliasloan25:
## Purpose 
Release so ClimaCoupler can use #3522 in https://github.com/CliMA/ClimaCoupler.jl/pull/1143

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "ClimaAtmos"
 uuid = "b2c96348-7fb7-4fe0-8da9-78d88439e717"
 authors = ["Climate Modeling Alliance"]
-version = "0.28.2"
+version = "0.28.3"
 
 [deps]
 ArgParse = "c7e460c6-2fb9-53a9-8c5b-16f535851c63"
@@ -42,10 +42,10 @@ ArtifactWrappers = "0.2"
 Artifacts = "1"
 AtmosphericProfilesLibrary = "0.1.7"
 ClimaComms = "0.6.4"
-ClimaCore = "0.14.19"
+ClimaCore = "0.14.23"
 ClimaDiagnostics = "0.2.4"
 ClimaParams = "0.10.17"
-ClimaTimeSteppers = "0.7.39"
+ClimaTimeSteppers = "0.7.40"
 ClimaUtilities = "0.1.14"
 CloudMicrophysics = "0.22.3"
 Dates = "1"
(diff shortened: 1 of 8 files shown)


Project.toml
```diff
@@ -44,7 +44,7 @@ ClimaComms = "0.6.4"
 ClimaCore = "0.14.19"
 ClimaDiagnostics = "0.2.4"
 ClimaParams = "0.10.17"
-ClimaTimeSteppers = "0.7.39"
+ClimaTimeSteppers = "=0.7.38"
```
Sbozzolo, line 47:
```suggestion
ClimaTimeSteppers = "<0.7.38"
```

Sbozzolo: If 0.7.39 is known to have a critical bug, the proper thing to do is release 0.7.40 with the changes reverted

juliasloan25: v0.8.0 has already been released - can we still go back and release 0.7.40?

Sbozzolo: Yes, we can create a branch in climatimesteppers and release in that branch.

If we don't release it and know that 0.7.39 has a critical bug, every package (or prior versions of ClimaAtmos) with compat 0.7 will always get a buggy version. If the bug is not super serious, we can skip this

juliasloan25: @dennisYatunin could you revert the change and release CTS v0.7.40?

dennisYatunin: The bug is definitely serious, but also is there an issue with just proceeding using CTS 0.8.0? I'm about to release 0.8.1 but 0.8.0 is ready to merge in

dennisYatunin: We need ClimaAtmos on 0.8.1 by the end of today so I'm not sure if it's useful to make a new 0.7.40

dennisYatunin (approved):

dennisYatunin:
I think you need to bump the ref counter for the MSE changes in the `diagnostic_edmfx_aquaplanet` job, and also increase the memory allocation for the `bm_prog_edmf_gpu` job (not sure why that one is necessary but it's probably not a big issue).

Sbozzolo:
> I think you need to bump the ref counter for the MSE changes in the `diagnostic_edmfx_aquaplanet` job, and also increase the memory allocation for the `bm_prog_edmf_gpu` job (not sure why that one is necessary but it's probably not a big issue).

The memory is GPU memory, so we need to request other types of GPUs or make the problem smaller. I am trying with the latter

PR #3549, opened 2025-01-27
Remove ᶜspecific from the cache

ph-kev:
This PR removes ᶜspecific from the cache.

### DIFF
--- a/src/cache/diagnostic_edmf_precomputed_quantities.jl
+++ b/src/cache/diagnostic_edmf_precomputed_quantities.jl
@@ -94,7 +94,6 @@ NVTX.@annotate function set_diagnostic_edmf_precomputed_quantities_bottom_bc!(
     n = n_mass_flux_subdomains(turbconv_model)
     (; ᶜΦ) = p.core
     (; ᶜp, ᶠu³, ᶜh_tot, ᶜK) = p.precomputed
-    (; q_tot) = p.precomputed.ᶜspecific
     (; ustar, obukhov_length, buoyancy_flux, ρ_flux_h_tot, ρ_flux_q_tot) =
         p.precomputed.sfc_conditions
     (; ᶜρaʲs, ᶠu³ʲs, ᶜKʲs, ᶜmseʲs, ᶜq_totʲs, ᶜtsʲs, ᶜρʲs) = p.precomputed
@@ -300,7 +299,6 @@ NVTX.@annotate function set_diagnostic_edmf_precomputed_quantities_do_integral!(
     (; dt) = p
     (; ᶜΦ) = p.core
     (; ᶜp, ᶠu³, ᶜts, ᶜh_tot, ᶜK) = p.precomputed
-    (; q_tot) = p.precomputed.ᶜspecific
     (;
         ᶜρaʲs,
         ᶠu³ʲs,
@@ -915,7 +913,6 @@ NVTX.@annotate function set_diagnostic_edmf_precomputed_quantities_env_closures!
     (; params) = p
     (; dt) = p
     (; ᶜp, ᶜu, ᶜts) = p.precomputed
-    (; q_tot) = p.precomputed.ᶜspecific
     (; ustar, obukhov_length) = p.precomputed.sfc_conditions
     (; ᶜρaʲs, ᶠu³ʲs, ᶜdetrʲs) = p.precomputed
     (; ᶜtke⁰, ᶠu³⁰, ᶜu⁰) = p.precomputed
@@ -1035,7 +1032,6 @@ NVTX.@annotate function set_diagnostic_edmf_precomputed_quantities_env_precipita
     microphys_0m_params = CAP.microphysics_0m_params(p.params)
     (; dt) = p
     (; ᶜts, ᶜSqₜᵖ⁰) = p.precomputed
-    (; q_tot) = p.precomputed.ᶜspecific
 
     # Environment precipitation sources (to be applied to grid mean)
     @. ᶜSqₜᵖ⁰ = q_tot_0M_precipitation_sources(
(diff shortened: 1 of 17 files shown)


src/cache/diagnostic_edmf_precomputed_quantities.jl
```diff
@@ -109,7 +108,9 @@ NVTX.@annotate function set_diagnostic_edmf_precomputed_quantities_bottom_bc!(
     u³_int_halflevel = Fields.field_values(Fields.level(ᶠu³, half))
     h_tot_int_level = Fields.field_values(Fields.level(ᶜh_tot, 1))
     K_int_level = Fields.field_values(Fields.level(ᶜK, 1))
-    q_tot_int_level = Fields.field_values(Fields.level(q_tot, 1))
+    @. p.scratch.ᶜtemp_scalar_3 = Y.c.ρq_tot / Y.c.ρ
+    q_tot_int_level =
+        Fields.field_values(Fields.level(p.scratch.ᶜtemp_scalar_3, 1))
```
charleskawczynski, line 113: I think there's a bigger question here: why do we need an extra cache variable for this at all? We already have `ᶜts`, which contains this variable (`q_tot`)

ph-kev: In this part of the code, it look like `q_tot` only exists if the moisture model is `EquilMoistModel` or `NonEquilMoistModel`. For this function specifically, it looks like there is no config that call this function which uses a `DryModel`. 

https://github.com/CliMA/ClimaAtmos.jl/blob/878d0def789b3248f40d339bbe5fde84d62265ed/src/cache/precomputed_quantities.jl#L313-L324

ph-kev:
Also, the reproducibility tests are failing. Should I expect this?

For this [buildkite](https://buildkite.com/clima/climaatmos-ci/builds/22418), it still fail the reproducibility tests despite the commit looking fine to me.

src/cache/diagnostic_edmf_precomputed_quantities.jl
```diff
@@ -336,13 +336,15 @@ NVTX.@annotate function set_diagnostic_edmf_precomputed_quantities_do_integral!(
         Fields.field_values(Fields.level(Fields.coordinate_field(Y.f).z, half))
 
     # integral
+    @. p.scratch.ᶜtemp_scalar_3 = Y.c.ρq_tot / Y.c.ρ
```
Sbozzolo, line 339: By the way, you can improve readability by defining
```
q_tot = p.scratch.ᶜtemp_scalar_3
```
so that expressions below will have `q_tot` instead of `p.scratch.ᶜtemp_scalar_3`

charleskawczynski:
> Also, the reproducibility tests are failing. Should I expect this?
> 
> For this [buildkite](https://buildkite.com/clima/climaatmos-ci/builds/22418), it still fail the reproducibility tests despite the commit looking fine to me.

 No, I think the repro tests should be passing. This is supposed to be a refactoring change, correct?

if so, I think you do need to make sure that you’ve rebased on the latest main.

charleskawczynski:
Actually, looking at the code changes, it's possible that we would expect to see changes due to machine precision error-- computing variables on the fly multiple times could result in changes.

ph-kev:
> Actually, looking at the code changes, it's possible that we would expect to see changes due to machine precision error-- computing variables on the fly multiple times could result in changes.

I found the reason why the reproducibility tests are failing. I changed this line to this and it passed now.

https://github.com/CliMA/ClimaAtmos.jl/blob/d3c0de17421ef513e1fe0bff0c8d27a17d4f66e4/src/prognostic_equations/hyperdiffusion.jl#L77

src/cache/diagnostic_edmf_precomputed_quantities.jl
ph-kev: I am not sure if I am calling that function right, but when I tried to use it, I am getting this error:
```
infil> TD.total_specific_humidity(thermo_params, p.precomputed.ᶜts)
ERROR: MethodError: no method matching total_specific_humidity(::Thermodynamics.Parameters.ThermodynamicsParameters{…}, ::ClimaCore.Fields.Field{…})
The function `total_specific_humidity` exists, but no method is defined for this combination of argument types.
```

charleskawczynski: You'd need to broadcast for it to work on the whole field: `TD.total_specific_humidity.(thermo_params, p.precomputed.ᶜts).`. But my reservation is about whether we actually need a cached variable for that field to begin with. Let's schedule some time to chat about this.

src/parameterized_tendencies/radiation/radiation.jl
```diff
@@ -393,12 +393,14 @@ function radiation_tendency!(Yₜ, Y, p, t, radiation_mode::RadiationDYCOMS)
...
     q_tot_isoline = FT(0.008)
+    @. p.scratch.ᶜtemp_scalar_3 = Y.c.ρq_tot / Y.c.ρ
+    q_tot = p.scratch.ᶜtemp_scalar_3
     Operators.column_reduce!(
         (nt1, nt2) ->
             abs(nt1.q_tot - q_tot_isoline) < abs(nt2.q_tot - q_tot_isoline) ?
             nt1 : nt2,
         isoline_z_ρ_q,
-        Base.broadcasted(NT ∘ tuple, ᶜz, Y.c.ρ, ᶜspecific.q_tot),
+        Base.broadcasted(NT ∘ tuple, ᶜz, Y.c.ρ, q_tot),
```
charleskawczynski, line 403:
```suggestion
        Base.broadcasted(NT ∘ tuple, ᶜz, Y.c.ρ, bc_q_tot),
```

PR #3555, opened 2025-01-29
add different sizes for dust and sea salt

szy21:
## Purpose 

Allows five size bins of dust and sea salt which matches RRTMGP after v0.19.2. Also supports RRTMGP before v0.19.2.
This will be merged first to fix downstream test in https://github.com/CliMA/RRTMGP.jl/pull/564. ci build with this rrtmgp branch: https://buildkite.com/clima/climaatmos-ci/builds/22742

## Content

----


szy21:
@Sbozzolo Could you take a look and see if there is an easy way to support both rrtmgp versions? I'm happy to change it if that's the case.

Sbozzolo (approved):

szy21:
This passed buildkite, and didn't change the ref counter, so I'm going to manually merge it in.

PR #3558, opened 2025-01-29
Skip buildkite if we only have markdown or github

Sbozzolo:
If we only change markdown files or files in the .github folder, skip buildkite


Sbozzolo:
Closing as implemented in a different way

PR #3559, opened 2025-01-30
Add simple precompilation example, split out CI driver

nefrathenrici:
This PR introduces a local package`PrecompileCI` that precompiles common methods in ClimaAtmos. For now, it only precompiles spaces for different spatial configurations, but this can be expanded to precompile more methods. This all happens in `.buildkite/PrecompileCI/src/PrecompileCI.jl`, and the package is then loaded by the driver to load in the precompiled methods.

This PR also adds a CI driver (`.buildkite/ci_driver.jl`) with the same functionality as the current driver.
The current driver `examples/hybrid/driver.jl` now just runs and plots. It also does not load `PrecompileCI`, so it can still be used in existing workflows.

To use, just import it before ClimaAtmos: 
`julia --project=.buildkite`, then
```julia
using PrecompileCI  # Loads the precompiled methods or precompiles if loaded for the first time
import ClimaAtmos as CA
config = CA.AtmosConfig()
sim = CA.get_simulation(config)
CA.solve_atmos!(sim)
```

I included this as a separate package because precompiling these methods when ClimaAtmos is loaded will add to the TTFX.

To see which methods are actually getting precompiled, see this [gist](https://gist.github.com/nefrathenrici/0d5d7b9799ec6052d02ab201cb444b69) or run:
`julia --project=.buildkite/PrecompileCI`
```julia
using PrecompileTools
PrecompileTools.verbose[] = true
include(".buildkite/PrecompileCI/src/PrecompileCI.jl")
```


.dev/climaformat.jl
```diff
@@ -3,7 +3,7 @@
 println("climaformat.jl has been discontinued in favor of JuliaFormatter")
 println("To use JuliaFormatter, add it to your base environment with: ")
 println()
-println("julia -e 'using Pkg. Pkg.add(\"JuliaFormatter\")'")
+println("julia -e 'using Pkg; Pkg.add(\"JuliaFormatter\")'")
```
charleskawczynski, line 6:
```suggestion
println("julia -e 'using Pkg; Pkg.add(\"JuliaFormatter\")'")
println()
println("If you already have it in your local environment, you can use")
println("julia -e 'using JuliaFormatter; format(\".\")")
```

charleskawczynski (approved):
I think it's worth trying this out, maybe we can loop over more configs? But I think it's a good starting point.

Sbozzolo:
With this PR, users using the driver will need to have `PrecompileCI` in their environments. This will break people's workflow.

Let's make sure we don't break people's workflow. We could take the opportunity to split the driver into `ci_driver.jl` (which contains `PrecompileCI`) and other `driver` that for the time being could just be running + plotting.

nefrathenrici:
> With this PR, users using the driver will need to have `PrecompileCI` in their environments. This will break people's workflow.
> 
> Let's make sure we don't break people's workflow. We could take the opportunity to split the driver into `ci_driver.jl` (which contains `PrecompileCI`) and other `driver` that for the time being could just be running + plotting.

This only breaks the case of including the driver without using the `.buildkite` Project. I think splitting out the driver is a larger task which would also break people's workflow and should be separate from this PR. As a temporary workaround I could add a check `get(ENV, "CI", false) && import PrecompileCI` to ensure that this is only loaded in actual CI runs.

charleskawczynski:
Did we measure what the impact of this is?

nefrathenrici:
I can follow-up and show the change in compilation time in a PR next week

PR #3557, opened 2025-01-29
Ci changes needed for the upcoming working fluid PR

trontrytel:
This is a peel off from #3536

It:
- removes support for `equilibrium moisture` + `1-moment microphysics` and `nonequilibrium` + `0-moment`
- makes the corresponding changes in the CI pipeline

PR #3560, opened 2025-01-30
tag v0.28.4

juliasloan25:
## Purpose 
patch release so we can use the updated CTS compat downstream in ClimaCoupler

### DIFF
--- a/.buildkite/Manifest-v1.11.toml
+++ b/.buildkite/Manifest-v1.11.toml
@@ -236,10 +236,10 @@ uuid = "0b7ba130-8d10-5ba8-a3d6-c5182647fed9"
 version = "1.21.6+2"
 
 [[deps.Bzip2_jll]]
-deps = ["Artifacts", "JLLWrappers", "Libdl", "Pkg"]
-git-tree-sha1 = "8873e196c2eb87962a2048b3b8e08946535864a1"
+deps = ["Artifacts", "JLLWrappers", "Libdl"]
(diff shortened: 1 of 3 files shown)


.buildkite/Manifest-v1.11.toml
szy21, line 1: Most of this file is removed - is it intended?

juliasloan25: no, I'm not sure what happened

juliasloan25: it's because there's no Project in the .bulidkite directory, maybe the one from examples/ was meant to be moved

juliasloan25: it will be fixed in https://github.com/CliMA/ClimaAtmos.jl/pull/3561

charleskawczynski: Yeah, we should wait to fix that before tagging/releasing a new version

NEWS.md
```diff
@@ -14,6 +14,9 @@ See the  [documentation](https://clima.github.io/ClimaAtmos.jl/dev/contributor_g
 
 ### Features
 
+v0.28.4
+-------
```
szy21, line 18: Could you add something like this for #3557?

> Only support equilibrium moisture + 0-moment microphysics and nonequilibrium + 1-moment microphysics

szy21 (approved):

PR #3561, opened 2025-01-30
Add buildkite project

charleskawczynski:
This PR adds the buildkite Project.toml ([xref](https://github.com/CliMA/ClimaAtmos.jl/pull/3547#issuecomment-2623615794), [xref](https://github.com/CliMA/ClimaAtmos.jl/pull/3547#issuecomment-2624654512)).

I actually did this programmatically, and doing so allowed me to delete some direct dependencies:

```julia
function julia_files(root)
	all_files = String[]
  for (root, dirs, files) in walkdir(root)
      for file in files
          endswith(file, ".jl") && push!(all_files, joinpath(root, file))
      end
  end
  return all_files
end

function find_unecessary_deps(root, deps)
  found = Dict{String,Bool}()
  for dep in deps
    found[dep] = false
  end
  jl_files = julia_files(root)
  N = length(jl_files)
  @show N
  for (i, file) in enumerate(jl_files)
    percent = i / N
  	jl_contents = readlines(file)
  	for dep in deps
  		found[dep] = found[dep] || occursin(dep, join(jl_contents, "\n"))
  	end
  end
  unfound = String[]
  for k in keys(found)
  	if !found[k]
  		push!(unfound, k)
  	end
  end
  return unfound
end

contents = readlines(".buildkite/Project.toml");
_deps = deleteat!(map(x->first(split(x, " = ")), contents), 1);
compat_start = findfirst(x->x=="", _deps)
if !isnothing(compat_start)
  _deps = _deps[1:compat_start-1]
end
@show length(_deps)
ud = find_unecessary_deps(".", _deps)
isempty(ud) || (using Pkg; Pkg.rm(ud))
```

This unfortunately won't remove packages that are mentioned in julia file doc strings, but I think it was still helpful


Sbozzolo:
Thank you! I forgot to commit it. 

Some depedencies are surprising. For example, how are we depending on `Automa`?

Also, some of the compats also seem wrong. For example, I don't think we are compatible with SciMLBase 1 or julia 1.7

juliasloan25:
Thank you! `examples/Project.toml` also still exists. Since this is meant to be replaced by the .buildkite environment, it seems like it should be removed

Sbozzolo:
> Thank you! `examples/Project.toml` also still exists. Since this is meant to be replaced by the .buildkite environment, it seems like it should be removed

I left it because people are used to using it, and there is no fundamental problem with having a Project.toml. You can still use it to instantiate your environment locally.

charleskawczynski:
> Automa

No idea, this came along when I instantiated the `.buildkite` environment (without a `.buildkite/Project.toml`), which pulls _all_ dependencies in. That's why I used that script to try and slim them back down to only the direct dependencies, but it's not air-tight.

Sbozzolo:
> > Automa
> 
> No idea, this came along when I instantiated the `.buildkite` environment (without a `.buildkite/Project.toml`), which pulls _all_ dependencies in. That's why I used that script to try and slim them back down to only the direct dependencies, but it's not air-tight.

I was comparing with the Project.toml I have locally I think that there's something wrong with how you got the dependencies. I see a lot of packages that can be removed (but my Project.toml also has package that can be removed). I'll merge the two lists

Sbozzolo:
Oh, I see it. It's the examples/Project.toml that has a lot of stuff that is not needed.

charleskawczynski:
I took a closer look, `Automa` is an indirect dependency. So this PR looks fine as is.

charleskawczynski:
#3563 is failing. Let's tackle these issues incrementally. This PR fixes one. If there are more, let's open issues and tackle the remaining separately.

PR #3566, opened 2025-01-31
update RRTMGP

szy21:
## Purpose 

And removes `use_pade_cloud_optics_mode`

## Content

----


szy21:
The result looks identical to me, but the MSE changes. I think the cloud optics update in RRTMGP results in some changes at floating point precision. I bumped the ref counter.

PR #3567, opened 2025-01-31
Working fluid part 2 - RHS updates

trontrytel:
Trying out the updtaed RHS in the CI

### DIFF
--- a/config/longrun_configs/longrun_moist_baroclinic_wave.yml
+++ b/config/longrun_configs/longrun_moist_baroclinic_wave.yml
@@ -10,4 +10,4 @@ dt_save_state_to_disk: "10days"
 toml: [toml/longrun_baroclinic_wave.toml]
 diagnostics:
   - short_name: [pfull, wa, va, ua, ta, rhoa, rv, hus, ke, clw, cli]
-    period: 1days
+    period: 10days
--- a/src/parameterized_tendencies/microphysics/microphysics_wrappers.jl
+++ b/src/parameterized_tendencies/microphysics/microphysics_wrappers.jl
@@ -7,20 +7,29 @@ import CloudMicrophysics.Microphysics2M as CM2
 import CloudMicrophysics.MicrophysicsNonEq as CMNe
 import CloudMicrophysics.Parameters as CMP
 
-# define some aliases and functions to make the code more readable
-const Iₗ = TD.internal_energy_liquid
-const Iᵢ = TD.internal_energy_ice
-const Lf = TD.latent_heat_fusion
+# Define some aliases and functions to make the code more readable
 const Tₐ = TD.air_temperature
 const PP = TD.PhasePartition
 const qᵥ = TD.vapor_specific_humidity
 qₜ(thp, ts) = TD.PhasePartition(thp, ts).tot
-qₗ(thp, ts) = TD.PhasePartition(thp, ts).liq
-qᵢ(thp, ts) = TD.PhasePartition(thp, ts).ice
-cᵥₗ(thp) = TD.Parameters.cv_l(thp)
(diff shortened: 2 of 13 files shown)


trontrytel:
@szy21 - this is the microphysics 1M RHS update to the working fluid equations. LMK if you would prefer to merge it and work on the advection piece next (making 1M microphysics slightly inconsistent for a little bit) or if we should wait and work on advection on a separate branch.

Also, I can clean up the cloud condensation by overwriting the parameter instead. I'll do it if we decide to actually try to merge it in

szy21 (commented):
I'm fine either way, so it's up to you if you want to merge it. But if you do, maybe keep 0M in the baroclinic wave so other people can use it for the dycore paper.

trontrytel:
Thanks! I'll go ahead clean up and merge it in. And then do the same for the velocity. It''ll make it easier for testing to not have to juggle between branches so much

config/longrun_configs/longrun_moist_baroclinic_wave.yml
```diff
@@ -10,4 +10,4 @@ dt_save_state_to_disk: "10days"
 toml: [toml/longrun_baroclinic_wave.toml]
 diagnostics:
   - short_name: [pfull, wa, va, ua, ta, rhoa, rv, hus, ke, clw, cli]
-    period: 1days
+    period: 10days
```
szy21, line 13: is this change intended?

src/parameterized_tendencies/microphysics/microphysics_wrappers.jl
```diff
@@ -117,105 +126,96 @@ Also returns the total energy source term due to the microphysics processes.
...
     # https://github.com/CliMA/ClimaCore.jl/issues/1786
-    @. Sqₜᵖ = ρ * FT(0)
+    @. Sqₗᵖ = ρ * FT(0)
+    @. Sqᵢᵖ = ρ * FT(0)
     @. Sqᵣᵖ = ρ * FT(0)
     @. Sqₛᵖ = ρ * FT(0)
-    @. Seₜᵖ = ρ * FT(0)
+
+    @. qᵣ = max(FT(0), qᵣ)
+    @. qₛ = max(FT(0), qₛ)
```
szy21, line 150: I think you will change the input to this function (e.g. c_specific.q_rai) if you do this - is this what you want?

src/parameterized_tendencies/microphysics/microphysics_wrappers.jl
```diff
@@ -292,32 +252,30 @@ Also returns the total energy source term due to the microphysics processes.
...
     mp,
     thp,
 )
-    FT = eltype(Sqₜᵖ)
+    FT = eltype(thp)
     sps = (mp.ps, mp.tv.snow, mp.aps, thp)
     rps = (mp.pr, mp.tv.rain, mp.aps, thp)
 
+    @. qᵣ = max(FT(0), qᵣ)
+    @. qₛ = max(FT(0), qₛ)
```
szy21, line 270: Same as above

szy21 (approved):
Looks good, thanks!

config/longrun_configs/longrun_moist_baroclinic_wave.yml
trontrytel: I thought it would be good to output that less frequently, given that the simulation is intended to run for 120 days in the long run test

src/parameterized_tendencies/microphysics/microphysics_wrappers.jl
trontrytel: Good point. I'll re-write it so that I don't overwrite the `specific` fields

PR #3568, opened 2025-01-31
Make geopotential a broadcasted object

charleskawczynski:
This is an alternative to #3455.


charleskawczynski:
I think this is actually not a great idea, it'll only make compilation worse.

PR #3569, opened 2025-01-31
Use lazy-broadcast for vertical transport terms

charleskawczynski:
This PR applies `@lazy` to some terms, which really helps us simplify dispatch.


nefrathenrici (approved):
This looks much cleaner to me

PR #3580, opened 2025-02-02
Use LazyBroadcast for edmf coriolis

charleskawczynski:
This PR adds the use of LazyBroadcast for the edmf coriolis force.


charleskawczynski:
Buildkite already passed CI: https://buildkite.com/clima/climaatmos-ci/builds/22676

PR #3583, opened 2025-02-03
increase slurm_time for several jobs in longrun

szy21:
## Purpose 

These jobs failed due to walltime limit.

## Content

----


szy21:
Merging it in manually as it only changes longrun pipeline.

PR #3586, opened 2025-02-03
Update moist baro wave config, [ci skip]

trontrytel:
The final config for equilibrium 0M moist baroclinic wave.

The model time step is 90s and the precipitation timescale is 120s


trontrytel:
Would it be ok to merge without running CI since it only modifies a longrun job config? @Sbozzolo @szy21

szy21 (approved):
We also need `deep_atmosphere: true`, but @akshaysridhar is changing it in a different PR I think.

szy21:
> Would it be ok to merge without running CI since it only modifies a longrun job config? @Sbozzolo @szy21

Yes, I think that's fine.

PR #3552, opened 2025-01-28
Modify velocity, kinetic energy, and curl terms to match dycore paper

dennisYatunin:
## Purpose 

This PR modifies the reconstructions of cell face velocity, cell center kinetic energy, and the curl terms in the velocity tendencies to exactly match the dycore paper.

----
- [x] I have read and checked the items on the review checklist.

PR #3593, opened 2025-02-05
Add tests and plots for steady-state velocity

dennisYatunin:
## Purpose 

This PR adds tests for the steady-state velocity in box setups with analytically-defined topography, as well as plots of errors and level/column averages. Together with #3552, this supercedes #3182.

The changes to the driver are:
- Add a `"check_steady_state"` flag to the config, which is `false` by default
- When the flag is true and the topography has a well-defined Fourier transform, compute the predicted steady-state velocity by integrating the Fourier transform of the frist-order velocity perturbation across all wavenumbers
  - The steady-state velocity on centers and faces is stored in the `AtmosCache`
- Generalize the topography elevation functions so that their parameters can be reused to compute analytic solutions
- Add two new types of horizontally periodic topography (2D and 3D cosine hills)
- Add a `ConstantBuoyancyFrequencyProfile` initial condition, which is currently required to compute analytic solutions
- Include the face velocity field `ᶠu` in the precomputed quantities, and use this value in comparisons against analytic solutions
  - Update the diagnostics to save the vertical component of velocity using `ᶠu` instead of `ᶜu`, in order to accurately capture behavior at the boundaries
- Add plots that show slices of velocity components as they approach steady-state, plots of the errors relative to analytic solutions, and plots of row-wise/column-wise RMS averages of the errors
  - Tests that involve a single mountain (Schar or Agnesi) also have plots that are zoomed in on the mountain
- Add an `@test` for simulations that run for longer than 1 day and can reach steady-state
  - For vertical velocity, check that the error is less than 0.05%
  - For horizontal velocity, check that the error is less than 0.5% for Float64 and 5% for Float32 (we should figure out why horizontal velocity diverges with Float32 at some point)

The changes to CI are:
- The Schar mountain test case is added back to our CI with the `ConstantBuoyancyFrequencyProfile` initial condition and the discretization outlined in the dycore paper
  - It is also split into separate Float32 and Float64 tests
- The Agnesi mountain test case is updated to use the same initial condition and discretization as the Schar mountain tests cases, with Float64 precision
- To test the topographic forcing for a specific wavenumber, I am also adding 3 tests with cosine topography
  - One test is 2D, another is extruded 2D, and the third is full 3D
- To test general issues with our steady-state solution, I am also adding a test with no topography
  - The gradual divergence of horizontal velocity observed in this test seems to be greatly exacerbated by the presence of topography and the use of Float32 precision
- All of these tests are run on GPUs because they have very small grid spacing
  - The analytic solutions for the Schar and Agnesi mountains also take a very long time to compute on CPUs

I'm planning to eventually remove the cosine topography tests, and instead use 3D and extruded 2D versions of the Schar mountain. However, these will take a while to run, so I will add them to the longruns in a future PR.

----


szy21 (commented):
I didn't read the steady_state_velocity related changes but the other code changes look good to me.

PR #3595, opened 2025-02-05
change rrtmgp version constraint in restart test

szy21:
## Purpose 

To fix downstream test in rrtmgp

## Content

----


szy21:
merging it in manually because it doesn't affect anything now

PR #3604, opened 2025-02-06
Lower slurm requirements for calibration tests

nefrathenrici:
This PR lowers the slurm resource requirements for the calibration test, we could enable the reservation for it now.

The job itself now takes ~20min instead of ~15.


Sbozzolo (approved):
Do you want to incorprate this back to normal CI, or keep it separate?

nefrathenrici:
Let's keep it separate for now, it is nice to have this run in parallel and simplify the environments for the main CI

PR #3613, opened 2025-02-08
Use lazy broadcast function instead of macro

charleskawczynski:
Suggested by @dennisYatunin, this PR switches to using `LazyBroadcast`'s `lazy` function, instead of the `@lazy` macro.

The main benefit is that functions are much simpler to understand than macros, which are capable of doing syntactic transformations and more.


dennisYatunin (commented):
Could you please use `@. lazy(` instead of `lazy.(@. `? I think it's a little more readable

charleskawczynski:
> Could you please use `@. lazy(` instead of `lazy.(@. `? I think it's a little more readable

Ah, sure!

charleskawczynski:
Done cc @dennisYatunin

dennisYatunin (approved):
Looks good now, thank you!

charleskawczynski:
The failure is unrelated (looks like it was a depot issue).

PR #3620, opened 2025-02-10
Prescribe effective radius, add to diagnostics output

trontrytel:
This PR:
- adds effective radius to radiative diagnostics. It is set to zero outside of clouds, so it only really makes sense to look at instantaneous output. This is meant to help with debugging. If we need a monthly mean output we should weigh it by the cloud fraction (?)
- allows for effective radius to be set by CloudMicrophysics.jl diagnostics module. For now I am setting it to constant 14um for liquid and 25 um for ice. I'l add the variable functions next, once I set up the amip simulation pipeline and take a look at the biases.

It needs the most recent CloudMicrophysics and ClimaParams to work


src/callbacks/callbacks.jl
```diff
@@ -203,8 +207,18 @@ NVTX.@annotate function rrtmgp_model_callback!(integrator)
                 max(cloud_fraction, eps(FT))
             @. ᶜfrac = cloud_fraction
             # RRTMGP needs effective radius in microns
-            @. ᶜreliq = FT(12)
-            @. ᶜreice = FT(25)
+            @. ᶜreliq = ifelse(
+                cloud_liquid_water_content > eps(FT),
```
szy21, line 211: I wonder if eps(FT) threshold is too large for Float32? Maybe we can use 0 here? Hopefully if doesn't matter too much when the cloud water is very small.

src/diagnostics/radiation_diagnostics.jl
```diff
@@ -857,3 +857,88 @@ add_diagnostic_variable!(
...
+            ) .* FT(1e-6)
+    end
+end
+
+add_diagnostic_variable!(
+    short_name = "reffclw",
+    long_name = "Effective radius for liquid clouds",
+    standard_name = "effective_radius_of_cloud_liquid_particles",
+    units = "m",
+    comments = "In-cloud ratio of the third moment over the second moment of the particle size distribution. Set to zero outside of clouds.",
```
szy21, line 900: If you do the ifelse thing in the end, let's add in the comments that it is set to zero when there is no cloud, so users know the monthly mean diagnostics is inaccurate.

szy21 (approved):

src/diagnostics/radiation_diagnostics.jl
trontrytel: I have it now in the comments. Would you prefer in the long name?

src/callbacks/callbacks.jl
trontrytel: Sounds good. will change that

src/diagnostics/radiation_diagnostics.jl
szy21: No this is good enough, thanks!

.buildkite/Project.toml
szy21, line 1: I don't fully understand how package works, but I think you shouldn't need CloudMicrophysics and RRTMGP specified here if the compact is specified in the other Project.toml?

trontrytel: I don't understand this either. It seems like some packages are specified there, for example `Timesteppers` or `ClimaCore` but not others. I guess it's the Manifest that matters anyway, since we are checking it in?

trontrytel: I'll try one more time without those and see if I can still use the correct package versions in the CI

PR #3625, opened 2025-02-12
Bump to 0.28.5

Sbozzolo:
Breaking at least because we removed the Frieson diffusion (maybe other reasons as well)


szy21:
I did remove all the `FriersonDiffusion` in ClimaCoupler before merging ClimaAtmos PR, but it makes sense to me if you think it's better to make it a breaking release.

szy21 (approved):

PR #3631, opened 2025-02-14
add AOD to diagnostics

szy21:
## Purpose 

Adds AOD to AerosolState and adds AOD diagnostics. This requires RRTMGP 0.21.0

## Content

----


src/diagnostics/radiation_diagnostics.jl
```diff
@@ -942,3 +942,46 @@ add_diagnostic_variable!(
...
+    state,
+    cache,
+    time,
+    radiation_mode::Union{
+        RRTMGPI.AllSkyRadiationWithClearSkyDiagnostics,
+        RRTMGPI.AllSkyRadiation,
+        RRTMGPI.ClearSkyRadiation,
+    },
+)
+    @assert cache.atmos.radiation_mode.aerosol_radiation "aerosol_radiation must be true to enable aerosol optical depth diagnostics"
```
szy21, line 965: @Sbozzolo Is there a better way to do this?

src/diagnostics/radiation_diagnostics.jl
```diff
@@ -942,3 +942,46 @@ add_diagnostic_variable!(
     comments = "In-cloud ratio of the third moment over the second moment of the particle size distribution. Set to zero outside of clouds.",
     compute! = compute_reffcli!,
 )
+
+###
+# Effective radius for ice clouds (3d)
```
Sbozzolo, line 947: This comment is incorrect

src/diagnostics/radiation_diagnostics.jl
Sbozzolo: We can use the same error system as the other diagnostics by dispatching over Val(cache.atmos.radiation_mode.aerosol_radiation)

Sbozzolo: Let me try

Sbozzolo: I think the alternative is uglier, so I think you can stick with this

szy21: Ok, thanks!

charleskawczynski (commented):
These changes look fine to me. I think two minor things need updated, but I'm fine with merging once they're addressed:

 - I think we do need to remove RRTMGP from the PrecompileCI environment (I probably added that by accident).

 - Also, we'll need to make a (breaking) release of RRTMGP before we can merge this.

szy21:
> These changes look fine to me. I think two minor things need updated, but I'm fine with merging once they're addressed:
> 
> * I think we do need to remove RRTMGP from the PrecompileCI environment (I probably added that by accident).
> * Also, we'll need to make a (breaking) release of RRTMGP before we can merge this.

Sounds good. Yes, I'll release RRTMGP first, and I will revert the environment changes and just update RRTMGP to the latest version.

PR #3640, opened 2025-02-18
Add MERRA_AOD artifact and accessor

imreddyTeja:
## Purpose 
Add the new MERRA_AOD artifact

## Content
Adds the artifact and a function that returns its path. 

----


szy21 (commented):
We won't use it in ClimaAtmos. We will use it in the future in the leaderboard plots and/or for calibration. Should we still add it to ClimaAtmos Artifacts.toml in this case?

imreddyTeja:
> We won't use it in ClimaAtmos. We will use it in the future in the leaderboard plots and/or for calibration. Should we still add it to ClimaAtmos Artifacts.toml in this case?

I don't think we should add it then.

PR #3644, opened 2025-02-19
Bugfix: DYAMOND ic with EDOnly

akshaysridhar:
Fix initial condition property (this change has been used in high-res simulations not in the CI pipeline). This pattern of overwriting initial conditions should be reviewed in the future.

### DIFF
--- a/config/gpu_configs/gpu_aquaplanet_dyamond_summer.yml
+++ b/config/gpu_configs/gpu_aquaplanet_dyamond_summer.yml
@@ -12,7 +12,8 @@ rad: "allskywithclear"
 insolation: "timevarying"
 dt_rad: "1hours"
 dt_cloud_fraction: "1hours"
-vert_diff: "DecayWithHeightDiffusion"
+turbconv: "edonly_edmfx"
+edmfx_sgs_diffusive_flux: true
 implicit_diffusion: true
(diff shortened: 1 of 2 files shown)


szy21:
Could you also change the ci config?

sriharshakandala:
Looks good to me!

.buildkite/pipeline.yml
```diff
@@ -282,6 +282,19 @@ steps:
         artifact_paths: "held_suarez_equil/output_active/*"
         agents:
           slurm_constraint: icelake|cascadelake|skylake|epyc
+
```
szy21, line 285: We already have `gpu_aquaplanet_dyamond_summer` in ci. Let's not add more tests. Could you just change it to use ed only? Thanks!

reproducibility_tests/ref_counter.jl
```diff
@@ -1,4 +1,4 @@
-215
+216
```
szy21, line 1: no need to change this either

.buildkite/pipeline.yml
szy21: feel free to move it to model_configs too if you think that's better.

szy21 (approved):

PR #3652, opened 2025-02-24
Automatically choose diagnostic grid

### LINKED ISSUE #3651: Automatically set diagnostics resolution based on model resolution by default
[Writing in progress]

Sbozzolo:
If `netcdf_interpolation_num_points` is not provided, `ClimaAtmos` will to determine it automatically by matching approximately choosing the same number of points as the model grid. In this, we distribute points equally along the two horizontal dimensions.

Closes #3651

### DIFF
--- a/src/callbacks/get_callbacks.jl
+++ b/src/callbacks/get_callbacks.jl
@@ -11,6 +11,7 @@ function get_diagnostics(
     (; dt, start_date) = sim_info
 
     FT = Spaces.undertype(axes(Y.c))
+    context = ClimaComms.context(axes(Y.c))
 
     # We either get the diagnostics section in the YAML file, or we return an empty list
     # (which will result in an empty list being created by the map below)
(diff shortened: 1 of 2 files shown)


src/callbacks/get_callbacks.jl
```diff
@@ -40,9 +41,16 @@ function get_diagnostics(
...
+        # Estimate the number of points we need to cover the entire domain
+        # ncolumns is the number of local columns
+        tot_num_columns =
+            ClimaComms.nprocs(context) * Fields.ncolumns(axes(Y.c))
+        if parsed_args["config"] == "plane"
+            num_points = tot_num_columns
+        else
+            num_points = round(Int, sqrt(tot_num_columns))
+        end
+        num_netcdf_points = (num_points, num_points, Spaces.nlevels(axes(Y.c)))
```
szy21, line 53: I think you want nlon = 2 * nlat, and nlon * nlat = total_num_columns?

szy21 (approved):

src/callbacks/get_callbacks.jl
```diff
@@ -40,9 +41,26 @@ function get_diagnostics(
...
+        elseif parsed_args["config"] == "box"
+            num2 = round(Int, sqrt(tot_num_columns))
+            num1 = num2
+        elseif parsed_args["config"] == "column"
+            # We need at least two points horizontally because our column is
+            # actually a box
+            num1, num2 = 2, 2
+        else
+            error("Uncaught case")
+        end
```
charleskawczynski, line 62: I just want to point out that every time we used `parsed_args` for if-else statements like this (instead of using, e.g., the `atmos` struct), we may be introducing more inference failures, because the compiler may not be able to guarantee that this won't result in an error.

Issue #3024, opened 2024-05-16
Problems with the user experience


Sbozzolo:
> :warning: Work in progress

This issue is meant to be a place where we can collect problems with the ClimaAtmos interface as we come across them.

##### No errors are produced when wrong key-value are set in the YAML file. 

For example, setting `approxmate_linear_solve_iters = 3` will silently run a simulation were the number of iterations for the approximate linear solve is 1.

##### Help messages are unhelpful

For example
```
divergence_damping_factor:
  help: "Divergence damping factor"
  value: 1.0
```
But even,
```
check_conservation:
  help: "Check conservation of mass and energy [`false` (default), `true`]"
  value: false
```
What does this mean? Is this check online/offline? Does it affec the evolution?

### It is not possible to opt-out everything and only turn on certain options

This leads to all sorts of stuff being there even if not needed (e.g., there should be no reference to clouds in a simulation of dry baroclinic waves).

##### Some options do nothing (ie, maintaing the list is a difficult task)

For example,
```
fps:
  help: "Frames per second for animations"
  value: 5
```

This is a symptom of the fact that a large list of options like what we have is not easy to maintain.

##### Incompatible parameters are allowed

Such as `deep_atmosphere = true` and `config = box`.

##### Too much information is in the defaults

Some options should really be in the config files. For example, `t_end`, `dt`. 

> :warning: Work in progress

akshaysridhar:
There was a collection of ideas here: https://github.com/CliMA/ClimaAtmos.jl/issues/427 which may also be helpful as we think about the general ClimaAtmos workflow.

Sbozzolo:
Tonight, I tried using ClimaAtmos as a library to set up a baroclinc wave experiment.

I falied. Major problems I encountereda re: parameters, cache, surface stuff.
```julia

# ClimaCore is responsible for the computational domain and the discretization
import ClimaCore

# ClimaTimeSteppers contain algorithms for the temporal discretization
import ClimaTimeSteppers as CTS

import SciMLBase

import ClimaAtmos as CA

# Define the floating-point type
FT = Float32

meter = meters = seconds = second = one(FT)
kilometers = 1000meters
day = days = 86400seconds
earth_radius = 6700kilometers

# Parameters of the grid

# Using a vertical grid stretched with an hyperbolic tangent
number_of_vertical_points = 10
minimum_elevation = 0meters
maximum_elevation = 30kilometers
height_of_bottom_cell = 500meters
[...]

PR #3655, opened 2025-02-25
Add dry baro wave example

charleskawczynski:
This PR is an attempt to add a dry baroclinic wave example, where we use ClimaAtmos as a library. This is very much in a draft state, there's no documentation, there's likely unused variables, but it compiles and runs.

There is:
 - effectively no arg parsing, so it's relatively type-stable.
 - no callbacks or diagnostics 😅, so we can't easily verify the results

One nice thing is that, while it only runs for 10 timesteps, the whole script takes ~80 seconds to run (most of which is compilation). I _think_ that this may be an improvement over the existing CI driver, but I've not yet timed it to compare.

cc @Sbozzolo / @dennisYatunin , who were interested in adding examples like this.

I wanted to see what a script like this might look like. It's pretty clear that an explicit version of this example would be hundreds of lines fewer, but I thought it'd be helpful if the example at least included implicit tendency complexity.

Oh, and I used chatgpt to make the intro ascii art.

### DIFF
--- a/.buildkite/Manifest-v1.11.toml
+++ b/.buildkite/Manifest-v1.11.toml
@@ -2,7 +2,7 @@
 
 julia_version = "1.11.3"
 manifest_format = "2.0"
-project_hash = "04945e2a7116cfacb2d57340f477dbe23e208227"
+project_hash = "a25db2ac38461b28c1b3a6176008dee03780354b"
 
 [[deps.ADTypes]]
 git-tree-sha1 = "fb97701c117c8162e84dfcf80215caa904aef44f"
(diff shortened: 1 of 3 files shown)


charleskawczynski:
> Here's my attempt: [#3024 (comment)](https://github.com/CliMA/ClimaAtmos.jl/issues/3024#issuecomment-2519420853)

Yeah, the issue with that is that it's not really hackable since you're still using the full set of tendencies in ClimaAtmos (also it being implicit makes things simpler, as we discussed). Let me try running that example, I'm curious how long it takes to run. In my example, you could hack in a tendency / change anywhere.

It might be nice if we added generic code injection into each of our methods. That might help strip out some pieces.

I think this example could be further simplified, but I would kind of want to generate a plot before doing more things that could introduce bugs

charleskawczynski:
Moving into a more functional form will help, too.

PR #3660, opened 2025-02-27
Prescribe N in r_eff liquid clouds

trontrytel:


src/callbacks/callbacks.jl
```diff
@@ -209,8 +209,14 @@ NVTX.@annotate function rrtmgp_model_callback!(integrator)
             # RRTMGP needs effective radius in microns
             @. ᶜreliq = ifelse(
                 cloud_liquid_water_content > FT(0),
-                CM.CloudDiagnostics.effective_radius_const(cmc.liquid) *
-                m_to_um_factor,
+                CM.CloudDiagnostics.effective_radius_Liu_Hallet_97(
+                    cmc.liquid,
+                    Y.c.ρ,
+                    cloud_liquid_water_content / max(FT(0.1), cloud_fraction),
```
szy21, line 215: 0.1 is pretty large. Does FT(eps) not work?

szy21 (approved):

trontrytel:
Dividing by FT(eps) makes r_eff sometimes go to pretty large values. This does not affect the simulation outcome too much because we clip the radius in the radiation lookup tables. 

I'll merge this PR in after the CI slurm and box errors are fixed. I will have to bump the ref counter because of the two test cases with radiation that we have in the CI. 

In the future we might consider clipping the r_eff computation on the CloudMicrophysics side. Especially when we upgrade to 2M scheme and start modeling the number concentrations

nefrathenrici (approved):
Thank you!

PR #3661, opened 2025-02-27
Add density in cloud condensate RHS

trontrytel:
This PR:
- fixes the density bug in cloud condensate formation tendensices
- sets the microphysics condensation and precipitation timescales to be longer than dt in different configs
- fixes tracer vertical advection bug


oalcabes (approved):
Looks good to me!

szy21 (approved):
Thanks!

PR #3662, opened 2025-02-27
Use two Newton iterations for implicit diffusion with ARS343

dennisYatunin:
## Purpose 

This PR changes the value of `max_newton_iters_ode` from 1 to 2 for all configurations with `implicit_diffusion` set to `true`. From the ClimaTimeSteppers convergence tests, we know that simulations with implicit diffusion and IMEX ARK timestepping typically require 2 Newton iterations to show correct convergence behavior. From recent testing in ClimaAtmos, it appears that they also require 2 iterations to avoid large conservation errors for specific humidity.

Since we were previously using SSPKnoth for our `ode_algo`, the value of `max_newton_iters_ode` was ignored. Now that we are using ARS343, the value is no longer ignored, and the default of 1 is only valid when `implicit_diffusion` is `false`. This PR adds a warning to `type_getters.jl` so that developers will be notified when the default number of iterations is too low.

----

### DIFF
--- a/config/model_configs/aquaplanet_equil_clearsky_tvinsol_0M_slabocean.yml
+++ b/config/model_configs/aquaplanet_equil_clearsky_tvinsol_0M_slabocean.yml
@@ -7,8 +7,9 @@ surface_setup: "DefaultMoninObukhov"
 vert_diff: "DecayWithHeightDiffusion"
 implicit_diffusion: true
 approximate_linear_solve_iters: 2
+max_newton_iters_ode: 2
 rad: "clearsky"
 insolation: "timevarying"
 dt_rad: "1hours"
 prognostic_surface: "PrognosticSurfaceTemperature"
-#check_conservation: true
+check_conservation: true
--- a/config/model_configs/aquaplanet_equil_clearsky_tvinsol_0M_slabocean_ft64.yml
+++ b/config/model_configs/aquaplanet_equil_clearsky_tvinsol_0M_slabocean_ft64.yml
@@ -7,8 +7,9 @@ surface_setup: "DefaultMoninObukhov"
 vert_diff: "DecayWithHeightDiffusion"
 implicit_diffusion: true
 approximate_linear_solve_iters: 2
+max_newton_iters_ode: 2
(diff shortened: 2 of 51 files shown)


dennisYatunin:
Testing longruns here: https://buildkite.com/clima/climaatmos-gpulongruns/builds/493

config/model_configs/aquaplanet_equil_clearsky_tvinsol_0M_slabocean_ft64.yml
```diff
@@ -7,6 +7,7 @@ surface_setup: "DefaultMoninObukhov"
 vert_diff: "DecayWithHeightDiffusion"
 implicit_diffusion: true
 approximate_linear_solve_iters: 2
+max_newton_iters_ode: 2
```
szy21, line 10: Could you uncomment `check_conservation: true` in this file?

config/model_configs/aquaplanet_equil_clearsky_tvinsol_0M_slabocean.yml
```diff
@@ -7,6 +7,7 @@ surface_setup: "DefaultMoninObukhov"
 vert_diff: "DecayWithHeightDiffusion"
 implicit_diffusion: true
 approximate_linear_solve_iters: 2
+max_newton_iters_ode: 2
```
szy21, line 10: Could you uncomment `check_conservation: true` in this file?

szy21 (approved):

szy21:
I cancelled the longrun because Akshay is testing it here: https://buildkite.com/clima/climaatmos-gpulongruns/builds/492

szy21:
Actually the one that @akshaysridhar has only changed a few configs. I'm restarting the build here: https://buildkite.com/clima/climaatmos-gpulongruns/builds/494

szy21:
But you can merge the PR without waiting for the longrun.

szy21:
Note that aquaplanet without EDMF is ~25% slower, so it is a pretty significant slowdown.

PR #3663, opened 2025-02-28
Split implicit and explicit precomputed quantities

dennisYatunin:
## Purpose 

This PR splits the precomputed quantities that need to be handled implicitly (i.e., updated on each Newton iteration) from those that can be handled explicitly (i.e., held constant throughout each Newton solve).

This change should improve performance for all simulations with more than one Newton iteration, since we are currently spending a significant chunk of time updating precomputed quantities that can be handled explicitly. Specifically, this should reduce the performance penalty that we have observed in #3662 from increasing the number of Newton iterations for configurations with implicit diffusion.

----

### DIFF
--- a/src/cache/precomputed_quantities.jl
+++ b/src/cache/precomputed_quantities.jl
@@ -5,32 +5,84 @@ import Thermodynamics as TD
 import ClimaCore: Spaces, Fields
 
 """
-    precomputed_quantities(Y, atmos)
+    implicit_precomputed_quantities(Y, atmos)
 
-Allocates and returns the precomputed quantities:
-    - `ᶜspecific`: the specific quantities on cell centers (for every prognostic
+Allocates precomputed quantities that are treated implicitly (i.e., updated
+on each iteration of the implicit solver). This includes all quantities related
+to velocity and thermodynamics that are used in the implicit tendency.
+
+The following grid-scale quantities are treated implicitly:
+    - `ᶜspecific`: specific quantities on cell centers (for every prognostic
         quantity `ρχ`, there is a corresponding specific quantity `χ`)
-    - `ᶜu`: the covariant velocity on cell centers
-    - `ᶠu³`: the third component of contravariant velocity on cell faces
-    - `ᶜK`: the kinetic energy on cell centers
-    - `ᶜts`: the thermodynamic state on cell centers
-    - `ᶜp`: the air pressure on cell centers
-    - `sfc_conditions`: the conditions at the surface (at the bottom cell faces)
-    - `ᶜh_tot`: the total enthalpy on cell centers
-
-If the `turbconv_model` is EDMFX, there also two SGS versions of every quantity
-except for `ᶜp` (we assume that the pressure is the same across all subdomains):
-    - `_⁰`: the value for the environment
-    - `_ʲs`: a tuple of the values for the mass-flux subdomains
-In addition, there are several other SGS quantities for the EDMFX model:
-    - `ᶜρa⁰`: the area-weighted air density of the environment on cell centers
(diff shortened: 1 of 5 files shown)


szy21:
tagging @costachris because there are prognostic edmf changes

szy21:
Is it expected to change MSE?

src/cache/precomputed_quantities.jl
```diff
@@ -426,6 +444,34 @@ NVTX.@annotate function set_precomputed_quantities!(Y, p, t)
...
+        set_prognostic_edmf_precomputed_quantities_draft!(Y, p, ᶠuₕ³, t)
+        set_prognostic_edmf_precomputed_quantities_environment!(Y, p, ᶠuₕ³, t)
+    end
+end
+
+"""
+    set_explicit_precomputed_quantities!(Y, p, t)
+
+Updates the precomputed quantities that are handled explicitly based on the
+current state `Y`.
```
charleskawczynski, line 462: Can we also include which functions this impacts? For example: `remaining_tendencies!`, ....

src/cache/precomputed_quantities.jl
```diff
@@ -369,37 +395,29 @@ function eddy_diffusivity_coefficient(C_E, norm_v_a, z_a, p)
 end
 
 """
-    set_precomputed_quantities!(Y, p, t)
+    set_implicit_precomputed_quantities!(Y, p, t)
```
charleskawczynski, line 398: Can we also include which functions this impacts? For example: `implicit_tendencies!`, ....

charleskawczynski (commented):
Overall the changes look fine, but I'm also concerned that there are mse changes. Any idea what is wrong there?

src/cache/precomputed_quantities.jl
dennisYatunin: The quantities updated in `set_explicit_precomputed_quantities!` are required to compute both `implicit_tendency!` and `remaining_tendency!`; they are just *handled* explicitly. This means that they are updated once per Newton solve, rather than once per Newton iteration. These quantities include things like diffusivity and surface conditions, which affect the implicit tendency when `diff_mode isa Implicit`.

src/cache/precomputed_quantities.jl
dennisYatunin: The quantities updated in `set_implicit_precomputed_quantities!` are required to compute both `implicit_tendency!` and `remaining_tendency!`; they are just handled implicitly (updated once per Newton iteration, rather than once per Newton solve). These quantities include everything related to velocity and thermodynamics, so they affect every tendency in our model.

src/cache/precomputed_quantities.jl
charleskawczynski: Yeah, I'm just saying can we update the doc string to explicitly say this

dennisYatunin:
Yes, we do expect MSE changes for simulations with implicit diffusion. We were previously updating the entire cache before computing the implicit tendency, but now we are only updating quantities related to velocity and thermodynamics. In particular, we are no longer updating the diffusivity, so its value in the implicit tendency is now reused from the previous explicit tendency. Since we are computing the implicit tendency with a different diffusivity, we expect the results to change. However, the diffusivity is mostly constant with respect to the state, so we only have small MSE changes and no meaningful differences in the summary plots.

I am going to add a simple heat equation test with fixed diffusivity in a separate PR, and then I will be able to confirm that these changes have no effect on implicit diffusion when diffusivity is constant.

charleskawczynski (approved):

szy21:
@costachris Could you take a look at the prognostic edmf relevant changes in this PR and see if it makes sense to you and/or make sure you understand the change?

costachris (approved):

PR #3666, opened 2025-02-28
get pressure from precomputed quantities in rrtmgp callback

szy21:
## Content

----

### DIFF
--- a/src/callbacks/callbacks.jl
+++ b/src/callbacks/callbacks.jl
@@ -74,7 +74,7 @@ NVTX.@annotate function rrtmgp_model_callback!(integrator)
     p = integrator.p
     t = integrator.t
 
-    (; ᶜts, cloud_diagnostics_tuple, sfc_conditions) = p.precomputed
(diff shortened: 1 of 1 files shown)


src/callbacks/callbacks.jl
```diff
@@ -105,9 +105,8 @@ NVTX.@annotate function rrtmgp_model_callback!(integrator)
     sfc_T = Fields.array2field(rrtmgp_model.surface_temperature, axes(sfc_ts))
     @. sfc_T = TD.air_temperature(thermo_params, sfc_ts)
 
-    ᶜp = Fields.array2field(rrtmgp_model.center_pressure, axes(Y.c))
+    rrtmgp_model.center_pressure .= Fields.field2array(ᶜp)
```
Sbozzolo, line 108: Was there an issue with the pressure or is this just an optimization?

szy21: There was no issue, but I think it's better not to calculate it again if we already have it in precomputed quantities.

PR #3670, opened 2025-03-02
Make surface flux fully explicit

### LINKED ISSUE #3587: Fix conservation with ARS
The conservation tests for `aquaplanet_rhoe_equil_clearsky_tvinsol_0M_slabocean.yml` and `aquaplanet_rhoe_equil_clearsky_tvinsol_0M_slabocean_ft64.yml` are broken after switching to ARS. The error is several orders of magnitude too large. It seems the conservation only breaks when `implicit_diffusion` is true. We should look into it.

szy21:
## Purpose 

Surface flux was used as a boundary condition for vertical diffusion tendency. When using implicit diffusion, it causes errors in conservation as surface flux is accumulated in the explicit stage, but is updated in the implicit stage. This PR moves surface flux to fully explicit by adding an explicit boundary source and setting the boundary condition of diffusion flux to 0 (except for tke, which is treated slightly differently now but we are not considering tke in conservation). This should be ok as we don't take into account any Jacobian of surface flux in the implicit solver now.

Because the surface flux was used as a boundary condition, whether the surface flux tendency is applied was controlled by `vert_diff` or `edmfx_sgs_diffusive_flux`. The PR adds a flag `disable_surface_flux_tendency` that specifies whether surface flux is applied. This is independent from `surface_setup`, which specifies the calculation of surface conditions. We need a separate flag to control whether surface flux is applied because surface conditions may be needed for things other than surface flux, for example, mixing length in TKE-based diffusion and EDMF updraft boundary conditions.

Also changes `diffuse_momentum` to `disable_momentum_vertical_diffusion` to make things slightly simpler.

Testing some longruns here: https://buildkite.com/clima/climaatmos-gpulongruns/builds/499

Closes #3587 

## Content

----

### DIFF
--- a/src/prognostic_equations/surface_flux.jl
+++ b/src/prognostic_equations/surface_flux.jl
@@ -1,0 +1,70 @@
+#####
+##### Surface flux tendencies
+#####
+
+import ClimaCore.Geometry: ⊗
+import ClimaCore.Operators as Operators
+
+"""
+    Boundary tendency for momentum, applied through the divergence operator
+    with zeros in the interior and user specified boundary values. The
+    resulting tendency is only non-zero at the boundaries.
+"""
+function boundary_tendency_momentum(ᶜρ, ᶜuₕ, ρ_flux_uₕ)
+    FT = eltype(ᶜρ)
+    ᶠgradᵥ = Operators.GradientC2F()
+    ᶜdivᵥ_uₕ = Operators.DivergenceF2C(
+        top = Operators.SetValue(C3(FT(0)) ⊗ C12(FT(0), FT(0))),
+        bottom = Operators.SetValue(ρ_flux_uₕ),
+    )
+    return @. lazy(ᶜdivᵥ_uₕ((0 * ᶠgradᵥ(ᶜuₕ))) / ᶜρ)
+end
+
+"""
+    Boundary tendency for scalars, applied through the divergence operator
+    with zeros in the interior and user specified boundary values. The
+    resulting tendency is only non-zero at the boundaries.
+"""
+function boundary_tendency_scalar(ᶜχ, ρ_flux_χ)
+    FT = eltype(ᶜχ)
+    ᶠgradᵥ = Operators.GradientC2F()
+    ᶜdivᵥ_χ = Operators.DivergenceF2C(
+        top = Operators.SetValue(C3(FT(0))),
+        bottom = Operators.SetValue(ρ_flux_χ),
+    )
+    return @. lazy(ᶜdivᵥ_χ(0 * ᶠgradᵥ(ᶜχ)))
+end
+
+"""
+   Apply surface flux tendency as an explicit boundary source.
+   When `disable_surface_flux_tendency` is true, no surface flux tendency is applied
+   no matter what the surface conditions are.
+"""
+function surface_flux_tendency!(Yₜ, Y, p, t)
+
+    p.atmos.disable_surface_flux_tendency && return
+
+    FT = eltype(Y)
+    (; ᶜh_tot, ᶜspecific, sfc_conditions) = p.precomputed
+
+    if !disable_momentum_vertical_diffusion(p.atmos.vert_diff)
+        btt =
+            boundary_tendency_momentum(Y.c.ρ, Y.c.uₕ, sfc_conditions.ρ_flux_uₕ)
+        @. Yₜ.c.uₕ -= btt
+    end
+
+    btt = boundary_tendency_scalar(ᶜh_tot, sfc_conditions.ρ_flux_h_tot)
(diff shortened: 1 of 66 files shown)


szy21:
@dennisYatunin I don't really want to create a new `surface_fllux_model` for this. Ideally I want to reuse `sfc_setup`, but I also don't want to make it more complicated. I can't use sfc_setup directly right now because I want no surface flux tendency for `NoSurface` (which is what I added in the PR), but I need to have surface flux for `PrescribedSurface` (for coupler). Right now they both set `sfc_steup` to nothing. Do you see an easy way to make it work? Otherwise I'm fine with what I have.

Sbozzolo:
Could you please make sure to document the feature you added?

The surface conditions code is incredibly messy and we must improve clarity.

src/solver/types.jl
```diff
@@ -561,6 +564,7 @@ Base.@kwdef struct AtmosModel{
     sfc_temperature::ST = nothing
     insolation::IN = nothing
     surface_model::SM = nothing
+    surface_flux_model::SFM = nothing
```
Sbozzolo, line 567: I would prefer we didn't introduce yet another "surface" model setting. It adds complexity and confusion and makes coupling harder.

All the configurations related to the surface should be in a single place (this should include `sfc_temperature` and `surface_albedo`) so that we can easily understand and control them, without worrying about interdependencies and interactions.

szy21: I don't think we can avoid a new surface model. Right now whether we apply surface flux or not is controlled by vertical diffusion, which is not ideal. So we need something else to specify it. This is different from `sfc_setup` as there are some cases where we want to have surface flux (for EDMF updraft calculations), but we don't want to apply the tendency. We may be able to incorporate it within `sfc_setup`, but that will make surface conditions more complicated. I will talk with Dennis to see what would be the easiest and not too confusing way for this.

src/solver/types.jl
```diff
@@ -560,6 +563,7 @@ Base.@kwdef struct AtmosModel{
     rayleigh_sponge::RS = nothing
     sfc_temperature::ST = nothing
     insolation::IN = nothing
+    disable_surface_flux_tendency::Bool = false
```
Sbozzolo, line 566: Could you add a docstring to describe what this does here?

src/surface_conditions/surface_setups.jl
```diff
@@ -1,5 +1,13 @@
 abstract type SurfaceSetup end
 
+"""
+    NoSurface()
+
+Used to indicate that there is no surface flux.
+"""
+struct NoSurface <: SurfaceSetup end
+(surface_setup::NoSurface)(params) = nothing
+
```
Sbozzolo, line 10: Is this needed for this PR?

szy21: no, I'll remove it.

src/prognostic_equations/vertical_diffusion_boundary_layer.jl
```diff
@@ -29,47 +21,32 @@ function vertical_diffusion_boundary_layer_tendency!(
...
-            bottom = Operators.SetValue(ρ_flux_χ),
+            bottom = Operators.SetValue(C3(FT(0))),
         )
         @. ᶜρχₜ_diffusion =
             ᶜdivᵥ_ρχ(-(ᶠinterp(Y.c.ρ) * ᶠinterp(ᶜK_h) * ᶠgradᵥ(ᶜχ)))
         @. ᶜρχₜ -= ᶜρχₜ_diffusion
-        if !(χ_name in (:q_rai, :q_sno))
-            @. Yₜ.c.ρ -= ᶜρχₜ_diffusion
-        end
+        @. Yₜ.c.ρ -= ᶜρχₜ_diffusion
```
szy21, line 50: @trontrytel here is the fix

src/prognostic_equations/surface_flux.jl
```diff
@@ -0,0 +1,70 @@
...
+    resulting tendency is only non-zero at the boundaries.
+"""
+function boundary_tendency_scalar(ᶜχ, ρ_flux_χ)
+    FT = eltype(ᶜχ)
+    ᶠgradᵥ = Operators.GradientC2F()
+    ᶜdivᵥ_χ = Operators.DivergenceF2C(
+        top = Operators.SetValue(C3(FT(0))),
+        bottom = Operators.SetValue(ρ_flux_χ),
+    )
+    return @. lazy(ᶜdivᵥ_χ((0 * ᶠgradᵥ(ᶜχ))))
```
Sbozzolo, line 35:
```suggestion
    return @. lazy(ᶜdivᵥ_χ(0 * ᶠgradᵥ(ᶜχ)))
```

Sbozzolo (approved):

PR #3672, opened 2025-03-03
change some longrun configs

szy21:
## Purpose 

Switch most longruns to TKE-based diffusion. Also reduce the frequency of diagnostics as some artifacts are too large to upload after #3652.

## Content

----


szy21:
Manually merging this as it only changes longrun configs.

PR #3673, opened 2025-03-05
Adapt caches

Sbozzolo:
ClimaCoupler is implementing a way to save and restart from checkpoints that relies on the beign able to save caches to disk. For this to happen, we first have to move it to the GPU. This commit adds the necessary Adapt methods.

This commit also fixes an out-of-memory test.

### DIFF
--- a/src/cache/cache.jl
+++ b/src/cache/cache.jl
@@ -70,6 +70,9 @@ struct AtmosCache{
     conservation_check::CONSCHECK
 end
 
+# Allow cache to be moved on the CPU. Used by ClimaCoupler to save checkpoints
+Adapt.@adapt_structure AtmosCache
+
 # Functions on which the model depends:
 # CAP.R_d(params)         # dry specific gas constant
 # CAP.kappa_d(params)     # dry adiabatic exponent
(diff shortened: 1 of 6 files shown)


src/cache/cache.jl
```diff
@@ -70,6 +70,9 @@ struct AtmosCache{
     conservation_check::CONSCHECK
 end
 
+# Allow cache to be moved on the CPU. Used by ClimaCoupler to save checkpoints
+ClimaCore.Adapt.@adapt_structure AtmosCache
```
charleskawczynski, line 74: I kind of think that Atmos should just directly depend on Adapt, it's a lightweight package, and code loading like this is a bit confusing (and could depend on ClimaCore versions).

@dennisYatunin, @imreddyTeja, @nefrathenrici, what do you think?

Sbozzolo: I am fine with that

nefrathenrici: Sounds good to me, Adapt does not introduce many dependencies

charleskawczynski (approved):
I think we just need to update the manifests

PR #3688, opened 2025-03-11
Add date to Pressel et al. reference

jty53:
## Purpose 
Add date to source code comment to clarify reference

## To-do

[ ] Verify Pressel et al., 2015 is intended reference

## Content

----
- [ x] I have read and checked the items on the review checklist.


szy21 (approved):
Thanks!

PR #3691, opened 2025-03-12
Remove data_loader in RRTMGPInterface

charleskawczynski:
This PR removes `data_loader`, which was originally added in to avoid explicitly depending on NCDatasets. But now that we do, I think there's no need for this abstraction.


szy21 (approved):

charleskawczynski:
Exit status says 255, which is a pass, despite the job reporting as failed. I'm going to merge and see what happens on the next PR.

PR #3692, opened 2025-03-12
Move precipitation cache to precomputed quantities

trontrytel:
Precipitation cache followed an old pattern and was:
- defined separately from precomputed quantities
- updated when the precipitation tendency was called.

This was making it difficult to track "who gets updated when" when coupling 1M microphysics with EDMF. This PR moves all precipitation cache into precomputed quantities

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "ClimaAtmos"
 uuid = "b2c96348-7fb7-4fe0-8da9-78d88439e717"
 authors = ["Climate Modeling Alliance"]
-version = "0.28.6"
+version = "0.29.0"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
--- a/src/cache/precipitation_precomputed_quantities.jl
+++ b/src/cache/precipitation_precomputed_quantities.jl
@@ -1,20 +1,57 @@
 #####
-##### Precomputed quantities
+##### Precomputed quantities for precipitation processes
 #####
-import CloudMicrophysics.Microphysics1M as CM1
+
 import CloudMicrophysics.MicrophysicsNonEq as CMNe
+import CloudMicrophysics.Microphysics1M as CM1
+
+import Thermodynamics as TD
+import ClimaCore.Operators as Operators
+import ClimaCore.Fields as Fields
+
+const Iₗ = TD.internal_energy_liquid
+const Iᵢ = TD.internal_energy_ice
 
 """
-    set_precipitation_precomputed_quantities!(Y, p, t)
+   Kin(ᶜw_precip, ᶜu_air)
 
-Updates the precipitation terminal velocity stored in `p`
-for the 1-moment microphysics scheme
+    - ᶜw_precip - teminal velocity of cloud consensate or precipitation
+    - ᶜu_air - air velocity
+
+Helper function to compute the kinetic energy of cloud condensate and
+precipitation.
 """
-function set_precipitation_precomputed_quantities!(Y, p, t)
-    @assert (p.atmos.precip_model isa Microphysics1Moment)
+function Kin(ᶜw_precip, ᶜu_air)
+    return @. lazy(
+        norm_sqr(
+            Geometry.UVWVector(0, 0, -(ᶜw_precip)) + Geometry.UVWVector(ᶜu_air),
+        ) / 2,
+    )
+end
 
-    (; ᶜwᵣ, ᶜwₛ) = p.precomputed
+"""
+    set_precipitation_velocities!(Y, p, moisture_model, precip_model)
+
+Updates the precipitation terminal velocity, cloud sedimentation velocity,
+and their contribution to total water and energy advection.
+"""
+function set_precipitation_velocities!(Y, p, _, _)
+    (; ᶜwₜqₜ, ᶜwₕhₜ) = p.precomputed
+    @. ᶜwₜqₜ = Geometry.WVector(0)
+    @. ᶜwₕhₜ = Geometry.WVector(0)
+    return nothing
+end
+function set_precipitation_velocities!(
+    Y,
+    p,
+    moisture_model::NonEquilMoistModel,
+    precip_model::Microphysics1Moment,
+)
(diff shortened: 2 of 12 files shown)


trontrytel:
I see that I'm breaking the coupler compatibility tests because I removed `precipitation` from `cache`. Could you point me to where `precipitation` Atmos cache is used in the Coupler? I can open a PR to fix this @juliasloan25 @Sbozzolo ? (Sorry, I couldn't find it myself)

szy21:
> I see that I'm breaking the coupler compatibility tests because I removed `precipitation` from `cache`. Could you point me to where `precipitation` Atmos cache is used in the Coupler? I can open a PR to fix this @juliasloan25 @Sbozzolo ? (Sorry, I couldn't find it myself)

Here, for example: https://github.com/CliMA/ClimaCoupler.jl/blob/710a9f5c703ae40d7caea88b10cbc9339a630b3f/experiments/ClimaEarth/components/atmosphere/climaatmos.jl#L180 (and probably other places too)

trontrytel:
experiments folder! Nvmd

src/cache/precipitation_precomputed_quantities.jl
```diff
@@ -1,20 +1,57 @@
...
+    set_precipitation_velocities!(Y, p, moisture_model, precip_model)
 
-    (; ᶜwᵣ, ᶜwₛ) = p.precomputed
+Updates the precipitation terminal velocity, cloud sedimentation velocity,
+and their contribution to total water and energy advection.
+"""
+function set_precipitation_velocities!(Y, p, _, _)
+    (; ᶜwₜqₜ, ᶜwₕhₜ) = p.precomputed
+    @. ᶜwₜqₜ = Geometry.WVector(0)
+    @. ᶜwₕhₜ = Geometry.WVector(0)
```
charleskawczynski, line 41: In this case, where are these values used?

src/cache/precipitation_precomputed_quantities.jl
```diff
@@ -57,5 +79,286 @@ function set_sedimentation_precomputed_quantities!(Y, p, t)
...
+        col_integrated_precip_energy_tendency,
+        ᶜS_ρe_tot,
+    )
+    # update surface precipitation fluxes in cache for coupler's use
+    thermo_params = CAP.thermodynamics_params(p.params)
+    T_freeze = TD.Parameters.T_freeze(thermo_params)
+    FT = eltype(p.params)
+    @. ᶜT = TD.air_temperature(thermo_params, ᶜts)
+    @. ᶜ3d_rain = ifelse(ᶜT >= T_freeze, ᶜS_ρq_tot, FT(0))
+    @. ᶜ3d_snow = ifelse(ᶜT < T_freeze, ᶜS_ρq_tot, FT(0))
```
charleskawczynski, line 325:
```suggestion
    ᶜ3d_rain = @. lazy(ifelse(ᶜT >= T_freeze, ᶜS_ρq_tot, FT(0)))
    ᶜ3d_snow @. lazy(ifelse(ᶜT < T_freeze, ᶜS_ρq_tot, FT(0)))
```
I think this might work, in which case, you could eliminate this variable from the cache entirely!

src/cache/precipitation_precomputed_quantities.jl
trontrytel: Yeah. They are used in https://github.com/CliMA/ClimaAtmos.jl/blob/aca95c135fd0cbd3722e5becd4a2f15a94e061d0/src/prognostic_equations/remaining_tendency.jl#L21

But good point. I'll write two versions of that function and then not need the zeros

szy21: You may need to write two versions for the implicit solver too, if we ever make it implicit. But that should be doable, just a bit more work.

szy21: (and if it's too much work, I'm fine with just setting it zero)

szy21 (commented):
I didn't read the details of the functions but the code change looks good to me. If there are MSE changes I can look at the code in more details.

src/cache/precipitation_precomputed_quantities.jl
trontrytel: Then maybe lets keep it as is for now. And once we settle on whether this should be implicit or explicit I can simplify it.

src/cache/precipitation_precomputed_quantities.jl
trontrytel: I think it did! Thank you

trontrytel:
> I didn't read the details of the functions but the code change looks good to me. If there are MSE changes I can look at the code in more details.

Thanks! I don't expect any MSE changes 🤞

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "ClimaAtmos"
 uuid = "b2c96348-7fb7-4fe0-8da9-78d88439e717"
 authors = ["Climate Modeling Alliance"]
-version = "0.28.6"
+version = "0.28.7"
```
szy21, line 4: Should we make it a breaking release?

trontrytel: I think a patch release is good enough? I think this way will be faster to propagate in the Coupler

trontrytel: Here is the PR I have in the Coupler https://github.com/CliMA/ClimaCoupler.jl/pull/1222

szy21: I'm fine either way - whatever @juliasloan25 prefers. And I think we should expect `[[deps.ClimaAtmos]]` in Manifest-v1.11.toml to change as well.

trontrytel: Sounds good. Lmk what is the least disruptive and relatively fast way to do it :)

juliasloan25: To follow best practices it should be a minor release since the interface is changing. One reason it would be nice to have it as a minor version is: in the coupler we'll support both this structure and the previous with an if statement (like `if pkgversion(ClimaAtmos) >= 0.29 integrator.p.precomputed.... else integrator.p.precipitation....`). When we eventually stop supporting both, we can set the compat in the coupler to `0.29` instead of specifying the patch `0.28.7`.

This is a pretty minor example (no pun intended), but the "best" practice is to make a minor release

szy21 (approved):
I think [[deps.ClimaAtmos]] in Manifest-v1.11.toml need to be changed. Everything else looks good!

PR #3693, opened 2025-03-12
Replace cdot with * using new ClimaCore branch

dennisYatunin:
## Purpose 

This PR tries out Clima/ClimaCore.jl#2232 in ClimaAtmos, replacing all uses of `⋅` with `*` in the implicit solver.

----


charleskawczynski (commented):
Why is this behavior changing?

dennisYatunin:
The behavior changes may be due to another recent PR in ClimaCore. I will try rerunning CI using the same ClimaCore branch, but with the original ⋅ operator instead of *. Hopefully this will exhibit the same behavior changes.

dennisYatunin:
CI exhibits the same behavior changes with the original ⋅ operator, so they must be due to some other recent PR in ClimaCore. I'll go ahead and merge in CliMA/ClimaCore.jl#2232.

charleskawczynski:
I think we can close this?

PR #3694, opened 2025-03-12
Refactor RRTMGP interface

charleskawczynski:
This PR:

 - Assumes `implied_values = :face`, which is the only setting that we currently support, and supporting `implied_values = :center` would involve changing pressure and temperature to live on cell faces, which would be very complicated in terms of a model feature.
 - Moves `radiation_mode` from a keyword arg to a positional arg, so that we can dispatch on it. This should allow us to simplify, and make assumptions inside, the `GrayRadiation`, `ClearSkyRadiation` etc. constructors.

This is technically a breaking change, since the `RRTMGPModel` interface is changing. We could add internal method constructors and pass it to a keyword constructor. Do people prefer that?

### DIFF
--- a/src/parameterized_tendencies/radiation/RRTMGPInterface.jl
+++ b/src/parameterized_tendencies/radiation/RRTMGPInterface.jl
@@ -404,12 +404,35 @@ array.
     - `center_z`: z-coordinate in m at cell centers
     - `face_z`: z-coordinate in m at cell faces
 """
-function RRTMGPModel(
+RRTMGPModel(
     params::RRTMGP.Parameters.ARP,
     context;
     ncol::Int,
     domain_nlay::Int,
     radiation_mode::AbstractRRTMGPMode = ClearSkyRadiation(),
+    interpolation::AbstractInterpolation = NoInterpolation(),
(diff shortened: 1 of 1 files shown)


src/parameterized_tendencies/radiation/RRTMGPInterface.jl
```diff
@@ -430,49 +449,9 @@ function RRTMGPModel(
...
-        error("please specify either center_pressure and center_temperature, \
-               or face_pressure and face_temperature, or all four values")
-    end
-
-    if implied_values != :face
-        if !(bottom_extrapolation isa SameAsInterpolation)
-            @warn "bottom_extrapolation is ignored if face_pressure and \
-                   face_temperature are specified"
-        end
+    if interpolation isa NoInterpolation
```
szy21, line 452: I think we can remove `NoInterpolation` too.

szy21 (approved):

src/parameterized_tendencies/radiation/RRTMGPInterface.jl
charleskawczynski: I'll do that in the next one.

PR #3696, opened 2025-03-13
Hoist lookup table logic

charleskawczynski:
This PR hoists the lookup table logic per radiation mode.

While this increases the number of lines of code, this helps break down `RRTMGPModel` into a much smaller function, and the interface to `lookup_tables` is nice in that it only requires

 - Radiation mode (mostly consists of Bools and Floats)
 - the ClimaComms device
 - the float type

So, this function does not need loads of other pieces within the scope of `RRTMGPModel`. This means that we can more easily interactively experiment with it and write unit tests for it.


charleskawczynski:
If we really want to focus on lines of code, we could write a small helper inside some of these functions and do:

```julia
    artifact(t, b, n) = NC.Dataset(RRTMGP.ArtifactPaths.get_lookup_filename(t, b)) do ds
        getproperty(RRTMGP.LookUpTables, n)(ds, FT, DA)
    end
    lookup_lw, idx_gases_lw = artifact(:gas, :lw, :LookUpLW)
    lookup_lw_cld = artifact(:cloud, :lw, :LookUpCld)
    lookup_lw_aero, idx_aerosol, idx_aerosize =
        if radiation_mode.aerosol_radiation
            artifact(:aerosol, :lw, :LookUpAerosolMerra)
        else
            (nothing, nothing, nothing)
        end
    lookup_sw, idx_gases_sw = artifact(:gas, :sw, :LookUpSW)
    lookup_sw_cld = artifact(:cloud, :sw, :LookUpCld)

    lookup_sw_aero, idx_aerosol, idx_aerosize =
        if radiation_mode.aerosol_radiation
            artifact(:aerosol, :sw, :LookUpAerosolMerra)
        else
            (nothing, nothing, nothing)
        end
```
It's nice in that nearly the whole function can fit on a page, but I'm not particularly a fan of writing helpers that dig into modules with symbols. Given that this is a somewhat niche function, I'm open to this "lines of code optimization"

szy21:
tagging @dennisYatunin as I'm not sure if I will get to this tonight, but if not I will review this tomorrow.

szy21 (approved):
What are the `# TODO: verify correctness`?

PR #3698, opened 2025-03-13
Encapsulate rrtmgp model kwargs

charleskawczynski:
This PR encapsulates the rrtmgp model kwargs, and will also allow us to dispatch on the radiation mode.

This should help make it more clear what kwargs are passed to `RRTMGPModel`, so that we can start tackling refactoring of the views (which all depend on the input kwargs).

I think more is needed, but this is a start and I want to make sure I haven't broken anything.


src/parameterized_tendencies/radiation/RRTMGPInterface.jl
```diff
@@ -574,6 +574,7 @@ function _RRTMGPModel(
     use_global_means_for_well_mixed_gases::Bool = false,
     kwargs...,
 )
+    @assert pkgversion(RRTMGP) > v"0.19.2"
```
szy21, line 577: We already have RRTMGP compat set to v0.21 in Project.toml, so maybe we can remove this line?

szy21 (approved):

PR #3699, opened 2025-03-13
Remove rrtmgp property overloading

charleskawczynski:
This PR removes the `getproperty` / `propertynames` overloading on the rrtmgp model. Currently, it's difficult to know, by inspection, what variables a given model has. Requiring explicit getproperty calls into the views makes inspecting the variables much easier.


charleskawczynski:
@dennisYatunin is alright with the essence of this PR.

szy21:
I think we use `p.radiation.rrtmgp_model.` in some places in the coupler. Maybe double check those?

charleskawczynski:
> I think we use `p.radiation.rrtmgp_model.` in some places in the coupler. And the coupler downstream test failed. Could you check that?

Ah, yeah, that's a good point. Hmm, let me think about this

PR #3701, opened 2025-03-14
Fix and test `TargetJobConfig`, remove duplicate config values

nefrathenrici:
## Content

----


config/gpu_configs/gpu_aquaplanet_dyamond_ws_1process.yml
```diff
@@ -5,7 +5,6 @@ h_elem: 30
 z_max: 60000.0
 z_elem: 63
 dz_bottom: 30.0
-rayleigh_sponge: true
```
szy21, line 8: Just me being picky - could we remove the second one below? I would like to keep both sponge configs close to each other so I can easily find them.

szy21 (approved):

charleskawczynski (approved):
Thank you!

PR #3702, opened 2025-03-14
add a scaling factor for precipitation diffusion

szy21:
## Purpose 

Add a scaling factor for hyperdiffusion and vertical diffusion of precipitating species so we can tune it. To preserve the current behavior the default of the hyperdiffusion factor is set to 0 and the default of the vertical diffusion factor is set to 1 in ClimaParams.

Note that I didn't add `tracer_vertical_diffusion_factor` to vert_diff_params because this will be used by EDMF as well, once we add vertical diffusion of precipitating species in EDMF.

## Content

----


src/prognostic_equations/implicit/implicit_solver.jl
```diff
@@ -759,10 +770,23 @@ function update_implicit_equation_jacobian!(A, Y, p, dtγ)
...
             ∂ᶜρq_err_∂ᶜρq = matrix[ρq_name, ρq_name]
-            @. ∂ᶜρq_err_∂ᶜρ =
-                dtγ * ᶜdiffusion_h_matrix ⋅ DiagonalMatrixRow(-(ᶜq) / ᶜρ)
-            @. ∂ᶜρq_err_∂ᶜρq +=
-                dtγ * ᶜdiffusion_h_matrix ⋅ DiagonalMatrixRow(1 / ᶜρ)
+            # ᶜdiffusion_h_matrix = ifelse(
+            #     q_name in (:q_rai, :q_sno),
+            #     ᶜdiffusion_h_matrix_scaled,
+            #     ᶜdiffusion_h_matrix,
+            # )
```
szy21, line 777: There seems to be some random allocation change here. Testing it now.

szy21:
> Thank you!
> 
> Is the additional allocation caused by adding `ᶜdiffusion_h_matrix_scaled` ?

No I think it is the ifelse thing I commented out. Let me try one more thing.

src/prognostic_equations/implicit/implicit_solver.jl
```diff
@@ -759,10 +772,27 @@ function update_implicit_equation_jacobian!(A, Y, p, dtγ)
...
+                q_name in (:q_rai, :q_sno),
+                ᶜdiffusion_h_matrix_scaled,
+                ᶜdiffusion_h_matrix,
+            )
             @. ∂ᶜρq_err_∂ᶜρ =
-                dtγ * ᶜdiffusion_h_matrix ⋅ DiagonalMatrixRow(-(ᶜq) / ᶜρ)
+                dtγ * ᶜtridiagonal_matrix_scalar ⋅ DiagonalMatrixRow(-(ᶜq) / ᶜρ)
             @. ∂ᶜρq_err_∂ᶜρq +=
-                dtγ * ᶜdiffusion_h_matrix ⋅ DiagonalMatrixRow(1 / ᶜρ)
+                dtγ * ᶜtridiagonal_matrix_scalar ⋅ DiagonalMatrixRow(1 / ᶜρ)
```
szy21, line 783: @trontrytel Both of these two versions work (without an increase in allocation). Compared to the commented out version, this one has fewer lines, but needs one more scratch variable. Which one do you prefer?

trontrytel: I don't have a strong preference. I would vote for the currently uncommented one.

PR #3704, opened 2025-03-16
set dt_save_to_sol default to Inf

szy21:
## Purpose 

`dt_save_to_sol` was default to 1 day, which causes memory errors when running long simulations (e.g. 10 year of 200km dry baroclinic wave). This PR sets the default to Inf. Also removes `dt_save_to_sol: Inf` and `dt_save_state_to_disk: Inf` in any config as those are the default.

## Content

----


Sbozzolo:
I removed this from the queue, I want to have a quick look to see if we can add a simple system to just the initial and final iteration (and maybe remove this feature completely)

PR #3708, opened 2025-03-18
make entrainment and detrainment implicit

szy21:
## Purpose 

Adds an option for treating entrainment and detrainment of updraft implicitly. Also removes an unused temporary variable.

## Content

----

### DIFF
--- a/src/prognostic_equations/implicit/implicit_solver.jl
+++ b/src/prognostic_equations/implicit/implicit_solver.jl
@@ -99,6 +99,7 @@ struct ImplicitEquationJacobian{
     F2 <: DerivativeFlag,
     F3 <: DerivativeFlag,
     F4 <: DerivativeFlag,
+    F5 <: DerivativeFlag,
     T <: Fields.FieldVector,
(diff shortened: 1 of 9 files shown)


src/prognostic_equations/implicit/implicit_solver.jl
```diff
@@ -995,6 +1002,21 @@ function update_implicit_equation_jacobian!(A, Y, p, dtγ)
                     dtγ * ᶠtridiagonal_matrix_c3 ⋅
                     DiagonalMatrixRow(adjoint(CT3(Y.f.sgsʲs.:(1).u₃))) - (I_u₃,)
             end
+
+            # entrainment and detrainment
+            (; ᶜentrʲs, ᶜdetrʲs, ᶜturb_entrʲs) = p
```
costachris, line 1007: could you add a comment mentioning the assumption (i.e w/z, dw/dz constant), so it's clear if we change entrainment later

costachris (approved):

PR #3710, opened 2025-03-18
Add option to restart from a non zero t and bump to 0.29.1

Sbozzolo:
ClimaCoupler handles restarts in its own way, but it was not possible to start a simulation from a `t_start` that is not zero without using the ClimaAtmos restart system. This commit adds this option and moves `t_start` inside `sim_info` for symmetry with `t_end` and `dt`.

### DIFF
--- a/src/solver/type_getters.jl
+++ b/src/solver/type_getters.jl
@@ -604,12 +604,19 @@ function get_sim_info(config::AtmosConfig)
     isnothing(restart_file) ||
         @info "Restarting simulation from file $restart_file"
     epoch = DateTime(parsed_args["start_date"], dateformat"yyyymmdd")
+    t_start_int = time_to_seconds(parsed_args["t_start"])
+    if !isnothing(restart_file) && t_start_int != 0
+        @warn "Non zero `t_start` passed with a restarting simulation. The provided `t_start` will be ignored."
+    end
     if parsed_args["use_itime"]
         dt = ITime(time_to_seconds(parsed_args["dt"]))
+        t_start = ITime(time_to_seconds(parsed_args["t_start"]), epoch = epoch)
         t_end = ITime(time_to_seconds(parsed_args["t_end"]), epoch = epoch)
-        (dt, t_end) = promote(dt, t_end)
+        # ITime(0) is added for backward compatibility (since t_start used to always be 0)
+        (dt, t_start, t_end, _) = promote(dt, t_start, t_end, ITime(0))
     else
         dt = FT(time_to_seconds(parsed_args["dt"]))
+        t_start = FT(time_to_seconds(parsed_args["t_start"]))
         t_end = FT(time_to_seconds(parsed_args["t_end"]))
     end
     sim = (;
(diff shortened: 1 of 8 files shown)


Sbozzolo:
Passing ClimaCoupler PR: https://github.com/CliMA/ClimaCoupler.jl/pull/1226

src/solver/type_getters.jl
```diff
@@ -600,12 +600,19 @@ function get_sim_info(config::AtmosConfig)
     isnothing(restart_file) ||
         @info "Restarting simulation from file $restart_file"
     epoch = DateTime(parsed_args["start_date"], dateformat"yyyymmdd")
+    t_start_int = time_to_seconds(parsed_args["t_start"])
+    if !isnothing(restart_file) && t_start_int != 0
+        @warn "Non zero `t_start` passed with a restarting simulation. The provided `t_start` will be ignored."
```
juliasloan25, line 605: I thought this PR was enabling using non-zero `t_start` in a restarted simulation, but this suggests it's still not supported

src/solver/type_getters.jl
```diff
@@ -600,12 +600,19 @@ function get_sim_info(config::AtmosConfig)
...
+    end
     if parsed_args["use_itime"]
         dt = ITime(time_to_seconds(parsed_args["dt"]))
+        t_start = ITime(time_to_seconds(parsed_args["t_start"]), epoch = epoch)
         t_end = ITime(time_to_seconds(parsed_args["t_end"]), epoch = epoch)
-        (dt, t_end) = promote(dt, t_end)
+        # ITime(0) is added for backward compatibility (since t_start used to always be 0)
```
juliasloan25, line 611: How does this provide backward compatibility? If `t_start` is 0, as it was before, wouldn't this be redundant?

juliasloan25 (approved):
I just had a couple of questions

src/solver/type_getters.jl
Sbozzolo: It enables non-zero `t_start` in a _manually_ restarted simulation (as we do in the coupler). If you restart from the Atmos checkpoints, `t_start` is already correctly set (because it is read from file), so there's no need to specify a different one.

src/solver/type_getters.jl
Sbozzolo: This is only to ensure that types are consistent. If you see, the output of `ITime(0)` is thrown away (but it contributes to the promotion in the line before)

Sbozzolo: Yes, because `ITime` is a parametrized type, so two `ITimes` might not be the same

src/solver/type_getters.jl
Sbozzolo: Added some additional documntation about this

szy21 (approved):
Thanks! What would be a use case for using a non-zero t_start when not restarting a simulation?

szy21:
And should we expect MSE changes?

Sbozzolo:
> And should we expect MSE changes?

No, I am trying to understand why that's happening.

> Thanks! What would be a use case for using a non-zero t_start when not restarting a simulation?

The main use case is this: https://github.com/CliMA/ClimaCoupler.jl/pull/1226

Issue #3712, opened 2025-03-19
Remove source term for mass from diffusion of cloud condensate and precipitation


szy21:
This includes both vertical diffusion and hyperdiffusion.

PR #3714, opened 2025-03-19
fix boundary condition for smagorinsky lilly

szy21:
## Purpose 

I didn't realize smagorinsky lilly diffusion has a boundary condition and missed this in #3670 .

## Content

----


akshaysridhar (commented):
These are zeroed out here such that 
```https://github.com/CliMA/ClimaAtmos.jl/blob/90d415be4253c3eadde2ac59c0be283382d7a705/src/prognostic_equations/remaining_tendency.jl#L112``` 
is the only place where the surface fluxes are then introduced (explicitly)?

szy21:
> https://github.com/CliMA/ClimaAtmos.jl/blob/90d415be4253c3eadde2ac59c0be283382d7a705/src/prognostic_equations/remaining_tendency.jl

Yes

akshaysridhar:
Could you please also add a note on the `surface_flux_tendency` explicit treatment in the docs? I don't see this here (maybe I'm missing its inclusion on a different page) https://clima.github.io/ClimaAtmos.jl/dev/equations/

szy21:
> Could you please also add a note on the `surface_flux_tendency` explicit treatment in the docs? I don't see this here (maybe I'm missing its inclusion on a different page) https://clima.github.io/ClimaAtmos.jl/dev/equations/

We don't have any surface flux or vertical diffusion in the doc now (see the todo here: https://github.com/CliMA/ClimaAtmos.jl/blob/90d415be4253c3eadde2ac59c0be283382d7a705/docs/src/equations.md?plain=1#L38), but I can add them when I have more time (or feel free to add them!).

akshaysridhar:
> > Could you please also add a note on the `surface_flux_tendency` explicit treatment in the docs? I don't see this here (maybe I'm missing its inclusion on a different page) https://clima.github.io/ClimaAtmos.jl/dev/equations/
> 
> We don't have any surface flux or vertical diffusion in the doc now (see the todo here:
> 
> https://github.com/CliMA/ClimaAtmos.jl/blob/90d415be4253c3eadde2ac59c0be283382d7a705/docs/src/equations.md?plain=1#L38
> ), but I can add them when I have more time (or feel free to add them!).

Thanks ; I'll add some notes along with fixes for #3712

PR #3705, opened 2025-03-17
Prognostic EDMF + 1M

trontrytel:
@szy21 - This is the code that I have right now that tries to add precipitation to edmf updrafts.

Included:
- Remove `xfail` from EDMF +1M Rico SCM case. Make that case use explicit diffusion for now. Add cloud condensate non-equil and precipitation in the updrafts/environment to diagnostics and CI plots.
- Add additional cached sgs variables:
       - environmental cloud condensate and precipitation: `ᶜq_liq⁰`, `ᶜq_ice⁰`, `ᶜq_rai⁰`, `ᶜq_sno⁰`
       - cloud and precipitation sources in the updrafts and environment: `ᶜSqₗᵖʲs`, `ᶜSqᵢᵖʲs`, `ᶜSqᵣᵖʲs`, `ᶜSqₛᵖʲs`, `ᶜSqₗᵖ⁰`, `ᶜSqᵢᵖ⁰`, `ᶜSqᵣᵖ⁰`, `ᶜSqₛᵖ⁰`
- Create a `TD.PhaseNonEquil_phq` state in the updraft and environment
- Set the `ᶜq_liqʲ_int_val` etc to the grid mean value. (Hacky?)
- Compute `compute_precipitation_sources!`, `compute_precipitation_sinks!`, `cloud_sources` in the updraft and environment.
-  Add `q_tot`, `q_liq`, `q_ice`, `q_rai`, `q_sno` to prognostic EDMF initial conditions.
- Add contributions from updraft and environment to the grid mean cloud and precipitation.
- Add tendencies to `Yₜ.c.sgsʲs.:($$j).q_liq`, `Yₜ.c.sgsʲs.:($$j).q_ice`, `Yₜ.c.sgsʲs.:($$j).q_rai`, `Yₜ.c.sgsʲs.:($$j).q_sno` from:
        - horizontal advection with the mean flow
        - vertical advection with updrafts
        - entrainment/detrainment
        - cloud and precipitation formation
        - diffusion and hyperdiffusion (with the corresponding multiplier for precip tracers to be able to switch it off)
- Add ED and MF contributions to `Yₜ.c.ρq_liq`, `Yₜ.c.ρq_ice`, `Yₜ.c.ρq_rai`, `Yₜ.c.ρq_sno`

Things left to do in the next PR:
- Add cloud and precipitation sedimentation
- Working TRMM SCM
[...]

PR #3716, opened 2025-03-19
Surface cloud sedimentation included in precipitation fluxes

trontrytel:
This PR:
- adds the contribution from sedimenting cloud condensate to `surface_rain_flux` and `surface_snow_flux` for the 1M microphysics. I would prefer to keep the names the same even though now they contain both precipitation and cloud fluxes. It's a slow process to change it in both Coupler and Atmos.


trontrytel:
I managed to remove 13 lines from https://github.com/CliMA/ClimaAtmos.jl/pull/3705

szy21 (approved):

PR #3709, opened 2025-03-18
Update dependencies

charleskawczynski:
This PR updates dependencies, including adapt and cuda. cc @glwagner.

PR #3719, opened 2025-03-20
Update buildkite manifest

imreddyTeja:
## Purpose 
Updates the dependencies following release of ClimaCore 0.14.27

## Content

----


imreddyTeja:
This duplicates #3709 , which will also update the ClimaCore dependency when the CUDA dependency is updated.

PR #3725, opened 2025-03-20
Print at exit only on buildkite

charleskawczynski:


glwagner (approved):

charleskawczynski:
The failure is unrelated to the change

PR #3732, opened 2025-03-21
Fix Jacobian for zero_velocity_tendency

dennisYatunin:
## Purpose 

We are currently using `zero_velocity_tendency!` in the EDMFX (prospct) advection test's tendency, but we are not making any corresponding update to the Jacobian. This PR adds a `zero_velocity_jacobian!` function that zeros out the corresponding Jacobian blocks.

----

### DIFF
--- a/src/prognostic_equations/zero_velocity.jl
+++ b/src/prognostic_equations/zero_velocity.jl
@@ -1,20 +1,49 @@
+import ClimaCore: MatrixFields
+import LinearAlgebra: UniformScaling
+
 ###
 ### edmfx advection test 
 ###
 
+# Turn off all momentum tendencies in the advection test.
 function zero_velocity_tendency!(Yₜ, Y, p, t)
-    # turn off all monmentum tendencies in the advection test
-    if p.atmos.advection_test
-        FT = eltype(Y)
-        n = n_mass_flux_subdomains(p.atmos.turbconv_model)
-
-        @. Yₜ.c.uₕ = C12(FT(0), FT(0))
-        @. Yₜ.f.u₃ = Geometry.Covariant3Vector(FT(0))
-        if p.atmos.turbconv_model isa PrognosticEDMFX
-            for j in 1:n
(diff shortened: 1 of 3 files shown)


szy21 (approved):

src/prognostic_equations/zero_velocity.jl
```diff
@@ -1,20 +1,49 @@
...
+                end
             end
         end
     end
-    return nothing
+end
+
+function set_identity_matrix_entry!(matrix_entry, row_name, col_name)
+    identity_matrix_entry_value = if row_name == col_name
+        # TODO: Add a method for one(::Axis2Tensor) to simplify this.
```
charleskawczynski, line 41: Is the suggestion to define `one(::Type{T}) where {T <: Axis2Tensor}`? or `one(::T) where {T <: Axis2Tensor}`?

charleskawczynski: I thought we had it. Ah, it looks like we currently only support `AxisVector`s:
```julia
Base.one(::T) where {T <: Geometry.AxisTensor} = one(T)
Base.one(::Type{T}) where {T′, A, S, T <: Geometry.AxisTensor{T′, 1, A, S}} =
    T(axes(T), S(one(T′)))
```

dennisYatunin: Following other definitions of `zero` and `one` in ClimaCore, I would probably define this as
```
one(::T) where {T <: Axis2Tensor} = one(T)
one(::Type{T}) where {T <: Axis2Tensor} = ...
```
I've already defined `one(::BandMatrixRow)`, so adding this will simplify the case for `row_name == col_name` to `-one(eltype(matrix_entry))`

dennisYatunin: It would also probably be good to generalize the method of `one` for `AxisVector`s to
```
one(::Type{T}) where {A, S, T <: Geometry.AxisTensor{<:Any, 1, A, S} = T(A, ones(S))
```
The version you gave above only works for single-component vectors.

PR #3739, opened 2025-03-24
Adjust CI to avoid timeouts

dennisYatunin:
## Purpose 

We have recently been seeing timeout errors for the following jobs: CI on Ubuntu with Julia 1.11, CI on Windows with Julia 1.11, and Lookahead on Ubuntu with the nightly release of Julia. This PR bumps the timeout limit for CI with Julia 1.11 to 60 minutes (while keeping the timeout for 1.10 at 45 minutes), and it switches the OS for the Lookahead job from Ubuntu to MacOS (we only need to test one OS for this job, so we can just use the one with the shortest compilation times).

----


szy21 (approved):
one of the tests takes > 50min, so this may make ci a bit longer, but I think it's ok. Could we look into how to make it shorter though (and why is there such a large difference between different operating systems)?

Sbozzolo:
I'd say that we can remove the lookahead job. It's not the most informative.

> one of the tests takes > 50min, so this may make ci a bit longer, but I think it's ok. Could we look into how to make it shorter though (and why is there such a large difference between different operating systems)?

Yes, it'd be good to look into this

Sbozzolo (approved):

dennisYatunin:
Ok, I've removed the lookahead job altogether (it was also timing out on MacOS so we would have had to increase the timeout limit past 60 minutes if we wanted to keep it). Unfortunately, it looks like there's also a failure in the downgrade job on Julia 1.10. That job also recently had another error that @nefrathenrici fixed, so maybe he'll know what's going on there.

PR #3741, opened 2025-03-25
Separate buoyancy and drag terms in edmf pressure closure

szy21:
## Purpose 

Right now we only take the Jacobians of the drag term in the edmf pressure closure, not the buoyancy (virtual mass) term. This PR separates the two terms in the pressure closure, so we can apply the tendencies separately and control whether to treat them implicitly. To be consistent with the Jacobians, we treat the tendency from the buoyancy term explicitly but the tendency from the drag term implicitly.

## Content

----

### DIFF
--- a/src/prognostic_equations/edmfx_closures.jl
+++ b/src/prognostic_equations/edmfx_closures.jl
@@ -52,46 +52,69 @@ function surface_flux_tke(
 end
 
 """
-   Return the nonhydrostatic pressure drag for updrafts [m/s2 * m]
+   Return the virtual mass term of the pressure closure for updrafts [m/s2 * m]
+
+   Inputs (everything defined on cell faces):
+   - params - set with model parameters
+   - ᶠbuoyʲ - covariant3 or contravariant3 updraft buoyancy
+"""
+function ᶠupdraft_nh_pressure_buoyancy(params, ᶠbuoyʲ)
+    turbconv_params = CAP.turbconv_params(params)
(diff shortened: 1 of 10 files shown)


szy21:
Hi Gemini could you resummarize the PR?

src/prognostic_equations/edmfx_closures.jl
```diff
@@ -52,46 +52,69 @@ function surface_flux_tke(
...
-
-    # Independence of aspect ratio hardcoded: α₂_asp_ratio² = FT(0)
-
     H_up_min = CAP.min_updraft_top(turbconv_params)
 
+    # Independence of aspect ratio hardcoded: α₂_asp_ratio² = FT(0)
     # We also used to have advection term here: α_a * w_up * div_w_up
-    return α_b * ᶠbuoyʲ +
-           α_d * (ᶠu3ʲ - ᶠu3⁰) * CC.Geometry._norm(ᶠu3ʲ - ᶠu3⁰, ᶠlg) /
+    return α_d * (ᶠu3ʲ - ᶠu3⁰) * CC.Geometry._norm(ᶠu3ʲ - ᶠu3⁰, ᶠlg) /
```
costachris, line 87: can you change `plume_height` to `scale_height` since that's what we're using now? Or otherwise make a note in the function signature that we're using one in place of another.

costachris (approved):

szy21:
Thanks! I added the comment and changed `plume_height` to `scale_height`.

PR #3743, opened 2025-03-25
calibratable number_conc for r_eff

trontrytel:
This PR adds a function to diagnose the number concentration of liquid water clouds based on aerosol loading (dust, sea salt and sulfate) and water. The diagnosed number concentration is used to compute the cloud effective radius for radiation. 

The default aerosol and `q_liq` coefficients are all zeros, so when no aerosol is present it should reduce to the currently prescribed number concentration. 

@nefrathenrici -  don't know why, but I was not able to pass the `aerosol_ml_parameters` as a whole struct. I was getting `ERROR: LoadError: ArgumentError: broadcasting over dictionaries and `NamedTuple`s is reserved`. I'm probably creating the parameter struct in Atmos wrong somehow. Maybe you can take a look?


src/callbacks/callbacks.jl
```diff
@@ -206,13 +206,64 @@ NVTX.@annotate function rrtmgp_model_callback!(integrator)
...
+                [:SSLT01, :SSLT02, :SSLT03, :SSLT04, :SSLT05]
+                if prescribed_aerosol_name in
+                   propertynames(p.tracers.prescribed_aerosols_field)
+                    aerosol_field = getproperty(
+                        p.tracers.prescribed_aerosols_field,
+                        prescribed_aerosol_name,
+                    )
+                    @. seasalt_aero_conc += aerosol_field
+                end
+            end
```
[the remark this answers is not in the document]
szy21: You may not have `prescribed_aerosols_field` in p.tracers, so this could cause a problem. You could either add another check `if :prescribed_aerosols_field in propertynames(p.tracers)`, or add a different flag for whether to use `ml_N_cloud_liquid_droplets` or not, and check that you have aerosols when this flag is true. It's probably easier to just add another check.

src/callbacks/callbacks.jl
```diff
@@ -206,13 +206,64 @@ NVTX.@annotate function rrtmgp_model_callback!(integrator)
...
+                    @. dust_aero_conc += aerosol_field
+                end
+            end
+
+            SO4_aero_conc = p.scratch.ᶜtemp_scalar_3
+            @. SO4_aero_conc = 0
+            if :SO4 in propertynames(p.tracers.prescribed_aerosols_field)
+                aerosol_field =
+                    getproperty(p.tracers.prescribed_aerosols_field, :SO4)
+                @. SO4_aero_conc += aerosol_field
```
szy21, line 243:
```suggestion
                @. SO4_aero_conc =
                    getproperty(p.tracers.prescribed_aerosols_field, :SO4)
```

src/parameterized_tendencies/microphysics/microphysics_wrappers.jl
```diff
@@ -34,6 +34,30 @@ function limit(q, dt, n::Int)
...
+)
+    # We can also add w, T, RH, w' ...
+    # Also consider lookind only at around cloud base height
+    FT = eltype(N₀)
+    return N₀ +
+           α_dust * log(max(c_dust, eps(FT)) / c₀_dust) +
+           α_seasalt * log(max(c_seasalt, eps(FT)) / c₀_seasalt) +
+           α_SO4 * log(max(c_SO4, eps(FT)) / c₀_SO4) +
+           α_q_liq * log(max(q_liq, eps(FT)))
+end
```
nefrathenrici, line 59:
```suggestion
function ml_N_cloud_liquid_droplets(
    cmc,
    c_dust,
    c_seasalt,
    c_SO4,
    q_liq,
)
    (; α_dust,
    α_seasalt,
    α_SO4,
    α_q_liq,
    c₀_dust,
    c₀_seasalt,
    c₀_SO4,) = cmc.aml
    N₀ = cmc.N_cloud_liquid_droplets
    # We can also add w, T, RH, w' ...
    # Also consider lookind only at around cloud base height
    FT = eltype(N₀)
    return N₀ +
           α_dust * log(max(c_dust, eps(FT)) / c₀_dust) +
           α_seasalt * log(max(c_seasalt, eps(FT)) / c₀_seasalt) +
           α_SO4 * log(max(c_SO4, eps(FT)) / c₀_SO4) +
           α_q_liq * log(max(q_liq, eps(FT)))
end
```

src/callbacks/callbacks.jl
```diff
@@ -206,13 +206,64 @@ NVTX.@annotate function rrtmgp_model_callback!(integrator)
...
+                        cmc.aml.c₀_dust,
+                        cmc.aml.c₀_seasalt,
+                        cmc.aml.c₀_SO4,
+                        cmc.N_cloud_liquid_droplets,
+                        dust_aero_conc,
+                        seasalt_aero_conc,
+                        SO4_aero_conc,
+                        cloud_liquid_water_content /
+                        max(eps(FT), cloud_fraction),
+                    ),
```
nefrathenrici, line 266:
```suggestion
                        ml_N_cloud_liquid_droplets((cmc,),
                            dust_aero_conc,
                            seasalt_aero_conc,
                            SO4_aero_conc,
                            cloud_liquid_water_content /
                            max(eps(FT), cloud_fraction),
                        ),
```

nefrathenrici (approved):
Thank you!

src/parameterized_tendencies/microphysics/microphysics_wrappers.jl
szy21: I don't have a good sense if we should use log q_liq here, and if we do, whether we should divide it by a reference q_liq as well. Maybe we can start with this and see what we get. Also, will negative N cause an issue? Do you want to clip it?

src/callbacks/callbacks.jl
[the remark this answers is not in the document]
trontrytel: Thanks! You are right. I added another check and re-wrote the loops to be more concise

src/callbacks/callbacks.jl
```diff
@@ -206,13 +206,64 @@ NVTX.@annotate function rrtmgp_model_callback!(integrator)
...
+                [:DST01, :DST02, :DST03, :DST04, :DST05]
+                if prescribed_aerosol_name in
+                   propertynames(p.tracers.prescribed_aerosols_field)
+                    aerosol_field = getproperty(
+                        p.tracers.prescribed_aerosols_field,
+                        prescribed_aerosol_name,
+                    )
+                    @. dust_aero_conc += aerosol_field
+                end
+            end
```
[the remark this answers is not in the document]
trontrytel: Rewritten

src/parameterized_tendencies/microphysics/microphysics_wrappers.jl
trontrytel: If N is negative the CloudMicrophysics will return zero for effective radius. Which will be clipped in the radiation lookup table to the smallest available size. So I think we should be safe(ish)

src/callbacks/callbacks.jl
trontrytel: Thank you!

PR #3745, opened 2025-03-25
remove derivative with respect to grid mean rho in edmf

szy21:
## Purpose 

Removes the derivative of updraft u3 and mse tendencies with respect to grid mean rho. This will make the implicit solver simpler after #3746.

## Content

----

### DIFF
--- a/reproducibility_tests/ref_counter.jl
+++ b/reproducibility_tests/ref_counter.jl
@@ -1,4 +1,4 @@
-228
+229
 
 # **README**
 #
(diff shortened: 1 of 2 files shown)


costachris (approved):

costachris:
These changes appear to produce stable 3-month aquaplanet simulations with reasonable updrafts at dt=50 and (h_elem: 16; 200 km). 

[waup_time_-1.pdf](https://github.com/user-attachments/files/19490569/waup_time_-1.pdf)
[arup_time_-1.pdf](https://github.com/user-attachments/files/19490571/arup_time_-1.pdf)

PR #3751, opened 2025-03-27
Remove unused dependencies.

sriharshakandala:
## Purpose 
Remove unused dependencies.

----

## Summary by CodeRabbit

- **Chores**
	- Streamlined the project configuration by removing unused external dependencies and simplifying compatibility requirements.
- **Refactor**
	- Aligned internal numerical integration routines with updated module references for improved consistency and performance.
	- Updated the method for generating Gauss-Hermite quadrature points to utilize a new source.


Sbozzolo:
Are you going to remove DiffEqBase from ClimaTimeSteppers next?

ClimaCore depends directly on GaussQuadrature, but GaussQuadrature has not been updated in 7 years, so maybe the correct thing to do is update ClimaCore to use FastGaussQuadrature and have it export the relevant functions (instead of accessing the package this way)

sriharshakandala:
> Are you going to remove DiffEqBase from ClimaTimeSteppers next?
> 
> ClimaCore depends directly on GaussQuadrature, but GaussQuadrature has not been updated in 7 years, so maybe the correct thing to do is update ClimaCore to use FastGaussQuadrature and have it export the relevant functions (instead of accessing the package this way)

- I can work on removing `DiffEqBase` from `ClimaTimeSteppers` in a different task. I would like to primarily focus on removing unused direct dependencies here
- `FastGaussQuadrature.jl` is missing some functionality we use from `GaussQuadrature.jl`. I have placed a feature request here: https://github.com/JuliaApproximation/FastGaussQuadrature.jl/issues/135 . For now, I would recommend using `GaussQuadrature.jl` everywhere consistently, and moving to `FastGaussQuadrature.jl` when it is ready for what we need.

akshaysridhar:
I have no strong opinions on the use of `FastGaussQuadrature` - given the maintenance status as @Sbozzolo points out, and how much time it may take to update this in ClimaCore we can make a decision on the use of `FastGaussQuadrature` in  `ClimaCore` (it seems the bickley jet DG problem is the only one affected in `ClimaCore`, if we agree to drop support for that example then we can move to `FastGaussQuadrature` in `ClimaCore`?).

Sbozzolo (approved):

PR #3755, opened 2025-03-28
change tracer_vertical_diffusion_factor to 0 in 1M jobs

szy21:
## Purpose 

Based on our discussion today, it is preferred to have no vertical diffusion on precipitation. This PR sets `tracer_vertical_diffusion_factor` in all the 1M tomls. If we are sure this is what we want in the future we can change the default in ClimaParameters.

## Content

----

### DIFF
--- a/toml/les_isdac.toml
+++ b/toml/les_isdac.toml
@@ -4,3 +4,6 @@ value = 2000.0
 [mixing_length_Prandtl_number_0]
 value = 0.33333
 type = "float"
+
+[tracer_vertical_diffusion_factor]
+value = 0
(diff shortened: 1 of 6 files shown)


akshaysridhar (approved):
Thanks: Following the TODO in L346 in `src/prognostic_equations/edmfx_sgs_flux.jl`  are these terms exercised by the `turbconv flags` in the implicit solver, and do we need to take a look at the solver?

szy21:
> Thanks: Following the TODO in L346 in `src/prognostic_equations/edmfx_sgs_flux.jl` are these terms exercised by the `turbconv flags` in the implicit solver, and do we need to take a look at the solver?

I don't think we need to change anything there. But for now, we will work on edmf + 1M explicitly first, and look at the implicit solver after that. We will deal with all the TODOs when we get to that.

PR #3760, opened 2025-04-01
specific contents jargon clarifications (to avoid humidity of snow, ice, rain, ...)

slayoo:
## Purpose 
Avoid misleading jargon featuring humidities of snow, humidities of water, etc by reserving the term "humidity" for humidity, and using specific contents for all other solid and liquid phase per-mass water contents. See related https://github.com/CliMA/CloudMicrophysics.jl/pull/540 

## To-do
- the PR corrects only docstrings and code comments, code changes for now just marked with TODO comments (likely need to be coordinated with Thermodynamics.jl PR: https://github.com/search?q=repo%3ACliMA%2FThermodynamics.jl%20specific%20humidity&type=code)

## Content
jargon changes in docs and code comments


szy21:
@trontrytel I like specific humidity as it is consistent with our paper and design doc. What do you think?

trontrytel:
I don't have a preference. We did update the docs etc in CloudMicrophysics. But I did drop the ball in ClimaAtmos, because it was a big change that was touching function names in Thermodynamics etc. Maybe we can leave it, unless Reviewer 2 agrees with @slayoo and does not like the naming convention?

PR #3767, opened 2025-04-04
GPU vel testing

trontrytel:


trontrytel:
@charleskawczynski - any 1M microphysics simulation should be ok for testing on GPU. For example `config/longrun_configs/longrun_moist_baroclinic_wave_1M.yml`

PR #3769, opened 2025-04-07
Check for equality with float(t) instead of t

ph-kev:
This fixed the bug that resulted in `NaN`s in the first time slice of any radiation-related diagnostics.

Related PR: https://github.com/CliMA/ClimaCoupler.jl/pull/1263

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved numerical comparison logic in surface property calculations to ensure consistent and reliable performance during evaluations.

### DIFF
--- a/src/cache/surface_albedo.jl
+++ b/src/cache/surface_albedo.jl
@@ -110,7 +110,7 @@ so the initial callback initialization doesn't lead to NaNs in the radiation mod
 Subsequently, the surface albedo will be updated by the coupler.
 """
 function set_surface_albedo!(Y, p, t, ::CouplerAlbedo)
-    if t == 0
+    if float(t) == 0
         FT = eltype(Y)
         # set initial insolation initial conditions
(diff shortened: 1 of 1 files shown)


src/cache/surface_albedo.jl
```diff
@@ -110,7 +110,7 @@ so the initial callback initialization doesn't lead to NaNs in the radiation mod
 Subsequently, the surface albedo will be updated by the coupler.
 """
 function set_surface_albedo!(Y, p, t, ::CouplerAlbedo)
```
Sbozzolo, line 112: There's another copy of this function in ClimaCoupler too.

This is entire function a hack and we should remove it. 

This function is here because there's no way to not set the albedo in ClimaAtmos, essentially.

Sbozzolo:
> Do we want to get this in? Or do we want to rewrite how albedo is set?

I think we want to get this in, and fix the albedo in the future

PR #3773, opened 2025-04-15
Remove unreferenced yaml parameter

charleskawczynski:
This PR removes an unreferenced yaml parameter.


Julians42:
Yes we can remove this for now. Tapio at one point requested a feature whereby you could run a model with the stretch grid from a larger model but only up to a certain height - e.g., run the stretch grid from a 70km model but only up to 40km, mostly for computational efficiency. I drafted this [here](https://github.com/CliMA/ClimaAtmos.jl/tree/j/truncation) but gave up on merging it a while ago.

Julians42 (approved):

PR #3774, opened 2025-04-15
Use default_num_points when available

Sbozzolo:
`ClimaDiagnostics` introduced a function to compute the default number of points for the NetCDFWriter. This commit uses this new function when available.

### DIFF
--- a/src/callbacks/get_callbacks.jl
+++ b/src/callbacks/get_callbacks.jl
@@ -33,26 +33,9 @@ function get_diagnostics(parsed_args, atmos_model, Y, p, sim_info, output_dir)
         num_netcdf_points =
             tuple(parsed_args["netcdf_interpolation_num_points"]...)
     else
-        # Estimate the number of points we need to cover the entire domain
-        # ncolumns is the number of local columns
-        tot_num_columns =
-            ClimaComms.nprocs(context) * Fields.ncolumns(axes(Y.c))
-        if parsed_args["config"] == "plane"
-            num1, num2 = tot_num_columns, 0
-        elseif parsed_args["config"] == "sphere"
-            num2 = round(Int, sqrt(tot_num_columns / 2))
-            num1 = 2num2
-        elseif parsed_args["config"] == "box"
-            num2 = round(Int, sqrt(tot_num_columns))
-            num1 = num2
-        elseif parsed_args["config"] == "column"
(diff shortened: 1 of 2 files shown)


nefrathenrici:
@Sbozzolo Is there a reason this hasn't been reviewed/merged?

Sbozzolo:
> @Sbozzolo Is there a reason this hasn't been reviewed/merged?

I ran into this problem with the macOS test and I couldn't debug it. I'll rebase and see if now this works

Sbozzolo:
This uncovered an edge case in ClimaAnalysis. It got fixed in https://github.com/CliMA/ClimaAnalysis.jl/commit/b014f0a502677f6e394843a9f070b6fbd0e4ac37
but it will need a new release

szy21:
@ph-kev Could you update and merge this PR?

ph-kev:
> Thanks! Could you add a comment explaining how this function can be used with `NetCDFWriter` or the [parsed arg](https://github.com/CliMA/ClimaAtmos.jl/blob/main/config/default_configs/default_config.yml#L304) `netcdf_interpolation_num_points`?

I am a little confused on what should be added for clarity. I think that reading the surrounding context would tell you how those two things. For example, after `num_netcdf_points` is initialized, it is passed as a keyword argument in `NetCDFWriter` and `netcdf_interpolation_num_points` is handled first before calling `default_netcdf_points`.

nefrathenrici:
> > Thanks! Could you add a comment explaining how this function can be used with `NetCDFWriter` or the [parsed arg](https://github.com/CliMA/ClimaAtmos.jl/blob/main/config/default_configs/default_config.yml#L304) `netcdf_interpolation_num_points`?
> 
> I am a little confused on what should be added for clarity. I think that reading the surrounding context would tell you how those two things. For example, after `num_netcdf_points` is initialized, it is passed as a keyword argument in `NetCDFWriter` and `netcdf_interpolation_num_points` is handled first before calling `default_netcdf_points`.

I think a comment or docstring explaining what this function is used for so that it's immediately clear to future developers, since currently you would need to search the codebase for the function name to see its purpose. It's not strictly necessary but I think it would be helpful.

Something like: "From the given space, calculate the number of diagnostic grid points. Used to set up the default netcdf diagnostic grid if not specified by the user."

PR #3788, opened 2025-04-29
Unconditionally update insolation tuple

Sbozzolo:
I am not sure why insolation was set to not be updated with `CouplerAlbedo`, but I don't think this is currently necessary.

The function `set_insolation_variables!` was manually called by ClimaCoupler as part of the coupling loop

https://github.com/CliMA/ClimaCoupler.jl/blob/a3b32d169137f7dad2edf33fd2f5e29ebd6d5356/experiments/ClimaEarth/components/atmosphere/climaatmos.jl#L394

My guess is that this is because we wanted to update it more frequently than a radiation timestep.

Maybe the point was to obtain `cos_zenith_angle`?

In this commit, I unconditionally update the insolation_tuple. I will also change ClimaCoupler to update it as it used to do, but I think that won't be needed after we remove `cos_zenith_angle` as an exchange field.

### DIFF
--- a/src/callbacks/callbacks.jl
+++ b/src/callbacks/callbacks.jl
@@ -223,10 +223,7 @@ NVTX.@annotate function rrtmgp_model_callback!(integrator)
         end
     end
 
-    if p.atmos.insolation isa IdealizedInsolation ||
-       !(p.atmos.surface_albedo isa CouplerAlbedo)
-        set_insolation_variables!(Y, p, t, p.atmos.insolation)
-    end
+    set_insolation_variables!(Y, p, t, p.atmos.insolation)
 
     if radiation_mode isa RRTMGPI.AllSkyRadiation ||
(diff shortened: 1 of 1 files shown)


Sbozzolo:
> It seems related to this block here wherein `surface_albedo` is set by the coupler.
> 
> https://github.com/CliMA/ClimaAtmos.jl/blob/acb0e0d4224207a1735e61336226e314d06477bc/src/cache/surface_albedo.jl#L106
> 
> @Sbozzolo The suggested change in this PR seems reasonable, but I'd like to make sure I understand the conditional logic that was set up in the original code!

The original code essentially handed-off the update of the insolation tuple to ClimaCoupler (when run as part of ClimaCoupler). We can still do that, but there is no harm in updating also when we call rrtgmp.

I also fixed this in ClimaCoupler:
https://github.com/CliMA/ClimaCoupler.jl/pull/1312

akshaysridhar (approved):

PR #3789, opened 2025-04-30
adding noneq implicit flag

oalcabes:
Add non-equilibrium cloud formation sources implicit flag (with no Jacobian blocks for now)

### DIFF
--- a/config/default_configs/default_config.yml
+++ b/config/default_configs/default_config.yml
@@ -202,6 +202,9 @@ precip_model:
 cloud_model:
   help: "Cloud model [`grid_scale`, `quadrature` (default), `quadrature_sgs`]"
   value: "quadrature"
+implicit_noneq_cloud_formation:
+  help: "Whether to treat the nonequilibrium cloud condensate tendency implicitly [`true`, `false` (default)]"
+  value: false
 perf_summary:
   help: "Flag for collecting performance summary information"
   value: false
--- a/src/solver/type_getters.jl
+++ b/src/solver/type_getters.jl
@@ -23,6 +23,11 @@ function get_atmos(config::AtmosConfig, params)
     moisture_model = get_moisture_model(parsed_args)
     precip_model = get_precipitation_model(parsed_args)
     cloud_model = get_cloud_model(parsed_args)
+
+    implicit_noneq_cloud_formation =
+        parsed_args["implicit_noneq_cloud_formation"]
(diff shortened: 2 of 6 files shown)


config/default_configs/default_config.yml
```diff
@@ -202,6 +202,9 @@ precip_model:
 cloud_model:
   help: "Cloud model [`grid_scale`, `quadrature` (default), `quadrature_sgs`]"
   value: "quadrature"
+implicit_noneq_moist:
+  help: "Whether to treat the nonequilibrium cloud condensate tendency implicitly [`true`, `false`]"
```
trontrytel, line 206:
```suggestion
  help: "Whether to treat the nonequilibrium cloud condensate tendency implicitly [`true`, `false` (default)]"
```

src/solver/type_getters.jl
```diff
@@ -93,6 +97,7 @@ function get_atmos(config::AtmosConfig, params)
         edmfx_model,
         precip_model,
         cloud_model,
+        noneq_mode = implicit_noneq_moist ? Implicit() : Explicit(),
```
trontrytel, line 100: `noneq` may refer to many things. Maybe something more precise like `cloud_formation_sources`?

config/default_configs/default_config.yml
```diff
@@ -202,6 +202,9 @@ precip_model:
 cloud_model:
   help: "Cloud model [`grid_scale`, `quadrature` (default), `quadrature_sgs`]"
   value: "quadrature"
+implicit_noneq_moist:
```
trontrytel, line 205: I would suggest something more descriptive, like maybe `implicit_noneq_cloud_formation` ?

trontrytel (approved):
LGTM! I would just suggest to change the naming of the flag and config file choice to something more tied to cloud formation sources. Words like nonequilibrium or moisture could be too broad. 

Also, it is meant to not add any Jacobian terms? Just to move the tendency and assume no changes to the Jacobian are needed?

oalcabes:
> LGTM! I would just suggest to change the naming of the flag and config file choice to something more tied to cloud formation sources. Words like nonequilibrium or moisture could be too broad.
> 
> Also, it is meant to not add any Jacobian terms? Just to move the tendency and assume no changes to the Jacobian are needed?

Thanks! Yes, I wanted to add a version without any Jacobian terms to start & keep it simple. Great suggestions -- I'll update.

PR #3792, opened 2025-05-05
Make nogw_callback no-op when there's no GW

Sbozzolo:
The code crashes when one calls `nogw_model_callback` without turning on gravity waves. This commit changes it so that it will just be a no-op.


akshaysridhar (approved):
Thanks @Sbozzolo. Good catch.

PR #3795, opened 2025-05-06
Update default config args 

akshaysridhar:
mesh_warp_type = "SLEVE"
deep_atmosphere = "true" (removes redundant config args)
Fixes some help strings in default configs and the longrun buildkite pipeline.


trontrytel (approved):
Thank you!

PR #3796, opened 2025-05-07
Restructure radiation callback

sriharshakandala:
## Purpose 
Restructure radiation callback

----


nefrathenrici (commented):
Thank you for restructuring the radiation callback, this is much more readable. I have two questions:
- Have you benchmarked the performance of this callback? I don't see anything that would particularly impact performance but it might be good to check.
- Could you provide documentation for the longer functions such as `update_cloud_properties` and `update_aerosol_concentrations`?

akshaysridhar:
> Thank you for restructuring the radiation callback, this is much more readable. I have two questions:
> 
>     * Have you benchmarked the performance of this callback? I don't see anything that would particularly impact performance but it might be good to check.
> 
>     * Could you provide documentation for the longer functions such as `update_cloud_properties` and `update_aerosol_concentrations`?

@nefrathenrici There are now some docstrings for those two functions. (Please note this is not a complete refactor just yet; we will be updating the interfaces and simplifying them through the new RRTMGP API usage over a series of PRs). I also don't think there is any evidence of serious performance regressions from the buildkite run reported above cc @sriharshakandala.

nefrathenrici (approved):
Thank you!

PR #3807, opened 2025-05-19
Use more of RRTMGP api

charleskawczynski:


trontrytel (approved):
It's nice to be able to use RRTMGP code here.

Before this is merged, could you make sure that it works with the `main` Coupler branch? Maybe even start one of the nightly amip runs? We recently had a couple of issues when changes in radiation related code were breaking things, but only at scale in amip runs in the Coupler.

PR #3810, opened 2025-05-20
Use `z_max` to compute metric scaling for RRTMGP

nefrathenrici:
## Purpose 

Fix radiation diagnostic scaling for deep atmosphere.

## Content

- Scale 3D radiation diagnostics by `((z_lev .+ planet_radius) ./ planet_radius) .^ (FT(2))`
- Scale 2D TOA radiation diagnostics by `((z_max .+ planet_radius) ./ planet_radius) .^ (FT(2))`
- Added a helper function `apply_geometric_scaling!` to apply the scaling term.

----

### DIFF
--- a/src/diagnostics/radiation_diagnostics.jl
+++ b/src/diagnostics/radiation_diagnostics.jl
@@ -1,5 +1,15 @@
 # This file is included in Diagnostics.jl
 
+"""
+    apply_geometric_scaling!(out, z, planet_radius, FT)
+
+Apply spherical shell geometric correction to radiative fluxes at radial height `z`.
+
+Helper function for scaling radiation diagnostics.
+"""
+function apply_geometric_scaling!(out, z, planet_radius, FT)
+    @. out *= ((z + planet_radius) / planet_radius)^(FT(2))
+end
 # Radiative fluxes
 
 ###
@@ -17,18 +27,25 @@ function compute_rsd!(
     time,
     radiation_mode::T,
 ) where {T <: RRTMGPI.AbstractRRTMGPMode}
+    z_lev = Fields.coordinate_field(axes(state.f)).z
+    planet_radius = CAP.planet_radius(cache.params)
+    FT = eltype(cache.params)
     if isnothing(out)
-        return copy(
+        out = copy(
             Fields.array2field(
                 cache.radiation.rrtmgp_model.face_sw_flux_dn,
                 axes(state.f),
             ),
         )
+        apply_geometric_scaling!(out, z_lev, planet_radius, FT)
+        return out
     else
         out .= Fields.array2field(
             cache.radiation.rrtmgp_model.face_sw_flux_dn,
             axes(state.f),
         )
+        @assert out !== nothing "Output field 'out' must not be `nothing` in this branch"
+        apply_geometric_scaling!(out, z_lev, planet_radius, FT)
     end
 end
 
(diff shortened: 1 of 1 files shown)


src/parameterized_tendencies/radiation/RRTMGPInterface.jl
```diff
@@ -915,8 +915,12 @@ function _RRTMGPModel(
             # Area ratio appears in denominator of RRTMGP scaling functions, 
             # we therefore pass the multiplicative inverse from ClimaAtmos to 
             # use mult ops instead of div in RRTMGP GPU kernels.
+            z_max = maximum(z_lev)
```
charleskawczynski, line 918:
```suggestion
            z_max = Spaces.z_max(axes(z_lev))
```
Can we not use this instead? calling `maximum(z_lev)` will launch an additional global reduction kernel.

nefrathenrici: Yes, thank you!

nefrathenrici: Actually, `z_lev` is just an array so `Spaces.z_max` doesn't work.

charleskawczynski: Do we not have any spaces / fields at this part in the code?

charleskawczynski: If we don't then maybe we should?

nefrathenrici: I don't think we have any space/field access since this is in `_RRTMGPModel` but we could definitely just pass a space or `z_max` in.

src/diagnostics/radiation_diagnostics.jl
```diff
@@ -1,5 +1,17 @@
 # This file is included in Diagnostics.jl
 
+"""
+    apply_geometric_scaling!(out, z, planet_radius, FT)
+
+Apply spherical shell geometric correction to radiative fluxes at radial height `z`.
+
+Helper function for scaling radiation diagnostics.
+"""
+function apply_geometric_scaling!(out, z, planet_radius, FT)
```
szy21, line 10: Would it be more clear if we make this function non-mutating (given how it is used in the diagnostics), or keep it mutating but change how it is used?

nefrathenrici: Yes, I think something like this that just mutates `out` is cleaner:
```julia

function compute_rsd!(
    out,
    state,
    cache,
    time,
    radiation_mode::T,
) where {T <: RRTMGPI.AbstractRRTMGPMode}
    z_lev = Fields.coordinate_field(axes(state.f)).z
    planet_radius = CAP.planet_radius(cache.params)
    FT = eltype(cache.params)
    if isnothing(out)
        out = copy(
            Fields.array2field(
                cache.radiation.rrtmgp_model.face_sw_flux_dn,
                axes(state.f),
            ),
        ),
        apply_geometric_scaling!(out, z_lev, planet_radius, FT)
        return out
    else
        out .= Fields.array2field(
            cache.radiation.rrtmgp_model.face_sw_flux_dn,
            axes(state.f),
        ),
        apply_geometric_scaling!(out, z_lev, planet_radius, FT)
    end
end
```

src/diagnostics/radiation_diagnostics.jl
```diff
@@ -1,5 +1,17 @@
...
+    @. out *= ((z + planet_radius) / planet_radius)^(FT(2))
+    return out
```
szy21, line 12:
```suggestion
```
(we don't need this anymore?)

src/diagnostics/radiation_diagnostics.jl
```diff
@@ -57,24 +75,25 @@ function compute_rsdt!(
...
-                Fields.array2field(
-                    cache.radiation.rrtmgp_model.face_sw_flux_dn,
-                    axes(state.f),
-                ),
-                nlevels + half,
-            ),
-        )
-    else
-        out .= Fields.level(
+        out = copy(
```
szy21, line 83: I think you accidentally removed Fields.level(..., nlevels+half) here and in other TOA diagnostics?

szy21 (approved):
Thanks!

PR #3815, opened 2025-05-21
Add preliminary implicit solver interface and documentation

dennisYatunin:
## Purpose

This PR is split off from #2730. It contains some preliminary setup for the new implicit solver interface, as well as documentation improvements and some code simplifications.

## Content

- Replace the `ImplicitEquationJacobian` with a more general `JacobianAlgorithm` that can be wrapped in a `Jacobian`. For now, the only available `JacobianAlgorithm` is `ManualSparseJacobian`, but I will also add `AutoDenseJacobian`, `AutoSparseJacobian`, and `DebugJacobian` in future PRs.
- Rename `implicit_solver.jl` to `manual_sparse_jacobian.jl`. The first part of this file, which now deals with the `JacobianAlgorithm` interface, will be moved into a separate file called `jacobian.jl` in another PR. Keeping everything in one file for now ensures that the Git history of the manual Jacobian approximation will be preserved.
- Add docstrings for `Jacobian` and `JacobianAlgorithm`, and move some of the original docstring for `ImplicitEquationJacobian` into a new documentation page for the implicit solver.
- Get rid of outdated code that drops unnecessary values from `p` for the Jacobian approximation, since this was only needed when we used `bycolumn`.
- Rearrange and remove some code from the Jacobian approximation to avoid inconsistent formatting and code duplication.
- Remove extraneous function definitions related to the ODE solver from `type_getters.jl`.
- Add a utility function that summarizes arbitrary data structures, and use it to print info about the `JacobianAlgorithm`.

----

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -35,6 +35,7 @@ makedocs(;
         "Ocean Surface Albedo Parameterization" => "surface_albedo.md",
         "Topography Representation" => "topography.md",
         "Tracers" => "tracers.md",
+        "Implicit Solver" => "implicit_solver.md",
         "Radiative Equilibrium" => "radiative_equilibrium.md",
         "Single Column Model" => "single_column_prospect.md",
         "Restarts and checkpoints" => "restarts.md",
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -43,10 +43,12 @@ ClimaAtmos.InitialConditions.Bomex
 ClimaAtmos.InitialConditions.Soares
 ```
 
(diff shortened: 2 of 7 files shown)


trontrytel (approved):
I don't think I'm qualified to review the code, but thank you for adding the documentation page! 

I think that a super helpful thing to add at some point would be a small tutorial that show how to move a simple relaxation like tendency from explicit to implicit. Maybe based on the changes we are doing for non-equilibrium cloud formation?

PR #3817, opened 2025-05-22
Try out columnwise

charleskawczynski:
Very much wip.

PR #3819, opened 2025-05-23
Add `assert_eltype` helper

charleskawczynski:
When working on #3817, I found some things that surprised me a bit. In particular, with the new helper (added in this PR), I believe that this will fail (let's see what CI says):

```julia
function surface_velocity(ᶠu₃, ᶠuₕ³)
    assert_eltype(ᶠu₃, Geometry.Covariant3Vector)
    assert_eltype(ᶠuₕ³, Geometry.Contravariant3Vector)
    sfc_u₃ = Fields.level(ᶠu₃.components.data.:1, half)
    sfc_uₕ³ = Fields.level(ᶠuₕ³.components.data.:1, half)
    sfc_g³³ = g³³_field(sfc_u₃)
    w₃ = @. lazy(-sfc_uₕ³ / sfc_g³³) # u³ = uₕ³ + w³ = uₕ³ + w₃ * g³³
    assert_eltype(w₃, Geometry.Covariant3Vector)
    return w₃
end
```
despite, where this function is called looks like `sfc_u₃ .= surface_velocity(Y.f.u₃, ᶠuₕ³)`. The variable being assigned appears to be a `Covariant3Vector`, however, the RHS of that expression doesn't seem to be.

I think the reason is because we reach into the components to get floats, and then the type of the output is actually a `Contravariant3Vector`, as it adopts the type of `g³³`.

This doesn't error, however, because ClimaCore has conversion methods for this sort of thing. I'm not sure if this is intended behavior, or if this will yield the intended result or not.

Assuming ClimaCore does this automatically, I'm actually surprised that this method isn't simply written as

```julia
function surface_velocity(ᶠu₃, ᶠuₕ³)
    assert_eltype(ᶠu₃, Geometry.Covariant3Vector)
    assert_eltype(ᶠuₕ³, Geometry.Contravariant3Vector)
    sfc_u₃ = Fields.level(ᶠu₃, half)
    sfc_uₕ³ = Fields.level(ᶠuₕ³, half)
    w₃ = @. lazy(-sfc_uₕ³ / sfc_u₃)
    assert_eltype(w₃, Geometry.Covariant3Vector)
    return w₃
end
```

Also, the `assert_eltype` should work for DataLayouts, Fields, and Broadcasted objects, so we can safely remove some of the other type assertions, which currently prevents some of the lazy machinery from working.

### DIFF
--- a/src/cache/precomputed_quantities.jl
+++ b/src/cache/precomputed_quantities.jl
@@ -284,15 +284,11 @@ end
 function surface_velocity(ᶠu₃, ᶠuₕ³)
     assert_eltype(ᶠu₃, Geometry.Covariant3Vector)
     assert_eltype(ᶠuₕ³, Geometry.Contravariant3Vector)
+    ᶠlg = Fields.local_geometry_field(axes(ᶠu₃))
     sfc_u₃ = Fields.level(ᶠu₃.components.data.:1, half)
     sfc_uₕ³ = Fields.level(ᶠuₕ³.components.data.:1, half)
     sfc_g³³ = g³³_field(sfc_u₃)
-    w₃ = @. lazy(-sfc_uₕ³ / sfc_g³³) # u³ = uₕ³ + w³ = uₕ³ + w₃ * g³³
-
-    # sfc_u₃ = Fields.level(ᶠu₃, half)
-    # sfc_uₕ³ = Fields.level(ᶠuₕ³, half)
-    # w₃ = @. lazy(-sfc_uₕ³ / sfc_u₃) # are metric terms automatically applied here?
-
+    w₃ = @. lazy(-C3(sfc_uₕ³ / sfc_g³³, ᶠlg)) # u³ = uₕ³ + w³ = uₕ³ + w₃ * g³³
     assert_eltype(w₃, Geometry.Covariant3Vector)
     return w₃
 end
--- a/src/cache/precomputed_quantities.jl
+++ b/src/cache/precomputed_quantities.jl
@@ -256,6 +256,7 @@ end
 
 # Interpolates the third contravariant component of Y.c.uₕ to cell faces.
 function compute_ᶠuₕ³(ᶜuₕ, ᶜρ)
+    assert_eltype(ᶜuₕ, Geometry.Covariant12Vector)
     ᶜJ = Fields.local_geometry_field(ᶜρ).J
     return @. lazy(ᶠwinterp(ᶜρ * ᶜJ, CT3(ᶜuₕ)))
 end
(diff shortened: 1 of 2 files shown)


dennisYatunin (commented):
@trontrytel and I are confused about the purpose of this PR. Is there a bug that you have identified in `surface_velocity`? I am not sure what that bug could be, as the surface vertical velocity follows what is in the dycore paper. I'm pretty sure this function needs to return a scalar Field, given how it is used in `set_velocity_at_surface!`.

Also, I am strongly opposed to adding type assertions throughout our code like this. They make it very challenging to properly use duck typing, e.g., to use `Covariant1Vector`s instead of `Covariant12Vector`s for horizontal velocity in 2D domains. Unnecessary type restrictions like this have been making it very difficult for me to implement automatic differentiation in ClimaAtmos, and I think we should avoid them wherever possible.

charleskawczynski:
> Is there a bug that you have identified in `surface_velocity`?

Potentially?

> I am not sure what that bug could be, as the surface vertical velocity follows what is in the dycore paper. And this function needs to return a scalar Field, given how it is used in `set_velocity_at_surface!`.

Shouldn't it return a `Covariant3Vector` field? That's the type of the destination.

> Also, I am strongly opposed to adding type assertions throughout our code like this. They make it very challenging to properly use duck typing, e.g., to use `Covariant1Vector`s instead of `Covariant12Vector`s for horizontal velocity in 2D domains. Unnecessary type restrictions have been making it very difficult for me to implement automatic differentiation in ClimaAtmos, and I think we should avoid them wherever possible.

I don't think that this type assertion will restrict autodiff stuff. Actually, I'd also like to remove some type assertions, like `Field` in `compute_kinetic`, as that doesn't allow it to work with broadcasted objects. This type assertion, however, will, as it also supports broadcasted objects, and there is no assertion on the underlying float type, just the axistensor type.

I see type assertions as a tool to help ensure correctness. I agree that we shouldn't use them if they restrict behavior, and as far as I can tell, this one should not.

dennisYatunin:
The destination is actually a scalar field, not a `Covariant3Vector` field. See the following lines in `set_velocity_at_surface!`: https://github.com/CliMA/ClimaAtmos.jl/blob/5e6f60138a7e76b13f014c7866784995743f6615/src/cache/precomputed_quantities.jl#L273-L274

So returning a `Covariant3Vector` field will likely break `set_velocity_at_surface!`. If you're really keen on returning a vector field from this function, you'll need to change `set_velocity_at_surface!` appropriately. However, I would suggest sticking with what we have now, since it is exactly what we wrote in the dycore paper.

And yes, this particular type assertion does not restrict autodiff; it only restricts the horizontal velocity type. I was just pointing out that most type assertions constrain the amount of duck typing we can perform, with the types of real numbers and horizontal velocities being two examples. If we want to check for correct types, we should do so through specific unit tests, instead of scattering restrictive type assertions around our code.

charleskawczynski:
I was under the assumption that the `₃`subscript indicated Covariant vector.

charleskawczynski:
I'll close, as it seems that people don't want this.

PR #3820, opened 2025-05-23
Add CAPE diagnostic

Julians42:
## Purpose 

Adds a diagnostic for computing convective available potential energy (CAPE), or the vertical integral of the buoyancy differential between a parcel lifted from the surface and the environment. Exemplified in the TRMM deep convection case.  Future TODOs and additional diagnostics are recorded in #3823.
## Content

----

### DIFF
--- a/src/diagnostics/core_diagnostics.jl
+++ b/src/diagnostics/core_diagnostics.jl
@@ -1349,3 +1349,62 @@ add_diagnostic_variable!(
         end
     end,
 )
+
+###
+# Convective Available Potential Energy (2d)
+###
+function compute_cape!(out, state, cache, time)
+    thermo_params = lazy.(CAP.thermodynamics_params(cache.params))
+    g = lazy.(TD.Parameters.grav(thermo_params))
+
+    # Get surface parcel properties
+    surface_thermal_state = lazy.(cache.precomputed.sfc_conditions.ts)
+    surface_q =
+        lazy.(TD.total_specific_humidity.(thermo_params, surface_thermal_state))
+    surface_θ_liq_ice =
+        lazy.(TD.liquid_ice_pottemp.(thermo_params, surface_thermal_state))
+
+    # Create parcel thermodynamic states at each level based on energy & moisture at surface
+    parcel_ts_moist =
+        lazy.(
+            TD.PhaseEquil_pθq.(
+                thermo_params,
+                cache.precomputed.ᶜp,
+                surface_θ_liq_ice,
+                surface_q,
+            )
+        )
+
+    # Calculate virtual temperatures for parcel & environment
+    parcel_Tv = lazy.(TD.virtual_temperature.(thermo_params, parcel_ts_moist))
+    env_Tv =
+        lazy.(TD.virtual_temperature.(thermo_params, cache.precomputed.ᶜts))
+
+    # Calculate buoyancy from the difference in virtual temperatures
+    ᶜbuoyancy = cache.scratch.ᶜtemp_scalar
+    ᶜbuoyancy .= g .* (parcel_Tv .- env_Tv) ./ env_Tv
+
+    # restrict to tropospheric buoyancy (generously below 20km) TODO: integrate from LFC to LNB 
+    FT = Spaces.undertype(axes(ᶜbuoyancy))
+    ᶜbuoyancy .=
+        ᶜbuoyancy .*
+        ifelse.(
+            Fields.coordinate_field(state.c.ρ).z .< FT(20000.0),
+            FT(1.0),
+            FT(0.0),
+        )
+
+    # Integrate positive buoyancy to get CAPE
+    out′ = isnothing(out) ? zeros(axes(Fields.level(state.f, half))) : out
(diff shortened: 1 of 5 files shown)


src/diagnostics/core_diagnostics.jl
```diff
@@ -1349,3 +1349,95 @@ add_diagnostic_variable!(
...
+    g = TD.Parameters.grav(thermo_params)
+
+    # Get surface parcel properties
+    surface_thermal_state = Fields.level(cache.precomputed.ᶜts, 1)
+    surface_q = TD.total_specific_humidity.(thermo_params, surface_thermal_state)
+    surface_θ_liq_ice = TD.liquid_ice_pottemp.(thermo_params, surface_thermal_state)
+
+    # Create parcel thermodynamic states at each level based on energy & moisture at surface
+    parcel_ts_moist = TD.PhaseEquil_pθq.(thermo_params, 
+                                        cache.precomputed.ᶜts.p,
```
trontrytel, line 1367: Don't understand here. You are passing in pressure at each level but first cell center theta and q?

src/diagnostics/core_diagnostics.jl
```diff
@@ -1349,3 +1349,95 @@ add_diagnostic_variable!(
...
+    thermo_params = CAP.thermodynamics_params(cache.params)
+    g = TD.Parameters.grav(thermo_params)
+
+    # Get surface parcel properties
+    surface_thermal_state = Fields.level(cache.precomputed.ᶜts, 1)
+    surface_q = TD.total_specific_humidity.(thermo_params, surface_thermal_state)
+    surface_θ_liq_ice = TD.liquid_ice_pottemp.(thermo_params, surface_thermal_state)
+
+    # Create parcel thermodynamic states at each level based on energy & moisture at surface
+    parcel_ts_moist = TD.PhaseEquil_pθq.(thermo_params,
```
akshaysridhar, line 1366: Do we want the thermo state constructor to reflect the moisture model type ? Or is the equilibrium thermo state assumption sufficient here ?

src/diagnostics/core_diagnostics.jl
```diff
@@ -1349,3 +1349,95 @@ add_diagnostic_variable!(
...
+    env_Tv = TD.virtual_temperature.(thermo_params, cache.precomputed.ᶜts)
+
+    # Calculate buoyancy from the difference in virtual temperatures
+    ᶜbuoyancy = cache.scratch.ᶜtemp_scalar
+    ᶜbuoyancy .= g .* (parcel_Tv .- env_Tv) ./ env_Tv
+
+    # Integrate positive buoyancy to get CAPE
+    if isnothing(out)
+        out = zeros(axes(Fields.level(state.f, half)))
+        Operators.column_integral_definite!(out, max.(ᶜbuoyancy, 0))
```
trontrytel, line 1382: Not sure if this (and below `min` for CIN) are valid. This should be integrated between surface and the level of free convection and between the level of free convection the equilibrium level. I think there might be some weird cases where you have more than one of those. See for example the discussion here: 

https://github.com/Unidata/MetPy/issues/862

src/diagnostics/core_diagnostics.jl
```diff
@@ -1349,3 +1349,95 @@ add_diagnostic_variable!(
...
+
+###
+# Convective Available Potential Energy (2d)
+###
+function compute_cape!(out, state, cache, time)
+    thermo_params = CAP.thermodynamics_params(cache.params)
+    g = TD.Parameters.grav(thermo_params)
+
+    # Get surface parcel properties
+    surface_thermal_state = Fields.level(cache.precomputed.ᶜts, 1)
```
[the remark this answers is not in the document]
Julians42: This mattered a lot. Order 30% difference in CAPE values and now we actually see a release of potential energy when we have convection. Thanks @trontrytel!!

src/diagnostics/core_diagnostics.jl
Julians42: I think equilibrium thermo state is fine here based on what I understand about cape. @trontrytel do you agree?

src/diagnostics/core_diagnostics.jl
```diff
@@ -1349,3 +1349,52 @@ add_diagnostic_variable!(
...
+    Operators.column_integral_definite!(out′, lazy.(max.(ᶜbuoyancy, 0)))
+    return out′
+end
+
+add_diagnostic_variable!(
+    short_name = "cape",
+    long_name = "Convective Available Potential Energy",
+    standard_name = "convective_available_potential_energy",
+    units = "J kg^-1",
+    comments = "Energy available to a parcel lifted moist adiabatically from the surface. We assume fully reversible phase changes and no precipitation.",
```
costachris, line 1398: While it's more standard to integrate between the LFC and EL when computing surface-based CAPE, just mention the assumptions or integration bounds in the comment depending on what you ultimately do.

costachris: I would at least have a minimum pressure so you don't include non-tropospheric CAPE in the integral (some simulations have model tops in the mesosphere). Otherwise, having the EL will prevent this.

PR #3824, opened 2025-05-28
Refactor and document ED pieces

costachris:
Consolidate ED logic in eddy_diffusion_closures, update buoyancy calculation, and add docstrings and comments

[ToDo] Added l_grid (right now should be proportional to dz) to MixingLength output struct; needs to be made consistent with definition elsewhere


src/solver/types.jl
```diff
@@ -343,6 +343,12 @@ struct MixingLength{FT}
     wall::FT
     tke::FT
     buoy::FT
+    l_grid::FT
+end
+
+struct EddyViscosityDiffusivity{FT}
```
trontrytel: Lmk if removing this struct helped with the allocations!

costachris: I ended up removing the `EddyViscosityDiffusivity` struct and creating separate functions for viscosity and diffusivity.

tapios (approved):
LGTM

PR #3825, opened 2025-05-28
Revert "Restructure radiation callback"

akshaysridhar:
Revert CliMA/ClimaAtmos.jl#3796 while we trace issues in the coupled setup. The original PR passed downstream Coupler checks and raised no flags in regression checks, so we may wish to revisit these. This should revive nightly AMIP builds.


szy21:
AMIP still fails with this PR: https://buildkite.com/clima/climacoupler-coarse-nightly-amip/builds/372#019718ba-456e-4255-a717-20857be6f44f. I guess this is not the reason then. I don't know what else changed between this Monday 7pm and Tuesday 7pm though.

szy21:
On a side note, this breaks faster than the current main, which means this PR changes the behavior (unless amip is non deterministic again), which is also surprising.

PR #3827, opened 2025-05-29
Add AutoDenseJacobian algorithm for implicit solver

dennisYatunin:
## Purpose 

This is a followup PR for #3815 that adds the next `JacobianAlgorithm` from #2730 -- the automatic dense Jacobian with LU factorization.

## Content

- Add ForwardDiff as a dependency, and bump ClimaComms to 0.6.8 to use the new `@threaded` macro
- Add the `AutoDenseJacobian`, which uses dual numbers to compute a dense representation of the Jacobian matrix, factorizes the matrix with `parallel_lu_factorize!`, and solves each linear system with `parallel_lu_solve!`
  - The computation of the dense matrix needs to be done in batches to avoid excessive compilation times
  - The general strategy for initializing the dual numbers and translating them into dense matrix entries will be similar for the upcoming `AutoSparseJacobian`
- Add some utility functions for defining custom kernels in `autodiff_utils.jl`, which will also be needed for `AutoSparseJacobian`
- Move code that is used for both `ManualSparseJacobian` and `AutoDenseJacobian` into a separate `jacobian.jl` file
- Override the method for `sqrt` from ForwardDiff so that the derivative of `x * sqrt(x)` at `x = 0` is `0` rather than `NaN`
- Remove some unnecessary `FT` conversions from `vertical_diffusion_boundary_layer_tendency!`, since they break when `FT` is a dual number
- Change `sgs_weight_function` so that it doesn't generate `NaN`s when evaluated with dual numbers
- Fix an accidental conversion to `Float64` in `smagorinsky_lilly_length`
- Add two new config arguments for using `AutoDenseJacobian`:
  - `use_dense_jacobian`, which replaces`ManualSparseJacobian` with `AutoDenseJacobian`
  - `update_jacobian_every`, which switches between updating once every timestep, every timestepper stage, and every linear solve (the latter has been our default choice so far, but it is not very performant for `AutoDenseJacobian`)
- Add several tests to the CI pipeline that run representative examples with `use_dense_jacobian: true` and `update_jacobian_every: dt`
- Fix header formatting in `api.md`

----


src/prognostic_equations/implicit/auto_dense_jacobian.jl
```diff
@@ -0,0 +1,300 @@
+import ForwardDiff
+import LinearAlgebra: diagind
+
+import ClimaComms
+import ClimaCore.DataLayouts: MArray
+
+"""
+    AutoDenseJacobian([max_simultaneous_derivatives])
```
sajjadazimi, line 8: This is a clear documentation explaining the logic behind the algorithm well and containing useful references. I was able to understand the algorithm and roughly what the code is meant to do based on the documentation.

src/prognostic_equations/implicit/auto_dense_jacobian.jl
```diff
@@ -0,0 +1,300 @@
...
+            getfield(p, cache_field_index)
+        end
+    end
+    p_dual = AtmosCache(p_dual_args...)
+
+    batches = Iterators.partition(scalar_level_pairs, batch_size)
+    for batch in ClimaComms.threadable(device, batches)
+        # Add a unique ε to Y for each scalar level index in this batch.
+        Y_dual .= Y
+        ClimaComms.@threaded device begin
```
sajjadazimi, line 118: This is unclear to me what this macro does exactly. Maybe a short comment would be useful.

src/prognostic_equations/implicit/auto_dense_jacobian.jl
```diff
@@ -0,0 +1,300 @@
...
+            for column_index in column_indices,
+                (ε_index, (_, (name_index, level))) in enumerate(batch)
+
+                partials = ntuple(i -> i == ε_index ? 1 : 0, batch_size_val)
+                dual_value = ForwardDiff.Dual{Jacobian}(0, partials...)
+                unrolled_applyat(name_index, scalar_names) do name
+                    field = MatrixFields.get_field(Y_dual, name)
+                    @inbounds point(field, level, column_index...)[] +=
```
sajjadazimi: Same as above!

src/prognostic_equations/implicit/auto_dense_jacobian.jl
```diff
@@ -0,0 +1,300 @@
...
+            end
+        end
+    end
+end
+
+function update_jacobian!(alg::AutoDenseJacobian, cache, Y, p, dtγ, t)
+    (; column_matrices, column_lu_factors, I_matrix, N_val) = cache
+    device = ClimaComms.device(Y.c)
+
+    # Set column_lu_factors to dtγ * J - I, where J is the dense Jacobian.
```
sajjadazimi: I don't understand why you do this.

src/prognostic_equations/implicit/autodiff_utils.jl
```diff
@@ -0,0 +1,80 @@
...
+# order to properly handle derivatives of `x * sqrt(x)`. Without this change,
+# the derivative of `x * sqrt(x)` is `NaN` when `x` is zero.
+@inline function Base.sqrt(d::ForwardDiff.Dual{Jacobian})
+    tag = Val(Jacobian)
+    x = ForwardDiff.value(d)
+    partials = ForwardDiff.partials(d)
+    val = sqrt(x)
+    deriv = iszero(x) ? zero(x) : inv(2 * val)
+    return ForwardDiff.dual_definition_retval(tag, val, deriv, partials)
+end
```
sajjadazimi, line 92: I understand that you need to avoid NaN for numerical purposes, but how do we know that replacing them with 0 does not change Newton's iteration solution?

sajjadazimi (commented):
Thank you!
The documentation is nicely written. It contains good references and explains the algorithm well. I understand what the code pieces are meant to do based on the documentation. I left a couple of other minor comments.

ray-chew:
Thanks, Dennis, for the detailed documentation.

As I have already asked you all my questions in person, I will not repeat them here. Since I am still in the process of understanding the code, my review may not be very detailed or helpful.

What I have done instead is to extend the `implicit_solver.jl` documentation based on my current understanding from your explanations, the existing documentation, and the code itself. This extension should reveal any of my misunderstandings.

ray-chew (approved):

dennisYatunin:
@sajjadazimi, I've added several code comments to clarify the snippets you've highlighted, and also an explanation of why we are allowed to override the derivative of `sqrt`. Ray has also extended `implicit_solver.md` with additional documentation. Let me know if any of the new comments and documentation are unclear.

ray-chew:
> @sajjadazimi, I've added several code comments to clarify the snippets you've highlighted, and also an explanation of why we are allowed to override the derivative of `sqrt`. Ray has also extended `implicit_solver.md` with additional documentation. Let me know if any of the new comments and documentation are unclear.

There are some typesetting issues in the modifications I made. I will fix them tomorrow, my first time typesetting math in Markdown.

docs/src/implicit_solver.md
```diff
@@ -1,15 +1,22 @@
 # Implicit Solver
+We use an implicit or split implicit-explicit (IMEX) timestepping scheme, and assume that we have the solution from the explicit step, ``Y_{prev}``. Now we want to solve the implicit part of the equation,
+```math
+Y^{n+1} = Y_{prev} + \Delta t * T_{imp}(Y^{n+1}),
+```
+where ``Δt * T_{imp}(Y)`` is the contribution from th implicit tendency ``T_{imp}`` between times ``t`` and ``t + Δt``.
```
charleskawczynski:
```suggestion
where ``Δt * T_{imp}(Y)`` is the contribution from the implicit tendency ``T_{imp}`` between times ``t`` and ``t + Δt``.
```

src/prognostic_equations/implicit/autodiff_utils.jl
```diff
@@ -0,0 +1,92 @@
...
+# In general, this change is valid when every function of sqrt(x) can be
+# expanded around x = 0 with a leading term of the form c * x^p * sqrt(x), where
+# c and p are constants and p ≥ -1/2. For example, this will be the case for any
+# linear function of sqrt(x), and more generally any linear function of
+# f(x) * sqrt(x), where f(x) has a Taylor series expansion around x = 0. On the
+# other hand, this change will not be valid for a function like sqrt(x) / x,
+# since its derivative at x = 0 should be ±Inf but this change makes it NaN. If
+# a function whose leading term has p < -1/2 is ever called by
+# set_implicit_precomputed_quantities! or implicit_tendency!, our strategy for
+# avoiding NaNs in the dissipation tendency derivative will need to be modified.
```
sajjadazimi: I think it's valid for p >= +1/2. The derivative of x^p * sort(x) is C * x^p / sort(x) which is ±Inf for p < 1/2.

src/prognostic_equations/implicit/autodiff_utils.jl
```diff
@@ -0,0 +1,92 @@
...
+
+# Every nonstandard rule should be defined using a method that specializes on
+# the tag Jacobian. Not specializing on any tag overwrites the generic method
+# for Dual in ForwardDiff and breaks precompilation, while specializing on the
+# default tag Nothing causes the type piracy test to fail.
+
+# Set the derivative of sqrt(x) to iszero(x) ? zero(x) : inv(2 * sqrt(x)) in
+# order to properly handle derivatives of x * sqrt(x). Without this change, the
+# derivative of the turbulent kinetic energy dissipation tendency, which is a
+# linear function of ᶜtke⁰ * sqrt(ᶜtke⁰), is NaN whenever tke⁰ = 0 at any point.
```
sajjadazimi: If we express the tke dissipation tendency as tke^1.5 would it solve the problem without overriding sqrt derivative?

src/prognostic_equations/implicit/autodiff_utils.jl
dennisYatunin: Good catch! I will update it.

src/prognostic_equations/implicit/autodiff_utils.jl
dennisYatunin: Yes I think that would also fix it, but we are trying to find an autodiff strategy that adds minimal restrictions to the code we can use in our parametrizations. We could just have a policy that developers need to write `x^FT(1.5)` instead of `x * sqrt(x)`, but this won't scale well if we keep finding more expressions that generate NaN derivatives in autodiff. If we don't fix the underlying sources of the NaN derivatives, we might end up in a situation where developers need to remember a bunch of esoteric rules whenever they use autodiff.

PR #3828, opened 2025-05-29
Make cape diagnostic optional

akshaysridhar:
modified:   src/diagnostics/default_diagnostics.jl

## Content

----

### DIFF
--- a/src/diagnostics/default_diagnostics.jl
+++ b/src/diagnostics/default_diagnostics.jl
@@ -209,7 +209,6 @@ function default_diagnostics(
         "lwp",
         "clwvi",
         "clivi",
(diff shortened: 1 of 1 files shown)


szy21 (approved):
Thanks! Maybe add cape to one ci case to make sure it can run?

akshaysridhar:
> Thanks! Maybe add cape to one ci case to make sure it can run?

It's run in the `trmm_column` example as part of the ci pipeline. There should be a corresponding AMIP result in a few hours as well.

Julians42 (approved):
Yes this looks fine. Hopefully this fixes the nightly tests. Thanks for tracking it down.

PR #3832, opened 2025-05-31
Loosen compute_kinetic types

charleskawczynski:
Peel off of #3817.

### DIFF
--- a/src/utils/utilities.jl
+++ b/src/utils/utilities.jl
@@ -58,11 +58,13 @@ individual velocity components:
     cell centers, and
  - `uᵥ` should be a `Covariant3Vector`-valued field at cell faces.
 """
-function compute_kinetic(uₕ::Fields.Field, uᵥ::Fields.Field)
+function compute_kinetic(uₕ, uᵥ)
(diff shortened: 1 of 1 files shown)


src/utils/utilities.jl
```diff
@@ -58,11 +58,13 @@ individual velocity components:
     cell centers, and
  - `uᵥ` should be a `Covariant3Vector`-valued field at cell faces.
 """
-function compute_kinetic(uₕ::Fields.Field, uᵥ::Fields.Field)
+function compute_kinetic(uₕ, uᵥ)
     @assert eltype(uₕ) <: Union{C1, C2, C12}
```
imreddyTeja, line 62: Do the asserts still need to be here/work? Is my understanding correct that this function is changed to accept broadcasted objects?

charleskawczynski: `eltype` is defined on `Broadcasted` objects, so removing the type restriction should not have any behavioral changes. I personally think it's good to keep the assertions around, so that users don't accidentally pass the wrong types in.

PR #3840, opened 2025-06-04
Docstrings in prognostic_equations

tapios:
## Purpose 

Adds docstrings to prognostic EDMF equations & forcing files; cleanup of logic in a few places; more consistent naming. 

Summary of Changes: 

1. Added comprehensive docstrings throughout prognostic equations modules for EDMF
2. Renamed edmf_coriolis → scm_coriolis,  remove redundant coordinate transformations in the cross product [behavior change]
3. Renamed horizontal_advection → horizontal_dynamics
4. Cleaned up entrainment/detrainment code, entr/detr only for updraft area above eps(FT)  [behavior change]
5. Split gradient Richardson number calculation into separate functions
6. Fixed hard-coded thermodynamic values

## To-do

- [x] Rebase after #3824 
- [x] Test
- [x] Stable with diagnostic EDMF AMIP nightly run [`amip_target_diagedmf.yml`]  

## Content

----


src/prognostic_equations/zero_velocity.jl
```diff
@@ -18,7 +39,29 @@ function zero_velocity_tendency!(Yₜ, Y, p, t)
...
+
+For matrix rows corresponding to velocity variables (grid-mean `uₕ`, `u₃`, and
+EDMFX updraft `u₃ʲ` if applicable):
+- Diagonal blocks (e.g., `∂(uₕ_tendency)/∂uₕ`) are set to represent `-I` (negative identity).
+- Off-diagonal blocks (e.g., `∂(uₕ_tendency)/∂(tracer)`) are set to zero.
+
+This effectively decouples the velocity evolution from other variables in the
+linearized system used by an implicit solver, or simplifies their implicit
+contribution to behave like a strong relaxation to zero tendency (if the Jacobian
+is for $Y' - \Delta t J Y'$ and $J = -I/\Delta t$). This is useful if velocities
```
charleskawczynski:
```suggestion
is for `Y' - \Delta t J Y'` and `J = -I/\Delta t`). This is useful if velocities
```
using `$` like that in a doc string is attempting to escape the doc string, can we change it to tick marks?

src/prognostic_equations/edmfx_boundary_condition.jl
```diff
@@ -25,28 +77,82 @@ function sgs_scalar_first_interior_bc(
...
+        return 4 * c_star_sq
     end
 end
 
+"""
+    approximate_inverf(x::FT) where {FT}
+
+Approximates the inverse error function, erf⁻¹(x).
+
+The approximation formula involves logarithmic and square root operations.
```
tapios:
```suggestion
The approximation formula involves logarithmic and square root operations and is 
based on a Pade approximation due to Sergei Winitzki (https://drive.google.com/file/d/0B2Mt7luZYBrwZlctV3A3eF82VGM/view?resourcekey=0-UQpPhwZgzP0sF4LHBDlLtgi)
```

src/prognostic_equations/edmfx_boundary_condition.jl
```diff
@@ -25,28 +77,82 @@ function sgs_scalar_first_interior_bc(
...
+Returns:
+- An approximation of erf⁻¹(x).
+
+Note: Numerical precision issues or invalid results may occur if `x` is too
+close to -1 or 1, due to `log(1 - x^2)`. The conditions for the terms under
+square roots (`term1^2 - term2 >= 0` and `term3 - term1 >= 0`) must also hold
+for the approximation to be valid.
+"""
 function approximate_inverf(x::FT) where {FT}
     # From Wikipedia
```
tapios:
```suggestion
    # From Sergei Winitzki (https://drive.google.com/file/d/0B2Mt7luZYBrwZlctV3A3eF82VGM/view?resourcekey=0-UQpPhwZgzP0sF4LHBDlLtgi)
```

src/prognostic_equations/forcing/external_forcing.jl
```diff
@@ -103,6 +217,9 @@ function external_forcing_cache(Y, external_forcing::GCMForcing, params, _)
         end
 
         function set_insolation!(cc_field)
+            # TODO: Clarify this (Zhaoyi?). If rsdt already is insolation on a 
+            # horizontal plane, why are we dividing by cos(SZA)? Is this to get 
+            # TSI from the GCM?
```
tapios: @szy21 Could you please have a look at this and clarify it?

src/prognostic_equations/advection.jl
```diff
@@ -40,14 +61,46 @@ NVTX.@annotate function horizontal_advection_tendency!(Yₜ, Y, p, t)
     end
 
     if use_prognostic_tke(p.atmos.turbconv_model)
-        @. Yₜ.c.sgs⁰.ρatke -= wdivₕ(Y.c.sgs⁰.ρatke * ᶜu⁰)
+        if p.atmos.turbconv_model isa EDOnlyEDMFX
```
trontrytel, line 64: Could we just have:
```
        if p.atmos.turbconv_model isa AbstractEDMF
            ᶜu_for_tke_advection = p.precomputed.ᶜu⁰
        else
            # Fallback for non-EDMFX models that might use prognostic TKE.
            # Defaulting to grid mean velocity ᶜu 
            ᶜu_for_tke_advection = ᶜu # Or consider an error if this case is invalid
        end
```

src/prognostic_equations/surface_temp.jl
```diff
@@ -41,19 +100,27 @@ function surface_temp_tendency!(Yₜ, Y, p, t, slab::PrognosticSurfaceTemperatur
         Q = FT(0)
     end
 
-    @. Yₜ.sfc.T -=
-        (sfc_rad_e_flux + sfc_turb_e_flux + Q) /
-        (ρ_ocean * cp_ocean * depth_ocean)
+    # TODO: The convention of "precipitation" being positive *upward* is very confusing.
```
trontrytel: I thought the idea was to use the same convention for the sign of fluxes in the entire code base. Converting them somewhere at the surface in Atmos may be more prone to errors

src/prognostic_equations/forcing/subsidence.jl
```diff
@@ -1,28 +1,70 @@
...
+
+The formulation `ᶜρ * (ᶜsubdivᵥ(Flux) - ᶜχ * ᶜsubdivᵥ(ᶠu³))` is equivalent to
+`ᶜρ * (ᶠu³ ⋅ ∇ᶜχ)`, implementing the advective form. The result is subtracted
+from `ᶜρχₜ`, effectively adding `ρ * (-ᶠu³ ⋅ ∇ᶜχ)` to it.
 
+Arguments:
+- `ᶜρχₜ`: Field for the tendency of the density-weighted scalar `ρχ`, modified in place.
+- `ᶜρ`: Cell-center density field.
+- `ᶠu³`: Face-valued field of prescribed vertical velocity (subsidence velocity `w`).
+        Typically, w < 0 for subsidence in an upward z coordinate.
```
tapios:
```suggestion
        Typically, `w < 0` for subsidence in an upward `z` coordinate.
```

src/prognostic_equations/implicit/implicit_tendency.jl
```diff
@@ -58,7 +58,7 @@ NVTX.@annotate function implicit_tendency!(Yₜ, Y, p, t)
     pressure_work_tendency!(Yₜ, Y, p, t, p.atmos.turbconv_model)
 
     # NOTE: This will zero out all momentum tendencies in the edmfx advection test
-    # please DO NOT add additional velocity tendencies after this function
+    # DO NOT add additional velocity tendencies after this function
```
trontrytel, line 61: We are less polite now :)

trontrytel (approved):
Thank you for adding all the docstrings! I didn't read through all of them, but the ones I did check were useful.

src/prognostic_equations/forcing/external_forcing.jl
szy21: Yes, this is because rrtmgp requires TSI and zenith angle and calculates downward SW fluxes at TOA, while the GCM forcing provides the downward SW fluxes at TOA (rsdt) and zenith angle.

src/prognostic_equations/advection.jl
```diff
@@ -1,22 +1,43 @@
...
 import ClimaCore.Fields as Fields
 import ClimaCore.Geometry as Geometry
 
-NVTX.@annotate function horizontal_advection_tendency!(Yₜ, Y, p, t)
+"""
+    horizontal_dynamics_tendency!(Yₜ, Y, p, t)
+
+Computes tendencies due to horizontal advection for prognostic variables of the
+grid mean and EDMFX subdomains, and also applies horizontal pressure gradient and 
+gravitational acceleration terms for momentum.
```
szy21:
```suggestion
gravitational acceleration terms for horizontal momentum.
```

src/prognostic_equations/advection.jl
szy21: There is no `p.precomputed.ᶜu⁰` for EDOnlyEDMFX (which is AbstractEDMF).

src/prognostic_equations/advection.jl
```diff
@@ -40,14 +61,46 @@ NVTX.@annotate function horizontal_advection_tendency!(Yₜ, Y, p, t)
...
     if use_prognostic_tke(p.atmos.turbconv_model)
-        @. Yₜ.c.sgs⁰.ρatke -= wdivₕ(Y.c.sgs⁰.ρatke * ᶜu⁰)
+        if p.atmos.turbconv_model isa EDOnlyEDMFX
+            ᶜu_for_tke_advection = ᶜu
+        elseif p.atmos.turbconv_model isa AbstractEDMF
+            ᶜu_for_tke_advection = p.precomputed.ᶜu⁰
+        else
+            # Fallback for non-EDMFX models that might use prognostic TKE.
+            # Defaulting to grid mean velocity ᶜu 
+            ᶜu_for_tke_advection = ᶜu # Or consider an error if this case is invalid
```
szy21: Probably better to have an error

src/prognostic_equations/forcing/external_forcing.jl
```diff
@@ -103,6 +217,9 @@ function external_forcing_cache(Y, external_forcing::GCMForcing, params, _)
         end
 
         function set_insolation!(cc_field)
+            # TODO: Clarify this (Zhaoyi?). If rsdt already is insolation on a
```
tapios:
```suggestion
```

src/prognostic_equations/forcing/external_forcing.jl
```diff
@@ -103,6 +217,9 @@ function external_forcing_cache(Y, external_forcing::GCMForcing, params, _)
         end
 
         function set_insolation!(cc_field)
+            # TODO: Clarify this (Zhaoyi?). If rsdt already is insolation on a 
+            # horizontal plane, why are we dividing by cos(SZA)? Is this to get
```
tapios:
```suggestion
```

src/prognostic_equations/forcing/external_forcing.jl
tapios:
```suggestion
    # rsdt is TOA insolation on a horizontal plane. We need total solar irradiance and the solar 
    # zenith angle separately. So compute `TSI = rsdt/cos(SZA)`.
```

src/prognostic_equations/advection.jl
```diff
@@ -85,6 +138,35 @@ NVTX.@annotate function horizontal_tracer_advection_tendency!(Yₜ, Y, p, t)
...
+- Vertical advection of grid-mean TKE (`ρatke⁰`) if `use_prognostic_tke` is true.
+
+Arguments:
+- `Yₜ`: The tendency state vector, modified in place.
+- `Y`: The current state vector.
+- `p`: Cache containing parameters, core fields (e.g., `ᶜf³`, `ᶠf¹²`, `ᶜΦ`),
+       precomputed fields (e.g., `ᶜu`, `ᶠu³`, `ᶜK`, EDMF velocities/TKE if applicable),
+       atmospheric model settings (`p.atmos.numerics` for upwinding schemes),
+       and scratch space.
+- `dt`: Time step size from `p.dt`.
```
szy21:
```suggestion
```
(dt is not an argument of this function)

src/prognostic_equations/advection.jl
```diff
@@ -184,6 +266,32 @@ NVTX.@annotate function explicit_vertical_advection_tendency!(Yₜ, Y, p, t)
     end
 end
 
+"""
+    edmfx_sgs_vertical_advection_tendency!(Yₜ, Y, p, t, turbconv_model::PrognosticEDMFX)
+
+Computes tendencies due to vertical advection and buoyancy for EDMF subgrid-scale
```
szy21:
```suggestion
Computes tendencies due to vertical advection and buoyancy for EDMFX subgrid-scale
```

src/prognostic_equations/eddy_diffusion_closures.jl
```diff
@@ -528,6 +527,38 @@ function mixing_length(
...
+- `ᶜN²_eff`: Effective squared Brunt-Väisälä frequency [1/s²].
+- `ᶜstrain_rate_norm`: Frobenius norm of the strain rate tensor, |S| [1/s].
+- `ε` is a small machine epsilon value to prevent division by zero.
+
+Returns:
+- The gradient Richardson number (dimensionless scalar).
+"""
+function gradient_richardson_number(params, ᶜN²_eff, ᶜstrain_rate_norm)
+    FT = eltype(params)
+    eps_FT = eps(FT)
```
szy21:
```suggestion
```

src/prognostic_equations/eddy_diffusion_closures.jl
```diff
@@ -528,6 +527,38 @@ function mixing_length(
...
+
+    # Calculate the denominator term for Ri, ensuring it's not zero
+    # Based on the formulation Ri = N^2 / max(2*|S|, eps)
+    ᶜshear_term_safe = max(2 * ᶜstrain_rate_norm, eps_FT)
```
szy21:
```suggestion
    ᶜshear_term_safe = max(2 * ᶜstrain_rate_norm, eps(FT))
```
(Up to you, but I feel there is no need to define a variable eps_FT.)

src/prognostic_equations/edmfx_boundary_condition.jl
```diff
@@ -2,6 +2,54 @@
...
+`SGS_scalar = Mean_scalar + C * σ`.
+
+The SGS variance is computed using Monin-Obukhov Similarity Theory via
+`get_first_interior_variance`. The coefficient `C` (here, `surface_scalar_coeff`)
+is determined by `percentile_bounds_mean_norm`, assuming the updraft samples
+from the upper tail (from percentile `1 - ᶜaʲ_int` to 1) of a Gaussian distribution
+of SGS fluctuations.
+
+This boundary condition is applied only when the surface buoyancy flux
+is positive (unstable conditions), indicating surface-driven updrafts.
```
szy21:
```suggestion
is positive (unstable conditions), indicating surface-driven updrafts. When the surface buoyancy flux is non-positive, the updraft scalar value is set to the grid-mean scalar.
```

src/prognostic_equations/edmfx_boundary_condition.jl
```diff
@@ -25,30 +77,86 @@ function sgs_scalar_first_interior_bc(
...
+        local_geometry, 
+    ) where {FT}
+
+Calculates the variance of a scalar quantity (σ²) at height `z` within the
+surface layer using Monin-Obukhov Similarity Theory (MOST).
+
+The calculation depends on stability:
+- For unstable conditions (Obukhov length `L < 0`):
+  σ² = C₁ * c²∗ * (1 - C₂ * z / L)^(-2/3)
+- For stable/neutral conditions (Obukhov length `L > 0`):
```
szy21:
```suggestion
- For stable/neutral conditions (Obukhov length `L >= 0`):
```

src/prognostic_equations/edmfx_boundary_condition.jl
```diff
@@ -25,30 +77,86 @@ function sgs_scalar_first_interior_bc(
...
+  σ² = C₁ * c²∗
+where `c∗ = -kinematic_scalar_flux / ustar` is the scalar flux scale.
+The constants C₁ and C₂ are empirical (here, C₁=4, C₂=8.3).
+
+Arguments:
+- `kinematic_scalar_flux`: Kinematic surface flux of the scalar (e.g., w'c'_sfc) [K⋅m/s or (kg/kg)⋅m/s].
```
szy21, line 109: Change `kinematic_scalar_flux` to `flux` here, or `flux` to `kinematic_scalar_flux` in the function?

src/prognostic_equations/edmfx_boundary_condition.jl
```diff
@@ -57,17 +165,56 @@ function approximate_inverf(x::FT) where {FT}
...
+- `high_percentile`: The upper percentile bound (e.g., 0.9 for the 90th percentile).
+
+Returns:
+- The mean of the standard normal distribution truncated between the quantiles
+  of `low_percentile` and `high_percentile`.
+"""
 function percentile_bounds_mean_norm(
     low_percentile::FT,
-    high_percentile,
+    high_percentile::FT,
```
szy21, line 212: I think @dennisYatunin was suggesting not to have extra FTs in functions?

src/prognostic_equations/edmfx_entr_detr.jl
```diff
@@ -1,24 +1,107 @@
...
-                                   relative humidity and buoyancy
-   - ᶜw⁰, ᶜRH⁰, ᶜbuoy⁰ - environment physical vertical velocity,
-                              relative humidity and buoyancy
-   - dt - timestep
+    entrainment(
+        thermo_params, turbconv_params, ᶜz, z_sfc, ᶜp, ᶜρ,
+        ᶜaʲ, ᶜwʲ, ᶜRHʲ, ᶜbuoyʲ, ᶜw⁰, ᶜRH⁰, ᶜbuoy⁰, ᶜtke⁰,
+        model_option::AbstractEntrainmentModel,
+    )
+
```
szy21: Move this docstring in front of the entrainment function (calculate_pi_groups)?

src/prognostic_equations/edmfx_entr_detr.jl
szy21, line 1: I think in general we don't need FT in FT(0). But maybe there is no harm in adding it. @dennisYatunin Could you clarify?

src/prognostic_equations/edmfx_sgs_flux.jl
```diff
@@ -181,7 +213,7 @@ function edmfx_sgs_mass_flux_tendency!(
             @. Yₜ.c.ρq_sno += vtt
         end
         # TODO - compute sedimentation and terminal velocities
-        # TODO - add w q_tot, w h_tot terms
+        # TODO - add w q_tot, w h_tot terms associated with sedimentation/falling
```
szy21, line 216:
```suggestion
        # TODO - add w q_tot, w h_tot terms associated with sedimentation/falling
        # TODO - add vertical momentum fluxes
```

src/prognostic_equations/edmfx_sgs_flux.jl
```diff
@@ -259,11 +292,37 @@ function edmfx_sgs_mass_flux_tendency!(
                 )
                 @. Yₜ.c.ρq_tot += vtt
             end
+            # TODO: add environment flux?
         end
```
szy21, line 297:
```suggestion
        end
        
        # TODO: add vertical momentum fluxes
```

src/prognostic_equations/forcing/large_scale_advection.jl
```diff
@@ -1,11 +1,39 @@
...
+Arguments:
+- `ᶜρ`: Cell-center air density field.
+- `thermo_params`: Thermodynamic parameters.
+- `ᶜts`: Cell-center thermodynamic state field.
+- `t`: Current simulation time.
+- `ls_adv`: `LargeScaleAdvection` object containing profile functions for tendencies,
+            or another type if large-scale advection is inactive.
+
+Returns:
+- A `ClimaCore.Fields.Field` representing the tendency `∂(ρq_tot)/∂t` due to
```
szy21: @dennisYatunin  Does @lazy return a ClimaCore Field?

src/prognostic_equations/forcing/large_scale_advection.jl
```diff
@@ -20,6 +48,37 @@ function large_scale_advection_tendency_ρq_tot(
...
+Arguments:
+- `ᶜρ`: Cell-center air density field.
+- `thermo_params`: Thermodynamic parameters.
+- `ᶜts`: Cell-center thermodynamic state field.
+- `t`: Current simulation time.
+- `ls_adv`: `LargeScaleAdvection` object containing profile functions for tendencies,
+            or another type if large-scale advection is inactive.
+
+Returns:
+- A `ClimaCore.Fields.Field` representing the tendency `∂(ρe_tot)/∂t` due to
```
szy21: same as above

src/prognostic_equations/forcing/subsidence.jl
```diff
@@ -1,28 +1,70 @@
...
-#####
-##### No subsidence
-#####
+"""
+    subsidence!(ᶜρχₜ, ᶜρ, ᶠu³, ᶜχ, scheme::Val)
 
-subsidence_tendency!(Yₜ, Y, p, t, ::Nothing) = nothing
+Computes the tendency contribution to a density-weighted scalar `ρχ` due to
+subsidence (vertical advection by a prescribed large-scale vertical velocity `ᶠu³`).
+The tendency is calculated as `ρ * (-w * ∂χ/∂z)`, where `w` is `ᶠu³` and `χ` is `ᶜχ`.
```
szy21: I don't think the second part is accurate, as it's the general coordinate gradient, not ∂ᶜχ/∂z.

src/prognostic_equations/remaining_tendency.jl
```diff
@@ -182,4 +268,20 @@ NVTX.@annotate function additional_tendency!(Yₜ, Y, p, t)
...
+    # NOTE: Precipitation tendencies should be applied before calling this function,
+    # because precipitation cache is used in this function
+    surface_temp_tendency!(Yₜ, Y, p, t, p.atmos.surface_model)
+
+    # NOTE: All ρa tendencies should be applied before calling this function
+    pressure_work_tendency!(Yₜ, Y, p, t, p.atmos.turbconv_model)
+
+    sl = p.atmos.smagorinsky_lilly
+    horizontal_smagorinsky_lilly_tendency!(Yₜ, Y, p, t, sl)
+    vertical_smagorinsky_lilly_tendency!(Yₜ, Y, p, t, sl)
```
szy21, line 281: Is there a particular reason to move these tendencies after gravity wave drag?

src/prognostic_equations/surface_flux.jl
```diff
@@ -1,14 +1,35 @@
...
+adjacent to the bottom boundary. The divergence effectively introduces the
+`ρ_flux_uₕ_surface` as a source/sink. The result is divided by density `ᶜρ`
+to yield a tendency for specific horizontal momentum `uₕ`.
+
+Arguments:
+- `ᶜρ`: Cell-center air density field.
+- `ᶜuₕ`: Cell-center horizontal velocity field (used for type/structure, not value in flux calc).
+- `ρ_flux_uₕ_surface`: The vertical flux of horizontal momentum through the bottom
+  boundary. This is a `ClimaCore.Geometry.AxisTensor` of type
+  `C3{Nothing} ⊗ C12{FT}` (e.g., representing surface stress `τ` as
```
szy21:
```suggestion
  `C3{FT} ⊗ C12{FT}` (e.g., representing surface stress `τ` as
```

src/prognostic_equations/surface_flux.jl
```diff
@@ -1,14 +1,35 @@
...
+  `e_3 ⊗ τ` if defined as flux into the domain, or simply
+  the stress vector `τ` if the `SetValue` operator handles the normal).
+   Conventionally, a positive flux might represent momentum transfer from the
+  atmosphere to the surface (drag).
```
szy21:
```suggestion
   A positive flux might represent momentum transfer from the
  surface to the atmosphere.
```

src/prognostic_equations/surface_flux.jl
```diff
@@ -1,14 +1,35 @@
...
+
+This function constructs a divergence term that is non-zero only in the grid cell
+adjacent to the bottom boundary. The divergence effectively introduces the
+`ρ_flux_uₕ_surface` as a source/sink. The result is divided by density `ᶜρ`
+to yield a tendency for specific horizontal momentum `uₕ`.
+
+Arguments:
+- `ᶜρ`: Cell-center air density field.
+- `ᶜuₕ`: Cell-center horizontal velocity field (used for type/structure, not value in flux calc).
+- `ρ_flux_uₕ_surface`: The vertical flux of horizontal momentum through the bottom
```
szy21, line 23:
```suggestion
- `ρ_flux_uₕ`: The vertical flux of horizontal momentum through the bottom boundary
```
(or change the function argument to ρ_flux_uₕ_surface)

src/prognostic_equations/surface_flux.jl
```diff
@@ -21,9 +42,27 @@ function boundary_tendency_momentum(ᶜρ, ᶜuₕ, ρ_flux_uₕ)
...
+adjacent to the bottom boundary, effectively introducing `ρ_flux_χ_surface` as a
+source/sink. When positive, the flux `ρ_flux_χ_surface` is directed from the surface to 
+the atmosphere, i.e., represents an atmospheric source.  
+
+Arguments:
+- `ᶜχ_field_template`: A cell-center scalar field of the same type as the quantity
+  `χ` for which the tendency is being computed (used for eltype and spatial structure,
+  not its values in the flux calculation).
+- `ρ_flux_χ_surface`: The vertical flux of the scalar quantity `χ` (density-weighted,
+  i.e., flux of `ρχ`) through the bottom boundary. This is a
```
szy21:
```suggestion
- `ᶜχ`: cell-center scalar field (used for eltype and spatial structure,
  not its values in the flux calculation).
- `ρ_flux_χ`: The vertical flux of the scalar quantity `χ` (density-weighted,
  i.e., flux of `ρχ`) through the bottom boundary. This is a
```
(or change the function argument to ρ_flux_χ_surface)

src/prognostic_equations/surface_temp.jl
```diff
@@ -63,11 +129,16 @@ function surface_temp_tendency!(Yₜ, Y, p, t, slab::PrognosticSurfaceTemperatur
...
         end
 
-        # precipitation
+        # 2. Precipitation (rain and snow, defined negative downward, so positive flux 
+        # from surface to atmosphere)
         P_liq = p.precomputed.surface_rain_flux
         P_snow = p.precomputed.surface_snow_flux
 
+        # Total water tendency for surface water:
+        # d(water)/dt = Precip_up + Precip_snow_up + TurbFlux_water_atm_to_sfc
```
szy21:
```suggestion
        # d(water)/dt = -(Precip_up + Precip_snow_up + TurbFlux_water_sfc_to_atm)
```

src/prognostic_equations/surface_temp.jl
```diff
@@ -63,11 +129,16 @@ function surface_temp_tendency!(Yₜ, Y, p, t, slab::PrognosticSurfaceTemperatur
...
         @. Yₜ.sfc.water -= P_liq + P_snow + sfc_turb_w_flux # d(water)/dt = P - E [kg/m²/s]
-
     end
+
+    return nothing
```
szy21: I think @dennisYatunin was suggesting not to add extra `return nothing`. Could you clarify?

src/prognostic_equations/vertical_diffusion_boundary_layer.jl
```diff
@@ -53,8 +110,8 @@ function vertical_diffusion_boundary_layer_tendency!(
         @. ᶜρχₜ_diffusion =
             ᶜdivᵥ_ρχ(-(ᶠinterp(Y.c.ρ) * ᶠinterp(ᶜK_h_scaled) * ᶠgradᵥ(ᶜχ)))
         @. ᶜρχₜ -= ᶜρχₜ_diffusion
-        # Exclude contributions from diffusion of condensate, precipitation 
-        # in mass tendency
+        # Only add contribution from total water diffusion to mass tencency
```
szy21:
```suggestion
        # Only add contribution from total water diffusion to mass tendency 
```

szy21 (approved):
Thanks for adding the docstring. I went through it briefly. I think gemini is a bit too verbose but overall it is pretty good and is helpful. I left a few minor comments.

src/prognostic_equations/remaining_tendency.jl
tapios: Does it matter?

szy21: It shouldn’t. Just curious.

src/prognostic_equations/forcing/large_scale_advection.jl
charleskawczynski: `@. lazy expr` returns a `Base.Broadcast.Broadcasted` object which, when materialized, will be a field.

src/prognostic_equations/edmfx_boundary_condition.jl
dennisYatunin: In general, it is better to avoid type restrictions like this for autodiff, since the types of certain arguments can change depending on whether they also encode partial derivatives. Most functions can extract floating point types from the parameter set (`FT = eltype(params)`), and there is no benefit to restricting the types of their other arguments. In this case, though, the parameter set is unavailable, so you need to make a choice of whether to grab `FT` from `low_percentile` or `high_percentile`. I can't imagine a situation where `low_percentile` will contain partial derivatives and `high_percentile` won't contain derivatives, or vice versa, so here it is okay to restrict their types to the same `FT` so that you don't need to arbitrarily choose one of them.

src/prognostic_equations/edmfx_entr_detr.jl
dennisYatunin: Yes in most cases there is no benefit to writing `FT(0)` or `FT(1)` instead of just 0 or 1. I can only think of two situations where you need the `FT`:
- Divisions, e.g., `1 / 2` is a `Float64`, but `FT(1) / 2` is an `FT`
- Conditionals, e.g., `bool ? value_of_type_FT : 0` is type-unstable when `bool` can't be statically inferred, but `bool ? value_of_type_FT : FT(0)` is type-stable

Most instances of `FT(0)` and `FT(1)` in this file don't fall into either of these categories, so the `FT`s can safely be removed. The are a few exceptions though, like the following conditional statement:
```
vel_diff_term = if elev_above_sfc > eps(FT)
   entr_coeff * abs(ᶜwʲ - ᶜw⁰) / elev_above_sfc
else
    FT(0)
end
```
Here the `FT` should not be removed because it would trigger a type-instability.

src/prognostic_equations/forcing/large_scale_advection.jl
dennisYatunin: Yeah it returns something that acts like a `Field` but isn't quite a `Field`. In this docstring and others you could replace "a ClimaCore Field" with something like "a ClimaCore Field or a lazy broadcast over ClimaCore Fields".

src/prognostic_equations/surface_temp.jl
dennisYatunin: Yes in this case `return nothing` isn't necessary. Sometimes it is needed to avoid type instabilities, though, like in the following example:
```
set_first_if_bool!(array, bool) = if bool
    array[1] = 1
end
```
This function returns `nothing` when `bool == true`, but it returns `1` when `bool == false`. To avoid this type instability, we can rewrite the function in one of two ways:
```
set_first_if_bool!(array, bool) = if bool
    array[1] = 1
    nothing
end
```
```
function set_first_if_bool!(array, bool)
    if bool
        array[1] = 1
    end
    return nothing
end
```
When the last part of a function is a conditional without an `else`-block and the `bool` isn't statically inferrable, `return nothing` may be needed. Otherwise, it can usually be avoided.

src/prognostic_equations/edmfx_entr_detr.jl
tapios: Thanks. So it sound like using `FT(0)` etc. is a good habit then, even if not always necessary.

PR #3841, opened 2025-06-04
ignore .log files

costachris:
gitignore .log files, such as those produced by calibrations


Julians42 (approved):
Support this as it's good to ignore log files produced from the `WorkerBackend` calibration method.

nefrathenrici (approved):

PR #3842, opened 2025-06-05
Add entr_inv_tau to PiGroups entr

costachris:
Add `entr_inv_tau` (additive constant) to PiGroups entrainment closure. This is needed to generate updrafts in the diagnostic EDMF.


nefrathenrici (approved):
Thank you!

PR #3843, opened 2025-06-06
Add sedimentation of rhoa, q_tot and moisture tracers to edmf

trontrytel:
This PR adds the sedimentation terms due to precipitation to rhoa, q_tot and other moisture tracers and enthalpy. Here are the derivation notes. 

PXL_20250606_000906473

TODO:
- [x] Add the terms to `h` advection
- [x] Debug snow in RICO
- [x] Test with TRMM

Next steps
- [ ] Add more `lazy` expressions. For example for the `w` of all the moisture tracers
- [x] Loop over tracers in a more concise way
- [x] Add the one missing term that is currently making the simulations unstable (or understand why we don't want it)
- [ ] Understand what terms need to be added to implicit solver
- [ ] Add additional detrainment like term to move mass from updrafts to the environment to ensure that the environmental values don't go too negative from advection. As is, right now, we don't have any means to ensure positivity in the environment while solving prognostic advection on the grid mean and updraft.

DyCOMS
[summary.pdf](https://github.com/user-attachments/files/21586888/summary.pdf)

RICO
[summary.pdf](https://github.com/user-attachments/files/21586854/summary.pdf)

TRMM
[summary.pdf](https://github.com/user-attachments/files/21586830/summary.pdf)


src/prognostic_equations/advection.jl
```diff
@@ -230,58 +232,111 @@ function edmfx_sgs_vertical_advection_tendency!(
...
-                edmfx_upwinding,
-            )
+            va = vertical_advection(ᶠu³ʲs.:($j), Y.c.sgsʲs.:($j).q_sno, edmfx_upwinding)
             @. Yₜ.c.sgsʲs.:($$j).q_sno += va
-        end
+
+            FT = eltype(params)
+            (; ᶜwₗʲs, ᶜwᵢʲs, ᶜwᵣʲs, ᶜwₛʲs, ᶜtsʲs) = p.precomputed
+
+            ᶜinv_ρ̂_∂ρ̂∂z = (@. lazy(divide_by_ρa(FT(1), Y.c.sgsʲs.:($$j).ρa, FT(0), Y.c.ρ, turbconv_model) * ᶜgradᵥ(ᶠinterp(Y.c.sgsʲs.:($$j).ρa))))
```
tapios: It's probably better to use an upwind gradient here, i.e., use an upwind biased interpolation for the reconstruction.

src/cache/prognostic_edmf_precomputed_quantities.jl
```diff
@@ -608,13 +608,60 @@ NVTX.@annotate function set_prognostic_edmf_precomputed_quantities_precipitation
...
+    FT = eltype(params)
 
     for j in 1:n
+
+        # compute terminal velocity for precipitation
+        @. ᶜwᵣʲs.:($$j) = CM1.terminal_velocity(
+            cmp.pr,
+            cmp.tv.rain,
+            ᶜρʲs.:($$j),
+            max(zero(Y.c.ρ), Y.c.sgsʲs.:($$j).q_rai),
```
tapios, line 557: Wouldn't it be better to have `max(zero...` inside the `terminal_velocity` function, so it does not need to be repeated in every call?

src/prognostic_equations/advection.jl
```diff
@@ -230,58 +232,111 @@ function edmfx_sgs_vertical_advection_tendency!(
...
+            ᶠwₗ³ʲs = (@. lazy( CT3(ᶠinterp(Geometry.WVector(-1 * ᶜwₗʲs.:($$j))))))
+            ᶠwᵢ³ʲs = (@. lazy( CT3(ᶠinterp(Geometry.WVector(-1 * ᶜwᵢʲs.:($$j))))))
+            ᶠwᵣ³ʲs = (@. lazy( CT3(ᶠinterp(Geometry.WVector(-1 * ᶜwᵣʲs.:($$j))))))
+            ᶠwₛ³ʲs = (@. lazy( CT3(ᶠinterp(Geometry.WVector(-1 * ᶜwₛʲs.:($$j))))))
+
+            ᶜa = (@. lazy(draft_area(Y.c.sgsʲs.:($$j).ρa, ᶜρʲs.:($$j))))
+
+            # Flux form vertical advection of rho * area with sedimantation contributions
```
tapios:
```suggestion
            # Flux form vertical advection of rho * area with sedimentation contributions
```

src/prognostic_equations/advection.jl
```diff
@@ -230,58 +232,111 @@ function edmfx_sgs_vertical_advection_tendency!(
...
+            @. Yₜ.c.sgsʲs.:($$j).ρa += vtt
+            vtt = vertical_transport(ᶜρʲs.:($j), ᶠwᵢ³ʲs, (@. lazy(ᶜa * Y.c.sgsʲs.:($$j).q_ice)), dt, edmfx_upwinding)
+            @. Yₜ.c.sgsʲs.:($$j).ρa += vtt
+            vtt = vertical_transport(ᶜρʲs.:($j), ᶠwᵣ³ʲs, (@. lazy(ᶜa * Y.c.sgsʲs.:($$j).q_rai)), dt, edmfx_upwinding)
+            @. Yₜ.c.sgsʲs.:($$j).ρa += vtt
+            vtt = vertical_transport(ᶜρʲs.:($j), ᶠwₛ³ʲs, (@. lazy(ᶜa * Y.c.sgsʲs.:($$j).q_sno)), dt, edmfx_upwinding)
+            @. Yₜ.c.sgsʲs.:($$j).ρa += vtt
+
+            # q_tot and moisture tracers advective form advection with sedimentation velocity
+            va = vertical_advection(ᶠwₗ³ʲs, Y.c.sgsʲs.:($j).q_liq, edmfx_upwinding)
```
tapios: Can we write the following block (and the preceding one above) as a loop of condensate species, so we do not need to repeat the same pattern so many times? (And can also add other condensate species, e.g., graupel, more easily later if we really want to.)

src/cache/prognostic_edmf_precomputed_quantities.jl
```diff
@@ -608,13 +608,60 @@ NVTX.@annotate function set_prognostic_edmf_precomputed_quantities_precipitation
...
     # TODO - can I re-use them between js and env?
     ᶜSᵖ = p.scratch.ᶜtemp_scalar
     ᶜSᵖ_snow = p.scratch.ᶜtemp_scalar_2
 
     n = n_mass_flux_subdomains(p.atmos.turbconv_model)
+    FT = eltype(params)
 
     for j in 1:n
+
+        # compute terminal velocity for precipitation
```
tapios, line 552: Looks like these terminal velocities are defined positive downward, aren't they? Either way, it would be good to make the sign convention clear in a docstring.

tapios (approved):
Basically looks good to me, modulo the minor comments I left (the most important among them may be to use upwinding also for the log(\hat rho) derivative). I also checked the derivations in your notes. They look correct.

src/cache/prognostic_edmf_precomputed_quantities.jl
trontrytel: After talking with @oalcabes and all our issues with clipped values affecting the derivatives for implicit solver, we decided to try to do all our clipping and handling of small negative numbers in Atmos.

I agree that the code looks more clunky, but on the other hand it does not assume that the developer knows what is clipped/not clipped in CloudMicrophysics package

src/prognostic_equations/advection.jl
trontrytel: Yes, I'll do it after I'm done with the PR. It's easier for me to debug when everything is written out like this. But I'll turn it into a concise loop over tracers and velocities for better readability afterwards

src/cache/prognostic_edmf_precomputed_quantities.jl
oalcabes: My argument is a little bit less about the implicit solver, and more that clippings that affect physical behavior should be done at a high level (ie, within Atmos rather than in the other repos) so that we aren't clipping multiple times throughout the codebase & so that it is clear to users what physics is occurring. It stems from me writing a limiter to handle negative values, only to learn that I was never passing in the negative values because values were getting clipped behind the scenes (which was essentially limiting things in a different way than I desired). Happy to think more about it though!

tapios: Sounds good to me!

szy21:
The derivation looks correct to me. I do think the code would be much simpler if you use e.g. wtqt instead of the sum of wjqjs. You will only need wtqt in the equations so there is no dividing by zero problem. You will have the dividing by zero problem in the implicit solver, but writing it as a sum doesn't help with that either. I'll take a look at the code now.

szy21:
Actually even for q_tot you can write everything in wtqt. The issue is only in the implicit solver I think.

src/cache/prognostic_edmf_precomputed_quantities.jl
```diff
@@ -538,13 +538,73 @@ NVTX.@annotate function set_prognostic_edmf_precomputed_quantities_precipitation
...
+        )
+        # compute their contirbutions to energy and total water advection
+        @. ᶜwₜʲs.:($$j) = ifelse(
+            Y.c.sgsʲs.:($$j).ρa * Y.c.sgsʲs.:($$j).q_tot > FT(0),
+            (
+                ᶜwₗʲs.:($$j) * Y.c.sgsʲs.:($$j).q_liq +
+                ᶜwᵢʲs.:($$j) * Y.c.sgsʲs.:($$j).q_ice +
+                ᶜwᵣʲs.:($$j) * Y.c.sgsʲs.:($$j).q_rai +
+                ᶜwₛʲs.:($$j) * Y.c.sgsʲs.:($$j).q_sno
+            ) / Y.c.sgsʲs.:($$j).ρa / Y.c.sgsʲs.:($$j).q_tot,
```
szy21: I don't think this should be divided by ρa?

src/cache/prognostic_edmf_precomputed_quantities.jl
```diff
@@ -538,13 +538,73 @@ NVTX.@annotate function set_prognostic_edmf_precomputed_quantities_precipitation
...
+                ᶜwᵢʲs.:($$j) *
+                Y.c.sgsʲs.:($$j).q_ice *
+                (Iᵢ(thp, ᶜtsʲs.:($$j)) + ᶜΦ) +
+                ᶜwᵣʲs.:($$j) *
+                Y.c.sgsʲs.:($$j).q_rai *
+                (Iₗ(thp, ᶜtsʲs.:($$j)) + ᶜΦ) +
+                ᶜwₛʲs.:($$j) *
+                Y.c.sgsʲs.:($$j).q_sno *
+                (Iᵢ(thp, ᶜtsʲs.:($$j)) + ᶜΦ)
+            ) / Y.c.sgsʲs.:($$j).ρa / abs(Y.c.sgsʲs.:($$j).mse),
```
szy21: same here

src/prognostic_equations/advection.jl
```diff
@@ -422,6 +427,284 @@ function edmfx_sgs_vertical_advection_tendency!(
...
+            ᶠwₗ³ʲs =
+                (@. lazy(CT3(ᶠinterp(Geometry.WVector(-1 * ᶜwₗʲs.:($$j))))))
+            ᶠwᵢ³ʲs =
+                (@. lazy(CT3(ᶠinterp(Geometry.WVector(-1 * ᶜwᵢʲs.:($$j))))))
+            ᶠwᵣ³ʲs =
+                (@. lazy(CT3(ᶠinterp(Geometry.WVector(-1 * ᶜwᵣʲs.:($$j))))))
+            ᶠwₛ³ʲs =
+                (@. lazy(CT3(ᶠinterp(Geometry.WVector(-1 * ᶜwₛʲs.:($$j))))))
+
+            ᶜa = (@. lazy(draft_area(Y.c.sgsʲs.:($$j).ρa, ᶜρʲs.:($$j))))
```
szy21: This is the same as the ᶜa_scalar defined earlier in this file.

src/prognostic_equations/advection.jl
```diff
@@ -422,6 +427,284 @@ function edmfx_sgs_vertical_advection_tendency!(
...
+            (; ᶜwₗʲs, ᶜwᵢʲs, ᶜwᵣʲs, ᶜwₛʲs, ᶜtsʲs) = p.precomputed
+
+            ᶠwₗ³ʲs =
+                (@. lazy(CT3(ᶠinterp(Geometry.WVector(-1 * ᶜwₗʲs.:($$j))))))
+            ᶠwᵢ³ʲs =
+                (@. lazy(CT3(ᶠinterp(Geometry.WVector(-1 * ᶜwᵢʲs.:($$j))))))
+            ᶠwᵣ³ʲs =
+                (@. lazy(CT3(ᶠinterp(Geometry.WVector(-1 * ᶜwᵣʲs.:($$j))))))
+            ᶠwₛ³ʲs =
+                (@. lazy(CT3(ᶠinterp(Geometry.WVector(-1 * ᶜwₛʲs.:($$j))))))
```
szy21: I think it would be a bit confusing if we need to remember ᶠwₗ³ʲs and ᶜwₗʲs have different signs but I'm fine with that.

src/prognostic_equations/advection.jl
```diff
@@ -422,6 +427,284 @@ function edmfx_sgs_vertical_advection_tendency!(
...
+            @. Yₜ.c.sgsʲs.:($$j).q_sno += va
+
+            # mse advective form advection with sedimentation velocity
+            (; ᶜΦ) = p.core
+            thp = CAP.thermodynamics_params(params)
+            ᶜmseₗ = (@. lazy(
+                Y.c.sgsʲs.:($$j).q_liq * (
+                    TD.internal_energy_liquid(thp, ᶜtsʲs.:($$j)) +
+                    TD.gas_constant_air(thp, ᶜtsʲs.:($$j)) *
+                    TD.air_temperature(thp, ᶜtsʲs.:($$j)) +
```
szy21:
```suggestion
                    TD.internal_energy_liquid(thp, ᶜtsʲs.:($$j))
```

szy21: (And I may call this variable qₗmseₗ to be clear but up to you)

src/prognostic_equations/advection.jl
szy21: Could you remove the `s` in the variable name though, as it is for one updraft?

src/prognostic_equations/advection.jl
```diff
@@ -422,6 +427,284 @@ function edmfx_sgs_vertical_advection_tendency!(
...
+                    -1 * Geometry.WVector(ᶜwᵣʲs.:($$j)),
+                    Y.c.sgsʲs.:($$j).ρa,
+                ),
+            ))
+            ᶜ∂ρ̂∂zₛ = (@. lazy(
+                upwind_biased_grad(
+                    -1 * Geometry.WVector(ᶜwₛʲs.:($$j)),
+                    Y.c.sgsʲs.:($$j).ρa,
+                ),
+            ))
```
szy21: I think all these ᶜ∂ρ̂∂z should be the same if all the ws have the same sign.

src/prognostic_equations/advection.jl
```diff
@@ -422,6 +427,284 @@ function edmfx_sgs_vertical_advection_tendency!(
...
+            @. Yₜ.c.sgsʲs.:($$j).q_liq -=
+                dot(
+                    ᶜinv_ρ̂ * ᶜ∂ρ̂∂zₗ,
+                    CT3(Geometry.WVector(-1 * ᶜwₗʲs.:($$j))),
+                ) * Y.c.sgsʲs.:($$j).q_liq
+            @. Yₜ.c.sgsʲs.:($$j).mse -=
+                dot(
+                    ᶜinv_ρ̂ * ᶜ∂ρ̂∂zₗ,
+                    CT3(Geometry.WVector(-1 * ᶜwₗʲs.:($$j))),
+                ) * (Y.c.sgsʲs.:($$j).q_liq * Y.c.sgsʲs.:($$j).mse - ᶜmseₗ)
```
szy21:
```suggestion
            @. Yₜ.c.sgsʲs.:($$j).mse -=
                dot(
                    ᶜinv_ρ̂ * ᶜ∂ρ̂∂zₗ,
                    CT3(Geometry.WVector(-1 * ᶜwₗʲs.:($$j))),
                ) * (ᶜmseₗ - Y.c.sgsʲs.:($$j).q_liq * Y.c.sgsʲs.:($$j).mse)
```
(I think it's a sign error here, but please double check)

src/prognostic_equations/advection.jl
```diff
@@ -422,6 +427,284 @@ function edmfx_sgs_vertical_advection_tendency!(
...
+
+            # mse, q_tot and moisture tracer terms proportional to velocity gradients
+            @. Yₜ.c.sgsʲs.:($$j).q_tot -=
+                ᶜdivᵥ(ᶠwₗ³ʲs) *
+                Y.c.sgsʲs.:($$j).q_liq *
+                (1 - Y.c.sgsʲs.:($$j).q_tot)
+            @. Yₜ.c.sgsʲs.:($$j).q_liq -= ᶜdivᵥ(ᶠwₗ³ʲs) * Y.c.sgsʲs.:($$j).q_liq
+            @. Yₜ.c.sgsʲs.:($$j).mse -=
+                ᶜdivᵥ(ᶠwₗ³ʲs) *
+                (Y.c.sgsʲs.:($$j).q_liq * Y.c.sgsʲs.:($$j).mse - ᶜmseₗ)
```
szy21:
```suggestion
            @. Yₜ.c.sgsʲs.:($$j).mse -=
                ᶜdivᵥ(ᶠwₗ³ʲs) *
                ( ᶜmseₗ - Y.c.sgsʲs.:($$j).q_liq * Y.c.sgsʲs.:($$j).mse)
```

src/prognostic_equations/advection.jl
szy21: Also, you have a Iₗ function defined somewhere and you use it in microphysics I think.

trontrytel:
I learned how to use `unrolled_foreach` As a result the code is MUCH shorter. But still not stable. 

Below are updated notes, so you can more easily match the equation terms with the lines of code.

updated_notes

src/prognostic_equations/advection.jl
```diff
@@ -370,58 +369,97 @@ function edmfx_sgs_vertical_advection_tendency!(
...
         @. Yₜ.c.sgsʲs.:($$j).ρa += vtt
 
-        va = vertical_advection(
-            ᶠu³ʲs.:($j),
-            Y.c.sgsʲs.:($j).mse,
-            edmfx_upwinding,
-        )
-        @. Yₜ.c.sgsʲs.:($$j).mse += va
+        # Advective form advection of mse and q_tot with the grid mean velocity
+        sgs_q_tot_mse = (@name(c.sgsʲs.:(1).mse), @name(c.sgsʲs.:(1).q_tot))
```
szy21, line 384: Maybe add a check for j==1 here and for other places where 1 is hardcoded, so that if anyone accidentally have n > 1, the tendency is not applied more than once. (Or you can move things outside the loop, like we have for the implicit solver, but that's a bigger change I think.)

src/prognostic_equations/advection.jl
```diff
@@ -370,58 +369,97 @@ function edmfx_sgs_vertical_advection_tendency!(
...
+        MatrixFields.unrolled_foreach(sgs_q_tot_mse) do qʲ_name
```
szy21: I prefer something like χʲ_name instead of qʲ_name so that I know it includes mse.

src/prognostic_equations/advection.jl
```diff
@@ -370,58 +369,97 @@ function edmfx_sgs_vertical_advection_tendency!(
...
-                Y.c.sgsʲs.:($j).q_rai,
-                edmfx_upwinding,
+            # TODO - add contibutions to sgs mass flux from tracer sedimentation
+            # TODO - add precipitation and cloud sedimentation in implicit solver/tendency with if/else
+
+            FT = eltype(params)
+            thp = CAP.thermodynamics_params(params)
+            (; ᶜΦ) = p.core
+            (; ᶜtsʲs) = p.precomputed
+            (; ᶜwₜʲs, ᶜwₕʲs) = p.precomputed # TODO - are those used?
```
szy21: they are not used:)

szy21 (approved):
Thanks! The code structure looks much better now.

src/prognostic_equations/advection.jl
szy21: Or just add a comment here, so we know where we need to change if we ever want more than one updraft.

src/prognostic_equations/advection.jl
trontrytel: Yes! One less letter - shorter lines 😻

trontrytel: Agreed on the sign confusion. I'm hoping to soon make it lazy. Then we will not even need to store `ᶜwₗʲs` and we will directly go to the `CT3(ᶠinterp(Geometry.WVector(-1 *` version via `lazy`

src/prognostic_equations/advection.jl
trontrytel: I just defined `Iₗ` locally in one function. I don't think I can use it here, unless we move it to some abbreviations file.

src/prognostic_equations/advection.jl
trontrytel: I wonder if I should just put there `-1` and be done with it. We are assuming that sedimentation velocity only goes down. This would be making more sense if we were summing the updraft velocity with the sedimentation components.

src/prognostic_equations/advection.jl
trontrytel: And they are gone. Also deleted the `rhoa` . Not sure why it was there

src/prognostic_equations/advection.jl
trontrytel: Done!

src/prognostic_equations/advection.jl
```diff
@@ -370,58 +369,127 @@ function edmfx_sgs_vertical_advection_tendency!(
...
-                ᶠu³ʲs.:($j),
-                Y.c.sgsʲs.:($j).q_sno,
-                edmfx_upwinding,
-            )
-            @. Yₜ.c.sgsʲs.:($$j).q_sno += va
+
+            MatrixFields.unrolled_foreach(
+                sgs_microphysics_tracers,
+            ) do (qʲ_name, name, wʲ_name)
+                MatrixFields.has_field(Y, qʲ_name) || return
```
szy21, line 416: Could we remove this? I think it's better that the model fails if we don't have all the tracers when running 1M instead of silently returning. I can remove it in one of my PRs if that sounds good to you.

PR #3844, opened 2025-06-09
Enthalpy perturbation in hyperdiffusion tendency

akshaysridhar:


src/cache/cache.jl
```diff
@@ -129,6 +129,46 @@ function build_cache(
...
+    q_tot_ref = similar(ᶜz)
+    cv_m = similar(ᶜz)
+    R_m = similar(ᶜz)
+    thermo_params = CAP.thermodynamics_params(params)
+    decay_scale_height = FT(8000)
+
+    # Reference State
+    temp_profile = TD.TemperatureProfiles.DecayingTemperatureProfile{FT}(
+        thermo_params,
+        FT(299),  # surface temperature
```
szy21: just curious - why not using 300?

trontrytel: Also interested :)

akshaysridhar: ARM, Gate, Rico use 299 or 299.8 in their surface setups `./src/surface_conditions/surface_setups.jl` so I started with 299 as some representative ref-state. 300K is also fine.

PR #3845, opened 2025-06-10
Reduce dependency on specific cache

charleskawczynski:
This PR defines:

```julia
@inline specific(ρχ, ρ) = ρχ / ρ
```
and attempts to use this functions where they can be easily applied.

### DIFF
--- a/src/cache/precomputed_quantities.jl
+++ b/src/cache/precomputed_quantities.jl
@@ -398,36 +398,34 @@ function thermo_state(
     return get_ts(ρ, p, θ, e_int, q_tot, q_pt)
 end
 
-function thermo_vars(moisture_model, precip_model, specific, K, Φ)
-    energy_var = (; e_int = specific.e_tot - K - Φ)
+function thermo_vars(moisture_model, precip_model, ᶜY, K, Φ)
+    energy_var = (; e_int = specific(ᶜY.ρe_tot, ᶜY.ρ) - K - Φ)
     moisture_var = if moisture_model isa DryModel
         (;)
     elseif moisture_model isa EquilMoistModel
-        (; specific.q_tot)
+        (; q_tot = specific(ᶜY.ρq_tot, ᶜY.ρ))
     elseif moisture_model isa NonEquilMoistModel
-        q_pt_args = (
-            specific.q_tot,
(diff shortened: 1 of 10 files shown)


src/cache/precomputed_quantities.jl
```diff
@@ -4,6 +4,14 @@
 import Thermodynamics as TD
 import ClimaCore: Spaces, Fields
 
+e_tot(ᶜY) = ᶜY.ρe_tot / ᶜY.ρ
+n_liq(ᶜY) = ᶜY.ρn_liq / ᶜY.ρ
```
trontrytel: We should also have `ᶜY.ρn_rai` there somewhere

trontrytel: It may be that we don't use the divided by rho version anywhere. But lets define the divide by function anyway, if we can.

trontrytel:
Thank you!

trontrytel (approved):

src/cache/precomputed_quantities.jl
charleskawczynski: I've adopted the pattern from #3851.

src/utils/variable_manipulations.jl
```diff
@@ -1,3 +1,47 @@
...
+- `ρχ`: The grid-mean density-weighted quantity (e.g., `ρe_tot`, `ρq_tot`).
+- `ρ`: The grid-mean density.
+- `ρaχ`: The density-area-weighted SGS quantity (e.g., `sgs.ρa * sgs.h_tot`).
+- `ρa`: The density-area product of the SGS component.
+- `ρχ_fallback`: The grid-mean density-weighted quantity used for the fallback value.
+- `ρ_fallback`: The grid-mean density used for the fallback value.
+- `turbconv_model`: The turbulence convection model, containing parameters for regularization (e.g., `a_half`).
+"""
+@inline specific(ρχ, ρ) = ρχ / ρ
+
```
tapios, line 44: You have the docstring here for the 2nd (SGS) method, but not the method itself. Might as well add the method too.

charleskawczynski: Ah, yeah, I can add that method too.

PR #3847, opened 2025-06-10
Remove instantiation of default project

charleskawczynski:
This PR removes the instantiation of the default environment, since it seems unused.


nefrathenrici (approved):
Thanks!

PR #3833, opened 2025-06-02
Start using columnwise

charleskawczynski:
@coderabbitai ignore

This PR starts using `columnwise!`, for the initial assignment of `Yₜ = 0`.

It looks like this needs a working version of https://github.com/CliMA/ClimaCore.jl/pull/2343 to work, since I'm seeing (via Cthulhu):

```
Advanced: dump [P]arams cache.
   %3 =  = < semi-concrete eval > getproperty(::StencilBroadcasted{…},::Symbol{…})::…
   %5 =  = < semi-concrete eval > getproperty(::StencilBroadcasted{…},::Symbol{…})::…
   %6 = instantiate_args(::Tuple{…})::…
   %8 =  = < semi-concrete eval > getproperty(::StencilBroadcasted{…},::Symbol{…})::Core.Const(nothing)
   %13 = axes(::StencilBroadcasted{…})::…
 • %31 = device(::FiniteDifferenceSpace{…})::Core.Const(ClimaCore.DeviceSideDevice())
```

In other words, not all broadcasted objects can be instantiated on the gpu with `DeviceSideDevice`.

PR #3852, opened 2025-06-12
Use columnwise for setting tendencies to zero

charleskawczynski:
I'm not getting much traction on #3833, so I'll open one more PR along these lines. This PR uses `columnwise!` to set `Yₜ` to zero. I've tried to cleanup / document functions so that this PR will pass CI and is merge-able.

### DIFF
--- a/src/prognostic_equations/remaining_tendency.jl
+++ b/src/prognostic_equations/remaining_tendency.jl
@@ -37,6 +37,115 @@ NVTX.@annotate function hyperdiffusion_tendency!(Yₜ, Yₜ_lim, Y, p, t)
     apply_hyperdiffusion_tendency!(Yₜ, Y, p, t)
 end
 
+using ClimaCore.RecursiveApply: rzero
+
+#####
+##### Cell center tendencies
+#####
+
+"""
+    ᶜremaining_tendency(ᶜY, ᶠY, p, t)
+
+Returns a Broadcasted object, for evaluating the cell center remaining
+tendency. This method calls `ᶜremaining_tendency(Val(name), ᶜY, ᶠY, p, t)` for
+all `propertynames` of `ᶜY`.
+"""
+function ᶜremaining_tendency(ᶜY, ᶠY, p, t)
+    names = propertynames(ᶜY)
+    tends = construct_tendencies(Val(names), ᶜremaining_tendency, ᶜY, ᶠY, p, t)
+    # We cannot broadcast over a NamedTuple, so we need to check that edge case
+    # first.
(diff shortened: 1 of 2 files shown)


src/utils/utilities.jl
```diff
@@ -496,3 +496,69 @@ function issphere(space)
...
+make_named_tuple(::Val{names}, vals...) where {names} = NamedTuple{names}(vals)
+
+"""
+    add_tend(∑tendencies, tendency)
+
+A helper function which returns `∑tendencies` when `tendency` is a
+`NullBroadcasted` and `lazy.(∑tendencies + tendency)` when `tendency` is not a
+`NullBroadcasted`.
+"""
+function add_tend end
```
glwagner:
```suggestion
function add_tendency end
```

src/utils/utilities.jl
```diff
@@ -496,3 +496,69 @@ function issphere(space)
...
+add_tend(∑tends, ::NullBroadcasted) = ∑tends
+
+"""
+    sub_tend(∑tendencies, tendency)
+
+A helper function which returns `∑tendencies` when `tendency` is a
+`NullBroadcasted` and `lazy.(∑tendencies - tendency)` when `tendency` is not a
+`NullBroadcasted`.
+"""
+function sub_tend end
```
glwagner:
```suggestion
function subtract_tendency end
```

guessing here -- not sure what "sub" represents

src/prognostic_equations/remaining_tendency.jl
```diff
@@ -37,6 +37,99 @@ NVTX.@annotate function hyperdiffusion_tendency!(Yₜ, Yₜ_lim, Y, p, t)
     apply_hyperdiffusion_tendency!(Yₜ, Y, p, t)
 end
 
+using ClimaCore.RecursiveApply: rzero
+
+#####
+##### Cell center tendencies
+#####
+function ᶜremaining_tendency(ᶜY, ᶠY, p, t)
```
glwagner, line 53: I can't figure out what this function does by its name so a docstring might help

src/utils/utilities.jl
charleskawczynski: That is correct. Yeah, perhaps more verbose names here is better.

src/prognostic_equations/remaining_tendency.jl
charleskawczynski: Yeah, I'll add one for this and the other top-level one that is passed to `columnwise!`.

glwagner (commented):
I don't feel qualified to deliver an authoritative review, so I've made a few comments about aesthetics.

PR #3856, opened 2025-06-14
add 2M number adjustment tendencies

sajjadazimi:
## Purpose 
Add 2M number adjustment tendencies

## To-do
Link to CCN budget

----


akshaysridhar:
The sticky CI was due to stale depot_path; this should be OK now!

akshaysridhar:
@sajjadazimi Can you please also add a comment about this PR to the `NEWS.md` file?

.buildkite/Manifest-v1.11.toml
```diff
@@ -1,6 +1,6 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.11.4"
+julia_version = "1.11.5"
```
trontrytel: Could we try to not change the things in the `Manifest` other than what is necessary for your PR. I.e., lets keep the julia version we use for testing to be `11.4`. Or if we choose to update, lets do it in a separate PR that just handles the dependencies.

PR #3857, opened 2025-06-16
Make ᶜspecific lazy and remove matching_subfields

charleskawczynski:
This PR extends `foreach_gs_tracer` to handle specific tracer variables (like the ones in `ᶜ∇²specific_tracers`), and uses it to replace the for-loop over `matching_subfields` in the hyperdiffusion tendency. It also replaces `specific_gs.(Y.c)` with the lazy function `ᶜspecific_gs_tracers(Y)`, which we can use to remove `ᶜspecific` from the cache in a future PR. Since `matching_subfields`, `specific_gs`, and some related functions are no longer needed, their definitions are removed.

This PR also fixes a minor bug and updates an inaccurate code comment in the hyperdiffusion tendency for precipitating species. Details can be found in the [comment](https://github.com/CliMA/ClimaAtmos.jl/pull/3857#issuecomment-3010751311) below.

### DIFF
--- a/src/prognostic_equations/hyperdiffusion.jl
+++ b/src/prognostic_equations/hyperdiffusion.jl
@@ -6,6 +6,22 @@ import ClimaCore.Geometry as Geometry
 import ClimaCore.Fields as Fields
 import ClimaCore.Spaces as Spaces
 
+"""
+    ν₄(hyperdiff, Y)
+
+A `NamedTuple` of the hyperdiffusivity `ν₄_scalar` and the hyperviscosity
+`ν₄_vorticity`. These quantities are assumed to scale with `h^3`, where `h` is
+the mean nodal distance, following the empirical results of Lauritzen et al.
+(2018, https://doi.org/10.1029/2017MS001257). When `h == 1`, these quantities
+are equal to `hyperdiff.ν₄_scalar_coeff` and `hyperdiff.ν₄_vorticity_coeff`.
+"""
+function ν₄(hyperdiff, Y)
+    h = Spaces.node_horizontal_length_scale(Spaces.horizontal_space(axes(Y.c)))
+    ν₄_scalar = hyperdiff.ν₄_scalar_coeff * h^3
+    ν₄_vorticity = hyperdiff.ν₄_vorticity_coeff * h^3
+    return (; ν₄_scalar, ν₄_vorticity)
+end
+
 hyperdiffusion_cache(Y, atmos) = hyperdiffusion_cache(
     Y,
     atmos.hyperdiff,
@@ -34,7 +50,7 @@ function hyperdiffusion_cache(
     gs_quantities = (;
         ᶜ∇²u = similar(Y.c, C123{FT}),
         ᶜ∇²specific_energy = similar(Y.c, FT),
-        ᶜ∇²specific_tracers = remove_energy_var.(specific_gs.(Y.c)),
+        ᶜ∇²specific_tracers = Base.materialize(ᶜspecific_gs_tracers(Y)),
     )
 
     # Sub-grid scale quantities
@@ -85,7 +101,7 @@ NVTX.@annotate function prep_hyperdiffusion_tendency!(Yₜ, Y, p, t)
 
     n = n_mass_flux_subdomains(turbconv_model)
     diffuse_tke = use_prognostic_tke(turbconv_model)
-    (; ᶜp, ᶜspecific) = p.precomputed
+    (; ᶜp) = p.precomputed
     (; ᶜh_ref) = p.core
     (; ᶜ∇²u, ᶜ∇²specific_energy) = p.hyperdiff
     if turbconv_model isa PrognosticEDMFX
@@ -125,20 +141,14 @@ NVTX.@annotate function apply_hyperdiffusion_tendency!(Yₜ, Y, p, t)
     (; hyperdiff, turbconv_model) = p.atmos
     isnothing(hyperdiff) && return nothing
 
-    (; ν₄_vorticity_coeff, ν₄_scalar_coeff, divergence_damping_factor) =
-        hyperdiff
-
-    h_space = Spaces.horizontal_space(axes(Y.c))
-    h_length_scale = Spaces.node_horizontal_length_scale(h_space) # mean nodal distance
-
-    ν₄_scalar = ν₄_scalar_coeff * h_length_scale^3
-    ν₄_vorticity = ν₄_vorticity_coeff * h_length_scale^3
+    (; divergence_damping_factor) = hyperdiff
+    (; ν₄_scalar, ν₄_vorticity) = ν₄(hyperdiff, Y)
 
     n = n_mass_flux_subdomains(turbconv_model)
     diffuse_tke = use_prognostic_tke(turbconv_model)
     ᶜJ = Fields.local_geometry_field(Y.c).J
     point_type = eltype(Fields.coordinate_field(Y.c))
-    (; ᶜp, ᶜspecific) = p.precomputed
+    (; ᶜp) = p.precomputed
     (; ᶜ∇²u, ᶜ∇²specific_energy) = p.hyperdiff
     if turbconv_model isa PrognosticEDMFX
         (; ᶜρa⁰) = p.precomputed
@@ -241,11 +251,12 @@ NVTX.@annotate function prep_tracer_hyperdiffusion_tendency!(Yₜ, Y, p, t)
     (; hyperdiff, turbconv_model) = p.atmos
     isnothing(hyperdiff) && return nothing
 
-    (; ᶜspecific) = p.precomputed
     (; ᶜ∇²specific_tracers) = p.hyperdiff
 
-    for χ_name in propertynames(ᶜ∇²specific_tracers)
-        @. ᶜ∇²specific_tracers.:($$χ_name) = wdivₕ(gradₕ(ᶜspecific.:($$χ_name)))
+    # TODO: Fix RecursiveApply bug in gradₕ to fuse this operation.
+    # ᶜ∇²specific_tracers .= wdivₕ.(gradₕ.(ᶜspecific_gs_tracers(Y)))
+    foreach_gs_tracer(Y, ᶜ∇²specific_tracers) do ᶜρχ, ᶜ∇²χ, _
+        @. ᶜ∇²χ = wdivₕ(gradₕ(specific(ᶜρχ, Y.c.ρ)))
     end
 
     if turbconv_model isa PrognosticEDMFX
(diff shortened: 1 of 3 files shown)


src/prognostic_equations/hyperdiffusion.jl
```diff
@@ -240,11 +240,13 @@ NVTX.@annotate function prep_tracer_hyperdiffusion_tendency!(Yₜ, Y, p, t)
...
-    (; ᶜspecific) = p.precomputed
     (; ᶜ∇²specific_tracers) = p.hyperdiff
 
-    for χ_name in propertynames(ᶜ∇²specific_tracers)
-        @. ᶜ∇²specific_tracers.:($$χ_name) = wdivₕ(gradₕ(ᶜspecific.:($$χ_name)))
+    for ρχ_name in filter(is_tracer_var, propertynames(Y.c))
+        χ_name = specific_name(ρχ_name)
+        ᶜρχ = getproperty(Y.c, ρχ_name)
+        ᶜχ = @. lazy(specific(ᶜρχ, Y.c.ρ))
+        @. ᶜ∇²specific_tracers.:($$χ_name) = wdivₕ(gradₕ(ᶜχ))
```
tapios: The logic here would be cleaner with the construct

``ᶜ∇²specific_tracers = remove_energy_var.(all_specific_gs.(Y.c)),``

from #3851. Please use that unless there's a reason not to.

src/utils/variable_manipulations.jl
```diff
@@ -42,6 +42,20 @@ Arguments:
...
+    elseif ρχ_name == :ρq_liq; return :q_liq
+    elseif ρχ_name == :ρq_ice; return :q_ice
+    elseif ρχ_name == :ρq_rai; return :q_rai
+    elseif ρχ_name == :ρn_liq; return :n_liq
+    elseif ρχ_name == :ρn_rai; return :q_rai
+    elseif ρχ_name == :ρq_sno; return :q_sno
+    else; error("Uncaught name: $ρχ_name")
+    end
+#! format: on
+
```
tapios: Same as above: helpers in #3851 are cleaner.

src/prognostic_equations/hyperdiffusion.jl
charleskawczynski: `ᶜ∇²specific_tracers = remove_energy_var.(all_specific_gs.(Y.c)),` uses broadcast, and would allocate these fields, which we don't want to do here.

If the issue is with the hard-coded names, I recommend we adjust the helpers to look like this:

```julia
"""
    all_specific_gs(gs)

Returns a tuple of symbols which can be used to access grid-mean tracers.

- `gs`: The grid-scale state, which must contain a `:ρ` field and other fields
    with a `:ρ` prefix.

Returns:
- A new `Tuple` containing symbols of only the specific quantities
  (e.g., `:q_tot`, `:e_tot`).
"""
@generated all_specific_gs_names(gs) =
    map(name -> remove_prefix(name, :ρ), relevant_relevant_gs_names(gs))

"""
    all_relevant_gs_names(gs)

Returns a tuple of symbols which are relevant for accessing grid-mean tracers.

- `gs`: The grid-scale state, which must contain a `:ρ` field and other fields
    with a `:ρ` prefix.

Returns:
- A new `Tuple` containing symbols of only the specific quantities
  (e.g., `:ρq_tot`, `:ρe_tot`).
"""
@generated all_relevant_gs_names(gs) =
    filter(name -> has_prefix(name, :ρ) && name != :ρ, Base._nt_names(gs))

"""
    tracer_pairs_gs

An iterator containing tracer names and their mass-weighted counterparts.

For example:
unrolled_foreach(tracer_pairs_gs(ᶜY)) do (χ_name, ρχ_name)
    ᶜρχ = ᶜY.:$(ρχ_name)
    ᶜχ = @. lazy(specific(ᶜρχ, ᶜY.ρ))
end
"""
tracer_pairs_gs(gs::Fields.Field) =
    zip(all_relevant_gs_names(gs), all_specific_gs_names(gs))
```

src/utils/variable_manipulations.jl
```diff
@@ -168,6 +168,33 @@ omitted from the result.
     return :(NamedTuple{$specific_gs_names}(($(specific_gs_values...),)))
 end
 
+"""
```
tapios: You need to remove the duplicative `specific_gs` above. It should be the same thing as `all_specific_gs`.

charleskawczynski: I recommend we leave the duplicate definition until we are able to fully remove `ᶜspecific` from the cache, at which point we can remove `specific_gs` trivially (as nothing will depend on it). If I were to try to remove it right now, it would take me some time to figure out how to make `all_specific_gs` be able to return a field (`all_specific_gs` returns a `NamedTuple` of `Broadcasted` objects) like `specific_gs` does.

tapios:
> > Is it alright if I break this back down into smaller function calls (which worked) and only do the generated function where it's needed? Or maybe @dennisYatunin has a better idea?
> 
> @dennisYatunin ?

As I mentioned, you'd want to use a (corrected form) of `foreach_tracer` here, restricted to grid-scale variables. Then this issue will disappear.

dennisYatunin:
@tapios, I've modified this PR to include the change you requested above, replacing the last use of `matching_subfields` in with an extended version of `foreach_gs_tracer` that can handle specific tracer variables. I also defined a lazy alternative to `specific_gs.(Y.c)`, which we can use to remove `ᶜspecific` from the cache in a future PR. The definitions of `matching_subfields`, `specific_gs`, and related functions have been removed since they are no longer necessary.

In the process of updating the hyperdiffusion tendency, I found a bug on lines [283--298](https://github.com/CliMA/ClimaAtmos.jl/blob/f20481a67210402f233438e572b362415366da2e/src/prognostic_equations/hyperdiffusion.jl#L283-L298) in the current version of `hyperdiffusion.jl`:
```
    ν₄_scalar = ν₄_scalar_coeff * h_length_scale^3
    (; ᶜ∇²specific_tracers) = p.hyperdiff
    for (ᶜρχₜ, ᶜ∇²χ, χ_name) in matching_subfields(Yₜ.c, ᶜ∇²specific_tracers)
        ν₄_scalar = ifelse(
            χ_name in (:q_rai, :q_sno, :n_rai),
            α_hyperdiff_tracer * ν₄_scalar,
            ν₄_scalar,
        )
        ...
    end
```
This loop modifies the value of `ν₄_scalar` whenever `χ_name in (:q_rai, :q_sno, :n_rai)`, changing the diffusion coefficient for all subsequent tracer variables. I've fixed this by assigning a new name `ν₄_scalar_for_χ` to the variable in the loop. @trontrytel, @akshaysridhar, this bugfix may be relevant for you.

I also changed the docstring on lines [301--302](https://github.com/CliMA/ClimaAtmos.jl/blob/f20481a67210402f233438e572b362415366da2e/src/prognostic_equations/hyperdiffusion.jl#L301-L302) to more accurately describe what the code is doing. @tapios, please take a look and let me know if the new version of this comment on lines [300--302](https://github.com/CliMA/ClimaAtmos.jl/blob/df895311301e9ced3d6cb608c8a972abb539e1c4/src/prognostic_equations/hyperdiffusion.jl#L300-L302) is incorrect.

src/prognostic_equations/hyperdiffusion.jl
```diff
@@ -276,31 +277,30 @@ NVTX.@annotate function apply_tracer_hyperdiffusion_tendency!(Yₜ, Y, p, t)
...
-        # Exclude contributions from hyperdiffusion of condensate, 
-        # precipitating species from mass tendency. 
-        if χ_name == :q_tot
+    foreach_gs_tracer(Yₜ, ᶜ∇²specific_tracers) do ᶜρχₜ, ᶜ∇²χ, ρχ_name
+        ν₄_scalar_for_χ =
+            ρχ_name in (@name(ρq_rai), @name(ρq_sno), @name(ρn_rai)) ?
+            ν₄_scalar_for_precip : ν₄_scalar
+        @. ᶜρχₜ -= ν₄_scalar_for_χ * wdivₕ(Y.c.ρ * gradₕ(ᶜ∇²χ))
+
+        # Precipitating species are part of q_tot, so they are included in the
```
tapios:
```suggestion
        # Take into account effect of differential total water diffusion on density
```

src/prognostic_equations/hyperdiffusion.jl
```diff
@@ -276,31 +277,30 @@ NVTX.@annotate function apply_tracer_hyperdiffusion_tendency!(Yₜ, Y, p, t)
...
+        # contribution of tracer hyperdiffusion to the mass tendency, even if
```
tapios:
```suggestion
```

src/prognostic_equations/hyperdiffusion.jl
```diff
@@ -276,31 +277,30 @@ NVTX.@annotate function apply_tracer_hyperdiffusion_tendency!(Yₜ, Y, p, t)
...
+        # they are not hyperdiffused themselves.
```
tapios:
```suggestion
```

src/prognostic_equations/hyperdiffusion.jl
```diff
@@ -276,31 +277,30 @@ NVTX.@annotate function apply_tracer_hyperdiffusion_tendency!(Yₜ, Y, p, t)
     (; hyperdiff, turbconv_model) = p.atmos
     isnothing(hyperdiff) && return nothing
 
+    # When α_hyperdiff_tracer is 0, precipitating species are not hyperdiffused.
     α_hyperdiff_tracer = CAP.α_hyperdiff_tracer(p.params)
     (; ν₄_scalar_coeff) = hyperdiff
     h_space = Spaces.horizontal_space(axes(Y.c))
     h_length_scale = Spaces.node_horizontal_length_scale(h_space) # mean nodal distance
     ν₄_scalar = ν₄_scalar_coeff * h_length_scale^3
```
tapios:
```suggestion
    # Hyperdiffusivity is assumed to scale with `h^3`, where `h` is the mean nodal distance, 
    # following the empirical results of Lauritzen et al. (2018, https://doi.org/10.1029/2017MS001257)
    # Move hyperviscosity and hyperdiffusivity calculation to a function, so that we do not duplicate code (and run the risk of introducing errors, see l. 134).
    ν₄_scalar = ν₄_scalar_coeff * h_length_scale^3
```

src/prognostic_equations/hyperdiffusion.jl
```diff
@@ -276,31 +277,30 @@ NVTX.@annotate function apply_tracer_hyperdiffusion_tendency!(Yₜ, Y, p, t)
     (; hyperdiff, turbconv_model) = p.atmos
     isnothing(hyperdiff) && return nothing
 
+    # When α_hyperdiff_tracer is 0, precipitating species are not hyperdiffused.
```
tapios:
```suggestion
    # α_hyperdiff_tracer rescales the hyperdiffusivity for precipitating species 
```
(Move this and the following line to above l. 286, where this is being used.)

tapios (approved):
LGTM. Good catch on the hyperdiffusivity bug; that's important. 

We should isolate the hyperviscosity/-diffusivity calculation to a function, and determine scalar diffusivity from hyperdiffusivity = hyperviscosity/Prandtl_number_hyperdiff, with the Prandtl number in the parameter file. You may consider introducing a function now, to avoid possible errors from the code duplication and inconsistent edits. We can also do it later.

dennisYatunin:
Ok, I've moved the computation of `ν₄_scalar` and `ν₄_vorticity` into a new function `ν₄(hyperdiff, Y)`, and I added your reference to its docstring. We can change this function to use the Prandtl number in a future PR.

PR #3858, opened 2025-06-16
Remove use of matching subfields

charleskawczynski:


src/parameterized_tendencies/les_sgs_models/smagorinsky_lilly.jl
```diff
@@ -107,19 +107,25 @@ function horizontal_smagorinsky_lilly_tendency!(Yₜ, Y, p, t, ::SmagorinskyLill
...
-        # Rain and snow does not affect the mass
-        if χ_name ∉ (:q_rai, :q_sno)
-            @. Yₜ.c.ρ += ᶜρχₜ_diffusion
+    UU.unrolled_foreach(propertynames(Yₜ.c)) do ρχ_name
+        if is_tracer_var(ρχ_name) && ρχ_name != :ρe_tot
+            ᶜρχ = getproperty(Y.c, ρχ_name)
+            ᶜρχₜ = getproperty(Yₜ.c, ρχ_name)
+            ᶜχ = @. lazy(specific(ᶜρχ, Y.c.ρ))
+            ᶜρχₜ_diffusion =
+                @. p.scratch.ᶜtemp_scalar = wdivₕ(Y.c.ρ * ᶜD_smag * gradₕ(ᶜχ))
```
tapios: Can we get rid of using `p.scratch` here too?

src/parameterized_tendencies/les_sgs_models/smagorinsky_lilly.jl
```diff
@@ -155,19 +161,23 @@ function vertical_smagorinsky_lilly_tendency!(Yₜ, Y, p, t, ::SmagorinskyLilly)
...
+            ᶜρχ = getproperty(Y.c, ρχ_name)
+            ᶜρχₜ = getproperty(Yₜ.c, ρχ_name)
+            ᶜχ = @. lazy(specific(ᶜρχ, Y.c.ρ))
+
+            ᶜdivᵥ_ρχ = Operators.DivergenceF2C(;
+                top = Operators.SetValue(C3(FT(0))),
+                bottom = Operators.SetValue(C3(FT(0))),
+            )
+
+            ᶜ∇ᵥρD∇χₜ = @. ᶜtemp_scalar = ᶜdivᵥ_ρχ(-(ᶠρ * ᶠD_smag * ᶠgradᵥ(ᶜχ)))
```
tapios: Can we get rid of `ᶜtemp_scalar` here by using `+=`?

src/prognostic_equations/advection.jl
```diff
@@ -213,10 +213,14 @@ NVTX.@annotate function explicit_vertical_advection_tendency!(Yₜ, Y, p, t)
...
-    for (ᶜρχₜ, ᶜχ, χ_name) in matching_subfields(Yₜ.c, ᶜspecific)
-        χ_name in (:e_tot, :q_tot) && continue
-        vtt = vertical_transport(ᶜρ, ᶠu³, ᶜχ, float(dt), tracer_upwinding)
-        @. ᶜρχₜ += vtt
+    UU.unrolled_foreach(propertynames(Yₜ.c)) do ρχ_name
+        if is_tracer_var(ρχ_name) && !(ρχ_name in (:ρe_tot, :ρq_tot))
+            ᶜρχ = getproperty(Y.c, ρχ_name)
+            ᶜρχₜ = getproperty(Yₜ.c, ρχ_name)
+            ᶜχ = @. lazy(specific(ᶜρχ, Y.c.ρ))
+            vtt = vertical_transport(ᶜρ, ᶠu³, ᶜχ, float(dt), tracer_upwinding)
```
tapios, line 219: Soon we'll need to remove`ᶠu³` from precomputed too. In that case, wouldn't it be better to apply `@. lazy` to `vertical_transport` (and replace `ᶜχ` here by `specific(ᶜρχ, Y.c.ρ)`)?

src/prognostic_equations/remaining_tendency.jl
```diff
@@ -166,11 +166,15 @@ NVTX.@annotate function additional_tendency!(Yₜ, Y, p, t)
...
-        @. ᶜρχₜ += vst_tracer
-        @. Yₜ.c.ρ += vst_tracer  # TODO: This doesn't look right for all tracers here. Remove?
+    UU.unrolled_foreach(propertynames(Yₜ.c)) do ρχ_name
+        if is_tracer_var(ρχ_name) && ρχ_name != :ρe_tot
+            ᶜρχ = getproperty(Y.c, ρχ_name)
+            ᶜρχₜ = getproperty(Yₜ.c, ρχ_name)
+            ᶜχ = @. lazy(specific(ᶜρχ, Y.c.ρ))
+            vst_tracer = viscous_sponge_tendency_tracer(ᶜρ, ᶜχ, viscous_sponge)
+            @. ᶜρχₜ += vst_tracer
+            @. Yₜ.c.ρ += vst_tracer  # TODO: This doesn't look right for all tracers here. Remove?
```
tapios: @akshaysridhar @szy21 This line looks wrong (adding density corrections for all tracers). Remove or make sure it's only applied to moisture tracers?

src/parameterized_tendencies/les_sgs_models/smagorinsky_lilly.jl
```diff
@@ -107,19 +107,25 @@ function horizontal_smagorinsky_lilly_tendency!(Yₜ, Y, p, t, ::SmagorinskyLill
...
-        ᶜρχₜ_diffusion =
-            @. p.scratch.ᶜtemp_scalar = wdivₕ(Y.c.ρ * ᶜD_smag * gradₕ(ᶜχ))
-        @. ᶜρχₜ += ᶜρχₜ_diffusion
-        # Rain and snow does not affect the mass
-        if χ_name ∉ (:q_rai, :q_sno)
-            @. Yₜ.c.ρ += ᶜρχₜ_diffusion
+    UU.unrolled_foreach(propertynames(Yₜ.c)) do ρχ_name
+        if is_tracer_var(ρχ_name) && ρχ_name != :ρe_tot
+            ᶜρχ = getproperty(Y.c, ρχ_name)
+            ᶜρχₜ = getproperty(Yₜ.c, ρχ_name)
```
tapios: This loop logic here is repeated several times. Encapsulate it in helper function (in `variable_manipulations`).

src/prognostic_equations/remaining_tendency.jl
szy21: Right, we should remove it. So, for this and any other diffusion, we should only apply density correction for qt? (i.e., not ql, qi, qr, qs, or other tracers such as aerosols.)

tapios: Yes, we should only apply it for `q_tot`, so if `ρχ_name == :ρq_tot`.

src/parameterized_tendencies/les_sgs_models/smagorinsky_lilly.jl
```diff
@@ -107,19 +107,25 @@ function horizontal_smagorinsky_lilly_tendency!(Yₜ, Y, p, t, ::SmagorinskyLill
...
+            ᶜχ = @. lazy(specific(ᶜρχ, Y.c.ρ))
+            ᶜρχₜ_diffusion =
+                @. p.scratch.ᶜtemp_scalar = wdivₕ(Y.c.ρ * ᶜD_smag * gradₕ(ᶜχ))
+            @. ᶜρχₜ += ᶜρχₜ_diffusion
+            # Rain and snow does not affect the mass
+            if ρχ_name ∉ (:ρq_rai, :ρq_sno)
```
tapios: This is also not correct. It should be 
```suggestion
            if ρχ_name == :ρq_tot
```

src/parameterized_tendencies/les_sgs_models/smagorinsky_lilly.jl
charleskawczynski: Which loop logic? The `is_tracer_var(ρχ_name) && ρχ_name != :ρe_tot` conditional? Or `ᶜρχ = getproperty(Y.c, ρχ_name)` and `ᶜρχₜ = getproperty(Yₜ.c, ρχ_name)` unpacking? Or both?

src/parameterized_tendencies/les_sgs_models/smagorinsky_lilly.jl
charleskawczynski: Sure, I think that's probably fine

src/parameterized_tendencies/les_sgs_models/smagorinsky_lilly.jl
charleskawczynski: For now I'll just update it to use `lazy` and avoid using `ᶜtemp_scalar`, but that should be possible, yes.

src/prognostic_equations/advection.jl
charleskawczynski: `vertical_transport` already returns a `lazy(expr)`, so no need to wrap that in a lazy call.

charleskawczynski: I'd like to stay focused on `ᶜspecific` for these PRs, so I'm not going to expand the scope to `ᶠu³`, as I know that's a bit of a rabbit hole (I've already explored it in #3833).

src/parameterized_tendencies/les_sgs_models/smagorinsky_lilly.jl
charleskawczynski: Ok, I'll apply this change, but I'll assume that any MSE changes will be attributed to this change.

src/prognostic_equations/remaining_tendency.jl
charleskawczynski: Ok, so this should be:

```julia
vst_tracer = viscous_sponge_tendency_tracer(ᶜρ, ᶜχ, viscous_sponge)
@. ᶜρχₜ += vst_tracer
if ρχ_name == :ρq_tot
    @. Yₜ.c.ρ += vst_tracer
end
```
?

src/parameterized_tendencies/les_sgs_models/smagorinsky_lilly.jl
tapios: If you add the foreach_scalar helper that Dennis wrote [here](https://github.com/CliMA/ClimaAtmos.jl/pull/3851/commits/9ca773ea719566d825090243618ec37544997270), you can just use `foreach_scalar (Yₜ.c, Y.c) do ᶜρχₜ, ᶜρχ, is_ρq_tot` here to replace these lines here and elsewhere.

src/parameterized_tendencies/les_sgs_models/smagorinsky_lilly.jl
```diff
@@ -107,22 +107,27 @@ function horizontal_smagorinsky_lilly_tendency!(Yₜ, Y, p, t, ::SmagorinskyLill
...
-            @. Yₜ.c.ρ += ᶜρχₜ_diffusion
+    for ρχ_name in propertynames(Yₜ.c)
+        if is_tracer_var(ρχ_name) && ρχ_name != :ρe_tot
+            ᶜρχ = getproperty(Y.c, ρχ_name)
+            ᶜρχₜ = getproperty(Yₜ.c, ρχ_name)
+            ᶜχ = @. lazy(specific(ᶜρχ, Y.c.ρ))
+            ᶜρχₜ_diffusion = @. lazy(wdivₕ(Y.c.ρ * ᶜD_smag * gradₕ(ᶜχ)))
+            @. ᶜρχₜ += ᶜρχₜ_diffusion
+            # Rain and snow does not affect the mass
+            if ρχ_name == :ρq_tot
```
tapios: This then becomes
```suggestion
            if is_ρq_tot 
```

src/parameterized_tendencies/les_sgs_models/smagorinsky_lilly.jl
```diff
@@ -107,22 +107,27 @@ function horizontal_smagorinsky_lilly_tendency!(Yₜ, Y, p, t, ::SmagorinskyLill
...
-        @. ᶜρχₜ += ᶜρχₜ_diffusion
-        # Rain and snow does not affect the mass
-        if χ_name ∉ (:q_rai, :q_sno)
-            @. Yₜ.c.ρ += ᶜρχₜ_diffusion
+    for ρχ_name in propertynames(Yₜ.c)
+        if is_tracer_var(ρχ_name) && ρχ_name != :ρe_tot
+            ᶜρχ = getproperty(Y.c, ρχ_name)
+            ᶜρχₜ = getproperty(Yₜ.c, ρχ_name)
+            ᶜχ = @. lazy(specific(ᶜρχ, Y.c.ρ))
+            ᶜρχₜ_diffusion = @. lazy(wdivₕ(Y.c.ρ * ᶜD_smag * gradₕ(ᶜχ)))
```
tapios: I think this is better as (to avoid nested lazy's)
```suggestion
            ᶜρχₜ_diffusion = @. lazy(wdivₕ(Y.c.ρ * ᶜD_smag * gradₕ(specific(ᶜρχ, Y.c.ρ))))
```

src/parameterized_tendencies/les_sgs_models/smagorinsky_lilly.jl
```diff
@@ -155,19 +160,23 @@ function vertical_smagorinsky_lilly_tendency!(Yₜ, Y, p, t, ::SmagorinskyLilly)
...
-            top = Operators.SetValue(C3(FT(0))),
-            bottom = Operators.SetValue(C3(FT(0))),
-        )
-
-        ᶜ∇ᵥρD∇χₜ = @. ᶜtemp_scalar = ᶜdivᵥ_ρχ(-(ᶠρ * ᶠD_smag * ᶠgradᵥ(ᶜχ)))
-        @. ᶜρχₜ -= ᶜ∇ᵥρD∇χₜ
-        # Rain and snow does not affect the mass
-        if χ_name == :q_tot
-            @. Yₜ.c.ρ -= ᶜ∇ᵥρD∇χₜ
+    for ρχ_name in propertynames(Yₜ.c)
```
tapios: And the following block becomes (Dennis's suggestion):
```suggestion
ᶜdivᵥ_ρχ = Operators.DivergenceF2C(;
    top = Operators.SetValue(C3(FT(0))),
    bottom = Operators.SetValue(C3(FT(0))),
)
foreach_tracer(Yₜ, Y) do ᶜρχₜ, ᶜρχ, is_ρq_tot
    ᶜ∇ᵥρD∇χₜ = @. lazy(ᶜdivᵥ_ρχ(-(ᶠρ * ᶠD_smag * ᶠgradᵥ(specific(ᶜρχ, Y.c.ρ)))))
    @. ᶜρχₜ -= ᶜ∇ᵥρD∇χₜ
    if is_ρq_tot
        @. Yₜ.c.ρ -= ᶜ∇ᵥρD∇χₜ
    end
end
```

src/prognostic_equations/remaining_tendency.jl
tapios: Yes (but you can use the `foreach_scalar` helper, which returns a boolean `is_ρq_tot`).

src/prognostic_equations/vertical_diffusion_boundary_layer.jl
```diff
@@ -96,24 +96,28 @@ function vertical_diffusion_boundary_layer_tendency!(
...
-            bottom = Operators.SetValue(C3(0)),
-        )
-        @. ᶜρχₜ_diffusion =
-            ᶜdivᵥ_ρχ(-(ᶠinterp(Y.c.ρ) * ᶠinterp(ᶜK_h_scaled) * ᶠgradᵥ(ᶜχ)))
-        @. ᶜρχₜ -= ᶜρχₜ_diffusion
-        # Only add contribution from total water diffusion to mass tendency 
-        # (exclude contributions from diffusion of condensate, precipitation) 
-        if χ_name == :q_tot
-            @. Yₜ.c.ρ -= ᶜρχₜ_diffusion
+    for ρχ_name in propertynames(Yₜ.c)
```
tapios: Same changes as above

src/prognostic_equations/surface_flux.jl
```diff
@@ -115,17 +115,21 @@ function surface_flux_tendency!(Yₜ, Y, p, t)
...
-        if χ_name == :q_tot
-            @. ρ_flux_χ = sfc_conditions.ρ_flux_q_tot
-        else
-            @. ρ_flux_χ = C3(FT(0))
-        end
-        btt = boundary_tendency_scalar(ᶜχ, ρ_flux_χ)
-        @. ᶜρχₜ -= btt
-        if χ_name == :q_tot
-            @. Yₜ.c.ρ -= btt
+    for ρχ_name in propertynames(Yₜ.c)
```
tapios: Same changes as above

src/prognostic_equations/vertical_diffusion_boundary_layer.jl
```diff
@@ -96,24 +96,28 @@ function vertical_diffusion_boundary_layer_tendency!(
...
+        if is_tracer_var(ρχ_name) && ρχ_name != :ρe_tot
+            ᶜρχ = getproperty(Y.c, ρχ_name)
+            ᶜρχₜ = getproperty(Yₜ.c, ρχ_name)
+            ᶜχ = @. lazy(specific(ᶜρχ, Y.c.ρ))
+            if ρχ_name in (:ρq_rai, :ρq_sno, :ρn_rai)
+                @. ᶜK_h_scaled = α_vert_diff_tracer * ᶜK_h
+            else
+                @. ᶜK_h_scaled = ᶜK_h
+            end
+            ᶜdivᵥ_ρχ = Operators.DivergenceF2C(
```
tapios: Move `ᶜdivᵥ_ρχ` definition out of loop.

src/prognostic_equations/advection.jl
tapios: Ok. My point was not to remove `ᶠu³` here but to use
```suggestion
            vtt = vertical_transport(ᶜρ, ᶠu³, specific(ᶜρχ, Y.c.ρ), float(dt), tracer_upwinding)
```
(and remove l. 220) unless there's a reason we cannot do that.

charleskawczynski: Do you mean `vtt = vertical_transport(ᶜρ, ᶠu³, lazy.(specific.(ᶜρχ, Y.c.ρ)), float(dt), tracer_upwinding)`? I thought that hoisting here kept the `vertical_transport` call a bit simpler to read.

src/parameterized_tendencies/les_sgs_models/smagorinsky_lilly.jl
szy21: @charleskawczynski If you want to keep this PR not behavior changing feel free to leave it as it is. I can go through the code base and make all the changes we need for diffusion once this PR is merged.

charleskawczynski: I think that might be a better way.

src/parameterized_tendencies/les_sgs_models/smagorinsky_lilly.jl
charleskawczynski: Is that going to fix the inference failure? Or is that just to remove code duplication? I'm wondering, will there be a plan to fuse this loop over all of these kernel launches? If so then this would need another refactor and `foreach_scalar` would basically need to be removed. Also, separately, is it worth adding yet another function barrier (more compilation time) for such an operation? This is actually why my latest commit removed the `unroll_foreach` call.

I'd prefer to keep things simple. I feel that these sort of changes are a bit off track of the original purpose of the PR. That said, I'm fine to apply this if this is what you want. Just let me know.

tapios: I think it's crucial to keep the code simple but avoiding duplication. Was there an inference failure here before? I thought that problem was elsewhere.

src/prognostic_equations/advection.jl
tapios: If you say vertical_transport is already lazy, why do we need to nest the lazy's? My intention was to avoid nested lazys that can generate compile problems. The syntax I'm suggesting follows the same pattern we use elsewhere (were we use `.@ lazy` on the outermost layer).

src/parameterized_tendencies/les_sgs_models/smagorinsky_lilly.jl
charleskawczynski: > I think it's crucial to keep the code simple but avoiding duplication. 

Ok, that's fine by me. So, should we bump the allocation limit? Or does @dennisYatunin have a suggestion for fixing inference with this? I'll try adding the `foreach_scalar` to start with, and see if the extra function barrier and `FieldMatrix` types fixes things.

> Was there an inference failure here before? I thought that problem was elsewhere.

Yes-- this PR (at the moment) is introducing inference failures via the loop over the `propertynames`.

tapios: Oh, right, the get property part. This should be fixed with the helper function, and it's one more reason to use it (to avoid repeating the `MatrixFields.get_field` operations that will not be familiar to folks).

src/parameterized_tendencies/les_sgs_models/smagorinsky_lilly.jl
charleskawczynski: What's the aversion to avoiding nested lazy expressions?

tapios: ^^^ @dennisYatunin. My understanding is that they can generate compilation problems if the nesting is too deep.

tapios: In any case, if we don't need them, we should not have them. Question is whether we need them here.

dennisYatunin: Nested lazy expressions should be roughly equivalent to manually inlined expressions from the compiler's perspective. I'm just not a fan of splitting expressions across multiple lines when it can be avoided, but I don't think it makes any practical difference.

charleskawczynski: The only difference is that we re-instantiate the broadcast operation, that's a pretty cheap operation (mostly pointer related). I can inline it, I think we may just get hit by the formatter.

charleskawczynski: I'll inline it.

src/prognostic_equations/vertical_diffusion_boundary_layer.jl
```diff
@@ -96,23 +96,22 @@ function vertical_diffusion_boundary_layer_tendency!(
...
     ᶜK_h_scaled = p.scratch.ᶜtemp_scalar_2
-    for (ᶜρχₜ, ᶜχ, χ_name) in matching_subfields(Yₜ.c, ᶜspecific)
-        χ_name == :e_tot && continue
-        if χ_name in (:q_rai, :q_sno, :n_rai)
+    ᶜdivᵥ_ρχ = Operators.DivergenceF2C(
+        top = Operators.SetValue(C3(0)),
+        bottom = Operators.SetValue(C3(0)),
+    )
+    foreach_scalar(Yₜ, Y) do ᶜρχₜ, ᶜρχ, is_ρq_tot
+        if ρχ_name in (:ρq_rai, :ρq_sno, :ρn_rai)
```
charleskawczynski: @tapios, this is the only remaining unresolved issue (`ρχ_name` is not emitted by `foreach_scalar`), as far as I can tell.

Do you want me to add `ρχ_name` to the args and update all of the calls accordingly?

tapios: Yes, why not.

src/utils/variable_manipulations.jl
```diff
@@ -51,6 +53,64 @@ Arguments:
     return ρa == 0 ? ρχ / ρ : weight * ρaχ / ρa + (1 - weight) * ρχ / ρ
 end
 
+"""
+    scalar_names(field)
```
tapios: One more thing: I changed my mind on the name. `tracer` is better than `scalar`, given density and energy are excluded:
```suggestion
    tracer_names(field)
```

src/utils/variable_manipulations.jl
```diff
@@ -51,6 +53,64 @@ Arguments:
...
+
+- A `Tuple` of `ClimaCore.MatrixFields.FieldName`s corresponding to the tracers.
+"""
+scalar_names(field) =
+    unrolled_filter(MatrixFields.top_level_names(field)) do name
+        !(name in (@name(ρ), @name(ρe_tot), @name(uₕ)))
+    end
+
+"""
+    foreach_scalar(f::F, Yₜ, Y) where {F}
```
tapios:
```suggestion
    foreach_tracer(f::F, Yₜ, Y) where {F}
```
and consistently throughout, `scalar` -> `tracer`

charleskawczynski:
@tapios / @dennisYatunin 

```julia
    foreach_tracer(Yₜ, Y) do ᶜρχₜ, ᶜρχ, ρχ_name, is_ρq_tot
        if !is_ρq_tot
            ᶜχ = @. lazy(specific(ᶜρχ, Y.c.ρ))
            vtt = vertical_transport(ᶜρ, ᶠu³, ᶜχ, float(dt), tracer_upwinding)
            @. ᶜρχₜ += vtt
        end
    end
```
does not seem to be equivalent to the original form:

```julia
    for (ᶜρχₜ, ᶜχ, χ_name) in matching_subfields(Yₜ.c, ᶜspecific)
        χ_name in (:e_tot, :q_tot) && continue
        vtt = vertical_transport(ᶜρ, ᶠu³, ᶜχ, float(dt), tracer_upwinding)
        @. ᶜρχₜ += vtt
    end
```
The error is:

```julia
ERROR: LoadError: MethodError: no method matching /(::@NamedTuple{ρatke::Float32}, ::Float32)
--
  | The function `/` exists, but no method is defined for this combination of argument types.
  |  
  | Closest candidates are:
  | /(::TaylorSeries.TaylorN{T}, ::S) where {T<:Union{Real, Complex}, S<:Union{Real, Complex}}
  | @ TaylorSeries /central/scratch/esm/slurm-buildkite/climaatmos-ci/depot/default/packages/TaylorSeries/azAPg/src/arithmetic.jl:801
  | /(::TaylorSeries.HomogeneousPolynomial{T}, ::S) where {T<:Union{Real, Complex}, S<:Union{Real, Complex}}
  | @ TaylorSeries /central/scratch/esm/slurm-buildkite/climaatmos-ci/depot/default/packages/TaylorSeries/azAPg/src/arithmetic.jl:801
  | /(::TaylorSeries.Taylor1{Rational{T}}, ::S) where {T<:Integer, S<:Union{Real, Complex}}
  | @ TaylorSeries /central/scratch/esm/slurm-buildkite/climaatmos-ci/depot/default/packages/TaylorSeries/azAPg/src/arithmetic.jl:793
  | ...
  |  
  | Stacktrace:
  | [1] specific
...
```
Should I change the condition to `ρχ_name in (@name(e_tot), @name(q_tot)) && continue`?

tapios:
> @tapios / @dennisYatunin
> 
> ```julia
>     foreach_tracer(Yₜ, Y) do ᶜρχₜ, ᶜρχ, ρχ_name, is_ρq_tot
>         if !is_ρq_tot
>             ᶜχ = @. lazy(specific(ᶜρχ, Y.c.ρ))
>             vtt = vertical_transport(ᶜρ, ᶠu³, ᶜχ, float(dt), tracer_upwinding)
>             @. ᶜρχₜ += vtt
>         end
>     end
> ```
> 
> does not seem to be equivalent to the original form:
> 
> ```julia
>     for (ᶜρχₜ, ᶜχ, χ_name) in matching_subfields(Yₜ.c, ᶜspecific)
>         χ_name in (:e_tot, :q_tot) && continue
>         vtt = vertical_transport(ᶜρ, ᶠu³, ᶜχ, float(dt), tracer_upwinding)
>         @. ᶜρχₜ += vtt
>     end
> ```
> 
> The error is:
> 
> ```julia
> ERROR: LoadError: MethodError: no method matching /(::@NamedTuple{ρatke::Float32}, ::Float32)
> --
>   | The function `/` exists, but no method is defined for this combination of argument types.
>   |  
>   | Closest candidates are:
>   | /(::TaylorSeries.TaylorN{T}, ::S) where {T<:Union{Real, Complex}, S<:Union{Real, Complex}}
>   | @ TaylorSeries /central/scratch/esm/slurm-buildkite/climaatmos-ci/depot/default/packages/TaylorSeries/azAPg/src/arithmetic.jl:801
>   | /(::TaylorSeries.HomogeneousPolynomial{T}, ::S) where {T<:Union{Real, Complex}, S<:Union{Real, Complex}}
>   | @ TaylorSeries /central/scratch/esm/slurm-buildkite/climaatmos-ci/depot/default/packages/TaylorSeries/azAPg/src/arithmetic.jl:801
>   | /(::TaylorSeries.Taylor1{Rational{T}}, ::S) where {T<:Integer, S<:Union{Real, Complex}}
>   | @ TaylorSeries /central/scratch/esm/slurm-buildkite/climaatmos-ci/depot/default/packages/TaylorSeries/azAPg/src/arithmetic.jl:793
>   | ...
>   |  
>   | Stacktrace:
>   | [1] specific
> ...
> ```
> 
> Should I change the condition to `ρχ_name in (@name(e_tot), @name(q_tot)) && continue`? (I'm wondering about the utility of `is_ρq_tot` if it's not covering all of the cases.

The problem seems to be that the function treats `ρatke` like other GS variables and then tries to extract `atke`, which doesn't make sense. So we should exclude SGS variables from the `foreach_tracer` loop.

Also, given that `ρχ_name` is now emitted, we should remove the now-redundant `is_ρq_tot` output argument and use `if ρχ_name == :q_tot` instead of `if is_ρq_tot`

tapios:
> > What would be cleaner is positively identifying the GS variables, i.e., those of the form `Y.c.ρ*`, except `ρ` and `ρe_tot`. Could you do that? Then this function could remain unchanged if one day we want to add other categories of tracers.
> 
> A regex-style scan of the `FieldName`'s `name_chain` would require converting the symbols to strings inside a generated function, which I wouldn't recommend. Also, it'd be difficult printing useful error messages with that code as that could cause undefined behavior (which we've run into in the past).
> 
> I kind of get the feeling that we're trying to make too many changes at once. [fe9e66d](https://github.com/CliMA/ClimaAtmos.jl/commit/fe9e66db66bcddc4ffa2c3ba8958da4f66ed30cc) worked, aside from a few jet failures, I feel like it'd be easier to bump the alloc / jet limits, merge that commit, and make these cleanup changes / reduce some duplication later so that we can make progress on the immediate goal of eliminating the dependency on `ᶜspecific`.

As I said, I do not like code duplication; it just creates problems down the line. What we are doing here is not complicated. 

I thought there were problems in the previous commit with `getproperty(Y.c, ρχ_name)`? If not, you can simply use 

``
if is_tracer_var(ρχ_name) && !(ρχ_name in (:ρe_tot, :ρq_tot))Add commentMore actions
            ᶜρχ = getproperty(Y.c, ρχ_name)
            ᶜρχₜ = getproperty(Yₜ.c, ρχ_name)
``
inside a redefined `foreach_tracer`.

src/prognostic_equations/advection.jl
charleskawczynski: Sorry for marking this resolved without answering here-- I think I hadn't refreshed the page to see your last comment. We need the nested lazy because we're removing the (dependence on the) `ᶜspecific` field. The input cannot be a field, because we don't want to materialize the field. Calling `specific(ᶜρχ, Y.c.ρ)` simply won't work because `/(::Field, ::Field)` is not defined outside of the context of a broadcast operation. Even if we did define `/(::Field, ::Field)` outside of the context of a broadcast operation, the result would be a field, which we don't want to materialize before calling `vertical_transport`. In other words, `ᶜχ` must be lazy to pass it into `vertical_transport`. I hope that makes sense.

Regarding the compile time comments, the _only_ difference (including considering compile-time) between
```julia
f = @. lazy(foo(bar(baz(z))))
@. x += div(grad(f))
```
and
```julia
@. x += div(grad(foo(bar(baz(z)))))
```
is that there is an extra call to `Base.Broadcast.instantiate(f)`, which is a very cheap operation.

I hope that clears things up, please let me know if that's not the case.

tapios:
> > As I said, I do not like code duplication; it just creates problems down the line. What we are doing here is not complicated.
> > I thought there were problems in the previous commit with `getproperty(Y.c, ρχ_name)`? If not, you can simply use
> > ```
> > if is_tracer_var(ρχ_name) && !(ρχ_name in (:ρe_tot, :ρq_tot))Add commentMore actions
> >             ᶜρχ = getproperty(Y.c, ρχ_name)
> >             ᶜρχₜ = getproperty(Yₜ.c, ρχ_name)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > inside a redefined `foreach_tracer`.
> > In any case, one way or another we need to replicate what `specific_gs(Y.c)` was doing before, which is extracting grid-scale tracers only. One wouldn't think this is a hard problem.
> 
> In principle, the commit before using/modifying `is_tracer_var` seemed like it should work. Looking at the error from that:
> 
> ```
> MethodError: no method matching /(::@NamedTuple{ρatke::Float32}, ::Float32)
> ```
> 
> It seems like `tracer_names` is somehow including `@name(sgs⁰.ρatke)` in the iterator, which (I think) it shouldn't. I'll have a closer look at MatrixFields to see how `MatrixFields.top_level_names` works.

Ok. Same problem as for `foreach_tracer`, then.

charleskawczynski:
Ok, I was able to reproduce it locally and I see:

```julia
julia> CA.tracer_names(Y.c)
(@name(ρq_tot), @name(sgs⁰))
```
So, I think we want `tracer_names` to not use `top_level_names` and instead use something like the leaves?

charleskawczynski:
> You could also keep using `top_level_names` and just filter out all names that start with `sgs`. There isn't a predefined function that makes an iterator of leaf names in `MatrixFields` (but it should be easy to write if you want it).

As in
```
tracer_names(field) =
    unrolled_filter(MatrixFields.top_level_names(field)) do name
        !(name in (@name(ρ), @name(ρe_tot), @name(uₕ), @name(sgs⁰), @name(sgsʲs)))
    end
```
?

src/utils/variable_manipulations.jl
```diff
@@ -51,6 +53,63 @@ Arguments:
     return ρa == 0 ? ρχ / ρ : weight * ρaχ / ρa + (1 - weight) * ρχ / ρ
 end
 
+"""
+    tracer_names(field)
+
+Filters and returns the names of the variables from a given state
+vector component, excluding `ρ`, `ρe_tot`, and `uₕ`
```
tapios:
```suggestion
vector component, excluding `ρ`, `ρe_tot`, and `uₕ` and SGS fields.
```

src/utils/variable_manipulations.jl
```diff
@@ -51,6 +53,63 @@ Arguments:
...
+"""
+tracer_names(field) =
+    unrolled_filter(MatrixFields.top_level_names(field)) do name
+        !(name in (@name(ρ), @name(ρe_tot), @name(uₕ), @name(sgs⁰), @name(sgsʲs)))
+    end
+
+"""
+    foreach_tracer(f::F, Yₜ, Y) where {F}
+
+Applies a given function `f` to each scalar variable (except `ρ` and  `ρe_tot`)
```
tapios:
```suggestion
Applies a given function `f` to each grid-scale scalar variable (except `ρ` and  `ρe_tot`)
```

src/utils/variable_manipulations.jl
```diff
@@ -51,6 +53,63 @@ Arguments:
...
+in the state `Y` and its corresponding tendency `Yₜ`.
+This utility abstracts the process of iterating over all scalars. It uses
+`tracer_names` to identify the relevant variables and `unrolled_foreach` to
+ensure a performant loop. For each tracer, it calls the provided function `f`
+with the tendency field, the state field, and a boolean flag indicating if
+the current tracer is `ρq_tot` (to allow for special handling).
+
+Arguments:
+
+- `f`: A function to apply to each scalar. It must have the signature `f
```
tapios:
```suggestion
- `f`: A function to apply to each grid-scale scalar. It must have the signature `f
```

tapios (approved):
LGTM. Thanks!

One minor detail: you may wish to consider renaming `foreach_tracer` to foreach_gs_tracer` for added clarity, but it's not crucial.

tapios (approved):
LGTM. Thanks, @charleskawczynski!

PR #3864, opened 2025-06-20
Remove viscosity, diffusivity, and mixing length from precomputed quantities

costachris:
Remove viscosity, diffusivity, and mixing length from precomputed quantities

### DIFF
--- a/src/cache/cloud_fraction.jl
+++ b/src/cache/cloud_fraction.jl
@@ -22,12 +22,9 @@ end
    Compute the grid scale cloud fraction based on sub-grid scale properties
 """
 NVTX.@annotate function set_cloud_fraction!(Y, p, ::DryModel, _)
-    (; ᶜmixing_length) = p.precomputed
     (; turbconv_model) = p.atmos
     FT = eltype(p.params)
-    if isnothing(turbconv_model)
-        compute_gm_mixing_length!(ᶜmixing_length, Y, p)
-    end
+
     p.precomputed.cloud_diagnostics_tuple .=
         ((; cf = FT(0), q_liq = FT(0), q_ice = FT(0)),)
 end
@@ -39,8 +36,9 @@ NVTX.@annotate function set_cloud_fraction!(
 )
     (; params) = p
     (; turbconv_model) = p.atmos
-    (; ᶜts, ᶜmixing_length, cloud_diagnostics_tuple) = p.precomputed
+    (; ᶜts, cloud_diagnostics_tuple) = p.precomputed
     thermo_params = CAP.thermodynamics_params(params)
+
     if isnothing(turbconv_model)
         if p.atmos.call_cloud_diagnostics_per_stage isa
            CallCloudDiagnosticsPerStage
@@ -53,7 +51,6 @@ NVTX.@annotate function set_cloud_fraction!(
             @. ᶜgradᵥ_θ_liq_ice =
                 ᶜgradᵥ(ᶠinterp(TD.liquid_ice_pottemp(thermo_params, ᶜts)))
         end
-        compute_gm_mixing_length!(ᶜmixing_length, Y, p)
     end
     if moist_model isa EquilMoistModel
         @. cloud_diagnostics_tuple = make_named_tuple(
@@ -85,7 +82,7 @@ NVTX.@annotate function set_cloud_fraction!(
 
     FT = eltype(params)
     thermo_params = CAP.thermodynamics_params(params)
-    (; ᶜts, ᶜmixing_length, cloud_diagnostics_tuple) = p.precomputed
+    (; ᶜts, cloud_diagnostics_tuple) = p.precomputed
     (; turbconv_model) = p.atmos
 
     if isnothing(turbconv_model)
@@ -100,8 +97,9 @@ NVTX.@annotate function set_cloud_fraction!(
             @. ᶜgradᵥ_θ_liq_ice =
                 ᶜgradᵥ(ᶠinterp(TD.liquid_ice_pottemp(thermo_params, ᶜts)))
         end
-        compute_gm_mixing_length!(ᶜmixing_length, Y, p)
     end
+
+    ᶜmixing_length = compute_gm_mixing_length(Y, p)
 
     diagnostic_covariance_coeff = CAP.diagnostic_covariance_coeff(params)
     @. cloud_diagnostics_tuple = quad_loop(
@@ -139,12 +137,15 @@ NVTX.@annotate function set_cloud_fraction!(
 
     FT = eltype(params)
     thermo_params = CAP.thermodynamics_params(params)
-    (; ᶜts, ᶜmixing_length, cloud_diagnostics_tuple) = p.precomputed
+    (; ᶜts, cloud_diagnostics_tuple) = p.precomputed
     (; turbconv_model) = p.atmos
 
     # TODO - we should make this default when using diagnostic edmf
     # environment
     diagnostic_covariance_coeff = CAP.diagnostic_covariance_coeff(params)
+
+    ᶜmixing_length_field = p.scratch.ᶜtemp_scalar
+    ᶜmixing_length_field .= ᶜmixing_length(Y, p)
 
     @. cloud_diagnostics_tuple = quad_loop(
         SG_quad,
@@ -152,7 +153,7 @@ NVTX.@annotate function set_cloud_fraction!(
         Geometry.WVector(p.precomputed.ᶜgradᵥ_q_tot),
         Geometry.WVector(p.precomputed.ᶜgradᵥ_θ_liq_ice),
         diagnostic_covariance_coeff,
-        ᶜmixing_length,
+        ᶜmixing_length_field,
         thermo_params,
     )
 
@@ -192,7 +193,7 @@ NVTX.@annotate function set_cloud_fraction!(
 
     FT = eltype(params)
     thermo_params = CAP.thermodynamics_params(params)
-    (; ᶜts⁰, ᶜmixing_length, cloud_diagnostics_tuple) = p.precomputed
+    (; ᶜts⁰, cloud_diagnostics_tuple) = p.precomputed
     (; ᶜρʲs, ᶜtsʲs, ᶜρa⁰, ᶜρ⁰) = p.precomputed
     (; turbconv_model) = p.atmos
 
@@ -200,13 +201,16 @@ NVTX.@annotate function set_cloud_fraction!(
     # environment
     diagnostic_covariance_coeff = CAP.diagnostic_covariance_coeff(params)
 
+    ᶜmixing_length_field = p.scratch.ᶜtemp_scalar
+    ᶜmixing_length_field .= ᶜmixing_length(Y, p)
+
     @. cloud_diagnostics_tuple = quad_loop(
         SG_quad,
         ᶜts⁰,
         Geometry.WVector(p.precomputed.ᶜgradᵥ_q_tot⁰),
         Geometry.WVector(p.precomputed.ᶜgradᵥ_θ_liq_ice⁰),
         diagnostic_covariance_coeff,
-        ᶜmixing_length,
+        ᶜmixing_length_field,
         thermo_params,
     )
 
--- a/src/diagnostics/core_diagnostics.jl
+++ b/src/diagnostics/core_diagnostics.jl
@@ -273,10 +273,40 @@ add_diagnostic_variable!(
     or from mixing length closure with EDMF SGS model.
     """,
     compute! = (out, state, cache, time) -> begin
-        if isnothing(out)
-            return copy(cache.precomputed.ᶜmixing_length)
+        turbconv_model = cache.atmos.turbconv_model
+        # TODO: consolidate remaining mixing length types
+        # (smagorinsky_lilly, dz) into a single mixing length function
+        if isa(turbconv_model, PrognosticEDMFX) ||
+           isa(turbconv_model, DiagnosticEDMFX) ||
+           isa(turbconv_model, EDOnlyEDMFX)
+            ᶜmixing_length_field = ᶜmixing_length(state, cache)
         else
-            out .= cache.precomputed.ᶜmixing_length
+            (; params) = cache
+            (; ᶜlinear_buoygrad, ᶜstrain_rate_norm) = cache.precomputed
+            ᶜdz = Fields.Δz_field(axes(state.c))
+            ᶜprandtl_nvec = @. lazy(
+                turbulent_prandtl_number(
+                    params,
+                    ᶜlinear_buoygrad,
+                    ᶜstrain_rate_norm,
+                ),
+            )
+            ᶜmixing_length_field = @. lazy(
+                smagorinsky_lilly_length(
+                    CAP.c_smag(params),
+                    sqrt(max(ᶜlinear_buoygrad, 0)),   # N_eff
+                    ᶜdz,
+                    ᶜprandtl_nvec,
+                    ᶜstrain_rate_norm,
+                ),
+            )
+        end
+
+
+        if isnothing(out)
+            return Base.materialize(ᶜmixing_length_field)
+        else
+            out .= ᶜmixing_length_field
         end
     end,
 )
--- a/src/diagnostics/edmfx_diagnostics.jl
+++ b/src/diagnostics/edmfx_diagnostics.jl
@@ -1176,10 +1176,12 @@ function compute_lmixw!(
     time,
     turbconv_model::Union{PrognosticEDMFX, DiagnosticEDMFX},
 )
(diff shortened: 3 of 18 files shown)


src/diagnostics/core_diagnostics.jl
```diff
@@ -273,10 +273,38 @@ add_diagnostic_variable!(
     or from mixing length closure with EDMF SGS model.
     """,
     compute! = (out, state, cache, time) -> begin
+        turbconv_model = cache.atmos.turbconv_model
+
+        if isa(turbconv_model, PrognosticEDMFX) ||
+           isa(turbconv_model, DiagnosticEDMFX) ||
+           isa(turbconv_model, EDOnlyEDMFX)
+            ᶜmixing_length = mixing_length(state, cache)
+        else
```
tapios, line 283: It's a little strange to have this calculation of the Smagorinsky-Lilly length here in the diagnostics, rather than where the mixing length is used dynamically. Wouldn't it be preferable for `mixing_length` to return Smag-Lilly when there is no TKE equation?

Maybe this doesn't need to be addressed right away, but if you agree it should be changed in the medium term, leaving a TODO and an explanation of why it is here would help.

Or could you just use `compute_gm_mixing_length` below? I know you are using the cached gradients here, but we want to remove them from the cache anyway.

src/prognostic_equations/gm_sgs_closures.jl
```diff
@@ -61,7 +61,7 @@ Modifies `ᶜmixing_length` in place. Also modifies fields in `p.precomputed`
 (like `ᶜlinear_buoygrad`, `ᶜstrain_rate_norm`) and uses `p.scratch` for
 intermediate calculations.
 """
-NVTX.@annotate function compute_gm_mixing_length!(ᶜmixing_length, Y, p)
+NVTX.@annotate function compute_gm_mixing_length!(Y, p)
```
tapios:
```suggestion
NVTX.@annotate function compute_gm_mixing_length(Y, p)
```
This function no longer modifies the input argument, so shouldn't the exclamation point be dropped?

src/prognostic_equations/implicit/manual_sparse_jacobian.jl
```diff
@@ -543,8 +543,26 @@ function update_jacobian!(alg::ManualSparseJacobian, cache, Y, p, dtγ, t)
...
+        (; ᶜp) = p.precomputed
+        if vert_diff isa DecayWithHeightDiffusion
+            ᶜK_h = compute_eddy_diffusivity_coefficient(Y.c.ρ, vert_diff)
+            ᶜK_u = ᶜK_h
+        elseif vert_diff isa VerticalDiffusion
+            ᶜK_h = compute_eddy_diffusivity_coefficient(Y.c.uₕ, ᶜp, vert_diff)
+            ᶜK_u = ᶜK_h
+        else
+            (; ᶜtke⁰,) = p.precomputed
+            ᶜmixing_length = p.scratch.ᶜtemp_scalar_6
```
tapios: @dennisYatunin Is there some way to encapsulate this recurring materialize in scratch pattern, so it's easy to remove later?

tapios (approved):
LGTM. This was a big lift. Thank you! (Dennis can comment on the question of using scratch storage.)

src/diagnostics/core_diagnostics.jl
costachris: I would like to open a separate PR that isolates the more significant behavior changes, namely just consolidating `gm_mixing_length` into the general `mixing_length` function and changing the default in the cloud fraction closures to use grid-mean (dz) rather than Smagorinsky Lilly. I will add the TODO here.

src/cache/temporary_quantities.jl
```diff
@@ -18,6 +18,8 @@ function temporary_quantities(Y, atmos)
         ᶜtemp_scalar_3 = Fields.Field(FT, center_space),
         ᶜtemp_scalar_4 = Fields.Field(FT, center_space),
         ᶜtemp_scalar_5 = Fields.Field(FT, center_space),
+        ᶜtemp_scalar_6 = Fields.Field(FT, center_space),
```
trontrytel: You should be able to re-use the `temp_scalars` we already have. Unless you need 7 temporary quantities at the same time

src/cache/cloud_fraction.jl
```diff
@@ -100,9 +97,10 @@ NVTX.@annotate function set_cloud_fraction!(
             @. ᶜgradᵥ_θ_liq_ice =
                 ᶜgradᵥ(ᶠinterp(TD.liquid_ice_pottemp(thermo_params, ᶜts)))
         end
-        compute_gm_mixing_length!(ᶜmixing_length, Y, p)
     end
 
+    ᶜmixing_length = compute_gm_mixing_length(Y, p)
```
trontrytel, line 102: Are you missing `@.` or a `lazy @` here? Did you allocate the memory for `ᶜmixing_length`?

src/cache/cloud_fraction.jl
```diff
@@ -139,13 +137,17 @@ NVTX.@annotate function set_cloud_fraction!(
...
     thermo_params = CAP.thermodynamics_params(params)
-    (; ᶜts, ᶜmixing_length, cloud_diagnostics_tuple) = p.precomputed
+    (; ᶜts, cloud_diagnostics_tuple) = p.precomputed
     (; turbconv_model) = p.atmos
 
     # TODO - we should make this default when using diagnostic edmf
     # environment
     diagnostic_covariance_coeff = CAP.diagnostic_covariance_coeff(params)
 
+    ᶜmixing_length = p.scratch.ᶜtemp_scalar_6
```
trontrytel: You can probably reuse the 1-5 temp scalars we already have

src/cache/cloud_fraction.jl
```diff
@@ -139,13 +137,17 @@ NVTX.@annotate function set_cloud_fraction!(
...
+    ᶜmixing_length_lazy = mixing_length(Y, p)
```
trontrytel: Do we need that line? Can't we just write `Base.Broadcast.materialize(mixing_length(Y, p))`?

src/diagnostics/core_diagnostics.jl
```diff
@@ -273,10 +273,39 @@ add_diagnostic_variable!(
...
+            (; params) = cache
+            (; ᶜlinear_buoygrad, ᶜstrain_rate_norm) = cache.precomputed
+            ᶜdz = Fields.Δz_field(axes(state.c))
+            ᶜprandtl_nvec = cache.scratch.ᶜtemp_scalar_2
+            @. ᶜprandtl_nvec = turbulent_prandtl_number(
+                params,
+                ᶜlinear_buoygrad,
+                ᶜstrain_rate_norm,
+            )
+            ᶜmixing_length = @. lazy(
```
trontrytel: Shouldn't this one be materialized, if we actually want to see the values in the output?

src/cache/cloud_fraction.jl
```diff
@@ -139,13 +137,17 @@ NVTX.@annotate function set_cloud_fraction!(
...
+    ᶜmixing_length = Base.Broadcast.materialize(ᶜmixing_length_lazy)
```
dennisYatunin:
```suggestion
    ᶜmixing_length .= mixing_length(Y, p)
```
This line will allocate as it is

src/cache/cloud_fraction.jl
```diff
@@ -192,14 +194,18 @@ NVTX.@annotate function set_cloud_fraction!(
...
     (; ᶜρʲs, ᶜtsʲs, ᶜρa⁰, ᶜρ⁰) = p.precomputed
     (; turbconv_model) = p.atmos
 
     # TODO - we should make this default when using diagnostic edmf
     # environment
     diagnostic_covariance_coeff = CAP.diagnostic_covariance_coeff(params)
 
+    ᶜmixing_length = p.scratch.ᶜtemp_scalar_6
+    ᶜmixing_length_lazy = mixing_length(Y, p)
+    ᶜmixing_length = Base.Broadcast.materialize(ᶜmixing_length_lazy)
```
dennisYatunin: Same as above

src/diagnostics/edmfx_diagnostics.jl
```diff
@@ -1177,9 +1177,9 @@ function compute_lmixw!(
     turbconv_model::Union{PrognosticEDMFX, DiagnosticEDMFX},
 )
     if isnothing(out)
-        return copy(cache.precomputed.ᶜmixing_length_tuple.wall)
+        return copy(mixing_length(state, cache, :wall))
```
trontrytel: Same as above. Why don't we have to materialize here if mixing length is lazy?

src/diagnostics/edmfx_diagnostics.jl
```diff
@@ -1272,10 +1272,18 @@ function compute_edt!(
     vert_diff::Union{VerticalDiffusion, DecayWithHeightDiffusion},
     turbconv_model::Nothing,
 )
+    (; vert_diff) = cache.atmos
+    (; ᶜp) = cache.precomputed
+
+    if vert_diff isa DecayWithHeightDiffusion
+        ᶜK_h = compute_eddy_diffusivity_coefficient(state.c.ρ, vert_diff)
```
trontrytel: Are you missing `@.` here?  Something like `ᶜK_h = @. compute_eddy_diffusivity_coefficient(state.c.ρ, vert_diff)` Is the diagnostic output correct?

src/diagnostics/edmfx_diagnostics.jl
```diff
@@ -1287,10 +1295,16 @@ function compute_edt!(
     vert_diff::Nothing,
     turbconv_model::Union{PrognosticEDMFX, DiagnosticEDMFX},
 )
+    turbconv_params = CAP.turbconv_params(cache.params)
+    (; ᶜtke⁰) = cache.precomputed
+
+    ᶜmixing_length = mixing_length(state, cache)
+    ᶜK_u = eddy_viscosity(turbconv_params, ᶜtke⁰, ᶜmixing_length)
```
trontrytel: same as above

src/diagnostics/core_diagnostics.jl
dennisYatunin: This is materialized a few lines down (`out .= ᶜmixing_length` or `copy(ᶜmixing_length)`)

src/diagnostics/core_diagnostics.jl
```diff
@@ -273,10 +273,39 @@ add_diagnostic_variable!(
...
+        else
+            (; params) = cache
+            (; ᶜlinear_buoygrad, ᶜstrain_rate_norm) = cache.precomputed
+            ᶜdz = Fields.Δz_field(axes(state.c))
+            ᶜprandtl_nvec = cache.scratch.ᶜtemp_scalar_2
+            @. ᶜprandtl_nvec = turbulent_prandtl_number(
+                params,
+                ᶜlinear_buoygrad,
+                ᶜstrain_rate_norm,
+            )
```
dennisYatunin:
```suggestion
            ᶜprandtl_nvec = @. lazy(turbulent_prandtl_number(
                params,
                ᶜlinear_buoygrad,
                ᶜstrain_rate_norm,
            ))
```
I think this can also be done lazily (no intermediate materialization)

src/diagnostics/edmfx_diagnostics.jl
dennisYatunin: `copy` calls `materialize!` internally

src/prognostic_equations/eddy_diffusion_closures.jl
```diff
@@ -527,6 +530,90 @@ function mixing_length(
...
+    (; ᶜtke⁰) = p.precomputed
+    (; ᶜlinear_buoygrad, ᶜstrain_rate_norm) = p.precomputed
+    ᶜz = Fields.coordinate_field(Y.c).z
+    z_sfc = Fields.level(Fields.coordinate_field(Y.f).z, Fields.half)
+    ᶜdz = Fields.Δz_field(axes(Y.c))
+    sfc_tke = Fields.level(ᶜtke⁰, 1)
+
+    ᶜprandtl_nvec_lazy = turbulent_prandtl_number(p)
+    ᶜprandtl_nvec = p.scratch.ᶜtemp_scalar_7
+    ᶜprandtl_nvec = Base.Broadcast.materialize(ᶜprandtl_nvec_lazy)
```
dennisYatunin: Can this also be done lazily?

src/prognostic_equations/eddy_diffusion_closures.jl
```diff
@@ -527,6 +530,90 @@ function mixing_length(
...
+        ᶜstrain_rate_norm,
+        ᶜPr,
+        ᶜtke_exch,
+        scale_blending_method,
+    )
+
+    return getproperty(ᶜmixing_length_tuple, P)
+end
+
+function mixing_length(Y, p, mixing_length_property = :master)
```
dennisYatunin: This is going to be type-unstable if the compiler doesn't feel like constant propagating the `mixing_length_property` symbol. If you want this to be type-stable, I suggest just returning the master mixing length from this function, and defining a separate function `mixing_length_components` that returns a `NamedTuple` of the `wall`, `tke`, and `buoy` components.

src/prognostic_equations/eddy_diffusion_closures.jl
```diff
@@ -612,6 +699,63 @@ function turbulent_prandtl_number(params, ᶜN²_eff, ᶜstrain_rate_norm)
     return min(max(prandtl_nvec, eps_FT), Pr_max)
 end
 
+
+function turbulent_prandtl_number(p)
```
dennisYatunin: It's a little confusing to give this non-pointwise function the same name as its pointwise counterpart above. I would suggest renaming this to something like `ᶜturbulent_prandtl_number`. You can also move the function definition above inside this function (as a closure) in order to keep everything together and avoid cluttering the namespace.

src/prognostic_equations/eddy_diffusion_closures.jl
```diff
@@ -612,6 +699,63 @@ function turbulent_prandtl_number(params, ᶜN²_eff, ᶜstrain_rate_norm)
...
+Returns:
+- The TKE exchange tendency term [m²/s³].
+"""
+function tke_exchange(Y, p)
+    (; turbconv_model) = p.atmos
+    n = n_mass_flux_subdomains(turbconv_model)
+    ᶜρa⁰ =
+        p.atmos.turbconv_model isa PrognosticEDMFX ? p.precomputed.ᶜρa⁰ : Y.c.ρ
+
+    ᶜtke_exch = p.scratch.ᶜtemp_scalar_2
```
dennisYatunin: This can also be done lazily. Here it's a bit trickier because of the loop over `1:n`, but you can implement this as an `unrolled_map`. For example,
```
ᶜtke_exch_summands =
    unrolled_map(UnrolledUtilities.StaticOneTo(n)) do j
            ᶜρaʲ =
                p.atmos.turbconv_model isa PrognosticEDMFX ? Y.c.sgsʲs.:($$j).ρa :
                ᶜρaʲs.:($$j)
            ᶜρaʲ * ᶜdetrʲs.:($$j) / ᶜρa⁰ * (
                1 / 2 * norm_sqr(ᶜinterp(ᶠu³⁰) - ᶜinterp(ᶠu³ʲs.:($$j))) -
                ᶜtke⁰
            )
    end
ᶜtke_exch = @. lazy(+(ᶜtke_exch_summands...))
```
This is a little confusing to read though, so I can add a utility function that generalizes this for you.

trontrytel:
We talked with @dennisYatunin about having an easy naming convention that would allow us to quickly tell apart point-wise functions vs lazy expressions. Think of it like `c` vs `f` naming convention for cell-centers vs cell-faces. I think this would help a lot with PR reviews as we go through the refactor.

Stay tuned for the message from our lead SE on how we could do this!

src/cache/cloud_fraction.jl
costachris: `compute_gm_mixing_length` returns a @. lazy(smagorinsky_lilly_length(...))

src/diagnostics/edmfx_diagnostics.jl
costachris: `compute_eddy_diffusivity_coefficient` returns a @. lazy eddy_diffusivity_coefficient(....) so it should be broadcasted

trontrytel: Yeah. As i mentioned below. We need a naming convention to be able to quickly tell if its a `lazy` function or a field, or a function. Otherwise it is difficult to review this stuff efficiently

src/prognostic_equations/eddy_diffusion_closures.jl
costachris: Making it lazy leads to the ` Kernel invocation uses too much parameter memory` error on the GPU.

PR #3865, opened 2025-06-22
only apply mass correction for qt in viscous sponge

szy21:
## Purpose 

I went through the codebase, and this seems to be the only one left that is not correct. Let me know if there are other places we need to change.

## Content

----


tapios (approved):
LGTM. Thank you!

PR #3870, opened 2025-06-25
Adds Diurnal ERA5 Driven SCM that repeats monthly forcing files 

Julians42:
## Purpose 

Adds a new ERA5 driven case that uses ClimaUtilities to repeat diurnal cycle ERA5 forcing ad nauseam. Case should improve signal to noise by allowing longer time integrals and solve the large data problem associated with downloading global ERA5 hourly data on pressure levels. 

## To-do 
- [ ] Address commented checks to ensure that the configuration is valid
- [ ] Fix ISDAC case: we think it's just an args error
- [ ] Are the names ok? `ReanalysisMonthlyAveragedDiurnal` is long. Utilities names are also long
- [ ] Make sure default yml is up to date

## Content

----
- [ ] - [ ] - [ ] - [ ] - [ ] - [ ] - [ ] - [ ] - [ ] - [ ] - [ ] - [ ] - [ ] - [ ] - [ ] - [ ] - [ ] - [ ] - [ ] - [ ] - [ ] - [ ] - [ ] - [ ] - [ ] - [ ] - [ ] - [ ] - [ ] - [ ] - [ ] - [ ] - [ ] - [ ] - [ ] - [ ] - [ ] - [ ] - [ ] - [ ] - [ ] - [ ]

### DIFF
--- a/docs/src/single_column_prospect.md
+++ b/docs/src/single_column_prospect.md
@@ -1,11 +1,25 @@
 # Single Column Models
-`ClimaAtmos.jl` supports several canonical test cases that are run in a single column model designed to verify how PROSPECT (EDMF) performs against other test cases. These cases include variants of `bomex`, `dycoms`, `rico`, `soares`, and `trmm` and can be found in the `config/model_configs` directory. To run, for example, the BOMEX test case execute the following:
+`ClimaAtmos.jl` supports several canonical test cases that are run in a single column model designed to verify how well PROSPCT (EDMF) is able to reproduce each of the convective schemes. These cases include variants of `bomex`, `dycoms`, `rico`, `soares`, `gabls`, `gate`,and `trmm` and can be found in the `config/model_configs` directory. The purpose of each simulation is summarized in the following table:
+
+| Abbreviation | Long Name | Cloud Regime | Reference |
+|--------------|-----------|--------------|-----------|
+| BOMEX | Barbados Oceanographic and Meteorological Experiment | Marine Cumulus | [Siebesma et al. (2003)](https://doi.org/10.1175/1520-0469(2003)60<1201:ALESIS>2.0.CO;2) |
+| DYCOMS | Dynamics and Chemistry of Marine Stratocumulus | Marine Stratocumulus | [Stevens et al. (2005)](https://doi.org/10.1175/MWR2930.1), [Ackerman et al. (2009)](https://doi.org/10.1175/2008MWR2582.1) |
+| RICO | Rain in Cumulus over the Ocean | Rainy Cumulus | [Rauber et al. (2007)](https://doi.org/10.1175/BAMS-88-12-1912) |
+| SOARES | Shallow Cumulus Convection | Shallow Cumulus | [Soares et al. (2004)](https://doi.org/10.1256/qj.03.223) |
+| GABLS | GEWEX Atmospheric Boundary Layer Study | Dry Convective Boundary Layer | [Kosović & Curry (2000)](https://doi.org/10.1175/1520-0469(2000)057<1052:ALESSO>2.0.CO;2) |
+| TRMM | Tropical Rainfall Measuring Mission | Deep Convection | [Grabowski et al. (2006)](https://doi.org/10.1256/qj.04.147) |
+| ARM_SGP | Atmospheric Radiation Measurement Southern Great Plains | Continental Shallow Cumulus | [Brown et al. (2002)](https://doi.org/10.1256/qj.01.202) |
+| LifeCycle | Life Cycle of Shallow Cumulus | Shallow Cumulus Life Cycle | [Tan et al. (2018)](https://doi.org/10.1002/2017MS001162) |
+| GATE_III | GARP Atlantic Tropical Experiment | Tropical Deep Convection | [Khairoutdinov et al. (2009)](https://doi.org/10.3894/JAMES.2009.1.15) |
+
+For example, to run the BOMEX test case execute the following:
 ```bash
-julia --project=examples examples/hybrid/driver.jl --config_file config/model_configs/prognostic_edmfx_bomex_column.yml --job_id bomex
+julia --project=.buildkite examples/hybrid/driver.jl --config_file config/model_configs/prognostic_edmfx_bomex_column.yml --job_id bomex
 ```
-It may also be helpful to run in interactive mode to be able to examine the simulation object, debug, and develop the code further. To enter debug mode run `julia --project=examples` and then in the REPL run:
+It may also be helpful to run in interactive mode to be able to examine the simulation object, debug, and develop the code further. To enter debug mode run `julia --project=.buildkite` and then in the REPL run:
 ```julia
-using Revise
+using Revise # if you are developing ClimaAtmos
 import ClimaAtmos as CA
 
 # get the configuration arguments
@@ -14,7 +28,7 @@ sol_res = CA.solve_atmos!(simulation) # run the simulation
 ```
 
 ## Externally-Driven Single Column Models
-Currently two versions of the externally driven single column model, `GCM` driven and `ReanalysisTimeVarying` are supported in `ClimaAtmos.jl`. They have been developed specifically for the purpose of realistic simulation and model calibration. Externally-driven means that the model is initialized and forced with data coming from a different simulation. This differs from setups like, for example, BOMEX or SOARES which have steady forcing or functional forcing, respectively.
+Currently three versions of the externally driven single column model, `GCM` driven, `ReanalysisTimeVarying` driven, and `ReanalysisMonthlyAveragedDiurnal` driven are supported in `ClimaAtmos.jl`. Externally-driven means that the model is initialized and forced with data coming from a different simulation. This differs from setups like, for example, BOMEX or SOARES which have steady forcing and low domain tops (~4km) or functional forcing, respectively. They have been developed specifically for the purpose of model calibration by recreating statistics that are close to either LES, for the `GCM` driven case only, or to observations.
 
 ### GCM-Driven Case
 For the `GCM` driven case we can run the experiment using the config file `config/model_configs/prognostic_edmfx_gcmdriven_column.yml` by running:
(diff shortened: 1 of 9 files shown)


akshaysridhar:
ISDAC issue (for starters) is indeed just the args : updating the cache function should fix that 
```external_forcing_cache(Y, external_forcing::ISDACForcing, params, _) = (;)```

There is a GPU error related to climacore column integral ops that may need to be updated (if you're running this case only on CPU this is not an issue).

File: src/utils/era5_observations_to_forcing_file.jl [hunk 1]
```diff
@@ -211,16 +247,24 @@ Note:
...
-    lat,
-    lon,
-    start_date,
+function generate_external_forcing_file(
+    parsed_args,
     forcing_file_path,
     FT;
     smooth_amount = 4,
     time_resolution = FT(3600), # size of accumulated variable period in seconds (3600 for hourly, 86400 for daily and monthly)
     data_dir = @clima_artifact("era5_hourly_atmos_raw"),
```
imreddyTeja: This default will no longer work with the addition of the daily and motnhly subdirs

src/utils/era5_observations_to_forcing_file.jl
```diff
@@ -43,13 +43,48 @@ function get_external_forcing_file_path(
        site_longitude != parsed_args["site_longitude"]
         @info "Rounded site latitude/longitude from ($(parsed_args["site_latitude"]), $(parsed_args["site_longitude"])) to ($(site_latitude), $(site_longitude)) for ERA5 quarter-degree resolution."
     end
-
+    if !isdir(joinpath(data_dir, "daily"))
+        mkdir(joinpath(data_dir, "daily"))
```
imreddyTeja: I don't think this function should automatically create a directory if it does not exist

src/utils/era5_observations_to_forcing_file.jl
```diff
@@ -43,13 +43,48 @@ function get_external_forcing_file_path(
...
+    end
     return joinpath(
         data_dir,
+        "daily",
         "tv_forcing_$(site_latitude)_$(site_longitude)_$(start_date)_$(end_date).nc",
```
imreddyTeja: I think this would be better if the function does not assume the structure of the data folders. So this could instead return 
```
joinpath(
        data_dir,
        "tv_forcing_....
)
```

and changing the default datadir to 
```
get(ENV, "BUILDKITE", "") == "true" ? mktempdir() :
               joinpath(@clima_artifact("era5_hourly_atmos_processed"), daily)
```

``

src/utils/era5_observations_to_forcing_file.jl
```diff
@@ -43,13 +43,48 @@ function get_external_forcing_file_path(
...
+    site_longitude = round(parsed_args["site_longitude"] * 4) / 4
+
+    if site_latitude != parsed_args["site_latitude"] ||
+       site_longitude != parsed_args["site_longitude"]
+        @info "Rounded site latitude/longitude from ($(parsed_args["site_latitude"]), $(parsed_args["site_longitude"])) to ($(site_latitude), $(site_longitude)) for ERA5 quarter-degree resolution."
+    end
+    if !isdir(joinpath(data_dir, "monthly"))
+        mkdir(joinpath(data_dir, "monthly"))
+    end
+    return joinpath(
```
imreddyTeja: Same comments as above

src/solver/model_getters.jl
```diff
@@ -465,10 +471,38 @@ function get_external_forcing_model(parsed_args, ::Type{FT}) where {FT}
...
         end
 
         ExternalDrivenTVForcing{FT}(external_forcing_file)
+    elseif external_forcing == "ReanalysisMonthlyAveragedDiurnal"
+        external_forcing_file =
+            get_external_monthly_forcing_file_path(parsed_args)
+        # ExternalDrivenMonthlyAveragedDiurnalForcing{FT}(external_forcing_file)
+        # generate single file from monthly averaged diurnal data if it doesn't exist
+        # we'll use ClimaUtilities.TimeVaryingInputs downstream to repeat the data. 
+        if !isfile(external_forcing_file)
```
imreddyTeja: It would be nice to  have something check that the external forcing file's time dim is as expected, like you did above

src/utils/era5_observations_to_forcing_file.jl
```diff
@@ -211,16 +247,24 @@ Note:
...
     FT;
     smooth_amount = 4,
     time_resolution = FT(3600), # size of accumulated variable period in seconds (3600 for hourly, 86400 for daily and monthly)
     data_dir = @clima_artifact("era5_hourly_atmos_raw"),
+    data_strs = [
+        "forcing_and_cloud_hourly_profiles",
+        "hourly_inst",
+        "hourly_accum",
+    ],
 )
```
imreddyTeja, line 301: update docstrings to reflect this addition

src/utils/era5_observations_to_forcing_file.jl
```diff
@@ -43,13 +43,48 @@ function get_external_forcing_file_path(
...
 
+"""
+    get_external_monthly_forcing_file_path(parsed_args; data_dir)
+
+Get the path to the external forcing file for a given site and start date.
+When using the BUILDKITE env, a temporary directory is used for the
+external forcing file. Otherwise, the file is expected to stored in the
+era5_hourly_atmos_processed artifact directory.
+"""
+function get_external_monthly_forcing_file_path(
```
imreddyTeja, line 63: This function is not tested, and it's making codecov complain

docs/src/single_column_prospect.md
```diff
@@ -1,11 +1,25 @@
...
+| LifeCycle | Life Cycle of Shallow Cumulus | Shallow Cumulus Life Cycle | [Tan et al. (2018)](https://doi.org/10.1002/2017MS001162) |
+| GATE_III | GARP Atlantic Tropical Experiment | Tropical Deep Convection | [Khairoutdinov et al. (2009)](https://doi.org/10.3894/JAMES.2009.1.15) |
+
+For example, to run the BOMEX test case execute the following:
 ```bash
-julia --project=examples examples/hybrid/driver.jl --config_file config/model_configs/prognostic_edmfx_bomex_column.yml --job_id bomex
+julia --project=.buildkite examples/hybrid/driver.jl --config_file config/model_configs/prognostic_edmfx_bomex_column.yml --job_id bomex
 ```
-It may also be helpful to run in interactive mode to be able to examine the simulation object, debug, and develop the code further. To enter debug mode run `julia --project=examples` and then in the REPL run:
+It may also be helpful to run in interactive mode to be able to examine the simulation object, debug, and develop the code further. To enter debug mode run `julia --project=.buildkite` and then in the REPL run:
```
imreddyTeja: What is the reasoning behind changing the project to buildkite here?

imreddyTeja (changes requested):
The updated docs are really nice. Code cov is complaining about a couple of things, but I think you only need to address the changes in era5_observations_to_forcing_file.jl

docs/src/single_column_prospect.md
Julians42: @imreddyTeja talked with @akshaysridhar and we'll leave it as `project=examples`. I feel like running from `.buildkite` is typically better for development since that's what we're doing integration tests on and so if I run everything from `.buildkite` I have less precompiles. Its definitely more natural for users to run with examples than some hidden file path.

src/utils/era5_observations_to_forcing_file.jl
Julians42: Agreed, but for simplicity I'm still testing it in the `ERA5 single day forcing file generation` test set using the default since I've combined the tests for the daily single and monthly single forcing generation file. LMK if that change resolves this comment or if I'm just missing what your intention is here...

src/solver/model_getters.jl
```diff
@@ -441,10 +446,10 @@ function get_external_forcing_model(parsed_args, ::Type{FT}) where {FT}
...
         ],
     )
-    if any(reanalysis_required_fields .== "ReanalysisTimeVarying")
-        @assert all(reanalysis_required_fields .== "ReanalysisTimeVarying") "All of external_forcing, surface_setup, surface_temperature and initial_condition must be set to ReanalysisTimeVarying."
-        @assert parsed_args["config"] == "column" "ReanalysisTimeVarying is only supported in column mode."
-    end
+    # if any(reanalysis_required_fields .== "ReanalysisTimeVarying")
+    #     @assert all(reanalysis_required_fields .== "ReanalysisTimeVarying") "All of external_forcing, surface_setup, surface_temperature and initial_condition must be set to ReanalysisTimeVarying."
+    #     @assert parsed_args["config"] == "column" "ReanalysisTimeVarying is only supported in column mode."
+    # end
```
imreddyTeja: This should be fixed before merging

src/solver/model_getters.jl
```diff
@@ -453,22 +458,48 @@ function get_external_forcing_model(parsed_args, ::Type{FT}) where {FT}
...
     elseif external_forcing == "ReanalysisTimeVarying"
-        external_forcing_file = get_external_forcing_file_path(parsed_args)
+        external_forcing_file =
+            get_external_daily_forcing_file_path(parsed_args)
         if !isfile(external_forcing_file) ||
-           !check_external_forcing_file_times(
-            external_forcing_file,
-            parsed_args,
-        )
+           !check_daily_forcing_times(external_forcing_file, parsed_args)
```
imreddyTeja, line 465: Both `!check_daily_forcing_times` and `check_monthly_forcing_times` will never return false. If the file fails the checks, an assertion error is thrown rather than returning false

src/utils/era5_observations_to_forcing_file.jl
imreddyTeja: I was referring to the default data_dir, not data_srts. In the test, and everywhere that `generate_external_forcing_file` is called supplies a `data_dir` kwarg.

src/utils/era5_observations_to_forcing_file.jl
```diff
@@ -196,31 +260,37 @@ and should contain 3 files:
...
-- This expansion is necessary for interpolation to the model grid and applies even to surface variables
-    in the current implementation.
+    `2x2x(pressure levels)x(time)` to be able to interpolate to the model grid.
 - The end time of the simulation is inferred from the start date and the simulation time, `t_end`.
 """
-function generate_external_era5_forcing_file(
-    lat,
-    lon,
-    start_date,
+function generate_external_forcing_file(
```
imreddyTeja, line 289: I didn't notice it in the previous PRs, but it is best practice in julia to avoid checking if something is nothing 
with `foo == nothing`. I believe this is because the behavior of nothing equality is inconsistent in older versions of julia.
Generally we use `isnothing(foo)` instead.

Could you change the lon/lat asserts to something like:
```
 @assert !isnothing(lon_index) "informative message"
```

or

```
 @assert Base.notnothing(lon_index) "informative message"
```

imreddyTeja (commented):
I think this is almost ready to merge from a software pov. Have you updated the `era5_hourly_atmos_raw` artifact on hpc and the cluster? 
What about  `era5_hourly_atmos_processed` (need to add monthly and daily subdirs)? 

Do you have a similar script for the daily data? I think we should create a ClimaArtifact folder for `era5_hourly_atmos_raw` with the download scripts (this can be done later). 

The names are long, but unless you can think of shorter names that convey similar info, I'd rather keep the long names than use less informative names

src/utils/era5_observations_to_forcing_file.jl
```diff
@@ -43,19 +46,49 @@ function get_external_forcing_file_path(
...
+    parsed_args;
+    data_dir = get(ENV, "BUILDKITE", "") == "true" ? mktempdir() :
+               joinpath(
+        @clima_artifact("era5_hourly_atmos_processed"),
+        "monthly",
+    ),
+)
+    start_date = parsed_args["start_date"]
+    # round to era5 quarter degree resolution for site selection
+    site_latitude = round(parsed_args["site_latitude"] * 4) / 4
```
costachris, line 73: why is the quarter degree rounding needed?

costachris (approved):

src/utils/era5_observations_to_forcing_file.jl
Julians42: ERA5 is at quarter degree resolution so we just round to the nearest point that ERA5 has and pull that data. Could do some interpolation but since the last PR already smoothed to improve stability I think the effect would be small.

src/utils/era5_observations_to_forcing_file.jl
```diff
@@ -196,31 +272,37 @@ and should contain 3 files:
...
 """
-function generate_external_era5_forcing_file(
-    lat,
-    lon,
-    start_date,
+function generate_external_forcing_file(
+    parsed_args,
     forcing_file_path,
     FT;
+    data_dir,
```
imreddyTeja: super minor, but this could be `input_data_dir` to match `generate_multiday_era5_external_forcing_file`, even though there is no `output_data_dir`.

Also the docstring should be updated to reflect that `data_dir` does not default to anything.

imreddyTeja (approved):
Looks good! I left a comment about two very small possible changes. Remember to clean up the commit history before merging.

PR #3874, opened 2025-06-27
 Restrict compat due to incompatibility with old pkg versions

nefrathenrici:
## Purpose 

Fixes the broken downgrade tests

## Content

- restrict compats for RRTMGP and Thermodynamics to the latest non-breaking version

----


nefrathenrici:
Superseded by #3873

PR #3876, opened 2025-06-30
Up deps: Params, Thermo and Micro

trontrytel:
Update:
- ClimaParams to `10.35`
- CloudMicrophysics to `24.2`
- Thermodynamics to `12.13`


trontrytel:
@juliasloan25 - Are there any other files I should be updating?

Project.toml
```diff
@@ -44,10 +44,10 @@ AtmosphericProfilesLibrary = "0.1.7"
 ClimaComms = "0.6.8"
 ClimaCore = "0.14.34"
 ClimaDiagnostics = "0.2.12"
-ClimaParams = "0.10.32"
+ClimaParams = "0.10.35"
```
juliasloan25, line 47: Is ClimaAtmos now incompatible with previous versions of these 3 packages? Since none of the atmos code was changed in this PR, I expect it's still compatible with the previous versions. If that's true, you can revert the changes in this file and just run `] up` to update the Manifest.

trontrytel: CloudMicrophysics is not compatible with older versions. Two parameter names were changed because of duplication

juliasloan25: If CloudMicrophysics v0.24.2 is only compatible with TD v0.12.13+ and ClimaPrams v0.10.35+, those should be specified in the compat of the CloudMicrophysics v0.24.2 Project.toml. ClimaAtmos is still compatible with any CloudMicrophysics, and it will use the right versions of the other dependencies based on the CloudMicrophysics version it's using.

So I think these need to be specified in CloudMicrophysics, but not in ClimaAtmos.

trontrytel: Ok, so I should ignore my knowledge that it will not work with older versions of ClimaParams and just update the CloudMicrophysics version? I'm just worried that it will lead to errors for other people when they update ClimaParams but not CloudMicrophysics. I think we should ahve made a breaking release of CloudMicrophysics at some point to prevent it from happening

trontrytel: Lets see if it works 🤞

juliasloan25: I'm sorry, you were right 😅. I thought the compat from CloudMicrophysics would be applied in atmos too but it looks like that's not what happens. My bad.

So maybe we add the CloudMicrophysics and ClimaParams compats here, but Thermodynamics can stay as is?

trontrytel: Sure! I'll try that. I totally get trying to have the package range as wide as possible. I really think I should have made a breaking release out there somewhere. 

I'll try what you are suggesting

trontrytel:
```
NonEquilibrium Moisture + 2-moment precipitation RHS terms: Error During Test at /home/runner/work/ClimaAtmos.jl/ClimaAtmos.jl/test/parameterized_tendencies/microphysics/precipitation.jl:172
  Got exception outside of a @test
  MethodError: no method matching cp_m(::Thermodynamics.Parameters.ThermodynamicsParameters{Float32}, ::Float32, ::Float32, ::Float32)
  
```
 
Apologies! Apparently `CloudMicrophysics` also now requires `Thermodynamics` `v12.13`, because the function to compute `cp_m` that avoids `PhasePartition` was only released then: https://github.com/CliMA/Thermodynamics.jl/pull/255

trontrytel: I'll fix the toml files there

PR #3877, opened 2025-06-30
Fix NonEq limiter clipping

oalcabes:


src/parameterized_tendencies/microphysics/microphysics_wrappers.jl
```diff
@@ -121,6 +126,8 @@ function cloud_sources(
 
     qᵥ = qₜ - qₗ - qᵢ - qᵣ - qₛ
 
+    qₛᵢ = TD.q_vap_saturation_from_density(thp, T, ρ, p_vapₛᵢ(thp, FT(T)))
```
trontrytel: No need to do `FT(T)` there. `T` should have correct type.

src/parameterized_tendencies/microphysics/microphysics_wrappers.jl
```diff
@@ -25,6 +25,10 @@ function limit(q, dt, n::Int)
     return max(FT(0), q) / float(dt) / n
 end
 
+# helper functions for finding saturation vapor pressures of liquid and ice
+p_vapₛₗ(tps, T) = TD.saturation_vapor_pressure(tps, T, TD.Liquid())
```
trontrytel: Those functions are only used once. So I would probably vote for not redefining them here, and just using the `TD.saturation_vapor_pressure` directly below

trontrytel (approved):
Thank you! Left some small comments

oalcabes:
Passing noneq test longruns here: https://buildkite.com/clima/climaatmos-gpulongruns/builds/594

PR #3878, opened 2025-07-01
rm pressure, h_tot from cache and begin refactoring velocity cache

akshaysridhar:


tapios:
Thanks, @akshaysridhar. This is a step in the right direction. However, it overlaps with what @costachris is working on: https://github.com/CliMA/ClimaAtmos.jl/compare/main...cc/remove_precomputed_new 

Please coordinate this with each other. It would be good to get Costa's changes in and then build on top of that.

PR #3883, opened 2025-07-03
Add WeatherQuest Atmos IC

Julians42:
## Purpose 

This commit adds functionality to run initial conditions from ERA5 data. We set the ic w=0 as hydrostatic w supplied by era5 is unphysically large for clima nonhydrostatic

## To-do

- [ ] Add artifact to climaartifacts to ensure download script is reproducible see https://github.com/CliMA/WeatherQuest/blob/main/processing/get_initial_conditions.py 
- [ ] add a buildkite job that automatically downloads new data to the artifact so it is available 
- [x] talk with someone on software about appropriate tests for this

## Content

----


config/default_configs/default_config.yml
```diff
@@ -163,6 +163,9 @@ bubble:
 start_date:
   help: "Start date of the simulation"
   value: "20100101"
+start_time:
+  help: "Start time in HHMM format. Used for intializing the weather model"
+  value: "0000"
```
ph-kev: Could we specify that this is only used for initializing the weather model?

Also, is there a reason that we don't include this information in `start_date`? Maybe I missed it, but do we also make the simulation start at `start_date + start_time`?

src/initial_conditions/initial_conditions.jl
```diff
@@ -192,7 +192,20 @@ struct MoistFromFile <: InitialCondition
     file_path::String
 end
 
-function (initial_condition::MoistFromFile)(params)
+"""
+    WeatherModel(start_date, start_time)
+
+An `InitialCondition` that initializes the model with an empty state, and then overwrites
+ it with the content of the weather model. We assume that the weather initial condition 
+ is stored in a correctly named and formatted file in some TODO artifact.
```
ph-kev: Could you add more explanation for what you mean by weather model? I am also not sure if that detail is relevant here, because the data is being loaded from a NetCDF file and any configuration would possibly want to do that. It just happen that we want to do it for the weather model.

Also, it would be nice to write down the assumptions made for the NetCDF file if it is not already there in the function.

src/initial_conditions/initial_conditions.jl
```diff
@@ -192,7 +192,20 @@ struct MoistFromFile <: InitialCondition
...
+ is stored in a correctly named and formatted file in some TODO artifact. 
+"""
+struct WeatherModel <: InitialCondition
+    start_date::String
+    start_time::String
+end
+
+
+function (initial_condition::Union{MoistFromFile, WeatherModel})(params)
     function local_state(local_geometry)
```
ph-kev, line 212: Is there a reason why we can't do
```suggestion
function (initial_condition::InitialCondition)(params)
    function local_state(local_geometry)
```

Note that I am assuming that `InitialCondition` is an abstract type.

toml/diagnostic_edmfx_0M.toml
```diff
@@ -20,4 +20,4 @@ value = 0.12
 value = 0
 
 [max_area_limiter_scale]
-value = 0
+value = 0
```
ph-kev: Could you add a new line here?

src/utils/weather_model.jl
```diff
@@ -0,0 +1,215 @@
...
+    )
+    lon_var = defVar(ncout, "lon", FT, ("lon",), attrib = lon_attrib)
+    lon_var[:] = lon
+
+    lat_attrib = Dict(
+        "Datatype" => string(FT),
+        "standard_name" => "latitude",
+        "long_name" => "latitude",
+        "units" => "degrees_north",
+        "stored_direction" => "decreasing",
```
ph-kev: You don't need to add "Datatype". That information can be found by loading the array and checking the element type of the array.

src/utils/weather_model.jl
```diff
@@ -0,0 +1,215 @@
...
+    return ic_data_path
+end
+
+
+"""
+    to_z_levels(source_file, target_file, target_levels, FT)
+
+Interpolate ERA5 data from native model levels to specified z-levels. Note that 
+to match MoistFromFile, we rename surface pressure, sp, to p. This should change 
+for consistency if we write a new initial condition dispatch.
```
ph-kev: It would be nice if you can write what one would need to implement for a new initial condition.

src/utils/weather_model.jl
```diff
@@ -0,0 +1,215 @@
+using NCDatasets
+using Dates
+import ClimaInterpolations.Interpolation1D: interpolate1d!, Linear, Flat
+
+"""
+    weather_model_data_path(start_date, start_time)
+
+Get the path to the weather model data for a given start date and time.
+If the data is not found, will attempt to generate it from raw data. If 
+the raw data is not found, throw an error
```
ph-kev: Is there a reason why we would prefer to compute this on the fly rather than compute all the files offline?

src/utils/weather_model.jl
```diff
@@ -0,0 +1,215 @@
...
+    v_var =
+        defVar(ncout, "v", FT, ("lon", "lat", "z"), attrib = ncin["v"].attrib)
+    v_var[:, :, :] =
+        interpz_3d(target_levels, source_z, FT.(ncin["v"][:, :, :, 1]))
+
+    # ERA5 w is from a hydrostatic model and so isn't meaningful for ClimaAtmos
+    # See https://agupubs.onlinelibrary.wiley.com/doi/full/10.1002/2017MS001059
+    w_var =
+        defVar(ncout, "w", FT, ("lon", "lat", "z"), attrib = ncin["w"].attrib)
+    w_var[:, :, :] .= FT(0)
```
ph-kev: You can use `dimnames` instead and I think you can use this everywhere if you are not permuting the dimensions of the array.

```suggestion
        defVar(ncout, "w", FT, dimnames(ncin["w"]), attrib = ncin["w"].attrib)
    w_var[:, :, :] .= FT(0)
```

src/utils/weather_model.jl
```diff
@@ -0,0 +1,215 @@
...
+    # Read and cast variables to FT type, replacing missing values with NaN
+    z_raw = ncin["z"][:, :, :, 1]
+    source_z = FT.(coalesce.(z_raw, NaN)) ./ grav # convert from geopotential height to height
+
+    # Create output file
+    ncout = NCDataset(target_file, "c", attrib = copy(ncin.attrib))
+
+    # Define dimensions
+    defDim(ncout, "lon", size(lon, 1))
+    defDim(ncout, "lat", size(lat, 1))
```
ph-kev: You should be able to use `length`.

```suggestion
    defDim(ncout, "lon", length(lon))
    defDim(ncout, "lat", length(lat))
```

src/utils/weather_model.jl
```diff
@@ -0,0 +1,215 @@
...
+end
+
+"""
+    interpz_3d(ztarget, zsource, fsource)
+
+Interpolate 3D field `fsource` from 3D source levels `zsource` to 1D target levels `ztarget`.
+"""
+function interpz_3d(ztarget, zsource, fsource)
+    nx, ny, nz = size(zsource)
+    # permute dimensions from (nx, ny, nz) to (nz, nx, ny) if needed
```
ph-kev, line 187: Would it be possible to add error handling here? I think you can check the dimensions of the variable and check it is what you expected.

src/initial_conditions/initial_conditions.jl
Julians42: The reason is that there are lots of other subtypes that require different setups. E.g., on the following lines we have: 
```julia
function (initial_condition::DryDensityCurrentProfile)(params)
    (; perturb) = initial_condition
    function local_state(local_geometry)
        FT = eltype(params)
        grav = CAP.grav(params)
        ndims = length(propertynames(local_geometry.coordinates))
        (; x, z) = local_geometry.coordinates
        x_c = FT(25600)
        x_r = FT(4000)
....
```

src/utils/weather_model.jl
Julians42: Ah yes this docstring is unclear - I cleaned it up. Basically if we wanted to copy paste the code in initial_conditions for dyamond `_overwrite_initial_conditions_from_file!` function we could do this, or generalize the names of the variables written.

src/utils/weather_model.jl
Julians42: @nefrathenrici @costachris Thoughts on this? the alternative is a buildkite script that runs automatically and generates the data and then does this processing. 
My thought would be data processing in atmos (so we can change / modify as needed -e.g., the w zeroing) and then buildkite just to get the raw era5 data.

nefrathenrici: I think it depends how long this takes, could you time this function? If it's reasonably fast then I think we should compute it online.

config/default_configs/default_config.yml
nefrathenrici: I think this argument is overly specific. Do we need this feature or could we just start a simulation at 0000 by default? Alternatively as @ph-kev said we should just bundle the start_time with start_date and parse it differently if the time is included: Something like "20100101-0000"

Project.toml
```diff
@@ -44,6 +45,7 @@ AtmosphericProfilesLibrary = "0.1.7"
 ClimaComms = "0.6.8"
 ClimaCore = "0.14.34"
 ClimaDiagnostics = "0.2.12"
+ClimaInterpolations = "0.1.0"
```
nefrathenrici, line 48: Glad to see we are using ClimaInterpolations.

src/callbacks/callbacks.jl
```diff
@@ -215,7 +215,7 @@ function set_insolation_variables!(Y, p, t, tvi::TimeVaryingInsolation)
     max_zenith_angle = FT(π) / 2 - eps(FT)
     irradiance = FT(CAP.tot_solar_irrad(params))
     au = FT(CAP.astro_unit(params))
-    # TODO: Where does this date0 come from?
+    # date0 references the start of the astronomial epoch J2000
```
nefrathenrici, line 218: Thanks for commenting this

toml/diagnostic_edmfx_era5_ic.toml
```diff
@@ -0,0 +1,29 @@
+[precipitation_timescale]
```
nefrathenrici, line 1: Why is this parameter file included? Do we need it for model stability?

nefrathenrici (commented):
This looks good to me overall, there are just a few things to simplify

toml/diagnostic_edmfx_era5_ic.toml
Julians42: Yes exactly - we needed to add a rayleigh sponge to keep vertical velocities at the model grid top from being large (simulation was crashing after an hour). The config is essentially the amip toml but with the sponge added. We could consider playing with how much of the domain is sponged @costachris maybe 20km from domain top (which is 28km high for the current config) is too low.

src/utils/weather_model.jl
costachris: a benefit of doing it on the fly is we can select an arbitrary date/time from the reanalysis record and generate a forecast in one go.

costachris (approved):

src/utils/weather_model.jl
Julians42: With ClimaInterpolations it takes about 15 seconds to generate (5-10x slower with Interpolations). @nefrathenrici that's short enough yeah?

src/initial_conditions/initial_conditions.jl
```diff
@@ -192,7 +192,22 @@ struct MoistFromFile <: InitialCondition
...
-function (initial_condition::MoistFromFile)(params)
+"""
+    WeatherModel(start_date)
+
+An `InitialCondition` that initializes the model with an empty state, and then overwrites
+it with the content of a NetCDF file that contains the initial conditions, stored in the 
+artifact `weather_model_ic`/raw/era5_raw_YYYYMMDD_HHMM.nc. We interpolate the initial 
+conditions from ERA5 pressure level grid to a z grid, saving to the artifact 
+`weather_model_ic`/init/era5_init_YYYYMMDD_HHMM.nc. It is then interpolated to the model
+ grid in `_overwrite_initial_conditions_from_file!`, which documents the required variables.
```
ph-kev: I think you meant `weather_model_ic/init/era5_init_YYYYMMDD_HHMM.nc`.

src/utils/weather_model.jl
nefrathenrici: I think that's acceptable.

nefrathenrici (approved):
Looks good, I think you just need to add `using Dates` in `test/utilities.jl`

ph-kev (changes requested):

src/utils/weather_model.jl
```diff
@@ -0,0 +1,186 @@
...
+
+Interpolate ERA5 data from native model levels to specified z-levels. Note that 
+to use _overwrite_initial_conditions_from_file! we rename surface pressure, sp, 
+to p. This allows us to share functionality with the Dyamond setup.
+
+# Arguments
+- `source_file` / `target_file`: Path to input / output NetCDF 
+- `target_levels`: Vector of target altitude levels (in meters)
+
+"""
```
ph-kev: Can you add the assumptions that you are making for `source_file` or add some error handling? I don't want bugs to appear because the ERA5 data changes for whatever reason.

PR #3884, opened 2025-07-03
Reduce cache: lazy velocity computations 

akshaysridhar:


szy21:
What's the status of this PR?

PR #3889, opened 2025-07-09
Improve CAM_SE hyperdiff Robustness

### LINKED ISSUE #3888: Divergence Damping Factor overwritten in CAM_SE hyperdiffusion.
EDIT: After talking with @akshaysridhar , we simply need to use `ClimaHyperdiffusion` and then throw a error when the user tries to pass the 3 coefficients using CAM_SE. Also need to update the documentation and default_config to make this more clear (at least I got confused). 

`divergence_damping_factor` is useful for damping gravity waves in global simulations and is currently not able to be modified if using the default `CAM_SE` hyperdiffusion setup. Perhaps we should also make the vorticity coefficients and scalars accessible @tapios ?
In `default_config.yml` we have
```yml
divergence_damping_factor:
  help: "Divergence damping factor"
  value: 1.0
```

in `src/solver_model_getters.jl` line 60:
```julia
    elseif hyperdiff_name in ("CAM_SE",)
        # To match hyperviscosity coefficients in:
        #    https://agupubs.onlinelibrary.wiley.com/doi/epdf/10.1029/2017MS001257
        #    for equation A18 and A19
        # Need to scale by (1.1e5 / (sqrt(4 * pi / 6) * 6.371e6 / (3*30)) )^3  ≈ 1.238
        @info "Using CAM_SE hyperdiffusion. vorticity_hyperdiffusion_coefficient, \
               scalar_hyperdiffusion_coefficient and divergence_damping_factor in the config \
               will be ignored."
        ν₄_vorticity_coeff = FT(0.150 * 1.238)
        ν₄_scalar_coeff = FT(0.751 * 1.238)
        divergence_damping_factor = FT(5)
        return ClimaHyperdiffusion(;
            ν₄_vorticity_coeff,
            ν₄_scalar_coeff,
            divergence_damping_factor,
        )
```
While we do throw a warning it would be better practice - and useful for the weather model ;) - to set this in the config.

Julians42:
## Purpose 

Improve the robustness of error handling for the CAM_SE hyperdiffusion setup and fixes a bug where the yaml does not accurately reflect the hyperdiffusion coefficients used to run the simulation. As we have access to limited values in the function, to do this we need to change the values in default config to the CAM_SE values as suggested by @akshaysridhar . 

Closes #3888 

## To-do

- [ ] Does the atol seem appropriate?
- [ ] Should we test more variants in the test or are these two sufficient?
- [ ] Could we be more succinct with the checks in model_getters? I was trying to write most clearly

## Content

----

### DIFF
--- a/config/default_configs/default_config.yml
+++ b/config/default_configs/default_config.yml
@@ -29,10 +29,10 @@ z_max:
   value: 30000.0
 vorticity_hyperdiffusion_coefficient:
   help: "Hyperdiffusion coefficient for vorticity (m s-1)"
-  value: 1.5
+  value: 0.1857
 scalar_hyperdiffusion_coefficient:
   help: "Hyperdiffusion coefficient for scalar (m s-1)"
-  value: 1.5
+  value: 0.929738
 # Topography
 topography:
   help: "Define the surface elevation profile [`NoWarp` (default), `Earth`, `DCMIP200`, `Hughes2023`, `Agnesi`, `Schar`, `Cosine2D`, `Cosine3D`]"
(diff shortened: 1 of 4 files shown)


akshaysridhar (approved):
This is reasonable; it clarifies the default behaviour and provides a useful message on exit.

Issue #3887, opened 2025-07-07
Abstract Out Reproducibility Tests


dennisYatunin:
@juliasloan25 has mentioned that ClimaCoupler currently lacks reproducibility test infrastructure, which would be useful for seeing how PRs in ClimaAtmos and other packages change our coupled simulations. There are some tests that compare simulations against reference solutions, but these errors are so large that they exhibit minimal changes even when the outputs of ClimaAtmos are significantly modified. Although it is helpful to have tests against reference solutions, reproducibility tests are more useful when the reference errors are large. In addition to ClimaCoupler, ClimaLand and other packages would also benefit from reproducibility tests.

In the process of abstracting out the reproducibility tests into a separate package, it would be good to make several updates:
- Generalize the infrastructure to work with other packages
  - Make a generic API that can be called from any CI pipeline with minimal setup
  - Add documentation and unit tests
- Only have CI jobs fail when errors are physically significant
  - Allow users to specify parameters for rescaling errors to determine if they are physically significant
  - Insignificant errors (e.g., roundoff errors from changing order of operations) should only be logged
- Consider moving all reproducibility tests into a single job per pipeline
  - This would make our CI results less noisy, with jobs that run simulations only failing when they crash
  - However, it would force users to spend time looking for artifacts of jobs that break reproducibility tests

PR #3893, opened 2025-07-15
Fix blank mse_summary

### LINKED ISSUE #3892: Reproducibility mse summary is missing
The "Print new mse tables" step of buildkite ci does not print any summaries.

imreddyTeja:
The mse_summary.jl script always prints no summaries or error. This is because the reproducibility files are now placed in a separate subfolder, which also contains a prog_state file. This commit updates the `get_computed_mses` default value for `subfolder` to reflect this, and ignores the prog_state file when searching for the most current comparison.

closes #3892

----

### DIFF
--- a/reproducibility_tests/reproducibility_utils.jl
+++ b/reproducibility_tests/reproducibility_utils.jl
@@ -722,7 +722,7 @@ where `is_mse_file` is `true`.
 """
 function get_computed_mses(;
     job_ids::Vector{String},
-    subfolder::String = "output_active",
+    subfolder::String = joinpath("output_active", "reproducibility_bundle"),
     is_mse_file::Function = default_is_mse_file,
     expected_filename_prefix = "computed_mse",
 )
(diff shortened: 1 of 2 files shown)


haakon-e:
I'm probably not the best to review this, so I'll just leave some thoughts:
- Based on your description, the changes seem reasonable.
- Based on your description of how the folder structure of reproducibility files has changed, would it make sense to add a test that verifies the current expectation of things, to ensure that this code is remembered if the folder structure ever changes again in the future?
- For example, what would happen if we changed the name of the `prog_state.hdf5` file? What would happen if we changed the name or contents of the `reproducibility_bundle` folder, etc.

imreddyTeja:
> I'm probably not the best to review this, so I'll just leave some thoughts:
> 
> * Based on your description, the changes seem reasonable.
> * Based on your description of how the folder structure of reproducibility files has changed, would it make sense to add a test that verifies the current expectation of things, to ensure that this code is remembered if the folder structure ever changes again in the future?
>   
>   * For example, what would happen if we changed the name of the `prog_state.hdf5` file?
>   * What would happen if we changed the name or contents of the `reproducibility_bundle` folder?
>   * etc...

Thanks for the reviews everyone. 

A test is definitely a good idea, and I will add one before merging. 
The reproducibility code as a whole contains quite a few hard-coded paths.
This should be solved with Issue #3887

PR #3904, opened 2025-07-28
DiagnosticEDMF + NonEquil + 1M

trontrytel:
This PR adds the `NonEquilibrium` moisture and `1Moment` microphysics to `DiagnosticEDMF`. The cloud and precipitation source terms are computed in the updraft and environment subdomains. Precipitation advection is only done for the grid mean values of cloud condensate and precipitation. We decided that, without the previous time-step memory, it doesn't make sense to compute advective tendencies in the  diagnostic edmf updraft subdomain.

I switched the defaults, so that now DyCOMS RF02, RICO and TRMM are running with NonEquil +1Moment for diagnostic EDMF in the CI. I think that many issues could be improved by calibrating some of the free parameters. For example, I can get deeper convection and snow/ice in TRMM by reducing the cloud relevant relaxation timescales. But that also leads to reducing the model time step. I'll do more testing and change those once I start testing in AMIP. I just want to move away from only testing in the single column. 

I will wait for this PR https://github.com/CliMA/ClimaAtmos.jl/pull/3913 to go in, to make sure I'm looking at fresh mse tables changes.

### DIFF
--- a/src/cache/diagnostic_edmf_precomputed_quantities.jl
+++ b/src/cache/diagnostic_edmf_precomputed_quantities.jl
@@ -77,6 +77,7 @@ NVTX.@annotate function set_diagnostic_edmfx_draft_quantities_level!(
     Φ_level,
 )
     FT = eltype(thermo_params)
+
     @. ts_level = TD.PhaseEquil_phq(
         thermo_params,
         p_level,
@@ -84,6 +85,34 @@ NVTX.@annotate function set_diagnostic_edmfx_draft_quantities_level!(
         q_tot_level,
         8,
         FT(0.0003),
+    )
+    @. ρ_level = TD.air_density(thermo_params, ts_level)
+    return nothing
+end
+NVTX.@annotate function set_diagnostic_edmfx_draft_quantities_level!(
+    thermo_params,
+    ts_level,
+    ρ_level,
+    mse_level,
+    q_tot_level,
+    q_liq_level,
+    q_ice_level,
+    q_rai_level,
+    q_sno_level,
+    p_level,
+    Φ_level,
+)
+    FT = eltype(thermo_params)
+
+    @. ts_level = TD.PhaseNonEquil_phq(
+        thermo_params,
+        p_level,
+        mse_level - Φ_level,
+        TD.PhasePartition(
+            q_tot_level,
+            q_liq_level + q_rai_level,
+            q_ice_level + q_sno_level,
+        ),
     )
     @. ρ_level = TD.air_density(thermo_params, ts_level)
     return nothing
@@ -129,7 +158,7 @@ NVTX.@annotate function set_diagnostic_edmf_precomputed_quantities_bottom_bc!(
     p,
     t,
 )
-    (; turbconv_model) = p.atmos
+    (; turbconv_model, moisture_model, microphysics_model) = p.atmos
     FT = eltype(Y)
     n = n_mass_flux_subdomains(turbconv_model)
     (; ᶜΦ) = p.core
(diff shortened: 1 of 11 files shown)


src/cache/diagnostic_edmf_precomputed_quantities.jl
```diff
@@ -225,15 +279,90 @@ NVTX.@annotate function set_diagnostic_edmf_precomputed_quantities_bottom_bc!(
...
-            thermo_params,
-            tsʲ_int_level,
-            ρʲ_int_level,
-            mseʲ_int_level,
-            q_totʲ_int_level,
-            p_int_level,
-            Φ_int_level,
-        )
+
+        if moisture_model isa NonEquilMoistModel && microphysics_model isa Microphysics1Moment
```
szy21: Do we check that we always use NonEquilMoist and 1M together somewhere in the code, and throw an error if not? It would be good to add it somewhere if we don't have it.

src/cache/precomputed_quantities.jl
```diff
@@ -170,6 +170,15 @@ function precomputed_quantities(Y, atmos)
         ᶜgradᵥ_θ_liq_ice = Fields.Field(C3{FT}, cspace),
     )
 
+    diagnostic_edmf_precipitation_quantities =
```
szy21:
```suggestion
    diagnostic_precipitation_sgs_quantities =
```
(to be consistent with the other names)

src/cache/temporary_quantities.jl
```diff
@@ -50,6 +50,9 @@ function temporary_quantities(Y, atmos)
         temp_data_level_3 = Fields.field_values(
             Fields.level(Fields.Field(FT, center_space), 1),
         ), # ρaʲu³ʲ_datah_tot
+        temp_data_level_3_tmp = Fields.field_values(
```
szy21:
```suggestion
        temp_data_level_4 = Fields.field_values(
```

src/diagnostics/edmfx_diagnostics.jl
```diff
@@ -693,6 +749,20 @@ function compute_rhoaen!(
         out .= TD.air_density.(thermo_params, cache.precomputed.ᶜts⁰)
     end
 end
+function compute_rhoaen!(
```
szy21: I don't think we need the diagnostics for environmental thermodynamic variables for diagnostic edmf as they are the same as grid-mean, and I don't think it is useful to save two variables that are exactly the same. Do you need them for anything?

src/cache/temporary_quantities.jl
szy21: Oh I see you are just using this for debugging

src/cache/diagnostic_edmf_precomputed_quantities.jl
```diff
@@ -837,6 +1041,163 @@ NVTX.@annotate function set_diagnostic_edmf_precomputed_quantities_do_integral!(
...
+                        q_iceʲ_prev_level,
+                    )
+                @info("q_ice entrainment ", extrema(ρaʲu³ʲ_dataq_tmp))
+                @. ρaʲu³ʲ_dataq_tmp = microphysics_sources(
+                        local_geometry_halflevel.J,
+                        local_geometry_prev_level.J,
+                        ρaʲ_prev_level,
+                        S_q_iceʲ_prev_level,
+                    )
+                @info("q_ice microphysics sources ", extrema(ρaʲu³ʲ_dataq_tmp))
```
szy21: Just remember to remove these before merging

szy21 (approved):
Thanks! The overall structure looks good to me - I didn't check all the sources and sinks in details but I'll trust you:) I only have some comments on the additional diagnostics.

src/diagnostics/edmfx_diagnostics.jl
trontrytel: I needed it for plotting in the CI. The script assumes that I have environmental precipitation in the output. But I'll update the plotting script instead of adding more to the output

PR #3905, opened 2025-07-29
Small refactor

trontrytel:
@szy21 - on the way to adding 1M in diagEDMF I did a small refactor of the big integral terms. I found it easier to define the helper functions for advection, entrainment, etc, because it was easier to see from which level the information was coming in. 

Lmk if this looks correct to you and if you would be ok to write it this way.

### DIFF
--- a/src/cache/diagnostic_edmf_precomputed_quantities.jl
+++ b/src/cache/diagnostic_edmf_precomputed_quantities.jl
@@ -5,6 +5,46 @@ import NVTX
 import Thermodynamics as TD
 import ClimaCore: Spaces, Fields, RecursiveApply
 
+###
+### Helper functions for the diagnosic edmf integral
+###
+# ᶠJ     - Jacobian at half level below (-1/2)
+# ᶜJₚ    - Jacobian at previous level below (-1)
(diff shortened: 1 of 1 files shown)


szy21:
I think it would be better to move the buoyancy term outside `mse_advection`, as it is not advection. If you do that, I wonder if you can unify the area / mse / tracer advection by passing ones as the tracer for the area advection. And I wonder if you can unify the entrainment and detrainment as well by passing the relevant updraft and environment quantities.

trontrytel:
> I think it would be better to move the buoyancy term outside `mse_advection`, as it is not advection. If you do that, I wonder if you can unify the area / mse / tracer advection by passing ones as the tracer for the area advection. And I wonder if you can unify the entrainment and detrainment as well by passing the relevant updraft and environment quantities.

I ended up unifying both the `advection` and `entrainment` functions for `area`, `mse` and `tracer`. Now it actually makes sense to define them and re-use them. Hopefully CI agrees

szy21:
Thanks! Let me try to understand why it is behavior changing.

szy21 (approved):
Thanks! I think this is helpful. Let's see if the test will pass.

PR #3912, opened 2025-08-01
Bugfix: microphysics sources when running with edmf

sajjadazimi:
## Purpose 
- When running with edmf cloud condensate tendencies should not be called.
- Sources in edmf subdomains are updated according to updating working fluid assumptions.

----

### DIFF
--- a/src/parameterized_tendencies/microphysics/cloud_condensate.jl
+++ b/src/parameterized_tendencies/microphysics/cloud_condensate.jl
@@ -2,7 +2,7 @@
 ##### DryModel, EquilMoistModel
 #####
 
-cloud_condensate_tendency!(Yₜ, Y, p, _, _) = nothing
+cloud_condensate_tendency!(Yₜ, Y, p, _, _, _) = nothing
 
 #####
 ##### NonEquilMoistModel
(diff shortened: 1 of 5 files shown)


src/parameterized_tendencies/microphysics/cloud_condensate.jl
```diff
@@ -170,3 +173,38 @@ function cloud_condensate_tendency!(
...
+#####
+##### PrognosticEDMF and DiagnosticEDMF
+#####
+
+function cloud_condensate_tendency!(
+    Yₜ,
+    Y,
+    p,
+    ::NonEquilMoistModel,
+    ::Union{NoPrecipitation, Microphysics0Moment},
```
trontrytel, line 186: We could just write it as 
`::Union{NoPrecipitation, Microphysics0Moment, Microphysics1Moment, Microphysics2Moment,}`

sajjadazimi: This won't work because the function `cloud_condensate_tendency!` is ambiguous. The various instances of the function are more specific for different input arguments. For example `cloud_condensate_tendency!(Yₜ, Y, p, ::NonEquilMoistModel, ::Microphysics1Moment, PrognosticEDMFX)` is ambiguous because the compiler does not know if it should use `cloud_condensate_tendency!(::Microphysics1Moment, _)` or `cloud_condensate_tendency!(::Union{..., Microphysics1Moment, ...}, ::PrognosticEDMF)`

PR #3913, opened 2025-08-01
Reducing the noise in ice and snow for warm cases

trontrytel:
While working with EDMF in single column dycoms, I noticed we were getting small values of ice and snow. This is unphysical, because we never reach cold temperatures in that single column simulation.

I tracked it back to two issues:
- For force == 0 limiters were sometimes giving us small tendency from roundoff errors of lim - sqrt(lim^2) not being exactly zero.
- We were still making ice for temperatures greater than freezing. Limiters don't make that go away. So I added an explicit statement that no ice is being formed when it's warm. You can think of it as lack of ice nucleating particles.

### DIFF
--- a/reproducibility_tests/ref_counter.jl
+++ b/reproducibility_tests/ref_counter.jl
@@ -1,4 +1,4 @@
-251
+252
 
 # **README**
 #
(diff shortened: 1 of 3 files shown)


sajjadazimi (approved):
Thank you!

PR #3914, opened 2025-08-01
remove examples project.toml

szy21:
## Purpose 

`examples/Project.toml` is out of date. I don't think it is still working so this PR removes it.

## Content

----


szy21:
Should we keep `examples/hybrid/driver.jl`? This one might still be working, but we will need to make sure to keep it updated.

Julians42:
Is there a recommended environment for external users? I've written documentation that points to using the examples `Project.toml`. Typically I use `.buildkite` but I think the only real difference between this and what `examples` should be is the PrecompileCI package.

Julians42:
@szy21 It was working a month ago for a single column pr I added but agree that it probably needs to be updated. Maybe we could just copy the .buildkite one and remove precomputed ci? Let me run a test or something ...

nefrathenrici (approved):
Thank you! It's good to remove the untested Project.toml. I think we could remove the examples driver as well, unless people still rely on it.

szy21:
> Thank you! It's good to remove the untested Project.toml. I think we could remove the examples driver as well, unless people still rely on it.

I like it as it shows we can run ClimaAtmos in three lines. However, I don't think we should keep something that is not tested. I'll remove it. If we want to add it back in the future, we should add tests as well.

szy21:
FYI @akshaysridhar I kept examples/topography_spectra.jl, but it is also not tested anywhere.

PR #3917, opened 2025-08-02
add common numerics configs

szy21:
## Purpose 

Add common configs of numerics. The extra numeric configurations added to different YAMLs are to ensure this PR does not change behavior. Configs under `cpu_scaling_pipeline` are not changed because it is not clear whether this pipeline is meaningful. EDMF column configs are not unified because each case has a different resolution. Also removes two unnecessary jobs in ci.

## To-do

- [x] Modify gpu configs
- [x] Add to news

## Content

----


docs/src/config_no_table.md
```diff
@@ -63,6 +63,36 @@ The `help` field is optional if you don't plan on making a permanent change to t
...
-# Default Configuration
+ClimaAtmos provides a set of common numerical configurations that can be used as building blocks for different types of simulations. These configurations are located in `config/common_configs/` and contain standardized settings for grid resolution, time stepping, and numerical schemes.
+
+## Available Common Configurations
+
+### Column Configurations
+- **`numerics_column_ze63.yml`**: Single column configuration with 63 vertical levels
+
+### Sphere Configurations
+- **`numerics_sphere_he6ze10.yml`**: Spherical configuration with 6 horizontal elements, 10 vertical levels, and 30km domain top
```
tapios: Maybe add the average horizontal resolution in km here and elsewhere so it's a bit clearer to the uninitiated.

szy21: Will do.

szy21:
cc @juliasloan25 for coupler yaml changes

PR #3919, opened 2025-08-03
fix conservation check for 1M

szy21:
## Purpose 

Adds accumulation of energy source and sink due to precipitation for 1M microphysics. Also, previously we separated surface flux tendency from vertical diffusion, but didn't change the condition in slab ocean tendency. This PR fixes it.

## Content

----


tapios (approved):
Looks good to me. Thanks for catching this!

PR #3920, opened 2025-08-04
Update REPL script documentation

akshaysridhar:
Simplified version of `.buildkite/ci_driver.jl` that shows one example of a debugging workflow using `Revise`.

### DIFF
--- a/docs/src/repl_scripts.md
+++ b/docs/src/repl_scripts.md
@@ -1,3 +1,72 @@
+# REPL script (debugging workflow example)
+
+
+```@example
+
+ca_dir = joinpath(@__DIR__, "..", "..")
+buildkite_env = joinpath(ca_dir, ".buildkite")
+using Pkg
+Pkg.activate(buildkite_env)
+
+# julia> using Revise # This is useful For REPL debugging. See also: Infiltrator.jl and Main.@infiltrate
+
+import ClimaAtmos as CA
+import SciMLBase: step!
+
+# If you wish to run your simulation on a `CUDA` device, use
+
+# import ClimaComms
+# ClimaComms.@import_required_backends
+# ENV["CLIMACOMMS_DEVICE"]="CUDA" 
+
+# (Note that the example below runs on a single CPU, which uses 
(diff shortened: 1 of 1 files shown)


.buildkite/Manifest-v1.11.toml
szy21, line 1: Could we revert the change in this file?

docs/src/repl_scripts.md
```diff
@@ -1,3 +1,33 @@
+# REPL script (debugging workflow example)
+
+```{julia}
+julia> ]
+pkg> activate .buildkite
+```
+
+```julia
+julia> using Revise # See also: Infiltrator.jl and Main.@infiltrate
```
szy21: I don't think Revise is a dependency of ClimaAtmos - how could this work as a test?

akshaysridhar: Switching to use `Literate.jl` should allow lines with # to be treated as comments and run all other lines as standard julia code, so we can still exercise the `AtmosConfig, AtmosSimulation, atmos_solve!(simulation)` and `step!(simulation.integrator)` methods - Revise is still an external reference as a doc example.

szy21: So is this being tested now? I'm fine with merging it either way, but I don't think it is tested.

trontrytel: I think we would see an error in the doc build? Would that be good enough warning?

akshaysridhar: Instructions within the `example` block are being run and should show corresponding output (there is an issue here in that I can get this to work locally but it fails at some albedo plot generation unrelated to this PR in the github actions job, prior to even reaching the REPL docs - I'll look into this shortly)

szy21: Oh I think we are testing it in the latest commit now - and now it is failing 😂

akshaysridhar:
> What's the status of this PR?

Rebased, if it passes checks it's ready to merge (the doc-build exercises the REPL commands)

PR #3922, opened 2025-08-05
add 2M sources to PrgnosticEDMF

sajjadazimi:
## Purpose 
Add 2M sources to PrgnosticEDMF

----


src/initial_conditions/atmos_state.jl
```diff
@@ -173,22 +173,41 @@ function turbconv_center_variables(
         TD.specific_enthalpy(ls.thermo_params, ls.thermo_state) +
         CAP.grav(ls.params) * ls.geometry.coordinates.z
     q_tot = TD.total_specific_humidity(ls.thermo_params, ls.thermo_state)
-    q_liq = TD.liquid_specific_humidity(ls.thermo_params, ls.thermo_state)
-    q_ice = TD.ice_specific_humidity(ls.thermo_params, ls.thermo_state)
+    q_liq = TD.liquid_specific_humidity(ls.thermo_params, ls.thermo_state) # - q_rai ?
```
trontrytel, line 176: I agree that we should subtract here. I don't think we have any initial conditions with non-zero precipitation. But better safe than sorry

trontrytel (approved):
This looks great! Thank you

PR #3927, opened 2025-08-06
Add diffusion to microphysics tracers in EDOnly and diagEDMF

trontrytel:
Also deletes the `ᶜspecific_env_value` for EDOnly and diagEDMF models. We should just be using the grid mean values there.


szy21 (approved):
Thanks!

PR #3928, opened 2025-08-06
add rayleigh sponge for edmf variables

szy21:
## Purpose 

Add rayleigh sponge to subgrid-scale variables, which relax updraft tracers to grid-mean and tke to 0.

## To-do

- [x] Update to the latest version of ClimaParams

## Content

----


src/parameterized_tendencies/radiation/radiation.jl
szy21, line 1: Ignore changes in this file. I will revert it.

config/model_configs/prognostic_edmfx_aquaplanet.yml
```diff
@@ -2,8 +2,6 @@
 z_max: 30000.0
 z_elem: 10
 dz_bottom: 500.0
-rayleigh_sponge: false
```
costachris, line 5: are these taken out because they will be specified in the Coupler? or because the sponges aren't needed for the aquaplanet

costachris (approved):
LGTM

PR #3930, opened 2025-08-07
add viscous sponge to updraft u3

szy21:
## Purpose 

We have viscous sponge on grid-mean u3 but not updraft u3. This PR fixes it so it is more consistent.

## Content

----


costachris (approved):
LGTM

PR #3932, opened 2025-08-07
Force precompilation only for `CPUSingleThreaded`

nefrathenrici:
## Content

----


szy21 (approved):
Thanks! Is this to help with the issue I got when running simulations myself?

PR #3933, opened 2025-08-08
Bump to v0.31.1

imreddyTeja:
Make a patch release

----


calibration/experiments/gcm_driven_scm/Manifest.toml
```diff
@@ -1,17 +1,18 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.10.5"
+julia_version = "1.11.5"
```
juliasloan25: I didn't even know these manifests existed - I assume they didn't need to use Julia 1.10 but they're just old, so this is probably fine

juliasloan25 (approved):

szy21:
Thanks! Do you plan to update the latest versions of the packages too?

imreddyTeja:
> Thanks! Do you plan to update the latest versions of the packages too?

I'm not sure I understand. The latest versions of which packages?

imreddyTeja:
> > > Thanks! Do you plan to update the latest versions of the packages too?
> > 
> > 
> > I'm not sure I understand. The latest versions of which packages?
> 
> New minor versions of climaparams and thermodynamics.

I can't update them because their version is restricted by SurfaceFluxes and Insolation. I could add them to the compat if you'd like, but they would be untested.

juliasloan25:
This is my bad - @szy21 when I said we'd update the versions of ClimaParams and TD I didn't realize they need to get updated in Insolation and SF first. Would you rather update those as well (and move this atmos release to tomorrow), or make this release now and update those packages + release next week?

imreddyTeja:
I think Tapio may have also wanted Atmos to use RootSolvers v1.0.0, but that also needs a ClimaCore release.

szy21:
> This is my bad - @szy21 when I said we'd update the versions of ClimaParams and TD I didn't realize they need to get updated in Insolation and SF first. Would you rather update those as well (and move this atmos release to tomorrow), or make this release now and update those packages + release next week?
> 
> I do think it would be helpful to release atmos today since it's been causing downstream coupler tests to fail in ClimaLand

Sure, feel free to make a release now. It would be good to update those packages as soon as possible though. Some of us may be able to help update different repos.

szy21 (approved):

Project.toml
```diff
@@ -44,16 +44,16 @@ ArgParse = "1"
 Artifacts = "1"
 AtmosphericProfilesLibrary = "0.1.7"
 ClimaComms = "0.6.9"
-ClimaCore = "0.14.37"
+ClimaCore = ">= 0.14.37"
 ClimaDiagnostics = "0.2.12"
 ClimaInterpolations = "0.1.0"
-ClimaParams = "0.10.35"
+ClimaParams = "0.10.35, 0.11"
```
imreddyTeja, line 50: @szy21 should support for the older versions be dropped, or is what I have here OK?

juliasloan25: I believe the values of some parameters changed but names didn't change so the interface hasn't changed. So it should be fine to continue allowing 0.10 and 0.11. But Zhaoyi will know for sure

szy21: Yes, what you have here is ok.

PR #3937, opened 2025-08-08
Update longrun_aquaplanet_progedmf.toml

szy21:
## Purpose 

EDMF_surface_area was set to 0 before as I was testing EDMF without updraft, but I forgot to remove it. This resulted in no updraft in the longrun. This PR fixes it. cc @costachris 

Also fixes some typos in parameters.

## Content

----


szy21:
manually merging it as it only changes longrun yaml

PR #3938, opened 2025-08-09
Only load PrecompileCI in CI builds

nefrathenrici:
## Content

----


nefrathenrici:
> Thanks! Should we change #3932 back?

No, I think both changes are useful for preventing precompilation issues.

PR #3943, opened 2025-08-11
Disable Aqua persistent tasks test

imreddyTeja:
v3.37 of RecursiveArrayTools causes a circular
dependency with julia 1.10. This temporarily
disables the aqua test that errors because of this.

See the [issue](https://github.com/SciML/RecursiveArrayTools.jl/issues/477) in RecursiveArrayTools.

## Content

----


szy21 (approved):
Thank you!

PR #3944, opened 2025-08-11
tmp testing to fix coupler test

trontrytel:
This is what I was talking about to tmp save the mixing length. Does that make sense @dennisYatunin ? Also, how do I check it for the Coupler @szy21 ?

### DIFF
--- a/src/cache/cloud_fraction.jl
+++ b/src/cache/cloud_fraction.jl
@@ -94,7 +94,9 @@ NVTX.@annotate function set_cloud_fraction!(
         end
     end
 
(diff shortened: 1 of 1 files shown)


szy21:
To test it in coupler I think you need to change the Manifest in the coupler here: https://github.com/CliMA/ClimaCoupler.jl/blob/main/experiments/ClimaEarth/Manifest-v1.11.toml and run ci. (In the same folder there is also Manifest.toml for julia 1.10? But hopefully it would be good enough to just test one of them.)

szy21:
This fixes it in the coupler: https://buildkite.com/clima/climacoupler-ci/builds/6866#_

PR #3946, opened 2025-08-12
Add Rainwater Path Diagnostic

ericmpham:
## Purpose 

Implementing rainwater path as a diagnostic variable.

## Content

- Implemented RWP in core_diagnostics.jl.
- Add as a diagnostic for "config/model_configs/prognostic_edmfx_dycoms_rf02_column.yml".

----

### DIFF
--- a/src/diagnostics/core_diagnostics.jl
+++ b/src/diagnostics/core_diagnostics.jl
@@ -1026,7 +1026,7 @@ add_diagnostic_variable!(
     short_name = "clwvi",
     long_name = "Condensed Water Path",
     standard_name = "atmosphere_mass_content_of_cloud_condensed_water",
-    units = "kg m-2",
+    units = "kg m^-2",
     comments = """
     Mass of condensed (liquid + ice) water in the column divided by the area of the column
     (not just the area of the cloudy portion of the column). It doesn't include precipitating hydrometeors.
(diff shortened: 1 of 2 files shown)


src/diagnostics/core_diagnostics.jl
```diff
@@ -1503,3 +1503,43 @@ add_diagnostic_variable!(
...
+        @. rw = state.c.ρq_rai
+        Operators.column_integral_definite!(out, rw)
+    end
+end
+
+add_diagnostic_variable!(
+    short_name = "rwp",
+    long_name = "Rainwater Path",
+    standard_name = "atmosphere_mass_content_of_rain_liquid_water",
+    units = "kg m-2",
```
akshaysridhar:
```suggestion
    units = "kg m^-2",
```

akshaysridhar:
> Closed and reopened because the branch was renamed to better fit the naming convention. Suggestions from previous PR have been added/committed to this branch.

@erphamucsd On the `TODO` block above : the pattern moving forward will rely on reducing dependency on cache variables - so, instead of precomputing `q_rai` we will use the `specific` function `https://github.com/CliMA/ClimaAtmos.jl/blob/97934873be37a2d92cb83627ff310d0cb995ae6c/src/utils/variable_manipulations.jl#L5` if necessary to get quantities from prognostic state variables. `LazyBroadcast` may also help with faster gpu code; you're using scratch space here so this is not strictly necessary.

PR #3947, opened 2025-08-12
Redefine sedimentation velocity in Prognostic EDMF (1M/2M)

trontrytel:
This PR fixes the computation of terminal velocities. They are now calculated consistently from the subdomain terminal velocities for both 1-moment and 2-moment microphysics.  

It also fixes a bug: after applying boundary conditions for the area fraction, the constraint  

$$
\langle \hat{\rho} u^j \rangle = \rho u
$$  

was not satisfied at the boundary. This introduced errors in the SGS mass flux computations. I now recompute $u^{3,0}$ so that the constraint holds.  

By the new definition of sedimentation velocities, the sum of fluxes of any tracer or energy over subdomains equals the flux at grid scale. As a result, there is no need to add sedimentation effects into the computation of SGS mass fluxes. The general form of the EDMF tracer flux is  

$$
F_\chi = \langle \hat{\rho} \\, a \\, w^j \chi^j \rangle - \rho w \chi ,
$$

where  
- $w = w_\text{air} + w_\chi$ is the vertical velocity, including both air motion and the sedimentation velocity of tracer $\chi$,  
- $\langle \cdot \rangle$ denotes the sum over subdomains, and  
- the second term is the resolved grid-scale flux.  

---

### Cancellation for tracers  

If we define the terminal velocities such that  

$$
\langle \hat{\rho} \\, a \\, w_\chi^j \chi^j \rangle = \rho w_\chi \chi ,
$$

then sedimentation contributions cancel in the SGS flux:  

$$
\begin{aligned}
F_\chi &= \langle \hat{\rho} \\, a \\, (w_\text{air}^j + w_\chi^j)\chi^j \rangle 
         - \rho (w_\text{air} + w_\chi)\chi \\
       &= \langle \hat{\rho} \\, a \\, w_\text{air}^j \chi^j \rangle - \rho w_\text{air} \chi .
\end{aligned}
$$

This cancellation holds for all tracers ($q_\text{liq}, q_\text{ice}, q_\text{rai}, q_\text{sno}$), their sum (total water sedimentation), and also for energy.  

---

### Energy case  

For total energy, the flux is  

$$
F_h = \Big\langle \hat{\rho} \big( w_\text{air}^j h_\text{tot}^j + 
 \sum_p w_p^j q_p^j h_p^j \big)\Big\rangle - \rho \Big( w_\text{air} h_\text{tot} + \sum_p w_p q_p h_p \Big) ,
$$

where $p$ runs over the precipitating species (liq, ice, rain, snow).  

We define the sedimentation velocity of energy as  

$$
w^j_h = \frac{\sum_p w^j_p q^j_p h^j_p}{h^j_\text{tot}} .
$$

The grid-scale sedimentation velocity of energy is defined analogously to tracers, such that the sum of subdomain energy fluxes equals the grid-scale energy flux. Therefore, as with tracers, energy sedimentation does not contribute to SGS fluxes.

### DIFF
--- a/src/cache/precipitation_precomputed_quantities.jl
+++ b/src/cache/precipitation_precomputed_quantities.jl
@@ -31,12 +31,12 @@ function Kin(ᶜw_precip, ᶜu_air)
 end
 
 """
-    set_precipitation_velocities!(Y, p, moisture_model, microphysics_model)
+    set_precipitation_velocities!(Y, p, moisture_model, microphysics_model, turbconv_model)
 
 Updates the precipitation terminal velocity, cloud sedimentation velocity,
 and their contribution to total water and energy advection.
 """
-function set_precipitation_velocities!(Y, p, _, _)
+function set_precipitation_velocities!(Y, p, _, _, _)
     (; ᶜwₜqₜ, ᶜwₕhₜ) = p.precomputed
     @. ᶜwₜqₜ = Geometry.WVector(0)
     @. ᶜwₕhₜ = Geometry.WVector(0)
@@ -47,6 +47,7 @@ function set_precipitation_velocities!(
     p,
     moisture_model::NonEquilMoistModel,
     microphysics_model::Microphysics1Moment,
+    _,
 )
     (; ᶜwₗ, ᶜwᵢ, ᶜwᵣ, ᶜwₛ, ᶜwₜqₜ, ᶜwₕhₜ, ᶜts, ᶜu) = p.precomputed
     (; ᶜΦ) = p.core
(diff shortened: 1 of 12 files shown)


.buildkite/Project.toml
```diff
@@ -25,6 +25,7 @@ Krylov = "ba0b0d4f-ebba-5204-a429-3ac8c609bfb7"
 MPI = "da04e1cc-30fd-572f-bb4f-1f8673147195"
 NCDatasets = "85f8d34a-cbdd-5861-8df4-14fed0d494ab"
 NullBroadcasts = "0d71be07-595a-4f89-9529-4065a4ab43a6"
+OhMyREPL = "5fb14364-9ced-5910-84b2-373655c76a03"
```
trontrytel: delete

sajjadazimi:
In addition to adding sedimentation effects in the SGS mass fluxes, this PR also fixes the computation of terminal velocities. They are now computed consistently from the subdomain terminal velocities, for both 1-moment and 2-moment microphysics.  

I also fixed a bug: after applying boundary conditions for area fraction, the constraint  
$$
\langle \hat{\rho} u \rangle = \rho u
$$  
was not satisfied at the boundary. This introduced errors in the SGS mass flux computations. I now recompute $u^{3,0}$ so that the constraint holds.

src/cache/precipitation_precomputed_quantities.jl
```diff
@@ -98,13 +99,114 @@ function set_precipitation_velocities!(
...
+        ᶜρ⁰,
+        max(zero(Y.c.ρ), ᶜq_liq⁰),
+    )
+    @. ᶜwₗ = ᶜρa⁰ * ᶜq_liq⁰ * ᶜw⁰
+    @. ᶜρχ = max(zero(Y.c.ρ), ᶜρa⁰ * ᶜq_liq⁰)
+    for j in 1:n
+        @. ᶜwₗ += Y.c.sgsʲs.:($$j).ρa * Y.c.sgsʲs.:($$j).q_liq * ᶜwₗʲs.:($$j)
+        @. ᶜρχ += max(zero(Y.c.ρ), Y.c.sgsʲs.:($$j).ρa * Y.c.sgsʲs.:($$j).q_liq)
+    end
+    @. ᶜwₗ = ifelse(ᶜρχ > eps(FT), ᶜwₗ / ᶜρχ, FT(0))
```
szy21: Would it be easier if you just write ᶜρχ as Y.c.ρq_liq? (It should be the same as what you have now?)

szy21: Resolved offline: we can't do it as subgrid scale tracer can be negative.

szy21:
I want to update dependencies (including CloudMicrophysics) before Monday. Will it affect this PR? I can wait if it will cause some issues.

src/cache/precipitation_precomputed_quantities.jl
szy21: I reopened this conversation as I'm not sure which one is more correct. It makes sense that the sedimentation velocity of grid-mean tracer should not be larger than both updraft and environment. However, this sedimentation velocity is applied to grid-mean tracer Y.c.ρχ, not ᶜρχ as defined here. Wouldn't it be more correct to define the sedimentation velocity to be total flux / Y.c.ρχ then?

sajjadazimi: Thanks for reopening — let me clarify my reasoning. The issue comes from negative mass in the environment: if we enforce flux equality, the grid-mean sedimentation velocity ends up larger than both the draft and environment, which is not physical. Instead, I assume equal terminal velocity across subdomains, which effectively treats the negative-mass environment as moving with the same velocity as the draft and grid mean. This avoids unphysical acceleration of the grid mean, and also improves numerical stability. Once the negative-mass issue is resolved, this distinction will no longer be relevant.

szy21: Makes sense, thanks!

PR #3949, opened 2025-08-12
Allow for Modifications to DYCOMS-RF02 Initial Conditions via Config

erphamucsd:
## Purpose 

Additional configuration variables were added to pass arguments to redefined APL functions that allow for changes in the initial conditions profiles.

## To-do

- Maybe we want to move some of the other documentation pages into examples. It may be out of the scope of this PR.
- Maybe we might rewrite the APL functions to allow for additional arguments, but that involves modifying that library. It may also be out of the scope of this PR.

## Content

- Update initial_conditions.jl, type_getters.jl, and default diagnostics.jl to allow for values to be passed to specify total water mixing ratio (qtot), liquid ice potential temperature (theta), and initial boundary layer height (zi).
- Add documentation, specifically an examples page. 

----


docs/src/contributor_guide.md
```diff
@@ -165,7 +165,16 @@ or clarification that you can take care of while you are here.
 
 Here is an example of a docstring:
 
-TODO: add example
+```
```
akshaysridhar, line 168: This example is helpful. Thanks for knocking out these stranded TODOs!

PR #3951, opened 2025-08-14
Allow for Modifications to DYCOMS-RF02 Initial Conditions via Config

ericmpham:
## Purpose 

Additional configuration variables were added to pass arguments to redefined APL functions that allow for changes in the initial conditions profiles.

## To-do

- Maybe we want to move some of the other documentation pages into examples. It may be out of the scope of this PR.
- Maybe we might rewrite the APL functions to allow for additional arguments, but that involves modifying that library. It may also be out of the scope of this PR.

## Content

- Update initial_conditions.jl, type_getters.jl, and default diagnostics.jl to allow for values to be passed to specify total water mixing ratio (qtot), liquid ice potential temperature (theta), and initial boundary layer height (zi).
- Add documentation, specifically an examples page.

----

### DIFF
--- a/config/model_configs/diagnostic_edmfx_dycoms_rf02_column.yml
+++ b/config/model_configs/diagnostic_edmfx_dycoms_rf02_column.yml
@@ -1,0 +1,41 @@
+initial_condition: DYCOMS_RF02
+subsidence: DYCOMS
+scm_coriolis: DYCOMS_RF02
+rad: DYCOMS
+surface_setup: DYCOMS_RF02
+turbconv: "diagnostic_edmfx"
+implicit_diffusion: true
+approximate_linear_solve_iters: 2
+edmfx_upwinding: first_order
+edmfx_entr_model: "Generalized"
+edmfx_detr_model: "Generalized"
+edmfx_sgs_mass_flux: true
+edmfx_sgs_diffusive_flux: true
+edmfx_nh_pressure: true
+edmfx_filter: true
+prognostic_tke: true
+moist: "nonequil"
+cloud_model: "grid_scale"
+precip_model: "1M"
+call_cloud_diagnostics_per_stage: true
+config: "column"
+x_elem: 2
+y_elem: 2
+z_elem: 30
+z_max: 1500
+z_stretch: false
+perturb_initstate: false
+dt: 10secs
+t_end: 5hours
+dt_save_state_to_disk: 10mins
+toml: [toml/diagnostic_edmfx_1M.toml]
+netcdf_interpolation_num_points: [8, 8, 30]
+diagnostics:
+  - short_name: [ts, ta, thetaa, ha, pfull, rhoa, ua, va, wa, hur, hus, cl, clw, cli, hussfc, evspsbl, pr, lwp, rwp]
+    period: 10mins
+  - short_name: [arup, waup, taup, thetaaup, haup, husup, hurup, clwup, cliup, husraup, hussnup, tke]
+    period: 10mins
+  - short_name: [entr, detr, lmix, bgrad, strain, edt, evu]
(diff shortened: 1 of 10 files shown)


erphamucsd:
"Is there a way to collapse the if-else statements for selecting initial conditions down into something cleaner?" was something that was sort of being discussed.

erphamucsd:
@akshaysridhar, @oalcabes, I think this is ready for review. Off the top of my head, I can't figure out a way to make the type_getters.jl ifelse statements less verbose. 

I was also wondering if maybe it would be better to directly update the APL library functions instead of importing them, then overwriting them, but that might be more trouble than it's worth. It kinda defeats the purpose of allowing other people to make similar modifications because then that means that updating the APL functions becomes part of the steps, and that means it's no longer contained within just ClimaAtmos.

Thanks guys!

config/model_configs/diagnostic_edmfx_dycoms_rf02_column.yml
```diff
@@ -0,0 +1,41 @@
...
+edmfx_upwinding: first_order
+edmfx_entr_model: "Generalized"
+edmfx_detr_model: "Generalized"
+edmfx_sgs_mass_flux: true
+edmfx_sgs_diffusive_flux: true
+edmfx_nh_pressure: true
+edmfx_filter: true
+prognostic_tke: true
+moist: "nonequil"
+cloud_model: "quadrature_sgs"
```
oalcabes: oops, this should be getting ran with grid_scale instead

docs/src/perturbation_experiments.md
```diff
@@ -0,0 +1,112 @@
+
+# Creating custom arguments for perturbation experiments
+In cases where it is of interest to run ensembles of the same model configuration but with perturbed initial conditions, it may be of use to add new keyword arguments to the .yml file. The idea is to allow modifications to initial conditions to be passed directly through the .yml file and then creating multiple copies of the files but with different variations and combinations of these parameters. 
+
+As an example, we will explore modifying the total water mixing ratio (`q_tot_0`), liquid-ice potential temperature profiles (`theta_0` & `theta_i`), and initial boundary layer height (`z_i`) in the DYCOMS-RF02 simulation setup.
```
oalcabes, line 5: It might be useful to talk to Nat, Costa & Julian about how they've implemented this for calibration -- I don't know how similar what you've done is to that?

config/model_configs/diagnostic_edmfx_dycoms_rf02_column.yml
ericmpham: is this the case for both prognostic and diagnostic or only diagnostic? This is referring to cloud_model right?

docs/src/perturbation_experiments.md
oalcabes: @nefrathenrici @costachris what are your opinions of doing things this way?

nefrathenrici: I think this is a great start (especially the initial conditions) but modifying the source code is probably time consuming and error-prone. I think the perturbation variables should be specified via parameters rather than parsed args, with an extra dycoms initial condition to read them in automatically.

nefrathenrici: If you want to merge this quickly, I would request a test case that can run as part of buildkite to ensure that this feature stays supported.

ericmpham: @nefrathenrici is what I put in the pipeline.yml okay? I consulted with Sajjad cause I was not really sure what to put for a test case. 

I agree that it's also a tedious process, I wish I had more time to restructure it.

szy21:
What's the status of this PR?

PR #3952, opened 2025-08-15
remove get_wstar function

szy21:
## Purpose 

`get_wstar` is only used once in the code for a few edmf single column cases. This PR removes it and calculates w_star in surface_conditions.jl directly.

## Content

----


tapios (approved):
LGTM

PR #3960, opened 2025-08-20
Synchronised parameter names with ClimaParams

ray-chew:
## Purpose 

Specifically, w.r.t. ClimaParams commit 74c2543

## Content

----


juliasloan25:
@szy21 do we still need this after #3972?

szy21:
> @szy21 do we still need this after #3972?

No we don’t. I left it open as it contains some other things Ray is testing I think.

PR #3962, opened 2025-08-20
Add 2M microphysics to diagnostic EDMF dycoms RF02

trontrytel:
Hi Folks! This PR adds 2-moment microphysics to diagnostic EDMF.

@sajjadazimi - could I ask you to take a look? I wasn't sure about two things:
1. What velocity I should be using in the aerosol activation. 
2. How to interact with global aerosol data. 

For the second point I temporarily hijacked two parameters: `seasalt_calibration_coefficient`, `reference_seasalt_aerosol_mass_concentration` to allow the users to prescribe the assumed mode size and concentration for seasalt aerosol. Once we make the breaking release of CliMA Params I will add proper parameters for it. 

We do get some clouds and drizzle and the results differ for high and low concentrations. But in the couple of configurations I tried we have been raining out a lot of the water very early on.

### DIFF
--- a/src/cache/diagnostic_edmf_precomputed_quantities.jl
+++ b/src/cache/diagnostic_edmf_precomputed_quantities.jl
@@ -204,7 +204,8 @@ NVTX.@annotate function set_diagnostic_edmf_precomputed_quantities_bottom_bc!(
     obukhov_length_sfc_halflevel = Fields.field_values(obukhov_length)
 
     if moisture_model isa NonEquilMoistModel &&
-       microphysics_model isa Microphysics1Moment
+       microphysics_model isa Microphysics1Moment ||
+       microphysics_model isa Microphysics2Moment
         (; ᶜq_liqʲs, ᶜq_iceʲs, ᶜq_raiʲs, ᶜq_snoʲs) = p.precomputed
 
         ᶜq_liq = @. lazy(specific(Y.c.ρq_liq, Y.c.ρ))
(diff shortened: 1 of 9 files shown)


src/cache/diagnostic_edmf_precomputed_quantities.jl
```diff
@@ -834,7 +926,85 @@ NVTX.@annotate function set_diagnostic_edmf_precomputed_quantities_do_integral!(
...
+                    q_liqʲ_prev_level,
+                    q_iceʲ_prev_level,
+                    q_raiʲ_prev_level,
+                    q_snoʲ_prev_level,
+                    ρʲ_prev_level,
+                    TD.air_temperature(thermo_params, tsʲ_prev_level),
+                    dt,
+                )
+                # Aerosol activation
+                # TODO - right now I'm hijacking the
```
sajjadazimi, line 983: We can read prescribed aerosol data from artifacts (Merra-2 dataset). For example the function ``microphysics_wrappers/compute_prescribed_aerosol_properties!`` reads sea salt and sulfate data and computes their number concentrations and the mean radius of seasalt. In order to use the data we need to pass a list of aerosols in the yml file.

src/cache/diagnostic_edmf_precomputed_quantities.jl
```diff
@@ -834,7 +926,85 @@ NVTX.@annotate function set_diagnostic_edmf_precomputed_quantities_do_integral!(
...
+                    sulfate_num_prev_level,
+                    q_totʲ_prev_level,
+                    q_liqʲ_prev_level + q_raiʲ_prev_level,
+                    q_iceʲ_prev_level + q_snoʲ_prev_level,
+                    n_liqʲ_prev_level + n_raiʲ_prev_level,
+                    ρʲ_prev_level,
+                    max(
+                        0,
+                        w_component.(
+                            Geometry.WVector.(u³ʲ_data_prev_halflevel)
```
sajjadazimi, line 1000: I think this is correct! We need to pass the velocity component along gravity and I think this is it.

sajjadazimi (approved):
Thank you! LGTM!
The velocity passed to aerosol activation seems correct. And for reading aerosol data we need to add the list of aerosols to yml and then use the function ``compute_prescribed_aerosol_properties!`` from microphysics wrappers .

PR #3963, opened 2025-08-21
rft: conditionally load microphysics parameters

haakon-e:
## Purpose 

This pull request updates the parameter creation logic in `src/parameters/create_parameters.jl` to optimize how microphysics parameters are loaded based on configuration arguments. The main improvement is that only the relevant microphysics parameter set is instantiated, reducing memory usage and improving efficiency.

Parameter loading optimization:

* The `ClimaAtmosParameters` function now accepts an optional `parsed_args` argument, allowing it to determine which microphysics model is in use and instantiate only the corresponding parameter set. (`src/parameters/create_parameters.jl`, [src/parameters/create_parameters.jlL18-R26](diffhunk://#diff-a9dec5e0780f17cbff9d5838bf3ef2feefae3ac2951e658359ce3581b8f00c54L18-R26))
* If `parsed_args` is provided, the function checks which microphysics model is selected and sets unused parameter sets (`microphysics_0m_params`, `microphysics_1m_params`, `microphysics_2m_params`) to `nothing`, ensuring only necessary parameters are loaded. (`src/parameters/create_parameters.jl`, [src/parameters/create_parameters.jlR57-R65](diffhunk://#diff-a9dec5e0780f17cbff9d5838bf3ef2feefae3ac2951e658359ce3581b8f00c54R57-R65))

This pull request is an attempt to resolve GPU kernel parameter memory issues, ref slack discussion [here](https://climate-machine.slack.com/archives/C0102PWRSGJ/p1755712606324659)

## To-do

* tests

----

### DIFF
--- a/src/parameters/create_parameters.jl
+++ b/src/parameters/create_parameters.jl
@@ -15,12 +15,15 @@ import StaticArrays as SA
 Construct the parameter set for any ClimaAtmos configuration.
 """
 ClimaAtmosParameters(config::AtmosConfig) =
-    ClimaAtmosParameters(config.toml_dict)
+    ClimaAtmosParameters(config.toml_dict, config.parsed_args)
 
 ClimaAtmosParameters(::Type{FT}) where {FT <: AbstractFloat} =
     ClimaAtmosParameters(CP.create_toml_dict(FT))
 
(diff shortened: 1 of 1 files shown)


nefrathenrici (commented):
I think this is fine, since the parameters can still be constructed without the AtmosConfig. Thank you!

PR #3968, opened 2025-08-22
rft: Increase formatting margin and enable source-based line joins for semantic grouping

haakon-e:
# Summary 
Adjust `.JuliaFormatter.toml` to (a) raise `margin` to 120 and (b) set `join_lines_based_on_source = true`, allowing selective horizontal grouping where it improves semantic readability while preserving existing vertical structure where it conveys meaning.

## Rationale
**Previous strict wrapping forced one-item-per-line patterns that:**
* Inflated vertical span of dense physics kernels.
* Obscured natural semantic clusters (e.g., related stress terms, gas families).
* Reduced scan speed during reviews (mental grouping had to be reconstructed). The new settings let authors explicitly signal grouping intent in source; the formatter no longer immediately re-expands those choices.

## Changed Options
* `margin`: `80` → `120` (headroom for grouped argument clusters without overflow)
`join_lines_based_on_source`: `false` → `true` (respect intentional horizontal grouping)

## Intended Benefits
* Faster cognitive parsing of tightly coupled argument sets.
* Reduced diff noise (smaller vertical churn for minor edits).
* Space to annotate clusters with concise inline comments without forcing wrap.
* Still respects margin to prevent unreadably long lines.

The option, `join_lines_based_on_source` is the key change in this PR (whether you set margin to 120 or 100 or 500 isn't really too important. A modest increase as proposed, simply allows more flexibility). 

With `join_lines_based_on_source = false`, if you have a function signature (or other expression) that is less than `margin` number of characters, it is _**forced**_ to be on one line. If the expression exceeds `margin`, then it is _**forced**_ to be split item-by-item on a new line.
With `join_lines_based_on_source = true`, the developer can choose themselves what makes the most sense for the code they're writing. For wrapper functions, putting things on one line is reasonable. For large function signatures for a physics closure, visually grouping related variables may make the most sense. The choice should be left to the expert, not an overly limiting formatter tool.

## Non-Goals
* Enforcing horizontal compaction everywhere.
* Reformatting legacy files wholesale.
* Introducing any behavioral code changes.

## Guidelines / Guardrails 
Use horizontal grouping when:
* Parameters form a natural physical or logical block (e.g., τ components, gas VMR family).
* Adjacent arguments share a common prefix or dimensional meaning. Retain one-per-line or minimal groups when:
* Arguments are heterogeneous and unrelated.
* Adding grouping would push near or beyond margin or harm diff clarity. Add a short comment header above multi-group blocks when it materially aids scanning.

**Note:** the reformatted files within this PR came about solely from running `JuliaFormatter.format(".")` after implementing the configuration changes. These came about due to the increased `margin` option, but are, as you can see, minimal.

### DIFF
--- a/.JuliaFormatter.toml
+++ b/.JuliaFormatter.toml
@@ -1,5 +1,6 @@
-margin = 80
+margin = 92
 indent = 4
 always_for_in = true
 whitespace_typedefs = true
 whitespace_ops_in_indices = true
+join_lines_based_on_source = true  # https://domluna.github.io/JuliaFormatter.jl/dev/#join_lines_based_on_source
--- a/src/parameterized_tendencies/gravity_wave_drag/orographic_gravity_wave.jl
+++ b/src/parameterized_tendencies/gravity_wave_drag/orographic_gravity_wave.jl
@@ -413,7 +413,7 @@ function calc_saturation_profile!(
         -(ᶠd2udz * τ_x + ᶠd2vdz * τ_y) / max(eps(FT), sqrt(τ_x^2 + τ_y^2)),
     )
     L1 = @. topo_L0 *
-       max(FT(0.5), min(FT(2.0), FT(1.0) - FT(2.0) * ᶠVτ * ᶠd2Vτdz / ᶠN^2))
+            max(FT(0.5), min(FT(2.0), FT(1.0) - FT(2.0) * ᶠVτ * ᶠd2Vτdz / ᶠN^2))
     # the coefficient FT(2.0) is the correction for coarse sampling of d2v/dz2
     FrU_clp0 = FrU_clp
     FrU_sat0 = FrU_sat
--- a/src/parameterized_tendencies/gravity_wave_drag/orographic_gravity_wave_helper.jl
+++ b/src/parameterized_tendencies/gravity_wave_drag/orographic_gravity_wave_helper.jl
@@ -242,7 +242,8 @@ function compute_OGW_info(Y, elev_data, earth_radius, γ, h_frac)
     hpoz = calc_hpoz_latlon(elev, lon, lat, earth_radius)
     hpoz = @. max(FT(0), hpoz)^(FT(2) - γ)
     hmax = @. (
-        abs(hpoz) * (γ + FT(2)) / (FT(2) * γ) * (FT(1) - h_frac^(FT(2) * γ)) / (FT(1) - h_frac^(γ + FT(2)))
+        abs(hpoz) * (γ + FT(2)) / (FT(2) * γ) * (FT(1) - h_frac^(FT(2) * γ)) /
+        (FT(1) - h_frac^(γ + FT(2)))
     )^(FT(1) / (FT(2) - γ))
     hmin = hmax .* h_frac
 
--- a/test/test_helpers.jl
+++ b/test/test_helpers.jl
@@ -150,8 +150,8 @@ function get_test_functions(cent_space, face_space)
     UVW = Geometry.UVWVector
     # Assemble (Cartesian) velocity
     ᶜu = @. UVW(Geometry.UVector(u)) +
-       UVW(Geometry.VVector(v)) +
-       UVW(Geometry.WVector(w))
+            UVW(Geometry.VVector(v)) +
+            UVW(Geometry.WVector(w))
     ᶠu = @. UVW(Geometry.UVector(ᶠu)) +
        UVW(Geometry.VVector(ᶠv)) +
        UVW(Geometry.WVector(ᶠw))
--- a/test/utilities.jl
+++ b/test/utilities.jl
@@ -50,8 +50,8 @@ end
     κ = zeros(cent_space)
     κ .= CA.compute_kinetic(uₕ, uᵥ)
     ᶜκ_exact = @. 1 // 2 *
-       cos(z)^2 *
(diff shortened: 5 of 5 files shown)


haakon-e:

- #3968 ◀
    - #3969

nefrathenrici:
Thank you for the thoughtful PR. I like the `join_lines_based_on_source` option, it has potential to make the code much more readable (especially the `Yₜ, Y, p, t,` and similar patterns). 
I agree that extra "headroom" is justifiable for horizontal groupings with long variable names, but increasing the margin to 120 is a large jump. I think we could increase it from 80 to 92 (the JuliaFormatter default) and see how we feel about it.

dennisYatunin:
I'm also down to change the `join_lines_based_on_source`, since it'll let us make some of our function signatures much more compact. I do like having things automatically collapse to one line when I remove arguments though; maybe there's some way to prevent automatic expansion to one item per line while still allowing automatic collapse? If not then setting `join_lines_based_on_source = true` should be fine.

As for the `margin` change, I think I prefer the current limit of 80 over 120. A limit of 80 allows two columns of code and a sidebar to fit on most monitors without line breaks, but anything over 100 will usually require line breaks to display two or more columns. I suppose this will be fine if you're only looking at one file at a time, but I don't know if that's the most typical workflow. If you want to increase it, 92 seems like a good compromise.

haakon-e:
Thanks for your thoughtful input @nefrathenrici and @dennisYatunin. 

Let me address the `margin` feedback. I fully support providing guidelines for how wide lines of code can be, for many reasons including those mentioned already. That said, there are numerous instances where code far exceed 80 (even 120) lines already, particularly in docstrings[^1]. I am of the view that there might be well-informed instances where it makes sense to deviate from guidelines[^2], so we should have guidelines not hard limits. Our developer docs already provide guidance on line width and I do not think we should make any changes to it. All this said, I again appreciate the feedback and will set the hard limit to `92`.

[^1]: examples: [code](https://github.com/CliMA/ClimaAtmos.jl/blob/4b075811ccd2878edc05a7d5f451842e272c1ff5/src/callbacks/get_callbacks.jl#L344), [code](https://github.com/CliMA/ClimaAtmos.jl/blob/4b075811ccd2878edc05a7d5f451842e272c1ff5/src/callbacks/get_callbacks.jl#L361), [code](https://github.com/CliMA/ClimaAtmos.jl/blob/4b075811ccd2878edc05a7d5f451842e272c1ff5/src/diagnostics/core_diagnostics.jl#L374), [code](https://github.com/CliMA/ClimaAtmos.jl/blob/4b075811ccd2878edc05a7d5f451842e272c1ff5/src/diagnostics/core_diagnostics.jl#L1503), [code](https://github.com/CliMA/ClimaAtmos.jl/blob/4b075811ccd2878edc05a7d5f451842e272c1ff5/src/solver/model_getters.jl#L374), [code](https://github.com/CliMA/ClimaAtmos.jl/blob/4b075811ccd2878edc05a7d5f451842e272c1ff5/src/solver/model_getters.jl#L466-L467), [code](https://github.com/CliMA/ClimaAtmos.jl/blob/4b075811ccd2878edc05a7d5f451842e272c1ff5/src/solver/type_getters.jl#L869), [docstring](https://github.com/CliMA/ClimaAtmos.jl/blob/4b075811ccd2878edc05a7d5f451842e272c1ff5/src/solver/types.jl#L720-L828), [docstring](https://github.com/CliMA/ClimaAtmos.jl/blob/4b075811ccd2878edc05a7d5f451842e272c1ff5/src/utils/era5_observations_to_forcing_file.jl#L269-L287), [docstring](https://github.com/CliMA/ClimaAtmos.jl/blob/4b075811ccd2878edc05a7d5f451842e272c1ff5/src/utils/era5_observations_to_forcing_file.jl#L533-L546), [docstring](https://github.com/CliMA/ClimaAtmos.jl/blob/4b075811ccd2878edc05a7d5f451842e272c1ff5/src/utils/variable_manipulations.jl#L263), [docstring](https://github.com/CliMA/ClimaAtmos.jl/blob/4b075811ccd2878edc05a7d5f451842e272c1ff5/src/utils/variable_manipulations.jl#L361), [docstring](https://github.com/CliMA/ClimaAtmos.jl/blob/4b075811ccd2878edc05a7d5f451842e272c1ff5/src/parameterized_tendencies/gravity_wave_drag/non_orographic_gravity_wave.jl#L571-L574), [docstring](https://github.com/CliMA/ClimaAtmos.jl/blob/4b075811ccd2878edc05a7d5f451842e272c1ff5/src/parameterized_tendencies/les_sgs_models/smagorinsky_lilly.jl#L12-L22), [docstring](https://github.com/CliMA/ClimaAtmos.jl/blob/4b075811ccd2878edc05a7d5f451842e272c1ff5/src/parameterized_tendencies/microphysics/microphysics_wrappers.jl#L380-L391), 

[^2]: You can already elide the juliaformatter completely on a per-file or per-block level, but I find the syntax somewhat [clunky](https://domluna.github.io/JuliaFormatter.jl/dev/#Turn-off/on-formatting).


----

> I do like having things automatically collapse to one line when I remove arguments though; maybe there's some way to prevent automatic expansion to one item per line while still allowing automatic collapse?

I don't this is currently possible, but opened an issue about it (https://github.com/domluna/JuliaFormatter.jl/issues/942) so we can update our formatter if this becomes possible in the future.

PR #3977, opened 2025-08-25
Add cloud top height diagnostic

oalcabes:
Hi everyone! I am trying to write this diagnostic for calculating cloud top height. The diagnostic needs two parameters: a threshold for minimum condensate to be considered a cloud, and an exponent for the heaviside function. I have currently hard coded these as 1e-10 and 4 respectively. Please let me know what you think and if there's a better way to do this!

### DIFF
--- a/src/diagnostics/core_diagnostics.jl
+++ b/src/diagnostics/core_diagnostics.jl
@@ -1201,6 +1201,67 @@ add_diagnostic_variable!(
     compute! = compute_clvi!,
 )
 
+###
+# Cloud top height
+###
+compute_cloud_top_height!(out, state, cache, time) = compute_cloud_top_height!(
+    out,
+    state,
+    cache,
+    time,
+    cache.atmos.moisture_model,
+)
+compute_cloud_top_height!(_, _, _, _, model::T) where {T} =
+    error_diagnostic_variable("cltz", model)
+
+#heaviside function version
+function compute_cloud_top_height!(
+    out,
+    state,
+    cache,
+    time,
+    moisture_model::T,
+) where {T <: Union{EquilMoistModel, NonEquilMoistModel}}
+
+    clw = @. lazy(state.c.ρ * cache.precomputed.cloud_diagnostics_tuple.q_liq)
+    cli = @. lazy(state.c.ρ * cache.precomputed.cloud_diagnostics_tuple.q_ice)
+    z = Fields.coordinate_field(state.c).z
+
+    FT = Spaces.undertype(axes(clw))
+
+    q_cond = @. lazy(clw + cli)
+
+    heaviside_num =
+        @. lazy(ifelse(q_cond > FT(1e-10), (q_cond * z)^(4) * z, FT(0)))
+
+    heaviside_denom =
+        @. lazy(ifelse(q_cond > FT(1e-10), (q_cond * z)^(4), FT(0)))
+
+    num = zeros(axes(Fields.level(state.f, half)))
+    denom = zeros(axes(Fields.level(state.f, half)))
+
+    Operators.column_integral_definite!(num, heaviside_num)
+    Operators.column_integral_definite!(denom, heaviside_denom)
+
+    if isnothing(out)
+        out = num ./ denom
+    else
+        out .= num ./ denom
+    end
+
+    return out
+end
+
+add_diagnostic_variable!(
+    short_name = "cltz",
+    long_name = "Cloud Top Height",
+    standard_name = "cloud_top_height", # NOT SURE IF STANDARD EXISTS
+    units = "m",
+    comments = """
+    The height of the cloud top based on some threshold.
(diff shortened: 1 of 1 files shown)


src/diagnostics/core_diagnostics.jl
```diff
@@ -1201,6 +1201,67 @@ add_diagnostic_variable!(
     compute! = compute_clvi!,
 )
 
+###
+# Cloud top height
```
Julians42, line 1205:
```suggestion
# Cloud top height (2d)
```

src/diagnostics/core_diagnostics.jl
```diff
@@ -1201,6 +1201,67 @@ add_diagnostic_variable!(
...
+        out .= num ./ denom
+    end
+
+    return out
+end
+
+add_diagnostic_variable!(
+    short_name = "cltz",
+    long_name = "Cloud Top Height",
+    standard_name = "cloud_top_height", # NOT SURE IF STANDARD EXISTS
```
Julians42, line 1258: Could check this table https://airtable.com/appYNLuWqAgzLbhSq/shrKcLEdssxb8Yvcp/tblL7dJkC3vl5zQLb

src/diagnostics/core_diagnostics.jl
```diff
@@ -1201,6 +1201,67 @@ add_diagnostic_variable!(
...
+    clw = @. lazy(state.c.ρ * cache.precomputed.cloud_diagnostics_tuple.q_liq)
+    cli = @. lazy(state.c.ρ * cache.precomputed.cloud_diagnostics_tuple.q_ice)
+    z = Fields.coordinate_field(state.c).z
+
+    FT = Spaces.undertype(axes(clw))
+
+    q_cond = @. lazy(clw + cli)
+
+    heaviside_num =
+        @. lazy(ifelse(q_cond > FT(1e-10), (q_cond * z)^(4) * z, FT(0)))
```
tapios, line 1235: This approach is not bad, but it will underestimate the cloud top (it will find the maximum of q_cond * z). Here's a bit more robust approach:

1. Define a logistic weight function `w(z) = 1 / (1 + exp(-k * (q_cond - q_threshold) )`, where `q_threshold` is a small threshold parameter (e.g., 1e-8).  This creates an approximately 1/0 cloud mask in a vector operation. The parameter `k` controls the steepness of the transition. 

2. Find the top of the mask `w(z)` in a soft reduction: `z_top = \int [w(z) exp(a*z)] z dz / \int [w(z) exp(a*z)] dz`. Here, `a` is a parameter that controls of how strongly the cloud top is favored in the integral. 

This approach will give a more accurate yet smooth measure of the cloud top.

src/diagnostics/core_diagnostics.jl
```diff
@@ -1201,6 +1201,67 @@ add_diagnostic_variable!(
...
+
+    heaviside_denom =
+        @. lazy(ifelse(q_cond > FT(1e-10), (q_cond * z)^(4), FT(0)))
```
tapios, line 1238: Ditto for exponent.

src/diagnostics/core_diagnostics.jl
```diff
@@ -1201,6 +1201,67 @@ add_diagnostic_variable!(
...
+
+    num = zeros(axes(Fields.level(state.f, half)))
+    denom = zeros(axes(Fields.level(state.f, half)))
+
+    Operators.column_integral_definite!(num, heaviside_num)
+    Operators.column_integral_definite!(denom, heaviside_denom)
+
+    if isnothing(out)
+        out = num ./ denom
```
tapios, line 1247: A more telling name instead of `out` (e.g., `z_clt`) would improve readability .

src/diagnostics/core_diagnostics.jl
```diff
@@ -1201,6 +1201,67 @@ add_diagnostic_variable!(
...
+
+    clw = @. lazy(state.c.ρ * cache.precomputed.cloud_diagnostics_tuple.q_liq)
+    cli = @. lazy(state.c.ρ * cache.precomputed.cloud_diagnostics_tuple.q_ice)
+    z = Fields.coordinate_field(state.c).z
+
+    FT = Spaces.undertype(axes(clw))
+
+    q_cond = @. lazy(clw + cli)
+
+    heaviside_num =
```
tapios, line 1234: It's not really a Heaviside function. Better call it `cloud_mask` or something like it (with the modifications below).

tapios (changes requested):
Thanks for putting this together. I outlined a more accurate approach that uses the same kinds of operations. I suggest you implement that (and experiment a bit with the parameters needed).

src/diagnostics/core_diagnostics.jl
oalcabes: thanks!!

PR #3978, opened 2025-08-26
Bugfix; use continue with for loops in sgs mass flux

sajjadazimi:
## Purpose 
Bugfix; use continue instead of return when using for loops instead of unrolled foreach


szy21 (approved):
Thank you!

PR #3983, opened 2025-08-26
set prognostic edmf bottom bc before environment

### LINKED ISSUE #3971: Fix calculation of environment u_3 in prognostic EDMF
Currently, u_3 at the boundary violates domain decomposition, and we recompute it when calculating the mass flux [here](https://github.com/CliMA/ClimaAtmos.jl/blob/f496d5ca327e7bb20fbb1b8d95e3143d9b13062c/src/prognostic_equations/edmfx_sgs_flux.jl#L55). We should move this to precomputed quantities.

### LINKED ISSUE #3981: Recomputing u_3 in sgs mass flux is problematic when environment area fraction is 0
I just realized [this](https://github.com/CliMA/ClimaAtmos.jl/blob/789a215e53f1157b1ba801a23735a0ee91e0450d/src/prognostic_equations/edmfx_sgs_flux.jl#L57) will cause an issue if the environment area fraction is close to 0 (which could happen in global runs). @sajjadazimi Could you modify it following e.g. [here](https://github.com/CliMA/ClimaAtmos.jl/blob/789a215e53f1157b1ba801a23735a0ee91e0450d/src/utils/variable_manipulations.jl#L488)?

szy21:
## Purpose 

This should solve the problem that domain decomposition is not correct at the boundary.
Closes #3981 Closes #3971 

## Content

----

### DIFF
--- a/src/cache/precomputed_quantities.jl
+++ b/src/cache/precomputed_quantities.jl
@@ -463,7 +463,6 @@ NVTX.@annotate function set_implicit_precomputed_quantities!(Y, p, t)
 
     if turbconv_model isa PrognosticEDMFX
         set_prognostic_edmf_precomputed_quantities_draft!(Y, p, ᶠuₕ³, t)
-        set_prognostic_edmf_precomputed_quantities_environment!(Y, p, ᶠuₕ³, t)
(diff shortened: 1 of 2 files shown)


szy21:
Oh wait, this is not correct. Let me try to fix it.

PR #3987, opened 2025-08-27
fix cached boundary values in env

### LINKED ISSUE #3971: Fix calculation of environment u_3 in prognostic EDMF

### LINKED ISSUE #3981: Recomputing u_3 in sgs mass flux is problematic when environment area fraction is 0

sajjadazimi:
## Purpose 
Fix cached boundary values in env by changing the order of functions calls in set_cache
Closes #3971 Closes #3981 

----

### DIFF
--- a/src/cache/precomputed_quantities.jl
+++ b/src/cache/precomputed_quantities.jl
@@ -411,20 +411,31 @@ end
 
 """
     set_implicit_precomputed_quantities!(Y, p, t)
+    set_implicit_precomputed_quantities_part1!(Y, p, t)
+    set_implicit_precomputed_quantities_part2!(Y, p, t)
 
-Updates the precomputed quantities that are handled implicitly based on the
-current state `Y`. This is called before each evaluation of either
-`implicit_tendency!` or `remaining_tendency!`, and it includes quantities used
+Update the precomputed quantities that are handled implicitly based on the
+current state `Y`. These are called before each evaluation of either
+`implicit_tendency!` or `remaining_tendency!`, and they include quantities used
 in both tedencies.
 
-This function also applies a "filter" to `Y` in order to ensure that `ᶠu³` is 0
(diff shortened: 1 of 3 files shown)


src/cache/precomputed_quantities.jl
```diff
@@ -502,6 +512,19 @@ NVTX.@annotate function set_explicit_precomputed_quantities!(Y, p, t)
 
     if turbconv_model isa PrognosticEDMFX
         set_prognostic_edmf_precomputed_quantities_bottom_bc!(Y, p, t)
+    elseif turbconv_model isa DiagnosticEDMFX
+        set_diagnostic_edmf_precomputed_quantities_bottom_bc!(Y, p, t)
```
szy21, line 529: The `bottom_bc` should be after `draft` (otherwise it is overwritten by the `draft` function). I think the easiest way to make this work is to move `bottom_bc` to implicit tendency, but after `draft`.

szy21: resolved offline: this works too.

szy21 (approved):
Thank you!

PR #3991, opened 2025-08-29
separate sedimentation of qt and mse in prognostic edmf

szy21:
## Purpose 

In preparation for making sedimentation of microphysics tracers implicit.

## Content

----


costachris (approved):
LGTM

szy21:
> Should we do the same for other microphysics tracers? I don't think that doing qt implicitly but doing for example rain explicitly will work well.

I want to do qt explicitly but other tracers implicitly, like what we do for the grid-mean.

PR #3999, opened 2025-09-02
Add new `AtmosSimulation` constructor

### LINKED ISSUE #3894: `AtmosSimulation` Constructor Refactor
This issue continues the refactoring to improve the user interface for setting up and running simulations. The ultimate goal is to move towards a more composable interface:
```julia
using ClimaAtmos
# Define the model, domain, and initial conditions as objects
model = AtmosModel(; moisture = Microphysics1Moment)
domain = SphereDomain(; h_elem = 30)
initial_condition = DryBaroclinicWave()
sim = AtmosSimulation(; model, domain, initial_condition)
solve_atmos!(sim)
```

The only way to construct an `AtmosSimulation` is via the monolithic `get_simulation` function, which requires an `AtmosConfig`. A new constructor which doesn't require `AtmosConfig` will be introduced. Responsibilities of `get_simulation` currently include comms_context setup, model and space creation, diagnostics initialization, and field/cache allocation.

### Core Changes

1.  **New `AtmosSimulation` Constructor**

    A new, composable `AtmosSimulation` constructor will be introduced. It will serve as the primary user-facing entry point and will have a signature similar to:
    ```julia
    AtmosSimulation(; model, domain, initial_condition, surface_setup, comms_context, params, dt, start_date, many_more_kwargs...)
    ```
    This constructor will be responsible for the final steps of setting up the simulation, including:
    *   Initializing diagnostics
    *   Allocating the state and the cache
    *   Creating the `integrator` object

2.  `AbstractDomain` structs

    To replace the use of parsed_args in `get_spaces`, new `AbstractDomain`s will be added. These structs will store all spatial information currently contained in `parsed_args` (e.g., `x_min`, `x_elem`, `z_stretch`, `nh_poly`, `dz_bottom`). The initial hierarchy matches `parsed_args["config"]` and will include:
    *   `SphereDomain`
    *   `ColumnDomain`
    *   `BoxDomain`
    *   `PlaneDomain`

###  **Refactoring of `parsed_args`-Dependent Functions**

  A significant part of this work involves refactoring the many "type getter" functions that currently depend on the `parsed_args`. These functions will have new methods that accept types/vals rather than parsed_args. The key functions include:

- `get_spaces`: Will be refactored to dispatch on the new AbstractDomain types.
- `get_callbacks`: Needs to be adapted to receive configuration (like output frequency) via explicit arguments or structs instead of parsed_args.
- `get_diagnostics`: Similar to callbacks, its behavior will need to be configured directly.  I expect this and the callbacks will be the biggest obstacle in the refactor, since it will be cumbersome to pass vectors of diagnostics/callbacks so we will need some convenience functions to simplify this aspect of simulation configuration.
- `get_surface_setup`: Turned into keyword argument: `AtmosSimulation(; surface_setup = ...)`
- `get_tracers`: Can be determined from the AtmosModel info for now since the only tracers available prescribed aerosols when applicable.
- `get_steady_state_velocity`: Will use val or types for topography info. 
- `get_state_restart`: Will be simplified to take just a file path and comms_ctx.
- `ode_configuration`: This will likely take many keyword arguments since it takes many parsed_args at the moment.

nefrathenrici:
## Purpose 

Adds a new AtmosSimulation constructor, closes #3894. 

The current `AtmosConfig`/`parsed_args` system is monolithic and hard to use programmatically. This PR introduces a more composable constructor that makes it easier to create simulations programmatically and reduces coupling between configuration parsing and simulation setup.

## Content

 - Added a new `AtmosSimulation` constructor that provides a cleaner API without requiring `AtmosConfig`/`parsed_args`
 - Introduced domain abstractions (`SphereDomain`, `ColumnDomain`, etc.) to replace parsed_args-based spatial configuration. This could use ClimaCore directly, as mentioned by @Sbozzolo.
 - Added `default_model_callbacks` to get callbacks from model info.
 - Added `common_callbacks` for standard simulation callbacks (NaN checks, conservation, etc.)
 - Refactored topography and mesh warp parsed_args into type hierarchies

## To-do

- Add unit tests
- Unify `get_callbacks` and `common_callbacks`
- Obtain tracers from model info
- Validate 
  - diagnostic/checkpoint frequencies match
  - reproducible restarts
  - callback frequencies with `dt`
- Add `netcdf_interpolation_num_points` and `netcdf_output_at_levels`
- Add support for scm forcing callbacks and other missing callbacks
- "verbose" option with all the logging from `get_simulation`
- More flexible timestepping configuration, some parsed_args are not supported
- Improve simulation validation with informative errors

----


src/simulation/domains.jl
```diff
@@ -0,0 +1,226 @@
...
+# Default constants for domain parameters
+const DEFAULT_Z_MAX = 30000.0
+const DEFAULT_Z_ELEM = 10
+const DEFAULT_DZ_BOTTOM = 500.0
+const DEFAULT_Z_STRETCH = true
+const DEFAULT_NH_POLY = 3
+const DEFAULT_BUBBLE = false
+const DEFAULT_DEEP_ATMOSPHERE = true
+const DEFAULT_H_ELEM = 6
+const DEFAULT_RADIUS = 6.371229e6
```
nefrathenrici: this should be a parameter

src/simulation/AtmosSimulations.jl
```diff
@@ -27,17 +31,265 @@ function Base.show(io::IO, sim::AtmosSimulation)
...
+    end
+
+    # TODO: Add steady state velocity
+    steady_state_velocity = nothing
+    p = build_cache(
+        Y, model, params, surface_setup, dt, start_date, tracers,
+        steady_state_velocity
+    )
+    discrete_hydrostatic_balance &&
+        set_discrete_hydrostatic_balanced_state!(Y, p)
```
nefrathenrici: unimportant

src/callbacks/get_callbacks.jl
```diff
@@ -207,165 +194,421 @@ function get_diagnostics(parsed_args, atmos_model, Y, p, sim_info, output_dir)
...
                         ClimaUtilities.TimeManager.date(t_start) :
                         start_date + Dates.Second(t_start),
         )
-        cond = let schedule = schedule
-            (u, t, integrator) -> schedule(integrator)
-        end
-        affect! = let output_dir = output_dir
-            (integrator) -> save_state_to_disk_func(integrator, output_dir)
-        end
-        callbacks = (callbacks..., SciMLBase.DiscreteCallback(cond, affect!))
```
nefrathenrici, line 298: Maybe this should be reverted, I'm not sure why the closure is needed.

src/callbacks/get_callbacks.jl
```diff
@@ -215,165 +201,421 @@ function get_diagnostics(parsed_args, atmos_model, Y, p, sim_info, output_dir)
     return diagnostics, writers, periods_reductions
 end
 
-function checkpoint_frequency_from_parsed_args(dt_save_state_to_disk::String)
-    if occursin("months", dt_save_state_to_disk)
-        months = match(r"^(\d+)months$", dt_save_state_to_disk)
+function parse_checkpoint_frequency(period::Number)
+    period == Inf && return Inf
+    # Treat number as seconds
+    # We use Millisecond to support fractional seconds, eg. 0.1
```
tapios: Do we really need fractional seconds for checkpointing? Seconds seems sufficient to me, and it would be less cumbersome without the factors of 1_000 in various places.

src/callbacks/get_callbacks.jl
```diff
@@ -215,165 +201,421 @@ function get_diagnostics(parsed_args, atmos_model, Y, p, sim_info, output_dir)
...
-        cond = let schedule = schedule
-            (u, t, integrator) -> schedule(integrator)
-        end
-        affect! = let output_dir = output_dir
-            (integrator) -> save_state_to_disk_func(integrator, output_dir)
-        end
-        callbacks = (callbacks..., SciMLBase.DiscreteCallback(cond, affect!))
+        cond = (u, t, integrator) -> schedule(integrator)
+        affect! = (integrator) -> save_state_to_disk_func(integrator, output_dir)
+        return (SciMLBase.DiscreteCallback(cond, affect!),)
```
tapios, line 295: Not directly related to this PR: But why do we need SciMLBase for callbacks (or anything really?)

src/simulation/AtmosSimulations.jl
```diff
@@ -27,17 +31,254 @@ function Base.show(io::IO, sim::AtmosSimulation)
...
+function AtmosSimulation{FT}(;
+    model = AtmosModel(),
+    params::Parameters.ClimaAtmosParameters = ClimaAtmosParameters(FT),
+    context::ClimaComms.AbstractCommsContext = ClimaComms.context(),
+    grid::Grids.AbstractGrid = SphereGrid(FT; radius = CAP.planet_radius(params), context),
+    initial_condition::ICs.InitialCondition = InitialConditions.DecayingProfile(),
+    dt = 600,
+    start_date = DateTime(2010, 1, 1),
+    t_start = 0,
+    t_end = 24 * 10 * 60 * 60,  # 10 days
```
tapios:
```suggestion
    t_end = 24 * 86400,  # 10 days
```
(In this line of business, 86400 should be hardwired in our brains for length of day. Better yet: make it a parameter.)

tapios (approved):
Looks good. Thank you. Definitely a useful step forward. Given that various aspects of the atmos model are about to change, it'll be good to get this in sooner (over potentially perfecting it further, which inevitably will come later).

src/callbacks/get_callbacks.jl
nefrathenrici: As far as I know, we have never needed fractional seconds for checkpointing. We can simplify this in a future PR.

tapios: Why not simplifying it here if we don't need it?

nefrathenrici: Sure, I can make the change now.

PR #4002, opened 2025-09-05
Add topography module and types

nefrathenrici:
## Content

----


src/topography/topography.jl
```diff
@@ -105,11 +107,107 @@ Surface elevation for 3D cosine hills.
 """
 function topography_cosine_3d(coord)
     FT = Geometry.float_type(coord)
-    (; x, y) = coord
-    (; h_max, λ) = cosine_params(FT)
```
dennisYatunin, line 109: Was there an inference issue that required changing this?

src/topography/topography.jl
```diff
@@ -105,11 +107,107 @@ Surface elevation for 3D cosine hills.
...
+    cos_params = cosine_params(FT)
+    h_max, λ = cos_params.h_max, cos_params.λ
     return topography_cosine(x, y, λ, λ, h_max)
 end
 
 topography_cosine(x, y, λ_x, λ_y, h_max) =
     h_max * cospi(2 * x / λ_x) * cospi(2 * y / λ_y)
 cosine_params(::Type{FT}) where {FT} = (; h_max = FT(25), λ = FT(25e3))
+
+abstract type AbstractTopography end
```
dennisYatunin, line 10: Giving this a method for `Base.broadcastable` would let you use `topo` instead of `(topo,)` in broadcast expressions.

src/topography/topography.jl
```diff
@@ -105,11 +107,107 @@ Surface elevation for 3D cosine hills.
...
+- `λ`: Wavelength parameter (m)
+- `a`: Mountain width parameter (m)
+"""
+Base.@kwdef struct ScharTopography{FT} <: AbstractTopography
+    h_max::FT = 25.0
+    x_center::FT = 50e3
+    λ::FT = 4e3
+    a::FT = 5e3
+end
+topo_func(::ScharTopography) = topography_schar
```
dennisYatunin:
```suggestion
topography_function((; h_max, x_center, λ, a)::ScharTopography) =
    (; x) -> h_max * exp(-(x - x_center)^2 / a^2) * cospi((x - x_center) / λ)^2
```
The methods of `topo_func` can be combined with the definitions of `topography_<...>`, using dispatch like in `steady_state_velocity`. To avoid having a lot of closures, you could also write this as
```suggestion
topography_function((; h_max, x_center, λ, a)::ScharTopography, (; x)) =
    h_max * exp(-(x - x_center)^2 / a^2) * cospi((x - x_center) / λ)^2
topography_function(topo) = Base.Fix1(topography_function, topo)
```
The topography parameters are currently being duplicated, e.g., in `ScharTopography` and `schar_params(FT)`. It would be good to reduce duplication by removing `schar_params(FT)` and using the fields of `ScharTopography` instead, like in this example.

dennisYatunin (approved):

src/topography/topography.jl
nefrathenrici: No, I think this can be reverted.

src/topography/topography.jl
nefrathenrici: Thanks, this is a much simpler pattern.

PR #4009, opened 2025-09-10
Use flux form terms for sedimentation

sajjadazimi:
## Purpose 
Use flux form terms for sedimentation because:
- Advective form is not stable for some setups
- The flux form is compatible with the grid mean equations

## To-do
Limit sources to avoid negative cloud liquid mass

----


sajjadazimi:
tempImagebP0d8K

PR #4011, opened 2025-09-11
change diffusion jacobian in the implicit solver

szy21:
## Purpose 

Set the tracer-density block due to diffusion to zero in the implicit solver Jacobian, as we think it is not important. What we currently have for this block is not correct because it doesn't calculate the full derivative with respect to rho. This seems to solve the inverted pressure gradient near the surface.

## Content

----

### DIFF
--- a/src/prognostic_equations/implicit/manual_sparse_jacobian.jl
+++ b/src/prognostic_equations/implicit/manual_sparse_jacobian.jl
@@ -619,11 +619,7 @@ function update_jacobian!(alg::ManualSparseJacobian, cache, Y, p, dtγ, t)
         end
 
         ∂ᶜρe_tot_err_∂ᶜρ = matrix[@name(c.ρe_tot), @name(c.ρ)]
-        @. ∂ᶜρe_tot_err_∂ᶜρ =
-            dtγ * ᶜdiffusion_h_matrix ⋅ DiagonalMatrixRow(
-                (-(1 + ᶜkappa_m) * ᶜe_tot - ᶜkappa_m * ∂e_int_∂q_tot * ᶜq_tot) /
-                ᶜρ,
-            )
+        @. ∂ᶜρe_tot_err_∂ᶜρ = zero(typeof(∂ᶜρe_tot_err_∂ᶜρ))
         @. ∂ᶜρe_tot_err_∂ᶜρe_tot +=
             dtγ * ᶜdiffusion_h_matrix ⋅ DiagonalMatrixRow((1 + ᶜkappa_m) / ᶜρ)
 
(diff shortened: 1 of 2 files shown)


src/prognostic_equations/implicit/manual_sparse_jacobian.jl
```diff
@@ -619,11 +619,13 @@ function update_jacobian!(alg::ManualSparseJacobian, cache, Y, p, dtγ, t)
         end
 
         ∂ᶜρe_tot_err_∂ᶜρ = matrix[@name(c.ρe_tot), @name(c.ρ)]
-        @. ∂ᶜρe_tot_err_∂ᶜρ =
-            dtγ * ᶜdiffusion_h_matrix ⋅ DiagonalMatrixRow(
-                (-(1 + ᶜkappa_m) * ᶜe_tot - ᶜkappa_m * ∂e_int_∂q_tot * ᶜq_tot) /
-                ᶜρ,
-            )
+        # TODO: Fix this block if we want to use it
```
tapios: If we are sure that tracer-mass consistency requires this to vanish, I'd just remove the code, rather than commenting it out.

src/prognostic_equations/implicit/manual_sparse_jacobian.jl
```diff
@@ -637,8 +639,10 @@ function update_jacobian!(alg::ManualSparseJacobian, cache, Y, p, dtγ, t)
                     ᶜ∂kappa_m∂q_tot *
                     (cp_d * T_0 + ᶜe_tot - ᶜK - ᶜΦ + ∂e_int_∂q_tot * ᶜq_tot)
                 ))
-            @. ∂ᶜρq_tot_err_∂ᶜρ =
-                dtγ * ᶜdiffusion_h_matrix ⋅ DiagonalMatrixRow(-(ᶜq_tot) / ᶜρ)
+            # TODO: Fix this block if we want to use it
```
tapios: Ditto.

src/prognostic_equations/implicit/manual_sparse_jacobian.jl
```diff
@@ -693,11 +699,11 @@ function update_jacobian!(alg::ManualSparseJacobian, cache, Y, p, dtγ, t)
             ∂ᶜρatke⁰_err_∂ᶜρ = matrix[@name(c.sgs⁰.ρatke), @name(c.ρ)]
             ∂ᶜρatke⁰_err_∂ᶜρatke⁰ =
                 matrix[@name(c.sgs⁰.ρatke), @name(c.sgs⁰.ρatke)]
+            # TODO: This ignores the diffusion tendency
```
tapios: Do we need the TODO?

src/prognostic_equations/implicit/manual_sparse_jacobian.jl
szy21: I left it here just in case we ever want to add it back. But I'm happy to remove it - it will be in the github history anyway.

tapios: I prefer the code clean...

PR #4012, opened 2025-09-12
Adds ability to run ERA5 simulations with temperature change

Julians42:
## Purpose 

In order to understand feedbacks on turbulence and convection closures under global warming we want to be able to run perturbation experiments. This PR introduces one setup that uses the ERA5 monthly diurnal setup and allows the user to increase the temperature (air and surface) by a specified amount with , e.g.,`era5_diurnal_warming: 4` for +4K experiments, in the config file. The specific humidity profile is adjusted by keeping the relative humidity constant. While this setup does not change large-scale dynamics, it does allow us to focus on feedbacks in just the boundary layer. 

## To-do

I have not added a buildkite configuration as an integration test because the base model is already tested. Instead I rely on unit tests to check that the forcing file created has the desired behavior (increased temperature and specific humidity). 

## Content

----

### DIFF
--- a/config/default_configs/default_config.yml
+++ b/config/default_configs/default_config.yml
@@ -271,6 +271,9 @@ site_latitude:
 site_longitude:
   help: "Site longitude for single column model. Used for externally driven time varying forcing model to generate the forcing file. Artifact support is currently for eastern Pacific region in July 2007 only. [`-149.0` (default)]"
   value: -149.0
+era5_diurnal_warming:
+  help: Applies a warming scenario by adding a constant temperature offset (in Kelvin). Supported only by the `ReanalysisMonthlyAveragedDiurnal` forcing model. This affects the target surface temperature for the entire simulation, the initial atmospheric temperature profile, and the relaxation temperature profile above `gcmdriven_relaxation_minimum_height`. Specific humidity is increased to hold relative humidity constant in the initial and relaxation profiles.
+  value: ~
 subsidence:
   help: "Subsidence [`nothing` (default), `Bomex`, `LifeCycleTan2018`, `Rico`, `DYCOMS`, `ISDAC`]"
   value: ~
--- a/src/solver/model_getters.jl
+++ b/src/solver/model_getters.jl
@@ -466,6 +466,10 @@ function get_external_forcing_model(parsed_args, ::Type{FT}) where {FT}
         @assert parsed_args["config"] == "column" "ReanalysisTimeVarying and ReanalysisMonthlyAveragedDiurnal are only supported in column mode."
         @assert all(reanalysis_required_fields .== "ReanalysisTimeVarying") "All of external_forcing, surface_setup, surface_temperature and initial_condition must be set to ReanalysisTimeVarying."
     end
+    if !isnothing(parsed_args["era5_diurnal_warming"])
+        @assert external_forcing == "ReanalysisMonthlyAveragedDiurnal" "era5_diurnal_warming is only supported for ReanalysisMonthlyAveragedDiurnal."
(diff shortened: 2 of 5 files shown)


test/test_helpers.jl
```diff
@@ -287,3 +295,60 @@ function create_mock_era5_datasets(
...
+end
+
+"""
+    shum_from_geopotential(geopotential_levels; q0 = 0.02, shum_scale_height = 2e4)
+
+Produce a typical specific humidity profile from geopotential levels.
+q0 is the specific humidity at the surface and shum_scale_height is the scale height in geopotential coordinates.
+The default values are a 2% mixing ratio at the surface and approximately a 2km scale height.
+"""
+function shum_from_geopotential(geopotential_levels; q0 = 0.02, shum_scale_height = 2e4)
```
szy21:
```suggestion
function shum_from_geopotential(geopotential_levels; q0 = 0.02, shum_scale_height = 2e3)
```

src/solver/model_getters.jl
```diff
@@ -466,6 +466,9 @@ function get_external_forcing_model(parsed_args, ::Type{FT}) where {FT}
         @assert parsed_args["config"] == "column" "ReanalysisTimeVarying and ReanalysisMonthlyAveragedDiurnal are only supported in column mode."
         @assert all(reanalysis_required_fields .== "ReanalysisTimeVarying") "All of external_forcing, surface_setup, surface_temperature and initial_condition must be set to ReanalysisTimeVarying."
     end
+    if parsed_args["era5_diurnal_warming"] isa Number
+        @assert external_forcing == "ReanalysisMonthlyAveragedDiurnal" "era5_diurnal_warming is only supported for ReanalysisMonthlyAveragedDiurnal."
+    end
```
imreddyTeja:
```suggestion
    if !isnothing(parsed_args["era5_diurnal_warming"])
        @assert external_forcing == "ReanalysisMonthlyAveragedDiurnal" "era5_diurnal_warming is only supported for ReanalysisMonthlyAveragedDiurnal."
        @assert parsed_args["era5_diurnal_warming"] isa Number "era5_diurnal_warming is expected to be a number, but was supplied as a $typeof(parsed_args["era5_diurnal_warming"])"
    end
```

test/test_helpers.jl
Julians42: The scale height here is in units of geopotential so 2e4 / g is about 2e3 m (which is the scale height we want to use). I'll change this to be more clear

Julians42: The profiles (with the slight change g=9.81 as opposed to 10) look like this:
Screenshot 2025-09-16 at 1 33 46 PM

config/default_configs/default_config.yml
```diff
@@ -271,6 +271,9 @@ site_latitude:
 site_longitude:
   help: "Site longitude for single column model. Used for externally driven time varying forcing model to generate the forcing file. Artifact support is currently for eastern Pacific region in July 2007 only. [`-149.0` (default)]"
   value: -149.0
+era5_diurnal_warming:
+  help: "Add this number of Kelvin to the ERA5 diurnal cycle and adjust specific humidity by holding relative humidity constant. Only supported for the `ReanalysisMonthlyAveragedDiurnal` external forcing model."
```
costachris: Can you clarify here how exactly temperature is perturbed (surface only, constant through profile). Also, this reads to me like it's added to the diurnal cycle, but I think you just mean it's added to the setup with a diurnal cycle (?).

Julians42: We discussed this offline and settled on a long but informative `help` string.

Julians42: Well online but offline ;)

Issue #3681, opened 2025-03-06
`vert_diff` config is marked as a `bool` but takes string values


Sbozzolo:
`vert_diff` config is marked as a `bool` but takes string values:
```julia
vert_diff:
  help: "Vertical diffusion [`false` (default), `VerticalDiffusion`, `true` (defaults to `VerticalDiffusion`), `DecayWithHeightDiffusion`]"
  value: false
```

This also triggers a warning when running a simulation

szy21:
I can change the default to nothing if that's better? (This will technically break the coupler as there are a few vert_diff: true there but I can change them too.)

juliasloan25:
Closed by #4017 - `vert_diff` now consistently takes Strings

PR #4017, opened 2025-09-16
Fixes in default config types

trontrytel:
Without this you get 

```
┌ Warning: Failed to convert `config_dict["prognostic_surface"] = PrognosticSurfaceTemperature` to default type Bool, keeping original value
└ @ ClimaAtmos ~/clones/ClimaAtmos.jl/src/solver/yaml_helper.jl:78
┌ Warning: Failed to convert `config_dict["vert_diff"] = DecayWithHeightDiffusion` to default type Bool, keeping original value
└ @ ClimaAtmos ~/clones/ClimaAtmos.jl/src/solver/yaml_helper.jl:78
```

and the model proceeds to run with the default.


akshaysridhar (approved):

ray-chew (approved):
Thanks for spotting and fixing these.

szy21:
FYI the model shouldn't proceed to run with the default. It keeps the "orginial value", which is the string you set in the yaml. It just can't convert it to bool. Here is the relevant issue: https://github.com/CliMA/ClimaAtmos.jl/issues/3681.

It is still good to fix it though, thanks! @nefrathenrici I wonder if the current fix will results in a warning if a user sets `vert_diff: false`? Maybe better to set the default to nothing? To be clear I'm fine with merging this, just wondering.

nefrathenrici:
> It is still good to fix it though, thanks! @nefrathenrici I wonder if the current fix will results in a warning if a user sets `vert_diff: false`? Maybe better to set the default to nothing? To be clear I'm fine with merging this, just wondering.

I think it's better to set the default to nothing to avoid the warning in the case you described, but it doesn't make a big difference.

PR #4020, opened 2025-09-17
Fix prognostic tv test by reducing io

Julians42:
## Purpose 

This pr addresses a slowdown in the ci test for `config/model_configs/prognostic_edmfx_tv_era5driven_column.yml`. In order to continue supporting ci plots for edmf for longer time periods, I also updated the `time_to_seconds` function to handle additional cases from `ClimaAnalysis.available_periods`.

## Content

----


trontrytel (approved):
Thank you! And looks like the other cases were a hiccup in the cluster. So lets not change any other times and just merge this

haakon-e (approved):
Thanks for adding some tests!

PR #4021, opened 2025-09-17
Use CommonSpaces/Grids

nefrathenrici:
## Content

----


post_processing/ci_plots.jl
```diff
@@ -1485,14 +1493,99 @@ end
...
+    )
+end
+
+function make_plots(
+    sim_type::Union{
+        EDMFColumnPlots,
+        EDMFColumnPlotsWithPrecip,
+    },
+    output_paths::Vector{<:AbstractString},
+)
```
dennisYatunin: It would probably be simpler to use one function that handles both boxes and columns, with an if-statement to distinguish between them in `map_comparison`, instead of defining a duplicate function that only handles columns

.buildkite/Manifest-v1.11.toml
```diff
@@ -1,6 +1,6 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.11.6"
+julia_version = "1.11.4"
```
dennisYatunin: Was this downgrade on purpose?

.buildkite/PrecompileCI/src/PrecompileCI.jl
```diff
@@ -14,34 +13,38 @@ import ClimaParams
...
+            periodic_x = true, periodic_y = true, topography,
+        )
+        plane_grid = CA.PlaneGrid(
+            FT; context, x_elem, x_max, z_elem, z_max, nh_poly, z_stretch,
+            dz_bottom, bubble, periodic_x = true,
+            topography,
+        )
+        column_grid = CA.ColGrid(
+            FT; context, z_elem, z_max, z_stretch, dz_bottom,
+        )
```
dennisYatunin, line 49: The keyword arguments of these constructors could be grouped in a more consistent way, like with horizontal discretization parameters on one row and vertical discretization parameters on another

post_processing/ci_plots.jl
dennisYatunin: It would probably be simpler to use one function that handles both boxes and columns, with an if-statement to distinguish between them in `map_comparison`, instead of duplicating the box function to handle columns

reproducibility_tests/ref_counter.jl
```diff
@@ -1,10 +1,10 @@
-277
+278
 
 # **README**
 #
 # What is the ref_counter?
 #
-# The ref_counter is part of reproduciability tests. The counter identifies a
+# The ref_counter is part of reproducability tests. The counter identifies a
```
dennisYatunin:
```suggestion
# The ref_counter is part of reproducibility tests. The counter identifies a
```

src/diagnostics/default_diagnostics.jl
```diff
@@ -168,22 +169,31 @@ function core_default_diagnostics(output_writer, duration, start_date)
...
+            average_func(core_diagnostics...; output_writer, start_date)...,
+            min_func("ts"; output_writer, start_date),
+            max_func("ts"; output_writer, start_date)]
+    else
+        [
+            average_func(core_diagnostics...; output_writer, start_date)...,
+            min_func("ts"; output_writer, start_date),
+            max_func("ts"; output_writer, start_date),
+        ]
+    end
```
dennisYatunin:
```suggestion
    core_diagnostics = [
        average_func(core_diagnostics...; output_writer, start_date)...,
        min_func("ts"; output_writer, start_date),
        max_func("ts"; output_writer, start_date),
    ]
    if topography
        # We need to compute the topography at the beginning of the simulation
        # (and only at the beginning), so we set output/compute_schedule_func to
        # false. It is still computed at the very beginning.
        orog_diagnostic = ScheduledDiagnostic(;
            variable = get_diagnostic_variable("orog"),
            output_schedule_func = (integrator) -> false,
            compute_schedule_func = (integrator) -> false,
            output_writer,
            output_short_name = "orog_inst",
        )
        push!(core_diagnostics, orog_diagnostic)
    end
```
This could be a bit less repetitive

src/prognostic_equations/advection.jl
```diff
@@ -240,7 +240,7 @@ NVTX.@annotate function explicit_vertical_advection_tendency!(Yₜ, Y, p, t)
 
     if point_type <: Geometry.Abstract3DPoint
         @. ᶜω³ = wcurlₕ(Y.c.uₕ)
-    elseif point_type <: Geometry.Abstract2DPoint
+    elseif point_type <: Union{Geometry.Abstract2DPoint, Geometry.ZPoint}
```
dennisYatunin:
```suggestion
    else
```
The vorticity expression below is correct for all non-3D points, not just 2D and Z points

src/simulation/grids.jl
```diff
@@ -0,0 +1,352 @@
...
+- `nh_poly`: Polynomial order
+- `bubble`: Enable bubble correction
+- `deep_atmosphere`: Enable deep atmosphere
+- `topography`: Topography type
+- `topography_damping_factor`: Topography damping
+- `mesh_warp_type`: Mesh warping type ([`SLEVEWarp`](@ref) or [`LinearWarp`](@ref))
+- `topo_smoothing`: Apply topography smoothing
+"""
+function SphereGrid(
+    FT;
```
dennisYatunin:
```suggestion
    ::Type{FT};
```
Making `FT` inferable will probably reduce the latency of all these functions.

src/simulation/grids.jl
```diff
@@ -0,0 +1,352 @@
...
+- `FT`: Floating point type
+
+# Keyword Arguments
+- `context`: Communications context
+- `z_elem`: Number of vertical elements
+- `z_max`: Maximum height
+- `z_stretch`: Whether to use vertical stretching
+- `dz_bottom`: Bottom layer thickness for stretching
+"""
+function ColGrid(
```
dennisYatunin:
```suggestion
function ColumnGrid(
```
This can have a clearer name if `CommonGrids.ColumnGrid` is qualified instead of being directly imported. ClimaCore functions are almost always qualified in ClimaAtmos, so these functions should probably do the same for `CommonGrids`.

src/simulation/grids.jl
```diff
@@ -0,0 +1,352 @@
...
+Create a hypsography function that handles topography integration.
+"""
+function hypsography_function_from_topography(
+    FT::Type{<:AbstractFloat},
+    topography::AbstractTopography,
+    topography_damping_factor,
+    mesh_warp_type::MeshWarpType,
+    topo_smoothing,
+)
+    return function (h_grid, z_grid)
```
dennisYatunin:
```suggestion
function hypsography_function_from_topography(
    ::Type{FT},
    topography::AbstractTopography,
    topography_damping_factor,
    mesh_warp_type::MeshWarpType,
    topo_smoothing,
) where {FT}
    function hypsography(::Type{FT}, h_grid, z_grid) where {FT}
        ...
    end
    return Base.Fix1(hypsography, FT)
```
If it turns out that making `FT` inferable does help with latency, you'll need to pull some shenanigans similar to this for it to be inferable in your hypsography closure

src/simulation/grids.jl
```diff
@@ -0,0 +1,352 @@
...
+            @info "Remapping Earth orography from ETOPO2022 data onto horizontal space"
+        else
+            z_surface = SpaceVaryingInput(topography_function(topography), h_space)
+            @info "Using $(nameof(typeof(topography))) orography"
+        end
+
+        # Apply topography-specific processing
+        if topography isa EarthTopography
+            mask(x::FT) where {FT} = x * FT(x > 0)
+            z_surface = @. mask(z_surface)
```
dennisYatunin:
```suggestion
            @. z_surface = min(z_surface, 0)
```
Writing this in terms of `min` will probably reduce latency, and there's no need to allocate a new `z_surface` field.

But more importantly, if the actual topography data includes negative values (I'm guessing landlocked areas below sea level), shouldn't we account for them in the final diffused surface elevation? It feels strange to just clip them to zero.

src/simulation/grids.jl
```diff
@@ -0,0 +1,352 @@
...
+            diff_courant = FT(0.05)
+            Δh_scale = Spaces.node_horizontal_length_scale(h_space)
+            κ = FT(diff_courant * Δh_scale^2)
+            maxiter = Int(round(log(topography_damping_factor) / diff_courant))
+            Hypsography.diffuse_surface_elevation!(z_surface; κ, dt = FT(1), maxiter)
+            # Coefficient for horizontal diffusion may alternatively be
+            # determined from the empirical parameters suggested by
+            # E3SM v1/v2 Topography documentation found here: 
+            # https://acme-climate.atlassian.net/wiki/spaces/DOC/pages/1456603764/V1+Topography+GLL+grids
+            z_surface = @. mask(z_surface)
```
dennisYatunin: If the values going into `diffuse_surface_elevation!` are all positive, you shouldn't need to clip them to 0 again.

However, I think it might be better to avoid clipping `z_surface` before its values are diffused, and only clip it afterward. That way, the negative values will be properly accounted for in the diffused surface elevation.

Moreover, I'm not even sure if we need to clip at all, since the ClimaCore operators should be able to handle negative values of `z`. But maybe there's something that breaks in ClimaAtmos if we don't clip?

src/solver/type_getters.jl
```diff
@@ -627,39 +538,71 @@ function auto_detect_restart_file(
...
+    return output_dir, final_restart_file
+end
+
+function get_sim_info(config::AtmosConfig)
+    (; comms_ctx, parsed_args) = config
+    FT = eltype(config)
+
+    (; job_id) = config
+
+    output_dir, restart_file = CA.setup_output_dir(
```
dennisYatunin:
```suggestion
    output_dir, restart_file = setup_output_dir(
```
This is defined in the same file so it shouldn't need to be qualified

src/solver/type_getters.jl
```diff
@@ -770,10 +713,100 @@ function get_comms_context(parsed_args)
     return comms_ctx
 end
 
+function get_mesh_warp_type(FT, parsed_args)
+    warp_type_str = parsed_args["mesh_warp_type"]
+    if warp_type_str == "SLEVE"
+        return SLEVEWarp{FT}(
+            eta = FT(parsed_args["sleve_eta"]),
+            s = FT(parsed_args["sleve_s"]),
```
dennisYatunin:
```suggestion
            eta = parsed_args["sleve_eta"],
            s = parsed_args["sleve_s"],
```
I think the `FT`s here are superfluous

src/solver/type_getters.jl
```diff
@@ -770,10 +713,100 @@ function get_comms_context(parsed_args)
...
+    elseif warp_type_str == "Linear"
+        return LinearWarp()
+    else
+        error(
+            "Unknown mesh warp type string: $warp_type_str. Supported types are 'SLEVE' and 'Linear'",
+        )
+    end
+end
+
+function get_grid(parsed_args, params, context)
```
dennisYatunin, line 737: It would be good to avoid duplicating the keyword arguments here. The simplest way to do this would probably be to gather them in a `NamedTuple` and then splat them into the constructors. For example, you could avoid duplicating the arguments related to columns and topography by doing something like
```julia
        kwargs = (
            z_elem = parsed_args["z_elem"],
            z_max = parsed_args["z_max"],
            z_stretch = parsed_args["z_stretch"],
            dz_bottom = parsed_args["dz_bottom"],
        )
        if parsed_args["config"] != "column"
            kwargs = (
                kwargs...,
                topography = get_topography(FT, parsed_args),
                topography_damping_factor = parsed_args["topography_damping_factor"],
                mesh_warp_type = get_mesh_warp_type(FT, parsed_args),
                topo_smoothing = parsed_args["topo_smoothing"],
            )
        end
```

src/solver/type_getters.jl
```diff
@@ -792,14 +825,16 @@ function get_simulation(config::AtmosConfig)
...
+            spaces = (; center_space = axes(Y.c), face_space = axes(Y.f))
             # Fix the t_start in sim_info with the one from the restart
             sim_info = merge(sim_info, (; t_start))
         end
         @info "Allocating Y: $s"
     else
-        spaces = get_spaces(config.parsed_args, params, config.comms_ctx)
+        spaces = get_spaces(grid)
     end
+    @info spaces.center_space.grid
```
dennisYatunin: If you're planning to leave this in the PR, maybe use `@show` instead of `@info`, or add a string to clarify what's being printed

src/surface_conditions/surface_conditions.jl
```diff
@@ -381,11 +381,11 @@ end
 #For non-RCEMIPII box models with prescribed surface temp, assume that the latitude is 0.
 function surface_temperature(
     ::Union{ZonallySymmetricSST, ZonallyAsymmetricSST},
-    coordinates::Union{Geometry.XZPoint, Geometry.XYZPoint},
+    coordinates::Union{Geometry.XZPoint, Geometry.XYZPoint, Geometry.ZPoint},
```
dennisYatunin:
```suggestion
    coordinates,
```
Unless the coordinate type restriction is being used for dispatch, it should probably be removed (both here and in the other methods for `surface_temperature`).

There's no need for ClimaAtmos to keep track of all point types that include `z`-coordinates. That kind of code should be limited to only ClimaCore's `Geometry` module, where the point types are defined. Mirroring the `Geometry` module's code duplication in ClimaAtmos will only introduce additional friction when we're trying to refactor it later.

src/utils/common_spaces.jl
```diff
@@ -3,30 +3,6 @@ using ClimaCore:
...
-        Geometry.YPoint(zero(y_max)),
-        Geometry.YPoint(y_max);
-        periodic = true,
-    )
-    domain = Domains.RectangleDomain(x_domain, y_domain)
-    return Meshes.RectilinearMesh(domain, x_elem, y_elem)
-end
-
 # h_elem is the number of elements per side of every panel (6 panels in total)
 function cubed_sphere_mesh(; radius, h_elem)
```
dennisYatunin: Are the `cubed_sphere_mesh` and `make_horizontal_space` functions still needed after this PR?

dennisYatunin (approved):
This is great! Excited to have columns that are actually real columns after this PR

src/solver/type_getters.jl
nefrathenrici: I thought this log line would replace the existing Resolution stats, but I can extract and log them to keep the existing format:
```
┌ Info: Resolution stats:
│   Nq = 4
│   h_elem = 6
│   z_elem = 10
│   ncols = 3456
└   ndofs_total = 34560
```
Here's what this PR prints (different grid from above):
```
┌ Info: ExtrudedFiniteDifferenceGrid:
│   horizontal:
│     context: SingletonCommsContext using CUDADevice
│     mesh: 100×1-element RectilinearMesh of RectangleDomain: x ∈ [0.0,100000.0] (periodic) × y ∈ [0.0,1000.0] (periodic)
│     quadrature: 4-point Gauss-Legendre-Lobatto quadrature
│   vertical:
└     mesh: 100-element IntervalMesh of IntervalDomain: z ∈ [0.0,21000.0] (:bottom, :top)
```

src/simulation/grids.jl
```diff
@@ -0,0 +1,338 @@
...
+                "z",
+                h_space,
+            )
+            @info "Remapping Earth orography from ETOPO2022 data onto horizontal space"
+        else
+            z_surface = SpaceVaryingInput(topography_function(topography), h_space)
+            @info "Using $(nameof(typeof(topography))) orography"
+        end
+
+        if topography isa EarthTopography
```
nefrathenrici, line 302: if topo_smoothing

PR #4022, opened 2025-09-17
Update WeatherModel initial conditions

costachris:
Add WeatherModel initial conditions for prognostic EDMF. Interpolate IC to arbitrary vertical grid. Generalize paths for IC.

### DIFF
--- a/config/model_configs/diagnostic_edmfx_era5_initial_condition.yml
+++ b/config/model_configs/diagnostic_edmfx_era5_initial_condition.yml
@@ -4,9 +4,7 @@ z_elem: 63
 dz_bottom: 30.0
 dt: 120secs
 t_end: 12hours
-# start_date: "20250627-0000"
-start_date: "20250831"
-era5_initial_condition_dir: "/net/sampo/data1/wxquest_data/initial_conditions"
+start_date: "20250627-0000"
 initial_condition: "WeatherModel"
 discrete_hydrostatic_balance: false # TODO: change to true once we fix discrete hydrostatic balance to work with orography
 topo_smoothing: true
--- a/config/model_configs/diagnostic_edmfx_era5_initial_condition.yml
+++ b/config/model_configs/diagnostic_edmfx_era5_initial_condition.yml
@@ -15,7 +15,6 @@ hyperdiff: ClimaHyperdiffusion
 divergence_damping_factor: 50.0 # large value to damp IC waves
 scalar_hyperdiffusion_coefficient: 0.929738
 vorticity_hyperdiffusion_coefficient: 0.1857
-
 insolation: "timevarying"
 surface_setup: DefaultMoninObukhov
 rad: allskywithclear
@@ -30,14 +29,14 @@ edmfx_detr_model: "Generalized"
 edmfx_nh_pressure: true
 edmfx_sgs_mass_flux: true
 edmfx_sgs_diffusive_flux: true
-moist: equil
-cloud_model: "grid_scale"
+cloud_model: "quadrature_sgs"
 precip_model: 0M
+moist: equil
 toml: [toml/diagnostic_edmfx_era5_ic.toml]
 use_itime: true
 output_default_diagnostics: false
 diagnostics:
-  - short_name: [ta, ua, wa, va, rhoa, hur, hus, clw, cli]
+  - short_name: [ts, ta, thetaa, ha, pfull, rhoa, ua, va, wa, hur, hus, cl, clw, cli, hussfc, evspsbl, pr]
     period: 60mins
   - short_name: [rsut, rlut, ts, pr, mslp]
     period: 60mins
--- a/config/model_configs/diagnostic_edmfx_era5_initial_condition.yml
+++ b/config/model_configs/diagnostic_edmfx_era5_initial_condition.yml
@@ -4,7 +4,9 @@ z_elem: 63
 dz_bottom: 30.0
 dt: 120secs
 t_end: 12hours
-start_date: "20250627-0000"
+# start_date: "20250627-0000"
+start_date: "20250831"
+era5_initial_condition_dir: "/net/sampo/data1/wxquest_data/initial_conditions"
 initial_condition: "WeatherModel"
 discrete_hydrostatic_balance: false # TODO: change to true once we fix discrete hydrostatic balance to work with orography
 topo_smoothing: true
@@ -36,7 +38,7 @@ toml: [toml/diagnostic_edmfx_era5_ic.toml]
 use_itime: true
 output_default_diagnostics: false
 diagnostics:
-  - short_name: [ts, ta, thetaa, ha, pfull, rhoa, ua, va, wa, hur, hus, cl, clw, cli, hussfc, evspsbl, pr]
+  - short_name: [ts, ta, thetaa, ha, pfull, rhoa, ua, va, wa, hur, hus, cl, clw, cli, hussfc, evspsbl]
     period: 60mins
-  - short_name: [rsut, rlut, ts, pr, mslp]
+  - short_name: [rsut, rlut, tas, pr, mslp]
     period: 60mins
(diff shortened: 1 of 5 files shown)


src/initial_conditions/initial_conditions.jl
```diff
@@ -377,7 +380,24 @@ function overwrite_initial_conditions!(
...
 )
     extrapolation_bc = (Intp.Periodic(), Intp.Flat(), Intp.Flat())
-    file_path = weather_model_data_path(initial_condition.start_date)
+
+    # Extract face coordinates and compute center midpoints
+    z_coords = Fields.axes(Y.c).grid.vertical_grid.topology.mesh.faces
+
+    # Compute center coordinates as midpoints between faces
+    face_z_values = [z.z for z in z_coords]
+    center_z_values = [
```
costachris: probably a better way to do this, but just getting the centers from the state fetches the full 3D grid.

src/initial_conditions/initial_conditions.jl
```diff
@@ -214,7 +215,9 @@ function (initial_condition::Union{MoistFromFile, WeatherModel})(params)
         grav = CAP.grav(params)
         thermo_params = CAP.thermodynamics_params(params)
 
-        T, p = FT(NaN), FT(NaN) # placeholder values
+        # Use safe defaults to avoid transient NaNs before overwrite
+        T = FT(CAP.T_0(params))
+        p = FT(CAP.MSLP(params))
```
Julians42: You do leave a comment but I don't understand why NaN's would come up here? I didn't run into this issue before. It doesn't really matter though since we're overwriting anyway (and the fields seem reasonably allocated in the simulations)...

config/model_configs/diagnostic_edmfx_era5_initial_condition.yml
```diff
@@ -30,14 +29,14 @@ edmfx_detr_model: "Generalized"
...
-cloud_model: "grid_scale"
+cloud_model: "quadrature_sgs"
 precip_model: 0M
+moist: equil
 toml: [toml/diagnostic_edmfx_era5_ic.toml]
 use_itime: true
 output_default_diagnostics: false
 diagnostics:
-  - short_name: [ta, ua, wa, va, rhoa, hur, hus, clw, cli]
+  - short_name: [ts, ta, thetaa, ha, pfull, rhoa, ua, va, wa, hur, hus, cl, clw, cli, hussfc, evspsbl, pr]
```
Julians42:
```suggestion
  - short_name: [ta, thetaa, ha, pfull, rhoa, ua, va, wa, hur, hus, cl, clw, cli, hussfc, evspsbl, pr]
```
You have `ts` twice - once on this line once on the next - I know its important but not that important ;)

src/initial_conditions/initial_conditions.jl
```diff
@@ -377,7 +380,24 @@ function overwrite_initial_conditions!(
...
+    z_coords = Fields.axes(Y.c).grid.vertical_grid.topology.mesh.faces
+
+    # Compute center coordinates as midpoints between faces
+    face_z_values = [z.z for z in z_coords]
+    center_z_values = [
+        (face_z_values[i] + face_z_values[i + 1]) / 2 for
+        i in 1:(length(face_z_values) - 1)
+    ]
+
+    target_levels = Array(center_z_values)
```
Julians42:
```suggestion
    target_levels = Fields.field2array(Fields.coordinate_field(Y.c).z)[:, 1]
```
Is maybe the shortest fix but is a little chaotic with the indexing

src/initial_conditions/initial_conditions.jl
Julians42: This is currently getting an approximate coordinate field from the unwarped mesh. One idea but that requires getting the z-levels from the warped mesh is to get a column coordinate field:
```julia
target_levels = Fields.field2array(Fields.coordinate_field(Y.c).z)[:, 1] # or any column index
```
This doesn't guarantee having 15 meters as the bottom level, e.g., starts out as
```julia
infil> Fields.field2array(Fields.coordinate_field(Y.c).z)[:, 1]
63-element Vector{Float32}:
    21.738264
    53.145973
    87.513336
   125.11682
   166.25824
```
I'm not sure if we care to get 15 meters at the bottom but if we do we could just argmin it like:
```julia
target_levels = Fields.field2array(Fields.coordinate_field(Y.c).z)[:, argmin(Fields.field2array(Fields.coordinate_field(Y.c).z))[2]]
```
Then we wouldn't be interpolating below the lowest level a second time ....

Julians42: @akshaysridhar and I chatted and maybe he has another idea that's more robust but I think this is a good balance between concise code and just doing it ...

src/initial_conditions/initial_conditions.jl
```diff
@@ -483,36 +502,125 @@ function _overwrite_initial_conditions_from_file!(
         )
     Y.c.uₕ .= C12.(Geometry.UVVector.(vel))
     Y.f.u₃ .= ᶠinterp.(C3.(Geometry.WVector.(vel)))
+
+    # Load cloud/precip water fields if present (fallback to zeros)
+    has_liq_ice = hasproperty(Y.c, :ρq_liq) && hasproperty(Y.c, :ρq_ice)
+    has_precip = hasproperty(Y.c, :ρq_rai) && hasproperty(Y.c, :ρq_sno)
+    # TEMP: diagnose 1M condensed/precip species at runtime (initialize to zero here)
+    diagnose_1m = true
```
Julians42: Currently this doesn't make sense with the config that's running on buildkite (the config has 0M) and so the code is relying on the check of `has_liq_ice` etc and is not consistent with the config. Ideally we should pass this all the way through from the `WeatherModel` struct when its constructed. 

I also think the if try blocks could be condensed into a helper function, which takes the pair (`:ᶜq_liq: "clwc"`), initializes to zeros and tries to fill, as this is a good amount of duplicated code.

src/utils/weather_model.jl
```diff
@@ -3,23 +3,60 @@ using Dates
 import ClimaInterpolations.Interpolation1D: interpolate1d!, Linear, Flat
 import ..parse_date
 
+
 """
-    weather_model_data_path(start_date)
+    weather_model_data_path(start_date, target_levels, era5_initial_condition_dir])
```
Julians42, line 8: Glad you got rid of the hardcoded `target_level` file - thank you!

src/utils/weather_model.jl
```diff
@@ -168,6 +189,22 @@ function to_z_levels(source_file, target_file, target_levels, FT)
...
+            var_var = defVar(
+                ncout,
+                var_name,
+                FT,
+                ("lon", "lat", "z"),
+                attrib = ncin[var_name].attrib,
+            )
+            var_var[:, :, :] = interpz_3d(target_levels, source_z, FT.(var_data))
+        end
+    end
```
Julians42, line 185: Should we rewrite the above code so that it uses this nice loop? E.g. why not `vcat(required_vars, opt_vars)` and then loop through all of them together?

Julians42 (changes requested):
At this point it might be easier to break off from the DYAMOND case so we're less restricted. I think this is causing problems for the `_overwrite_initial_conditions_from_file!` where you have to write a lot of `if else` statements. Happy to take some of this on if you are busy.

src/initial_conditions/initial_conditions.jl
costachris: I removed the species initialization for now, since it's unclear whether we will need that in the end.

src/initial_conditions/initial_conditions.jl
costachris: reverted to NaN

costachris:
> At this point it might be easier to break off from the DYAMOND case so we're less restricted. I think this is causing problems for the `_overwrite_initial_conditions_from_file!` where you have to write a lot of `if else` statements. Happy to take some of this on if you are busy.

Agreed. I consolidated the logic into the existing `overwrite_initial_conditions!` for WeatherModel, so we no longer rely on the common `_overwrite_initial_conditions_from_file!`

src/initial_conditions/initial_conditions.jl
costachris: this is much cleaner but doesn't work on the GPU, looking for alternatives

PR #4023, opened 2025-09-18
Add cloud top height diagnostic

oalcabes:
This diagnostic calculates a resolution invariant cloud top height. It is sensitive to three parameters that have been added to ClimaParams -- "cloud_top_threshold," "cloud_top_k," and "cloud_top_a." Results are sensitive to parameter choices. I hand tuned to get somewhat decent default parameters with these results: 
Screenshot 2025-09-18 at 9 52 14 AM


src/diagnostics/core_diagnostics.jl
```diff
@@ -1114,6 +1114,64 @@ add_diagnostic_variable!(
...
+    long_name = "Cloud Top Height",
+    standard_name = "",
+    units = "m",
+    comments = """
+    The resolution invariant height of the cloud top. This is calculated with
+    tuned parameters "cloud_top_threshold," "cloud_top_k," "cloud_top_a."
+    "cloud_top_k" should be chosen such that
+    w(z) = 1 / (1 + exp(-k * (q_cond - q_threshold) ) is close to 0 for
+    values q_c << "cloud_top_threshold" and close to 1 for
+    q_c ~ "cloud_top_threshold."
```
tapios:
```suggestion
Cloud-top height derived from a GPU-friendly weighted average, avoiding index searches. A soft cloud mask (sigmoid function) and an exponential weighting are used to identify the top of the cloud layer.
```

src/diagnostics/core_diagnostics.jl
```diff
@@ -1114,6 +1114,64 @@ add_diagnostic_variable!(
...
 )
 
+###
+# Cloud top height
+###
+compute_cloud_top_height!(out, state, cache, time) =
+    compute_cloud_top_height!(out, state, cache, time, cache.atmos.moisture_model)
+compute_cloud_top_height!(_, _, _, _, model::T) where {T} =
+    error_diagnostic_variable("cltz", model)
+
```
tapios, line 1124: Add docstring, e.g.:

"""
    compute_cloud_top_height!(out, state, cache, time, moisture_model)

Calculates a resolution-insensitive, GPU-friendly cloud-top height.

This function computes a weighted average of height over each column, avoiding
explicit index searches. A sigmoid function provides a "soft" cloud mask based
on the total condensate density (liquid + ice), and an exponential term weights
the average toward the highest altitudes within that mask.

The physical parameters controlling this calculation (threshold `thresh`, steepness `k`,
and exponential factor `a`) are specified in the `ct_constants` struct, which is
retrieved from `cache.params`.

# Arguments
- `out`: Output `Field` to store the computed cloud-top height. Can be `nothing`,
  in which case a new `Field` is allocated.
- `state`: The prognostic state `Vector`.
- `cache`: The cache containing parameters and precomputed quantities.
- `time`: The current simulation time.
- `moisture_model`: The moisture model struct, used for dispatch.

# Returns
- The `out` `Field` containing the calculated cloud-top height in meters.
"""

src/diagnostics/core_diagnostics.jl
```diff
@@ -1114,6 +1114,64 @@ add_diagnostic_variable!(
...
+function compute_cloud_top_height!(
+    out,
+    state,
+    cache,
+    time,
+    moisture_model::T,
+) where {T <: Union{EquilMoistModel, NonEquilMoistModel}}
+
+    clw = @. lazy(state.c.ρ * cache.precomputed.cloud_diagnostics_tuple.q_liq)
+    cli = @. lazy(state.c.ρ * cache.precomputed.cloud_diagnostics_tuple.q_ice)
```
tapios:
```suggestion
    q_liq = cache.precomputed.cloud_diagnostics_tuple.q_liq
    q_ice = cache.precomputed.cloud_diagnostics_tuple.q_ice
```
(Avoid one lazy multiplication.)

src/diagnostics/core_diagnostics.jl
```diff
@@ -1114,6 +1114,64 @@ add_diagnostic_variable!(
...
+    z = Fields.coordinate_field(state.c).z
+
+    ct_constants = CAP.microphysics_cloud_params(cache.params).ct_constants
+
+    q_cond = @. lazy(clw + cli)
```
tapios:
```suggestion
    # Condensate density (kg/m^3)
    q_cond = @. lazy(state.c.ρ * (q_liq + q_ice))
```

src/diagnostics/core_diagnostics.jl
```diff
@@ -1114,6 +1114,64 @@ add_diagnostic_variable!(
...
+    w = @. lazy(1 / (1 + exp(-ct_constants.k * (q_cond - ct_constants.thresh))))
```
tapios:
```suggestion
    # 1. Create the "cloudiness" mask using the sigmoid function
    w = @. lazy(1 / (1 + exp(-ct_constants.k * (q_cond - ct_constants.thresh))))
```

src/diagnostics/core_diagnostics.jl
```diff
@@ -1114,6 +1114,64 @@ add_diagnostic_variable!(
...
+
+    numerator = @. lazy(w * exp(ct_constants.a * z) * z)
+    denominator = @. lazy(w * exp(ct_constants.a * z))
```
tapios:
```suggestion
    # 2. Create a numerically stabilized exponential weight to favor higher altitudes
    az = @. lazy(ct_constants.a * z)
    max_az = maximum(az) # This prevents overflow in the exp() call
    exp_az_stabilized = @. lazy(exp(az - max_az))
    
    # 3. Combine weights into a single common term 
    common_weight = @. lazy(w * exp_az_stabilized)
    
    # Define numerator and denominator for the weighted average 
    numerator = @. lazy(common_weight * z)
    # The denominator is just the common_weight itself
    denominator = common_weight
```

src/diagnostics/core_diagnostics.jl
```diff
@@ -1114,6 +1114,64 @@ add_diagnostic_variable!(
...
+
```
tapios, line 1144:
```suggestion
# Perform column integration
```

src/diagnostics/core_diagnostics.jl
```diff
@@ -1114,6 +1114,64 @@ add_diagnostic_variable!(
...
+    num = zeros(axes(Fields.level(state.f, half)))
+    denom = zeros(axes(Fields.level(state.f, half)))
+
+    Operators.column_integral_definite!(num, numerator)
+    Operators.column_integral_definite!(denom, denominator)
+
```
tapios, line 1169:
```suggestion
    # Handle the no-cloud case to prevent division by zero
    is_cloudy = Fields.level(denom, 1) .> eps(eltype(denom))
    result = @. ifelse(is_cloudy, num / denom, 0.0)
```

src/diagnostics/core_diagnostics.jl
```diff
@@ -1114,6 +1114,64 @@ add_diagnostic_variable!(
...
+    if isnothing(out)
+        out = num ./ denom
```
tapios:
```suggestion
        out = result
```

src/diagnostics/core_diagnostics.jl
```diff
@@ -1114,6 +1114,64 @@ add_diagnostic_variable!(
...
+    else
+        out .= num ./ denom
```
tapios:
```suggestion
        out .= result
```

src/diagnostics/core_diagnostics.jl
oalcabes: > Add docstring, e.g.:
> 
> """ compute_cloud_top_height!(out, state, cache, time, moisture_model)
> 
> Calculates a resolution-insensitive, GPU-friendly cloud-top height.
> 
> This function computes a weighted average of height over each column, avoiding explicit index searches. A sigmoid function provides a "soft" cloud mask based on the total condensate density (liquid + ice), and an exponential term weights the average toward the highest altitudes within that mask.
> 
> The physical parameters controlling this calculation (threshold `thresh`, steepness `k`, and exponential factor `a`) are specified in the `ct_constants` struct, which is retrieved from `cache.params`.
> # Arguments
> 
>     * `out`: Output `Field` to store the computed cloud-top height. Can be `nothing`,
>       in which case a new `Field` is allocated.
> 
>     * `state`: The prognostic state `Vector`.
> 
>     * `cache`: The cache containing parameters and precomputed quantities.
> 
>     * `time`: The current simulation time.
> 
>     * `moisture_model`: The moisture model struct, used for dispatch.
> 
> 
> # Returns
> 
>     * The `out` `Field` containing the calculated cloud-top height in meters.
>       """

None of the rest of these diagnostic functions have docstrings, instead there is the comments section we edited above. I'll make sure most of this information is in the comments.

tapios: It's a good idea, then, to start adding docstrings to the functions, especially those like this one, which are not entirely straightforward to understand.

src/diagnostics/core_diagnostics.jl
oalcabes: @tapios why do you want this to be per volume? (Doing it this way would require changing the threshold and k constants to also be per volume, which I wasn't initially doing)

tapios: It's the same you had before, isn't it? There were two multiplications by density in l. 1133/1134 before, and I replaced by the one multiplication here. I was wondering why it was there but thought you had reasons for it. It does make more sense to me to have the thresholds be on specific humidity (i.e., by mass).

oalcabes: Oops you're right! I'll just remove it.

src/diagnostics/core_diagnostics.jl
```diff
@@ -1114,6 +1114,87 @@ add_diagnostic_variable!(
...
+) where {T <: Union{EquilMoistModel, NonEquilMoistModel}}
+
+    q_liq = cache.precomputed.cloud_diagnostics_tuple.q_liq
+    q_ice = cache.precomputed.cloud_diagnostics_tuple.q_ice
+    z = Fields.coordinate_field(state.c).z
+
+    ct_constants = CAP.microphysics_cloud_params(cache.params).ct_constants
+
+    # Condensate density (kg/m^3)
+    q_cond = @. lazy(state.c.ρ * (q_liq + q_ice))
```
sajjadazimi, line 1140: It's better to show densities (per volume) with something else maybe `L` or `ρq`

src/diagnostics/core_diagnostics.jl
```diff
@@ -1114,6 +1114,87 @@ add_diagnostic_variable!(
...
+
+    # 1. Create the "cloudiness" mask using the sigmoid function
+    w = @. lazy(1 / (1 + exp(-(state.c.ρ * ct_constants.k) * (q_cond - (state.state.c.ρ * ct_constants.thresh)))))
+
+    # 2. Create a numerically stabilized exponential weight to favor higher altitudes
+    az = @. lazy(ct_constants.a * z)
+    max_az = maximum(az) # This prevents overflow in the exp() call
+    exp_az_stabilized = @. lazy(exp(az - max_az))
```
sajjadazimi, line 1148: The parameter `a` is dimensional (1/m) unless the variable `exp_az_stabilized` is given by `exp(a(z/zmax - 1))`

src/diagnostics/core_diagnostics.jl
```diff
@@ -1114,6 +1114,87 @@ add_diagnostic_variable!(
...
+    q_ice = cache.precomputed.cloud_diagnostics_tuple.q_ice
+    z = Fields.coordinate_field(state.c).z
+
+    ct_constants = CAP.microphysics_cloud_params(cache.params).ct_constants
+
+    # Condensate density (kg/m^3)
+    q_cond = @. lazy(state.c.ρ * (q_liq + q_ice))
+
+    # 1. Create the "cloudiness" mask using the sigmoid function
+    w = @. lazy(1 / (1 + exp(-(state.c.ρ * ct_constants.k) * (q_cond - (state.state.c.ρ * ct_constants.thresh)))))
```
sajjadazimi, line 1143: You can also try multiplying the mask by condensate mass signal itself to get rid of noise.

sajjadazimi: Also, here k is dimensional (`m^6`) too.

trontrytel:
Would be great to add it in! @oalcabes - what was blocking it last time and do you need help with it?

PR #4025, opened 2025-09-18
Sa/fix negative clw

sajjadazimi:
## Purpose 
Resolve negative cloud mass by:
- unwinding sgs mass flux for tracers
- using flux form updraft velocity advection of tracers
- setting minimum detr equal to minimum entr
- ensuring updraft sedimentation is done consistently with the gs sedimentation

With these fixes, we almost don't get any negative tracers. An exception is the cell at the bottom boundary where applying the BC that updates updraft area fraction violates tracer mass conservation.

----

### DIFF
--- a/src/prognostic_equations/advection.jl
+++ b/src/prognostic_equations/advection.jl
@@ -250,11 +250,14 @@ NVTX.@annotate function explicit_vertical_advection_tendency!(Yₜ, Y, p, t)
     ᶜρ = Y.c.ρ
 
     # Full vertical advection of passive tracers (like liq, rai, etc) ...
-    foreach_gs_tracer(Yₜ, Y) do ᶜρχₜ, ᶜρχ, ρχ_name
-        if !(ρχ_name in (@name(ρe_tot), @name(ρq_tot)))
-            ᶜχ = @. lazy(specific(ᶜρχ, Y.c.ρ))
-            vtt = vertical_transport(ᶜρ, ᶠu³, ᶜχ, float(dt), tracer_upwinding)
-            @. ᶜρχₜ += vtt
+    # If sgs_mass_flux is true, the advection term is computed from the sum of SGS fluxes
+    if p.atmos.edmfx_model.sgs_mass_flux isa Val{false}
+        foreach_gs_tracer(Yₜ, Y) do ᶜρχₜ, ᶜρχ, ρχ_name
+            if !(ρχ_name in (@name(ρe_tot), @name(ρq_tot)))
+                ᶜχ = @. lazy(specific(ᶜρχ, Y.c.ρ))
+                vtt = vertical_transport(ᶜρ, ᶠu³, ᶜχ, float(dt), tracer_upwinding)
+                @. ᶜρχₜ += vtt
+            end
         end
     end
     # ... and upwinding correction of energy and total water.
@@ -349,7 +352,7 @@ function edmfx_sgs_vertical_advection_tendency!(
     (; params) = p
     n = n_prognostic_mass_flux_subdomains(turbconv_model)
     (; dt) = p
-    (; edmfx_upwinding) = p.atmos.numerics
+    (; edmfx_upwinding, tracer_upwinding) = p.atmos.numerics
     (; ᶠu³ʲs, ᶠKᵥʲs, ᶜρʲs) = p.precomputed
     (; ᶠgradᵥ_ᶜΦ) = p.core
 
@@ -358,6 +361,8 @@ function edmfx_sgs_vertical_advection_tendency!(
     ᶠz = Fields.coordinate_field(Y.f).z
     ᶜu₃ʲ = p.scratch.ᶜtemp_C3
     ᶜKᵥʲ = p.scratch.ᶜtemp_scalar_2
+    ᶠJ = Fields.local_geometry_field(axes(Y.f)).J
+
     for j in 1:n
         # TODO: Add a biased GradientF2F operator in ClimaCore
         @. ᶜu₃ʲ = ᶜinterp(Y.f.sgsʲs.:($$j).u₃)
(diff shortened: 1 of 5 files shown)


src/prognostic_equations/edmfx_sgs_flux.jl
```diff
@@ -145,19 +145,17 @@ function edmfx_sgs_mass_flux_tendency!(
                     MatrixFields.has_field(Y, ρχ_name) || continue
 
                     ᶜχʲ = MatrixFields.get_field(Y, χʲ_name)
-                    ᶜρχ = MatrixFields.get_field(Y, ρχ_name)
-                    ᶜχ = (@. lazy(specific(ᶜρχ, Y.c.ρ)))
 
-                    @. ᶠu³_diff = ᶠu³ʲs.:($$j) - ᶠu³
+                    @. ᶠu³_diff = ᶠu³ʲs.:($$j)
```
szy21: Could you change this for diagnostic EDMF as well?

src/diagnostics/core_diagnostics.jl
```diff
@@ -419,15 +419,28 @@ function compute_clw!(
...
 end
+function compute_clw!(
+    out,
+    state,
+    cache,
+    time,
+    moisture_model::NonEquilMoistModel,
+)
+    if isnothing(out)
+        return state.c.ρq_liq ./ state.c.ρ
```
trontrytel: I think the cloud diagnostics tuple should be computed based on the state if running with non-equilibrium model. There are also different options about how to compute it - grid mean, quadrature, etc. It may be good to keep the output consistent with that choice.

Or if we want add a second output - something like raw_clw that would always be based on state and only work for non-eq. 

But if we do that, could you also add it for ice?

toml/prognostic_edmfx_1M.toml
```diff
@@ -23,7 +23,7 @@ value = 0.1
 value = 0
 
 [detr_inv_tau]
-value = 0
+value = 0.0001
```
trontrytel, line 26: Would be good to encode somewhere the knowledge that this should be equal to `entr_inv_tau`.

src/prognostic_equations/advection.jl
```diff
@@ -250,11 +250,13 @@ NVTX.@annotate function explicit_vertical_advection_tendency!(Yₜ, Y, p, t)
     ᶜρ = Y.c.ρ
 
     # Full vertical advection of passive tracers (like liq, rai, etc) ...
-    foreach_gs_tracer(Yₜ, Y) do ᶜρχₜ, ᶜρχ, ρχ_name
-        if !(ρχ_name in (@name(ρe_tot), @name(ρq_tot)))
-            ᶜχ = @. lazy(specific(ᶜρχ, Y.c.ρ))
-            vtt = vertical_transport(ᶜρ, ᶠu³, ᶜχ, float(dt), tracer_upwinding)
-            @. ᶜρχₜ += vtt
+    if p.atmos.edmfx_model.sgs_mass_flux isa Val{false}
```
trontrytel, line 254: Could you add a comment here that says that when sgs model is used the vertical advection is done for both at the same time?

src/prognostic_equations/advection.jl
```diff
@@ -424,7 +428,7 @@ function edmfx_sgs_vertical_advection_tendency!(
                     FT(1),
                     Y.c.sgsʲs.:($$j).ρa,
                     FT(0),
-                    Y.c.ρ,
+                    ᶜρʲs.:($$j),
```
trontrytel, line 432: Is this correct? I thought we should tend to grid mean density if area is small?

src/prognostic_equations/advection.jl
```diff
@@ -446,24 +450,24 @@ function edmfx_sgs_vertical_advection_tendency!(
                 ᶜqʲ = MatrixFields.get_field(Y, qʲ_name)
                 ᶜqʲₜ = MatrixFields.get_field(Yₜ, qʲ_name)
                 ᶜwʲ = MatrixFields.get_field(p.precomputed, wʲ_name)
-                ᶠw³ʲ = (@. lazy(CT3(ᶠinterp(Geometry.WVector(-1 * ᶜwʲ)))))
                 ᶜaqʲ = (@. lazy(ᶜa * ᶜqʲ))
 
-                # Advective form advection of tracers with grid-mean velocity
-                va = vertical_advection(
+                # Flux form advection of tracers with grid-mean velocity
```
trontrytel: and updrafts? (for the comment)

src/prognostic_equations/advection.jl
```diff
@@ -527,24 +530,24 @@ function edmfx_sgs_vertical_advection_tendency!(
                 ᶜχʲ = MatrixFields.get_field(Y, χʲ_name)
                 ᶜχʲₜ = MatrixFields.get_field(Yₜ, χʲ_name)
                 ᶜwʲ = MatrixFields.get_field(p.precomputed, wʲ_name)
-                ᶠw³ʲ = (@. lazy(CT3(ᶠinterp(Geometry.WVector(-1 * ᶜwʲ)))))
                 ᶜaχʲ = (@. lazy(ᶜa * ᶜχʲ))
 
-                # Advective form advection of tracers with grid-mean velocity
-                va = vertical_advection(
+                # Flux form advection of tracers with grid-mean velocity
```
trontrytel: Same as above

src/diagnostics/core_diagnostics.jl
sajjadazimi: The tuple is computed based on the sate, however at some point it uses TD functions for the env that clip negative values. So the output can be different if env values are negative. After fixes in this PR clw values are positive so I think it is safe to change this function back to what it was for consistency.

src/prognostic_equations/advection.jl
sajjadazimi: When area is small we tend to zero (not gird mean density). The density is only used for computing area fraction.

trontrytel (approved):

PR #4029, opened 2025-09-23
Fix updraft values for vanishing ar

sajjadazimi:
## Purpose 
Return constant entr and detr values instead of zero for vanishing `ar` conditions:
This allows updraft variables follow GS variables where `ar` is negligible. For small `ar` entrainment coefficient will not be zero that results in a tendency that brings updraft tracer and energy values close to the grid mean values. This is important to prevent the spread of errors from the top boundary downward into the domain where non-trivial values reside (see figure). With this fix now single column setups with 1M and 2M are numerically stable (tested for several days).

## To-do
Resolve `q_tot` fluctuations related to fast varying `ar`. This is possibly due to not using any upwinding scheme is in sgs mass flux computations.

----

### DIFF
--- a/src/prognostic_equations/edmfx_entr_detr.jl
+++ b/src/prognostic_equations/edmfx_entr_detr.jl
@@ -139,9 +139,11 @@ function entrainment(
     ::PiGroupsEntrainment,
 )
     FT = eltype(thermo_params)
-    # Entrainment is not well-defined or should be zero if updraft area is negligible.
+    entr_inv_tau = CAP.entr_inv_tau(turbconv_params)
+    # Entrainment is not well-defined if updraft area is negligible.
+    # Fix at entr_inv_tau to ensure some mixing with the environment.
     if ᶜaʲ <= eps(FT)
-        return 0
+        return entr_inv_tau
     end
 
     elev_above_sfc = ᶜz - z_sfc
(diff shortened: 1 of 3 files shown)


szy21:
We added this filter before because when a is negative, a positive entrainment will make it more negative (at some point the threshold was changed from zero to eps(FT) but I don't remember why). Will we have the problem if we remove this filter?

costachris:
Another issue we've had with prognostic EDMF is "zombie" updrafts that stick around after period of deep convection (with little to no updraft velocity, etc). Having some kind of persistent detrainment detr_inv_tau  slightly larger entr_inv_tau or minimum detrainment helped with that.

src/prognostic_equations/edmfx_entr_detr.jl
```diff
@@ -466,11 +449,6 @@ function detrainment(
     max_area_limiter_power = CAP.max_area_limiter_power(turbconv_params)
     a_max = CAP.max_area(turbconv_params)
 
-    # If ᶜρaʲ (updraft effective density) is zero or negligible, detrainment is zero.
-    if ᶜρaʲ <= eps(FT) # Consistent check
```
costachris, line 470: Do we want to at least keep a ᶜρaʲ < 0, return 0 in the entrainment functions (since a< 0 is already unphysical and entrainment when a < 0 will make area even more negative)

sajjadazimi:
When I change the condition to return zero for only negative `a` (and not `a < eps(FT)`) one of the implicit jobs in the buildkite that fails. I guess it's because of the division by `rhoa` in the detrainment computations. I think it's better to keep the filter but return the fixed values `entr_inv_tau` and `detr_inv_tau` instead of zero for small area fractions (`a < eps(FT)` or `\rho a < eps(FT)`)

PR #4037, opened 2025-10-01
Implementing double-sided limiters

trontrytel:
Triangle limiter flips the sign of the tendency if we are limiting by negative numbers. The tendency in that case becomes unbounded. It is possible in our simulations to have spurious negative tracers.

This PR changes the triangle limiter behavior. If limiting by negative number the source and destination categories are swapped. We try to limit this new force by the amount available in the old destination category. If both source and destination are negative, we return zero, as there is nothing the limiter can do.

If no second limiter is provided for negative tracer case, the tendency is set to zero. This is used for example when limiting the number concentration tendencies.

### DIFF
--- a/perf/flame.jl
+++ b/perf/flame.jl
@@ -40,7 +40,7 @@ ProfileCanvas.html_file(joinpath(output_dir, "flame.html"), results)
 allocs_limit = Dict()
 allocs_limit["flame_baroclinic_wave_moist_gpu"] = 1_243_048
 allocs_limit["flame_default"] = 99_928
-allocs_limit["flame_default_1m"] = 527_712
+allocs_limit["flame_default_1m"] = 596_096
 allocs_limit["flame_diagnostics"] = 10_677_144
 allocs_limit["flame_aquaplanet_diagedmf"] = 11_644_128
 allocs_limit["flame_aquaplanet_progedmf"] = 774_712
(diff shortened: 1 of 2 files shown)


trontrytel:
Testing in CI here and also with high resolution MBW

src/prognostic_equations/limited_tendencies.jl
```diff
@@ -31,35 +31,61 @@ NVTX.@annotate function limiters_func!(Y, p, t, ref_Y)
...
+        # This results in the tendecy switching signs. In that case we need to try to limit by the
+        # amound available in the destination category.
+        force1 = force + allowed_source_amount - sqrt(force^2 + allowed_source_amount^2)
+        force2 = -force1 + limit_neg - sqrt(force1^2 + limit_neg^2)
+        return -force2
+    else
+        # If both the source and destination are negatve, we set the force to zero
+        # (There is nothing the limiter can do in this case.)
+        return FT(0)
+    end
```
sajjadazimi, line 93: When force is negative we might get a rate larger than limit_neg from the first condition. I think we should use max(0, force). So when force is negative then the rate is computed based on allowed_source_amount (if allowed_source_amount is positive return 0, if negative return something).

sajjadazimi (approved):

trontrytel:
There are a couple of simulations that are flagged for having different results by our reproducibility tests. To my eye the changes are insignificant. One flame graph shows that with this PR we are increasing allocations by 2% for 1-moment microphysics job. 

I vote for merging it as is. I will just update the documentation to explain the more complicated formulation of the limiter.

PR #4038, opened 2025-10-02
add sensible and latent heat flux diagnostics

szy21:
## Purpose 

I also think we can just remove `hfes`. What do others think?

## Content

----


nefrathenrici (approved):
Thank you! I think a few test cases need to save the hfss diagnostic .

costachris (approved):
LGTM

costachris:
I'm okay with removing hfes now that we have LHF/SHF.

PR #4044, opened 2025-10-03
Fix density oscillations

sajjadazimi:
## Purpose 
Fix density fluctuations by using downwind density info in updraft buoyancy eq

Here are `\rho^j - \rho` plots after 1hr simulation of the single column trmm with prognostic EDMF when updating the updraft velocity by using density:
- upwind of the flow,
- interpolating the two sides,
- downwind of the flow,
respectively:

Screenshot 2025-10-03 at 3 15 45 PM

Screenshot 2025-10-03 at 3 12 44 PM

Screenshot 2025-10-03 at 3 14 20 PM

### DIFF
--- a/src/initial_conditions/initial_conditions.jl
+++ b/src/initial_conditions/initial_conditions.jl
@@ -66,7 +66,7 @@ function column_indefinite_integral(
     f::Function,
     ϕ₀::FT,
     zspan::Tuple{FT, FT};
-    nelems = 100, # sets resolution for integration
+    nelems = 1000, # sets resolution for integration
 ) where {FT <: Real}
     # --- Make a space for integration:
     z_domain = Domains.IntervalDomain(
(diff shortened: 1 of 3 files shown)


sajjadazimi:
## Intuitive explanation of perturbation behavior (upwind vs downwind)

Consider an upward flow with a positive density perturbation at a cell center (ρ′ > 0 at i).

### Upwind scheme

* Velocity at the face is updated using the upstream (upwind) density.
* A denser fluid leads to a reduction in velocity (u′ < 0).
* Reduced outflow from the perturbed cell causes the density perturbation to grow.
* **Result:** perturbations amplify → unstable.

### Downwind scheme

* Velocity at the face is updated using the downstream (downwind) density.
* The negative velocity perturbation reduces inflow into the perturbed cell.
* The excess density gradually decays.
* **Result:** perturbations damp → stable.

PR #4055, opened 2025-10-14
Update `benchmark_step.jl` for CUDA benchmarking with useful kernel names

petebachant:
This uses https://github.com/CliMA/ClimaCore.jl/pull/2376 to provide more useful CUDA kernel names in benchmarks.

## TODO

- [x] Can we format the kernel names with anything other than underscores as separators? --> No, all non-alphanumeric characters get converted to underscores.
- [x] Do we want to do this on the benchmarks that use `benchmark.jl`, not `benchmark_step.jl`? --> we could, but probably not useful for CI yet.
- [x] Update Buildkite pipeline to use this feature
- [x] Switch to non-dev ClimaCore

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -20,7 +20,7 @@ steps:
   - label: "init :computer:"
     key: "init_cpu_env"
     concurrency: 1
-    concurrency_group: 'depot/climaatmos-ci'
+    concurrency_group: "depot/climaatmos-ci"
     command:
       - "echo $$JULIA_DEPOT_PATH"
 
@@ -41,13 +41,11 @@ steps:
 
   - group: "Reproducibility infrastructure"
     steps:
-
       - label: ":computer: Test reproducibility infrastructure"
         command: "julia --color=yes --project=.buildkite test/unit_reproducibility_infra.jl"
 
   - group: "Radiation"
     steps:
-
       - label: ":computer: single column radiative equilibrium gray"
         command: >
           julia --color=yes --project=.buildkite .buildkite/ci_driver.jl
@@ -98,7 +96,6 @@ steps:
 
   - group: "Gravity wave"
     steps:
-
       - label: ":computer: non-orographic gravity wave parameterization unit test 3d"
         command: "julia --color=yes --project=.buildkite test/parameterized_tendencies/gravity_wave/non_orographic_gravity_wave/nogw_test_3d.jl"
         artifact_paths: "nonorographic_gravity_wave_test_3d/*"
@@ -132,7 +129,6 @@ steps:
 
   - group: "Column Examples"
     steps:
-
       - label: ":computer: single column hydrostatic balance float64"
         command: >
           julia --color=yes --project=.buildkite .buildkite/ci_driver.jl
@@ -149,7 +145,6 @@ steps:
 
   - group: "Box Examples"
     steps:
-
       - label: ":computer: Box hydrostatic balance"
         command: >
           julia --color=yes --project=.buildkite .buildkite/ci_driver.jl
@@ -188,7 +183,6 @@ steps:
 
   - group: "Plane Examples"
     steps:
-
       - label: ":computer: Density current experiment"
         command: >
           julia --color=yes --project=.buildkite .buildkite/ci_driver.jl
@@ -287,7 +281,6 @@ steps:
 
   - group: "Conservation check"
     steps:
-
       - label: ":computer: baroclinic wave check conservation"
         command: >
           julia --color=yes --project=.buildkite .buildkite/ci_driver.jl --config_file $CONFIG_PATH/baroclinic_wave_conservation.yml
@@ -324,7 +317,6 @@ steps:
 
   - group: "Sphere Examples (Dycore)"
     steps:
-
       - label: ":computer: hydrostatic balance float64"
         command: >
           julia --color=yes --project=.buildkite .buildkite/ci_driver.jl
@@ -394,7 +386,6 @@ steps:
 
   - group: "Sphere Examples (Aquaplanet)"
     steps:
-
       - label: ":umbrella: aquaplanet nonequil allsky monin_obukhov varying insol gravity wave (gfdl_restart) high top 1-moment"
         command: >
           julia --color=yes --project=.buildkite .buildkite/ci_driver.jl
@@ -462,7 +453,6 @@ steps:
 
   - group: "Sphere Examples (Topography)"
     steps:
-
       - label: ":computer: baroclinic wave topography (dcmip)"
         command: >
           julia --color=yes --project=.buildkite .buildkite/ci_driver.jl
@@ -489,7 +479,6 @@ steps:
 
   - group: "Restarting"
     steps:
-
       - label: ":computer: test restart"
         command: >
           julia --color=yes --project=.buildkite test/restart.jl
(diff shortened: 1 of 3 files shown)


perf/benchmark_step.jl
```diff
@@ -30,13 +30,64 @@ Y₀ = deepcopy(integrator.u);
...
+        catch
+            @warn "Unrecognized boolean env var value; using default" var = var val = raw default =
+                default
+            return default
+        end
+    end
+end
+
+# If we're running on CUDA, use CUDA's profiler
+if ENV["CLIMACOMMS_DEVICE"] == "CUDA" && device isa ClimaComms.CUDADevice
```
petebachant: @imreddyTeja any idea why this would fail here on CPU with:

```
ERROR: LoadError: UndefVarError: `CUDA` not defined in `Main`
```

.buildkite/pipeline.yml
```diff
@@ -41,13 +41,11 @@ steps:
 
   - group: "Reproducibility infrastructure"
     steps:
-
```
petebachant, line 44: These changes were made by a YAML auto-formatter in VS Code. Is there a style guide I might be breaking here?

perf/benchmark_step.jl
imreddyTeja: Did you see this while using the `.buildkite` project?

petebachant: It's happening in the [Buildkite pipeline](https://buildkite.com/clima/climaatmos-ci/builds/26389/steps/canvas?sid=0199e41a-e6cf-4c47-a35d-4752e93e05cf), which should be using that project.

imreddyTeja: I think this is because `CUDA.@profile` is a macro, which gets expanded when the if-else statement is parsed. 

Moving `import CUDA` outside of the conditional should fix it, but that will always import CUDA, regardless if running on CPU or GPU. 

Another option would be to wrap `CA.benchmark_step!(integrator, Y₀, n_steps)`  in a function and explicitly call `CUDA.profile_internally` or `CUDA.profile_externally`

.buildkite/pipeline.yml
imreddyTeja: I'm not sure... this is something I have been wondering as well. I considered following [this](https://github.com/buildkite/agent/blob/main/.buildkite/pipeline.yml) example, which is used in Buildkite's docs.

perf/benchmark_step.jl
```diff
@@ -30,13 +30,64 @@ Y₀ = deepcopy(integrator.u);
...
+        @info "Using external CUDA profiler"
+        CUDA.@profile external = true begin
+            e = CUDA.@elapsed begin
+                CA.benchmark_step!(integrator, Y₀, n_steps)
+            end
+        end
+    else
+        @info "Using internal CUDA profiler"
+        CUDA.@profile external = false begin
+            e = CUDA.@elapsed begin
```
imreddyTeja, line 52: What does the `e = CUDA.@elapsed...` inside the ` CUDA.@profile` do?

petebachant: The intent is to return the elapsed time of the call for the info statement below. I still need to figure out how to get the tabular output for the kernels that took the longest time. Let me know if you know how to do that. It looks like the `ClimaAtmos.benchmark_step` function doesn't return the same data that `ClimaTimeSteppers.benchmark_step` does. The latter is used in `benchmark.jl` and the former in `benchmark_step.jl`, though both scripts are calling their respective `benchmark_step` method 🤔

imreddyTeja: I think the tabular output you are referring to only works with the internal profiler. Also, with 
```julia
p = CUDA.@profile external=false $any_expression$
```
`p` will be the profiling results as `CUDA.ProfileResults` type. With the external profiler:

```julia
e = CUDA.@profile external=true $any_expression$
```
e is whatever `any_expression` returns.

perf/benchmark_step.jl
```diff
@@ -30,13 +30,64 @@ Y₀ = deepcopy(integrator.u);
...
+    if use_external_profiler
+        @info "Using external CUDA profiler"
+        CUDA.@profile external = true begin
+            e = CUDA.@elapsed begin
+                CA.benchmark_step!(integrator, Y₀, n_steps)
+            end
+        end
+    else
+        @info "Using internal CUDA profiler"
+        CUDA.@profile external = false begin
```
imreddyTeja: The profiler output needs to be printed for the tabular output to show:

```julia

### shows tabular output because the profiled results are the returned value
julia> if true
       CUDA.@profile external=false CUDA.@elapsed some_cuarray .+= 2
       end

Profiler ran for 219.11 µs, captu....
#### no results shown because the info statement's return is the return value
julia> if true
       CUDA.@profile external=false CUDA.@elapsed some_cuarray .+= 2
       @info "some info"
       end
[ Info: some info

### profiler results explicitly shown
julia> if true
       p = CUDA.@profile external=false CUDA.@elapsed some_cuarray .+= 2
       println(p) # or @show p
       @info "some info"
       end
Profiler ran for 178.1 µs, capturing 78 events.

Host-side activity: calling CUDA APIs took 95.61 µs (53.68% of the trace)
┌──────────┬────────────┬───────┬───────────────────────────────────────┬────────────────────┐
│ Time (%) │ Total time │ Calls │ Time distribution                     │ Name               │
...
```

imreddyTeja:
> @dennisYatunin @imreddyTeja thoughts on the timeouts here? Should I increase the limit or disable kernel renaming?

What is the advantage of using kernel renaming in climaatmos-ci for buildkite steps that don't profile? Is the idea to run the profiler at the end of each simulation? 

Comparing the buildkite for this PR to the main's last buildkite run shows a ~40% slowdown, but everything after the first step doesn't seem to be affected. I'm not sure if that cost is worth it at the moment.

.buildkite/pipeline.yml
```diff
@@ -1233,7 +1214,6 @@ steps:
 
   - group: "Benchmarks"
     steps:
-
       - label: ":computer: Benchmark: CPU baroclinic wave moist"
```
imreddyTeja, line 1217: I think this label is incorrect

perf/benchmark_step.jl
```diff
@@ -14,10 +14,12 @@ include(joinpath("perf", "benchmark_step.jl"));
 redirect_stderr(IOContext(stderr, :stacktrace_types_limited => Ref(false)))
 import ClimaComms
 ClimaComms.@import_required_backends
+import ClimaCore
```
imreddyTeja: Is this import used?

imreddyTeja:
> Alright, so now that [CliMA/ClimaCore.jl#2376](https://github.com/CliMA/ClimaCore.jl/pull/2376) is merged, do I need to make a ClimaCore release, or should I simply update the `.buildkite` project to refer to a specific ClimaCore Git commit?
> 
> I suppose I could revert any Buildkite changes here and just keep the updates to the benchmarking script.

I think we should make a ClimaCore release. It would be nice if we updated the prettytables compat before releasing.
After the release and updating the manifest, I think this PR is good to go.

perf/benchmark_step.jl
petebachant: Good catch. It's no longer used. `ClimaComms` was also imported twice. We really need to figure out how to get linting working!

imreddyTeja (approved):

PR #4058, opened 2025-10-15
Anisotropic Minimum Dissipation method (LES)

akshaysridhar:
Alternative parameterization for subfilter scale fluxes. May provide more favourable tracer diffusivity where weak shear and strong tracer gradients are present.

### DIFF
--- a/config/default_configs/default_config.yml
+++ b/config/default_configs/default_config.yml
@@ -163,6 +163,12 @@ vert_diff:
 hyperdiff:
   help: "Hyperdiffusion. Use `CAM_SE` for sensible default values and ClimaHyperdiffusion for user control. [`CAM_SE` (default), `ClimaHyperdiffusion` (or `true`), `none` (or `false`)]"
   value: "CAM_SE"
+amd_les:
+  help: "AMD LES closure [`true`, `false` (default)]"
+  value: false
+c_amd: 
+  help: "Model coefficient for AMD-LES closure (TODO: Move to parameters.toml)"
+  value: 0.29
 smagorinsky_lilly:
   help: "Smagorinsky-Lilly diffusive closure [`false` (default), `true`]"
   value: false
--- a/src/parameterized_tendencies/les_sgs_models/anisotropic_minimum_dissipation.jl
+++ b/src/parameterized_tendencies/les_sgs_models/anisotropic_minimum_dissipation.jl
@@ -1,0 +1,364 @@
+#####
+##### AMD Diffusion
+#####
+
+import ClimaCore.Fields as Fields
+import ClimaCore.Operators as Operators
+import ClimaCore: Geometry
+import LinearAlgebra: tr
+
+"""
+    set_amd_precomputed_quantities!(Y, p)
+
+Placeholder for precomputed quantities in the Anisotropic-Minimum-Dissipation method. 
+Returns `nothing`. This function is included for simple extensions in debugging workflows.
+"""
+function set_amd_precomputed_quantities!(Y, p)
+    nothing
+end
+
+horizontal_amd_tendency!(Yₜ, Y, p, t, ::Nothing) = nothing
+vertical_amd_tendency!(Yₜ, Y, p, t, ::Nothing) = nothing
+
+"""
+    horizontal_amd_tendency!(Yₜ,Y, p, t, ::AnisotropicMinimumDissipation)
+
+Anisotropic Minimum Dissipation (AMD) Subgrid-Scale Model
+
+This module implements the Anisotropic Minimum Dissipation (AMD) subgrid-scale (SGS) model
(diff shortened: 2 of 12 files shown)


config/default_configs/default_config.yml
```diff
@@ -163,6 +163,9 @@ vert_diff:
 hyperdiff:
   help: "Hyperdiffusion. Use `CAM_SE` for sensible default values and ClimaHyperdiffusion for user control. [`CAM_SE` (default), `ClimaHyperdiffusion` (or `true`), `none` (or `false`)]"
   value: "CAM_SE"
+amd_les:
```
trontrytel, line 166: Would it make sense to switch the logic to something like

`horizontal_diffusion: "Horizontal diffusion closure choice. [none (default), SmagorinskiLilly, AMD]`

and combine it with the Smagorinski_lilly flag when setting up the model based on the config flags?

Or do you think it possible that the users will want to run with both Smag and amd?

akshaysridhar: I don't expect to use both models simultaneously other than in some debugging workflow: In this case switching the config logic as you suggest is reasonable!

src/parameterized_tendencies/les_sgs_models/anisotropic_minimum_dissipation.jl
```diff
@@ -0,0 +1,329 @@
...
+```
+
+To remove this tendency in debugging workflows, comment (or delete) the call to this function
+in `remaining_tendencies.jl`
+
+"""
+function horizontal_amd_tendency!(Yₜ, Y, p, t, ::AnisotropicMinimumDissipation)
+    (; atmos, precomputed, scratch, params) = p
+    FT = eltype(Y)
+    c_amd = FT(1/√12)
```
trontrytel: Do we want to make it a parameter in ClimaParams at some point?

src/parameterized_tendencies/les_sgs_models/anisotropic_minimum_dissipation.jl
```diff
@@ -0,0 +1,329 @@
...
+        @. ∂̂ᶜχ += @. ᶜΔ_z * Geometry.project(axis_uvw, ᶜgradᵥ(ᶠinterp(ᶜχ)))
+        ᶜD_amd = @. ᶜtemp_scalar = max(FT(0), -c_amd * 
+                                   ((ᶜ∂̂u_uvw * ∂̂ᶜχ ⊗ ∇ᶜχ).components.data.:1 +
+                                   (ᶜ∂̂u_uvw * ∂̂ᶜχ ⊗ ∇ᶜχ).components.data.:5 + 
+                                   (ᶜ∂̂u_uvw * ∂̂ᶜχ ⊗ ∇ᶜχ).components.data.:9) / 
+                                   max(eps(FT), CA.norm_sqr(gradₕ(ᶜχ))))
+        ᶜρχₜ_diffusion = @. lazy(wdivₕ(Y.c.ρ * ᶜD_amd * gradₕ(ᶜχ)))
+        @. ᶜρχₜ += ᶜρχₜ_diffusion
+        # Rain and snow does not affect the mass
+        if ρχ_name == @name(ρq_tot)
```
trontrytel, line 171: Not sure if thats what we want in the end. I'll change it in case the results are better with all contributions in.

src/parameterized_tendencies/les_sgs_models/anisotropic_minimum_dissipation.jl
```diff
@@ -0,0 +1,329 @@
...
+                               max(eps(FT), CA.norm_sqr(gradₕ(ᶜh_tot))))
+    @. Yₜ.c.ρe_tot += wdivₕ(Y.c.ρ * ᶜD_amd * gradₕ(ᶜh_tot))
+
+    # Tracer diffusion and associated mass changes
+    foreach_gs_tracer(Yₜ, Y) do ᶜρχₜ, ᶜρχ, ρχ_name
+        ᶜχ = @. lazy(specific(ᶜρχ, Y.c.ρ))
+        ∇ᶜχ = @. lazy(Geometry.project(axis_uvw, gradₕ(ᶜχ)))
+        ∂̂ᶜχ = @. ᶜtemp_UVW = Δ_h * Geometry.project(axis_uvw, gradₕ(ᶜχ)) 
+        @. ∂̂ᶜχ += @. ᶜΔ_z * Geometry.project(axis_uvw, ᶜgradᵥ(ᶠinterp(ᶜχ)))
+        ᶜD_amd = @. ᶜtemp_scalar = max(FT(0), -c_amd *
```
trontrytel: It seems to me that once we are done with the initial debugging and testing there could be much more code re-use between Smagorinski and amd options. Is it correct to say that they both are just ways to compute `D`. And everything else on how that `D` is applied in the tendencies is the same?

Lets do that when we switch the user choice logic and allow to use only one of them at a time

Issue #4046, opened 2025-10-06
GPU compilation error using 1M microphysics


nefrathenrici:
Some configurations with 1M result in a compilation error on GPU.

[Example](https://buildkite.com/clima/climacoupler-longruns/builds/948#019983fd-2a45-4022-8e29-040eb532d02c):
```
ERROR: LoadError: InvalidIRError: compiling MethodInstance for ClimaCoreCUDAExt.copyto_stencil_kernel!(::Field{…}, ::Base.Broadcast.Broadcasted{…}, ::ClimaCore.Spaces.ExtrudedFiniteDifferenceSpace{…}, ::NTuple{…}, ::ClimaCore.DataLayouts.UniversalSize{…}, ::ClimaCore.DataLayouts.NoMask, ::CartesianIndices{…}) resulted in invalid LLVM IR
Reason: unsupported dynamic function invocation (call to *)
Stacktrace:
 [1] *
   @ ./operators.jl:596
 [2] compute_Δ𝛼_linvanleer
   @ /scratch/clima/slurm-buildkite/climacoupler-longruns/948/depot/default/packages/ClimaCore/y7ZbY/src/Operators/finitedifference.jl:1498
 [3] multiple call sites
   @ unknown:0
```

trontrytel:
I spend some more time with it, and I really don't understand why `vanleer_limiter` is not working for microphysics tracers. In the interest of time I'll add a separate flag and set it for `first_order` for now. Once we decide it's really needed to make it work with `vanleer` I can spend more time on it

PR #4062, opened 2025-10-17
ADD emdfx_tracer_upwinding flag, set it to first_order

trontrytel:
`vanleer_limiter` upwinding option does not compile on GPUs for microphysics tracers in EDMF. See https://github.com/CliMA/ClimaAtmos.jl/issues/4046

This PR tries to cleanup the logic of upwinding flags and give us different configuration options untill we solve the underlying GPU compiler issue.

For the atmos runs without EDMF we will now have `energy_q_tot_upwinding` and `tracer_upwinding`. They should both be set to `vanleer_limiter` unless there is a good reason not to.

For atmos runs with EDMF ideally we would do the same. But because right now we can't I have:
- `energy_q_tot_upwinding` set to `vanleer_limiter` but `tracer_upwinding` set to `first_order`
- `edmfx_mse_q_tot_upwinding` and `edmfx_tracer_upwinding` set to `first_order`.

One "behavioral change" is that `tracer_upwinding` was affecting `q_tot` and microphysics tracers on the grid mean. But from all our discussions it seems like we want to group energy and total water together. I don't think it should make any changes in behavior because those two flags are usually set to the same value anyway.

### DIFF
--- a/config/longrun_configs/amip_target_diagedmf.yml
+++ b/config/longrun_configs/amip_target_diagedmf.yml
@@ -13,7 +13,6 @@ prescribed_aerosols: ["CB1", "CB2", "DST01", "DST02", "DST03", "DST04", "DST05",
 surface_setup: "DefaultMoninObukhov"
 turbconv: "diagnostic_edmfx"
 prognostic_tke: true
-edmfx_upwinding: "first_order"
 edmfx_entr_model: "Generalized"
 edmfx_detr_model: "Generalized"
 edmfx_nh_pressure: true
--- a/calibration/experiments/gcm_driven_scm/model_config_diagnostic.yml
+++ b/calibration/experiments/gcm_driven_scm/model_config_diagnostic.yml
@@ -6,7 +6,6 @@ turbconv: "diagnostic_edmfx"
 implicit_diffusion: true
 approximate_linear_solve_iters: 2
 prognostic_tke: true
-edmfx_upwinding: "first_order"
 rayleigh_sponge: true
 edmfx_entr_model: "PiGroups"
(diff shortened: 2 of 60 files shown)


trontrytel:
Lets see how many CI cases work with that change 🤞

config/longrun_configs/amip_target_diagedmf.yml
```diff
@@ -12,8 +12,8 @@ aerosol_radiation: true
 prescribed_aerosols: ["CB1", "CB2", "DST01", "DST02", "DST03", "DST04", "DST05", "OC1", "OC2", "SO4", "SSLT01", "SSLT02", "SSLT03", "SSLT04", "SSLT05"]
 surface_setup: "DefaultMoninObukhov"
 turbconv: "diagnostic_edmfx"
+tracer_upwinding: first_order
```
szy21: This is not used in 0M, so it's not changing anything. It might be good to remove it to have one less config, but it's fine to have it too. (same for all other 0M yamls)

trontrytel: Updated! If the Ci is still unchanged I'll merge it in. 

But could you look at the `src/prognostic_equations/implicit/manual_sparse_jacobian.jl` I'm wondering if I should now update it too. We are checking the `is_third_order = p.atmos.numerics.edmfx_mse_q_tot_upwinding == Val(:third_order)` but now we have more options combinations...

szy21: This is ok for now. We will just use first order upwinding in the Jacobian for edmf mse, qt, and tracers. We can change it if needed in the future.

PR #4069, opened 2025-10-23
Update to new release of CloudMicrophysics

trontrytel:


trontrytel:
The only case that is breaking is `prognostic_edmfx_rico_column_2M` This is not unexpected since the changes on CM.jl side made the clipping on small numbers less restrictive than before for 2M scheme. 

After discussing with @sajjadazimi, we are ok to temporarily make it a soft fail. We want to make the 1M scheme stable first. The 2M scheme will probably require more thought on how to set epsilons correctly for both mass and number concentrations.

PR #4054, opened 2025-10-14
rft: split Smagorinsky diffusive model to horizontal and vertical components

haakon-e:
This pull request extends the configurability of the Smagorinsky-Lilly turbulence model by allowing users to specify whether the model should be applied in the horizontal direction only, vertical direction only, both together, or each separately. This enhancement is reflected in configuration options, documentation, and the underlying implementation.

**Smagorinsky-Lilly model configurability:**

* Added support for new configuration options for the Smagorinsky-Lilly model: `UV` (horizontal only), `W` (vertical only), and `UV_W` (horizontal and vertical treated separately), in addition to the existing `UVW` (all directions). (`config/default_configs/default_config.yml`, `src/solver/model_getters.jl`, `src/solver/types.jl`) [[1]](diffhunk://#diff-a507a5b1b0f8375040a7694778c1ce9ad5ec72a4465b76b7ba5d2aa693dcb574L173-R173) [[2]](diffhunk://#diff-ab240b0d863b064ad1bd28f7844952c3809af57b3424ed31794a4835f6349e47R179-R187) [[3]](diffhunk://#diff-2f94c7adf07b3cbc2396d99087f28bfb189a773943c7e1994754d81ef01e1d66R249-R251)
* Updated the assertion in `get_smagorinsky_lilly_model` to allow the new configuration symbols. (`src/solver/model_getters.jl`)
* Improved documentation for the Smagorinsky-Lilly model to describe all possible configuration flags and their effects. (`src/solver/model_getters.jl`, `src/solver/types.jl`) [[1]](diffhunk://#diff-ab240b0d863b064ad1bd28f7844952c3809af57b3424ed31794a4835f6349e47R179-R187) [[2]](diffhunk://#diff-2f94c7adf07b3cbc2396d99087f28bfb189a773943c7e1994754d81ef01e1d66R249-R251)

**Implementation logic:**

* Updated logic in `set_smagorinsky_lilly_precomputed_quantities!` to correctly compute horizontal and vertical length scales (`ᶜL_h`, `ᶜL_v`) depending on the chosen model configuration. (`src/parameterized_tendencies/les_sgs_models/smagorinsky_lilly.jl`)
[...]

PR #4071, opened 2025-10-23
rft: add face density method

haakon-e:
Add `ᶠface_density(ᶜρ)`, the Jacobian-weighted face density `ᶠinterp(ᶜρ * ᶜJ) / ᶠJ` (Eq. 104 of the dycore paper), and use it in the vertical advection, sedimentation, and manual sparse Jacobian routines instead of combining cell and face Jacobians and densities at each call site. Also add `WVec` and `UV` as shorthand for `Geometry.WVector` and `Geometry.UVVector` (`UV` has no call site in this PR).

Pure refactor: the substitutions are algebraically identical, so simulations are unaffected.

About two dozen sites use `ᶠinterp(ᶜρ)` without Jacobian weighting. Whether any of them should use `ᶠface_density` instead is left for discussion; such a change would affect results and must be applied to the tendency and the Jacobian together.

### DIFF
--- a/src/prognostic_equations/water_advection.jl
+++ b/src/prognostic_equations/water_advection.jl
@@ -53,9 +53,6 @@ function vertical_advection_of_water_tendency!(Yₜ, Y, p, t)
     (; ᶜu, ᶜT) = p.precomputed
     thp = CAP.thermodynamics_params(params)
 
-    ᶜJ = Fields.local_geometry_field(Y.c).J
-    ᶠJ = Fields.local_geometry_field(Y.f).J
-
     microphysics_tracers = (
         (@name(ρq_lcl), @name(ᶜwₗ)),
         (@name(ρq_icl), @name(ᶜwᵢ)),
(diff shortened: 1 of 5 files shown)


haakon-e:

- #4071 ◀
    - #4054

src/prognostic_equations/water_advection.jl
```diff
@@ -67,33 +64,23 @@ function vertical_advection_of_water_tendency!(Yₜ, Y, p, t)
         (name == @name(ρq_icl) || name == @name(ρq_sno)) ? TD.internal_energy_ice :
         nothing
 
-    ᶠρ = p.scratch.ᶠtemp_scalar
+    ᶠρ = ᶠface_density(Y.c.ρ)
```
dennisYatunin, line 67: This is changing `ᶠρ` from a materialized field to a lazy field. This saves a kernel launch, but it also means that `ᶠρ` has the same memory overhead as 3 fields instead of 1. Check the performance of GPU tests (particularly those with high memory overhead like the full AMIP test) to make sure this doesn't deteriorate performance.

dennisYatunin (commented):
The `WVector -> WVec` and `UVVector -> UV` changes feel like they're a bit too far down the abbreviation rabbit hole, but the `ᶠρ` changes look great. There was one behavior change that I noticed, so check its performance impact if you're committed to it.

Regarding the sites that use `ᶠinterp(ᶜρ)` without Jacobian weighting, those should all match up with the dycore paper. Discrete conservation requires a duality between the density weighting of certain scalars and velocity components, so some need to be weighted while others need to be unweighted (i.e., you could swap the weights around, but you can't use the weights everywhere). Anything that doesn't match the dycore paper should probably be changed.

dennisYatunin (approved):

PR #4075, opened 2025-10-24
reduct t_end for some longruns

szy21:
## Purpose 

So that they will finish in a day

## Content

----


szy21:
Manually merging it as it only changes longrun config.

PR #4078, opened 2025-10-28
Fix dt float type instability

oalcabes:
Remove use of float(dt) and replace with FT(dt) to prevent type instability

### DIFF
--- a/post_processing/jacobian_summary.jl
+++ b/post_processing/jacobian_summary.jl
@@ -19,7 +19,7 @@ function print_jacobian_summary(integrator)
 
     FT = eltype(Y)
     γs = filter(!iszero, CA.LinearAlgebra.diag(tableau_coefficients))
-    dtγ = FT(float(dt) * γs[end])
+    dtγ = FT(dt) * FT(γs[end])
     scalar_names = CA.scalar_field_names(Y)
     block_keys = Iterators.product(scalar_names, scalar_names)
 
--- a/reproducibility_tests/ref_counter.jl
+++ b/reproducibility_tests/ref_counter.jl
@@ -1,4 +1,4 @@
-270
+271
 
 # **README**
 #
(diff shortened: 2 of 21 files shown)


src/prognostic_equations/implicit/manual_sparse_jacobian.jl
```diff
@@ -680,7 +680,7 @@ function update_jacobian!(alg::ManualSparseJacobian, cache, Y, p, dtγ, t)
             ᶜmixing_length_field .= ᶜmixing_length(Y, p)
 
             @inline tke_dissipation_rate_tendency(tke⁰, mixing_length) =
-                tke⁰ >= 0 ? c_d * sqrt(tke⁰) / mixing_length : 1 / float(dt)
+                tke⁰ >= 0 ? c_d * sqrt(tke⁰) / mixing_length : 1 / typeof(tke⁰)(dt)
```
haakon-e, line 683: I think you can use `FT` here.

haakon-e (approved):

src/prognostic_equations/implicit/manual_sparse_jacobian.jl
oalcabes: I can't because FT is not defined in the function

reproducibility_tests/ref_counter.jl
```diff
@@ -20,6 +20,9 @@
 
 
 #=
+271
+- Change float(dt) to FT(t) to improve type stability
```
haakon-e:
```suggestion
- Change float(dt) to FT(dt) to improve type stability
```

post_processing/jacobian_summary.jl
```diff
@@ -19,7 +19,7 @@ function print_jacobian_summary(integrator)
 
     FT = eltype(Y)
     γs = filter(!iszero, CA.LinearAlgebra.diag(tableau_coefficients))
-    dtγ = FT(float(dt) * γs[end])
+    dtγ = FT(FT(dt) * γs[end])
```
dennisYatunin:
```suggestion
    dtγ = FT(dt) * FT(γs[end])
```
This avoids an intermediate Float64 conversion

PR #4080, opened 2025-10-29
Implicit sgs advection and sedimentation

sajjadazimi:
## Purpose 
This PR fixes the implicit treatment of advection and sedimentation of sgs tracers. Plus other minor improvements.

----


szy21 (approved):
Thanks! At some point we can talk about for those terms we don't take into account in the Jacobian, whether we should move them out of implicit tendency. But I think it's fine for now.

PR #4089, opened 2025-11-09
remove some assert and clean up some edmf code

szy21:
## Purpose 

I took a look at the `@assert` we have in EDMF. Even though most of them are fine and removing them wouldn't give us any speedup right now, I took the opportunity to clean up some of them. Most of them are removed because we already check them in precomputed_quantities, so we don't need to check again at each timestep.

## Content

----

### DIFF
--- a/src/cache/cloud_fraction.jl
+++ b/src/cache/cloud_fraction.jl
@@ -309,8 +309,6 @@ function quad_loop(
 
     function f(x1_hat, x2_hat)
         FT = eltype(x1_hat)
-        @assert(x1_hat >= FT(0))
-        @assert(x2_hat >= FT(0))
         _ts = thermo_state(thermo_params; p = p_c, θ = x1_hat, q_tot = x2_hat)
         hc = TD.has_condensate(thermo_params, _ts)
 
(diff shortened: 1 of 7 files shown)


src/diagnostics/radiation_diagnostics.jl
```diff
@@ -45,7 +45,6 @@ function compute_rsd!(
             cache.radiation.rrtmgp_model.face_sw_flux_dn,
             axes(state.f),
         )
-        @assert !isnothing(out) "Output field 'out' must not be `nothing` in this branch"
```
szy21, line 48: I don't know why it is here, but we don't have it for any other radiation diagnostics.

src/cache/cloud_fraction.jl
```diff
@@ -309,8 +309,6 @@ function quad_loop(
 
     function f(x1_hat, x2_hat)
         FT = eltype(x1_hat)
-        @assert(x1_hat >= FT(0))
-        @assert(x2_hat >= FT(0))
```
szy21, line 313: These errors are not very useful for debugging.

tapios (approved):
LGTM

PR #4094, opened 2025-11-11
fix sedimentation detr

sajjadazimi:
## Purpose 
Modify tracer sedimentation equations and remove sedimentation detraiment function

## Content

----


trontrytel (approved):
For drizzling Sc clouds we might want to consider the inflow from environment too. Just leaving this thought here for future fixes, if needed

PR #4098, opened 2025-11-12
Improve Initialization from ERA5 and MSLP Diagnostic

costachris:
Improvements to WeatherQuest ERA5 initial conditions: 
- Adjust for ERA5/ClimaAtmos topo differences when initializing from surface pressure. 
- Improve initialization by including upstream 3D lat/lon/z interp, extrapolation beyond 1000mb surface, and high resolution custom vertical grid. These changes are needed to reduce vertical interpolation errors because upstream Clima utilities lack tri-linear interpolation to the spectral element space, resulting in what were otherwise significant interpolation errors. 
- Account for lapse rate in mslp sea level reduction (diagnostics).

The colorbars vary, but note each improvement significantly reduces the RMSE/Bias.
Current 😭
mslp_v1
Topography correction in surface pressure init 👀
mslp_v2
Topography correction + account for lapse rate in MSLP diagnostic 🥳
mslp_v3

### DIFF
--- a/src/diagnostics/core_diagnostics.jl
+++ b/src/diagnostics/core_diagnostics.jl
@@ -1647,24 +1647,32 @@ add_diagnostic_variable!(
 function compute_mslp!(out, state, cache, time)
     thermo_params = CAP.thermodynamics_params(cache.params)
     g = TD.Parameters.grav(thermo_params)
-    R_m_surf = Fields.level(
-        lazy.(TD.gas_constant_air.(thermo_params, cache.precomputed.ᶜts)),
-        1,
-    )
+    ts_level = Fields.level(cache.precomputed.ᶜts, 1)
+    R_m_surf = @. lazy(TD.gas_constant_air(thermo_params, ts_level))
 
-    # get pressure, temperature, and height at the lowest atmospheric level
     p_level = Fields.level(cache.precomputed.ᶜp, 1)
-    t_level = Fields.level(
-        lazy.(TD.air_temperature.(thermo_params, cache.precomputed.ᶜts)),
-        1,
-    )
+    t_level = @. lazy(TD.air_temperature(thermo_params, ts_level))
     z_level = Fields.level(Fields.coordinate_field(state.c.ρ).z, 1)
 
-    # compute sea level pressure using the hypsometric equation
+    # Reduce to mean sea level using hypsometric formulation with lapse rate adjustment
+    # Using constant lapse rate Γ = 6.5 K/km, with virtual temperature
+    # represented via R_m_surf. This reduces biases over
+    # very cold or very warm high-topography regions.
+    FT = Spaces.undertype(Fields.axes(state.c.ρ))
+    Γ = FT(6.5e-3) # K m^-1
+
+    #   p_msl = p_z0 * [1 + Γ * z / T_z0]^( g / (R_m Γ))
+    # where:
+    #   - p_z0 pressure at the lowest model level
+    #   - T_z0 air temperature at the lowest model level
+    #   - R_m moist-air gas constant at the surface (R_m_surf), which
+    #     accounts for virtual-temperature effects in the exponent
+    #   - Γ constant lapse rate (6.5 K/km here)
+
     if isnothing(out)
-        return @. p_level * exp(g * z_level / (R_m_surf * t_level))
+        return p_level .* (1 .+ Γ .* z_level ./ t_level) .^ (g / Γ ./ R_m_surf)
     else
-        @. out = p_level * exp(g * z_level / (R_m_surf * t_level))
+        out .= p_level .* (1 .+ Γ .* z_level ./ t_level) .^ (g / Γ ./ R_m_surf)
     end
 end
 
@@ -1673,7 +1681,7 @@ add_diagnostic_variable!(
     long_name = "Mean Sea Level Pressure",
     standard_name = "mean_sea_level_pressure",
     units = "Pa",
-    comments = "Mean sea level pressure computed from the hypsometric equation",
+    comments = "Mean sea level pressure computed using a lapse-rate-dependent hypsometric reduction (ERA-style; Γ=6.5 K/km with virtual temperature via moist gas constant).",
     compute! = compute_mslp!,
 )
 
--- a/src/initial_conditions/initial_conditions.jl
+++ b/src/initial_conditions/initial_conditions.jl
@@ -447,73 +447,233 @@ function overwrite_initial_conditions!(
     )
 end
 
+"""
+    correct_surface_pressure_for_topography!(
+        p_sfc,
+        file_path,
+        face_space,
+        Y,
+        ᶜT,
+        ᶜq_tot,
+        thermo_params,
+        regridder_kwargs;
+        surface_altitude_var = "z_sfc",
+    )
+
+Adjusts the surface pressure field `p_sfc` to account for mismatches between
+ERA5 (file) surface altitude and the model orography when specifying pressure.
+
+    Δz = z_model_surface - z_sfc
+
+and applies a hydrostatic correction at the surface using the local moist gas
+constant and temperature at the surface:
+
+    p_sfc .= p_sfc .* exp.(-Δz * g ./ (R_m_sfc .* T_sfc))
+
+where:
+- `g` is gravitational acceleration from `thermo_params`
+- `R_m_sfc` is the moist-air gas constant evaluated from `ᶜq_tot` at the surface
+- `T_sfc` is the air temperature from `ᶜT` at the surface
+
+Returns `true` if the correction is applied; returns `false` if the surface
+altitude field cannot be loaded.
+
+Arguments
+- `p_sfc`: face field of surface pressure to be corrected (modified in-place)
+- `file_path`: path to the ERA5-derived initialization NetCDF file
+- `face_space`: face space of the model grid (for reading/regridding)
+- `Y`: prognostic state, used to obtain model surface height
+- `ᶜT`: center field of temperature
+- `ᶜq_tot`: center field of total specific humidity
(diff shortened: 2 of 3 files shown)


src/utils/weather_model.jl
Julians42, line 136: If it breaks without `surface_geopotential` catch it with the `@assert` here

src/utils/weather_model.jl
```diff
@@ -159,12 +159,28 @@ function to_z_levels(source_file, target_file, target_levels, FT)
...
     # Simply repeat the surface values for all levels
-    surf_map = Dict("skt" => "skt", "sp" => "p")
+    surf_map = Dict("skt" => "skt", "sp" => "p", "surface_geopotential" => "z_sfc")
     for (src_name, dst_name) in surf_map
-        var_obj =
-            defVar(ncout, dst_name, FT, ("lon", "lat", "z"), attrib = ncin[src_name].attrib)
+        # Choose attributes; for z_sfc, set clean altitude attributes
+        var_attrib = if dst_name == "z_sfc"
+            Dict(
+                "standard_name" => "surface_altitude",
```
Julians42, line 230: I assume the data you're writing is geopotential based on the ERA5 name? If that is the case the units are gravity * height (m^2/s^2) and would be good to at least rename the long_name and the units. I would also use zg_sfc even though ERA5 uses z for geopotential because then it's very clear.

src/utils/weather_model.jl
```diff
@@ -159,12 +159,28 @@ function to_z_levels(source_file, target_file, target_levels, FT)
...
+        else
+            ncin[src_name].attrib
+        end
+        var_obj = defVar(ncout, dst_name, FT, ("lon", "lat", "z"), attrib = var_attrib)
+        # Read first time slice and coalesce; follow same convention as sp (use [:, :, 1])
+        data2d = FT.(coalesce.(ncin[src_name][:, :, 1], NaN))
+        # Convert geopotential to meters if necessary
+        if dst_name == "z_sfc"
+            data2d .= data2d ./ grav
+        end
```
Julians42, line 244:
```suggestion
        end
        # Duplicate 2D surface field across all target vertical levels
```
I think this comment would be helpful here

src/utils/weather_model.jl
Julians42: Ah ok I see what is happening now, you're correct. Not sure I have an immediate suggestion other that it is a little confusing because there are two if statements in this loop to catch `z_sfc` since it needs a new var_attrib and needs to be divided by gravity. Possibly separating it out could be cleaner / easier to read if a little longer.

src/utils/weather_model.jl
```diff
@@ -159,12 +159,28 @@ function to_z_levels(source_file, target_file, target_levels, FT)
...
+                "long_name" => "surface altitude derived from ERA5",
+                "units" => "m",
+                "source_variable" => src_name,
+            )
+        else
+            ncin[src_name].attrib
+        end
+        var_obj = defVar(ncout, dst_name, FT, ("lon", "lat", "z"), attrib = var_attrib)
+        # Read first time slice and coalesce; follow same convention as sp (use [:, :, 1])
+        data2d = FT.(coalesce.(ncin[src_name][:, :, 1], NaN))
```
Julians42, line 240: Do we need a coalesce? Are there instances when the data is NaN and our model will still run? If not then we could consider asserting not NaN

src/initial_conditions/initial_conditions.jl
```diff
@@ -371,6 +371,136 @@ function overwrite_initial_conditions!(
...
+    var_name = "z_sfc"
+    ᶠz_surface = nothing
+    # Explicitly check if `z_sfc` exists in the input file; skip if not found
+    has_z_sfc = NC.NCDataset(file_path) do ds
+        haskey(ds, var_name)
+    end
+    if !has_z_sfc
+        @warn "Skipping topographic correction because variable `$var_name` is missing from $(file_path)."
+        return false
+    end
```
Julians42: I think this `haskey` check is maybe better done in `overwrite_initial_conditions!`? Perhaps also passing the variable name `z_sfc` as a function argument would make it more generalizable

src/initial_conditions/initial_conditions.jl
```diff
@@ -371,6 +371,136 @@ function overwrite_initial_conditions!(
...
+    p_sfc_finite = filter(isfinite, p_sfc_arr)
+
+    ᶠz_model_surface = Fields.level(Fields.coordinate_field(Y.f).z, Fields.half)
+    ᶠΔz = zeros(face_space)
+    @. ᶠΔz = ᶠz_model_surface - ᶠz_surface
+
+    # Log surface height statistics
+    z_surface_arr = Array(Fields.field2array(ᶠz_surface))
+    z_model_arr = Array(Fields.field2array(ᶠz_model_surface))
+    Δz_arr = Array(Fields.field2array(ᶠΔz))
```
Julians42: Shouldn't `field2array` already return an array - i.e., does `Array()` actually do anything here?

src/initial_conditions/initial_conditions.jl
```diff
@@ -371,6 +371,136 @@ function overwrite_initial_conditions!(
...
+        one(FT),
+    )
+
+    @. p_sfc =
+        p_sfc *
+        ifelse(
+            isfinite(ᶠΔz),
+            exp(FT(-1) * ᶠΔz * grav / denom),
+            one(FT),
+        )
```
Julians42: Repeated code. either delete the first bit or simplify the second

src/initial_conditions/initial_conditions.jl
```diff
@@ -371,6 +371,136 @@ function overwrite_initial_conditions!(
...
+    ᶠR_m = ᶠinterp.(ᶜR_m)
+    ᶠR_m_sfc = Fields.level(ᶠR_m, Fields.half)
+
+    ᶠT = ᶠinterp.(ᶜT)
+    ᶠT_sfc = Fields.level(ᶠT, Fields.half)
+
+    denom = copy(ᶠR_m_sfc)
+    @. denom = ᶠR_m_sfc * ᶠT_sfc
+    denom_floor = FT(10)
+    @. denom = ifelse(isfinite(denom) && denom > denom_floor, denom, denom_floor)
```
Julians42: This is written confusingly. Could you either comment to explain why we need so many steps (e.g. why is  floor and finite checks needed?) Why is 10 chosen as a floor?

src/initial_conditions/initial_conditions.jl
```diff
@@ -371,6 +371,136 @@ function overwrite_initial_conditions!(
...
+    if ᶠz_surface === nothing
+        return false
+    end
+
+    FT = eltype(thermo_params)
+    grav = thermo_params.grav
+
+    # Log initial surface pressure statistics
+    p_sfc_arr = Array(Fields.field2array(p_sfc))
+    p_sfc_finite = filter(isfinite, p_sfc_arr)
```
Julians42: Again is Array needed after field2array? Is isfinite needed if its surface pressure - shouldn't it be finite? I also think one line and a clearer comment would help

src/diagnostics/core_diagnostics.jl
Julians42: Maybe update the comment to note that this formulation is similar to ERA5 and uses a lapse-rate temperature dependent hypsometric calculation instead of just the standard one

src/diagnostics/core_diagnostics.jl
```diff
@@ -1660,11 +1660,26 @@ function compute_mslp!(out, state, cache, time)
...
+    Γ = FT(6.5e-3) # K m^-1
+    ϵ = FT(1e-6)   # small floor to avoid non-positive base
+
+    # bracket = max(ϵ, 1 + Γ z / T(z))
+    bracket = similar(t_level)
+    @. bracket = max(ϵ, 1 + Γ * z_level / t_level)
+
+    # exponent = g / (R_m Γ)
+    exponent = similar(R_m_surf)
+    @. exponent = g / (R_m_surf * Γ)
```
Julians42: This will allocate new arrays which is slow. Maybe @dennisYatunin could recommend a fix?

src/diagnostics/core_diagnostics.jl
```diff
@@ -1660,11 +1660,26 @@ function compute_mslp!(out, state, cache, time)
     )
     z_level = Fields.level(Fields.coordinate_field(state.c.ρ).z, 1)
 
-    # compute sea level pressure using the hypsometric equation
+    # Reduce to mean sea level using a standard-lapse hypsometric formulation (ERA-style)
```
Julians42: Could you write out the whole equation here for clarity? Then it matters less how you break up the computation for performance. e.g., "p_msl = p(z1)(1+\Gamma z1 / t)^(g/(Rm*\Gamma) is the ERA5 standard calculation for MSLP...".

Julians42 (commented):
Left some comments to improve clarity and then one performance optimization we should do in the diagnostics for Dennis. Glad this is looking so much better!

src/diagnostics/core_diagnostics.jl
costachris: I consolidated the logic and reduced the number of new variables.  But also diagnostics are computed on the CPU so I'm not sure we need to worry about allocations as much here (note we already materialize a lot of our lazy fields throughout diagnostics as is). But maybe @dennisYatunin can comment how much it matters in the end.

src/initial_conditions/initial_conditions.jl
costachris: I was using this to log and debug something. It's removed!

src/initial_conditions/initial_conditions.jl
costachris: The 2 clippings are removed + the logic is consolidated

Julians42 (approved):

PR #4099, opened 2025-11-13
Expose ClimaParams strict option

costachris:
Expose ClimaParams strict option. Needed for running coupled runs with tomls containing both land and atmosphere parameters in ClimaCoupler.jl.

### DIFF
--- a/config/default_configs/default_config.yml
+++ b/config/default_configs/default_config.yml
@@ -289,6 +289,9 @@ subsidence:
 toml:
   help: "TOML file(s) used to override model parameters"
   value: []
+strict_params:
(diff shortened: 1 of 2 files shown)


config/default_configs/default_config.yml
```diff
@@ -289,6 +289,9 @@ subsidence:
 toml:
   help: "TOML file(s) used to override model parameters"
   value: []
+strict_climaparams:
+  help: "Strict mode for ClimaParams."
```
nefrathenrici:
```suggestion
  help: "Raise an error if any overridden parameters are unused."
```

PR #4105, opened 2025-11-14
Moisture fixer

trontrytel:
Adds a tendency that adjusts negative small cloud and precipitation values towards zero at the expense of water vapor. Switched off by default to prevent it from masking other problems.

### DIFF
--- a/src/prognostic_equations/remaining_tendency.jl
+++ b/src/prognostic_equations/remaining_tendency.jl
@@ -7,7 +7,7 @@ state vector `Y`.
 This function follows a sequence:
 1. Prepares hyperdiffusion tendencies for tracers (stored in `Yₜ_lim`).
 2. Prepares hyperdiffusion tendencies for other state variables (e.g., momentum, energy, stored in `Yₜ`).
-3. If Direct Stiffness Summation (DSS) is required and hyperdiffusion is active, performs DSS on the 
+3. If Direct Stiffness Summation (DSS) is required and hyperdiffusion is active, performs DSS on the
    prepared hyperdiffusion tendencies.
 4. Applies the (potentially DSSed) hyperdiffusion tendencies to `Yₜ_lim` and `Yₜ`.
 
@@ -356,7 +356,11 @@ NVTX.@annotate function additional_tendency!(Yₜ, Y, p, t)
     horizontal_amd_tendency!(Yₜ, Y, p, t, amd)
     vertical_amd_tendency!(Yₜ, Y, p, t, amd)
 
-    # NOTE: This will zero out all momentum tendencies in the EDMFX advection test, 
+    # Optional tendency to bring negative small tracers back from negative
+    # at the cost of water vapor.
+    moisture_fixer_tendency!(Yₜ, Y, p, t, moisture_model, microphysics_model)
+
+    # NOTE: This will zero out all momentum tendencies in the EDMFX advection test,
     # where velocities do not evolve
     # DO NOT add additional velocity tendencies after this function
     zero_velocity_tendency!(Yₜ, Y, p, t)
--- a/src/solver/types.jl
+++ b/src/solver/types.jl
@@ -610,6 +610,7 @@ Base.@kwdef struct AtmosWater{MM, PM, CM, NCFM, CCDPS}
     cloud_model::CM = nothing
     noneq_cloud_formation_mode::NCFM = nothing
     call_cloud_diagnostics_per_stage::CCDPS = nothing
+    moisture_fixer::Bool = false
 end
 
 """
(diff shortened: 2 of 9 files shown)


src/prognostic_equations/remaining_tendency.jl
```diff
@@ -356,7 +356,30 @@ NVTX.@annotate function additional_tendency!(Yₜ, Y, p, t)
...
+            moisture_species,
+        ) do (ρq_name)
+            ᶜρq = MatrixFields.get_field(Y, ρq_name)
+            ᶜρqₜ = MatrixFields.get_field(Yₜ, ρq_name)
+            ᶜq = @. lazy(specific(ᶜρq, Y.c.ρ))
+            # Increase the grid mean small tracers if negative,
+            # using mass from grid mean vapor.
+            @. ᶜρqₜ += Y.c.ρ * moisture_fixer(ᶜq, qᵥ, p.dt)
+        end
+    end
```
szy21: I think it would be better to move this to a function under `src/parameterized_tendencies/microphysics`.

src/solver/types.jl
```diff
@@ -604,12 +604,13 @@ end
...
 Groups moisture-related models and types.
 """
-Base.@kwdef struct AtmosWater{MM, PM, CM, NCFM, CCDPS}
+Base.@kwdef struct AtmosWater{MM, PM, CM, NCFM, CCDPS, MF}
     moisture_model::MM = nothing
     microphysics_model::PM = nothing
     cloud_model::CM = nothing
     noneq_cloud_formation_mode::NCFM = nothing
     call_cloud_diagnostics_per_stage::CCDPS = nothing
+    moisture_fixer::MF = false
```
szy21:
```suggestion
    moisture_fixer::Bool = false
```
(I don't think we need an extra type here)

src/prognostic_equations/remaining_tendency.jl
```diff
@@ -356,7 +356,30 @@ NVTX.@annotate function additional_tendency!(Yₜ, Y, p, t)
...
-    # NOTE: This will zero out all momentum tendencies in the EDMFX advection test, 
+    if moisture_model isa NonEquilMoistModel &&
+       microphysics_model isa Microphysics1Moment &&
+       p.atmos.water.moisture_fixer
+        moisture_species = (@name(c.ρq_liq), @name(c.ρq_ice),
+            @name(c.ρq_rai), @name(c.ρq_sno),
+        )
+        qᵥ = @. lazy(
+            specific(Y.c.ρq_tot - Y.c.ρq_liq - Y.c.ρq_ice - Y.c.ρq_rai - Y.c.ρq_sno,
+                Y.c.ρ),
```
sajjadazimi: Maybe we can define a function to return vapor specific humidity as it is used multiple times at different places. This way we don't have to check the microphysics model; we can have a list of tracers and return if the tracer does not exist and fix if it exists.

src/prognostic_equations/remaining_tendency.jl
```diff
@@ -356,7 +356,30 @@ NVTX.@annotate function additional_tendency!(Yₜ, Y, p, t)
     horizontal_amd_tendency!(Yₜ, Y, p, t, amd)
     vertical_amd_tendency!(Yₜ, Y, p, t, amd)
 
-    # NOTE: This will zero out all momentum tendencies in the EDMFX advection test, 
+    if moisture_model isa NonEquilMoistModel &&
+       microphysics_model isa Microphysics1Moment &&
```
sajjadazimi: We can apply the fixer for both 1M or 2M microphysics models.

sajjadazimi (approved):
Thank you! I just left a comment about how we compute qv. We can define a function for computing qv (depending on microphyiscs model) and then apply the fixer on a list of tracers (moisture_species) without checking moisture and microphyiscs models. But feel free to merge as is and leave this to future PRs.

src/prognostic_equations/remaining_tendency.jl
trontrytel: Technically there is one, but in Thermodynamics and I don't want to use it because it's clipping it to positive numbers.  I guess I have one in ThermodynamicsInterface in CloudMicrophysics. I'll use that one.

src/prognostic_equations/remaining_tendency.jl
trontrytel: Will do

src/prognostic_equations/remaining_tendency.jl
trontrytel: I'll put it in a separate file in parameterized tendencies. I want to make it clear that it's a numerics tendency and not a microphysics tendency.

src/solver/types.jl
trontrytel: Will do!

src/prognostic_equations/remaining_tendency.jl
szy21: ok!

PR #4106, opened 2025-11-14
pedmf boundary conditions

sajjadazimi:
## Purpose 
This PR updates the treatment of boundary conditions for prognostic EDMF:
1- Apply surface fluxes to both grid-mean and updraft fields equally.
2- Seed a small updraft area fraction in the first cell when the surface buoyancy flux is positive, allowing it to grow even if initially zero.
3- Instead of overwriting the first-cell area fraction, increase the entrainment rate when buoyancy flux is positive, ensuring the area reaches the prescribed value in one timestep. This ensures that tracer masses are conserved.
4- Instead of overwriting mse and q_tot, assume a near-surface entrainment in the first cell, with tracer values of the entraining fluid provided by the function `sgs_scalar_first_interior_bc` (mean + perturbation).

## Content

----

### DIFF
--- a/src/cache/prognostic_edmf_precomputed_quantities.jl
+++ b/src/cache/prognostic_edmf_precomputed_quantities.jl
@@ -116,150 +116,6 @@ NVTX.@annotate function set_prognostic_edmf_precomputed_quantities_draft!(
 end
 
 """
-    set_prognostic_edmf_precomputed_quantities_bottom_bc!(Y, p, ᶠuₕ³, t)
-
-Updates velocity and thermodynamics quantities at the surface in each SGS draft.
-"""
-NVTX.@annotate function set_prognostic_edmf_precomputed_quantities_bottom_bc!(
-    Y,
-    p,
-    t,
-)
-    (; moisture_model, turbconv_model, microphysics_model) = p.atmos
-
-    FT = Spaces.undertype(axes(Y.c))
-    n = n_mass_flux_subdomains(turbconv_model)
-    thermo_params = CAP.thermodynamics_params(p.params)
-    turbconv_params = CAP.turbconv_params(p.params)
-
-    (; ᶜΦ,) = p.core
-    (; ᶜp, ᶜK, ᶜtsʲs, ᶜρʲs, ᶜts) = p.precomputed
-    (; ustar, obukhov_length, buoyancy_flux) = p.precomputed.sfc_conditions
-
-    for j in 1:n
-        ᶜtsʲ = ᶜtsʲs.:($j)
-        ᶜmseʲ = Y.c.sgsʲs.:($j).mse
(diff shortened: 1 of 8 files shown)


tapios (approved):
LGTM. It'll be good to see how well this work across a range of SCM simulations. This PR looks conceptually and code-wise fine to me.

src/cache/prognostic_edmf_precomputed_quantities.jl
```diff
@@ -411,6 +268,27 @@ NVTX.@annotate function set_prognostic_edmf_precomputed_quantities_explicit_clos
...
+            Fields.field_values(Fields.level(Y.c.sgsʲs.:($j).ρa, 1))
+        @. ᶜaʲ_int_val = draft_area(sgsʲs_ρa_int_val, sgsʲs_ρ_int_val)
+        entr_int_val = Fields.field_values(Fields.level(ᶜentrʲs.:($j), 1))
+        detr_int_val = Fields.field_values(Fields.level(ᶜdetrʲs.:($j), 1))
+        @. entr_int_val = limit_entrainment(
+            ifelse(
+                buoyancy_flux_val < 0 || ᶜaʲ_int_val >= $(FT(turbconv_params.surface_area)),
+                entr_int_val,
+                detr_int_val +
+                ($(FT(turbconv_params.surface_area)) / ᶜaʲ_int_val - FT(1)) / FT(dt),
```
szy21: Add an eps to ᶜaʲ_int_val to be safe? Also do you need the FTs in this function?

src/prognostic_equations/remaining_tendency.jl
```diff
@@ -284,6 +284,7 @@ NVTX.@annotate function additional_tendency!(Yₜ, Y, p, t)
     if p.atmos.sgs_entr_detr_mode == Explicit()
         edmfx_entr_detr_tendency!(Yₜ, Y, p, t, p.atmos.turbconv_model)
     end
+    edmfx_first_interior_entr_tendency!(Yₜ, Y, p, t, p.atmos.turbconv_model)
```
szy21, line 287: Does this need to be implicit?

src/cache/prognostic_edmf_precomputed_quantities.jl
sajjadazimi: The function `limit_entrainment` returns a safe value even if the input is inf. For the FTs I think the one around number one is not necessary.

src/prognostic_equations/remaining_tendency.jl
sajjadazimi: entrainment and detrainment are written in a safe way to avoid numerical instabilities even if done explicitly. As CI implicit tests seem to pass without a problem I think it's fine to leave making this function implicit to future PRs if necessary.

PR #4108, opened 2025-11-15
remove dyamond longrun

szy21:
## Purpose 

We don't need it anymore.

## Content

----


szy21:
Manually merging this in as it only changes the longrun pipeline

PR #4111, opened 2025-11-20
Fix bg_gradient lh bug

trontrytel:
This PR fixes the latent heat bug discovered by @jbphyswx in our buoyancy gradients.  

It would be great to rewrite parts of that file to get rid of the getter functions at the top and the bg_model, and to reduce the number of functions we define. But maybe we can do it in the next PRs.

### DIFF
--- a/src/prognostic_equations/eddy_diffusion_closures.jl
+++ b/src/prognostic_equations/eddy_diffusion_closures.jl
@@ -152,10 +152,20 @@ function buoyancy_gradients(
             error("Unsupported moisture model")
         end
 
-        phase_part = TD.PhasePartition(thermo_params, ts_sat)
-        lh = TD.latent_heat_liq_ice(thermo_params, phase_part)
-        cp_m = TD.cp_m(thermo_params, ts_sat)
         t_sat = get_t_sat(thermo_params, bg_model)
+        phase_part = TD.PhasePartition(thermo_params, ts_sat)
+        lh =
+            (
+                get_ql_sat(thermo_params, bg_model) *
+                TD.latent_heat_vapor(thermo_params, t_sat) +
+                get_qi_sat(thermo_params, bg_model) *
+                TD.latent_heat_sublim(thermo_params, t_sat)
+            ) /
+            (
+                get_ql_sat(thermo_params, bg_model) +
+                get_qi_sat(thermo_params, bg_model) + sqrt(eps(FT))
+            )
+        cp_m = TD.cp_m(thermo_params, ts_sat)
         qv_sat = get_qv_sat(thermo_params, bg_model)
         ∂b∂θli_sat = (
             ∂b∂θv *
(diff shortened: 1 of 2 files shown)


src/prognostic_equations/eddy_diffusion_closures.jl
```diff
@@ -152,10 +152,22 @@ function buoyancy_gradients(
...
+        # TODO - add here also precipiotation once the code is simplifies
+        # and we don't rely on thermo state to get the condensed species.
+        lh =
+            (
+                get_ql_sat(thermo_params, bg_model) *
+                TD.latent_heat_vapor(thermo_params, t_sat) +
+                get_qi_sat(thermo_params, bg_model) *
+                TD.latent_heat_sublim(thermo_params, t_sat)
+            ) /
+            (get_ql_sat(thermo_params, bg_model) + get_qi_sat(thermo_params, bg_model))
```
szy21: add an eps to make sure this is not divided by zero?

szy21 (approved):
Looks good, just one comment.

src/prognostic_equations/eddy_diffusion_closures.jl
trontrytel: Good call. I thought I would be able to avoid it since we are only in that function if we have some non-zero cloud condensate. But I guess the numerics were not playing well with me - I actually need and `sqrt(eps(FT))` in the denominator to make the implicit Rico single column test case work

szy21: sorry, I meant max(..., eps(FT)), but I can change it later. (or does max not work?)

jbphyswx: Did it change the results of any of the CI? I'm not super familiar with all the plots.
Also was wondering if it should converge to a real `L_v` at  `q=0` instead of `0`, not that it should make a huge difference in the presence of any real cloud. I guess in equilibrium, condensate converges to 0 at the same time as the supersaturation, but that's less true in noneq.

`sqrt(eps(Float64))` is only `1.49e-8` which is not unheard of for thin ice clouds, and if we ever ran `Float32` simulations is `3e-4` which isn't insignificant (though I don't know if that's intended to be possible)

When I trialed fixing `bg` I just set to `lh` to `Lv` at `ql+qi < eps(FT)` -- I guess you could also just add `+eps(FT)L_v` or something similar to the numerator and denominator if you wanted it to be more numerically stable  and continuous.

szy21: prognostic edmf dycoms in ci changed. There are probably some other changes too. And we always run Float32. I agree we should do this more carefully. Could you work together with @trontrytel to make this better?

szy21: Oh, I have an idea. We can use TD.liquid_fraction here? That way Lh is always between Lv and Ls.

PR #4112, opened 2025-11-20
ci: JET v0.10, v0.11 in .buildkite compat

haakon-e:
- this enables support for Julia v1.12
- retain JET v0.9 to support Julia v1.11

### DIFF
--- a/.buildkite/Project.toml
+++ b/.buildkite/Project.toml
@@ -40,11 +40,15 @@ StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 YAML = "ddb6d928-2868-570f-bddf-ab3f9cf99eb6"
 
+[sources]
+ClimaAtmos = {path = ".."}
+PrecompileCI = {path = "PrecompileCI"}
+
 [compat]
 CairoMakie = "0.11 - 0.15"
 ClimaCoreSpectra = "0.1"
 ClimaCoreTempestRemap = "0.3"
-JET = "0.9"
+JET = "0.9, 0.10, 0.11"
 PrettyTables = "2"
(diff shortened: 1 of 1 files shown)


haakon-e:
I have not run the entirety of ci, no. Perhaps we could chat about how to do this sometime?
- I know I can update the `climacommon` version in `.buildkite/pipeline.yml`, but I don't want to mess up the precompile cache for all other runs, even if just to test running the entirety of ci on v1.12.

This proposed change is mostly motivated by wanting to make use of quality-of-life improvements in v1.12; the simulations I've been running so far have not complained. 

It is impossible to precompile the `.buildkite` environment in v1.12 without this change, so note that this PR does not propose we upgrade ci from 1.11 to 1.12, rather it is intended for users who'd like to start working with v1.12 before a formal upgrade. I think we can upgrade the entire pipeline to v1.12 at a later time.

imreddyTeja:
> I have not run the entirety of ci, no. Perhaps we could chat about how to do this sometime?
> 
> * I know I can update the `climacommon` version in `.buildkite/pipeline.yml`, but I don't want to mess up the precompile cache for all other runs, even if just to test running the entirety of ci on v1.12.
> 
> This proposed change is mostly motivated by wanting to make use of quality-of-life improvements in v1.12; the simulations I've been running so far have not complained.
> 
> It is impossible to precompile the `.buildkite` environment in v1.12 without this change, so note that this PR does not propose we upgrade ci from 1.11 to 1.12, rather it is intended for users who'd like to start working with v1.12 before a formal upgrade. I think we can upgrade the entire pipeline to v1.12 at a later time.

I'm not completely confident in this, but I think precompile caches are in different folders for different Julia minor versions. 
I was mainly asking about this to guess how easy adding "official" support for 1.12 will be.

haakon-e:

- #4112 ◀
    - #4168

PR #4118, opened 2025-11-25
clean up buoyancy gradient

szy21:
## Purpose 

This is the first part of buoyancy gradient clean up. This PR removes getter functions and some if-else logic. The result is unchanged. In the next PR(s) I will use the actual cloud fraction for the weight in the buoyancy gradient calculation, which will change the result.

## Content

----

### DIFF
--- a/src/prognostic_equations/eddy_diffusion_closures.jl
+++ b/src/prognostic_equations/eddy_diffusion_closures.jl
@@ -7,28 +7,6 @@ import Thermodynamics.Parameters as TDP
 import ClimaCore.Geometry as Geometry
 import ClimaCore.Fields as Fields
 
-get_t_sat(thermo_params, x::EnvBuoyGradVars) =
-    TD.air_temperature(thermo_params, x.ts)
-get_p(thermo_params, x::EnvBuoyGradVars) = TD.air_pressure(thermo_params, x.ts)
-get_ρ(thermo_params, x::EnvBuoyGradVars) = TD.air_density(thermo_params, x.ts)
-get_en_cld_frac(thermo_params, x::EnvBuoyGradVars) =
-    ifelse(TD.has_condensate(thermo_params, x.ts), 1, 0)
-get_θ_liq_ice_sat(thermo_params, x::EnvBuoyGradVars) =
-    TD.liquid_ice_pottemp(thermo_params, x.ts)
-get_qt_sat(thermo_params, x::EnvBuoyGradVars) =
-    TD.total_specific_humidity(thermo_params, x.ts)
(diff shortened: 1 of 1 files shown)


src/prognostic_equations/eddy_diffusion_closures.jl
```diff
@@ -124,54 +102,27 @@ function buoyancy_gradients(
...
-            (lh / cp_m / t_sat * ∂b∂θli_sat - ∂b∂θv) *
-            get_θ_sat(thermo_params, bg_model)
-    else
-        ∂b∂θli_sat = FT(0)
-        ∂b∂qt_sat = FT(0)
-    end
+    ts = bg_model.ts
+    p = TD.air_pressure(thermo_params, ts)
+    Π = TD.exner_given_pressure(thermo_params, p)
+    ∂b∂θv = g * (R_d * TD.air_density(thermo_params, ts) / p) * Π
```
tapios, line 108: Isn't this more easily written as `g * (R_d /R_m) * Π/T`, or better yet `g/θv` with virtual potential temperature from Thermodynamics.jl?

src/prognostic_equations/eddy_diffusion_closures.jl
```diff
@@ -124,54 +102,27 @@ function buoyancy_gradients(
...
+    lh =
+        λ * TD.latent_heat_vapor(thermo_params, t_sat) +
+        (1 - λ) * TD.latent_heat_sublim(thermo_params, t_sat)
+    cp_m = TD.cp_m(thermo_params, ts)
+    qv_sat = TD.vapor_specific_humidity(thermo_params, ts)
+    qt_sat = TD.total_specific_humidity(thermo_params, ts)
+    ∂b∂θli_sat = (
+        ∂b∂θv *
+        (1 + Rv_over_Rd * (1 + lh / R_v / t_sat) * qv_sat - qt_sat) /
+        (1 + lh * lh / cp_m / R_v / t_sat / t_sat * qv_sat)
```
tapios, line 121: Why is `lh * lh` better than `lh^2`? (same for temperature)

src/prognostic_equations/eddy_diffusion_closures.jl
szy21: Yes, I tested it and it changed the result, just from the order of computation I think. To keep this PR not behavior changing, I changed it back, but I'm going to change this in my next PR.

src/prognostic_equations/eddy_diffusion_closures.jl
szy21: I don't know. Sometimes ^2 allocates, but I think it's ok here. This will cause a behavior change too - I'll change it in my next PR.

PR #4121, opened 2025-11-25
fix sgs mass flux of microphysics tracers in diagnostic edmf

szy21:
## Purpose 

Adds back the grid-mean advection of microphysics tracers in diagnostic EDMF and uses the difference for sgs mass flux. Also removes the j loop for microphysics tracers as we always use the first updraft for the tracers, so it is a bit misleading.

Note: This is just one way to fix it. I chose this so it is similar to qt flux. But it is different from how we calculate flux in prognostic EDMF. We can change it to a different way if we think it is better.

## Content

----

### DIFF
--- a/src/prognostic_equations/edmfx_sgs_flux.jl
+++ b/src/prognostic_equations/edmfx_sgs_flux.jl
@@ -135,29 +135,27 @@ function edmfx_sgs_mass_flux_tendency!(
                 (@name(c.ρn_liq), @name(c.sgsʲs.:(1).n_liq), @name(n_liq)),
                 (@name(c.ρn_rai), @name(c.sgsʲs.:(1).n_rai), @name(n_rai)),
             )
-            for j in 1:n
-                # TODO using unrolled_foreach here allocates! (breaks the flame tests
-                # even though they use 0M microphysics)
-                # MatrixFields.unrolled_foreach(
-                #     microphysics_tracers,
-                # ) do (ρχ_name, χʲ_name, _)
-                for (ρχ_name, χʲ_name, _) in microphysics_tracers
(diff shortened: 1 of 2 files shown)


src/prognostic_equations/edmfx_sgs_flux.jl
```diff
@@ -302,10 +305,10 @@ function edmfx_sgs_mass_flux_tendency!(
                         )
                     vtt = vertical_transport(
                         ᶜρʲs.:($j),
-                        ᶠu³ʲs.:($j),
+                        ᶠu³_diff,
                         ᶜa_scalar,
                         dt,
-                        edmfx_tracer_upwinding,
+                        edmfx_sgsflux_upwinding,
```
szy21: You don't have to change this. `edmfx_sgsflux_upwinding` is None by default and it was better for 0m previously, but I didn't test it for 1M.

src/prognostic_equations/edmfx_sgs_flux.jl
```diff
@@ -276,40 +274,41 @@ function edmfx_sgs_mass_flux_tendency!(
...
                     )
-                    ᶜρχₜ = MatrixFields.get_field(Yₜ, ρχ_name)
-                    @. ᶜρχₜ += vtt
-                end
+                vtt = vertical_transport(
+                    ᶜρʲs.:(1),
+                    ᶠu³_diff,
+                    ᶜa_scalar,
+                    dt,
+                    edmfx_sgsflux_upwinding,
```
szy21, line 308: Note: I changed the flag here and this is central difference, which looks slightly better in dycoms. But you can try if `edmfx_tracer_upwinding` is better for other scms / global run.

PR #4123, opened 2025-11-26
Rewrite updraft_cloud_diagnostics as a point function

daverumph:
Rewrite updraft_cloud_diagnostics in cloud_fraction.jl as a point function
so that it behaves better as a CUDA kernel

## Content
No change in behavior.  Should be a bit faster running on GPU.

----
- [ X] I have read and checked the items on the review checklist.


src/cache/cloud_fraction.jl
```diff
@@ -123,6 +123,23 @@ NVTX.@annotate function set_cloud_fraction!(
...
+    FT = eltype(thermo_params)
+    cf = FT(0)
+    q_liq = FT(0)
+    q_ice = FT(0)
+    for j in 1:n
+        aʲ = draft_area(ρaʲs.:($j), ρʲs.:($j))
+        if TD.has_condensate(thermo_params, tsʲs.:($j))
+            cf += aʲ
+        end
+        (; liq, ice) = TD.PhasePartition(thermo_params, tsʲs.:($j))
```
haakon-e, line 136:
```suggestion
    unrolled_map(ρaʲs, ρʲs, tsʲs) do ρaʲ, ρʲ, tsʲ
        aʲ = draft_area(ρaʲ, ρʲ)
        if TD.has_condensate(thermo_params, tsʲ)
            cf += aʲ
        end
        (; liq, ice) = TD.PhasePartition(thermo_params, tsʲ)
```
I wonder if unrolling the loop is even a bit faster? 
at least a little less intimidating without the `.:($j)` 🙂

trontrytel:
This is probably very stale by now. Should we close it?

PR #4135, opened 2025-12-04
use the correct a_m in mixing length

szy21:
## Purpose 

The Businger a_m was used in the stable branch of mixing length. This PR corrects it to use Gryanik a_m. Also modifies the stability function to use the same optimization as the one in SurfaceFluxes.

## Content

----


tapios (approved):
Looks good. But isn't there a way to use `phi` directly from SurfaceFluxes, with the surface flux parameterization being part of a config choice and passing the same config choice through here?

PR #4139, opened 2025-12-05
fix jacobian terms related to dpdqt

szy21:
## Purpose 

Makes the jacobian terms that related to dp/dqt more consistent with the updraft. These includes ∂ᶠu₃_err_∂ᶜρq_tot and ∂ᶠᶜρe_tot_err_∂ᶜρq_tot. Also fixes a bug we had before in ∂ᶠᶜρe_tot_err_∂ᶜρq_tot.

I think there might be a small error in ∂ᶠᶜρe_tot_err_∂ᶜρ too. I will look at that next.

## Content

----

### DIFF
--- a/src/prognostic_equations/implicit/manual_sparse_jacobian.jl
+++ b/src/prognostic_equations/implicit/manual_sparse_jacobian.jl
@@ -412,18 +412,19 @@ function update_jacobian!(alg::ManualSparseJacobian, cache, Y, p, dtγ, t)
     Δcv_v = FT(CAP.cv_v(params)) - cv_d
     T_0 = FT(CAP.T_0(params))
     R_d = FT(CAP.R_d(params))
-    ΔR_v = FT(CAP.R_v(params)) - R_d
+    R_v = FT(CAP.R_v(params))
+    ΔR_v = R_v - R_d
     cp_d = FT(CAP.cp_d(params))
     Δcp_v = FT(CAP.cp_v(params)) - cp_d
+    e_int_v0 = FT(CAP.e_int_v0(params))
     LH_v0 = FT(CAP.LH_v0(params))
     LH_s0 = FT(CAP.LH_s0(params))
-    R_v = FT(CAP.R_v(params))
     Δcp_l = FT(CAP.cp_l(params) - CAP.cp_v(params))
     Δcp_i = FT(CAP.cp_i(params) - CAP.cp_v(params))
     # This term appears a few times in the Jacobian, and is technically
     # minus ∂e_int_∂q_tot
     thermo_params = CAP.thermodynamics_params(params)
-    ∂e_int_∂q_tot = T_0 * (Δcv_v - R_d) - FT(CAP.e_int_v0(params))
+    ∂e_int_∂q_tot = T_0 * (Δcv_v - R_d) - e_int_v0
     ᶜh_tot = @. lazy(
         TD.total_specific_enthalpy(
             thermo_params,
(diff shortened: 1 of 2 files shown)


config/model_configs/baroclinic_wave_equil.yml
szy21, line 1: I will revert this

src/prognostic_equations/implicit/manual_sparse_jacobian.jl
```diff
@@ -448,13 +452,9 @@ function update_jacobian!(alg::ManualSparseJacobian, cache, Y, p, dtγ, t)
...
         TD.gas_constant_air(thermo_params, ᶜts) / TD.cv_m(thermo_params, ᶜts)
 
-    ᶜ∂kappa_m∂q_tot = p.scratch.ᶜtemp_scalar_2
-    # Using abs2 because ^2 results in allocation
-    @. ᶜ∂kappa_m∂q_tot =
-        (
-            ΔR_v * TD.cv_m(thermo_params, ᶜts) -
-            Δcv_v * TD.gas_constant_air(thermo_params, ᶜts)
-        ) / abs2(TD.cv_m(thermo_params, ᶜts))
+    T = @. lazy(TD.air_temperature(thermo_params, ᶜts))
```
szy21, line 450: This works on GPU, I'll change the one in microphysics jacobian too. This is a very simple function so I don't think it will make things slower.

src/prognostic_equations/implicit/manual_sparse_jacobian.jl
```diff
@@ -92,11 +92,14 @@ function jacobian_cache(alg::ManualSparseJacobian, Y, atmos)
         is_in_Y(@name(c.sgs⁰.ρatke)) ? (@name(c.sgs⁰.ρatke),) : ()
     sfc_if_available = is_in_Y(@name(sfc)) ? (@name(sfc),) : ()
 
-    condensate_names = (
+    condensate_mass_names = (
```
szy21: Not really needed in this PR. I'll revert it.

src/prognostic_equations/implicit/manual_sparse_jacobian.jl
```diff
@@ -446,13 +447,9 @@ function update_jacobian!(alg::ManualSparseJacobian, cache, Y, p, dtγ, t)
...
-    ᶜ∂kappa_m∂q_tot = p.scratch.ᶜtemp_scalar_2
-    # Using abs2 because ^2 results in allocation
-    @. ᶜ∂kappa_m∂q_tot =
-        (
-            ΔR_v * TD.cv_m(thermo_params, ᶜts) -
-            Δcv_v * TD.gas_constant_air(thermo_params, ᶜts)
-        ) / abs2(TD.cv_m(thermo_params, ᶜts))
+    T = @. lazy(TD.air_temperature(thermo_params, ᶜts))
+    ᶜ∂RmT∂q = p.scratch.ᶜtemp_scalar_2
+    @. ᶜ∂RmT∂q = ᶜkappa_m * (-e_int_v0 - R_d * T_0 - Δcv_v * (T - T_0)) + ΔR_v * T
```
sajjadazimi: This is d(rho R T)/d(rho q). Maybe we can call it dp/d(rho q) as well.

szy21: I just realized d(rho R T)/d(rho q) is the same as d(R T) / dq when rho is fixed. But I'll call it dp / d(rho q) to be more clear.

PR #4143, opened 2025-12-06
use universal functions in mixing length

### LINKED ISSUE #2465: Use universal function from SurfaceFluxes for l_W
For the wall constraint mixing length [here](https://github.com/CliMA/ClimaAtmos.jl/blob/d9ebb2253e14b7da1b2fccdff9c10a7da8c849ed/src/prognostic_equations/edmfx_closures.jl#L223), it would be better to use the universal function from SurfaceFluxes.

szy21:
## Purpose 

Uses universal functions from SurfaceFluxes directly in the mixing length calculation. Closes #2465 

## To-do

- [x] Remove Gryanik parameters if this works

## Content

----


szy21:
The one we have before has a few numeric safety checks - I hope SurfaceFluxes will handle them?

PR #4150, opened 2025-12-09
Add constant horizontal diffusion option for testing

trontrytel:
I would like to add constant horizontal diffusion option to ClimaAtmos for testing.


trontrytel:
Going to clear the deopt so that new ClimaParams are used

config/default_configs/default_config.yml
```diff
@@ -169,6 +169,9 @@ amd_les:
 c_amd:
   help: "Model coefficient for AMD-LES closure (TODO: Move to parameters.toml)"
   value: 0.29
+constant_horizontal_dissipation:
```
petebachant: Diffusion and dissipation appear to be used interchangeably here. Is that intentional?

szy21: I think it would be good to use diffusion consistently.

trontrytel: Sounds good. I was kind of waiting to see which one you would prefer ;)

szy21 (approved):
We talked offline - We can merge this in, but it would be nice to combine the different options of horizontal (and vertical) diffusion at some point, so we don't duplicate the code for the diffusive tendencies.

PR #4151, opened 2025-12-09
Update ClimaDiagnostics in compat

ph-kev:
This updates ClimaDiagnostics to v0.3. The only breaking change is that the diagnostics are saved at the start of the reduction period instead of at the end.

I don't think this affects anything, but the postprocessing of the NetCDF files. I couldn't find any postprocessing of the diagnostics that depends on the exact time values beside the calibration.

In general, any plots with instantaneous variables should be unchanged, but any plots with time reduction should differ by single time period. For example, a plot with `time = 8d` would change to `time = 7d` if the time reduction is daily.

ClimaDiagnostics v0.2.13 is still supported, because ClimaCoupler does not support v0.3 yet.


juliasloan25 (commented):
This looks good to me! I would just want to hear from Costa about the experiment you mentioned before merging.

@szy21 can we tag a patch release so we can update the coupler too, or should we wait since things are broken right now?

szy21:
> This looks good to me! I would just want to hear from Costa about the experiment you mentioned before merging.
> 
> @szy21 can we tag a patch release so we can update the coupler too, or should we wait since things are broken right now?

We can tag but it will be a breaking release because of this: https://github.com/CliMA/ClimaAtmos.jl/pull/4198. Maybe we can tag a release when the coupler surface flux PR is ready?

szy21:
> > > This looks good to me! I would just want to hear from Costa about the experiment you mentioned before merging.
> > > @szy21 can we tag a patch release so we can update the coupler too, or should we wait since things are broken right now?
> > 
> > 
> > We can tag but it will be a breaking release because of this: #4198. Maybe we can tag a release when the coupler surface flux PR is ready?
> 
> Sounds good. Teja and I are also working on updating the Makie compats in ClimaCore/ClimaAtmos/ClimaLand. It will be good to make releases of those packages when those are in too

Sounds good. We can tag the next minor release after all of these are done.

nefrathenrici (approved):
This looks good to me. @costachris can chime in, but I think we should add ClimaDiagnostics and a compat "=0.2.15" to the calibration experiment [Project.toml](https://github.com/CliMA/ClimaAtmos.jl/blob/main/calibration/experiments/gcm_driven_scm/Project.toml). That should help keep the pipeline reproducible.

calibration/experiments/gcm_driven_scm/Project.toml
```diff
@@ -19,5 +20,7 @@ NCDatasets = "85f8d34a-cbdd-5861-8df4-14fed0d494ab"
 YAML = "ddb6d928-2868-570f-bddf-ab3f9cf99eb6"
 
 [compat]
-ClimaCalibrate = "=0.0.3"
+ClimaAtmos = "=0.27.6"
+ClimaCalibrate = "=0.1.4"
+ClimaDiagnostics = "=0.2.15"
```
ph-kev, line 25: I wasn't sure about the versions of the packages. I don't think there was anything breaking going from ClimaCalibrate 0.0.3 to 0.1.4.

cc: @nefrathenrici

PR #4153, opened 2025-12-09
Refactor atmospheric gases

oalcabes:
This PR refactors the way atmospheric gases are treated in the code. Namely, it:

- gives all atmospheric gases two options: either to be time varying or fixed. Only two gases have time varying implementations currently -- ozone and co2
- sets fixed values for atmospheric gases from values given in toml, rather than from RRTMGP artifact
- removes individual model structs for ozone and co2

### DIFF
--- a/config/default_configs/default_config.yml
+++ b/config/default_configs/default_config.yml
@@ -337,17 +337,14 @@ deep_atmosphere:
 restart_file:
   help: "Path to HDF5 file to use as simulation starting point"
   value: ~
-prescribe_ozone:
-  help: "Prescribe time and spatially varying ozone from a file [`false` (default), `true`]"
-  value: false
-co2_model:
-  help: "What CO2 concentration to use for RRTGMP. When fixed, it is set to 397.547 parts per million. Otherwise, it is read from the MaunaLoa measuraments. [`nothing` (default), `Fixed`, `MaunaLoa`]"
-  value: ~
 detect_restart_file:
   help: "When true, try finding a restart file and use it to restart the simulation. Only works with ActiveLink."
   value: false
 prescribed_aerosols:
   help: "Which aerosols to add. List of keys from the data file (e.g., CB1, CB2)."
+  value: []
+time_varying_trace_gases:
+  help: "Trace gases that vary in time, e.g. [`CO2`, `O3`]. This setting is only relevant when using RRTMGP. Currently, only `CO2` and `O3` are supported. All other trace gases set fixed to default values from ClimaParams."
   value: []
 # TODO: remove once https://github.com/CliMA/ClimaAtmos.jl/issues/2873 is closed
 call_cloud_diagnostics_per_stage:
--- a/docs/src/tracers.md
+++ b/docs/src/tracers.md
@@ -1,13 +1,13 @@
 # Aerosols
 
-`ClimaAtmos` implements two modes for ozone profiles. These are only relevant
-for the radiation transfer, and only when RRTMGP is used.
+# Trace Gases
+`ClimaAtmos` implements two modes for each ozone and carbon dioxide: one time varying and one time invariant. These are only relevant for the radiation transfer, and only when RRTMGP is used. All other atmospheric gases are held fixed with default values from RRTMPG that can be changed in the toml file.
 
-### Idealized Ozone Profile
+### Time Invariant Ozone Profile
 
-The `IdealizedOzone` type of ozone uses the `idealized_ozone` function to
-compute an idealized ozone profile based on the work of `Wing2018`. This
-profile is not time-varying.
+The time invariant type of ozone uses the `idealized_ozone` function to
+compute an idealized ozone profile based on the work of `Wing2018`.
+This option is default.
 
 The `idealized_ozone` function returns the ozone concentration in volume mixing
 ratio (VMR) at a given altitude `z`.
@@ -26,34 +26,28 @@ ozone = ClimaAtmos.idealized_ozone.(z)
 lines(ozone, z)
 ```
 
-### Prescribed Ozone Profile
+### Time Varying Ozone Profile
 
-The `PrescribedOzone` type of ozone uses CMIP6 forcing data to prescribe ozone
+The time varying ozone profile uses CMIP6 forcing data to prescribe ozone
 as read from files. A high-resolution, multi-year file is available in the
 `ozone_concentrations` artifact. This file is not small, so you have to obtain
 independently. Please, refer to `ClimaArtifacts` for more information. If the
 file is not found, a low-resolution, single-year version is used. This is not
-advised for production simulations.
+advised for production simulations. This option is enabled with by adding `"O3"`
+to the `time_varying_gases` config argument list, ie: `time_varying_gases: ["O3"]`.
 
 We interpolate the data from file in time every time radiation is called. The
 interpolation used is the `LinerPeriodFilling` from `ClimaUtilities`. This is a
 linear period-aware interpolation that preserves the annual cycle.
 
-### Prescribed CO2 Profile
+### Time Invariant CO2
 
-In addition to ozone, `ClimaAtmos` can prescribe CO2 concentration using data
+By default, CO2 concentrations are set to 397.547 ppm. The number can be altered
+by changing the `CO2_fixed_value` parameter in the toml file.
+
+### Time Varying CO2
+
+`ClimaAtmos` can prescribe CO2 concentration using data
 from [Mauna Loa CO2 measurements](https://gml.noaa.gov/ccgg/trends/data.html).
-This option is enabled with `MaunaLoaCO2`. Alternatively, `FixedCO2`
-utilizes a constant value that can be prescribed (by default, 397.547 ppm).
-
-### More docstrings
-
-```@docs
-ClimaAtmos.AbstractOzone
-ClimaAtmos.IdealizedOzone
-ClimaAtmos.PrescribedOzone
-
-ClimaAtmos.AbstractCO2
-ClimaAtmos.FixedCO2
-ClimaAtmos.MaunaLoaCO2
-```
+This option is enabled with by adding `"CO2"` to the `time_varying_gases`
+config argument list, ie: `time_varying_gases: ["CO2"]`.
--- a/src/cache/tracer_cache.jl
+++ b/src/cache/tracer_cache.jl
@@ -4,8 +4,7 @@ import ClimaUtilities.TimeVaryingInputs
 import ClimaUtilities.TimeVaryingInputs: TimeVaryingInput, LinearInterpolation
 import Interpolations as Intp
 
-ozone_cache(_, _, _) = (;)
-function ozone_cache(::PrescribedOzone, Y, start_date)
+function ozone_cache(Y, start_date)
     o3 = similar(Y.c.ρ)
     extrapolation_bc = (Intp.Periodic(), Intp.Flat(), Intp.Flat())
     prescribed_o3_timevaryinginput = TimeVaryingInput(
@@ -20,8 +19,7 @@ function ozone_cache(::PrescribedOzone, Y, start_date)
     return (; o3, prescribed_o3_timevaryinginput)
 end
 
-co2_cache(_, _, _) = (;)
-function co2_cache(::MaunaLoaCO2, Y, start_date)
+function co2_cache(Y, start_date)
     FT = Spaces.undertype(axes(Y.c))
     # ClimaUtilities < v0.1.21 can only write to Arrays that are on the same
     # device as the space
@@ -57,7 +55,7 @@ function co2_cache(::MaunaLoaCO2, Y, start_date)
     return (; co2, prescribed_co2_timevaryinginput)
 end
 
-function tracer_cache(Y, atmos, prescribed_aerosol_names, start_date)
+function tracer_cache(Y, prescribed_aerosol_names, time_varying_trace_gases, start_date)
     if !isempty(prescribed_aerosol_names)
         target_space = axes(Y.c)
 
@@ -104,7 +102,9 @@ function tracer_cache(Y, atmos, prescribed_aerosol_names, start_date)
     else
         aerosol_cache = (;)
     end
-    o3_cache = ozone_cache(atmos.ozone, Y, start_date)
-    co2_cache_nt = co2_cache(atmos.co2, Y, start_date)
+
+    o3_cache = :O3 ∈ Symbol.(time_varying_trace_gases) ? ozone_cache(Y, start_date) : (;)
+    co2_cache_nt = :CO2 ∈ Symbol.(time_varying_trace_gases) ? co2_cache(Y, start_date) : (;)
+
     return (; aerosol_cache..., o3_cache..., co2_cache_nt...)
 end
--- a/src/parameterized_tendencies/radiation/radiation.jl
+++ b/src/parameterized_tendencies/radiation/radiation.jl
@@ -39,18 +39,6 @@ radiation_tendency!(Yₜ, Y, p, t, ::Union{Nothing, HeldSuarezForcing}) = nothin
 ##### RRTMGP Radiation
 #####
 
-#########
-# Ozone #
-#########
-
-function center_vmr_o3(::IdealizedOzone, ᶜz)
-    ᶜvolume_mixing_ratio_o3_field = idealized_ozone.(ᶜz)
-    return Fields.field2array(ᶜvolume_mixing_ratio_o3_field)
-end
-
-# Initialized in callback
-center_vmr_o3(::PrescribedOzone, _) = NaN
-
 """
     idealized_ozone(z::FT)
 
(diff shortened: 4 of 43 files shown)


config/default_configs/default_config.yml
```diff
@@ -334,18 +334,15 @@ deep_atmosphere:
...
-  help: "What CO2 concentration to use for RRTGMP. When fixed, it is set to 397.547 parts per million. Otherwise, it is read from the MaunaLoa measuraments. [`nothing` (default), `Fixed`, `MaunaLoa`]"
-  value: ~
 detect_restart_file:
   help: "When true, try finding a restart file and use it to restart the simulation. Only works with ActiveLink."
   value: false
 prescribed_aerosols:
   help: "Which aerosols to add. List of keys from the data file (e.g., CB1, CB2)."
   value: []
+time_varying_trace_gases:
+  help: "Trace gases that vary in time. Currently, only \"CO2\" and \"O3\" are supported. All other trace gases set fixed to default values from ClimaParams."
```
trontrytel: Could you add to the help string an example on how to set this flag correctly?

src/parameterized_tendencies/radiation/radiation.jl
```diff
@@ -164,133 +142,139 @@ function rrtmgp_model_kwargs(
...
-                kwargs = (;
-                    kwargs...,
-                    center_cloud_liquid_water_path = NaN, # initialized in callback
-                    center_cloud_ice_water_path = NaN, # initialized in callback
-                    center_cloud_fraction = NaN, # initialized in callback
-                    center_cloud_liquid_effective_radius = NaN, # initialized in callback
-                    center_cloud_ice_effective_radius = NaN, # initialized in callback
-                )
-            end
+    center_volume_mixing_ratio_o3 =
```
trontrytel, line 138: We don't need a corresponding one for CO2?

src/parameterized_tendencies/radiation/radiation.jl
```diff
@@ -164,133 +142,139 @@ function rrtmgp_model_kwargs(
...
+        center_volume_mixing_ratio_h2o = NaN, # initialize in tendency
+        center_relative_humidity = NaN, # initialized in callback
+        center_volume_mixing_ratio_o3,
+        NamedTuple{atmospheric_gas_vmr_names}(atmospheric_gas_vmrs)...,
+    )
 
-        if aerosol_radiation
+    if !(radiation_mode isa RRTMGPI.ClearSkyRadiation)
+        kwargs = (; kwargs..., ice_roughness = 2)
+        if radiation_mode.idealized_clouds # icy cloud on top and wet cloud on bottom
```
trontrytel:
```suggestion
        if radiation_mode.idealized_clouds # icy cloud on top and liquid cloud on bottom
```

src/parameterized_tendencies/radiation/radiation.jl
oalcabes: Nope! o3 is just special bc it has vertical structure

src/cache/tracer_cache.jl
```diff
@@ -104,7 +102,18 @@ function tracer_cache(Y, atmos, prescribed_aerosol_names, start_date)
...
-    return (; aerosol_cache..., o3_cache..., co2_cache_nt...)
+
+    if :o3 in Symbol.(lowercase.(time_varying_trace_gases))
+        o3_cache = ozone_cache(Y, start_date)
+    else
+        o3_cache = (;)
+    end
+
+    if :co2 in Symbol.(lowercase.(time_varying_trace_gases))
+        co2_cache = co2_cache(Y, start_date)
```
szy21: I forgot why I need to name it `co2_cache_nt` before, but maybe there was some issue when you have the same name as the function.

src/parameterized_tendencies/radiation/radiation.jl
```diff
@@ -90,20 +78,8 @@ function idealized_ozone(z::FT) where {FT}
...
-#######
-
-center_vmr_co2(co2::FixedCO2) = co2.value
-
-# Initialized in callback
-center_vmr_co2(::MaunaLoaCO2) = NaN
-
 function rrtmgp_model_kwargs(
     space,
-    radiation_mode::RRTMGPI.GrayRadiation,
```
szy21, line 104: why is radiation_mode removed here?

src/cache/tracer_cache.jl
oalcabes: good catch! perhaps this will fix a bug I'm getting

src/parameterized_tendencies/radiation/radiation.jl
oalcabes: It just wasn't being used, I added it in an if statement later on bc I thought that might be cleaner. Happy to change it back if this way is better, it just was requiring me to pass a bunch of stuff that function didn't need

szy21: Oh I see you changed how `rrtmgp_model_kwargs` is called. It's ok then.

src/parameters/create_parameters.jl
```diff
@@ -96,6 +100,7 @@ function ClimaAtmosParameters(
         parameters...,
         thermodynamics_params,
         rrtmgp_params,
+        atmospheric_gas_params,
```
szy21: maybe call it trace_gas_params (although technically o2 and n2 are not trace gases)? atmospheric seems redundant here.

oalcabes: yeah I originally had trace_gas_params but then changed it bc of o2 and n2. can change it back if you think that's fine

src/cache/tracer_cache.jl
```diff
@@ -104,7 +102,18 @@ function tracer_cache(Y, atmos, prescribed_aerosol_names, start_date)
...
+    else
+        o3_cache = (;)
+    end
+
+    if :co2 in Symbol.(lowercase.(time_varying_trace_gases))
+        co2_cache_nt = co2_cache(Y, start_date)
+    else
+        co2_cache_nt = (;)
+    end
+
```
szy21: These will make a couple of configs (with typos) works: e.g. co2, CO2, Co2, cO2. Is that what you want? In general, I prefer it to be more strict, so it is more clear to the users when they set up something wrong.

oalcabes: I thought it could be nice, but if you're normally more strict can keep it that way for consistency

szy21: Ah, it's ok if you prefer this way. Up to you.

config/default_configs/default_config.yml
```diff
@@ -334,18 +334,15 @@ deep_atmosphere:
...
-  help: "What CO2 concentration to use for RRTGMP. When fixed, it is set to 397.547 parts per million. Otherwise, it is read from the MaunaLoa measuraments. [`nothing` (default), `Fixed`, `MaunaLoa`]"
-  value: ~
 detect_restart_file:
   help: "When true, try finding a restart file and use it to restart the simulation. Only works with ActiveLink."
   value: false
 prescribed_aerosols:
   help: "Which aerosols to add. List of keys from the data file (e.g., CB1, CB2)."
   value: []
+time_varying_trace_gases:
+  help: "Trace gases that vary in time, e.g. [\"CO2\", \"O3\"]. Currently, only \"CO2\" and \"O3\" are supported. All other trace gases set fixed to default values from ClimaParams."
```
haakon-e:
```suggestion
time_varying_trace_gases:
  help: 'Trace gases that vary in time, e.g. ["CO2", "O3"]. Currently, only "CO2" and "O3" are supported. All other trace gases set fixed to default values from ClimaParams.'
```

src/parameters/create_parameters.jl
```diff
@@ -257,6 +262,29 @@ function prescribed_aerosol_parameters(toml_dict)
...
+        :CFC22_fixed_value => :CFC22_fixed_value,
+        :HFC143A_fixed_value => :HFC143A_fixed_value,
+        :HFC125_fixed_value => :HFC125_fixed_value,
+        :HFC23_fixed_value => :HFC23_fixed_value,
+        :HFC32_fixed_value => :HFC32_fixed_value,
+        :HFC134A_fixed_value => :HFC134A_fixed_value,
+        :CF4_fixed_value => :CF4_fixed_value,
+        :NO2_fixed_value => :NO2_fixed_value,
+    )
+    return CP.get_parameter_values(toml_dict, name_map, "ClimaAtmos")
```
haakon-e:
```suggestion
    names = (
        "CO2_fixed_value", "N2O_fixed_value", "CO_fixed_value", "CH4_fixed_value",
        "O2_fixed_value", "N2_fixed_value", "CF4_fixed_value", "NO2_fixed_value",
        "CCL4_fixed_value", "CFC11_fixed_value", "CFC12_fixed_value", "CFC22_fixed_value",
        "HFC143A_fixed_value", "HFC125_fixed_value", "HFC23_fixed_value", "HFC32_fixed_value",
        "HFC134A_fixed_value",
    )
    return CP.get_parameter_values(toml_dict, names, "ClimaAtmos")
```
It looks like you're not actually renaming any of the parameters, so you can just give them as a list.

config/default_configs/default_config.yml
haakon-e: I think it would be helpful to highlight that setting (or not setting) this configuration option is only relevant when using RRTMGP

docs/src/tracers.md
```diff
@@ -1,13 +1,13 @@
...
+`ClimaAtmos` implements two modes for each ozone and carbon dioxide: one time varying and one time invariant. These are only relevant for the radiation transfer, and only when RRTMGP is used. All other atmospheric gases are held fixed with default values from RRTMPG that can be changed in the toml file.
 
-### Idealized Ozone Profile
+### Time Invariant Ozone Profile
 
-The `IdealizedOzone` type of ozone uses the `idealized_ozone` function to
-compute an idealized ozone profile based on the work of `Wing2018`. This
-profile is not time-varying.
+The time invariant type of ozone uses the `idealized_ozone` function to
+compute an idealized ozone profile based on the work of `Wing2018`.
```
haakon-e, line 9: add xref?

docs/src/tracers.md
```diff
@@ -26,34 +26,28 @@ ozone = ClimaAtmos.idealized_ozone.(z)
...
-advised for production simulations.
+advised for production simulations. This option is enabled with by adding `"O3"`
+to the `time_varying_gases` config argument list, ie: `time_varying_gases: ["O3"]`.
 
 We interpolate the data from file in time every time radiation is called. The
 interpolation used is the `LinerPeriodFilling` from `ClimaUtilities`. This is a
 linear period-aware interpolation that preserves the annual cycle.
 
-### Prescribed CO2 Profile
+### Time Invariant CO2 Profile
```
haakon-e: not a profile

src/cache/tracer_cache.jl
```diff
@@ -104,7 +102,18 @@ function tracer_cache(Y, atmos, prescribed_aerosol_names, start_date)
...
+        o3_cache = (;)
+    end
+
+    if :CO2 in Symbol.(time_varying_trace_gases)
+        co2_cache_nt = co2_cache(Y, start_date)
+    else
+        co2_cache_nt = (;)
+    end
+
     return (; aerosol_cache..., o3_cache..., co2_cache_nt...)
```
haakon-e:
```suggestion
    o3_cache =  "O3" ∈ time_varying_trace_gases ? ozone_cache(Y, start_date) : (;)
    co2_cache = "CO2" ∈ time_varying_trace_gases ? co2_cache(Y, start_date) : (;)
    return (; aerosol_cache..., o3_cache..., co2_cache...)
```
simpler variant?

src/cache/tracer_cache.jl
```diff
@@ -104,7 +102,18 @@ function tracer_cache(Y, atmos, prescribed_aerosol_names, start_date)
     else
         aerosol_cache = (;)
     end
-    o3_cache = ozone_cache(atmos.ozone, Y, start_date)
-    co2_cache_nt = co2_cache(atmos.co2, Y, start_date)
+
+    if :O3 in Symbol.(time_varying_trace_gases)
```
haakon-e: It might be good to add a check in `check_case_consistency` to verify that no other tracers than `O3` and `CO2` are present in `time_varying_trace_gases`, since that's the only which are supported for now.

src/parameterized_tendencies/radiation/radiation.jl
```diff
@@ -141,15 +117,17 @@ end
...
     radiation_mode::RRTMGPI.AbstractRRTMGPMode,
-    ozone::AbstractOzone,
-    co2::AbstractCO2,
     include_z::Bool,
 )
     ᶜspace = Spaces.center_space(space)
     ᶠspace = Spaces.face_space(space)
     FT = Spaces.undertype(space)
 
+    trace_gas_params = CAP.trace_gas_params(params)
```
haakon-e: could also just pass `trace_gas_params` into `rrtmgp_model_kwargs` directly, since `params` isn't used for anything else in this function.

src/parameterized_tendencies/radiation/radiation.jl
```diff
@@ -164,133 +142,142 @@ function rrtmgp_model_kwargs(
     end
```
haakon-e:
```suggestion

```
latitude is no longer used, remove this block & the definition of `bottom_coords`

src/parameterized_tendencies/radiation/radiation.jl
```diff
@@ -164,133 +142,142 @@ function rrtmgp_model_kwargs(
...
         end
+    end
+    kwargs = (;
+        kwargs...,
+        use_global_means_for_well_mixed_gases = true,
+        center_volume_mixing_ratio_h2o = NaN, # initialize in tendency
+        center_relative_humidity = NaN, # initialized in callback
+        center_volume_mixing_ratio_o3,
+        NamedTuple{trace_gas_vmr_names}(trace_gas_vmrs)...,
+    )
```
haakon-e, line 180:
```suggestion
    trace_gas_names = (
        "CO2", "N2O", "CO", "CH4", "O2", "N2", "O3",
        "CCL4", "CFC11", "CFC12", "CFC22",
        "HFC143A", "HFC125", "HFC23", "HFC32", "HFC134A",
        "CF4", "NO2",
    )
    trace_gas_vmr_names = @. Symbol(:volume_mixing_ratio_, lowercase(trace_gas_names))
    trace_gas_vmrs = map(trace_gas_names) do gas_name
        if gas_name in time_varying_trace_gases
            NaN
        elseif gas_name == "O3"
            Fields.field2array(idealized_ozone.(ᶜz))
        else
            getfield(trace_gas_params, Symbol(gas_name, :_fixed_value))
        end
    end
    kwargs = (;
        kwargs...,
        use_global_means_for_well_mixed_gases = true,
        center_volume_mixing_ratio_h2o = NaN, # initialize in tendency
        center_relative_humidity = NaN, # initialized in callback
        NamedTuple{trace_gas_vmr_names}(trace_gas_vmrs)...,
    )
```
a bit more concise suggestion that also treats ozone together with all other trace gases.

src/parameterized_tendencies/radiation/radiation.jl
haakon-e: If you remove the `radiation_mode` argument, I think you should change the name of the function to e.g. `rrtmgp_model_kwargs_gray_radiation` or something similar, and ideally add a brief docstring or comment.

This function has no documentation, so without the hint from dispatch, it's not at all clear in what circumstances this function is called.

src/parameterized_tendencies/radiation/radiation.jl
```diff
@@ -344,8 +330,22 @@ function radiation_model_cache(
...
+    else
+        kwargs =
+            rrtmgp_model_kwargs(
+                axes(Y.c),
+                params,
+                time_varying_trace_gas_names,
+                radiation_mode,
+                include_z,
+            )
+    end
```
haakon-e, line 333:
```suggestion
    if radiation_mode isa RRTMGPI.GrayRadiation
        kwargs = rrtmgp_model_kwargs(axes(Y.c), include_z)
    else
        kwargs = rrtmgp_model_kwargs(
            axes(Y.c), params, time_varying_trace_gas_names, radiation_mode, include_z,
        )
    end
```
a bit less verbose formatting

src/solver/type_getters.jl
```diff
@@ -46,27 +46,11 @@ function get_atmos(config::AtmosConfig, params)
...
-    co2 = get_co2(parsed_args)
-    with_rrtmgp = radiation_mode isa RRTMGPI.AbstractRRTMGPMode
-    if with_rrtmgp && isnothing(co2)
-        @warn (
-            "co2_model set to nothing with an RRTMGP model. Resetting to FixedCO2"
-        )
-        co2 = FixedCO2()
-    end
-    (!isnothing(co2) && !with_rrtmgp) &&
-        @warn ("$(co2) does nothing if RRTMGP is not used")
```
haakon-e, line 68: It would be good to add this check to `check_case_consistency`, warning or erroring if you specify co2 or o3 in `time_varying_trace_gases` if RRTMGP is not enabled

src/solver/types.jl
```diff
@@ -109,70 +109,6 @@ struct RCEMIPIIInsolation <: AbstractInsolation end
...
-
-"""
-    MuanaLoaCO2
-
-Implement a time-varying CO2 profile as read from disk.
-
-The data from the Mauna Loa CO2 measurements is used. It is a assumed that the
-concentration is constant.
-"""
-struct MaunaLoaCO2 <: AbstractCO2 end
```
haakon-e, line 174: deleting code, yay!

src/parameterized_tendencies/radiation/radiation.jl
oalcabes: I am wary of using strings here as I believe things need to be symbols. Ozone has a different input structure into kwargs so I can't make it nice like this unfortunately

haakon-e: If there's reason to use symbols (I'm not aware of why) then I'd suggest to convert the `time_varying_trace_gases` list to a list of tuples here, to be consistent:
```julia
function get_tracers(parsed_args)
    aerosol_names = Tuple(parsed_args["prescribed_aerosols"])
    time_varying_trace_gas_names = Tuple(Symbol.(parsed_args["time_varying_trace_gases"]))
    return (; aerosol_names, time_varying_trace_gas_names)
end
```

PR #4158, opened 2025-12-11
Clean up the cloud fraction computation

trontrytel:
This PR:
- Removes the  `quadrature_sgs` option for computing cloud fraction. The remaining options are `grid_scale` and `quadrature` (default). The functionality of `quadrature_sgs` is retained via `quadrature` option and the `turbconv_model` choice. I.e. if the user is running with a `turbconv_model` and `quadrature` cloud, they will get the `quadrature_sgs` behavior.
- Deletes the multiple dispatch options for `set_cloud_fraction!` that were based on `turbconv_model`. I think this was duplicating the code for doing the quadratures and adding updraft contributions. To me the flow for all `quadrature` options is similar. The small differences arising from assuming environment area to be one and storing updraft properties in `precomputed` vs `state` structs can be handled by `if/else` within one function. AFAIK they will be handled at compile time and should not affect the performance. 

I kept the quadrature summation separate, even though it is only used once. Just in case we will want to re-use it in other parts of the code.

### DIFF
--- a/src/cache/cloud_fraction.jl
+++ b/src/cache/cloud_fraction.jl
@@ -5,7 +5,7 @@ import ClimaCore.RecursiveApply: rzero, ⊞, ⊠
 """
     Helper function to populate the cloud diagnostics named tuple
 """
-function make_named_tuple(t1, t2, t3)
+function make_cloud_fraction_named_tuple(t1, t2, t3)
     return NamedTuple{(:cf, :q_liq, :q_ice)}(tuple(t1, t2, t3))
 end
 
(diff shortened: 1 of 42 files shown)


trontrytel:
Not sure why the restart tests are failing. Will look into it tomorrow.

src/cache/precomputed_quantities.jl
```diff
@@ -529,13 +529,14 @@ NVTX.@annotate function set_explicit_precomputed_quantities!(Y, p, t)
...
+    # The buoyancy gradient depends on the cloud fraction, and the cloud fraction
+    # depends on the mixing length, which depends on the buoyancy gradient.
+    # We break this circular dependency by using cloud fraction from the previous time step in the
+    # buoyancy gradient calculation. This breaks reproducible restart in general,
     # but we support reproducible restart with grid-scale cloud by recalculating the cloud fraction here.
-    if p.atmos.cloud_model isa GridScaleCloud
-        set_cloud_fraction!(Y, p, p.atmos.moisture_model, p.atmos.cloud_model)
+    if p.atmos.cloud_model isa GridScaleCloud ||
+       p.atmos.cloud_model isa QuadratureCloud
+        set_cloud_fraction!(Y, p, moisture_model, cloud_model)
```
Julians42: I don't understand the change here? For all methods don't we need to set cloud fraction and so why previously is this call just for grid scale clouds? 

Second thought is that if we need this for all methods, would it be best to just check is `p.atmos.cloud_model` exists as opposed to an `or` statement for each model type? (i.e., we want this for all moist simulations)

src/cache/cloud_fraction.jl
```diff
@@ -302,32 +129,126 @@ function quad_loop(
...
+            make_cloud_fraction_named_tuple(
+                ifelse(TD.has_condensate(thermo_params, ᶜts), 1, 0),
+                TD.PhasePartition(thermo_params, ᶜts).liq,
+                TD.PhasePartition(thermo_params, ᶜts).ice,
+            )
+    else
+        q_liq = @. lazy(specific(Y.c.ρq_liq, Y.c.ρ))
+        q_ice = @. lazy(specific(Y.c.ρq_ice, Y.c.ρ))
+        @. p.precomputed.cloud_diagnostics_tuple =
+            make_cloud_fraction_named_tuple(ifelse(q_liq + q_ice > 0, 1, 0), q_liq, q_ice)
```
Julians42: cf will be stored in the tuple as int. This is currently breaking CI. You'll need to unpack `FT` as well
```suggestion
            make_cloud_fraction_named_tuple(ifelse(q_liq + q_ice > 0, FT(1), FT(0)), q_liq, q_ice)
```

src/cache/cloud_fraction.jl
```diff
@@ -302,32 +129,126 @@ function quad_loop(
...
+)
+    thermo_params = CAP.thermodynamics_params(p.params)
+    diagnostic_covariance_coeff = CAP.diagnostic_covariance_coeff(p.params)
+    turbconv_model = p.atmos.turbconv_model
+
+    if turbconv_model isa PrognosticEDMFX
+        local_ts = p.precomputed.ᶜts⁰
+    else
+        local_ts = p.precomputed.ᶜts
+    end
```
Julians42: I find these differences in cache thermal state annoying. (not a statement about this pr)

src/cache/cloud_fraction.jl
```diff
@@ -302,32 +129,126 @@ function quad_loop(
...
+
+    # TODO - tmp fix for the Coupler test. To be removed soon.
+    ᶜmixing_length_field = p.scratch.ᶜtemp_scalar
+    ᶜmixing_length_field .= compute_gm_mixing_length(Y, p)
```
Julians42: Reminder to address todo. We should use `ᶜmixing_length(Y, p)` since this is the most general form

src/cache/cloud_fraction.jl
trontrytel: Yeah I agree. Some of them are more necessary than others - like weather variables are part of the state vector or cached. But here the code would be much cleaner if we used the same name in prognostic EDMF and everywhere else.

Julians42 (commented):
Thanks for simplifying the cloud fraction options. Left a few comments but should be straightforward to address.

src/cache/cloud_fraction.jl
szy21: This is not used anywhere?

src/cache/cloud_fraction.jl
szy21:
```suggestion
    ᶜts = turbconv_model isa PrognosticEDMFX ? p.precomputed.ᶜts⁰ : p.precomputed.ᶜts
```
I just think this is a bit nicer and consistent with other part of the codebase. And I kept the superscript in the variable name for the convention we have for fields.

src/cache/cloud_fraction.jl
szy21: I'm not sure what this TODO means. but I think we should call `ᶜmixing_length(Y, p)` if `turbconv_model isa AbstractEDMF`, and `compute_gm_mixing_length(Y, p)` if not?

src/cache/cloud_fraction.jl
```diff
@@ -302,32 +129,126 @@ function quad_loop(
...
+    # ... and add contributions from the updrafts if using EDMF.
+    if turbconv_model isa PrognosticEDMFX || turbconv_model isa DiagnosticEDMFX
+        n = n_mass_flux_subdomains(turbconv_model)
+        (; ᶜρʲs, ᶜtsʲs) = p.precomputed
+        for j in 1:n
+            if turbconv_model isa PrognosticEDMFX
+                local_draft_ρa = @. lazy(Y.c.sgsʲs.:($$j).ρa)
+            else
+                local_draft_ρa = @. lazy(p.precomputed.:($$j).ᶜρaʲs)
+            end
```
szy21:
```suggestion
ᶜρaʲ =
                turbconv_model isa PrognosticEDMFX ? Y.c.sgsʲs.:($j).ρa :
                p.precomputed.ᶜρaʲs.:($j)
```
(same as the previous comment)

src/cache/precomputed_quantities.jl
szy21: We probably don't need to change here - does it break restart if you don't change it?

szy21 (commented):
Thanks for cleaning up the code! I left a few comments. I think this should be good after addressing the comments but I'm happy to take a look again if needed.

src/cache/precomputed_quantities.jl
trontrytel: Yeah, I came here because I was being flagged by a restart test. But I changed a couple of things afterwards, so I'll try again.

src/cache/cloud_fraction.jl
```diff
@@ -296,38 +123,131 @@ function quad_loop(
...
+            @. p.precomputed.ᶜgradᵥ_θ_liq_ice =
+                ᶜgradᵥ(ᶠinterp(TD.liquid_ice_pottemp(thermo_params, local_ts)))
+        end
+    end
+
+    # Compute SGS cloud fraction diagnostics based on environment quadrature points ...
+    @. p.precomputed.cloud_diagnostics_tuple =
+        compute_cloud_fraction_quadrature_diagnostics(
+            qc.SG_quad,
+            local_ts,
```
szy21:
```suggestion
           ᶜts,
```

src/cache/cloud_fraction.jl
```diff
@@ -296,38 +123,131 @@ function quad_loop(
...
+            thermo_params,
+        )
+    # ... weight by environment area fraction if using PrognosticEDMFX (assumed 1 otherwise) ...
+    if turbconv_model isa PrognosticEDMFX
+        ᶜρa⁰ = @. lazy(ρa⁰(Y.c.ρ, Y.c.sgsʲs, p.atmos.turbconv_model))
+        @. p.precomputed.cloud_diagnostics_tuple *= NamedTuple{(:cf, :q_liq, :q_ice)}(
+            tuple(
+                draft_area(ᶜρa⁰, TD.air_density(thermo_params, local_ts)),
+                draft_area(ᶜρa⁰, TD.air_density(thermo_params, local_ts)),
+                draft_area(ᶜρa⁰, TD.air_density(thermo_params, local_ts)),
```
szy21:
```suggestion
                draft_area(ᶜρa⁰, TD.air_density(thermo_params, ᶜts)),
                draft_area(ᶜρa⁰, TD.air_density(thermo_params, ᶜts)),
                draft_area(ᶜρa⁰, TD.air_density(thermo_params, ᶜts)),
```

src/cache/precomputed_quantities.jl
```diff
@@ -529,13 +529,13 @@ NVTX.@annotate function set_explicit_precomputed_quantities!(Y, p, t)
...
-    # The buoyancy gradient depends on the cloud fraction, and the cloud fraction 
-    # depends on the mixing length, which depends on the buoyancy gradient. 
-    # We break this circular dependency by using cloud fraction from the previous time step in the 
-    # buoyancy gradient calculation. This breaks reproducible restart in general, 
+    # The buoyancy gradient depends on the cloud fraction, and the cloud fraction
+    # depends on the mixing length, which depends on the buoyancy gradient.
+    # We break this circular dependency by using cloud fraction from the previous time step in the
+    # buoyancy gradient calculation. This breaks reproducible restart in general,
     # but we support reproducible restart with grid-scale cloud by recalculating the cloud fraction here.
     if p.atmos.cloud_model isa GridScaleCloud
```
szy21:
```suggestion
    if cloud_model isa GridScaleCloud
```

tapios:
Looks basically good to me. Just one question on design:

Can we separate the calculation of the (co-)variances for q and \theta_l from the cloud fraction calculation more cleanly? That is, compute covariances separately, then use them (rather than gradients) as inputs for CF calculations? The advantage would be to avoid code duplication when, for example, we use an ML CF that depends on covariances.

tapios (approved):

trontrytel:
> Looks basically good to me. Just one question on design:
> 
> Can we separate the calculation of the (co-)variances for q and \theta_l from the cloud fraction calculation more cleanly? That is, compute covariances separately, then use them (rather than gradients) as inputs for CF calculations? The advantage would be to avoid code duplication when, for example, we use an ML CF that depends on covariances.

Yes! I was planning to do it in a separate PR and also add an option to have a user prescribed value there. Kind of a first step towards replacing  it with data driven model.

PR #4162, opened 2025-12-12
add an option for reproducible restart

szy21:
## Purpose 

Previously, we only supported reproducible restart with grid scale cloud. This PR adds an option so that for reproducible restart, we still use partial cloud for radiation, but calculate grid scale cloud for buoyancy gradient. I think this is better. And we can use this option if we have similar problems in the future.

## Content

----

### DIFF
--- a/docs/src/restarts.md
+++ b/docs/src/restarts.md
@@ -26,8 +26,12 @@ particularly useful for
 
 !!! note
 
-    The simulation can only be restarted in a reproducible way when using `GridScaleCloud`
-    for `cloud_model`.
+    By default, the simulation cannot be restarted in a reproducible way. To
+    enable reproducible restarts, you need to set `reproducible_restart` to `true`. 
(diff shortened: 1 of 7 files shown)


docs/src/restarts.md
```diff
@@ -26,8 +26,12 @@ particularly useful for
 
 !!! note
 
-    The simulation can only be restarted in a reproducible way when using `GridScaleCloud`
-    for `cloud_model`.
+    By default, the simulation cannot be restarted in a reproducible way. To
+    enable reproducible restarts, you need to set `reproducible_restart` to `true`. 
+    When `reproducible_restart` is true, `ClimaAtmos` recalculates the grid_scale 
+    cloud fraction and use it in the buoyancy gradient calculation to ensure deterministic
```
tapios:
```suggestion
    cloud fraction and uses it in the buoyancy gradient calculation to ensure deterministic 
```

PR #4164, opened 2025-12-12
Revert "Refactor atmospheric gases"

szy21:
Reverts CliMA/ClimaAtmos.jl#4153, which seems to break prognostic edmf amip


szy21:
This surprisingly didn't break regression test, because I think by reverting the ref counter to the previous one (decreasing by one), it is comparing with the previous reference. I will still increase the ref counter to avoid any potential issues.

PR #4175, opened 2025-12-16
ft: add support for lazy diagnostics

haakon-e:
This pull request adds support for lazy diagnostics in ClimaAtmos, aligning with the new functionality introduced in ClimaDiagnostics.jl v0.2.13. The main changes update the way diagnostic variables are registered and documented, improving flexibility and backward compatibility.

**Diagnostics API and Documentation Improvements:**

* Refactored the `add_diagnostic_variable!` function to accept both `compute` (new, 3-argument) and `compute!` (legacy, 4-argument) keyword arguments for defining diagnostic computations, supporting backward compatibility with ClimaDiagnostics v0.2.13. The function signature and internal logic were updated to handle both forms and provide better warning messages when overwriting diagnostics. [[1]](diffhunk://#diff-0049530136103293ec2026f423f0cd61169c394e32e8c21a14a9641700c58af3L14-R21) [[2]](diffhunk://#diff-0049530136103293ec2026f423f0cd61169c394e32e8c21a14a9641700c58af3L43-R89)
* Expanded and clarified the documentation for diagnostic variable registration, including detailed descriptions for all keyword arguments and a note explaining backward compatibility with the new and old compute function signatures. [[1]](diffhunk://#diff-0049530136103293ec2026f423f0cd61169c394e32e8c21a14a9641700c58af3L31-R35) [[2]](diffhunk://#diff-0049530136103293ec2026f423f0cd61169c394e32e8c21a14a9641700c58af3L43-R89)

**Release Notes:**

* Added a release note announcing support for lazy diagnostics as introduced in ClimaDiagnostics.jl v0.2.13.

PR #4178, opened 2025-12-17
ft: add support for lazy diagnostics

haakon-e:
see #4175, which somehow lost all its commits


haakon-e:
@szy21 this is the same as #4175, but somehow I rebased locally onto a state that deleted all my commits then pushed.

szy21 (approved):

PR #4180, opened 2025-12-17
Improve GPU performance by pulling out some common subexpressions

daverumph:
Increment jet_test_nfailures by 2 to account for new kernels. Add two new scratch locations for subexpressions.
Make one calculation eager so that it isn't continually redone within kernels.

## Purpose 
Sedimentation and the microphysics tracers loop were affected. Three new scratch variables were introduced.

## To-do
Profile the advection kernels from manual_sparse_jacobian.jl:1041 and 1051 to see
what can be done to speed those up.

## Content
nsys profiling revealed that four kernels were taking up over 25% of the total time running
prognostic 1M. With these changes, they account for a few percent of the changes, and overall
step times dropped about 33%.
Because this is just a performance rewrite, there is no new functionality, and no new tests
are needed.

----
- [X] I have read and checked the items on the review checklist.

### DIFF
--- a/src/prognostic_equations/implicit/manual_sparse_jacobian.jl
+++ b/src/prognostic_equations/implicit/manual_sparse_jacobian.jl
@@ -481,9 +481,9 @@ function update_jacobian!(alg::ManualSparseJacobian, cache, Y, p, dtγ, t)
     @. ᶜkappa_m =
         TD.gas_constant_air(thermo_params, ᶜts) / TD.cv_m(thermo_params, ᶜts)
 
-    T = @. lazy(TD.air_temperature(thermo_params, ᶜts))
+    ᶜT = @. lazy(TD.air_temperature(thermo_params, ᶜts))
     ᶜ∂p∂ρq_tot = p.scratch.ᶜtemp_scalar_2
-    @. ᶜ∂p∂ρq_tot = ᶜkappa_m * (-e_int_v0 - R_d * T_0 - Δcv_v * (T - T_0)) + ΔR_v * T
+    @. ᶜ∂p∂ρq_tot = ᶜkappa_m * (-e_int_v0 - R_d * T_0 - Δcv_v * (ᶜT - T_0)) + ΔR_v * ᶜT
 
     if use_derivative(topography_flag)
         @. ∂ᶜK_∂ᶜuₕ = DiagonalMatrixRow(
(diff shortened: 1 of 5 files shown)


src/cache/temporary_quantities.jl
```diff
@@ -112,6 +121,7 @@ function temporary_quantities(Y, atmos)
             Y.c,
             BidiagonalMatrixRow{Adjoint{FT, C3{FT}}},
         ),
+        ᶜtridiagonal_matrix_c3 = similar(Y.c, TridiagonalMatrixRow{FT}),
```
imreddyTeja: This name is misleading. The "c3" make me think it would be a tridiagonal matrix with elements of type Covariant3Vector

src/cache/temporary_quantities.jl
```diff
@@ -92,6 +92,15 @@ function temporary_quantities(Y, atmos)
                 ClimaCore.Geometry.WVector{FT},
             },
         ),
+        ᶠsed_tracer_advection = similar(
+            Y.f,
+            ClimaCore.MatrixFields.BandMatrixRow{
+                ClimaCore.Utilities.PlusHalf{Int64}(0),
+                1,
+                ClimaCore.Geometry.WVector{FT},
+            },
```
imreddyTeja, line 102:
```suggestion
            MatrixFields.BandMatrixRow{
                half,
                1,
                ClimaCore.Geometry.WVector{FT},
            },
```

src/prognostic_equations/implicit/manual_sparse_jacobian.jl
```diff
@@ -1041,10 +1044,7 @@ function update_jacobian!(alg::ManualSparseJacobian, cache, Y, p, dtγ, t)
             )
 
                 ᶜa = (@. lazy(draft_area(Y.c.sgsʲs.:(1).ρa, ᶜρʲs.:(1))))
-                ᶜ∂a∂z =
-                    @. lazy(
-                        ᶜprecipdivᵥ(ᶠinterp(ᶜJ) / ᶠJ * ᶠright_bias(Geometry.WVector(ᶜa))),
-                    )
+                ᶜ∂a∂z = @. ᶜprecipdivᵥ(ᶠinterp(ᶜJ) / ᶠJ * ᶠright_bias(Geometry.WVector(ᶜa)))
```
imreddyTeja: This line will allocate

src/cache/temporary_quantities.jl
```diff
@@ -92,6 +92,15 @@ function temporary_quantities(Y, atmos)
...
+        ),
+        ᶜtracer_advection = similar(Y.c, BidiagonalMatrixRow{Adjoint{FT, C3{FT}}}),
```
imreddyTeja: The other MatrixFields have "_matrix_" in their name. Same comment applies to  `ᶠsed_tracer_advection`

src/cache/temporary_quantities.jl
daverumph: Changed

src/cache/temporary_quantities.jl
daverumph: I followed the same convention as the prior definition.  The result is equivalent.

src/cache/temporary_quantities.jl
daverumph: Changed as you suggest

PR #4181, opened 2025-12-17
use grid-mean u for TKE advection

szy21:
## Purpose 

I forgot to change environment u to grid-mean u when changing environment TKE to grid-mean TKE.

## Content

----


tapios:
> I forgot to change the vertical advection - now more code has been removed! I will move TKE out of sgs maybe over the holidays so we won't be confused again:)

Good idea! It's confusing to have it there.

PR #4182, opened 2025-12-18
Combine operations in constant horizontal diffusion

petebachant:
These changes reduce the slowdown from constant horizontal diffusion from ~2% to ~1% (logs [here](https://github.com/petebachant/clima-horizontal-diffusion/tree/main/.calkit/slurm/logs); see `run*.out`). I'll definitely need a review to ensure it's mathematically equivalent though.

## TODO

- [ ] Remove environment changes (I can't get a globally installed Revise to import properly)
- [ ] Rebase into one commit

### DIFF
--- a/src/parameterized_tendencies/les_sgs_models/constant_horizontal_diffusion.jl
+++ b/src/parameterized_tendencies/les_sgs_models/constant_horizontal_diffusion.jl
@@ -25,19 +25,15 @@ function horizontal_constant_diffusion_tendency!(
     ρD = @. lazy(ρ * D)
 
     # Total energy diffusion
-    ᶜh_tot = @. lazy(
-        TD.total_specific_enthalpy(
-            thermo_params,
-            ᶜts,
-            specific(Y.c.ρe_tot, ρ),
+    @. Yₜ.c.ρe_tot += wdivₕ(
+        ρD * gradₕ(
+            TD.total_specific_enthalpy(thermo_params, ᶜts, specific(Y.c.ρe_tot, ρ)),
         ),
     )
-    @. Yₜ.c.ρe_tot += wdivₕ(ρD * gradₕ(ᶜh_tot))
 
     # Tracer diffusion
     foreach_gs_tracer(Yₜ, Y) do ᶜρχₜ, ᶜρχ, ρχ_name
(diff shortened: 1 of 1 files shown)


src/parameterized_tendencies/les_sgs_models/constant_horizontal_diffusion.jl
```diff
@@ -17,25 +17,23 @@ function horizontal_constant_diffusion_tendency!(
...
 
     # Total energy diffusion
-    ᶜh_tot = @. lazy(
-        TD.total_specific_enthalpy(
-            thermo_params,
-            ᶜts,
-            specific(Y.c.ρe_tot, Y.c.ρ),
+    @. Yₜ.c.ρe_tot += wdivₕ(
+        ρD * gradₕ(
+            TD.total_specific_enthalpy(thermo_params, ᶜts, specific(Y.c.ρe_tot, ρ)),
```
szy21, line 30: cc @haakon-e . If this does improve performance, maybe it suggests we shouldn't do things like this in one of your PRs?
```
ᶜe_tot = @. lazy(...)
ᶜh_tot = @. lazy(...)
```

haakon-e: It'd be a shame if `@. lazy(...)` isn't "free" as I have long thought. 

It'd be good to understand this a bit more. I see that this removes
```julia
ᶜh_tot = @. lazy(...)
```
but adds
```julia
ρD = @. lazy(ρ * D)
```
If it indeed is the case that the former hurts performance but the latter is totally fine, then I worry it'll be quite the challenge to write reasonably performant, let alone readable, code.

----

Even more surprising is the appearance of 
```julia
ρD = @. lazy(ρ * D)
```
but deletion of 
```julia
ᶜχ = @. lazy(specific(ᶜρχ, Y.c.ρ))
```
which literally just executes `ᶜρχ / Y.c.ρ`.

trontrytel: Yeah. How to guess which `lazy` is expensive? Do we have any good rule for that?

PR #4185, opened 2025-12-22
Add ml cloud fraction closure 

Julians42:
## Purpose 

adds machine learning cloud fraction. 42% error reduction in SCM calibration to LES over quadrature approach. 

performance_graph

## To-do

1. Change NN weights to be online calibrated weights (currently offline calibrated) by choosing the best particle.

2. Performance was really slow in the diagnostic case that I tested - either my allocation request on HPC was bad or performance / allocations are really bad. It was fine in a previous branch `j/mlcf` (prognostic edmfx in costa's aiwq setup in coupler was equal speed to quadrature) before rebasing to Anna's version (although I don't think any changes she made would affect performance here). 

3. There is stuff to clean up but I want to get it running ci before tomorrow. 

## Content

----


src/cache/cloud_fraction.jl
```diff
@@ -272,3 +272,130 @@ NVTX.@annotate function set_cloud_fraction!(
...
+    #dθli_dz
+    ᶜ∇θ =
+        dot.(
+            Geometry.WVector.(p.precomputed.ᶜgradᵥ_θ_liq_ice),
+            Ref(ClimaCore.Geometry.WVector(FT(1.0))),
+        )
+    # distance to saturation in temperature space 
+    θli = p.scratch.ᶜtemp_scalar_2
+    θli .= TD.liquid_ice_pottemp.(thermo_params, ᶜts)
+    delta_θli = FT(0.1)
```
tapios: Add some code comments to explain what's happening here: Approximate gradient dq_sat/dθli by one-sided finite difference.

Wouldn't it make more sense to take a negative `delta_θli`, given that saturation occurs with cooling?

src/cache/cloud_fraction.jl
```diff
@@ -272,3 +272,130 @@ NVTX.@annotate function set_cloud_fraction!(
...
+    # (; ᶜρʲs, ᶜtsʲs, cloud_diagnostics_tuple) = p.precomputed
+    # ᶜρa⁰ = @. lazy(ρa⁰(Y.c.ρ, Y.c.sgsʲs, turbconv_model))
+    FT = eltype(p.params)
+
+    # COMPUTE quantities needed to form pi groups 
+
+    # mixing length
+    ᶜmixing_length_field = p.scratch.ᶜtemp_scalar
+    ᶜmixing_length_field .= ᶜmixing_length(Y, p)
+    # distance to saturation in q space
```
tapios: It would be cleaner and clearer to move the distance to saturation calculations to their own function.

src/cache/cloud_fraction.jl
```diff
@@ -272,3 +272,130 @@ NVTX.@annotate function set_cloud_fraction!(
...
+    end
+
+    cf = p.scratch.ᶜtemp_scalar_3
+    cf .= apply_cf_nn.(Ref(cloud_ml.model), π_1, π_2, π_3, π_4)
+    #Main.@infiltrate
+    # overwrite with the ML computed cloud fraction, leaving q_liq, q_ice computed via quadrature
+    p.precomputed.cloud_diagnostics_tuple.cf .= cf
+
+    # ... and add contributions from the updrafts if using EDMF.
+    if turbconv_model isa PrognosticEDMFX || turbconv_model isa DiagnosticEDMFX
```
tapios: Also, better to make the updraft cloud fraction its own function (we'll need it in several places)

src/cache/cloud_fraction.jl
Julians42: Sounds good - I'll clear up the code. I was debating using centered difference since that's more accurate but its more costly (another allocation/ computation where I don't think it really matters). 

@tapios - In a cloud, saturation temperature is reached with warming so I think the reverse argument also applies in that case. If you want I can switch to minus and recalibrate? (otherwise I'll just leave as plus) I don't intuitively see which side will give less error...

src/cache/cloud_fraction.jl
```diff
@@ -272,3 +272,130 @@ NVTX.@annotate function set_cloud_fraction!(
...
+            ),
+        )
+    end
+
+    (; params) = p
+    # (; ᶜρʲs, ᶜtsʲs, cloud_diagnostics_tuple) = p.precomputed
+    # ᶜρa⁰ = @. lazy(ρa⁰(Y.c.ρ, Y.c.sgsʲs, turbconv_model))
+    FT = eltype(p.params)
+
+    # COMPUTE quantities needed to form pi groups
```
haakon-e: Most of this function except these lines forming pi groups and calling the NN seems to be identical to `set_cloud_fraction!` for `QuadratureCloud`. Perhaps it would be nicer to use that function directly, i.e.:
```Julia
NVTX.@annotate function set_cloud_fraction!(
    Y,
    p,
    ::Union{EquilMoistModel, NonEquilMoistModel},
    cloud_model::Union{QuadratureCloud, CloudML},
)
# ...

if cloud_model isa CloudML
    # overwrite with the ML computed cloud fraction, leaving q_liq, q_ice computed via quadrature
    @. p.precomputed.cloud_diagnostics_tuple.cf = set_ml_cloud_fraction(cloud_model, ᶜmixing_length_field, Y.c, thermo_params, ᶜts, p.precomputed.ᶜgradᵥ_q_tot, p.precomputed.ᶜgradᵥ_θ_liq_ice)
end

# ...
```
and then define `set_ml_cloud_fraction` as the pointwise version of the code below. This has the added benefit of avoiding all the allocations you're introducing, which may be the cause of the performance slowdown you're observing.

src/solver/model_getters.jl
```diff
@@ -357,13 +360,26 @@ function get_microphysics_model(parsed_args)
...
 
-function get_cloud_model(parsed_args)
+function get_cloud_model(parsed_args, params)
     cloud_model = parsed_args["cloud_model"]
     FT = parsed_args["FLOAT_TYPE"] == "Float64" ? Float64 : Float32
     return if cloud_model == "grid_scale"
         GridScaleCloud()
     elseif cloud_model == "quadrature"
         QuadratureCloud(SGSQuadrature(FT))
+    elseif cloud_model == "cloud_ml"
```
haakon-e: It might be better to give it a more descriptive name here, since the ML architecture choice is so specific, e.g. `Schmitt2026MLCloudFraction` or something similar.

src/solver/model_getters.jl
```diff
@@ -357,13 +360,26 @@ function get_microphysics_model(parsed_args)
...
+        nn_filepath = joinpath(
+            @clima_artifact("cloud_fraction_nn"),
+            "arch_2layers_8nodes.jld2",
```
haakon-e: could consider moving the strings `"cloud_fraction_nn"` and `"arch_2layers_8nodes.jld2"` to the config/toml file so that you can easily change them later as you refine the model.

src/solver/model_getters.jl
```diff
@@ -1,3 +1,6 @@
+using Flux
```
haakon-e, line 1: is `using Flux` needed because you do `JLD2.load` below?

Artifacts.toml
```diff
@@ -80,3 +80,11 @@ git-tree-sha1 = "789c8dd160b022a6c714a102c1495530c7a9d20c"
...
 [weather_model_ic]
 git-tree-sha1 = "1c05bc305e79ceaab8c537be4f8c4a26465996a3"
+
+
+[cloud_fraction_nn]
+git-tree-sha1 = "c2412372bdf40ba73f496dc0fe427d063e31533b"
+
+    [[cloud_fraction_nn.download]]
+    sha256 = "a0b8b682539d975c1cc331549d0222a0eda121f5b2977cc63486368a7dc07885"
+    url = "https://caltech.box.com/shared/static/vdb6r7cv5tj6uhrphn5a4352smsxdqgs.gz"
```
haakon-e:
```suggestion
    url = "https://caltech.box.com/shared/static/vdb6r7cv5tj6uhrphn5a4352smsxdqgs.gz"

```
line break

src/cache/cloud_fraction.jl
szy21: I also think something like this would be better.

src/cache/cloud_fraction.jl
```diff
@@ -272,3 +272,130 @@ NVTX.@annotate function set_cloud_fraction!(
...
+        dot.(
+            Geometry.WVector.(p.precomputed.ᶜgradᵥ_q_tot),
+            Ref(ClimaCore.Geometry.WVector(FT(1.0))),
+        )
+    #dθli_dz
+    ᶜ∇θ =
+        dot.(
+            Geometry.WVector.(p.precomputed.ᶜgradᵥ_θ_liq_ice),
+            Ref(ClimaCore.Geometry.WVector(FT(1.0))),
+        )
```
szy21: You can probably use the `projected_vector_data` function here, although it won't be much simpler.

src/cache/cloud_fraction.jl
```diff
@@ -272,3 +272,130 @@ NVTX.@annotate function set_cloud_fraction!(
...
+
+    function apply_cf_nn(model, π_1::FT, π_2::FT, π_3::FT, π_4::FT) where {FT}
+        return clamp((model(SA.SVector(π_1, π_2, π_3, π_4))[]), FT(0.0), FT(1.0))
+    end
+
+    cf = p.scratch.ᶜtemp_scalar_3
+    cf .= apply_cf_nn.(Ref(cloud_ml.model), π_1, π_2, π_3, π_4)
+    #Main.@infiltrate
+    # overwrite with the ML computed cloud fraction, leaving q_liq, q_ice computed via quadrature
+    p.precomputed.cloud_diagnostics_tuple.cf .= cf
```
szy21: This should be multiplied by ᶜρa⁰ for prognostic EDMFX if it's the environment cloud fraction (and I think it should be, as the updraft is treated separately below).

src/solver/model_getters.jl
Julians42: @haakon-e `cloud_fraction_nn` is the name of the artifact so will keep fixed, but arch_2layers_8nodes could be for example set in the config and seems like a good idea to me if we want to test different architectures down the line

src/solver/model_getters.jl
Julians42: We need it for the reconstruction of the network (just after the load were we do `network(weights)`). @imreddyTeja should we be doing `import Flux` or import it somewhere else?

src/cache/cloud_fraction.jl
```diff
@@ -272,3 +285,89 @@ NVTX.@annotate function set_cloud_fraction!(
...
+        projected_vector_data.(
+            C3,
+            p.precomputed.ᶜgradᵥ_θ_liq_ice,
+            Fields.level(Fields.local_geometry_field(Y.c)),
+        )
+
+    q_v = p.scratch.ᶜtemp_scalar_4
+    q_v .= specific.(Y.c.ρq_tot, Y.c.ρ)
+    # Saturation state at current thermodynamic state
+    q_sat = p.scratch.ᶜtemp_scalar_5
```
szy21: ᶜtemp_scalar_5 is used in ᶜmixing_length called on line 301. It is materialized, so it is probably ok to reuse it here, but it would be good to make sure. An easy way to check is to change the temp_scalar here and see if the result is the same.

src/cache/cloud_fraction.jl
```diff
@@ -272,3 +285,89 @@ NVTX.@annotate function set_cloud_fraction!(
...
+
+    # distance to saturation in temperature space 
+    Δθli, θli_sat, dqsatdθli =
+        saturation_distance(q_v, q_sat, ᶜts, θli, thermo_params, FT(0.1))
+
+    # form the pi groups 
+    π_1 = (q_sat .- q_v) ./ q_sat
+    π_2 = Δθli ./ θli_sat
+    π_3 = @. (((dqsatdθli * ᶜ∇θ - ᶜ∇q) * ᶜmixing_length_field) / q_sat)
+    π_4 = @. (ᶜ∇θ * ᶜmixing_length_field) / θli_sat
```
szy21: I think these will allocate?

src/cache/cloud_fraction.jl
```diff
@@ -272,3 +285,89 @@ NVTX.@annotate function set_cloud_fraction!(
...
+
+    cf = p.scratch.ᶜtemp_scalar_6
+    cf .= apply_cf_nn.(Ref(cloud_ml.model), π_1, π_2, π_3, π_4)
+    return cf
+end
+
+function saturation_distance(q_v, q_sat, ᶜts, θli, thermo_params, Δθli_fd)
```
szy21: I think this function allocates. Could you make it a point-wise function?

szy21: Actually, maybe it is better to make the entire set_ml_cloud_fraction point-wise, if you can. That will avoid the allocations.

Julians42: @szy21 I edited to make what I could pointwise to reduce allocations. As there isn't a function for pointwise mixing length and I couldn't find one, or didn't see how, for `projected_vector_data` I had to add an extra function to keep the original `set_cloud_fraction!` function clean. Let me know if you can think of a better way (e.g., with one function) or better naming conventions.

src/cache/cloud_fraction.jl
```diff
@@ -216,7 +218,7 @@ NVTX.@annotate function set_cloud_fraction!(
     Y,
     p,
     ::Union{EquilMoistModel, NonEquilMoistModel},
-    qc::QuadratureCloud,
+    qc::Union{QuadratureCloud, MLCloud},
```
tapios, line 221: Why do we have both Schmitt2026ML and MLCloud? Seems confusing.

tapios (approved):
LGTM

src/cache/cloud_fraction.jl
Julians42: Good point - I agree it's currently not consistent with the naming convention for the other cloud fraction parameterizations. I think the most consistent is snake case without "cloud" in the name, so just dispatching on `cloud_model: ml`. My only concern, as @haakon-e pointed out was that we may have other ml models (@trontrytel was going to work on one for the covariance) so then it becomes ambiguous and since the architecture / pi groups will be specific to a paper I write next year. We could leave it simple for now and just put as `ml` and try and disambiguate when another ml cloud case arises? Thoughts?

tapios: I'll leave it up to you how to handle this. As a general rule, I think we should avoid naming parameterizations after people, because it leads to hesitancy downstream to change the parameterizations--a stasis I have seen many times that is not healthy. But I have no objection and strong thoughts how to handle it here. It may help to make clearer that only cloud fraction is ML--in principle other quantities (condensate path, effective radius etc.) could be too. So something like MLCloudFraction may work. But again: your call. it's not crucially important.

src/cache/cloud_fraction.jl
szy21: This is looking good. I can take a look tonight / tomorrow. But if you are in a hurry to get this in I'm fine with merging it. (I didn't look at the changes in packages in Project and Manifest - maybe it would be good if a software engineer looks at them?)

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "ClimaAtmos"
 uuid = "b2c96348-7fb7-4fe0-8da9-78d88439e717"
 authors = ["Climate Modeling Alliance"]
-version = "0.33.0"
+version = "0.33.1"
```
szy21: Is the version change on purpose?

src/cache/cloud_fraction.jl
szy21: I think we can call it MLCloud (or MLCloudFraction) and worry about other options when we have them, but up to you.

src/cache/cloud_fraction.jl
```diff
@@ -272,3 +287,112 @@ NVTX.@annotate function set_cloud_fraction!(
...
+        compute_ml_cloud_fraction.(
+            Ref(cloud_ml.model),
+            ᶜmixing_length_field,
+            ᶜ∇q,
+            ᶜ∇θ,
+            Y.c.ρq_tot,
+            Y.c.ρ,
+            ᶜts,
+            Ref(thermo_params),
+            Ref(FT),
```
szy21:
```suggestion
            thermo_params
```
(I'm not sure, but I don't think you need Ref here. And you can get FT from thermo_params.)

src/cache/cloud_fraction.jl
```diff
@@ -272,3 +287,112 @@ NVTX.@annotate function set_cloud_fraction!(
...
+function compute_ml_cloud_fraction(
+    nn_model,
+    ᶜmixing_length_field,
+    ᶜ∇q,
+    ᶜ∇θ,
+    ρq_tot,
+    ρ,
+    ᶜts,
+    thermo_params,
+    FT,
```
szy21:
```suggestion
```

src/cache/cloud_fraction.jl
```diff
@@ -272,3 +287,112 @@ NVTX.@annotate function set_cloud_fraction!(
...
+)
+    # Saturation state at current thermodynamic state
```
szy21, line 344:
```suggestion
    FT = eltype(thermo_params)
    # Saturation state at current thermodynamic state
```

src/cache/cloud_fraction.jl
```diff
@@ -272,3 +287,112 @@ NVTX.@annotate function set_cloud_fraction!(
...
+        )
+end
+
+function compute_ml_cloud_fraction(
+    nn_model,
+    ᶜmixing_length_field,
+    ᶜ∇q,
+    ᶜ∇θ,
+    ρq_tot,
+    ρ,
```
szy21:
```suggestion
    q_tot
```

src/cache/cloud_fraction.jl
```diff
@@ -272,3 +287,112 @@ NVTX.@annotate function set_cloud_fraction!(
...
+        )
+
+    p.precomputed.cloud_diagnostics_tuple.cf .=
+        compute_ml_cloud_fraction.(
+            Ref(cloud_ml.model),
+            ᶜmixing_length_field,
+            ᶜ∇q,
+            ᶜ∇θ,
+            Y.c.ρq_tot,
+            Y.c.ρ,
```
szy21:
```suggestion
            specific.(Y.c.ρq_tot, Y.c.ρ),
```

src/cache/cloud_fraction.jl
```diff
@@ -272,3 +287,112 @@ NVTX.@annotate function set_cloud_fraction!(
...
+    thermo_params,
+    FT,
+)
+    # Saturation state at current thermodynamic state
+    q_sat = TD.q_vap_saturation(thermo_params, ᶜts)
+
+    # Liquid–ice potential temperature at current thermodynamic state
+    θli = TD.liquid_ice_pottemp(thermo_params, ᶜts)
+
+    q_v = specific(ρq_tot, ρ)
```
szy21:
```suggestion
```
(and use q_tot, which is an input argument for q_v below)

szy21: And note q_tot can be negative so you may want to clip it.

szy21 (approved):
The physics part looks good to me now. It may be good if @dennisYatunin can take a look at the package and autodiff change. But I'm fine with merging it.

Issue #4147, opened 2025-12-08
Investigate slowdown due to horizontal diffusion (AMD and Smagorinsky)


petebachant:
Preliminary results from @trontrytel:

<img width="2000" height="1200" alt="Image" src="https://github.com/user-attachments/assets/5174b8f9-cfea-4a89-b5a3-205b85ad187b" />

petebachant:
I just ran a quick benchmark with Smagorinsky and it was a ~10% slowdown, so not as bad as AMD I guess.

szy21:
Yes, we can use the constant horizontal diffusion.

szy21:
I think we will use constant horizontal diffusion for now. Smagorinsky and AMD will be used for configs other than AMIP, and if there are obvious things to optimize, it would still be good to do that. But if not, we can focus on other parts of the code now.

PR #4186, opened 2025-12-22
Lazify cell-centered stress and add horizontal-only strain rate for Smagorinsky-Lilly UV diffusion

petebachant:
This is a [~2% performance improvement](https://github.com/petebachant/clima-horizontal-diffusion/blob/main/results/summary.csv) on 1 GPU in UV mode. Not huge, but maybe worth it if it's still correct.

Towards #4147 

## TODO

- [ ] If valuable, rebase and remove env changes.
- [ ] Verify that omitting vertical components of strain rate does not impact horizontal diffusion.


src/utils/utilities.jl
```diff
@@ -82,6 +82,59 @@ state.
...
+ - `ᶜε`: Preallocated `UVW x UVW` tensor field for the strain rate at cell centers
+ - `ᶜu`: Velocity field at cell centers.
+
+See also [`compute_strain_rate_center_full!`](@ref) for the full calculation.
+
+# Notes:
+- This variant only computes horizontal gradients, reducing register pressure and local memory usage
+- Recommended for use when only horizontal diffusion is needed
+"""
+function compute_strain_rate_center_horizontal!(ᶜε, ᶜu)
```
haakon-e: do these (face/center) need to be separate functions? They should be identical.
Horizontal operations don't change spaces from centers to faces and vice versa, so only having one function for `compute_strain_rate_horizontal!(...)` should be sufficient.

Also, because these only do horizontal operations, they could be lazy instead of writing memory twice.

src/parameterized_tendencies/les_sgs_models/smagorinsky_lilly.jl
```diff
@@ -61,9 +61,17 @@ function set_smagorinsky_lilly_precomputed_quantities!(Y, p, model)
...
+    # If only using UV components, don't do vertical terms
+    if is_smagorinsky_UVW_coupled(model)
+        compute_strain_rate_center_full!(ᶜS, ᶜu, ᶠu)
+        compute_strain_rate_face_full!(ᶠS, ᶜu, ᶠu)
+    else
+        # Horizontal-only variants take two arguments:
+        # centers use ᶜu; faces use ᶠu
+        compute_strain_rate_center_horizontal!(ᶜS, ᶜu)
+        compute_strain_rate_face_horizontal!(ᶠS, ᶠu)
+    end
```
haakon-e: This misses the case if we run `SmagorinskyLilly{W}` (only apply smagorinsky diffusion in the vertical direction). 

Though in the case of `SmagorinskyLilly{UV}`, I think this may lead to different result in the tendency computation. Could you check the values you get when you do e.g.:
```julia
ᶜτ_smag = @. lazy(-2 * ᶜνₜ_h * ᶜS)
# ...
@. Yₜ.c.uₕ -= C12(wdivₕ(ᶜρ * ᶜτ_smag) / ᶜρ)
```
while using the `UV`-configuration

.buildkite/Manifest-v1.11.toml
```diff
@@ -2212,6 +2212,16 @@ git-tree-sha1 = "62389eeff14780bfe55195b7204c0d8738436d64"
 uuid = "ae029012-a4dd-5104-9daa-d747884805df"
 version = "1.3.1"
 
+[[deps.Revise]]
```
haakon-e, line 2464: I added Revise in a separate PR (#4187), so you can remove this here

src/parameterized_tendencies/les_sgs_models/smagorinsky_lilly.jl
```diff
@@ -9,7 +9,7 @@ import ClimaCore: Geometry
 """
     lilly_stratification_correction(p, ᶜS)
 
-Return a lazy representation of the Lilly stratification correction factor 
+Return a lazy representation of the Lilly stratification correction factor
```
haakon-e, line 12: these unrelated changes are a bit distracting. Is there any way you could disable them?

haakon-e (commented):
Thanks for testing this!

We should make sure the conditional evaluation remains numerically equivalent, though I think few of them are covered by CI. So may need to run some tests manually (or add to ci).

.buildkite/Manifest-v1.11.toml
petebachant: Yep, it's in the PR's TODO list to revert the env changes if the rest looks good.

src/parameterized_tendencies/les_sgs_models/smagorinsky_lilly.jl
petebachant: This could also be problematic if any of the cached vertical components like `ᶜS_norm_v` are used elsewhere.

petebachant: Added a TODO item to verify this by actually running code, but it looks like the vertical terms should be zeroed out by the projection in `strain_rate_norm`.

petebachant: Reran baseline and modified code with the latest `main` merged in and they're tough to compare because the baseline is all NaN at 2048 steps, and the modified code is not. Not sure if there's a bug in the current version on `main` or if it's just numerical stability: https://github.com/petebachant/clima-horizontal-diffusion/blob/e2f44a52f3a7c546fcb3b0d2a4c23ed5d2fbd30b/results/compare_smag_uh.log

PR #4189, opened 2025-12-23
Fix initialization of EveryCalendarDtSchedule for diagnostics

### LINKED ISSUE #4188: Off by one timestep for diagnostics when restarting
For this [buildkite run](https://buildkite.com/clima/climacoupler-amip/builds/223#019b0efc-3580-4d5e-bc41-012f849ba511) where a restart was used, the diagnostics produced from the restarted simulation are off by one coupler timestep. For this run, the times of the diagnostics are shifted by two minutes or one coupler timestep.

To reproduce this bug, you can

1. Run the simulation and create a checkpoint (simulation time at checkpoint should be in sync with diagnostics output schedule)
2. Restart from the checkpoint

This happens when restarting from both state and state + cache.

ph-kev:
closes #4188 - This PR fixes the initialization of `EveryCalendarDtSchedule`. Without this fix, the times and dates of the diagnostics after a restart would be off by a single timestep. Note that the first time and date of the diagnostics produced from a restarted simulation is still wrong, because of a bug in ClimaDiagnostics. See https://github.com/CliMA/ClimaDiagnostics.jl/issues/149

### DIFF
--- a/src/callbacks/get_callbacks.jl
+++ b/src/callbacks/get_callbacks.jl
@@ -120,13 +120,20 @@ function get_diagnostics(parsed_args, atmos_model, Y, p, sim_info, output_dir)
                     CA.promote_period.(Dates.Second(period_seconds))
             end
 
+            (; start_date, t_start) = sim_info
+            date_last =
+                t_start isa ITime ?
+                ClimaUtilities.TimeManager.date(t_start) :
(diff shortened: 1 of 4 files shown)


szy21:
Thank you! I have been wondering about this for a while.

PR #4192, opened 2025-12-28
remove and modify some single column setup

szy21:
## Purpose 

This is the first of a series of PRs that modify ClimaAtmos to use the new SurfaceFluxes. This PR removes some single column setups that are not used or tested. It also prescribes ustar for the cases that prescribe heat fluxes, as only prescribing heat fluxes is not supported anymore.

## Content

----

### DIFF
--- a/src/surface_conditions/surface_setups.jl
+++ b/src/surface_conditions/surface_setups.jl
@@ -36,17 +36,6 @@ function (::DefaultExchangeCoefficients)(params)
 end
 
 # All of the following are specific to TC cases
-struct Nieuwstadt end
-function (::Nieuwstadt)(params)
-    FT = eltype(params)
-    T = FT(300)
-    p = FT(1e5)
-    q_vap = FT(0)
-    z0 = FT(0.16)
-    θ_flux = FT(0.06)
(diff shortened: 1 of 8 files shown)


src/surface_conditions/surface_setups.jl
```diff
@@ -310,6 +234,7 @@ function (::ISDAC)(params)
     p = FT(102000)  # Pa
     lhf = shf = FT(0)  # neglect heat fluxes
     z0 = FT(4e-4)  # m  surface roughness length
-    parameterization = MoninObukhov(; z0, fluxes = HeatFluxes(; lhf, shf))
+    ustar = FT(0.05) # not specified in the literature, use a small value for stable condition
```
szy21: reasonable?

tapios: yes

tapios: But maybe a simpler way of dealing with the boundary conditions here is to either (a) just prescribe the roughness length and let the heat fluxes be what they want to be, or (b) prescribe zero heat fluxes and `Cd = 1e-3` (and remove the surface roughness length above, which is not needed)

szy21: I'll just prescribe the roughness length.

PR #4193, opened 2025-12-29
output cloud liquid and ice from the state directly

szy21:
## Purpose 

Get the cloud liquid and ice diagnostics from the state directly when using NonEquilMoistModel, rather than from the cloud_diagnostics_tuple (which uses quadratures). I will change the cloud_diagnostics_tuple very soon, after I'm done with the current 1M testing.

## Content

----


tapios (approved):
Looks good! (I'm not familiar with the column integral logic but I trust you got this right.)

PR #4196, opened 2026-01-04
some more cleanup of surface conditions

szy21:
## Purpose 

Some more cleanup before changing to the new SurfaceFluxes.
- Remove deardorff gustiness specification in ClimaAtmos. This will be set in the new SurfaceFluxes
- Remove the option that sets surface thermo state with prescribed surface pressure. The new SurfaceFluxes doesn't require a surface thermo state, and sets surface density internally.

Note these changes the behavior of some EDMF cases slightly, but I think it's ok.

## Content

----


tapios (approved):
LGTM

PR #4197, opened 2026-01-05
Trim trailing whitespace in `parameterized_tendencies`

petebachant:
The purpose here is to not obscure the diff in #4186 and future-proof against similar issues. I assume these aren't meaningful since they're mostly trailing single spaces, and Julia only respects trailing double spaces as new line indicators in docstrings?

### DIFF
--- a/src/parameterized_tendencies/gravity_wave_drag/non_orographic_gravity_wave.jl
+++ b/src/parameterized_tendencies/gravity_wave_drag/non_orographic_gravity_wave.jl
@@ -331,7 +331,7 @@ function non_orographic_gravity_wave_forcing(
     ᶜv_p1 = p.scratch.ᶜtemp_scalar_4
     ᶜbf_p1 = p.scratch.ᶜtemp_scalar_5
 
-    FT = eltype(ᶜρ) # Define the floating point type 
+    FT = eltype(ᶜρ) # Define the floating point type
 
     # Using interpolate operator, generate the field of ρ,u,v,z with on level shifted up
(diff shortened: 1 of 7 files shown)


src/parameterized_tendencies/gravity_wave_drag/non_orographic_gravity_wave.jl
```diff
@@ -375,7 +375,7 @@ function non_orographic_gravity_wave_forcing(
 
     mask_u = StaticBitVector{nc}(_ -> true)
     mask_v = StaticBitVector{nc}(_ -> true)
-    #We use StaticBitVector here because the unrolled_reduce function in Julia can cause memory allocation issues when the mask has more than 32 elements。 
+    #We use StaticBitVector here because the unrolled_reduce function in Julia can cause memory allocation issues when the mask has more than 32 elements。
```
petebachant: This is a weird one. Two spaces after what looks like a subscript o instead of a period?

nefrathenrici: I think this can just be changed to a period.

szy21: It might be a Chinese period by accident 😂

trontrytel:
FYI: You can ignore whitespace when comparing files on GitHub by adding the ?w=1 parameter to the URL

petebachant:
> FYI: You can ignore whitespace when comparing files on GitHub by adding the ?w=1 parameter to the URL

That's a great tip I've never heard before, thanks!

PR #4198, opened 2026-01-06
modify surface conditions and update SurfaceFluxes

szy21:
## Purpose 

Update to use the new SurfaceFluxes.

Also comments out some broken tests on main (that are unrelated to this PR): baroclinic_wave_equil_amd (it runs for an hour then fails due to regression test) and test restart_AtmosSimulation GPU manytests

Closes #4224 

## Content

----


src/surface_conditions/surface_conditions.jl
```diff
@@ -168,12 +168,14 @@ function surface_state_to_conditions(
     surf_state = surface_state(wrapped_sfc_setup, surface_local_geometry, interior_z, t)
     parameterization = surf_state.parameterization
     (; coordinates) = surface_local_geometry
-    FT = eltype(thermo_params)
+    Φ_sfc = SFP.grav(surface_fluxes_params) * coordinates.z
```
tapios: Can we please just have a function for geopotential that we call here and elsewhere where we need it? I'd like to make the code easy to change to a geopotential with nonconstant g.

src/surface_conditions/surface_conditions.jl
```diff
@@ -185,54 +187,48 @@ function surface_state_to_conditions(
...
+    )
     if atmos.moisture_model isa DryModel
-        ts = TD.PhaseDry_ρT(thermo_params, ρ, T)
+        q_vap = 0
+        ts = TD.PhaseDry_ρT(thermo_params, ρ_sfc, T_sfc)
     else
         # Assume that the surface is water with saturated air directly
         # above it.
-        q_vap_sat = TD.q_vap_saturation_generic(thermo_params, T, ρ, TD.Liquid())
+        q_vap_sat = TD.q_vap_saturation_generic(thermo_params, T_sfc, ρ_sfc, TD.Liquid())
```
tapios, line 187:
```suggestion
        q_vap_sat = TD.q_vap_saturation(thermo_params, T_sfc, ρ_sfc, TD.Liquid())
```

src/surface_conditions/surface_conditions.jl
```diff
@@ -185,54 +187,48 @@ function surface_state_to_conditions(
...
-    interior_T = TD.air_temperature(thermo_params, interior_ts)
-    ρ = interior_ρ * (T / interior_T)^(cv / R)
+    u_int = SA.SVector(interior_u, interior_v)
+    u_sfc = SA.SVector(u, v)
+
+    ρ_int = TD.air_density(thermo_params, interior_ts)
+    T_int = TD.air_temperature(thermo_params, interior_ts)
+    q_tot_int = TD.total_specific_humidity(thermo_params, interior_ts)
+    q_liq_int = TD.liquid_specific_humidity(thermo_params, interior_ts)
+    q_ice_int = TD.ice_specific_humidity(thermo_params, interior_ts)
```
tapios, line 170: Might as well as the functional form of TD here right away. (Or else we change it later.)

src/surface_conditions/surface_conditions.jl
szy21: will change it after updating to the latest TD

src/surface_conditions/surface_conditions.jl
szy21: will change it after updating to the latest TD

src/surface_conditions/surface_conditions.jl
szy21: I think I can pass in Φ_sfc directly from cache here. I'll change to that.

szy21: actually nevermind I don't have it in the cache. I'll add a function for geopotential.

PR #4209, opened 2026-01-12
Update scm setups - Small box

costachris:
Use small box for single column simulations. Addresses 0 horizontal velocity in current setups.

### DIFF
--- a/src/prognostic_equations/dss.jl
+++ b/src/prognostic_equations/dss.jl
@@ -69,7 +69,7 @@ See also:
 
 # Example structure for Y and p.ghost_buffer:
 # Y_state = (c = center_fields, f = face_fields)
-# params = (ghost_buffer = (c = центр_dss_buffer, f = face_dss_buffer), ...)
+# params = (ghost_buffer = (c = center_dss_buffer, f = face_dss_buffer), ...)
 
 dss!(Y_state, params, t_current)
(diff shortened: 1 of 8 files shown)


src/prognostic_equations/dss.jl
```diff
@@ -69,7 +69,7 @@ See also:
 
 # Example structure for Y and p.ghost_buffer:
 # Y_state = (c = center_fields, f = face_fields)
-# params = (ghost_buffer = (c = центр_dss_buffer, f = face_dss_buffer), ...)
+# params = (ghost_buffer = (c = center_dss_buffer, f = face_dss_buffer), ...)
```
costachris, line 72: @haakon-e

src/utils/utilities.jl
```diff
@@ -580,6 +580,14 @@ function issphere(space)
            Domains.SphereDomain
 end
 
+function isbox(space)
+    h_space = Spaces.horizontal_space(space)
+    return Meshes.domain(Spaces.topology(h_space)) isa Domains.RectangleDomain
+end
```
haakon-e, line 586: perhaps instead of this, the `iscolumn` method could (for now) be updated to check that the horizontal domain has 1 elements and poly order 1? Not sure if practicable.

costachris: I agree that's more concise. The column may be supported again after upstream ClimaCore changes, so I'll leave it for now to keep the box usage explicit.

PR #4210, opened 2026-01-13
Wip on moisture fixer in dss

trontrytel:
What we discussed today - apply "moisture fixer" inside dss to fix small tracer at the expense of water vapor

### DIFF
--- a/src/prognostic_equations/dss.jl
+++ b/src/prognostic_equations/dss.jl
@@ -20,7 +20,7 @@ Perform a weighted Direct Stiffness Summation (DSS) on components of the state `
 This function applies DSS to `ClimaCore.Field`s (or structures of `Field`s)
 typically named `.c` (center-located) and `.f` (face-located) within the state
 object `Y`. The DSS operation is essential in `ClimaCore` for ensuring that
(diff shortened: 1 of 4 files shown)


src/prognostic_equations/dss.jl
```diff
@@ -79,12 +79,27 @@ dss!(Y_state, params, t_current)
...
 end
 
+function enforce_nonnegative_tracers!(Y, p, t)
+    if p.atmos.enforce_nonnegative_tracers
+        @assert :ρq_tot in propertynames(Y.c)
+        for tracer_name in (:ρq_liq, :ρq_ice, :ρq_rai, :ρq_sno)
+            if tracer_name in propertynames(Y.c)
+                ρq_tot = Y.c.ρq_tot
+                tracer = getproperty(Y.c, tracer_name)
+                @. tracer += min(max(0, -tracer), max(0, ρq_tot))
```
szy21, line 96: Note that the sum of the change in tracers (e.g. if all four are negative) can still be larger than ρq_tot because it is constrained individually. I guess we don't want to have an artificial order for the clipping. But then, do we need the constraint of ρq_tot at all?

szy21:
superseded by #4211

PR #4212, opened 2026-01-13
chore(ci): remove JET tests

haakon-e:
This pull request removes all dependencies and related CI/test scripts for the JET static analysis tool from the project. The main focus is on cleaning up the codebase and build configuration by eliminating JET-related files, dependencies, and CI steps.


dennisYatunin (approved):
Thanks for removing these! The number of JET failures for the current main branch is 236, but it doesn't look like any of those failures are correlated with actual performance issues. While tests like this are useful for simple functions in ClimaCore, they appear to mostly be catching instabilities in the high-level interface of ClimaAtmos and its various dependencies. If these instabilities aren't affecting performance on GPUs (either runtime or latency), removing the tests is a good idea.

haakon-e:
Seems like buildkite does not complete gpu jobs. Need to merge #4215 then rerun ci here.

haakon-e:
@szy21 said I can merge without tests

PR #4215, opened 2026-01-14
chore(ci): update buildkite gpu syntax

haakon-e:
- caltech central cluster requires (as of 2025/01/14) that the type of gpu requested is specified.
- all jobs now run on nvidia p100 gpus

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -32,7 +32,8 @@ steps:
 
     agents:
       slurm_cpus_per_task: 8
-      gres: "gpu:p100:1"
+      slurm_gres: "gpu:p100:1"
+      slurm_partition: "gpu"
     env:
       JULIA_NUM_PRECOMPILE_TASKS: 8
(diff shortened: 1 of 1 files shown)


.buildkite/pipeline.yml
```diff
@@ -805,6 +805,7 @@ steps:
           - exit_status: 255
         agents:
           slurm_gpus_per_task: 1
+          gres: "gpu:nvidia_h200:2"
           slurm_ntasks: 2
```
haakon-e, line 813: If someone could confirm that these lines do what is expected, that'd be nice.

tapios:
> @tapios can you confirm that the gpu type (h200) I set buildkite to run with is reasonable. I don't know the pricing.

That's the most expensive GPU. It would be better if you can run this on p100, which is much more available too.

haakon-e:
> > @tapios can you confirm that the gpu type (h200) I set buildkite to run with is reasonable. I don't know the pricing.
> 
> That's the most expensive GPU. It would be better if you can run this on p100, which is much more available too.

Thanks. changed to `p100`.

PR #4225, opened 2026-01-19
Enable `VerticalMassBorrowingLimiter` in ClimaAtmos (CCv0.14.48)

akshaysridhar:
Enable vertical water borrowing limiters via the existing `tracer_nonnegativity_method` flag.

## Content

----


src/cache/cache.jl
```diff
@@ -93,6 +93,8 @@ function build_cache(
     aerosol_names,
     time_varying_trace_gas_names,
     steady_state_velocity,
+    vwb_thresholds = nothing,
+    vwb_species = nothing,
```
haakon-e: Since this function doesn't have a docstring, I think it would be more user friendly if these variable names are spelled out.

src/cache/cache.jl
haakon-e, line 135: You could consider moving the conditional code up here, as an `elseif`.

src/cache/cache.jl
```diff
@@ -135,7 +137,23 @@ function build_cache(
...
 
-    numerics = (; limiter, tracer_nonnegativity_limiter)
+    # Parse and create vertical_water_borrowing_limiter if configured
+    # Thresholds and species are passed as function arguments (not stored in AtmosModel for GPU compatibility)
+    # Check if tracer_nonnegativity_method is TracerNonnegativityVerticalWaterBorrowing
+    vertical_water_borrowing_limiter = nothing
+    vertical_water_borrowing_species = vwb_species
+
+    if atmos.water.tracer_nonnegativity_method isa TracerNonnegativityVerticalWaterBorrowing
+        thresholds = isnothing(vwb_thresholds) ? (FT(0.0),) : vwb_thresholds
```
haakon-e: I think it'd be nicer if the thresholds are passed as NamedTuples to the `VerticalMassBorrowingLimiter`. Then we don't have to carry around the species separately. This would require loosening the type restrictions of
```julia
struct VerticalMassBorrowingLimiter{T <: Tuple}
    q_min::T
end
```
to `T <: Union{Tuple, NamedTuple}`, or just `T <: NamedTuple`

src/prognostic_equations/limited_tendencies.jl
```diff
@@ -12,21 +40,51 @@ and `ref_Y.c.ρ`). Subsequently, `ClimaCore.Limiters.apply_limiter!` is called t
 specific quantity `ρχ/ρ` within bounds or if it only modifies `ρχ` based on `ρ`),
 ensuring the tracer quantities adhere to these bounds.
 
+For the VerticalMassBorrowingLimiter (via PR 2383 in ClimaCore.jl), if `p.numerics.vertical_water_borrowing_limiter`
```
haakon-e:
```suggestion
For the VerticalMassBorrowingLimiter (via [PR 2383](https://github.com/CliMA/ClimaCore.jl/pull/2383) in ClimaCore.jl), if `p.numerics.vertical_water_borrowing_limiter`
```

config/default_configs/default_config.yml
```diff
@@ -181,8 +181,9 @@ tracer_nonnegativity_method:
     - `elementwise_constraint`: Enforce nonnegativity by instantaneously redistributing tracer mass within an element
     - `vapor_constraint`: Enforce nonnegativity by instantaneously redistributing tracer mass between vapor (q_tot) and each tracer
     - `vapor_tendency`: Weakly enforce nonnegativity by applying a tendency to each tracer to redistribute tracer mass between vapor (q_tot) and each tracer
+    - `vertical_water_borrowing`: Enforce nonnegativity using VerticalMassBorrowingLimiter (PR 2383), which redistributes tracer mass vertically
```
haakon-e:
```suggestion
    - `vertical_water_borrowing`: Enforce nonnegativity using VerticalMassBorrowingLimiter [PR 2383](https://github.com/CliMA/ClimaCore.jl/pull/2383), which redistributes tracer mass vertically
```

src/prognostic_equations/limited_tendencies.jl
```diff
@@ -12,21 +40,51 @@ and `ref_Y.c.ρ`). Subsequently, `ClimaCore.Limiters.apply_limiter!` is called t
...
+
+    # Apply general limiter if configured
     if !isnothing(limiter)
         for ρχ_name in filter(is_tracer_var, propertynames(Y.c))
             Limiters.compute_bounds!(limiter, ref_Y.c.:($ρχ_name), ref_Y.c.ρ)
             Limiters.apply_limiter!(Y.c.:($ρχ_name), Y.c.ρ, limiter)
         end
     end
+
+    # Apply vertical water borrowing limiter if configured (PR 2383)
```
haakon-e:
```suggestion
    # Apply vertical water borrowing limiter if configured [PR 2383](https://github.com/CliMA/ClimaCore.jl/pull/2383)
```

src/cache/cache.jl
```diff
@@ -135,7 +137,23 @@ function build_cache(
...
+        vertical_water_borrowing_limiter = Limiters.VerticalMassBorrowingLimiter(thresholds)
+    end
+
+    numerics = (;
+        limiter,
```
haakon-e: Could we rename `limiter` to `sem_quasimonotone_limiter` everywhere?

src/solver/type_getters.jl
```diff
@@ -243,7 +243,24 @@ function get_numerics(parsed_args, FT)
     edmfx_tracer_upwinding =
         Val(Symbol(parsed_args["edmfx_tracer_upwinding"]))
 
-    limiter = parsed_args["apply_limiter"] ? CA.QuasiMonotoneLimiter() : nothing
+    limiter =
+        parsed_args["apply_sem_quasimonotone_limiter"] ? CA.QuasiMonotoneLimiter() : nothing
+
+    # Parse vertical_water_borrowing_limiter configuration (PR 2383)
```
haakon-e:
```suggestion
    # Parse vertical_water_borrowing_limiter configuration [PR 2383](https://github.com/CliMA/ClimaCore.jl/pull/2383)
```

src/solver/types.jl
```diff
@@ -34,18 +34,21 @@ Methods for enforcing tracer nonnegativity.
...
-There are three methods for enforcing tracer nonnegativity:
+There are four methods for enforcing tracer nonnegativity:
 - `TracerNonnegativityElementConstraint{qtot}`: Enforce nonnegativity by instantaneously redistributing
     tracer mass within an element (i.e. horizontally)
 - `TracerNonnegativityVaporConstraint{qtot}`: Enforce nonnegativity by instantaneously redistributing
     tracer mass between vapor (`q_vap = q_tot - q_cond`) and each tracer
 - `TracerNonnegativityVaporTendency`: Enforce nonnegativity by applying a tendency to each tracer,
     exchanging tracer mass between vapor (`q_vap`) and each tracer over time
+- `TracerNonnegativityVerticalWaterBorrowing`: Enforce nonnegativity using VerticalMassBorrowingLimiter
+    (PR 2383), which redistributes tracer mass vertically. Note: `qtot` parameter is not applicable to this method.
```
haakon-e:
```suggestion
    [PR 2383](https://github.com/CliMA/ClimaCore.jl/pull/2383), which redistributes tracer mass vertically. Note: `qtot` parameter is not applicable to this method.
```

src/solver/types.jl
```diff
@@ -533,6 +556,13 @@ Base.@kwdef struct AtmosNumerics{EN_UP, TR_UP, ED_UP, SG_UP, ED_TR_UP, TDC, RR,
 
     limiter::LIM = nothing
 
+    """Vertical water borrowing limiter (VerticalMassBorrowingLimiter from PR 2383).
```
haakon-e:
```suggestion
    """Vertical water borrowing limiter (VerticalMassBorrowingLimiter from [PR 2383](https://github.com/CliMA/ClimaCore.jl/pull/2383)).
```

src/solver/type_getters.jl
```diff
@@ -243,7 +243,24 @@ function get_numerics(parsed_args, FT)
...
+    # Store as empty type marker - thresholds and species stored separately in cache for GPU compatibility
```
haakon-e: Are tuples not GPU compatible? I think it would be clearer if we reuse the same struct for this and the other water limiters/constraints

src/solver/types.jl
```diff
@@ -512,12 +515,32 @@ struct Implicit <: AbstractTimesteppingMode end
 
 struct QuasiMonotoneLimiter end # For dispatching to use the ClimaCore QuasiMonotoneLimiter.
 
+"""
+    VerticalMassBorrowingLimiter
+
+Type marker for the ClimaCore VerticalMassBorrowingLimiter (PR 2383).
+The limiter is created in the cache with the thresholds (stored separately for GPU compatibility).
+"""
+struct VerticalMassBorrowingLimiter end
```
haakon-e: do we need both `VerticalMassBorrowingLimiter` and `TracerNonnegativityVerticalWaterBorrowing`?

src/solver/type_getters.jl
```diff
@@ -1051,6 +1069,52 @@ function get_simulation(config::AtmosConfig)
...
+        if haskey(config.parsed_args, "vertical_water_borrowing_thresholds") &&
+           !isnothing(config.parsed_args["vertical_water_borrowing_thresholds"])
+            thresholds_config = config.parsed_args["vertical_water_borrowing_thresholds"]
+            if thresholds_config isa Vector
+                tuple(FT.(thresholds_config)...)
+            else
+                (FT(thresholds_config),)
+            end
+        else
+            (FT(0.0),)  # Default threshold
```
haakon-e: Is the default threshold, a 1-tuple, compatible with the default species (apply to all)? Would be good to unit test some of these options.

config/default_configs/default_config.yml
```diff
@@ -226,9 +227,15 @@ initial_condition:
 perturb_initstate:
   help: "Add a perturbation to the initial condition [`false`, `true` (default)]"
   value: true
-apply_limiter:
-  help: "Apply a horizontal limiter to every tracer [`false` (default), `true`]"
+apply_sem_quasimonotone_limiter:
+  help: "Apply a horizontal SEM quasi-monotone limiter to every tracer [`false` (default), `true`]"
   value: false
+vertical_water_borrowing_thresholds:
```
szy21: Will we experiment with this? If we almost never need to change this, I would prefer fewer options.

src/cache/cache.jl
imreddyTeja: Yeah, I agree with this. I'll open an issue in ClimaCore

src/prognostic_equations/limited_tendencies.jl
```diff
@@ -12,21 +40,51 @@ and `ref_Y.c.ρ`). Subsequently, `ClimaCore.Limiters.apply_limiter!` is called t
...
+    # Also note: species filtering is done here, not passed to apply_limiter! (which doesn't support it)
+    if !isnothing(vertical_water_borrowing_limiter)
+        for ρχ_name in filter(is_tracer_var, propertynames(Y.c))
+            # Filter tracers based on species configuration
+            if _should_apply_limiter_to_tracer(ρχ_name, vertical_water_borrowing_species)
+                # PR 2383: Usage: limiter = Limiters.VerticalMassBorrowingLimiter((0.0,))
```
imreddyTeja: ClimaCore should be changed to support species filtering, but in the case where `isnothing(species)`, the limiter is still being applied to each species individually, even though it could be done as a batch.

src/cache/cache.jl
imreddyTeja: Is `vwb_thresholds` expected to be a tuple of length 1 here, or the same length as `vwb_species`?

config/longrun_configs/longrun_aquaplanet_allsky_1M_vwb.yml
szy21, line 1: I think it's better to finalize the numerics and change `longrun_aquaplanet_1M` directly. I'm fine with adding it for temporary testing.

src/prognostic_equations/limited_tendencies.jl
szy21: I think it's also fine to not refer to ClimaCore PR here.

src/prognostic_equations/limited_tendencies.jl
szy21: same here (and elsewhere in the PR)

src/prognostic_equations/limited_tendencies.jl
```diff
@@ -1,3 +1,31 @@
...
+
+Returns:
+- `true` if the limiter should be applied to this tracer, `false` otherwise
+"""
+function _should_apply_limiter_to_tracer(ρχ_name, species)
+    if isnothing(species)
+        return true  # Apply to all tracers
+    elseif species isa Tuple
+        return ρχ_name in species
+    elseif species isa Function
```
szy21: what's an example for `species isa Function`?

src/solver/types.jl
```diff
@@ -519,12 +522,32 @@ struct Implicit <: AbstractTimesteppingMode end
...
```
imreddyTeja: I know this is following the pattern, but I find it odd to have a struct with the same name in ClimaAtmos and ClimaCore

src/solver/types.jl
szy21: This is a little confusing to me - If I understand it correctly we choose one of the first three options, but we can use the fourth option (vertical borrowing) together with one of the others? (It's fine if you want to change this later.)

config/default_configs/default_config.yml
```diff
@@ -181,8 +181,9 @@ tracer_nonnegativity_method:
     - `elementwise_constraint`: Enforce nonnegativity by instantaneously redistributing tracer mass within an element
     - `vapor_constraint`: Enforce nonnegativity by instantaneously redistributing tracer mass between vapor (q_tot) and each tracer
     - `vapor_tendency`: Weakly enforce nonnegativity by applying a tendency to each tracer to redistribute tracer mass between vapor (q_tot) and each tracer
+    - `vertical_water_borrowing`: Enforce nonnegativity using VerticalMassBorrowingLimiter (PR 2383), which redistributes tracer mass vertically
     - `~` (default): Do not enforce nonnegativity
-    Note: append `_qtot` to the method name to enforce nonnegativity of q_tot as well.
+    Note: append `_qtot` to the method name to enforce nonnegativity of q_tot as well (not applicable to `vertical_water_borrowing`).
```
oalcabes, line 186: So is vertical_water_borrowing by default enforcing nonnegativity on qtot (as per "vertical_water_borrowing_species" below)? Maybe that config argument can be changed to work the same as these ones?

src/prognostic_equations/limited_tendencies.jl
akshaysridhar: It's a bit contrived but if we have a longer list of vars something like this to "exclude" elements: 
```
julia> species = ρχ_name -> ρχ_name ∉ ("tracer1", "tracer2")
#11 (generic function with 1 method)

julia> ρχ_name = ("total", "liquid", "ice", "snow", "tracer1", "tracer2", "tracer3", "tracer4")
("total", "liquid", "ice", "snow", "tracer1", "tracer2", "tracer3", "tracer4")

julia> for x in ρχ_name
           @info "Limiter?",  species(x)
       end
[ Info: ("Limiter?", true)
[ Info: ("Limiter?", true)
[ Info: ("Limiter?", true)
[ Info: ("Limiter?", true)
[ Info: ("Limiter?", false)
[ Info: ("Limiter?", false)
[ Info: ("Limiter?", true)
[ Info: ("Limiter?", true)
```

I don't think this is strictly necessary (since parsed_args would handle the species lists from `yaml` files through the other options).

config/default_configs/default_config.yml
akshaysridhar: I expect we'll just use the default 0 values, so I can simplify this by removing the thresholds.

trontrytel:
This is a broader question that we discussed today with @haakon-e and @sajjadazimi. 

In the horizontal and vertical water borrowing algorithms when we adjust the tracers but don't adjust density and energy, wouldn't that be seen from the model point of view as a phase change? Say I move `q_liq` from one nodal point to the other, either horizontally or vertically. If I don't adjust the total energy and density in those two points, it will look like a phase change that happened in two nodal points - `q_liq` was changed but `q_tot` stays the same, therefore it was `q_vap` that was inherently adjusted. 

And, do we want to adjust density and energy, or just hide it under the rug of algorithm inaccuracy and move on?

src/solver/types.jl
```diff
@@ -524,7 +527,18 @@ struct SmoothMinimumBlending <: AbstractScaleBlendingMethod end
...
+    TR_UP,
+    ED_UP,
+    SG_UP,
+    ED_TR_UP,
+    TDC,
+    RR,
+    LIM,
+    DM,
+    HD,
+}
```
haakon-e, line 541: nit:
```suggestion
Base.@kwdef struct AtmosNumerics{
    EN_UP, TR_UP, ED_UP, SG_UP, ED_TR_UP, TDC, RR, LIM, DM, HD
}
```
would this also pass formatter?

.buildkite/Manifest-v1.11.toml
```diff
@@ -1,6 +1,6 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.11.8"
+julia_version = "1.11.6"
```
akshaysridhar: I locally use 1.11.6 per the recommended version docs; but I'll clean this up prior to final ci-checks.

config/default_configs/default_config.yml
akshaysridhar: I've updated the docstring to clarify usage.

PR #4230, opened 2026-01-21
Insolation simplifications and update to 1.1

tapios:
## Purpose 
* Updates to Insolation 1.1 API
* Simplifies atmos code by separating concerns: all radiation and orbital geometry calculations and helper variables are handled by Insolation
* Removes a test that is a pure insolation test and covered by Insolation.jl
* Not related: Sets a restart test to soft fail as it fails randomly

## Content

----


src/callbacks/callbacks.jl
```diff
@@ -201,56 +202,58 @@ function set_insolation_variables!(Y, p, t, ::IdealizedInsolation)
...
     if eltype(bottom_coords) <: Geometry.LatLongZPoint
-        @. insolation_tuple = instantaneous_zenith_angle(
-            d,
-            δ,
-            η_UTC,
-            bottom_coords.long,
-            bottom_coords.lat,
-        ) # the tuple is (zenith angle, azimuthal angle, earth-sun distance)
+        # Calculate insolation for each grid point
+        @. begin
```
szy21: maybe don't use this begin - end

src/parameterized_tendencies/radiation/radiation.jl
```diff
@@ -409,14 +408,6 @@ function get_cloud_cache(::PrescribedCloudInRadiation, Y, start_date)
     return (; prescribed_clouds_field, prescribed_cloud_timevaryinginputs)
 end
 
-insolation_cache(_, _) = (;)
-function insolation_cache(::TimeVaryingInsolation, Y)
-    FT = Spaces.undertype(axes(Y.c))
-    return (;
-        insolation_tuple = similar(Spaces.level(Y.c, 1), Tuple{FT, FT, FT})
-    )
-end
```
szy21, line 418: we need it:)

src/callbacks/callbacks.jl
```diff
@@ -189,7 +190,7 @@ function set_insolation_variables!(Y, p, t, ::ExternalTVInsolation)
 
     # set insolation variables from the values within the fields
     rrtmgp_model.cos_zenith .= Fields.field2array(coszen)
-    rrtmgp_model.weighted_irradiance .= Fields.field2array(rsdt)
+    rrtmgp_model.weighted_irradiance .= Fields.field2array(rsdt)  # rsdt = solar flux
```
szy21: This looks wrong but I'm not sure. I'll check with Costa later.

szy21 (approved):

src/callbacks/callbacks.jl
```diff
@@ -218,39 +221,37 @@ function set_insolation_variables!(Y, p, t, tvi::TimeVaryingInsolation)
...
+            insolation_params,
+        )
+        
     else
-        # assume that the latitude and longitude are both 0 for flat space,
-        # so that insolation_tuple is a constant Field
-        insolation_tuple .=
-            Ref(instantaneous_zenith_angle(d, δ, η_UTC, FT(0), FT(0)))
+        # assume that the latitude and longitude are both 0 for flat space
+        insolation_tuple .= Ref(Insolation.insolation(
```
szy21, line 241: Changing this back so that it passes the regression test.

src/callbacks/callbacks.jl
```diff
@@ -189,7 +189,8 @@ function set_insolation_variables!(Y, p, t, ::ExternalTVInsolation)
 
     # set insolation variables from the values within the fields
     rrtmgp_model.cos_zenith .= Fields.field2array(coszen)
-    rrtmgp_model.weighted_irradiance .= Fields.field2array(rsdt)
+    # TODO: is this correct?
```
szy21: I think I'll just change this as it looks wrong to me.

src/callbacks/callbacks.jl
```diff
@@ -167,13 +167,13 @@ function set_insolation_variables!(Y, p, t, ::RCEMIPIIInsolation)
...
     rrtmgp_model.cos_zenith .= cosd(FT(42.05))
-    rrtmgp_model.weighted_irradiance .= FT(551.58)
+    rrtmgp_model.toa_flux .= FT(551.58)  # solar flux (W m⁻²)
 end
 
 function set_insolation_variables!(Y, p, t, ::GCMDrivenInsolation)
     (; rrtmgp_model) = p.radiation
     rrtmgp_model.cos_zenith .= Fields.field2array(p.external_forcing.cos_zenith)
-    rrtmgp_model.weighted_irradiance .=
+    rrtmgp_model.toa_flux .=
```
tapios, line 176: The terminology here is confusing. We have toa_flux on the left, which seems fine. But then why is it `insolation` on the right? Is it not insolation but also flux on the right?

src/callbacks/callbacks.jl
tapios: So it was wrong all along?

src/callbacks/callbacks.jl
```diff
@@ -218,39 +217,37 @@ function set_insolation_variables!(Y, p, t, tvi::TimeVaryingInsolation)
...
     cos_zenith =
         Fields.array2field(rrtmgp_model.cos_zenith, axes(bottom_coords))
-    weighted_irradiance = Fields.array2field(
-        rrtmgp_model.weighted_irradiance,
+    toa_flux = Fields.array2field(
+        rrtmgp_model.toa_flux,
         axes(bottom_coords),
     )
+
+    # Use Insolation API: F, S, μ, ζ = insolation(datetime, lat, lon, params)
```
tapios:
```suggestion
    # Use Insolation API: insolate_tuple = insolation(datetime, lat, lon, params)
```

src/utils/era5_observations_to_forcing_file.jl
```diff
@@ -481,7 +464,16 @@ function generate_external_forcing_file(
     end
 
     # add coszen
-    coszen_list = get_coszen_inst.(lat, lon, tvforcing["valid_time"][:], FT)
+    # Compute coszen and solar flux using Insolation.jl
+    coszen_list = map(tvforcing["valid_time"][:]) do date
+        F, S, μ, ζ = Insolation.insolation(
```
tapios, line 469: Is this ok here, without returning to `insolation_tuple`?

szy21: This looks like a pointwise function so I think it's ok.

src/callbacks/callbacks.jl
szy21: Yes I think what in p.external_forcing.insolation is actually flux.

szy21: I can try changing it to p.external_forcing.toa_flux

PR #4231, opened 2026-01-22
remove grid-scale thermo state

szy21:
## Purpose 

Removes grid-scale thermo state and uses new Thermodynamics API.

Removes ts and adds ᶜT, ᶜq_tot_safe, ᶜq_liq_rai, ᶜq_ice_sno to the cache. Also adds h_tot back to the cache because using `@lazy` results in gpu error. The difference between ᶜq_tot_safe and ᶜq_tot (a lazy broadcasting) is ᶜq_tot_safe is clipped to 0. The difference between ᶜq_liq_rai, ᶜq_liq and ᶜq_rai is ᶜq_liq_rai = ᶜq_liq + ᶜq_rai and clipped to 0. Same for  ᶜq_ice_sno. I am open to different names. ᶜT, ᶜq_tot_safe, ᶜq_liq_rai, ᶜq_ice_sno essentially create the previous thermo state. Clipping the humidities to 0 recovers the previous behavior of thermo state.

I will remove subgrid thermo state and remaining TD.PhaseEquil in separate PRs.

## Content

----


test/gpu_horizontal_dynamics_mwe.jl
szy21, line 1: ignore this

test/gpu_split_divh_mwe.jl
szy21, line 1: ignore this

src/cache/precomputed_quantities.jl
```diff
@@ -498,8 +576,16 @@ NVTX.@annotate function set_implicit_precomputed_quantities!(Y, p, t)
...
     end
-    @. ᶜts = ts_gs(thermo_args..., Y.c, ᶜK, ᶜΦ, Y.c.ρ)
-    @. ᶜp = TD.air_pressure(thermo_params, ᶜts)
+    # Compute all thermodynamic state variables in one call (avoids multiple saturation_adjustment calls)
+    ᶜthermo_state = p.scratch.ᶜtemp_thermo_state
+    @. ᶜthermo_state = thermo_state_gs(thermo_args..., Y.c, ᶜK, ᶜΦ, Y.c.ρ)
+    @. ᶜT = ᶜthermo_state.T
+    @. ᶜq_tot_safe = ᶜthermo_state.q_tot_safe
+    @. ᶜq_liq_rai = ᶜthermo_state.q_liq_rai
+    @. ᶜq_ice_sno = ᶜthermo_state.q_ice_sno
```
szy21: ClimaCore doesn't support a single fused broadcast, and this is the only way to get the cached variables without multiple  thermo_state_gs (saturation_adjustment) calls. However, this doesn't work with autodiff. Having multiple thermo_state_gs calls work, but that will be slower. I'll look into it more and see what we can do.

src/prognostic_equations/advection.jl
```diff
@@ -72,10 +71,12 @@ NVTX.@annotate function horizontal_dynamics_tendency!(Yₜ, Y, p, t)
...
     end
 
-    ᶜΦ_r = @. lazy(phi_r(thermo_params, ᶜts))
-    ᶜθ_v = @. lazy(theta_v(thermo_params, ᶜts))
-    ᶜθ_vr = @. lazy(theta_vr(thermo_params, ᶜts))
-    ᶜΠ = @. lazy(dry_exner_function(thermo_params, ᶜts))
+    (; ᶜq_tot_safe) = p.precomputed
+    ᶜΦ_r = @. lazy(phi_r(thermo_params, ᶜp))
+    ᶜθ_v = p.scratch.ᶜtemp_scalar
+    @. ᶜθ_v = theta_v(thermo_params, ᶜT, ᶜp, ᶜq_tot_safe, ᶜq_liq_rai, ᶜq_ice_sno)
```
szy21, line 77: Using lazy here doesn't work on gpu (same problem as h_tot)

src/cache/precomputed_quantities.jl
szy21: The only thing I have tried that doesn't allocate and works for autodiff is to have separate functions T_gs, q_tot_safe_gs, q_liq_rai_gs, q_ice_sno_gs, then call
```
ᶜT = T_gs(...)
@. ᶜq_tot_safe = q_tot_safe_gs(...)
...
```
But this calls saturation adjustment multiple times. I'll ask Dennis about this.

szy21: The simulation isn’t slower with multiple calls so maybe it’s ok.

src/cache/cloud_fraction.jl
```diff
@@ -337,20 +374,18 @@ function compute_ml_cloud_fraction(
...
     FT = eltype(thermo_params)
-    # Saturation state at current thermodynamic state
-    q_sat = TD.q_vap_saturation(thermo_params, ᶜts)
-
-    # Liquid–ice potential temperature at current thermodynamic state
-    θli = TD.liquid_ice_pottemp(thermo_params, ᶜts)
-
+    # Saturation state at current thermodynamic state (via θ, p, q_tot)
+    ts_current = TD.PhaseEquil_pθq(thermo_params, p_c, θli, q_tot)
+    q_sat = TD.q_vap_saturation(thermo_params, ts_current)
```
tapios, line 384: It might be more straightforward to pass in temperature rather than pressure, and then compute `q_sat` and other quantities we need from that. Wouldn't that avoid use of `PhaseEquil...` here? (Same for the other cloud_fraction function above.)

Perhaps your intention was to do that, or something like it, in a separate PR, and that's fine too.

src/cache/cloud_fraction.jl
```diff
@@ -359,17 +394,16 @@ function compute_ml_cloud_fraction(
     π_4 = (ᶜ∇θ * ᶜmixing_length_field) / θli_sat
 
     return apply_cf_nn(nn_model, π_1, π_2, π_3, π_4)
-
 end
 
-function saturation_distance(q_tot, q_sat, ᶜts, θli, thermo_params, Δθli_fd)
+function saturation_distance(q_tot, q_sat, p_c, θli, thermo_params, Δθli_fd)
```
tapios, line 399: Similar to above: I think this gets easier to change, and to reduce reliance on equilibrium constructors, by passing in temperature rather than pressure.

src/cache/precomputed_quantities.jl
```diff
@@ -431,12 +438,73 @@ function thermo_vars(::NonEquilMoistModel, ::Any, ᶜY, K, Φ)
...
+    q_tot = specific(ᶜY.ρq_tot, ᶜY.ρ)
+    return max(0, q_tot)
+end
+
+# q_liq_rai getters
+function q_liq_rai_gs(thermo_params, ::DryModel, ᶜY, K, Φ, ρ)
+    return zero(eltype(thermo_params))
+end
+
+function q_liq_rai_gs(thermo_params, ::EquilMoistModel, ᶜY, K, Φ, ρ)
```
tapios: Can you combine the temperature and humidity getters to return a tuple of all? That way, we do not need to call SA multiple times for EquilMoist?

src/cache/precomputed_quantities.jl
```diff
@@ -498,8 +566,16 @@ NVTX.@annotate function set_implicit_precomputed_quantities!(Y, p, t)
         # @. ᶜK += Y.c.ρtke / Y.c.ρ
         # TODO: We should think more about these increments before we use them.
     end
-    @. ᶜts = ts_gs(thermo_args..., Y.c, ᶜK, ᶜΦ, Y.c.ρ)
-    @. ᶜp = TD.air_pressure(thermo_params, ᶜts)
+    # Compute thermodynamic state variables using individual getter functions.
+    # Note: For EquilMoistModel, this calls saturation_adjustment 3 times per grid point
```
tapios: See comment above: why not combining getters so that we call SA only once?

src/diagnostics/core_diagnostics.jl
```diff
@@ -1322,26 +1302,18 @@ add_diagnostic_variable!(
...
         return out
     else
         cp = CAP.cp_d(cache.params)
         dse = cache.scratch.ᶜtemp_scalar
-        @. dse =
-            state.c.ρ * (
-                cp * TD.air_temperature(thermo_params, cache.precomputed.ᶜts) +
-                cache.core.ᶜΦ
-            )
+        @. dse = state.c.ρ * (cp * ᶜT + cache.core.ᶜΦ)
```
tapios: Why not using TD function? (I know it subtracts T_0, but no harm changing that, right?)

src/parameterized_tendencies/gravity_wave_drag/orographic_gravity_wave.jl
```diff
@@ -107,15 +106,16 @@ function orographic_gravity_wave_tendency!(Yₜ, Y, p, t, ::OrographicGravityWav
...
         parent(topo_k_pbl[colidx]) .=
             get_pbl(ᶜp[colidx], ᶜT[colidx], ᶜz[colidx], grav, cp_d)
     end
 
     # buoyancy frequency at cell centers
     parent(ᶜdTdz) .= parent(Geometry.WVector.(ᶜgradᵥ.(ᶠinterp.(ᶜT))))
-    ᶜN = @. (grav / ᶜT) * (ᶜdTdz + grav / TD.cp_m(thermo_params, ᶜts)) # this is actually ᶜN^2
+    (; ᶜq_tot_safe, ᶜq_liq_rai, ᶜq_ice_sno) = p.precomputed
+    ᶜcp_m = @. lazy(TD.cp_m(thermo_params, ᶜq_tot_safe, ᶜq_liq_rai, ᶜq_ice_sno))
+    ᶜN = @. (grav / ᶜT) * (ᶜdTdz + grav / ᶜcp_m) # this is actually ᶜN^2
```
tapios, line 118: Good to leave a TODO here: We need a central helper for N2, so that we do not compute it in multiple ways in different places.

.buildkite/longruns_gpu/pipeline.yml
szy21, line 1: revert this

src/prognostic_equations/eddy_diffusion_closures.jl
```diff
@@ -96,18 +108,18 @@ function buoyancy_gradients(
...
-    ts = bg_model.ts
-    ∂b∂θv = g / TD.virtual_pottemp(thermo_params, ts)
+    (; T, ρ, q_tot, q_liq, q_ice) = bg_model
+    ∂b∂θv = g / TD.virtual_pottemp(thermo_params, T, ρ, q_tot, q_liq, q_ice)
 
-    T = TD.air_temperature(thermo_params, ts)
-    λ = TD.liquid_fraction(thermo_params, ts)
+    # TODO: is this correct for equilmoist?
+    λ = TD.liquid_fraction(thermo_params, T, q_liq, q_ice)
     lh =
```
tapios: Use `latent_heat_mixed`. (Better yet: we should probably provide a `latent_heat` method in TD that takes temperature and humidities as input.)

src/cache/precomputed_quantities.jl
szy21: I did try to combine it here: https://github.com/CliMA/ClimaAtmos.jl/pull/4231#discussion_r2726498552. But this doesn't work for autodiff, as I need to give the tuple a type (of FTs). I'll ask Dennis if there is a way to make it work. FWIW this doesn't make things slower (maybe because there are only 2 iterations in saturation adjustment.)

tapios: SA returns a tuple. Why can't we use that directly, with duck typing? In any case, yes talk with Dennis.

tapios: If nothing else works, you can change the humidity getters to using temperature (from SA) as input. At least then it won't iterate again.

src/cache/cloud_fraction.jl
szy21: Yes, I want to remove the PhaseEquil in the next PR, partially because I didn't know how to deal with this field of named tuple. But now I think I know:)

PR #4233, opened 2026-01-22
Remove thermal state from initial conditions

tapios:
## Purpose 

Also removes discrete hydrostatic balance

## Content

----


szy21:
We need to change the local_state in `set_discrete_hydrostatic_balanced_state!`. But should we just delete it? We don't need it, and it was derived with the previous formulation of pressure gradient.

PR #4238, opened 2026-01-27
Advection kernel splits

daverumph:
## Purpose 
Improve GPU performance of AMIP with 1 moment precip based on profiling without changing behavior

## Content
Pull out common subexpression and store in shared memory; split one kernel into three.
All kernels in update_jacobian are now less than 1.2% of total step time.

----
- [X] I have read and checked the items on the review checklist.

### DIFF
--- a/src/prognostic_equations/implicit/manual_sparse_jacobian.jl
+++ b/src/prognostic_equations/implicit/manual_sparse_jacobian.jl
@@ -1088,15 +1088,20 @@ function update_jacobian!(alg::ManualSparseJacobian, cache, Y, p, dtγ, t)
                         ) - (I,)
                     ∂ᶜχʲ_err_∂ᶠu₃ʲ =
                         matrix[χʲ_name, @name(f.sgsʲs.:(1).u₃)]
-                    @. ∂ᶜχʲ_err_∂ᶠu₃ʲ =
+                    # pull out and store for performance
+                    @. ᶜtracer_advection_matrix =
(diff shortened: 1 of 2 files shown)


src/prognostic_equations/implicit/manual_sparse_jacobian.jl
```diff
@@ -1092,15 +1092,20 @@ function update_jacobian!(alg::ManualSparseJacobian, cache, Y, p, dtγ, t)
                         ) - (I,)
                     ∂ᶜχʲ_err_∂ᶠu₃ʲ =
                         matrix[χʲ_name, @name(f.sgsʲs.:(1).u₃)]
-                    @. ∂ᶜχʲ_err_∂ᶠu₃ʲ =
+                    # pull out and store in shmem for performance
```
imreddyTeja: I don't think this is putting data in shmem. It should be going to a cache that lives in global memory.

src/prognostic_equations/implicit/manual_sparse_jacobian.jl
```diff
@@ -1469,30 +1474,35 @@ function update_jacobian!(alg::ManualSparseJacobian, cache, Y, p, dtγ, t)
 
                         ∂ᶜρχ_err_∂ᶜρa =
                             matrix[ρχ_name, @name(c.sgsʲs.:(1).ρa)]
+                        # pull out and store in shmem for kernel performance
```
imreddyTeja: same comment here

PR #4241, opened 2026-01-28
remove large artifacts

szy21:
## Purpose 

The uploading artifacts sometimes hang when the artifacts are too large, resulting in random failures in buildkite.

## Content

----


szy21:
Actually maybe not bumping the ref counter will fix it after all, as https://buildkite.com/clima/climaatmos-ci/builds/27850 didn't move main result. Let me try.

PR #4242, opened 2026-01-28
Modify Buildkite pipeline to ignore .tar artifacts

imreddyTeja:
Uploading the nc_files.tar and hdf5_files.tar can
cause some buildkite jobs to fail. This makes them only upload .pdf, .dat, .toml, and .yml files.

This is a follow up to #4241 . Instead of deleting the large files, it ignores them. 

----


szy21 (approved):
Thank you!

szy21:
The test that failed seems random. Feel free to mark it as soft-fail.

PR #4243, opened 2026-01-28
Add functionality to change compute schedule

### LINKED ISSUE #4066: Do not accumulate diagnostics at each timestep
This is not necessary (even though this doesn't seem to cost much). Accumulating the diagnostics hourly or longer would be good enough.

ph-kev:
closes #4066 - This PR allows the user to change the compute schedule via the `compute_every` keyword in the YAML file.


ph-kev:
Okay, I will see what can be done for the default diagnostics in another PR. For the different compute schedules, I just remembered that the diagnostics are forced to be unique, so it wouldn't crash the simulation or anything like that. I don't remember which diagnostics are removed though, so I will check that and see if anything need to be updated in ClimaAtmos.

Issue #4223, opened 2026-01-17
test restart_AtmosSimulation GPU manytests is failing on main


szy21:
See https://buildkite.com/clima/climaatmos-ci/builds/27628#019bca5b-6ae8-4ad8-a8e0-63e4014470c5. I'll mark it as soft-fail now. @nefrathenrici could you look at it?

szy21:
fails again here: https://buildkite.com/clima/climaatmos-ci/builds/27703#019be264-c6b7-48f7-b538-7e811d911992

PR #4246, opened 2026-01-29
Fix restart_AtmosSimulation test

nefrathenrici:
This PR fixes https://github.com/CliMA/ClimaAtmos.jl/issues/4223 by reconstructing the grid for each new simulation within a restart test. Reusing the same grid across restarted simulations caused nondeterministic results and test failure.
I ran the test 30 times and did not see a failure, so I think this can close the issue.


nefrathenrici:
Closing for now as it didn't actually solve the flaky test.

PR #4248, opened 2026-01-29
change default timestepper to ARS222

szy21:
## Purpose 

From @sajjadazimi:)

> ARS222 is stiffly accurate; the solution that is found in Newton is used as the initial condition of the next step. Based on the reference paper being stiffly accurate helps with extremely stiff problems. The accuracy is the same for both. So I think I like ARS222 better.

## Content

----


szy21:
we should really remove all the mbw_scaling_configs... I'll do it next.

sajjadazimi (approved):

PR #4255, opened 2026-01-30
Update compute schedule of default diagnostics

### LINKED ISSUE #4253: Update compute schedules for default diagnostics
Following the discussion in https://github.com/CliMA/ClimaAtmos.jl/pull/4243, the default diagnostics should be updated, so that precipitation and radiation variables (pr, prra, prsn, prw, rsd, rsdt, rsds, rsu, rsut, rsus, rld, rlds, rlu, rlut, rlus, rsdcs, rsdscs, rsucs, rsutcs, rsuscs, rldcs, rldscs, rldscs, rlucs, rlutcs) are computed every hour and all the other variables are computed every 6 hour.

The choice of compute schedule is also determined by the length of the simulation, since for a shorter simulation, it might not make to compute the field every hour or six hours.

ph-kev:
close #4253 - This PR changes the default compute schedule from `EveryStepSchedule` to a frequency of six hours for most diagnostics. For precipitation and radiation diagnostics, the frequency is one hour.

### DIFF
--- a/src/diagnostics/default_diagnostics.jl
+++ b/src/diagnostics/default_diagnostics.jl
@@ -93,9 +93,12 @@ function common_diagnostics(
         start_date + Dates.Second(t_start)
     return vcat(
         map(short_names) do short_name
-            return ScheduledDiagnostic(
-                variable = get_diagnostic_variable(short_name),
-                compute_schedule_func = EveryStepSchedule(),
+            variable = get_diagnostic_variable(short_name)
+            return ScheduledDiagnostic(;
+                variable,
+                compute_schedule_func = make_compute_schedule(variable, period,
+                    start_date,
+                    date_last),
                 output_schedule_func = EveryCalendarDtSchedule(
                     period;
                     reference_date = start_date,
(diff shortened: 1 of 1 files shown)


src/diagnostics/default_diagnostics.jl
```diff
@@ -109,30 +124,66 @@ function common_diagnostics(
...
+    period,
+    start_date,
+    date_last;
+    model = nothing,
+)
+    if !(period isa Dates.Month || (period isa Dates.Day && Dates.value(period) > 1))
+        return EveryStepSchedule()
+    end
+
+    is_precip_diag = model isa Union{EquilMoistModel, NonEquilMoistModel}
```
szy21: Is this making only the precipitation and radiation diagnostics 1h, or is it making all the variables in a simulation running with EquilMoistModel or NonEquilMoistModel or AbstractRRTMGPMode 1h?

ph-kev: It is the latter. I can make it only the precipitation and radiation diagnostics 1h if this solution isn't okay.

szy21: Yes we want only the precipitation and radiation diagnostics 1h. But if it's complicated, just making everything 1h is fine.

nefrathenrici (approved):
Looks good to me

PR #4260, opened 2026-02-02
SGS Quadrature Integration and Microphysics Refactor

tapios:
## Purpose 
This PR introduces subgrid-scale (SGS) quadrature integration for microphysics tendencies and performs a major reorganization of the microphysics module structure. The SGS quadrature approach accounts for thermodynamic variability in temperature and moisture, capturing SGS variations and improving representation of phase changes.

## Key Changes
1. SGS Quadrature Integration (sgs_quadrature.jl)
  * New microphysics_tendencies_quadrature function integrates point-wise microphysics over joint PDF of (`T`, `q_tot`)
  * Supports 1M and 2M microphysics schemes via CloudMicrophysics.BulkMicrophysicsTendencies API
  * Performs quadrature over saturation adjustment for equilibrium moisture models

2. Module Reorganization
  * Consolidated microphysics tendencies into unified `tendency.jl` entry point
  * Generally consolidated microphysics calculations into microphysics directory (e.g., rather than in cache, although the cache uses the physics functions)

3. CloudMicrophysics v0.31 API Migration
  * Updated to CloudMicrophysics v0.31.1+
  * Migrated to consolidated parameter containers
  * Updated aerosol activation function signatures

4. Bulk microphysics tendencies uses MM2015 relaxation-to-equilibrium in the fused tendency API. 
  * The simple relaxation terms are currently not supported in atmos. Support for them should be added in CloudMicrophysics.
  * Implicit timestepping for the relaxation terms with `p.atmos.noneq_cloud_formation_mode == Implicit()` is not currently handled correctly and needs to be revised. 

5. Output diagnostics changes from liquid-ice potential temperature variance (and q_t covariance) to temperature variance and covariance (temperature is more natural for microphysics calculations)

## Bug Fixes 
1. Inconsistent density scaling in cloud liquid self-collection limiter: In the old 2M precipitation code, the `cloud_liquid_self_collection` limiter incorrectly divided `nₗ` by `ρ` ([Line 642](https://github.com/CliMA/ClimaAtmos.jl/blob/92ae51cc97ddd18727546650ec1d863a59de8e52/src/parameterized_tendencies/microphysics/microphysics_wrappers.jl#L642))
 
2. Aerosol `Mode_κ` constructor argument mismatch: there were mismatches in constructor arguments. These are not entirely fixed. The upstream aerosol code in CMP needs to be updated.

Closes #4262 #4293 
 

----


src/cache/diagnostic_edmf_precomputed_quantities.jl
```diff
@@ -1396,16 +1337,16 @@ NVTX.@annotate function set_diagnostic_edmf_precomputed_quantities_top_bc!(
...
             S_q_totʲ_level = Fields.field_values(Fields.level(ᶜS_q_totʲ, i_top))
             @. S_q_totʲ_level = 0
         end
         if microphysics_model isa Microphysics1Moment &&
            moisture_model isa NonEquilMoistModel
-            ᶜS_q_liqʲ = p.precomputed.ᶜSqₗᵖʲs.:($j)
-            ᶜS_q_iceʲ = p.precomputed.ᶜSqᵢᵖʲs.:($j)
-            ᶜS_q_raiʲ = p.precomputed.ᶜSqᵣᵖʲs.:($j)
-            ᶜS_q_snoʲ = p.precomputed.ᶜSqₛᵖʲs.:($j)
+            ᶜS_q_liqʲ = p.precomputed.ᶜSqₗᵐʲs.:($j)
```
szy21, line 1379: These are just renaming - do you want to change them to use the BulkMicrophysicsTendency too?

szy21: Sorry, I missed the code change. Ignore it.

src/parameterized_tendencies/microphysics/microphysics_wrappers.jl
```diff
@@ -507,286 +391,384 @@ function aerosol_activation_sources(
...
         )
-    @. Snₗᵖ = Sᵖ + S₂ᵖ
-    # rain number adjustment for mass limits
-    @. Sᵖ = CM2.number_increase_for_mass_limit(
-        mp.sb.numadj,
-        mp.sb.pdf_r.xr_max,
-        qᵣ,
-        ρ,
-        ρ * nᵣ,
+    end
```
szy21: not needed as there is a dispatch below?

src/parameterized_tendencies/microphysics/tendency.jl
```diff
@@ -0,0 +1,541 @@
...
+    (; ᶜT) = p.precomputed
+    (; ᶜSqₗᵐ, ᶜSqᵢᵐ, ᶜSqᵣᵐ, ᶜSqₛᵐ) = p.precomputed
+    (; params, dt) = p
+    thp = CAP.thermodynamics_params(params)
+    cmp = CAP.microphysics_1m_params(params)
+
+    # --- Cloud condensation/evaporation (if explicit mode) ---
+    if p.atmos.noneq_cloud_formation_mode == Explicit()
+        # Compute all 1M tendencies via BMT (cloud + precipitation)
+        result = @. BMT.bulk_microphysics_tendencies(
```
szy21: Is this the sum of all tendencies, or just cloud condensate tendency?

src/cache/precomputed_quantities.jl
```diff
@@ -493,6 +535,25 @@ NVTX.@annotate function set_implicit_precomputed_quantities!(Y, p, t)
         @. ᶜT = ᶜsa_result.T
         @. ᶜq_liq_rai = ᶜsa_result.q_liq
         @. ᶜq_ice_sno = ᶜsa_result.q_ice
+
+        # Two-pass SGS: recompute condensate using SGS quadrature over (T, q_tot)
+        # Note: In implicit phase, covariance cache may not be populated yet;
+        # get_covariances will compute on-the-fly if cache is empty
+        if microphysics_model isa QuadratureMicrophysics
+            ᶜq′q′, ᶜT′T′, ᶜT′q′ = get_covariances(Y, p, thermo_params)
+            sgs_result = @. compute_sgs_saturation_adjustment(
```
szy21: This allocates

src/cache/prognostic_edmf_precomputed_quantities.jl
```diff
@@ -863,54 +760,43 @@ NVTX.@annotate function set_prognostic_edmf_precomputed_quantities_precipitation
...
         ᶜq_rai⁰,
-        ᶜq_sno⁰,
-        ᶜρ⁰,
-        ᶜT⁰,
-        dt,
+        ᶜn_rai⁰,
+    )
+
+    # Extract tendencies with coupled limiting (preserve mass/number ratio)
+    ᶜldt = @. limited_tendency_2m(
```
szy21: This and all ᶜldt allocate, but maybe we don't need to worry too much about 2M now.

src/cache/precomputed_quantities.jl
```diff
@@ -493,6 +535,25 @@ NVTX.@annotate function set_implicit_precomputed_quantities!(Y, p, t)
         @. ᶜT = ᶜsa_result.T
         @. ᶜq_liq_rai = ᶜsa_result.q_liq
         @. ᶜq_ice_sno = ᶜsa_result.q_ice
+
+        # Two-pass SGS: recompute condensate using SGS quadrature over (T, q_tot)
+        # Note: In implicit phase, covariance cache may not be populated yet;
+        # get_covariances will compute on-the-fly if cache is empty
+        if microphysics_model isa QuadratureMicrophysics
+            ᶜq′q′, ᶜT′T′, ᶜT′q′ = get_covariances(Y, p, thermo_params)
```
szy21: I think this may also allocate - I'm actually surprised this works. I'll look at it more.

src/cache/precipitation_precomputed_quantities.jl
```diff
@@ -614,36 +623,87 @@ function set_precipitation_velocities!(
...
+        ᶜS_ρq_tot * (
+            BMT.bulk_microphysics_tendencies(
+                BMT.Microphysics0Moment(),
+                cm_params,
+                thermo_params,
+                ᶜT,
+                ᶜq_liq_rai,
+                ᶜq_ice_sno,
+            ).e_int_precip + ᶜΦ
+        )
```
szy21: These allocate a lot

szy21: (Commenting this out fixes flame graph: perf target. I can look into how to fix this later)

perf/alloc_profile.jl
szy21, line 1: remember to remove this file

config/default_configs/default_config.yml
```diff
@@ -245,9 +245,18 @@ cloud_model:
 cloud_nn_architecture:
   help: "Cloud neural network architecture to use when `cloud_model` is set to `MLCloud` [`arch_2layers_8nodes.jld2` (default), etc.]"
   value: "arch_2layers_8nodes.jld2"
-implicit_noneq_cloud_formation:
-  help: "Whether to treat the nonequilibrium cloud condensate tendency implicitly [`true`, `false` (default)]"
-  value: false
+implicit_microphysics:
```
szy21, line 248: This is a placeholder that doesn't change anything now, right? If so maybe good to make it clear in the help message.

src/cache/diagnostic_edmf_precomputed_quantities.jl
```diff
@@ -1500,99 +1451,89 @@ NVTX.@annotate function set_diagnostic_edmf_precomputed_quantities_env_precipita
     microphysics_model::Microphysics0Moment,
 )
```
szy21, line 1493: We don't use quadrature for diagnostic EDMF 0M?

src/cache/precipitation_precomputed_quantities.jl
```diff
@@ -614,36 +629,74 @@ function set_precipitation_velocities!(
...
-            dt,
-            Y.c.ρq_tot / Y.c.ρ,
-            ᶜq_tot_safe,
-            ᶜq_liq_rai,
-            ᶜq_ice_sno,
-        )
-    @. ᶜS_ρe_tot =
-        ᶜS_ρq_tot *
-        e_tot_0M_precipitation_sources_helper(thermo_params, ᶜT, ᶜq_liq_rai, ᶜq_ice_sno, ᶜΦ)
+        Y.c.ρ * limit_sink(ᶜmp_result.dq_tot_dt, Y.c.ρq_tot / Y.c.ρ, dt)
```
szy21: You use limit_sink(ᶜmp_result.dq_tot_dt, Y.c.ρq_tot / Y.c.ρ, dt, 1) in other places - do you want to do that here?

src/cache/precomputed_quantities.jl
```diff
@@ -244,13 +279,18 @@ function precomputed_quantities(Y, atmos)
...
-        ᶜgradᵥ_θ_liq_ice = Fields.Field(C3{FT}, cspace),
-    )
+    # Gradient fields for covariance computation (used in cloud fraction/microphysics)
+    # Only allocate if NOT already included via advective_sgs_quantities (PrognosticEDMFX)
+    sgs_quantities =
+        atmos.turbconv_model isa PrognosticEDMFX ? (;) :
+        (;
+            ᶜgradᵥ_q_tot = Fields.Field(C3{FT}, cspace),
+            ᶜgradᵥ_θ_liq_ice = Fields.Field(C3{FT}, cspace),
+        )
```
szy21: maybe more clear to always have sgs_quantities and remove the two fields from advective_sgs_quantities

src/parameterized_tendencies/microphysics/cloud_fraction.jl
```diff
@@ -0,0 +1,825 @@
...
+    # Compute lazy θ-based covariances (shared with diagnostics)
+    ᶜq′q′_lazy, ᶜθ′θ′_lazy, ᶜθ′q′_lazy = compute_θ_covariance(Y, p, thermo_params)
+
+    # Materialize into cache fields
+    @. ᶜq′q′ = ᶜq′q′_lazy
+    @. ᶜT′T′ = ᶜθ′θ′_lazy  # temporarily holds θ′θ′
+    @. ᶜT′q′ = ᶜθ′q′_lazy  # temporarily holds θ′q′
+
+    # Transform θ→T covariances in-place
+    ᶜ∂T_∂θ = compute_∂T_∂θ!(p.scratch.ᶜtemp_scalar_2, Y, p, thermo_params)
```
szy21: I think it's easier to read if we change it to something like
```
 ᶜ∂T_∂θ = p.scratch.ᶜtemp_scalar_2
compute_∂T_∂θ!(ᶜ∂T_∂θ, Y, p, thermo_params)
```
and remove the return in the mutating function. but up to you.

src/parameterized_tendencies/microphysics/cloud_fraction.jl
```diff
@@ -0,0 +1,825 @@
...
+    # Use cached covariances only when per-step caching is active
+    # (same condition as _cache_covariances in set_explicit_precomputed_quantities!)
+    _is_cached =
+        _uses_sgs_covariances(cloud_model, microphysics_model) && 
+            microphysics_model isa QuadratureMicrophysics
+    if _is_cached
+        # Use cached covariances (populated by set_covariance_cache!)
+        (; ᶜT′T′, ᶜq′q′, ᶜT′q′) = p.precomputed
+        return (ᶜq′q′, ᶜT′T′, ᶜT′q′)
+    else
```
szy21: should we just always cache the covariance?

src/parameterized_tendencies/microphysics/cloud_fraction.jl
```diff
@@ -0,0 +1,825 @@
...
+
+# Returns
+Tuple: `(ᶜq_liq_mean, ᶜq_ice_mean)` as lazy field expressions.
+"""
+function _get_condensate_means_equil(p, turbconv_model)
+    if turbconv_model isa PrognosticEDMFX
+        (; ᶜq_liq_rai⁰, ᶜq_ice_sno⁰) = p.precomputed
+        return ᶜq_liq_rai⁰, ᶜq_ice_sno⁰
+    else
+        (; ᶜq_liq_rai, ᶜq_ice_sno) = p.precomputed  # TODO: Check this. Shouldn't we use environment variables for DiagnosticEDMFX too?
```
szy21: we assume the environment is the same as grid-mean in diagnostic edmf

test/Project.toml
```diff
@@ -0,0 +1,398 @@
+[deps]
```
trontrytel: Do we need this file?

src/cache/diagnostic_edmf_precomputed_quantities.jl
```diff
@@ -825,27 +821,34 @@ NVTX.@annotate function set_diagnostic_edmf_precomputed_quantities_do_integral!(
...
                     microphys_0m_params,
-                    dt,
-                    q_totʲ_prev_level,
-                    q_tot_safeʲ_prev_level,
+                    thermo_params,
+                    Tʲ_prev_level,
                     q_liq_raiʲ_prev_level,
                     q_ice_snoʲ_prev_level,
                 )
+                @. S_q_totʲ_prev_level = limit_sink(
```
trontrytel, line 846: This is duplicating the number of stored variables: we now have individual `S_q_totʲ_prev_level` and `mp_result_prev_level.dq_tot_dt` ?

src/cache/diagnostic_edmf_precomputed_quantities.jl
```diff
@@ -1483,7 +1434,7 @@ end
 """
     set_diagnostic_edmf_precomputed_quantities_env_precip!(Y, p, t)
 
-Updates the precipitation sources in precomputed quantities stored in `p` for diagnostic edmfx.
+Updates the microphysics tendency precomputed quantities stored in `p` for diagnostic EDMFX.
```
trontrytel, line 1460: I think we are pretty committed at this point to support:
- equilibrium + 0M
- non-equilibrium + 1M
- non-equilibrium + 2M + P3

It would be good to get rid of the separate `moisture_model` and `microphysics_model` choices, since they are never separate.

src/cache/prognostic_edmf_precomputed_quantities.jl
```diff
@@ -448,37 +448,40 @@ NVTX.@annotate function set_prognostic_edmf_precomputed_quantities_precipitation
...
     n = n_mass_flux_subdomains(p.atmos.turbconv_model)
+    (; ᶜmp_result) = p.precomputed
     for j in 1:n
-        @. ᶜSqₜᵖʲs.:($$j) = q_tot_0M_precipitation_sources(
-            cmp,
-            dt,
-            Y.c.sgsʲs.:($$j).q_tot,
-            ᶜq_tot_safeʲs.:($$j),
+        # Materialize BMT result first to avoid NamedTuple property access in broadcast
+        @. ᶜmp_result = BMT.bulk_microphysics_tendencies(
```
trontrytel: Same comment as in diagnostic edmf - `ᶜmp_result` and `ᶜSqₜᵐʲs` duplicate the number of stored variables

src/parameterized_tendencies/microphysics/cloud_fraction.jl
```diff
@@ -0,0 +1,825 @@
...
+
+Note: For PrognosticEDMFX, gradients are computed from grid-scale variables
+rather than environmental variables. This is an approximation; ideally
+covariances would use environmental gradients since they represent sub-grid
+fluctuations within the environment.
+
+# Returns
+Tuple `(ᶜq′q′, ᶜθ′θ′, ᶜθ′q′)` of lazy field expressions.
+"""
+function compute_θ_covariance(Y, p, thermo_params)
```
trontrytel: Is this function used anywhere else than in `compute_covariance`? If not, maybe we don't need to define a separate one?

src/parameterized_tendencies/microphysics/cloud_fraction.jl
```diff
@@ -0,0 +1,825 @@
...
+    (; ᶜgradᵥ_q_tot, ᶜgradᵥ_θ_liq_ice) = p.precomputed
+
+    # Compute gradients for non-EDMF cases (EDMF gradients are precomputed)
+    if isnothing(turbconv_model)
+        needs_gradients =
+            p.atmos.call_cloud_diagnostics_per_stage isa CallCloudDiagnosticsPerStage ||
+            p.atmos.microphysics_model isa QuadratureMicrophysics ||
+            p.atmos.cloud_model isa Union{QuadratureCloud, MLCloud}
+        if needs_gradients
+            (; ᶜT, ᶜq_tot_safe, ᶜq_liq_rai, ᶜq_ice_sno) = p.precomputed
```
trontrytel, line 84: Is `ᶜq_tot_safe` some clipped version of `q_tot`? If so, should we do the same for `q_liq_rai`, `q_ice_sno`?

szy21: q_liq_rai and q_ice_sno are also clipped (I could have used a better name)

trontrytel: I like the `_safe` as a clear sign that this is a clipped variable. But maybe we can look at the variable names later. See if we want to change to the convention in CM.jl or not

test/parameterized_tendencies/microphysics/bmt_integration.jl
```diff
@@ -0,0 +1,275 @@
...
+                    FT(0.0),     # q_ice
+                    FT(0.0),     # q_rai
+                    FT(0.0),     # q_sno
+                )
+
+                # With no hydrometeors, tendencies should be zero or near-zero
+                @test isfinite(result.dq_lcl_dt)
+                @test isfinite(result.dq_icl_dt)
+                @test isfinite(result.dq_rai_dt)
+                @test isfinite(result.dq_sno_dt)
```
szy21:
```suggestion
                @test result.dq_rai_dt == 0
                @test result.dq_sno_dt == 0
```

src/cache/precipitation_precomputed_quantities.jl
```diff
@@ -166,8 +172,8 @@ function set_precipitation_velocities!(
         @. ᶜimplied_env_mass_flux -=
             Y.c.sgsʲs.:($$j).ρa * Y.c.sgsʲs.:($$j).q_liq * ᶜwₗʲs.:($$j)
     end
-    # average
-    @. ᶜwₗ = ifelse(ᶜρχ > ϵ_numerics(FT), ᶜwₗ / ᶜρχ, FT(0))
+    # average (clamp to prevent spurious negatives from numerical errors at low mass)
+    @. ᶜwₗ = ifelse(ᶜρχ > sqrt(eps(FT)), max(ᶜwₗ / ᶜρχ, zero(ᶜwₗ / ᶜρχ)), zero(ᶜwₗ))
```
trontrytel: Here and below, why not use `ϵ_numerics`?

test/parameterized_tendencies/microphysics/cloud_fraction.jl
```diff
@@ -0,0 +1,123 @@
...
+                    @test cf > FT(0.99)
+                end
+
+                @testset "Zero variance, nonzero condensate → cf = 1" begin
+                    cf = CA.compute_cloud_fraction_sd(
+                        thp, T, ρ,
+                        FT(1e-3), FT(0),
+                        FT(0), FT(0), FT(0),
+                    )
+                    @test cf > FT(0.9)
```
szy21:
```suggestion
                    @test cf > FT(0.99)
```

test/parameterized_tendencies/microphysics/cloud_fraction.jl
```diff
@@ -0,0 +1,123 @@
...
+                        FT(1e-3), FT(5e-4),
+                        FT(1.0), FT(1e-6), FT(0),
+                    )
+                    cf_liq = CA.compute_cloud_fraction_sd(
+                        thp, FT(260.0), ρ,
+                        FT(1e-3), FT(0),
+                        FT(1.0), FT(1e-6), FT(0),
+                    )
+                    # Max overlap: combined cf ≥ liquid-only cf
+                    @test cf_both >= cf_liq - eps(FT)
```
szy21:
```suggestion
                    @test cf_both >= cf_liq
```

src/cache/precipitation_precomputed_quantities.jl
tapios: ϵ_numerics is too small; it gave errors in unit tests here at low mass. I can try a compromise if you think 1e-4 is too large.

src/parameterized_tendencies/microphysics/cloud_fraction.jl
tapios: It's used in diagnostics. I had thought about (and would prefer) to change diagnostics to T'T' (rather than theta_li^2), but didn't want to do it here in case people are looking at it.

src/cache/prognostic_edmf_precomputed_quantities.jl
tapios: Yes, I know. It's one of the items in #4285 to change that (again, I tried to minimize number of changes here :)

test/parameterized_tendencies/microphysics/tendency.jl
szy21, line 1: I'm not sure where the numbers for allocation in this file come from. We can keep them here now, but maybe move the allocation test to a different place later.

test/parameterized_tendencies/microphysics/tendency_limiters.jl
```diff
@@ -0,0 +1,376 @@
...
+            @test limit(-0.001, dt, 2) == 0.0
+        end
+
+        @testset "type stability" begin
+            @test eltype(limit(Float32(q), Float32(dt), 2)) == Float32
+            @test eltype(limit(Float64(q), Float64(dt), 2)) == Float64
+        end
+    end
+
+    @testset "triangle_inequality_limiter()" begin
```
szy21: Is triangle_inequality_limiter still used anywhere?

test/restart.jl
```diff
@@ -104,6 +104,15 @@ function test_restart(test_dict; job_id, comms_ctx, more_ignore = Symbol[])
...
             :data_handler,
+            # Scratch field for microphysics NamedTuple results
+            :ᶜmp_result,
+            # Covariance fields depend on scratch state
+            :ᶜT′T′,
+            :ᶜq′q′,
+            :ᶜT′q′,
+            # RRTMGP internal arrays may differ due to RNG state
+            :rrtmgp_model,
+            :ᶠradiation_flux,
```
szy21: did it fail if you don't ignore these two fields? It would be good to keep them so we know radiation is reading the correct cloud field. If it fails, it may be because I forgot to add radiation_reset_rng_seed: true in the restart test (although for our initial condition, there should be no cloud in the restart test).

test/restart.jl
```diff
@@ -144,6 +153,15 @@ function test_restart(test_dict; job_id, comms_ctx, more_ignore = Symbol[])
...
             :rc,
+            # Scratch field for microphysics NamedTuple results
+            :ᶜmp_result,
+            # Covariance fields depend on scratch state
+            :ᶜT′T′,
+            :ᶜq′q′,
+            :ᶜT′q′,
+            # RRTMGP internal arrays are not deterministic through fill_with_nans!
+            :rrtmgp_model,
+            :ᶠradiation_flux,
```
szy21: same here, it would be good to keep. I can take a look at this.

src/cache/precipitation_precomputed_quantities.jl
trontrytel: 1e-4 seems large for microphysics tracers. In the tests I was doing, the maximum value of those tracers was somewhere between 1e-3 and 1e-4

src/parameterized_tendencies/microphysics/cloud_fraction.jl
szy21: yeah, when I added them I use q_tot_safe because there is already q_tot (from specific). I use e.g. q_liq_rai because there is only q_liq or q_rai from specific. but happy to change them

src/cache/precomputed_quantities.jl
```diff
@@ -128,42 +128,75 @@ function precomputed_quantities(Y, atmos)
...
         sfc_conditions = similar(Spaces.level(Y.f, half), SCT),
     )
-    cloud_diagnostics_tuple =
-        similar(Y.c, @NamedTuple{cf::FT, q_liq::FT, q_ice::FT})
     # Cloud fraction is used to calculate buoyancy gradient, so we initialize it to 0 here.
-    @. cloud_diagnostics_tuple.cf = FT(0)
+    ᶜcloud_fraction = similar(Y.c, FT)
+    @. ᶜcloud_fraction = FT(0)
+
+    # SGS covariances for cloud fraction (S&D) and microphysics quadrature
```
trontrytel: What does S&D stand for?

src/cache/precipitation_precomputed_quantities.jl
tapios: I want limiting of sources and sinks for other tendencies. 0M is special because it's just a sink.

szy21: No I just meant for EDMF updraft and environment with 0M you are passing in n = 1 as an argument, but here you are not, so it's using the default n = 3.

src/cache/precomputed_quantities.jl
tapios: Sommeria & Deardorff. I'll add it.

src/cache/precipitation_precomputed_quantities.jl
tapios: n = 1 seems extreme. It means we would allow depleting all atmospheric water in one time step. A less wide bound seems better. (Though in practice, this should never matter, with typical autoconversion timescales.)

tapios: I'll change the other places to also use the default.

test/parameterized_tendencies/microphysics/tendency_limiters.jl
tapios: No. We can remove it. I just left it there for reference for now.

test/parameterized_tendencies/microphysics/tendency.jl
tapios: Trial and error. You mean move to buildkite CI? Or a different file. I find it helpful to have allocation tests in the unit tests, where they can be run quickly.

src/cache/precipitation_precomputed_quantities.jl
tapios: I'll try to see how small we can make it. The results for this evaluation were platform dependent when the limit was too small.

src/cache/diagnostic_edmf_precomputed_quantities.jl
tapios: Good catch! (I had missed it.) Thank you. I added it now.

src/cache/diagnostic_edmf_precomputed_quantities.jl
tapios: Good point. I'll add it to the refactor issue.

src/parameterized_tendencies/microphysics/cloud_fraction.jl
tapios: added to issue

src/parameterized_tendencies/microphysics/cloud_fraction.jl
tapios: The conditional was there (and needed) because of this `call_cloud_diagnostics_per_stage` logic. But since we treat this is true anyway, we can remove those checks and always cache covariances. 

I removed `call_cloud_diagnostics_per_stage` and always use the cache for the covariances.

test/parameterized_tendencies/microphysics/tendency.jl
szy21: no I meant a different file in unit test that tests the allocations. But we don't need to change it now.

src/cache/diagnostic_edmf_precomputed_quantities.jl
tapios: Yes, will fix in #4285

src/parameterized_tendencies/microphysics/cloud_fraction.jl
tapios: I changed diagnostics to temperature based and simplified to covariances to having them always cached.

src/parameterized_tendencies/microphysics/tendency_limiters.jl
```diff
@@ -0,0 +1,317 @@
...
+- `q_source`: Available source quantity for this species [kg/kg]
+- `q_sink`: Available sink quantity (species being depleted) [kg/kg]
+- `dt`: Timestep [s]
+- `n_sink`: Number of timesteps for sink depletion (default: 5)
+- `n_source`: Number of timesteps for source depletion (default: 20)
+
+# Example
+```julia
+# Limit BMT tendency: sources bounded by sat_excess + partner, sinks by species
+Sqₗᵐ = smooth_tendency_limiter(dq_lcl_dt, sat_excess + q_ice, q_liq, dt)
```
trontrytel: Taking cloud liquid water as an example, cloud ice is a possible source of liquid water but only in temperatures warmer than freezing. Otherwise , due to WBF process, its actually a sink for water.

Similarly in collisions, the available source quantity is made of different species depending on the temperature.

src/parameterized_tendencies/microphysics/tendency_limiters.jl
```diff
@@ -0,0 +1,317 @@
...
+- `S`: Tendency [kg/kg/s], can be positive (source) or negative (sink)
+- `q_source`: Available source quantity for this species [kg/kg]
+- `q_sink`: Available sink quantity (species being depleted) [kg/kg]
+- `dt`: Timestep [s]
+- `n_sink`: Number of timesteps for sink depletion (default: 5)
+- `n_source`: Number of timesteps for source depletion (default: 20)
+
+# Example
+```julia
+# Limit BMT tendency: sources bounded by sat_excess + partner, sinks by species
```
trontrytel: Taking cloud liquid water as an example, if we limit the bulk cloud liquid tendency, shouldn't we also limit the corresponding tendency to other species? Or is the idea to treat water vapor as a garbage collector and assume that all the mismatch from scaling one tendency but not the other will be masked by it?

src/prognostic_equations/remaining_tendency.jl
```diff
@@ -279,17 +279,7 @@ NVTX.@annotate function additional_tendency!(Yₜ, Y, p, t)
...
-            Yₜ,
-            Y,
-            p,
-            p.atmos.moisture_model,
-            p.atmos.microphysics_model,
-            p.atmos.turbconv_model,
-        )
-    end
-
+    # EDMF updraft microphysics tendencies (applied to updraft prognostic variables)
```
trontrytel, line 283: Should we change `edmfx_precipitation_tendency!` to `edmfx_microphysics_tendency` to be consistent with other name changes?

src/parameterized_tendencies/microphysics/tendency.jl
```diff
@@ -0,0 +1,316 @@
...
+        @. Yₜ.c.ρq_sno += Y.c.sgsʲs.:($$j).ρa * ᶜSqₛᵐʲs.:($$j)
+    end
+end
+
+#####
+##### QuadratureMicrophysics
+#####
+
+# For NonEquilMoistModel + QuadratureMicrophysics{Microphysics1Moment}:
+# The cache (ᶜSqₗᵐ, etc.) already includes ALL tendencies (cloud + precip)
```
trontrytel: Is that the pattern we want for the other tendencies - to do the summation from different sub-domains in the cache?

src/cache/precipitation_precomputed_quantities.jl
tapios: Changed ϵ_numerics to cbrt(floatmin), which seems to work.

src/parameterized_tendencies/microphysics/tendency.jl
tapios: Seems easiest to me. Do you have other preferences?

src/parameterized_tendencies/microphysics/tendency_limiters.jl
tapios: I have gone through various iterations of treating vapor as gc, or preserving sums of condensate tendencies. It doesn't seem to matter much in practice--the bottom line is, results are always sensitive to limiter details. So I opted for simply for now (pending an implicit solver, where hopefully this becomes irrelevant).

src/parameterized_tendencies/microphysics/tendency_limiters.jl
tapios: I changed the docstring. This is from earlier attempts to mimic what we had before for individual tendencies. Now I'm simply using q_tot as the total reservoir, with large enough n--it's a simple metric.

src/parameterized_tendencies/microphysics/microphysics_wrappers.jl
```diff
@@ -209,167 +93,160 @@ function compute_precipitation_sources!(
...
+    # Temperature-rate limiter:
+    # Phase changes imply temperature tendencies via latent heat release/absorption
+    Lv_over_cp = TD.Parameters.LH_v0(tps) / TD.Parameters.cp_d(tps)
+    Ls_over_cp = TD.Parameters.LH_s0(tps) / TD.Parameters.cp_d(tps)
+
+    # max 5 K temperature change per timestep
+    # TODO: Arbitrary limit; remove when microphysics is implicitly handled
+    dT_dt_max = FT(5) / dt
+
+    dT_dt = Lv_over_cp * dq_lcl_dt + Ls_over_cp * dq_icl_dt
```
trontrytel: Not all cloud liquid/ice changes are phase changes

src/parameterized_tendencies/microphysics/microphysics_wrappers.jl
```diff
@@ -209,167 +93,160 @@ function compute_precipitation_sources!(
...
+
+# Returns
+NamedTuple with limited tendencies: `(dq_lcl_dt, dq_icl_dt, dq_rai_dt, dq_sno_dt)`
+"""
+@inline function apply_1m_tendency_limits(mp_result, tps, q_tot, q_liq, q_ice, q_rai, q_sno, dt)
+    FT = typeof(q_tot)
+    q_vap = q_tot - q_liq - q_ice - q_rai - q_sno
+
+    # Mass-conservation limits
+    dq_lcl_dt = smooth_tendency_limiter(mp_result.dq_lcl_dt, q_vap + q_ice, q_liq, dt)
```
trontrytel: Should the limiters affect the Jacobian terms? Or is the idea to switch those limiters off when we move everything to implicit?

src/parameterized_tendencies/microphysics/microphysics_wrappers.jl
```diff
@@ -209,167 +93,160 @@ function compute_precipitation_sources!(
...
+    dq_icl_dt = smooth_tendency_limiter(mp_result.dq_icl_dt, q_vap + q_liq, q_ice, dt)
+    dq_rai_dt = smooth_tendency_limiter(mp_result.dq_rai_dt, q_liq + q_sno, q_rai, dt)
+    dq_sno_dt = smooth_tendency_limiter(mp_result.dq_sno_dt, q_ice + q_rai, q_sno, dt)
```
trontrytel: It depends on the temperature whether `q_rai` or `q_ice` are a source for snow. Similar with the other processes. Maybe it would be better to just limit by the sink in each of them?

I'm still worried that we limit the tendency to one tracer without limiting the corresponding tendency for another tracer involved in this process. I hope that assuming that water vapor can accommodate for the mismatch is a good assumption

src/parameterized_tendencies/microphysics/microphysics_wrappers.jl
tapios: Right. But I think the equivalent temperature change is a convenient metric for deciding when a change is too large.

src/parameterized_tendencies/microphysics/microphysics_wrappers.jl
tapios: We shouldn't need them when the microphysics are implicit. But it's good to keep them as a defensive strategy, just with generous limits.

src/parameterized_tendencies/microphysics/sgs_quadrature.jl
```diff
@@ -0,0 +1,614 @@
...
+"""
+struct SGSQuadrature{N, A, W, D <: AbstractSGSDistribution, FT} <: AbstractSGSamplingType
+    a::A             # quadrature points
+    w::W             # quadrature weights
+    dist::D          # distribution type
+    default_corr::FT # default T-q correlation when covariance is zero
+    T_min::FT        # minimum temperature for physical validity [K]
+    function SGSQuadrature(
+        ::Type{FT};
+        quadrature_order = 2,
```
trontrytel: We are setting `quadrature_order = 2` as default here and in the `default_config`. But it is set to 3 in `solver/types.jl`

src/parameterized_tendencies/microphysics/tendency.jl
trontrytel: No, this looks good. I'll clean up in the next PR

src/parameterized_tendencies/microphysics/sgs_quadrature.jl
tapios: Good catch. Thanks!

src/parameterized_tendencies/microphysics/microphysics_wrappers.jl
tapios: I changed this to uniform rescaling. (Makes not much of a difference though how it's done.)

test/restart.jl
tapios: Works without this now.

test/restart.jl
tapios: Works without this now. Removed it.

src/diagnostics/core_diagnostics.jl
```diff
@@ -902,9 +902,13 @@ function compute_pr!(
...
         NoPrecipitation,
         Microphysics0Moment,
+        QuadratureMicrophysics{<:Microphysics0Moment},
         Microphysics1Moment,
+        QuadratureMicrophysics{<:Microphysics1Moment},
         Microphysics2Moment,
+        QuadratureMicrophysics{<:Microphysics2Moment},
         Microphysics2MomentP3,
+        QuadratureMicrophysics{<:Microphysics2MomentP3},
     },
```
szy21, line 912: maybe this is everything and I can just remove the check? I'll try

szy21: actually let's just clean it up in the future, when we unify microphysics_model and moisture_model.

src/parameterized_tendencies/microphysics/tendency.jl
```diff
@@ -0,0 +1,366 @@
...
+
+    # Tendencies are already limited in the cache; just scale by density
+    # specific sources are for the environment, so we need to multiply by ρ_env
+    # ρ_env = ρ - sum(ρa_up)
+    n = n_mass_flux_subdomains(p.atmos.turbconv_model)
+    ᶜρa_up_sum = p.scratch.ᶜtemp_scalar
+    @. ᶜρa_up_sum = 0
+    for j in 1:n
+        @. ᶜρa_up_sum += ᶜρaʲs.:($$j)
+    end
```
szy21: do we want to do this for diagnostic edmf (in many other places, we are assuming environment \rhoa = \rho)? If you want to do this, you can use the ρa⁰ function (with `@lazy`).

tapios: It's ok to assume that environmental scalar = grid-mean scalar. It's an odd assumption for rho a, because it violates the subdomain decomposition assumption and leads to double counting of tendencies (e.g., here, if updraft and environmental tendencies were the same, we'd effectively add the updraft tendency twice).

PR #4263, opened 2026-02-03
Add 1D negative tracers statistics to output

trontrytel:
Adds a way to output a vertical profile of minimum value of a tracer, mean negative value of a tracer and a fraction of the domain that has negative tracer. 

Requires ClimaDiagnostics v0.3.1

### DIFF
--- a/src/diagnostics/negative_scalars_diagnostics.jl
+++ b/src/diagnostics/negative_scalars_diagnostics.jl
@@ -1,0 +1,156 @@
+# Aggregating the information about negative tracers at each vertical level
+
+import Statistics: mean
+
+"""
+    compute_min_per_level!(out, field)
+
+Computes the minimum value of a field at each vertical level and stores it in `out`.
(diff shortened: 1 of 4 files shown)


src/diagnostics/negative_scalars_diagnostics.jl
```diff
@@ -0,0 +1,145 @@
...
+        field_level = Fields.level(field, i)
+        Fields.level(out′, i) .= minimum(field_level)
+    end
+    return out′
+end
+
+"""
+    compute_negative_mean_per_level!(out, field)
+
+Computes the mean of negative values of a field at each vertical level.
```
szy21, line 22: The mean includes the zeros so this is not the mean of negative values (i.e. it is affected by the negative fraction), right?

trontrytel: Fixed! Thank you

Issue #4040, opened 2025-10-02
Output `yaml` file should point to output `toml` file


haakon-e:
To allow easier restarting from existing runs, it would be convenient if the `yaml` file in the output directory point to the `toml` file that is stored in the same directory.

Currently, the `yaml` file points to whatever path was specified in the original run. If that original file is moved, then restarting from a `yaml` file will fail.

AmanKashyap0807:
Hi @haakon-e and CliMA team,

I’m interested in working on this issue and had a couple of questions before I get started:

Is there any ongoing work or existing branch/PR related to this already?
Do you have a preferred approach for how the output yaml should reference the output toml (e.g., relative path within the output directory, any specific conventions to follow)?
If it’s still available, I’d be happy to put together a PR addressing this.

haakon-e:
Hi @AmanKashyap0807,

You're very welcome to contribute a PR to fix this issue, thank you for wanting to contribute!
To my knowledge, there is currently no ongoing work to address this.

I think it'd be safest to write the full path, as I think a relative directory would be interpreted by Julia as relative to where the Julia process is launched, which may vary.

Note that the toml file in the output directory, typically is named `<job_id>_parameters.toml`, where `<job_id>` is a string provided during model initialization, and corresponds to the parent output directory. For example, if you're running a simulation named `bomex_box`, you'd get an output structure like so:

```
# -v- contains all simulation output
/home/user/ClimaAtmos.jl/output/bomex_box/output_0000/
# -v- a copy of the final yaml file used to run the simulation
/home/user/ClimaAtmos.jl/output/bomex_box/output_0000/bomex_box.yml  
# -v- a copy of the toml file that we want to reference
/home/user/ClimaAtmos.jl/output/bomex_box/output_0000/bomex_box_parameters.toml
```

Currently, `bomex_box.yml` would contain a line like:
```yaml
toml:
  - "toml/bomex_box.toml"
```
I think a PR should modify that to something like:
```yaml
toml:
  - "/home/user/ClimaAtmos.jl/output/bomex_box/output_0000/bomex_box_parameters.toml"
```

Note that it is possible to provide a list of toml files to initialize the simulation, but I don't do this in practice. I'll let @nefrathenrici comment if there is any special care needed to handle that case, or if referring to a single toml file from the output directory is sufficient.

nefrathenrici:
[...]

PR #4264, opened 2026-02-03
Updated the yaml output to reference the toml file in the output

AmanKashyap0807:
directory to prevent restart failures when original files are moved

## Purpose 
This PR addresses issue #4040. Currently, the output YAML file generated after a simulation references the paths of the original input TOML files. If those original files are moved or deleted, restarting the simulation from the output YAML fails.

This change ensures that the output YAML points to the absolute path of the parameters TOML file that is automatically saved in the output directory ( <job_id>_parameters.toml ). This makes the output folder self-contained and allows for robust restarts even if the original input files are modified or removed.

## To-do
- Modify `get_simulation` in type_getters.jl to update the toml path in output config
- use abspath to ensure the path is unambiguous regardless of the execution directory

## Content
The fix is implemented in `type_getters.jl` withing the get_simulation function.
Prior to writing the output YAML file, I created a copy of config.parsed_args to avoid modifying the runtime configuration object (which seemed safer than in-place mutation). In this copy, I updated the toml key to be a list containing the absolute path to the local `*_parameters.toml` file.

As suggested by @haakon-e and @nefrathenrici , I used the full absolute path (abspath(...)) for the new TOML entry. This avoids potential ambiguity with relative paths depending on where the Julia process is launched.

## Verification Step:
I verified this manually using a small "box" simulation configuration:

1. Ran a simulation using an existing TOML file ( box_density_current_test.toml ) as input.
2. Verified that the generated output YAML now points to the file in the output directory instead of the original input file.
3. Renamed the original input TOML file to simulate deletion.
4. Successfully restarted the simulation using the generated output YAML 

Let me know if any adjustment are needed, or if you want any test script.

----

### DIFF
--- a/src/solver/type_getters.jl
+++ b/src/solver/type_getters.jl
@@ -988,12 +988,16 @@ function get_simulation(config::AtmosConfig)
     output_dir = sim_info.output_dir
     @info "Simulation info" job_id output_dir
 
+    output_toml_file = joinpath(output_dir, "$(job_id)_parameters.toml")
     CP.log_parameter_information(
         config.toml_dict,
-        joinpath(output_dir, "$(job_id)_parameters.toml"),
+        output_toml_file,
         strict = config.parsed_args["strict_params"],
     )
-    YAML.write_file(joinpath(output_dir, "$job_id.yml"), config.parsed_args)
+
+    output_args = copy(config.parsed_args)
+    output_args["toml"] = [abspath(output_toml_file)]
+    YAML.write_file(joinpath(output_dir, "$job_id.yml"), output_args)
 
     if sim_info.restart
         s = @timed_str begin
--- a/test/test_output_yaml_path.jl
+++ b/test/test_output_yaml_path.jl
@@ -1,0 +1,39 @@
+import ClimaAtmos as CA
+import YAML
+using Test
+
+@testset "Output YAML Path Test" begin
+    mktempdir() do output_dir
+        job_id = "test_yaml_path_fix"
+
+        input_toml_path = joinpath(@__DIR__, "..", "toml", "bomex_box_rhoe.toml")
+
+        config_dict = Dict(
+            "job_id" => job_id,
+            "output_dir" => output_dir,
+            "config" => "box",
+            "toml" => [input_toml_path],
+            "x_elem" => 2,
+            "y_elem" => 2,
+            "z_elem" => 4,
+            "dt" => "1s",
+            "t_end" => "2s",
+            "dt_save_state_to_disk" => "2s",
+        )
+
+        config = CA.AtmosConfig(config_dict; job_id)
+
+        simulation = CA.get_simulation(config)
+        real_output_dir = simulation.output_dir
+
+        output_yaml_path = joinpath(real_output_dir, "$(job_id).yml")
+        output_toml_path = joinpath(real_output_dir, "$(job_id)_parameters.toml")
+
+        @test isfile(output_yaml_path)
+        @test isfile(output_toml_path)
+
+        yaml_data = YAML.load_file(output_yaml_path)
+        @test yaml_data["toml"] == [abspath(output_toml_path)]
+        @test yaml_data["toml"] != [input_toml_path]
+    end
+end
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -33,6 +33,8 @@ if TEST_GROUP in ("infrastructure", "all")
     @safetestset "Variable manipulations" begin @time include("variable_manipulations_tests.jl") end
     @safetestset "Parameter tests" begin @time include("parameter_tests.jl") end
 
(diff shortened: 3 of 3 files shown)


src/solver/type_getters.jl
```diff
@@ -995,7 +995,11 @@ function get_simulation(config::AtmosConfig)
         joinpath(output_dir, "$(job_id)_parameters.toml"),
```
haakon-e: It looks like you could make a variable name like:
```Julia
output_toml_file = joinpath(output_dir, "$(job_id)_parameters.toml")
```
and reference that both in the call to `CP.log_parameter_information`, and the code you wrote below (but with `abspath(output_toml_file)`. That way, it's slightly clearer that we're referring to the same file.

AmanKashyap0807: Thanks for the suggestion, That makes sense
I’ll update the code to introduce `output_toml_file` and reuse it in both places.

test/test_output_yaml_path.jl
```diff
@@ -0,0 +1,58 @@
...
+using Logging
+
+ClimaComms.@import_required_backends
+
+device = ClimaComms.device()
+context = ClimaComms.context(device)
+ClimaComms.init(context)
+
+@testset "Output YAML Path Test" begin
+    output_dir = mktempdir()
```
haakon-e: Could you rewrite this as
```suggestion
    mktempdir() do output_dir
```
then indent the rest of the test, and replace `rm(output_dir, recursive=true)`by `end`.
The `do`-block cleans up the temporary directory automatically.

test/test_output_yaml_path.jl
```diff
@@ -0,0 +1,58 @@
...
+import YAML
+import ClimaComms
+using Test
+using Logging
+
+ClimaComms.@import_required_backends
+
+device = ClimaComms.device()
+context = ClimaComms.context(device)
+ClimaComms.init(context)
```
haakon-e:
```suggestion
import ClimaAtmos as CA
import YAML
using Test
```
I think you only need these three.

test/test_output_yaml_path.jl
```diff
@@ -0,0 +1,58 @@
...
+        "t_end" => "2s",
+        "dt" => "1s",
+        "dt_save_state_to_disk" => "2s",
+        "toml" => [input_toml_path],
+        "x_max" => 1e4,
+        "y_max" => 1e4,
+        "z_max" => 1e4,
+        "x_elem" => 4,
+        "y_elem" => 4,
+        "z_elem" => 10,
```
haakon-e: I think a simpler config would suffice:

```suggestion
        "job_id" => job_id,
        "output_dir" => output_dir,
        "toml" => [input_toml_path],
        # Simple grid for quick testing
        "config" => "box",
        "x_elem" => 2,
        "y_elem" => 2,
        "z_elem" => 4,
```

test/test_output_yaml_path.jl
```diff
@@ -0,0 +1,58 @@
...
+ClimaComms.init(context)
+
+@testset "Output YAML Path Test" begin
+    output_dir = mktempdir()
+    job_id = "test_yaml_path_fix"
+    
+    input_toml_path = joinpath(output_dir, "input.toml")
+    open(input_toml_path, "w") do io
+        print(io, "")
+    end
```
haakon-e: For simplicity, here you can just refer one of our existing toml files. E.g.:

```suggestion
    input_toml_path = "toml/bomex_box_rhoe.toml"
```

test/test_output_yaml_path.jl
haakon-e, line 1: Can you add something like
```
@safetestset "Check TOML path" begin @time include("test_output_yaml_path.jl") end
```
in `test/runtests.jl` on L35 (just after the `"Parameter tests"` test), so this test becomes part of CI:

https://github.com/CliMA/ClimaAtmos.jl/blob/3028e2d49e818be8cb3e9c1baf3773bce0beafbe/test/runtests.jl#L34-L36

test/test_output_yaml_path.jl
AmanKashyap0807: Thanks for the suggestions

I’ve updated the test to use the mktempdir() do pattern for cleanup, simplified the imports, switched to an existing TOML file, reduced the simulation size, and added the test to runtests.jl so it runs in CI.

I’ve verified the test runs successfully locally. Please let me know if you would like any further changes.

haakon-e:
@AmanKashyap0807 Thanks! Can you rebase to latest main and squash your changes into one commit?

haakon-e (approved):

AmanKashyap0807:
Hi @haakon-e,

Just checking in,  it looks like all GitHub CI checks have passed, and only the `buildkite/climaatmos-ci` job is still pending.

Since auto-merge is enabled, I just wanted to ask if anything further is needed from my side.

Happy to rebase again if required.

Thanks!

PR #4267, opened 2026-02-04
Remove unneeded test/Project.toml

nefrathenrici:
This PR:
- Removes the unused and unmaintained GPU scaling pipeline
- Removes the `test/Project.toml` and adds the dependencies to `[extras]` in the base `Project.toml`.


nefrathenrici:
> I guess we still have two environment, one with `julia --project` and one with `julia --project=.buildkite`?

Yes, this only removes the test Project. We could move the `.buildkite` Project to `test` and use that for the unit tests, but it breaks for 1.10 and below with the `can not merge projects` [error](https://discourse.julialang.org/t/testing-package-where-the-test-depends-on-local-deved-package-fails-in-julia-1-4-2-with-error-can-not-merge-projects/41611)

test/parameterized_tendencies/gravity_wave/non_orographic_gravity_wave/nogw_test_single_column.jl
```diff
@@ -6,7 +6,7 @@ comparison (Jan/Apr/Jul/Oct profiles). It is NOT included in the automated test
 suite because it has no @test assertions - it only generates plots for manual verification.
 
 To run manually:
-    julia --project=test test/parameterized_tendencies/gravity_wave/non_orographic_gravity_wave/nogw_test_single_column.jl
+    julia --project test/parameterized_tendencies/gravity_wave/non_orographic_gravity_wave/nogw_test_single_column.jl
```
ph-kev, line 9: I don't think this will always work if this script is using any test dependencies. The user might need to do `using TestEnv; TestEnv.activate()`.

ph-kev (approved):

PR #4269, opened 2026-02-04
Make negative stats work on GPU, add more stats

trontrytel:
This PR makes the output work on GPUs. 

It also adds more stats like the sum of negative numbers, and the corresponding positive sums, means etc.

### DIFF
--- a/src/diagnostics/negative_scalars_diagnostics.jl
+++ b/src/diagnostics/negative_scalars_diagnostics.jl
@@ -17,6 +17,20 @@ function compute_min_per_level!(out, field)
 end
 
 """
+    compute_max_per_level!(out, field)
+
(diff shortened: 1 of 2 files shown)


src/diagnostics/negative_scalars_diagnostics.jl
```diff
@@ -90,20 +169,55 @@ const DIAGNOSTIC_TYPES = [
...
+        "Vertical Profile of Sum of Positive",
+        "kg/kg",
+        "Sum of positive",
+    ),
     (
         "_neg_frac",
         compute_negative_fraction_per_level!,
         "Vertical Profile of Negative Fraction",
-        "%",
+        "-",
```
szy21: We have been just using "" for unitless diagnostics.

PR #4271, opened 2026-02-04
clip inputs to vapor_specific_humidity in precipitation sinks

szy21:
## Purpose 

This was missed in #4231. It was clipped in Thermodynamics before, but we need to clip them in ClimaAtmos now.

## Content

----


src/parameterized_tendencies/microphysics/microphysics_wrappers.jl
sajjadazimi: Is this line missing intentionally?

szy21: nope, I missed it. Thanks!

PR #4276, opened 2026-02-06
Create IC type to allow initializing AMIP with ERA5 on model levels

costachris:
Create IC type to allow initializing AMIP with ERA5 on model levels. Add artifact which corresponds to default AMIP start date (Jan 1, 2010). Enables change of default IC for nightly and longruns in Coupler: https://github.com/CliMA/ClimaCoupler.jl/pull/1712
Also removes unused `ᶜspecific` in the precomputed cache.


config/longrun_configs/amip_target_diagedmf.yml
```diff
@@ -1,3 +1,5 @@
+initial_condition: "AMIPFromERA5"
+start_date: "20100101"
```
szy21: This is the default start date so you don't need to add it

PR #4278, opened 2026-02-06
fix downgrade tests

juliasloan25:
## Purpose 
#4267 broke the downgrade tests. I think it may be because of the old CairoMakie compat, so I'm raising that here.

cc @szy21


Project.toml
```diff
@@ -79,7 +79,7 @@ YAML = "0.4"
 julia = "1.9"
 Aqua = "0.8"
 ArtifactWrappers = "0.2"
-CairoMakie = "0.10, 0.11"
+CairoMakie = "0.11 - 0.15"
 ClimaReproducibilityTests = "0.1.1"
```
ph-kev: If you want to use range specifiers, you can do
```suggestion
CairoMakie = "=0.11, 0.11 - 0.15"
ClimaReproducibilityTests = "0.1.1"
```

ph-kev: I tested this locally and CairoMakie v0.11 should work for the downgrade test.

szy21: I tried 0.11, and it seems it still failed?

PR #4279, opened 2026-02-06
Add spectra to plot-group : `AquaplanetPlots` (includes diag-EDMF / EDOnly runs) 

akshaysridhar:
Monthly averaged spectra based on available diagnostic variables (ua, ta, hus). Creates one additional pdf in artifact output group.


szy21:
Does this control what is plotted for the amip run (in coupler)? also could you add wa spectra?

szy21 (approved):

PR #4213, opened 2026-01-13
rcemipii testing

haakon-e:
config

rcemip box debug plots

hack to turn off lilly correction

viz cond/evap tendency

wip: nonnegativity element limiter

## Content

----
- [ ] I have read and checked the items on the review checklist.

PR #4286, opened 2026-02-09
fix: rcemip config

haakon-e:
## Purpose 

Improve structuring of rcemip toml file and set `zd_rayleigh` to a reasonable value given the model domain top.

This will require an update to the ref counter.


haakon-e:

- #4286 ◀
    - #4213
        - #4390

oalcabes:
LGTM! At some point we may want to think of automating the SST_mean to be in alignment with the initial condition profile given. I realized as is, it's possible to run with an initial condition with a profile of 295K, but to run that correctly you still must edit the toml file SST_mean to be  295. Probably not the most important thing, could do this change now or in a future edit (or just keep it as is and put in a reminder to change it).

oalcabes (approved):

PR #4287, opened 2026-02-09
Use MacOS runners only on PR merges

ph-kev:
From https://docs.github.com/en/actions/reference/limits#job-concurrency-limits-for-github-hosted-runners, the concurrency limit for MacOS is 5 across the entire organization. This PR limits CI testing for MacOS to only PR merges.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,7 @@
 name: ci
 on:
   push:
+    branches: [main]
     tags: '*'
   pull_request:
   merge_group:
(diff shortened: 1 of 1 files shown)


ph-kev:
> Thanks! so it only runs on the main branch now?

The tests for MacOS should only run when the PR is merged if I did it correctly.

.github/workflows/ci.yml
```diff
@@ -28,8 +29,10 @@ jobs:
           - '1.11'
         os:
           - ubuntu-latest
-          - macOS-latest
           - windows-latest
+          - ${{ github.event_name == 'push' && 'macos-latest' || '' }}
```
nefrathenrici, line 38: Could you add a comment explaining why we have this line?

PR #4289, opened 2026-02-10
Implicit pedmf: remove entr limits

sajjadazimi:
## Purpose 
Remove entr limits when entrainment is implicit; only entr - detr must be limited to `E-D < 1/dt`. This helps using large entrainment rates where area fraction is negligible for numerical instability.

----


src/cache/temporary_quantities.jl
```diff
@@ -22,6 +22,15 @@ function implicit_temporary_quantities(Y, atmos)
...
         ᶠtemp_UVWxUVW = Fields.Field(typeof(uvw_vec * uvw_vec'), face_space), # ᶠstrain_rate
+        temp_data_level = Fields.field_values(
+            Fields.level(Fields.Field(FT, center_space), 1),
+        ),
+        temp_data_level_2 = Fields.field_values(
+            Fields.level(Fields.Field(FT, center_space), 1),
+        ),
+        temp_data_level_3 = Fields.field_values(
+            Fields.level(Fields.Field(FT, center_space), 1),
+        ),
```
szy21, line 33: are these used?

src/cache/prognostic_edmf_precomputed_quantities.jl
```diff
@@ -346,33 +337,69 @@ NVTX.@annotate function set_prognostic_edmf_precomputed_quantities_explicit_clos
...
 
-        # If the surface buoyancy flux is positive, increase entrainment in the first cell
-        # so that the updraft area grows to at least `surface_area` within one timestep.
-        # Otherwise (stable surface), leave entrainment unchanged.
+            @. ᶜdetrʲs.:($$j) = limit_detrainment(
+                ᶜdetrʲs.:($$j),
+                draft_area(Y.c.sgsʲs.:($$j).ρa, ᶜρʲs.:($$j)),
+                dt,
+            )
+        else
```
szy21, line 356: You may want to limit entrainment before calculating detrainment, because detrainment can depend on entrainment, but it's probably not very important.

src/prognostic_equations/edmfx_entr_detr.jl
```diff
@@ -666,6 +657,8 @@ limit_entrainment(entr::FT, a, dt) where {FT} = max(
 )
 limit_detrainment(detr::FT, a, dt) where {FT} =
     max(min(detr, FT(0.9) * 1 / dt), 0)
+limit_detrainment(detr::FT, entr::FT, a, dt) where {FT} =
+    max(detr, entr - FT(0.9) * 1 / dt)
```
szy21, line 661: shouldn't this be entr + FT(0.9) * 1 / dt?

src/cache/temporary_quantities.jl
sajjadazimi: These are used in `edmfx_first_interior_entr_tendency!` and now that I moved this function to the implicit tendency, autodial only works if they are returned here. These are the list of scratch variables that will be filled by duals during autodiff computations. So any scratch variable that is used in implicit tendency must be returned here.

src/prognostic_equations/edmfx_entr_detr.jl
sajjadazimi: `E - D < 0.9 / dt` so `E - 0.9 / dt < D`. What it does is that it increase detr so that the growth rate of area fraction is not more that `0.9/dt` to keep our solver well-defined. It is also important when the solver is implicit because: `\rho a (E-D) dt = \rho a - \rho a_0 \implis \rho a = \rho a_0 / (1 - (E-D) \cdot dt)`

szy21: Oh, right, it's the maximum

PR #4293, opened 2026-02-11
Remove old code that was relying on PhasePartition

trontrytel:


trontrytel:
Very likely. I just want to have it as a reminder - I still have some obsolete code in CM.jl I need to delete once we stop using it

PR #4295, opened 2026-02-12
Vertical water borrowing : mass , energy consistency. 

akshaysridhar:


haakon-e:
We should do this for the horizontal limiter too

PR #4296, opened 2026-02-12
remove unused cache in prognostic edmf

szy21:
## Purpose 

ᶠnh_pressure₃_buoyʲs is not used anymore after TKE is changed to grid-mean. ᶠnh_pressure_dragʲs is only used in the nh pressure tendency, so we don't need to cache it. This PR removes them.

## Content

----


src/cache/precomputed_quantities.jl
sajjadazimi: These are scratch matrices used in Jacobian computations. Maybe we can move them to scratch.

szy21: will do

Issue #4234, opened 2026-01-23
Separate model setup from core atmosphere model


tapios:
# The Climate Modeling Alliance
### Purpose
Currently, the ClimaAtmos source entangles core model code and setup specific to certain simulations. This makes it very hard to configure new simulations (what needs to be specified is scattered over the code base) and make the core code difficult to read, improve, and maintain. 

It's important that the concerns of the model itself and specific setups are separated. This SDI achieves this for initial conditions, while at the same time considerably simplifying the code structure (e.g., functions in favor of layered structs). It can be expanded to include, thereafter, boundary conditions and forcing functions for specific setups. 

### Cost/Benefits/Risks

The key benefit will be clearer code that will be much easier to develop further and maintain. A risk is that this will impact coupling of model components and will require some work to re-configure this. (I believe this is modest and ultimately will benefit the coupling too). 

### People and Personnel
- Lead: @tapios 
- Collaborators:
- Reviewers: @nefrathenrici  @dennisYatunin @szy21 @haakon-e 

#### Core Simplification
Current: `initial_condition → LocalState → atmos_center_variables → Y`
Proposed: `initial_condition → NamedTuple of prognostic variables → Y`

#### Output from `initial_condition`

Variable | Description | Required?
-- | -- | --
ρ | Density | Always
ρe_tot | Total energy density | Always
uₕ | Horizontal velocity | Always
ρq_tot | Total moisture | If moist
ρq_liq, ρq_ice | Condensate | If nonequil
ρq_rai, ρq_sno | Precipitation | If precip model
ρtke | TKE | If PrognosticTKE (default: 0)
SGS variables | Updraft area, MSE, etc. | If prognostic EDMFX (default: 0)
Tracer variables | ρχ for each tracer | If tracers defined

**Note**: turbconv will be renamed to PrognosticTKE

### API Design

```
# setups/Bomex/setup.jl
struct Bomex end
# Center variables (cell centers)
function center_initial_condition(::Type{Bomex}, params, local_geometry; prognostic_tke=false)
    FT = eltype(params)
    (; z) = local_geometry.coordinates
    # ... compute prognostic variables
    return (; ρ, ρe_tot, uₕ, ρq_tot)
end
# Face variables (cell faces) - vertical velocity only
function face_initial_condition(::Type{Bomex}, params, local_geometry)
    FT = eltype(params)
    return (; u₃ = Geometry.WVector(FT(0)))
end
# Surface/boundary conditions
[...]

PR #4297, opened 2026-02-12
Add Setups module to replace ICs and unify config setup

nefrathenrici:
This PR implements part of #4234 
## PR Summary
This PR introduces the Setups module (src/setups/) as the new canonical location for simulation initial conditions, replacing the legacy InitialConditions module. It uses a two-layer architecture:

- Layer 1 (physics-only): Each setup implements`center_initial_condition`, `face_initial_condition`, and `surface_condition`, returning plain NamedTuples with no knowledge of the atmos model. Setups can also implement optional SCM forcing methods: `subsidence_forcing`, `large_scale_advection_forcing`, and `coriolis_forcing` to completely unify cases like Bomex, Rico, DYCOMS etc. This will probably be further extended for other setup needs.
- Layer 2 (model-aware, setup-independent assembly): prognostic_variables.jl converts the physical state NamedTuple into model-specific prognostic variables by dispatching on moisture, microphysics, and turbconv model types.

**ICs migrated so far:** Bomex, Rico, GCMDriven (3 of ~27 IC types)

### New files created within src/setups:

- Setups.jl: module definition, abstract type, interface functions, initial_state glue
- common.jl: physical_state, perturb_coeff, thermodynamic helpers
- hydrostatic.jl: utilities duplicated from ICs: hydrostatic_pressure_profile, ColumnInterpolatableField
- prognostic_variables.jl: Layer 2 assembly with dispatch on all model types
- overwrite_from_file.jl: shared infrastructure for field-level ICs
- Bomex.jl, Rico.jl, GCMDriven.jl: individual setup implementations

### Integration path in type_getters.jl and model_getters.jl:

- get_setup_type() returns a Setups.X(...) or nothing
  - When setup is not nothing, use setup to build initial state and surface state
  - When nothing: falls through to legacy get_initial_condition / get_surface_setup
- get_scm_coriolis, get_subsidence_model, and get_large_scale_advection_model all take an optional Setup used to set the proper config.

### Remaining IC Types to Migrate:

Pointwise column ICs with prognostic_tke
- GABLS
- Soares
- DYCOMS_RF01
- DYCOMS_RF02
- TRMM_LBA
- SimplePlume
- ISDAC (also takes perturb_initstate)

Pointwise ICs with perturb_initstate
- DryBaroclinicWave (also takes deep_atmosphere)
- MoistBaroclinicWave (also takes deep_atmosphere)
- MoistBaroclinicWaveWithEDMF (also takes deep_atmosphere)
- DecayingProfile
- MoistAdiabaticProfileEDMFX

No-arg ICs
- ConstantBuoyancyFrequencyProfile
- IsothermalProfile
- DryDensityCurrentProfile
- RisingThermalBubbleProfile
- PrecipitatingColumn
- RCEMIPIIProfile_295 / _300 / _305 (this can probably be unified further)
- ShipwayHill2012

File-based / field-level overwrite ICs
- MoistFromFile (file path as IC name)
- ReanalysisTimeVarying (InterpolatedColumnProfile)
- WeatherModel
- AMIPFromERA5


src/setups/Setups.jl
```diff
@@ -0,0 +1,220 @@
...
+import ..Microphysics2Moment
+import ..Microphysics2MomentP3
+import ..PrognosticEDMFX
+import ..DiagnosticEDMFX
+import ..EDOnlyEDMFX
+import ..n_mass_flux_subdomains
+import ..PrescribedSST
+import ..SlabOceanSST
+
+abstract type AbstractSetup end
```
haakon-e: It's not clear to me if we need an `abstract type` at all. The fallback functions below could simply read 
```julia
face_initial_condition(setup, local_geometry, params) = (; w = eltype(params)(0))
#...
overwrite_initial_state!(setup, Y, thermo_params) = nothing
#...
function initial_state(setup, params, atmos_model, center_space, face_space)
#...
```
and then, each model setup, e.g. `Bomex.jl`, could be
```julia
struct Bomex{P}
    prognostic_tke::Bool
    profiles::P
end
#...
center_initial_condition(setup::Bomex, local_geometry, params) = ...
# etc.
```
That is, just duck typing, instead of enforcing a type structure. This way, other people can implement their own cases without explicitly subtyping `AbstractSetup`. The contract lies in the setups returning `NamedTuple`s with expected fields anyways, that's not something enforced by being a subtype of `AbstractSetup`.

I think only the `initial_state` function would have to stay in `src`, though `get_setup_type` would need to change somehow.

----
That said, if there are compelling reasons to keep the abstract supertype, I'd suggest updating it's name to something like
```suggestion
abstract type ModelSetup end
```
would be a bit more descriptive? or `InitialConditionSetup`, `CaseSetup`, or just `Setup`?

src/setups/Setups.jl
haakon-e, line 1: From #4234:

> > Phase 1:
> >
> > * What is the purpose of the Setups.jl module. What functionality resides there?
>
> To specify the setup of a simulation--as distinct from the model source code. This includes config, initial condition, boundary conditions. It can also include additional forcing functions (i.e., tendencies of prognostic variables). **They do not belong into the source code of the model itself.**[^1]

[^1]: emphasis mine

My understanding is that the `setups` module and folder should reside outside the `ClimaAtmos.jl/src` directory, e.g. in `ClimaAtmos.jl/setups`

src/setups/Bomex.jl
```diff
@@ -0,0 +1,83 @@
+"""
+    Bomex
+
+The Bomex setup described in [Holland1973](@cite), with a hydrostatically
+balanced pressure profile. Profiles are sourced from AtmosphericProfilesLibrary.
+
+The `profiles` field stores precomputed atmospheric profile functions (computed
+at construction time before broadcasting).
```
haakon-e, line 8: Would be useful if this, and other docstrings, contained an example of how to construct it

src/setups/Bomex.jl
```diff
@@ -0,0 +1,83 @@
...
+    profiles::P
+end
+Bomex(; prognostic_tke, thermo_params) =
+    Bomex(prognostic_tke, bomex_profiles(thermo_params))
+
+"""
+    bomex_profiles(params)
+
+Precompute the atmospheric profiles for the Bomex case. Returns a NamedTuple of
+interpolatable profile functions of height `z`.
```
haakon-e: A suggestion for slightly more comprehensive docs.
```suggestion
    bomex_profiles(thermo_params)

Precompute the atmospheric profiles for the Bomex case, given thermodynamic parameters `thermo_params`. 

Returns a NamedTuple of interpolatable profile functions of height `z`.
```

src/setups/Setups.jl
```diff
@@ -0,0 +1,220 @@
...
+            lg,
+            params,
+            atmos_model,
+        )
+    face_ic =
+        lg -> face_prognostic_variables(
+            face_initial_condition(setup, lg, params),
+            lg,
+            atmos_model,
+        )
```
haakon-e: Perhaps slightly cleaner to just make these proper functions?

```suggestion
    center_ic(lg) = center_prognostic_variables(
        center_initial_condition(setup, lg, params), lg, params, atmos_model,
    )
    face_ic(lg) = face_prognostic_variables(
        face_initial_condition(setup, lg, params), lg, atmos_model,
    )
```

src/setups/common.jl
```diff
@@ -0,0 +1,145 @@
...
+# Perturbation coefficient
+# ============================================================================
+
+"""
+    perturb_coeff(point)
+
+Return a geometry-dependent perturbation coefficient for the given coordinate
+point. Used by setups that apply spatial perturbations to initial profiles.
+"""
+perturb_coeff(p::Geometry.AbstractPoint{FT}) where {FT} = FT(0)
```
haakon-e: It's not clear to me that these functions are sufficiently generic to be broadly useful for many cases. I'd suggest each case defines these locally, or their names are updated to something more reflective of what they do.

src/setups/common.jl
```diff
@@ -0,0 +1,145 @@
...
+perturb_coeff(p::Geometry.LatLongZPoint{FT}) where {FT} = sind(p.long)
+perturb_coeff(p::Geometry.XZPoint{FT}) where {FT} = sin(p.x)
+perturb_coeff(p::Geometry.XYZPoint{FT}) where {FT} = sin(p.x)
+
+# ============================================================================
+# Thermodynamic helpers
```
haakon-e: It's not clear to me if we need these helpers all the time we have an API for this in Thermodynamics.jl

src/setups/common.jl
```diff
@@ -0,0 +1,145 @@
...
+Construct a NamedTuple representing the physical state at a grid point.
+This is the return type of `center_initial_condition` — it describes the
+thermodynamic and kinematic state without any knowledge of the atmos model.
+
+The assembly layer (`prognostic_variables.jl`) converts this into model-specific
+prognostic variables.
+
+## Required arguments
+- `T`: Temperature (K)
+- `p`: Pressure (Pa)
```
haakon-e: pressure is only used to calculate density, so in principle either of them could be provided here.

src/setups/common.jl
haakon-e, line 1: I would find it more helpful if this file had a more descriptive name, depending on what actually ends up being in here.

haakon-e (commented):
Thanks for the PR! I assume it's based on the ideas suggested in #4234, so would be good to link back to it. Left some broad comments for you to consider (and some scattered specific suggestions).

src/setups/Setups.jl
nefrathenrici: I agree, the abstract type can be removed.

src/setups/Setups.jl
nefrathenrici: I agree with the intent to separate out the setups, but I'm not sure how to do it properly. Maybe the modularization this PR provides is sufficient? We can move the setups module outside of `src` but I am unclear how that would change how we interact with it. If the module is still always included in ClimaAtmos, it makes little difference whether it lives in `src` or at the top level. We could try to conditionally load the setups, but I am not sure what purpose that would serve.

haakon-e: > If the module is still always included in ClimaAtmos, it makes little difference whether it lives in `src` or at the top level.

I agree with this. What I was getting at is that `ClimaAtmos` won't need to know about the module at all. Only the `ci_driver.jl` and other model scripts would need to know about them.

src/setups/prognostic_variables.jl
```diff
@@ -0,0 +1,257 @@
...
+# ============================================================================
+
+"""
+    uniform_subdomains(nt::NamedTuple, turbconv_model)
+
+Create `n` identical subdomain copies of `nt`, where `n` is the number of
+mass-flux subdomains in `turbconv_model`.
+"""
+uniform_subdomains(nt, turbconv_model) =
+    ntuple(_ -> nt, Val(n_mass_flux_subdomains(turbconv_model)))
```
dennisYatunin:
```suggestion
    ntuple(Returns(nt), Val(n_mass_flux_subdomains(turbconv_model)))
```
(for slightly lower latency)

src/setups/common/prognostic_variables.jl
```diff
@@ -0,0 +1,257 @@
...
+"""
+    face_prognostic_variables(face_state, local_geometry, atmos_model)
+
+Convert a face-state NamedTuple (from `face_initial_condition`) into the face
+prognostic NamedTuple required by `atmos_model`.
+"""
+function face_prognostic_variables(face_state, local_geometry, atmos_model)
+    FT = typeof(face_state.w)
+    u₃ = C3(Geometry.WVector(face_state.w), local_geometry)
+    w_draft = Geometry.WVector(get(face_state, :w_draft, zero(FT)))
```
dennisYatunin: It would be more consistent with the computations on centers to auto-fill the physical state on faces with `w_draft = 0` before calling this function.

Alternatively, you could implement the center computations using the end-stage zero-padding you have here for `w_draft`. If the GPU compiler fails to inline part of the current implementation for centers, the resulting kernel may pass around zero-padding registers that it never ends up using. Only adding zero-padding at the last step of the computation may avoid this issue. But I'm not sure if that's actually an issue without testing it, and this won't matter much for overall performance.

src/setups/prognostic_variables.jl
```diff
@@ -0,0 +1,257 @@
...
+function turbconv_face_variables(u₃, w_draft, lg, turbconv_model::PrognosticEDMFX)
+    return (; sgsʲs = uniform_subdomains((; u₃ = C3(w_draft, lg)), turbconv_model))
+end
+
+# ============================================================================
+# Surface field
+# ============================================================================
+
+"""
+    atmos_surface_field(surface_space, surface_model)
```
dennisYatunin: It would be more consistent with the center/face computations to define `surface_prognostic_variables` here, and broadcast that over the `local_geometry` field in `initial_state`. There isn't really a meaningful distinction between defining variables in 3D domains vs 2D domains, so they might as well use the same implementation.

You could then also dispatch off the type of `local_geometry.coord` instead of using `if :lat in propertynames(...)` in the method for `SlabOceanSST`, if that would be more succinct.

src/setups/Setups.jl
dennisYatunin: If these setups are only intended for use with `ci_dirver.jl`, then I agree that this module might as well live in the same place as `ci_driver.jl`. But I was under the impression that we wanted to use some of these setups in the unit tests for `ClimaAtmos`, in which case this module should live in either `src` or `test` because it is required to run the unit tests for `src`.

src/solver/model_getters.jl
```diff
@@ -441,7 +448,22 @@ function get_subsidence_model(parsed_args, radiation_mode, FT)
...
+    parsed_args, ::Type{FT}; setup_type = nothing,
+) where {FT}
+    # Setup path: if setup provides large-scale advection, use it
+    if !isnothing(setup_type)
+        data = Setups.large_scale_advection_forcing(setup_type, FT)
+        if !isnothing(data)
+            prof_dqtdt = (thermo_params, ᶜp, t, z) -> data.prof_dqtdt(z)
+            prof_dTdt =
+                (thermo_params, ᶜp, t, z) ->
+                    data.prof_dTdt(TD.exner_given_pressure(thermo_params, ᶜp), z)
```
dennisYatunin:
```suggestion
            prof_dqtdt = (thermo_params, p, t, z) -> data.prof_dqtdt(z)
            prof_dTdt =
                (thermo_params, p, t, z) ->
                    data.prof_dTdt(TD.exner_given_pressure(thermo_params, p), z)
```
`ᶜp` is for the pressure field, `p` is for a single value of pressure

src/solver/model_getters.jl
```diff
@@ -420,7 +420,14 @@ function get_call_cloud_diagnostics_per_stage(parsed_args)
     end
 end
 
-function get_subsidence_model(parsed_args, radiation_mode, FT)
+function get_subsidence_model(parsed_args, radiation_mode, FT; setup_type = nothing)
+    # Setup path: if setup provides subsidence, use it
+    if !isnothing(setup_type)
+        prof = Setups.subsidence_forcing(setup_type, FT)
+        !isnothing(prof) && return Subsidence(prof)
```
dennisYatunin:
```suggestion
        subsidence = Setups.subsidence_forcing(setup_type, FT)
        !isnothing(subsidence) && return subsidence
```
It would be a bit simpler to return a `Subsidence` object from the setup, so that no further manipulation is necessary. This also applies to `LargeScaleAdvection` and `SCMCoriolis` below.

src/solver/type_getters.jl
```diff
@@ -492,7 +529,45 @@ function get_steady_state_velocity(params, Y, topo, initial_condition, mesh_warp
     return (; ᶜu, ᶠu)
 end
 
-function get_surface_setup(parsed_args)
+"""
+    _assemble_surface_state(sfc_data::NamedTuple, FT)
```
dennisYatunin: This functionality also feels like it belongs in the `Setups` module, with similar auto-filling and conditional variable inclusion. The `SurfaceState` could be returned directly from that module, like with `subsidence_forcing` and similar submodels.

src/setups/GCMDriven.jl
```diff
@@ -0,0 +1,89 @@
...
+    z_gcm, vars, T_sfc = NC.NCDataset(external_forcing_file) do ds
+        site = ds.group[cfsite_number]
+        z = vec(gcm_height(site))
+        profiles = (
+            gcm_driven_profile_tmean(site, "ta"),
+            gcm_driven_profile_tmean(site, "ua"),
+            gcm_driven_profile_tmean(site, "va"),
+            gcm_driven_profile_tmean(site, "hus"),
+            # Convert specific volume (alpha) to density: ρ = 1/α
+            vec(mean(1 ./ site["alpha"][:, :], dims = 2)),
```
dennisYatunin:
```suggestion
            vec(mean(inv, site["alpha"][:, :], dims = 2)),
```
This intermediate array does not need to be allocated.

src/setups/GCMDriven.jl
```diff
@@ -0,0 +1,89 @@
...
+
+    T_prof, u_prof, v_prof, q_tot_prof, ρ_prof = map(vars) do value
+        Intp.extrapolate(
+            Intp.interpolate((z_gcm,), value, Intp.Gridded(Intp.Linear())),
+            Intp.Flat(),
+        )
+    end
+
+    profiles = (; T = T_prof, u = u_prof, v = v_prof, q_tot = q_tot_prof, ρ = ρ_prof)
+    return GCMDriven(external_forcing_file, cfsite_number, profiles, T_sfc)
```
dennisYatunin:
```suggestion
    T, u, v, q_tot, ρ = map(vars) do value
        Intp.extrapolate(
            Intp.interpolate((z_gcm,), value, Intp.Gridded(Intp.Linear())),
            Intp.Flat(),
        )
    end
    return GCMDriven(external_forcing_file, cfsite_number, (; T, u, v, q_tot, ρ), T_sfc)
```
I'm pretty sure `profiles` is being boxed in the current implementation, since the same variable name also appears inside the closure above. If that's the case, you could improve latency by avoiding that variable name here.

src/setups/InterpolatedColumnProfile.jl
```diff
@@ -0,0 +1,87 @@
...
+- `external_forcing_file`: Path to the NetCDF file containing external forcing data
+- `start_date`: Date string in format "yyyymmdd" specifying which time to use
+
+## Returns
+An `InterpolatedColumnProfile` instance with interpolators for T, u, v, q_tot, and ρ.
+
+The function reads the file, finds the time index closest to `start_date`, extracts
+vertical profiles at that time, and builds 1D interpolators for each variable.
+"""
+function interpolated_column_profile(external_forcing_file::String, start_date::String)
```
dennisYatunin: Most of this looks like a duplicate of the `GCMDriven` function. It would be good to use the same implementation in both cases.

src/setups/overwrite_from_file.jl
```diff
@@ -0,0 +1,278 @@
...
+                file_path, "v", center_space; svi_kwargs...,
+            ),
+            SpaceVaryingInputs.SpaceVaryingInput(
+                file_path, "w", center_space; svi_kwargs...,
+            ),
+        )
+    Y.c.uₕ .= C12.(Geometry.UVVector.(vel))
+    Y.f.u₃ .= ᶠinterp.(C3.(Geometry.WVector.(vel)))
+    e_kin = similar(ᶜT)
+    e_kin .= compute_kinetic(Y.c.uₕ, Y.f.u₃)
```
dennisYatunin:
```suggestion
    e_kin = compute_kinetic(Y.c.uₕ, Y.f.u₃)
```
This intermediate field does not need to be allocated.

src/setups/overwrite_from_file.jl
```diff
@@ -0,0 +1,278 @@
...
+# Internal helpers (shared by overwrite_from_file! and WeatherModel)
+# ============================================================================
+
+"""
+    _hydrostatic_pressure(p_sfc, ᶜT, ᶜq_tot, face_space, thermo_params)
+
+Compute face pressure by hydrostatic integration from surface pressure.
+Solves ∂(ln p)/∂z = -g/(Rₘ(q)T) using `column_integral_indefinite!`.
+"""
+function _hydrostatic_pressure(p_sfc, ᶜT, ᶜq_tot, face_space, thermo_params)
```
dennisYatunin: Unless this is to avoid name conflicts with some other `hydrostatic_pressure` function, prefixing with underscores isn't particularly helpful, especially if the function is exposed enough to have its own docstring. Same goes for the other functions below.

src/setups/overwrite_from_file.jl
```diff
@@ -0,0 +1,278 @@
...
+    regridder_kwargs =
+        if isnothing(extrapolation_bc) && isnothing(interpolation_method)
+            ()
+        elseif isnothing(interpolation_method)
+            (; extrapolation_bc)
+        elseif isnothing(extrapolation_bc)
+            (; interpolation_method)
+        else
+            (; extrapolation_bc, interpolation_method)
+        end
```
dennisYatunin:
```suggestion
    regridder_kwargs = filter(!isnothing, (; extrapolation_bc, interpolation_method))
```

src/setups/Rico.jl
```diff
@@ -0,0 +1,100 @@
...
+    p_val = FT(profiles.p(z))
+    T = FT(TD.air_temperature(thermo_params, TD.pθ_li(), p_val, θ_val, q_tot_val))
+
+    tke_val = prognostic_tke ? FT(0) : FT(profiles.tke(z))
+
+    return physical_state(;
+        T, p = p_val, q_tot = q_tot_val,
+        u = FT(profiles.u(z)), v = FT(profiles.v(z)),
+        tke = tke_val,
+    )
```
dennisYatunin:
```suggestion
    (; θ, q_tot, p, u, v, tke) = map(profile -> profile(z), profiles)
    T = TD.air_temperature(thermo_params, TD.pθ_li(), p, θ, q_tot)
    return physical_state(; T, p, q_tot, u, v, tke = prognostic_tke ? FT(0) : tke)
```
None of the `FT` conversions here are needed, since the `profiles` are generated using the same `FT`.

src/setups/Bomex.jl
```diff
@@ -0,0 +1,94 @@
...
+
+    # Evaluate profiles at z
+    θ = FT(profiles.θ(z))
+    q_tot = FT(profiles.q_tot(z))
+    p = FT(profiles.p(z))
+    T = FT(TD.air_temperature(thermo_params, TD.pθ_li(), p, θ, q_tot))
+
+    tke = prognostic_tke ? FT(0) : FT(profiles.tke(z))
+
+    return physical_state(; T, p, q_tot, u = FT(profiles.u(z)), tke)
```
dennisYatunin:
```suggestion
    (; θ, q_tot, p, u, tke) = map(profile -> profile(z), profiles)
    T = TD.air_temperature(thermo_params, TD.pθ_li(), p, θ, q_tot)
    return physical_state(; T, p, q_tot, u, tke = prognostic_tke ? FT(0) : tke)
```
None of these `FT` conversions are needed either.

dennisYatunin (approved):

src/solver/model_getters.jl
nefrathenrici: I prefer the current pattern as it prevents all Setups using subsidence from needing the additional boilerplate of calling `Subsidence(profile)`.

src/setups/GCMDriven.jl
nefrathenrici: I ended up putting this function in a helper since it's needed for other setups

src/setups/common.jl
```diff
@@ -0,0 +1,182 @@
...
+        tke = FT(0),
+    )
+end
+
+# ============================================================================
+# Thermodynamic helpers
+# ============================================================================
+
+"""
+    total_specific_energy(params, T, local_geometry; u=0, v=0, q_tot=0, q_liq=0, q_ice=0)
```
tapios:
```suggestion
    total_energy(params, T, local_geometry; u=0, v=0, q_tot=0, q_liq=0, q_ice=0)
```
We have moved away from using `specific` for the energy functions in TD. Would be good to do the same here. (All energies we compute are specific, so it can be omitted.)

src/setups/common.jl
```diff
@@ -0,0 +1,182 @@
...
+
+Compute the specific total energy `e_tot = e_int + e_kin + e_pot` from
+temperature, geometry, and optional velocity/moisture fields.
+
+This is the per-unit-mass total energy (multiply by `ρ` to get `ρe_tot`).
+"""
+function total_specific_energy(
```
tapios:
```suggestion
function total_energy(
```

src/setups/common.jl
```diff
@@ -0,0 +1,182 @@
...
+end
+
+"""
+    moist_static_energy(params, T, local_geometry; q_tot=0, q_liq=0, q_ice=0)
+
+Compute the moist static energy `h = c_p T + g z + L_v q_tot` (enthalpy + geopotential).
+
+Used for initializing EDMFX subdomain thermodynamic variables.
+"""
+function moist_static_energy(
```
tapios: Use `moist_static_energy` from TD to ensure consistency.

src/setups/MoistFromFile.jl
```diff
@@ -0,0 +1,33 @@
+"""
```
tapios, line 1: I think it would be good to move files such as this, which contain generic helpers, to a Common subdirectory (or something similar), so we keep helper infrastructure together and distinct from specific case setups.

tapios (approved):
Thanks, @nefrathenrici. This is a large step in a good direction.

docs/src/api.md
```diff
@@ -78,7 +32,40 @@ ClimaAtmos.SLEVEWarp
...
+ClimaAtmos.Setups.DryDensityCurrentProfile
+ClimaAtmos.Setups.RisingThermalBubbleProfile
+ClimaAtmos.Setups.MoistAdiabaticProfileEDMFX
+ClimaAtmos.Setups.GCMDriven
+ClimaAtmos.Setups.InterpolatedColumnProfile
+ClimaAtmos.Setups.MoistFromFile
+ClimaAtmos.Setups.WeatherModel
+ClimaAtmos.Setups.AMIPFromERA5
+ClimaAtmos.Setups.ColumnInterpolatableField
+```
```
haakon-e: I think this list of cases (i.e. structs) can be automatically generated by:

~~~julia
```@autodocs
Modules = [Setups]
Order   = [:type]
```
~~~

see [`@autodocs-block`](https://documenter.juliadocs.org/stable/man/syntax/#@autodocs-block) for details. With this syntax, the `API.md` page does not need to be updated whenever we add new simulation cases.

src/prognostic_equations/forcing/subsidence.jl
```diff
@@ -65,31 +65,9 @@ If `subsidence_model` is `Nothing`, no subsidence tendency is applied.
 """
 subsidence_tendency!(Yₜ, Y, p, t, ::Nothing) = nothing    # No subsidence
 
-
-"""
```
haakon-e, line 69: Why remove the docstring?

src/prognostic_equations/scm_coriolis.jl
```diff
@@ -19,17 +19,11 @@ large-scale forcing.
 
 Arguments:
 - `ᶜuₕ`: The cell-centered horizontal velocity field [m/s].
-- `scm_coriolis`: An object containing Coriolis parameters and geostrophic wind
-                   profiles. If not of type `SCMCoriolis`, a `NullBroadcasted`
-                   is returned, indicating no Coriolis forcing.
-
-Returns:
```
haakon-e, line 26: Keep docstring about the returned object?

src/prognostic_equations/forcing/subsidence.jl
nefrathenrici: The same docstring is provided above the method `subsidence_tendency!(Yₜ, Y, p, t, ::Nothing) = nothing `

docs/src/api.md
nefrathenrici: Oh that is good to know! I decided to just group them manually in the `setups.md` file

PR #4305, opened 2026-02-17
Update updraft buoyancy and edmf filters

sajjadazimi:
## Purpose 
This PR updates buoyancy equations and edmf filters to make prognostic edmf more stable:
- precompute updraft buyancy; 
- apply edmf filter as a callback; 
- mix updraft with grid-mean when it is unphysical (area fraction or velocity is negative or negligible or when buoyancy is negative)

----

### DIFF
--- a/src/cache/prognostic_edmf_precomputed_quantities.jl
+++ b/src/cache/prognostic_edmf_precomputed_quantities.jl
@@ -251,6 +251,7 @@ NVTX.@annotate function set_prognostic_edmf_precomputed_quantities_explicit_clos
         ᶜentrʲs,
         ᶜdetrʲs,
         ᶜturb_entrʲs,
+        ᶠρ_diffʲs,
         ᶠnh_pressure₃_buoyʲs,
     ) = p.precomputed
     (; ustar, obukhov_length) = p.precomputed.sfc_conditions
@@ -408,6 +409,8 @@ NVTX.@annotate function set_prognostic_edmf_precomputed_quantities_explicit_clos
         else
             @. detr_int_val = limit_detrainment(detr_int_val, entr_int_val, ᶜaʲ_int_val, dt)
         end
+
+        @. ᶠρ_diffʲs.:($$j) = min(0, ᶠinterp(ᶜρʲs.:($$j) - Y.c.ρ)) / ᶠinterp(ᶜρʲs.:($$j))
 
         # The buoyancy term in the nonhydrostatic pressure closure is always applied
         # for prognostic edmf. The tendency is combined with the buoyancy term in the
--- a/src/prognostic_equations/advection.jl
+++ b/src/prognostic_equations/advection.jl
@@ -358,7 +358,7 @@ function edmfx_sgs_vertical_advection_tendency!(
     n = n_prognostic_mass_flux_subdomains(turbconv_model)
     (; dt) = p
     (; edmfx_mse_q_tot_upwinding, edmfx_tracer_upwinding) = p.atmos.numerics
-    (; ᶠu³ʲs, ᶠKᵥʲs, ᶜρʲs) = p.precomputed
+    (; ᶠu³ʲs, ᶠKᵥʲs, ᶜρʲs, ᶠρ_diffʲs) = p.precomputed
     (; ᶠgradᵥ_ᶜΦ) = p.core
 
     FT = eltype(p.params)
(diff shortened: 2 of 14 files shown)


src/cache/prognostic_edmf_precomputed_quantities.jl
```diff
@@ -402,6 +403,8 @@ NVTX.@annotate function set_prognostic_edmf_precomputed_quantities_explicit_clos
             @. detr_int_val = limit_detrainment(detr_int_val, entr_int_val, ᶜaʲ_int_val, dt)
         end
 
+        @. ᶠρ_diffʲs.:($$j) = min(0, ᶠinterp(ᶜρʲs.:($$j) - Y.c.ρ)) / ᶠinterp(ᶜρʲs.:($$j))
```
tapios, line 413: Is it necessary to limit updraft buoyancy to be positive? At the updraft top, we usually get convective overshoots, with negative buoyancy.

src/prognostic_equations/mass_flux_closures.jl
```diff
@@ -204,33 +204,72 @@ function edmfx_vertical_diffusion_tendency!(
...
- - Relax ρa to ρ when a is larger than one
-
-This function modifies the tendency `Yₜ` in place based on the current state `Y`,
-parameters `p`, time `t`, and the turbulence convection model `turbconv_model`.
-It specifically targets the vertical velocity (`u₃`) and the product of density and area fraction (`ρa`)
-for each sub-domain in the EDMFX model.
+Apply EDMF physical constraints: immediately mix the updraft with the environment if
+  - area fraction is negative or negligible (smaller than eps)
+  - updraft velocity is negative or negligible
+  - updraft air is heavier than the grid mean (negative buoyancy)
```
tapios, line 214: Again, is the buoyancy filtering necessary? Would it not suffice to have strong mixing with the environment for negative buoyancy, without a hard clipping?

tapios (approved):
LGTM

src/cache/prognostic_edmf_precomputed_quantities.jl
szy21: If we need it, maybe we can change the threhold from 0 to min(0,  -wʲ / dt), if that's slightly more physical?

src/prognostic_equations/advection.jl
```diff
@@ -382,8 +382,7 @@ function edmfx_sgs_vertical_advection_tendency!(
         # and calcuate the buoyancy term relative to the grid-mean density.
         # We also include the buoyancy term in the nonhydrostatic pressure closure here.
         @. Yₜ.f.sgsʲs.:($$j).u₃ -=
-            (1 - α_b) * (ᶠinterp(ᶜρʲs.:($$j) - Y.c.ρ) * ᶠgradᵥ_ᶜΦ) /
-            ᶠinterp(ᶜρʲs.:($$j)) + ᶠgradᵥ(ᶜKᵥʲ)
+            (1 - α_b) * ᶠρ_diffʲs.:($$j) * ᶠgradᵥ_ᶜΦ + ᶠgradᵥ(ᶜKᵥʲ)
```
szy21, line 385: Do we still want the buoyancy term to be implicit if we remove the Jacobian?

src/prognostic_equations/mass_flux_closures.jl
```diff
@@ -204,33 +204,72 @@ function edmfx_vertical_diffusion_tendency!(
...
+            @. Y.f.sgsʲs.:($$j).u₃ =
+                C3(max(Y.f.sgsʲs.:($$j).u₃.components.data.:1, 0))
+
+            # clip updraft velocity to zero if the updraft air is heavier than the grid-mean
+            @. Y.f.sgsʲs.:($$j).u₃ =
+                ifelse(ᶠinterp(ᶜρʲs.:($$j) - Y.c.ρ) > 0, C3(0), Y.f.sgsʲs.:($$j).u₃)
+
+            # clip updraft area fraction to zero if the cell-averaged velocity is negligible.
+            @. Y.c.sgsʲs.:($$j).ρa = ifelse(
+                ᶜinterp(Y.f.sgsʲs.:($$j).u₃.components.data.:1) <= eps(FT),
```
szy21, line 246: any reason why it is <= eps(FT) here but < eps(FT) for area fraction below?

PR #4311, opened 2026-02-18
Semi-implicit limiting of sedimentation sinks

tapios:
## Purpose 

Backward Euler limiter for mass correction terms in updrafts. 

## Content

----

### DIFF
--- a/src/parameterized_tendencies/microphysics/tendency_limiters.jl
+++ b/src/parameterized_tendencies/microphysics/tendency_limiters.jl
@@ -32,31 +32,6 @@ max_rate = limit(q_rai, dt, 3)
 end
 
 """
-    implicit_sink_rate(rate, dt)
-
-Semi-implicit (backward Euler) limiter for a linear decay tendency.
-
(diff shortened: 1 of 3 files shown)


src/parameterized_tendencies/microphysics/tendency_limiters.jl
```diff
@@ -212,7 +210,7 @@ NamedTuple with limited tendencies: `(dq_lcl_dt, dq_icl_dt, dq_rai_dt, dq_sno_dt
     )
     dq_sno_dt = tendency_limiter(
         mp_tendency.dq_sno_dt,
-        limit(q_ice + q_rai, dt, n_source),
+        limit(q_ice, dt, n_source),
```
szy21, line 213: why do we remove q_rai here?

szy21 (approved):

src/parameterized_tendencies/microphysics/tendency_limiters.jl
tapios: Because we rarely convert rain to snow. (It's not an important change though.)

Issue #4285, opened 2026-02-08
Remove call_cloud_diagnostics_per_stage option (treat as always true)


tapios:

trontrytel:
Would it be ok to close this issue? I think we have done the core of the work. And we will continue working on 2MP3 microphysics and simplifying the interface? cc @tapios

tapios:
Fine to close.

Issue #4007, opened 2025-09-09
Missing terms for 2M microphysics


szy21:
In the process of stabilizing/improving 1M microphysics, we added something to 1M tracers that should be applied to 2M tracers (number concentration) as well. These include:

- Rayleigh sponge of subgrid scale tracers
- Implicit updraft advection and sedimentation
- Implicit entrainment

cc @sajjadazimi @trontrytel. I will try to remember to add things for 2M microphysics if they are easy to add in future PRs, and at some point, I may go through the code and add everything. But if you have some time, feel free to take it on.

PR #4320, opened 2026-02-20
Wip on moisture_model and microphysics_model refactor

trontrytel:
This PR addresses the first issue from list of https://github.com/CliMA/ClimaAtmos.jl/issues/4285

- Combines the `moisture_model` with `microphysics_model` (keeping the `microphysics_model` name). The supported options are: `DryModel`, `EquilibriumMicrophysics0M`, `NonEquilibriumMicrophysics1M`, `NonEquilibriumMicrophysics2M` (to be deleted when P3 is fully functional), `NonEquilibriumMicrophysics2MP3`.

- Separates `SGSQuadrature` from `microphysics_model`. About 50 places require only microphysics information, while only about 10 require quadrature information. This pattern lead to cumbersome unpacking the type to get to base type.


src/prognostic_equations/hyperdiffusion.jl
```diff
@@ -223,26 +219,22 @@ function dss_hyperdiffusion_tendency_pairs(p)
...
-        Union{Microphysics2Moment, QuadratureMicrophysics{Microphysics2Moment}} ?
+        p.atmos.microphysics_model isa NonEquilibriumMicrophysics2M ?
         (
             p.hyperdiff.ᶜ∇²q_liqʲs => buffer.ᶜ∇²q_liqʲs,
             p.hyperdiff.ᶜ∇²q_iceʲs => buffer.ᶜ∇²q_iceʲs,
             p.hyperdiff.ᶜ∇²q_raiʲs => buffer.ᶜ∇²q_raiʲs,
             p.hyperdiff.ᶜ∇²q_snoʲs => buffer.ᶜ∇²q_snoʲs,
-            p.hyperdiff.ᶜ∇²q_liqʲs => buffer.ᶜ∇²q_liqʲs,
-            p.hyperdiff.ᶜ∇²q_raiʲs => buffer.ᶜ∇²q_raiʲs,
+            p.hyperdiff.ᶜ∇²n_liqʲs => buffer.ᶜ∇²n_liqʲs,
```
trontrytel, line 236: TODO - check this

src/solver/model_getters.jl
```diff
@@ -1,18 +1,33 @@
...
 
+function get_sgs_quadrature(parsed_args, params = nothing)
+    use_sgs_quadrature = get(parsed_args, "use_sgs_quadrature", false)
+    use_sgs_quadrature || return nothing
+    FT = parsed_args["FLOAT_TYPE"] == "Float64" ? Float64 : Float32
+    distribution = get_sgs_distribution(parsed_args)
+    quadrature_order = get(parsed_args, "quadrature_order", 2)
+    T_min = isnothing(params) ? FT(150) : FT(CAP.T_min_sgs(params))
+    q_max = isnothing(params) ? FT(0.1) : FT(CAP.q_max_sgs(params))
+    return SGSQuadrature(FT; quadrature_order, distribution, T_min, q_max)
```
trontrytel, line 28: TODO - check where SGSQuadrature is defined

src/solver/type_getters.jl
```diff
@@ -136,13 +116,13 @@ function get_atmos(config::AtmosConfig, params)
     end
 
     atmos = AtmosModel(;
-        # AtmosWater - Moisture, Precipitation & Clouds
-        moisture_model,
+        # AtmosWater - Microphysics & Clouds
```
trontrytel: Keep the comment

src/solver/types.jl
```diff
@@ -895,14 +854,14 @@ end
...
     AtmosWater
 
-Groups moisture-related models and types.
+Groups moisture and microphysics-related models and types.
 """
-@kwdef struct AtmosWater{MM, PM, CM, MTTS, TNM}
-    moisture_model::MM = DryModel()
-    microphysics_model::PM = NoPrecipitation()
+@kwdef struct AtmosWater{PM, CM, MTTS, TNM, SQ}
+    microphysics_model::PM = DryModel()
```
trontrytel: ::MM

src/initial_conditions/atmos_state.jl
```diff
@@ -172,11 +163,9 @@ end
 function turbconv_center_variables(
     ls,
     turbconv_model::PrognosticEDMFX,
-    moisture_model::NonEquilMoistModel,
-    microphysics_model::Union{Microphysics1Moment, Microphysics2Moment},
+    microphysics_model::Union{NonEquilibriumMicrophysics1M, NonEquilibriumMicrophysics2M},
     gs_vars,
 )
-    # TODO - Instead of dispatching, should we unify this with the above function?
```
trontrytel, line 179: Keep the comment

src/diagnostics/default_diagnostics.jl
```diff
@@ -277,65 +272,73 @@ function default_diagnostics(
...
-    return [average_func(precip_diagnostics...; output_writer, start_date, t_start)...]
+    return [
+        _moist_default_diagnostics(duration, start_date, t_start; output_writer)...,
+        average_func(precip_diagnostics...; output_writer, start_date, t_start)...,
+    ]
 end
 
 function default_diagnostics(
-    ::Union{Microphysics2Moment, QuadratureMicrophysics{Microphysics2Moment}},
+    ::NonEquilibriumMicrophysics2MP3,
```
nefrathenrici, line 315: This methods does not add precip diagnostics ["husra", "hussn", "cdnc", "ncra"]. 
Is there a reason `NonEquilibriumMicrophysics2MP3` and `NonEquilibriumMicrophysics2M` have two separate methods? Should they have the same default diagnostics?

src/prognostic_equations/hyperdiffusion.jl
```diff
@@ -47,18 +47,14 @@ function hyperdiffusion_cache(
...
         (;
             ᶜ∇²q_liqʲs = similar(Y.c, NTuple{n, FT}),
             ᶜ∇²q_iceʲs = similar(Y.c, NTuple{n, FT}),
             ᶜ∇²q_raiʲs = similar(Y.c, NTuple{n, FT}),
             ᶜ∇²q_snoʲs = similar(Y.c, NTuple{n, FT}),
         ) :
-        moisture_model isa NonEquilMoistModel &&
-        microphysics_model isa
-        Union{Microphysics2Moment, QuadratureMicrophysics{Microphysics2Moment}} ?
+        microphysics_model isa NonEquilibriumMicrophysics2M ?
```
nefrathenrici, line 57: Should this check for `NonEquilibriumMicrophysics2MP3` as well?

src/surface_conditions/surface_conditions.jl
```diff
@@ -340,7 +340,7 @@ function tensor_from_components(f₁₃, f₂₃, L, n₃ = surface_normal(L))
 end
 
 """
-    surface_conditions_type(moisture_model, FT)
+    surface_conditions_type(microphysics_model, FT)
```
nefrathenrici:
```suggestion
    surface_conditions_type(atmos_model, FT)
```

src/prognostic_equations/remaining_tendency.jl
```diff
@@ -170,9 +169,7 @@ NVTX.@annotate function additional_tendency!(Yₜ, Y, p, t)
             )
             @. Yₜ.c.sgsʲs.:($$j).q_tot += rst_sgs_q_tot
         end
-        if moisture_model isa NonEquilMoistModel &&
-           microphysics_model isa
-           Union{Microphysics1Moment, QuadratureMicrophysics{Microphysics1Moment}}
+        if microphysics_model isa NonEquilibriumMicrophysics1M
```
nefrathenrici, line 172: Is a case for Microphysics2M needed here?

src/cache/diagnostic_edmf_precomputed_quantities.jl
```diff
@@ -1498,10 +1447,11 @@ NVTX.@annotate function set_diagnostic_edmf_precomputed_quantities_env_precipita
 
     # Get T-based variances (from cache)
     (; ᶜT′T′, ᶜq′q′) = p.precomputed
+    sgs_quad = something(p.atmos.sgs_quadrature, GridMeanSGS())
```
tapios, line 1450: We should make sure 0M always uses quadrature; as it is, it doesn't make sense without it (gives zero condensate and precip almost everywhere, unless grid box saturates). So, no `GridMeanSGS` default.

src/cache/precipitation_precomputed_quantities.jl
```diff
@@ -646,24 +629,38 @@ properties. When running with edmf model this means summing the microphysics
...
-    @. ᶜmp_tendency = BMT.bulk_microphysics_tendencies(
-        BMT.Microphysics0Moment(),
-        cm_params,
-        thermo_params,
-        ᶜT,
-        ᶜq_liq_rai,
-        ᶜq_ice_sno,
-    )
+    sgs_quad = p.atmos.sgs_quadrature
+    if !isnothing(sgs_quad)
```
tapios, line 641: Isn't this handled better by dispatch over the compile time parameter? In any case, 0M without quadrature shouldn't be an option.

src/diagnostics/default_diagnostics.jl
trontrytel: Thanks for looking here. For some reason the diagnostics module felt more convoluted than other places in the code. 

The goal is to have only one option `2MP3`. But right now the `P3` part is still a work in progress, whereas the `2M` part is more or less done. Thats why there isn't much default output for `P3`. Hopefully soon we can combine them.

src/prognostic_equations/hyperdiffusion.jl
trontrytel: Lets add it when we actually make it work in Atmos

src/surface_conditions/surface_conditions.jl
trontrytel: Thank you!

src/cache/diagnostic_edmf_precomputed_quantities.jl
szy21: what if we run LES?

src/prognostic_equations/remaining_tendency.jl
trontrytel: Yes. I think its on the TODO list here: https://github.com/CliMA/ClimaAtmos.jl/issues/4007 

But I wanted to keep this already long PR simple, and fix the 2M missing issues once we are closer to actually using it. Or when I have the time (whichever comes sooner)

src/cache/diagnostic_edmf_precomputed_quantities.jl
trontrytel: I think the minimal change is to add an error if someone wants to run 0M without quadratures. 

```
 if microphysics_model isa EquilibriumMicrophysics0M && isnothing(sgs_quadrature)
       error("EquilibriumMicrophysics0M requires use_sgs_quadrature: true. " *
             "GridMeanSGS fallback is not supported for 0-moment microphysics.")
   end
```

The alternative is to change the `get_sgs_quadrature` function to always return the non-grid_mean option if running with 0M. But I think this is a worse solution, because it would be potentially silently overwriting the config file choice without warning the user.

src/cache/precipitation_precomputed_quantities.jl
trontrytel: I think it does get handled during compile time, because `p.atmos.sgs_quadrature` is known at compile time (?). cc @nefrathenrici

PR #4322, opened 2026-02-21
pedmf area bc update

sajjadazimi:
## Purpose 
Include sgs mass-outflow estimate at the lower boundary to offset advective area loss and allow relaxation toward prescribed surface_area.

## Content

----


src/cache/prognostic_edmf_precomputed_quantities.jl
```diff
@@ -351,14 +351,14 @@ NVTX.@annotate function set_prognostic_edmf_precomputed_quantities_explicit_clos
         entr_int_val = Fields.field_values(Fields.level(ᶜentrʲs.:($j), 1))
         detr_int_val = Fields.field_values(Fields.level(ᶜdetrʲs.:($j), 1))
         @. entr_int_val = ifelse(
-            buoyancy_flux_val < 0 || ᶜaʲ_int_val <= 0 ||
+            buoyancy_flux_val < 0 ||
```
szy21, line 354: why don't we need ᶜaʲ_int_val <= 0 anymore?

sajjadazimi: Because a bit earlier we clip area fraction to eps if buoyancy flux is non-negative (line 346). So this condition here (a <= 0) is redundant.

PR #4328, opened 2026-02-25
Clean up microphysics cache

trontrytel:
This PR addresses the next issue from the list https://github.com/CliMA/ClimaAtmos.jl/issues/4285
- renames `precipitation_precomputed_quantities.jl` to `microphysics_cache.jl` 
- moves the `set_diagnostic_edmf_precomputed_quantities_env_precipitation!` and `set_prognostic_edmf_precomputed_quantities_precipitation!` to `microphysics_cache.jl`
- moves updraft tracers sedimentation velocity computations from `set_microphysics_tendency_cache!` to `set_precipitation_velocities!` along with the rest of precipitation velocities computations for microphysics. 

`set_precipitation_velocities` function is not merged with `set_microphysics_tendency_cache`, because we can re-use the same velocity computation logic for no-EDMF and diagnostic-EDMF configurations. 

The velocity computations in `set_precipitation_velocities` are repetetive, but difficult to turn into a more concise loop because of the small differences between microphysics species. I'll think on how to refactor that part.


src/cache/microphysics_cache.jl
```diff
@@ -893,8 +1197,157 @@ function set_microphysics_tendency_cache!(
...
+        ᶜSnₗᵐʲs,
+        ᶜSnᵣᵐʲs,
+        ᶜρʲs,
+        ᶜTʲs,
+        ᶜuʲs,
+    ) = p.precomputed
+    (; ᶜSqₗᵐ⁰, ᶜSqᵢᵐ⁰, ᶜSqᵣᵐ⁰, ᶜSqₛᵐ⁰, ᶜSnₗᵐ⁰, ᶜSnᵣᵐ⁰, ᶜu⁰, ᶜmp_tendency) =
+        p.precomputed
+    (; ᶜT⁰, ᶜp, ᶜq_tot_safe⁰, ᶜq_liq_rai⁰, ᶜq_ice_sno⁰) = p.precomputed
+    (; ᶜwₗʲs, ᶜwᵢʲs, ᶜwᵣʲs, ᶜwₛʲs, ᶜwₙₗʲs, ᶜwₙᵣʲs, ᶜuʲs) =
```
nefrathenrici:
```suggestion
    (; ᶜwₗʲs, ᶜwᵢʲs, ᶜwᵣʲs, ᶜwₛʲs, ᶜwₙₗʲs, ᶜwₙᵣʲs) =
```

ᶜuʲs is extracted from p.precomputed twice here, one can be removed

src/cache/microphysics_cache.jl
```diff
@@ -666,34 +781,57 @@ function set_microphysics_tendency_cache!(Y, p, ::EquilibriumMicrophysics0M, _)
...
+        cm0,
+        thp,
+        Y.c.ρ,
+        ᶜT,
+        ᶜq_tot_safe,
+        ᶜT′T′,
+        ᶜq′q′,
+        correlation_Tq(p.params),
+    )
+    ᶜq_tot = @. lazy(specific(Y.c.ρq_tot, Y.c.ρ)) # TODO why not q_tot_safe?
```
nefrathenrici, line 819: Is there a reason not to use q_tot_safe? I assume it will be behavior-changing, so this can be done in a future PR.

src/cache/microphysics_cache.jl
```diff
@@ -623,39 +721,56 @@ end
...
+
+**Updrafts** use direct BMT evaluation (no SGS quadrature) because:
+1. Updrafts are coherent turbulent structures with more homogeneous thermodynamic properties
+2. Updraft area fraction is usually small (~1-10%), so SGS variance within updrafts has limited
+impact on the grid-mean tendency
+
+**Environment** uses SGS quadrature integration (when `sgs_quadrature` is configured)
+because the environment dominates the grid-mean variance. The quadrature captures subgrid-scale
+fluctuations in temperature and moisture, which is important for threshold processes like
+condensation/evaporation at cloud edges.
```
nefrathenrici, line 737: Thank you for the clear docstring!

src/cache/microphysics_cache.jl
```diff
@@ -623,39 +721,56 @@ end
 """
     set_microphysics_tendency_cache!(Y, p, microphysics_model, turbconv_model)
 
-Computes the cache needed for microphysics tendencies. When run without edmf
-model this involves computing microphysics sources based on the grid mean
-properties. When running with edmf model this means summing the microphysics
-sources from the sub-domains.
+When run without EDMF, this involves computing microphysics sources based on the grid mean
+properties. The model supports summing over SGS quardature points to sample the variability.
```
nefrathenrici:
```suggestion
properties. The model supports summing over SGS quadrature points to sample the variability.
```

src/cache/microphysics_cache.jl
```diff
@@ -42,8 +42,11 @@ Uses cbrt(floatmin(FT)) to detect effectively-zero specific humidities.
 """
     set_precipitation_velocities!(Y, p, microphysics_model, turbconv_model)
 
-Updates the precipitation terminal velocity, cloud sedimentation velocity,
+Updates the grid mean precipitation terminal velocity, cloud sedimentation velocity,
 and their contribution to total water and energy advection.
+
+For prognostic EDMF it also computes the sedimentation velocities in sub-domains
+and ensures that the grid-scale flux is equal to thesum of sub-grid-scale fluxes.
```
nefrathenrici:
```suggestion
and ensures that the grid-scale flux is equal to the sum of sub-grid-scale fluxes.
```

src/cache/microphysics_cache.jl
```diff
@@ -129,23 +134,24 @@ function set_precipitation_velocities!(
...
     # that the value remains between sgs terminal velocity values (important for stability).
     ᶜρχ = p.scratch.ᶜtemp_scalar_2
     # scratch for adding energy fluxes over subdomains
     ᶜρwₕhₜ = p.scratch.ᶜtemp_scalar_3
 
-    # Compute gs sedimentation velocity based on subdomain velocities (assuming gs flux 
-    # equals sum of sgs fluxes)
+    # Compute gs sedimentation velocity based on subdomain velocities
+    # assuming grid-scale flux equals to the sum of sub-grid-scale fluxes.
+    # TODO - below code is very repetetive between liquid, ice, rain and snow,
```
nefrathenrici, line 147: I think we can continue to clean this up in a future PR.

src/cache/microphysics_cache.jl
trontrytel: Yeah, I don't know why. But I wanted to test it in a separate PR and keep this big one easy and non-behavior changing.

szy21: It's because I thought we want to limit the tendency with the actual ᶜq_tot, not the clipped one. But it shouldn't matter for this case. ᶜmp_tendency.dq_tot_dt should be 0 when ᶜq_tot is negative.

PR #4329, opened 2026-02-26
Add options for monthly, weekly, and daily for the period

ph-kev:
This PR adds the option for passing "monthly", "weekly", and "daily" for the period. This is useful for accounting for spinup when the spinup is not exactly the same units as the period (e.g. 1 week spinup and n months of simulation).

For example, if `period: monthly`, `reduction_time: average` are used, and the simulation begins on `2010-01-15`, then the first time saved represent the time average of the second half of January, the second time saved represents the time average of February, and so on.

### DIFF
--- a/src/callbacks/get_callbacks.jl
+++ b/src/callbacks/get_callbacks.jl
@@ -242,22 +242,34 @@ function parse_frequency_to_schedule(
         return CAD.DivisorSchedule(steps)
     end
 
+    date_last =
+        t_start isa ITime ?
+        ClimaUtilities.TimeManager.date(t_start) :
+        start_date + Dates.Second(t_start)
+
     if occursin("months", frequency_str)
         months = match(r"^(\d+)months$", frequency_str)
         isnothing(months) && error(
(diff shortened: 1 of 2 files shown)


src/callbacks/get_callbacks.jl
```diff
@@ -242,22 +242,34 @@ function parse_frequency_to_schedule(
...
         months = match(r"^(\d+)months$", frequency_str)
         isnothing(months) && error(
             "$(frequency_str) has to be of the form <NUM>months, e.g. 2months for 2 months",
         )
         period_dates = Dates.Month(parse(Int, first(months)))
+    elseif frequency_str == "monthly"
+        period_dates = Dates.Month(1)
+        date_last = Dates.firstdayofmonth(date_last)
+    elseif frequency_str == "weekly"
+        period_dates = Dates.Week(1)
```
ph-kev, line 260: For WQ, I think there needs to be support for any day (not necessarily the beginning of the week). I am not sure if it should be in this PR or not.

szy21 (approved):

PR #4334, opened 2026-02-28
Make microphysics implicit v3

tapios:
## Purpose 
* Makes microphysics implicit
* Tweaks limiters
* Makes (lognormal) quadrature default
* Needs (small) allocation increase because of additional Jacobian terms
* 2M microphysics is not fully implicit yet (requires larger changes to Jacobian structure for number density terms)

Maybe the third time's a charm...
-->

## Content

----

### DIFF
--- a/src/parameterized_tendencies/microphysics/cloud_fraction.jl
+++ b/src/parameterized_tendencies/microphysics/cloud_fraction.jl
@@ -277,7 +277,7 @@ end
 )
     FT = typeof(Q_hat)
     # Coefficient of variation (protect against division by zero)
-    C_v = sig_s / max(q_tot, ϵ_numerics(FT))
+    C_v = sig_s / max(q_tot, q_min(FT))
 
     # Base coefficient corresponds to Gaussian limit (Cv -> 0)
     coeff = (FT(π) / sqrt(FT(6))) * cf_steepness_scale
@@ -355,6 +355,9 @@ NVTX.@annotate function set_cloud_fraction!(
         isnothing(p.atmos.sgs_quadrature) ? GaussianSGS() :
         p.atmos.sgs_quadrature.dist
 
+    corr_Tq = correlation_Tq(p.params)
+    cf_steepness_scale = CAP.cloud_fraction_steepness_scale(p.params)
+
     @. p.precomputed.ᶜcloud_fraction = compute_cloud_fraction_sd(
         thermo_params,
         ᶜT_mean,
@@ -364,8 +367,8 @@ NVTX.@annotate function set_cloud_fraction!(
         ᶜq_ice,
         ᶜT′T′,
         ᶜq′q′,
-        correlation_Tq(p.params),
-        CAP.cloud_fraction_steepness_scale(p.params),
+        corr_Tq,
+        cf_steepness_scale,
         sgs_dist,
     )
 


src/solver/model_getters.jl
```diff
@@ -396,11 +396,11 @@ Parse the SGS distribution type from configuration.
 - `"mean"` → `GridMeanSGS()` (grid-mean only, no SGS sampling)
 """
 function get_sgs_distribution(parsed_args)
-    dist_name = get(parsed_args, "sgs_distribution", "gaussian")
-    return if dist_name in (nothing, "gaussian")
-        GaussianSGS()
-    elseif dist_name == "lognormal"
+    dist_name = get(parsed_args, "sgs_distribution", "lognormal")
+    return if dist_name in (nothing, "lognormal")
```
trontrytel: Maybe it would be a better pattern to throw an error if the distribution is not specified?

test/parameterized_tendencies/microphysics/tendency_limiters.jl
```diff
@@ -261,11 +263,11 @@ import ClimaAtmos:
             @test abs(limited.dq_sno_dt) <= abs(mp_tendency.dq_sno_dt)
         end
 
-        @testset "temperature-rate limiting" begin
-            # Large condensation → combined temperature limit kicks in.
-            # Rain/snow tendencies are tiny relative to their n_sink=5 budgets
-            # so mass limiting doesn't interfere with them.
-            # Budget for q_rai=0.01: 0.01/(600*5) = 3.3e-6
+        @testset "temperature-rate limiting (removed in implicit branch)" begin
+            # Temperature-rate limiter was removed; tendencies should pass
```
trontrytel: This and below comments - I would stay away from describing how the model used to be (mentioning the removed temperature limiting, etc) and just describe the current logic.

src/cache/prognostic_edmf_precomputed_quantities.jl
```diff
@@ -60,13 +60,16 @@ NVTX.@annotate function set_prognostic_edmf_precomputed_quantities_environment!(
...
         @. ᶜq_tot_safe⁰ = max(ᶜq_liq_rai⁰ + ᶜq_ice_sno⁰, ᶜq_tot⁰)
         ᶜh⁰ = @. lazy(ᶜmse⁰ - ᶜΦ)  # specific enthalpy
-        @. ᶜT⁰ = TD.air_temperature(
-            thermo_params,
-            TD.ph(),
-            ᶜh⁰,
-            ᶜq_tot_safe⁰,
-            ᶜq_liq_rai⁰,
-            ᶜq_ice_sno⁰,
+        @. ᶜT⁰ = max(
```
trontrytel, line 63: Do we need to clip the temperature here? If we are running into unreasonably col temperatures, then its probably a sign of other things going wrong.

src/cache/microphysics_cache.jl
```diff
@@ -34,10 +34,11 @@ end
 """
     ϵ_numerics(FT)
 
-Smallest threshold for specific humidity comparisons.
-Uses cbrt(floatmin(FT)) to detect effectively-zero specific humidities.
+Smallest threshold for specific humidity comparisons and Jacobian
+denominator floors.  Hardcoded to 1e-9 so that the behavior is identical
```
trontrytel: Why do we want the epsilon to be the same for different precision types?

test/parameterized_tendencies/microphysics/tendency_limiters.jl
tapios: Actually, this is gone in the explicit branch too, so the test can just go.

src/cache/prognostic_edmf_precomputed_quantities.jl
tapios: Yes, I prefer to keep it. I do not want the model to fail just because it hit a strange value in a Newton iteration (but can get out of it). If it can't get out of it, it'll fail somewhere else.

src/cache/microphysics_cache.jl
tapios: Because we use it to limit Jacobians, which provide a physical damping on sinks, and I prefer this to be reproducible across precisions. cbrt(floatmin) was much too small for safe implicit solves at Float64.

src/cache/precomputed_quantities.jl
```diff
@@ -563,6 +627,14 @@ NVTX.@annotate function set_implicit_precomputed_quantities!(Y, p, t)
     elseif !(isnothing(turbconv_model))
         # Do nothing for other turbconv models for now
     end
+
+    # When microphysics is implicit, refresh ᶜS_ρq_tot / ᶜS_ρe_tot and the
+    # surface precipitation fluxes so that they reflect the current Y.
+    # The surface flux fields have Dual-typed copies in
+    # implicit_precomputed_quantities, so AD can write into them safely.
+    if p.atmos.microphysics_tendency_timestepping == Implicit()
```
trontrytel, line 632: @dennisYatunin approved feedback - Can we move it to the beginning of `update_jacobian!` for manual sparse Jacobian? This way we would avoid the awkward need to check `if eltype(ᶜS_ρq_tot) === eltype(ᶜ∂Sq_tot)` to know if we are in autodiff or manual jacobian configuration.

src/cache/microphysics_cache.jl
```diff
@@ -718,23 +719,203 @@ function set_precipitation_velocities!(
...
+    (; dt) = p
+    (; ᶜS_ρq_tot, ᶜS_ρe_tot, ᶜmp_tendency, ᶜ∂Sq_tot) = p.precomputed
+    (; ᶜΦ) = p.core
+    # No limit_sink needed: the S/q Jacobian naturally suppresses tendencies
+    # as q→0, and limiting introduces discontinuities that hurt Newton convergence.
+    @. ᶜS_ρq_tot = Y.c.ρ * ᶜmp_tendency.dq_tot_dt
+    @. ᶜS_ρe_tot = ᶜS_ρq_tot * (ᶜmp_tendency.e_int_precip + ᶜΦ)
+    # Pre-compute Jacobian diagonal coefficient S/|q| for update_microphysics_jacobian!
+    # Only compute Jacobian coefficient when not in autodiff mode
+    # (ᶜ∂Sq_tot is Float but ᶜS_ρq_tot may contain Dual numbers during autodiff)
```
trontrytel: Moving this function to `update_jacobian!` for manual sparse Jacobian would make this awkward type checking not needed.

src/cache/microphysics_cache.jl
```diff
@@ -718,23 +719,203 @@ function set_precipitation_velocities!(
...
+- **1M/2M**: full recompute via `set_microphysics_tendency_cache!`.  A cheaper
+  path (refreshing only the Jacobian derivatives) is not used because the BMT
+  Jacobian is bounded and provides no barrier at q=0: Newton can converge to
+  negative q_liq/q_ice when the frozen tendency is a large sink.
+- **default**: no-op (microphysics not active or not implicit).
+"""
+refresh_microphysics_source!(Y, p, _, _) = nothing
+
+# 0M grid-mean (non-EDMF):
+function refresh_microphysics_source!(
```
trontrytel: We are also setting the surface fluxes in this function call, not just the sources. Maybe a name like `update_implicit_microphysics_cache!` would be better?

src/cache/microphysics_cache.jl
```diff
@@ -718,23 +719,203 @@ function set_precipitation_velocities!(
...
+    end
+    set_precipitation_surface_fluxes!(
+        Y, p,
+        EquilibriumMicrophysics0M(),
+    )
+    return nothing
+end
+
+# 0M + DiagnosticEDMFX: mirrors set_microphysics_tendency_cache! for DiagnosticEDMFX —
+# re-aggregate the frozen per-subdomain specific tendencies (ᶜSqₜᵐ⁰, ᶜSqₜᵐʲs) with the
```
trontrytel:
```suggestion
# re-aggregate per-subdomain specific tendencies that were not updated in time (ᶜSqₜᵐ⁰, ᶜSqₜᵐʲs) with the
```

src/cache/microphysics_cache.jl
```diff
@@ -718,23 +719,203 @@ function set_precipitation_velocities!(
...
+# current density (ρ changes at each Newton iterate; precomputed area fractions do not).
+# Without this dispatch the wildcard (0M, _) method is chosen, which uses the grid-mean
+# ᶜmp_tendency and ignores the EDMF area weighting, corrupting the implicit residual.
```
trontrytel:
```suggestion
```

src/cache/microphysics_cache.jl
```diff
@@ -749,10 +930,11 @@ function set_microphysics_tendency_cache!(Y, p, ::EquilibriumMicrophysics0M, _)
     # TODO - The logic in 1m all configs and 0m EDMF is:
     # SG_quad = something(p.atmos.sgs_quadrature, GridMeanSGS())
     # and then always call the quadrature path anyway.
-    # Do we want to do the same here?
+    # We should do the same here. Concern is that ᶜT′T′/ᶜq′q′
```
trontrytel, line 933: Will do!

src/cache/microphysics_cache.jl
```diff
@@ -718,23 +719,203 @@ function set_precipitation_velocities!(
...
+    ::EquilibriumMicrophysics0M,
+    _,
+)
+    (; dt) = p
+    (; ᶜS_ρq_tot, ᶜS_ρe_tot, ᶜmp_tendency, ᶜ∂Sq_tot) = p.precomputed
+    (; ᶜΦ) = p.core
+    # No limit_sink needed: the S/q Jacobian naturally suppresses tendencies
+    # as q→0, and limiting introduces discontinuities that hurt Newton convergence.
+    @. ᶜS_ρq_tot = Y.c.ρ * ᶜmp_tendency.dq_tot_dt
+    @. ᶜS_ρe_tot = ᶜS_ρq_tot * (ᶜmp_tendency.e_int_precip + ᶜΦ)
```
trontrytel, line 766: Why don't we need to update the `mp_tendency` here for the 0M microphysics calls?

trontrytel: Just calling `set_microphysics_tendency_cache!` here, the same way as is done in `1M` configuration would reduce significantly the code duplication introduced here. If not, maybe we should introduce some helper functions that can be re-used between the microphysics cache and refresh functions in the 0M configuration.

src/prognostic_equations/surface_temp.jl
```diff
@@ -2,6 +2,42 @@
...
+    surface_precipitation_tendency!(Yₜ, Y, p, t, surface_model, microphysics_model)
+
+Applies the surface water and energy deposition from precipitation.
+
+When microphysics tendencies are treated implicitly, this function is called
+from `implicit_tendency!` (rather than from the explicit `surface_temp_tendency!`)
+so that both the atmospheric water removal and the surface deposition use the
+same cached `ᶜS_ρq_tot`, preserving conservation across IMEX stages.
+"""
+surface_precipitation_tendency!(Yₜ, Y, p, t, ::PrescribedSST, _) = nothing
```
trontrytel: We could just have ?

`surface_precipitation_tendency!(Yₜ, Y, p, t, _, _) = nothing`

src/prognostic_equations/implicit/implicit_tendency.jl
```diff
@@ -13,6 +13,38 @@ NVTX.@annotate function implicit_tendency!(Yₜ, Y, p, t)
...
+            t,
+            p.atmos.microphysics_model,
+            p.atmos.turbconv_model,
+        )
+        # Apply surface precipitation tendency here alongside atmospheric
+        # removal so both sides of the water/energy budget use the same
+        # cached ᶜS_ρq_tot. Placing the deposition in T_exp would cause
+        # an IMEX stage mismatch (T_imp sees stale ᶜS_ρq_tot, T_exp sees
+        # fresh one), breaking conservation.
+        surface_precipitation_tendency!(
```
trontrytel, line 35: Do we also need it in `remaining_tendency` when running explicit?

src/prognostic_equations/surface_temp.jl
```diff
@@ -2,6 +2,42 @@
...
+surface_precipitation_tendency!(Yₜ, Y, p, t, _, ::DryModel) = nothing
+surface_precipitation_tendency!(Yₜ, Y, p, t, ::PrescribedSST, ::DryModel) =
+    nothing
+surface_precipitation_tendency!(Yₜ, Y, p, t, ::SlabOceanSST, ::DryModel) =
+    nothing
+
+function surface_precipitation_tendency!(
```
trontrytel, line 19: Maybe it would be cleaner to define an implicit and explicit version of this tendency? Rather than setting `pet`, `P_liq` and `P-ice` to zero and keeping the explicit version below unchanged

src/parameterized_tendencies/microphysics/microphysics_jacobian.jl
```diff
@@ -0,0 +1,92 @@
...
+    # For sources (Sq > 0) with |q| > ε: S/q captures accretion-type
+    # ∂S/∂q = S/q exactly, converting the otherwise unstable Picard iteration
+    # (q^{k+1} = q^n + dtγ·S(q^k), which diverges when dtγ·k_acc·q_liq > 1)
+    # into a one-step Newton solve, even when dtγ·S/q > 1 (fast conversion).
+    #
+    # For sources (Sq > 0) with |q| ≈ 0: return zero to avoid the 1/ε
+    # singularity.  In this regime (e.g. autoconversion creating the first
+    # rain drops from q_rain = 0), S is independent of q, so J = I is exact
+    # and the Newton step q^{k+1} = q^n + dtγ·S converges in one iteration.
+    return ifelse(Sq >= zero(FT) && aq <= ε, zero(FT), Sq / max(aq, ε))
```
trontrytel, line 26: Do we still need `max(aq, ε)` if we are already setting the return to zero if `aq <= eps` ?

src/parameterized_tendencies/microphysics/microphysics_jacobian.jl
```diff
@@ -0,0 +1,92 @@
...
+end
+@inline function _jac_coeff(Sq, q)
```
trontrytel: Looks like only `_jac_coeff(Sq, q)` is used in the source code. Can we then combine it with the above `_jac_coeff(Sq, q, ε)` version and just define one? The same comment for below `_jac_coeff_from_ratio`

src/parameterized_tendencies/microphysics/microphysics_jacobian.jl
```diff
@@ -0,0 +1,92 @@
...
+    ε = ϵ_numerics(typeof(Sq))
+    return _jac_coeff(Sq, q, ε)
+end
+
+@inline function _jac_coeff_from_ratio(Sq, ρq, ρ, ε)
```
trontrytel: This repeats the code and logic of `_jac_coeff`. I think we can just have one function, and divide by rho before calling it.

src/parameterized_tendencies/microphysics/microphysics_jacobian.jl
```diff
@@ -0,0 +1,92 @@
...
+already in specific (per-mass) form but the state `ρq` is density-weighted.
+The state is divided by `ρ` inside the fused broadcast to form the specific
+value `q = ρq / ρ`, giving the Jacobian diagonal `Sq / |q|`.
+
+!!! note
+    The division by `ρ` is performed inside the broadcast to avoid creating
+    intermediate lazy fields, which would break broadcast fusion and cause
+    heap allocations.
+"""
+@inline function add_microphysics_jacobian_entry!(∂, dtγ, Sq, ρq, ρ)
```
trontrytel, line 81: If we can divide by rho before calling this function, we would only need one version of `dd_microphysics_jacobian_entry!`

src/cache/microphysics_cache.jl
tapios: It's kept fixed during Newton iterations right now, to avoid quadrature during each iteration. It seems to work fine. But strictly speaking, yes, one should update it like for 1M.

src/prognostic_equations/implicit/implicit_tendency.jl
tapios: When running explicit microphysics, `surface_temp_tendency!` includes the precipitation energy and water contributions.

src/cache/microphysics_cache.jl
trontrytel: Sounds good. We can leave the repetetive code for now. I'll clean it up when merging the next refactoring PR

src/parameterized_tendencies/microphysics/microphysics_jacobian.jl
tapios: We still need the max(aq, ε) for sinks (S<0).

src/parameterized_tendencies/microphysics/microphysics_jacobian.jl
tapios: We could, but we'd need another intermediate lazy field, which have caused problems with broadcast fusion. I'll leave this for now with a comment; maybe someone more experienced can streamline it later.

trontrytel (approved):

PR #4336, opened 2026-02-28
Store all parameters in arrays

dennisYatunin:
## Purpose 
This PR adds arrays for storing parameters, so that they can be loaded into GPU kernels from constant memory instead of parameter memory.

## To-do
Update the `DataLayouts` module in ClimaCore.jl to handle nonuniform data types. This is necessary because the parameters for Insolation.jl contain an integer `DateTime`, which cannot currently be stored in a `CuArray`-backed `DataF` of floating-point values.

----

### DIFF
--- a/src/parameters/Parameters.jl
+++ b/src/parameters/Parameters.jl
@@ -8,6 +8,7 @@ import SurfaceFluxes as SF
 
 abstract type AbstractClimaAtmosParameters end
 const ACAP = AbstractClimaAtmosParameters
+const ACAPField = CC.Fields.Field{<:CC.DataLayouts.DataF{<:ACAP}}
 
 abstract type AbstractTurbulenceConvectionParameters end
 const ATCP = AbstractTurbulenceConvectionParameters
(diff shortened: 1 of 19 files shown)


src/parameters/Parameters.jl
```diff
@@ -18,9 +18,19 @@ const ASTP = AbstractSurfaceTemperatureParameters
...
+# TODO: Check if this works with the new DataLayouts code.
+const fake_point_space = CC.Spaces.PointSpace(CC.DataLayouts.DataF(nothing))
+point_field(x) = CC.Fields.Field(CC.DataLayouts.DataF(x), fake_point_space)
+Base.broadcastable(param_set::ACAP) = point_field(param_set)
+Base.broadcastable(param_set::ATCP) = point_field(param_set)
+Base.broadcastable(param_set::ASTP) = point_field(param_set)
+Base.broadcastable(param_set::IP.InsolationParameters) = point_field(param_set)
+Base.broadcastable(param_set::TD.Parameters.ThermodynamicsParameters) =
+    point_field(param_set)
+Base.broadcastable(param_set::CM.Parameters.ParametersType) =
```
haakon-e: Would this cause invalidations in CloudMicrophysics? There, we set this to `tuple(x)`

dennisYatunin: This code was just for testing. The final PR will replace the entire ClimaAtmosParameters struct with a PointField during model initialization. Allocating a new field on every broadcast like I did here would not be as efficient.

PR #4337, opened 2026-03-02
PEDMF: decouple uj3 implicit solution

sajjadazimi:
## Purpose 
First solve for uj3 and then use the solution to solve for the rest of the variables with only one Newton iteration.

## To-do
Clean up

----


.buildkite/Manifest-v1.11.toml
```diff
@@ -1,8 +1,8 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.11.8"
+julia_version = "1.11.5"
```
szy21: We shouldn't change this. (If you revert the change and update any dependency you need with julia 1.11.8, this won't be changed)

src/solver/type_getters.jl
```diff
@@ -582,6 +583,19 @@ function ode_configuration(::Type{FT}, ode_name, update_jacobian_every,
                 error("Unknown value of `update_jacobian_every`: \
                        $(update_jacobian_every)")
             end,
+        )
+        newtons_method = CTS.NewtonsMethod(;
+            max_iters = 1,
```
szy21: Could you add a comment in the default_config that `max_newton_iters_ode` only controls updraft w?

szy21 (approved):
Thanks! I only skimmed through it. It would be great if someone else could take a look, but I'm also fine with merging as it works:)

akshaysridhar:
cc @imreddyTeja https://buildkite.com/clima/climaatmos-ci/builds/28789#019cb527-75ef-48aa-b9c2-cf29299b53a5/L558 (issue with PrettyTables: `got unsupported keyword arguments "columns_width", "vertical_lines_at_data_columns` )

PR #4347, opened 2026-03-08
Refactor diagnostics: lazy compute API, radiation helpers, and unit tests

haakon-e:
This PR refactors diagnostic variable computation functions to use the lazy `compute` API and adds a unit test suite covering all registered diagnostics.

## Compute API migration

Most diagnostics previously used `compute!`, which required every function to handle two cases — returning a copy when `out` is `nothing`, and writing in-place otherwise. These are replaced with `compute`, which returns a field or lazy broadcast directly; ClimaDiagnostics materialises the result as needed. The handful of diagnostics that write scalar outputs into a pre-allocated buffer (conservation integrals) retain `compute!`, but are simplified to the short-circuit form `isnothing(out) && return [val]; out .= val`.

## Radiation refactor

Repeated geometric-scaling and array-to-field logic is factored into three shared helpers (`ᶠradiation_field_3d`, `ᶠradiation_field_toa`, `ᶠradiation_field_sfc`), reducing each radiation compute function from ~20 lines to 2–3. The mutating `apply_geometric_scaling!` helper is replaced by a pure scalar `geometric_scaling(z, r)`. Also fixes a field name bug in `rldscs` (`cache.radiation_mode` → `cache.radiation`).

## Unit tests

`test/diagnostics/unit_diagnostics.jl` is a unit test suite for all registered diagnostics. Fixtures are collected in a `states` dict keyed by symbol; `case(name, keys)` entries cover multiple dispatch paths per diagnostic in a single declaration. An exhaustiveness check ensures every registered diagnostic is either tested or explicitly listed in `SKIP_CASES`.

## Changed files

- `src/diagnostics/conservation_diagnostics.jl`
- `src/diagnostics/core_diagnostics.jl`
- `src/diagnostics/gravitywave_diagnostics.jl`
- `src/diagnostics/radiation_diagnostics.jl`
- `src/utils/abbreviations.jl` — adds `[U/V/W]Vec` and `UV` abbreviations for
  `Geometry.[U/V/W]Vector` and `Geometry.UVVector`
- `test/diagnostics/unit_diagnostics.jl` — new file

### DIFF
--- a/src/cache/tracer_cache.jl
+++ b/src/cache/tracer_cache.jl
@@ -67,7 +67,7 @@ function tracer_cache(Y, prescribed_aerosol_names, time_varying_trace_gases, sta
         # The keys in the merra2_aerosols.nc file have to match the ones passed with the
         # configuration. The file also has to be defined on the globe and provide
         # time series of lon-lat-z data.
-        prescribed_aerosol_names_as_symbols = Symbol.(prescribed_aerosol_names)
+        prescribed_aerosol_names_as_symbols = Symbol.(Tuple(prescribed_aerosol_names))
         target_space = axes(Y.c)
         extrapolation_bc = (Intp.Periodic(), Intp.Flat(), Intp.Flat())
         timevaryinginputs = [
--- a/src/diagnostics/core_diagnostics.jl
+++ b/src/diagnostics/core_diagnostics.jl
@@ -1,102 +1,67 @@
-# This file is included in Diagnostics.jl
-#
-# README: Adding a new core diagnostic:
-#
-# In addition to the metadata (names, comments, ...), the most important step in adding a
-# new DiagnosticVariable is defining its compute! function. `compute!` has to take four
-# arguments: (out, state, cache, time), and has to write the diagnostic in place into the
-# `out` variable.
-#
-# Often, it is possible to compute certain diagnostics only for specific models (e.g.,
-# humidity for moist models). For that, it is convenient to adopt the following pattern:
-#
-# 1. Define a catch base function that does the computation we want to do for the case we know
-# how to handle, for example
-#
-# function compute_hur!(
-#     out,
-#     state,
-#     cache,
-#     time,
-#     ::T,
-# ) where {T <: MoistMicrophysics}
-#     thermo_params = CAP.thermodynamics_params(cache.params)
-#     out .= TD.relative_humidity.(thermo_params, cache.precomputed.ᶜT, state.c.ρ, cache.precomputed.ᶜq_tot_nonneg, cache.precomputed.ᶜq_liq, cache.precomputed.ᶜq_ice))
-# end
-#
-# 2. Define a function that has the correct signature and calls this function
-#
-# compute_hur!(out, state, cache, time) =
-#     compute_hur!(out, state, cache, time, cache.atmos.microphysics_model)
-#
-# 3. Define a function that returns an error when the model is incorrect
-#
-# compute_hur!(_, _, _, _, model::T) where {T} =
-#     error_diagnostic_variable("relative_humidity", model)
-#
-# We can also output a specific error message
-#
-# compute_hur!(_, _, _, _, model::T) where {T} =
-#     error_diagnostic_variable("relative humidity makes sense only for moist models")
+#=
+This file is included in Diagnostics.jl
 
-# General helper functions for undefined diagnostics for a particular model
-error_diagnostic_variable(
-    message = "Cannot compute $variable with model = $T",
-) = error(message)
-error_diagnostic_variable(variable, model::T) where {T} =
+README: Adding a new core diagnostic:
+
+In addition to the metadata (names, comments, ...), the most important step
+in adding a new DiagnosticVariable is defining its compute function. 
+`compute` takes three arguments: (state, cache, time), 
+and returns the diagnostic value, preferring lazy fields when possible.
+
+Often, it is possible to compute certain diagnostics only for specific models
+(e.g., humidity for moist models). For that, it is convenient to adopt the following pattern:
+
+1. Define a function for the case we know how to handle:
+
+function compute_hur(state, cache, time, ::MoistMicrophysics)
+    tps = CAP.thermodynamics_params(cache.params)
+    (; ᶜT, ᶜq_tot_nonneg, ᶜq_liq, ᶜq_ice) = cache.precomputed
+    @. lazy(TD.relative_humidity(tps, ᶜT, state.c.ρ, ᶜq_tot_nonneg, ᶜq_liq, ᶜq_ice))
+end
+
+2. Define a dispatch function with the correct signature:
+
+compute_hur(state, cache, time) =
+    compute_hur(state, cache, time, cache.atmos.microphysics_model)
+
+3. Define an error fallback for unsupported models:
+
+compute_hur(state, cache, time, model) = error_diagnostic_variable("hur", model)
+
+=#
+
+# General helper functions for undefined diagnostics
+error_diagnostic_variable(message = "Cannot compute variable") = error(message)
+error_diagnostic_variable(variable, ::T) where {T} =
     error_diagnostic_variable("Cannot compute $variable with model = $T")
 
 ###
 # Density (3d)
 ###
-add_diagnostic_variable!(
-    short_name = "rhoa",
-    long_name = "Air Density",
-    standard_name = "air_density",
-    units = "kg m^-3",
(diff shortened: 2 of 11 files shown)


src/cache/tracer_cache.jl
```diff
@@ -67,7 +67,7 @@ function tracer_cache(Y, prescribed_aerosol_names, time_varying_trace_gases, sta
         # The keys in the merra2_aerosols.nc file have to match the ones passed with the
         # configuration. The file also has to be defined on the globe and provide
         # time series of lon-lat-z data.
-        prescribed_aerosol_names_as_symbols = Symbol.(prescribed_aerosol_names)
+        prescribed_aerosol_names_as_symbols = Symbol.(Tuple(prescribed_aerosol_names))
```
haakon-e, line 70: fixes an issue I discovered while writing the unit tests, where this function would fail if the aerosol names are passed as a vector instead of tuple. This path is not hit with yaml files, but may be hit with runscripts. This explicitly converts the vector into a tuple

src/setups/common/prognostic_variables.jl
```diff
@@ -151,7 +151,10 @@ function turbconv_center_variables(
         )
     else  # NonEquilibriumMicrophysics2M
         sgsʲs = uniform_subdomains(
-            (; ρa, mse, q_tot, q_lcl = q_liq, q_icl = q_ice, q_rai, q_sno, n_liq, n_rai),
+            (; ρa, mse, q_tot,
+                q_lcl = q_liq, q_icl = q_ice, q_rai, q_sno,
+                n_lcl = n_liq, n_rai,
+            ),
```
haakon-e, line 157: the only real change here is setting `n_lcl = n_liq`, consistent with e.g. #4389

src/utils/abbreviations.jl
```diff
@@ -12,6 +12,9 @@ const CT2 = Geometry.Contravariant2Vector
 const CT12 = Geometry.Contravariant12Vector
 const CT3 = Geometry.Contravariant3Vector
 const CT123 = Geometry.Contravariant123Vector
+const UV = Geometry.UVVector
+const VVec = Geometry.VVector
+const WVec = Geometry.WVector
```
haakon-e: non-exhaustive list of new, useful abbreviations.

src/utils/abbreviations.jl
```diff
@@ -12,6 +12,10 @@ const CT2 = Geometry.Contravariant2Vector
 const CT12 = Geometry.Contravariant12Vector
 const CT3 = Geometry.Contravariant3Vector
 const CT123 = Geometry.Contravariant123Vector
+const UVec = Geometry.UVector
+const VVec = Geometry.VVector
+const WVec = Geometry.WVector
+const UV = Geometry.UVVector
```
haakon-e, line 18: non-exhaustive list of new, useful abbreviations.

test/runtests.jl
```diff
@@ -40,6 +40,9 @@ if TEST_GROUP in ("infrastructure", "all")
     @safetestset "Model getters" begin @time include("solver/model_getters.jl") end
     @safetestset "AtmosModel Constructor" begin @time include("solver/atmos_model_constructor.jl") end
     @safetestset "Topography tests" begin @time include("topography.jl") end
+
+    # Diagnostics unit tests
+    @safetestset "Diagnostics unit tests" begin @time include("diagnostics/unit_diagnostics.jl") end
```
haakon-e, line 50: The primary issue with the introduced set of unit tests is that, due to the volume of the diagnostics, and that all tests within a test group are run serially (instead of in parallel), another ~8 minutes of GitHub CI time is added to the infrastructure tests, increasing them from >20 minutes to >30 minutes.

haakon-e (commented):
a few notes for review

src/setups/common/prognostic_variables.jl
trontrytel: I tried not to touch 2MP3 in the PR I merged. But hopefully at some point it would be using variables:
`q_tot`, `q_lcl`, `q_rai`, `q_ice`?

test/runtests.jl
trontrytel: Would it make sense to split the infrastructure test group into two? Or do we think we will wait more for the resources to be assigned than for the tests to finish?

src/diagnostics/radiation_diagnostics.jl
```diff
@@ -1,1145 +1,505 @@
...
 
-add_diagnostic_variable!(
-    short_name = "rsu",
+add_diagnostic_variable!(short_name = "rsu", units = "W m^-2",
     long_name = "Upwelling Shortwave Radiation",
     standard_name = "surface_upwelling_shortwave_flux_in_air",
-    units = "W m^-2",
-    comments = "Upwelling shortwave radiation",
-    compute! = compute_rsu!,
+    comments = "Upwelling shortwave radiation at the surface",
```
trontrytel: Are we sure nobody downstream expects it to be a 3D field for some obscure reason?

src/diagnostics/core_diagnostics.jl
```diff
@@ -1141,404 +627,216 @@ add_diagnostic_variable!(
...
-    end
+    # vertical integral of saturation specific humidity
+    sat_vi = cache.scratch.ᶠtemp_field_level
+    sat = @. lazy(
+        state.c.ρ * TD.q_vap_saturation(thermo_params, ᶜT, state.c.ρ, ᶜq_liq, ᶜq_ice),
+    )
+    Operators.column_integral_definite!(sat_vi, sat)
+    # saturation-weighted vertical integral
+    hur_weighted = @. lazy(state.c.ρ * ᶜq_vap / sat_vi)
+    out = cache.scratch.ᶠtemp_field_level
```
trontrytel, line 812: Would it be safer to make it use something else than `ᶠtemp_field_level`? This is used to save `sat_vi`, which is then used in `hur_weighted`. with lazy expressions, I'm not sure when exactly things get materialized

src/diagnostics/core_diagnostics.jl
```diff
@@ -1,1138 +1,624 @@
...
-    end
-end
+compute_husra(state, cache, time) =
+    compute_husra(state, cache, time, cache.atmos.microphysics_model)
+compute_husra(_, _, _, model) = error_diagnostic_variable("husra", model)
 
-add_diagnostic_variable!(
-    short_name = "husra",
+compute_husra(state, _, _,
+    ::Union{NonEquilibriumMicrophysics1M, NonEquilibriumMicrophysics2M},
```
trontrytel, line 563: We don't want it for 2MP3?

trontrytel (approved):
Thank you!

src/setups/common/prognostic_variables.jl
haakon-e: yes. This will follow in the actual P3 PR. Here, the change is needed to make tests pass

src/diagnostics/radiation_diagnostics.jl
haakon-e: Thanks for spotting this error, fixed back to 3d.

src/diagnostics/core_diagnostics.jl
haakon-e: `column_integral_definite!` ([`integrals.jl:15-23`](https://github.com/CliMA/ClimaCore.jl/blob/27c98a15209a57d944a17622fd5ff81f95939b50/src/Operators/integrals.jl#L15-L23)) calls `column_reduce!`, which dispatches to `single_column_reduce!` ([`integrals.jl:182-204`](https://github.com/CliMA/ClimaCore.jl/blob/27c98a15209a57d944a17622fd5ff81f95939b50/src/Operators/integrals.jl#L182-L204)):
```Julia
# integrals.jl:182-204
function single_column_reduce!(f, transform, _output, _input, init, space)
    first_level = left_idx(space)
    last_level = right_idx(space)
    reduced_value = init
    @inbounds for level in next_level:last_level              # READS all levels first
        reduced_value = f(reduced_value, get_level_value(space, _input, level))
    end
    Fields.field_values(_output)[] = transform(reduced_value)  # WRITES once, after loop
end
```
All vertical-level reads complete before the single scalar write to output. Thus, it's safe to use `ᶠtemp_field_level` in both the input and output.

src/diagnostics/core_diagnostics.jl
haakon-e: I'll remove 2MP3 anyways in the P3 PR.

szy21:
@copilot review this PR

PR #4382, opened 2026-03-20
Convective gravity wave source based on Beres et al. (2004) driven by the prognostic EDMF scheme

ray-chew:
## Purpose
Add a Beres et al. (2004) convective gravity-wave source to the Alexander–Dunkerton (1999) non-orographic GW scheme, coupled to prognostic EDMFX. The convective source is additive: AD99 supplies the background broad spectrum everywhere, and Beres supplies a convective spectrum whose shape and amplitude are set by EDMF heating. Each source is launched and broken independently (each with its own intermittency), so deposition is computed twice per column with active convective gravity waves.

The Beres scheme is enabled with `non_orographic_gravity_wave: true` + `nogw_beres_source: true` (requires EDMF).

## Content
- **Beres convective source** (`src/parameterized_tendencies/gravity_wave_drag/non_orographic_gravity_wave.jl`):   launches at the convective envelope top; the heating amplitude `Q0` is derived from  the EDMF updraft mass-flux divergence (in-cloud amplitude with cloud coverage/area fraction weighting).
- **Unified Beres/AD99 kernel** with a moment-matched half-sine heating envelope.
- **Steady (ν ≈ 0) stationary-wave component**, triggered by choice of the phase speed `c` grid, now always computed and  deposited at `c ≈ 0`.
- **Optional canonical latent heating** (`nogw_beres_heating_latent`, requires   `microphysics_model: 1M`) as an alternative to the dry static energy flux-divergence Yanai-Q₁ source. 
- **Parameters migrated to ClimaParams TOML** under the `nogw_beres_*` prefix  (`sigma_x`, `nu_min/max`, `h_heat_min`, `Q0_threshold`, `scale_factor`, …).
- **Diagnostics**: the CliMA dycore gets only `utendnogw`/`vtendnogw`. The source-internal verification fields (e.g., `Q0`, `h_heat`, `z_bot/z_top`, `a_cover`, `Q_conv`, half-sine,  launched-spectrum summaries) are activated by toggling `nogw_beres_detailed_diagnostics`.
- **Docs** are now split into `non_orographic_gravity_wave.md` and `orographic_gravity_wave.md`, as the gravity wave documentation files are getting longer.
- **Tests / CI**: unit, single-column, and sphere-integration suites, plus two Buildkite jobs and the `single_column_beres_nogw_test` / `beres_squall_integration_test` model configs.
- **`ref_counter`** updated as the steady-source phase-speed grid (`cmax/dc = 125`) shifts the NOGW bins for every NOGW run (AD99 included).

## Tuning note
`nogw_beres_scale_factor` (α) is the one free amplitude knob. Separate tuning for ρ₀/(Lτ), the |Q̂|² weight, and tuning are all combined into this one tuning parameter. It may need calibration so the Beres convective and AD99 background spectra sit at comparable magnitudes.

## Verification
The 9-panel moment-matching figure below shows the half-sine source centroid tracking the EDMF in-cloud heating, with sensible drag in the deposition panels for an instantaneous snapshot at day 120 of a 0M aquaplanet free run with prognostic EDMF.

image

----


docs/src/non_orographic_gravity_wave.md
szy21, line 1: Should we remove the current docs/src/gravity_wave.md?

src/callbacks/callbacks.jl
```diff
@@ -336,6 +336,21 @@ function reset_graceful_exit(output_dir)
...
+                field = getproperty(sub, fn)
+                if any(isnan, parent(field))
+                    n_nan = count(isnan, parent(field))
+                    n_tot = length(parent(field))
+                    @info "NaN found in Y.$pn.$fn: $n_nan / $n_tot elements"
+                end
+            end
+        end
+        error("Found NaN")
+    end
```
szy21, line 362: does this work on GPU?

src/config/model_getters.jl
```diff
@@ -292,7 +319,63 @@ function get_non_orographic_gravity_wave_model(
...
+                z_bot_Q_threshold = FT(bsp.z_bot_Q_threshold),
+                z_bot_floor = FT(bsp.z_bot_floor),
+                beres_steady_dc_frac = FT(bsp.steady_dc_frac),
+                beres_L_system = FT(bsp.L_system),
+                # Two deprecation notes:
+                # beres_steady_source defaults true on the struct (no YAML switch); it
+                # is toggled implicitly by the phase-speed grid (deposits only with a c=0
+                # bin, see the warning below). moment_matched is the sole active
+                # envelope mode (legacy area_threshold is deprecated but retained).
+                moment_envelope = true,
```
szy21: should we remove the moment_envelope params and the area_threshold branch in the code? It seems we don't have moment_envelope: false anywhere.

config/model_configs/beres_squall_integration_test.yml
szy21, line 1: Maybe add the toml config directly to the test file instead of adding another yaml file?

config/longrun_configs/longrun_aquaplanet_allsky_progedmf_0M.yml
szy21, line 1: let's add these configs to amip_target.yml; that run contains the most complicated physics. (You can remove them here, or keep them if you want.)

src/diagnostics/gravitywave_diagnostics.jl
akshaysridhar, line 1: The added list of diagnostics is quite useful, thank you! Could you please enable a subset as generated artifacts (the current [ci-build](https://buildkite.com/clima/climaatmos-ci/builds/30423#019f06c5-0cb8-48cd-9fbf-d92d3ddd3dbf) appears to have none for the two relevant test cases)

src/parameterized_tendencies/gravity_wave_drag/non_orographic_gravity_wave.jl
```diff
@@ -8,6 +8,55 @@ import ClimaCore.Spaces as Spaces
...
+import CloudMicrophysics.BulkMicrophysicsTendencies as BMT
+
+"""
+    _beres_latent_heating(mp, thp, ρ, T, q_tot, q_lcl, q_icl, q_rai, q_sno)
+
+Beres' canonical transport-free latent heating for one updraft:
+
+    Q_lat = (1/cp^{(j)}) Σ_p L_p R_p^{(j)}   [K/s]
+
+where `p` runs over the microphysical phase-change processes (e.g., condenstation, evaporation). `L_p` is the latent heat released or absorbed by process `p` [J/kg] and `R_p^{(j)}` is the mass-mixing-ratio tendency [kg/kg/s]. These rates can be built from CloudMicrophysics 1-moment aggregated phase-change tendencies `dq_{lcl,icl,rai,sno}_dt`. `cp^{(j)}` is the subdomain moist heat capacity `TD.cp_m(thp, q_tot, q_lcl, q_icl)`
```
akshaysridhar:
```suggestion
where `p` runs over the microphysical phase-change processes (e.g., condensation, evaporation). `L_p` is the latent heat released or absorbed by process `p` [J/kg] and `R_p^{(j)}` is the mass-mixing-ratio tendency [kg/kg/s]. These rates can be built from CloudMicrophysics 1-moment aggregated phase-change tendencies `dq_{lcl,icl,rai,sno}_dt`. `cp^{(j)}` is the subdomain moist heat capacity `TD.cp_m(thp, q_tot, q_lcl, q_icl)`
```

src/parameterized_tendencies/gravity_wave_drag/non_orographic_gravity_wave.jl
```diff
@@ -54,6 +103,30 @@ function non_orographic_gravity_wave_cache(Y, gw::NonOrographicGravityWave)
             uforcing = zero(Y.c.ρ),
             vforcing = zero(Y.c.ρ),
             gw_ncval = Val(nc),
+            # Beres fields (always allocated)
```
akshaysridhar: Do we always need to allocate these fields ? Is it a reasonable option to selectively allocate cache based on, e.g. `gw::NonOrographicGravityWave{FT, <:BeresSourceParams})` or `gw::NonOrographicGravityWave{FT, Nothing})`

src/parameterized_tendencies/gravity_wave_drag/non_orographic_gravity_wave.jl
```diff
@@ -277,6 +396,452 @@ function non_orographic_gravity_wave_compute_tendency!(
...
+                result_field,
+                input1;
+                init = reduce_init,
+            ) do (I0p, I1p, I2p, _4, _5, _6, _7), (z, Q_ic, dz)
+                qpdz = ifelse(z >= _z_floor, max(Q_ic, _zero) * dz, _zero)
+                return (
+                    I0p + qpdz,
+                    I1p + z * qpdz,
+                    I2p + z * z * qpdz,
+                    _4,
```
akshaysridhar: What is the purpose of `_4 ... _7` here ?

akshaysridhar: Ah, they're assigned meaning in Pass2 via
```
        N_sum_prev,
        m_sum_prev,
        Qic_int_prev,
        a_sum_prev,
```

akshaysridhar: I'd suggest adding a comment tying the tuple organisation here (Pass 1) to the code that follows (in Pass 2)

src/parameterized_tendencies/gravity_wave_drag/non_orographic_gravity_wave.jl
```diff
@@ -277,6 +396,452 @@ function non_orographic_gravity_wave_compute_tendency!(
...
+    if n_updrafts == 0
+        # No EDMF — Beres inactive everywhere
+        p.non_orographic_gravity_wave.gw_beres_active .= 0
+        p.non_orographic_gravity_wave.gw_a_cover .= 0
+        p.non_orographic_gravity_wave.gw_Q_conv_ic .= 0
+        p.non_orographic_gravity_wave.gw_halfsine .= 0
+        p.non_orographic_gravity_wave.gw_launch_flux .= 0
+        p.non_orographic_gravity_wave.gw_c_centroid .= 0
+        return
+    end
```
szy21: We probably don't need this check as we already check turbconv must be prognostic EDMF when nogw_beres_source is true when setting up the config.

src/parameterized_tendencies/gravity_wave_drag/non_orographic_gravity_wave.jl
```diff
@@ -277,6 +396,452 @@ function non_orographic_gravity_wave_compute_tendency!(
...
+    ᶜa_scalar_ic = p.scratch.ᶜtemp_scalar_5
+    ᶜρa_sum = p.scratch.ᶜtemp_scalar_6
+
+    # For DiagnosticEDMFX, ρa is in p.precomputed
+    # For PrognosticEDMFX, ρa is in Y.c.sgsʲs.:($j)
+    has_prognostic_sgs =
+        hasproperty(Y.c, :sgsʲs) && n_updrafts > 0
+    if !has_prognostic_sgs && haskey(p.precomputed, :ᶜρaʲs)
+        ᶜρaʲs_all = p.precomputed.ᶜρaʲs
+    end
```
szy21: DiagnosticEDMFX has been removed, so you can remove the branch for getting ᶜρaʲs here and below

src/parameterized_tendencies/gravity_wave_drag/non_orographic_gravity_wave.jl
```diff
@@ -277,6 +396,452 @@ function non_orographic_gravity_wave_compute_tendency!(
...
+        @. ᶜa_scalar_ic =
+            ifelse(ᶜρʲs.:($$j) > eps(FT), cp_d * (ᶜTʲs.:($$j) - ᶜT), FT(0))
+
+        # DSE anomaly × area fraction: cp_d·(Tʲ − T̄) · (ρaʲ/ρʲ)
+        @. ᶜa_scalar =
+            ifelse(
+                ᶜρʲs.:($$j) > eps(FT),
+                ᶜa_scalar_ic * (ᶜρaʲ / ᶜρʲs.:($$j)),
+                FT(0),
+            )
```
szy21, line 474: Do we need the `ᶜρʲs.:($$j) > eps(FT)` check here?

szy21:
Should the Beres source be added on top of AD99, or replace it?

ray-chew:
> Should the Beres source be added on top of AD99, or replace it?

It should be added on top of it. AD99 is now fully responsible for the background source only, while Beres is the convective GW source (which uses the AD99 propagator for breaking and deposition).

docs/src/non_orographic_gravity_wave.md
ray-chew: Done. Removed `docs/src/gravity_wave.md`. It was already superseded by the split into `non_orographic_gravity_wave.md` + `orographic_gravity_wave.md`. They are already registered in `make.jl`.

src/callbacks/callbacks.jl
ray-chew: Yes, should be. `any(isnan, parent(...))` is the same GPUArrays reduction we already use across the GW code, and the only new op, `count(isnan, parent(field))`, dispatches to `Base.count(pred, ::AnyGPUArray)` in GPUArrays.

src/config/model_getters.jl
ray-chew: I kept it as I thought maybe we might have to revert to the old threshold computations. But moments matching seems to be robust enough. Removed the `moment_envelope` switch and the whole deprecated area_threshold branch, plus its exclusive param `z_bot_Q_threshold`. Cleaned up in `types.jl`, `Parameters.jl`, `create_parameters.jl`, `model_getters.jl`, and `non_orographic_gravity_wave.jl`. The `nogw_beres_z_bot_Q_threshold` default in ClimaParam has also been removed.

config/model_configs/beres_squall_integration_test.yml
ray-chew: Done. Remove the `beres_squall_integration_test.yml` fixture and set the toml override inline in `test_beres_sphere_integration.jl`. I kept `toml/nogw_beres_test.toml` rather than re-typing its 4 values into the test, since that toml is shared with the single-column Beres test.

config/longrun_configs/longrun_aquaplanet_allsky_progedmf_0M.yml
ray-chew: Added `non_orographic_gravity_wave: true`, `nogw_beres_source: true`, `dt_nogw: "400secs"` to `amip_target.yml`. I kept it enabled on `longrun_aquaplanet_allsky_progedmf_{0/1}M.yml` too, since that's the run behind the manuscript QBO results. Left the source on the default DSE-Q₁ heating (not the 1M latent path) so both runs behave identically.

src/diagnostics/gravitywave_diagnostics.jl
ray-chew: Thanks! This is now covered by the "Beres NOGW: convective source verification plot (GPU)" job. It runs `sphere_beres_verification.yml` with `nogw_beres_detailed_diagnostics: true`, writes the GW source diagnostics as NetCDF, and uploads `sphere_beres_verification/output_active/*` with both the diagnostic files and the rendered 9-panel (.png and .pdf).

src/parameterized_tendencies/gravity_wave_drag/non_orographic_gravity_wave.jl
ray-chew: Done, thank you!

src/parameterized_tendencies/gravity_wave_drag/non_orographic_gravity_wave.jl
ray-chew: Oh right, I overlooked these. These fields aren't shared with AD99, so an AD99-only NOGW run was allocating them for nothing (3 full 3D fields + ~13 level fields). I moved them behind a `_beres_cache_fields(Y, gw.beres_source)` helper that returns `(;)` when `beres_source === nothing`. All runtime access was already behind `!isnothing(gw_beres_source)`, and the diagnostics already dispatch on `{FT, Nothing}`, so this just aligns the cache with that assumption.

src/parameterized_tendencies/gravity_wave_drag/non_orographic_gravity_wave.jl
ray-chew: Added a comment tying the two together as you suggested, thansk!

src/parameterized_tendencies/gravity_wave_drag/non_orographic_gravity_wave.jl
ray-chew: Removed it. You're right, the config setup already errors unless `turbconv = prognostic_edmfx`.

src/parameterized_tendencies/gravity_wave_drag/non_orographic_gravity_wave.jl
ray-chew: Done, removed the fallback.

src/parameterized_tendencies/gravity_wave_drag/non_orographic_gravity_wave.jl
ray-chew: I think we should keep it. The one place it matters is the top of the domain: at ~45–80 km the pressure is tiny, and in Float32 `ρʲ = p/(R·T)` can drop to ~1e-7, right around eps(Float32). That's why the `/ ᶜρʲs` divisions just below genuinely need the guard there. Added a comment to that effect.

szy21 (approved):

PR #4412, opened 2026-04-02
Cleanup in 0M and 1M non-quadrature and quadrature paths

trontrytel:
This PR unifies the microphysics tendency evaluation logic for 0-moment and 1-moment schemes. Previously, grid-mean and SGS-quadrature evaluations used separate functions with inconsistent signatures and duplicate external limiting logic. 

- Both 0M and 1M schemes now have two pathways - using the evaluator on quadrature points or directly calling BMT (when in updrafts or running without SGS fluctuations). We can't just have one pathway, because the arguments passed (variances) and the logic (saturation adjustment vs passing in cloud condensate) is inherently different.
- I added one definition of what it means without SGS fluctuations.
- All the if/else statements that I added are known at compile time and should not result in any run-time slow-downs.
- I put the explicit limiters inside the microphysics_tendencies to avoid code duplication.
- I changed the tendencies and limiters from mutating things in place to returning the result (in line with what we discussed in microphysics refactor issue https://github.com/CliMA/ClimaAtmos.jl/issues/4285)

Out of curiosity, I also added some simple timing test for the direct BMT vs non-quadrature vs quadrature runs. The test accuracy is not super high, because I did not want to add `BenchmarkTools` to the dependencies. But here is a quick glance at the quadrature penalty we have for 0M and 1M:

```
┌ Info: 0M Performance (10000 calls)
│   t_direct_0m = 0.0036955
│   t_wrapper_0m = 0.000869083
│   timings_0m =
│    Dict{Int64, Float64} with 3 entries:
│      5 => 0.0130378
│      3 => 0.00598933
└      1 => 0.00185075
┌ Info: 1M Performance (10000 calls)
│   t_direct_1m = 0.022510667
│   t_wrapper_1m = 0.014284125
│   timings_1m =
│    Dict{Int64, Float64} with 3 entries:
│      5 => 0.397994
│      3 => 0.168257
└      1 => 0.039695
```

### DIFF
--- a/src/cache/diagnostic_edmf_precomputed_quantities.jl
+++ b/src/cache/diagnostic_edmf_precomputed_quantities.jl
@@ -800,52 +800,38 @@ NVTX.@annotate function set_diagnostic_edmf_precomputed_quantities_do_integral!(
             nh_pressure³_dragʲ_data_prev_halflevel =
                 nh_pressure³_dragʲ_prev_halflevel.components.data.:1
 
-            # Microphysics sources and sinks. To be applied in updraft continuity,
+            # Microphysics sources and sinks. To be applied in continuity,
             # moisture and energy equations for updrafts and grid mean.
-
-            # 0-moment microphysics: sink of q_tot from precipitation removal
             if microphysics_model isa EquilibriumMicrophysics0M
-                @. ᶜmp_tendencyʲ_prev_level.dq_tot_dt =
-                    BMT.bulk_microphysics_tendencies(
-                        BMT.Microphysics0Moment(),
-                        microphys_0m_params,
-                        thermo_params,
-                        Tʲ_prev_level,
-                        q_liqʲ_prev_level,
-                        q_iceʲ_prev_level,
-                        TD.q_vap_saturation(
-                            thermo_params, Tʲ_prev_level, ρʲ_prev_level,
-                        ),
-                    )
-                @. ᶜmp_tendencyʲ_prev_level.dq_tot_dt = limit_sink(
-                    ᶜmp_tendencyʲ_prev_level.dq_tot_dt,
-                    q_totʲ_prev_level, dt,
-                )
-                @. ᶜmp_tendencyʲ_prev_level.e_tot_hlpr =
-                    e_tot_0M_precipitation_sources_helper(
-                        thermo_params,
-                        Tʲ_prev_level,
-                        q_liq_prev_level,
-                        q_ice_prev_level,
-                        Φ_prev_level,
-                    )
-                # 1-moment microphysics: cloud water (liquid and ice) and
-                # precipitation (rain and snow) tendencies. q_tot is constant, because
-                # all the species are considered a part of the working fluid.
-            elseif microphysics_model isa NonEquilibriumMicrophysics1M
-                # Microphysics tendencies from the updrafts (using fused BMT API)
-                compute_1m_precipitation_tendencies!(
-                    ᶜmp_tendencyʲ_prev_level,
+                # Sink of total water and energy from precipitation removal.
+                @. ᶜmp_tendencyʲ_prev_level = microphysics_tendencies_0m(
+                    microphys_0m_params,
+                    thermo_params,
                     ρʲ_prev_level,
-                    q_totʲ_prev_level,
+                    Tʲ_prev_level,
+                    q_tot_nonnegʲ_prev_level,
+                    q_liqʲ_prev_level,
+                    q_iceʲ_prev_level,
+                    Φ_prev_level,
+                    $(p.atmos.microphysics_tendency_timestepping),
+                    dt,
+                )
+            elseif microphysics_model isa NonEquilibriumMicrophysics1M
+                # Cloud water (liquid and ice) and precipitation
+                # (rain and snow) tendencies. Total water is constant, because
+                # all water species are considered part of the working fluid.
+                @. ᶜmp_tendencyʲ_prev_level = microphysics_tendencies_1m(
+                    ρʲ_prev_level,
+                    q_tot_nonnegʲ_prev_level,
                     q_lclʲ_prev_level,
                     q_iclʲ_prev_level,
                     q_raiʲ_prev_level,
                     q_snoʲ_prev_level,
                     Tʲ_prev_level,
-                    dt,
                     microphys_1m_params,
                     thermo_params,
+                    $(p.atmos.microphysics_tendency_timestepping),
+                    dt,
                 )
             end
 
--- a/src/parameterized_tendencies/microphysics/microphysics_wrappers.jl
+++ b/src/parameterized_tendencies/microphysics/microphysics_wrappers.jl
@@ -61,12 +61,10 @@ struct Microphysics0MEvaluator{CMP, SAE, FT}
     sat_eval::SAE
     Φ::FT
 end
-
 function Microphysics0MEvaluator(cm_params, thermo_params, ρ, Φ)
     sat_eval = SaturationAdjustmentEvaluator(thermo_params, ρ)
     return Microphysics0MEvaluator(cm_params, sat_eval, Φ)
 end
-
 @inline function (eval::Microphysics0MEvaluator)(T_hat, q_hat)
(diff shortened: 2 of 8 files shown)


src/cache/diagnostic_edmf_precomputed_quantities.jl
```diff
@@ -805,47 +805,35 @@ NVTX.@annotate function set_diagnostic_edmf_precomputed_quantities_do_integral!(
...
                 )
-                @. ᶜmp_tendencyʲ_prev_level.e_tot_hlpr =
-                    e_tot_0M_precipitation_sources_helper(
-                        thermo_params,
-                        Tʲ_prev_level,
-                        q_liq_prev_level,
-                        q_ice_prev_level,
-                        Φ_prev_level,
-                    )
                 # 1-moment microphysics: cloud water (liquid and ice) and
```
tapios: It's confusing to have the code comment above the elseif branch to which is belongs--it makes it look like it belongs to the branch above (0M); the indentation contributes to that impression. I think this would be clearer below the corresponding elseif (and same comment applies to the other branches).

src/parameterized_tendencies/microphysics/microphysics_wrappers.jl
```diff
@@ -90,107 +88,84 @@ end
...
-    microphysics_tendencies_quadrature_0m(
-        SG_quad, cm_params, thermo_params,
-        ρ, T_mean, q_tot_mean, T′T′, q′q′, corr_Tq, Φ,
-    )
+    microphysics_tendencies_0m(SG_quad, cmp, thp, ρ, T, q_tot, T′T′, q′q′, corr_Tq, Φ, tst, dt)
+    microphysics_tendencies_0m(cmp, thp, ρ, T, q_tot, q_liq, q_ice, Φ, tst, dt)
+
+Computes 0-moment microphysics tendencies.
+
+When using SGS-quadratures, the tendecies are integrated over the joint PDF of (T, q_tot).
```
tapios:
```suggestion
When using SGS-quadratures, the tendencies are integrated over the joint PDF of (T, q_tot).
```

src/parameterized_tendencies/microphysics/microphysics_wrappers.jl
```diff
@@ -90,107 +88,84 @@ end
...
+At each quadrature point condensate is diagnosed from saturation excess,
```
tapios:
```suggestion
At each quadrature point, condensate is diagnosed from saturation excess,
```

src/parameterized_tendencies/microphysics/microphysics_wrappers.jl
```diff
@@ -90,107 +88,84 @@ end
...
+then 0M precipitation removal tendencies are computed.
 
-Compute SGS-averaged 0-moment microphysics tendencies by integrating over
-the joint PDF of (T, q_tot). At each quadrature point, condensate is diagnosed
-from saturation excess, then 0M precipitation removal tendencies are computed.
+The option without SGS-quadratures can be used in the EDMF updrafts, or to compute
+grid-mean tendency without taking account of fluctuations. It computes the
+0M precipitation removal tendecies based on provided point-value of temperature
```
tapios:
```suggestion
0M precipitation removal tendecies based on the provided point-values of temperature
```

src/parameterized_tendencies/microphysics/microphysics_wrappers.jl
```diff
@@ -90,107 +88,84 @@ end
...
+and specific humidities.
+
+When running with explicit microphysics timestepping the total water sink is limited
```
tapios:
```suggestion
When running with explicit microphysics timestepping, the total water sink is limited
```

src/parameterized_tendencies/microphysics/microphysics_wrappers.jl
```diff
@@ -90,107 +88,84 @@ end
...
+    evaluator = Microphysics0MEvaluator(cmp, thp, ρ, Φ)
     # Integrate over quadrature points; both dq_tot_dt and e_tot_hlpr
     # are averaged over the SGS distribution.
-    return integrate_over_sgs(
-        evaluator, SG_quad, q_tot_mean, T_mean, q′q′, T′T′, corr_Tq,
+    (; dq_tot_dt, e_tot_hlpr) = integrate_over_sgs(
+        evaluator, SG_quad, q_tot_nonneg, T, q′q′, T′T′, corr_Tq,
     )
+    # Apply limiter if using explicit microphysics timestepping
+    if tst isa Explicit
```
tapios, line 136: This is fine in principle. But it assumes we will never use limiters with implicit timestepping. Certainly, that's reasonable for 0M. But will it generally be the case so that this is the pattern we want?

src/parameterized_tendencies/microphysics/microphysics_wrappers.jl
```diff
@@ -90,107 +88,84 @@ end
...
-NamedTuple with `dq_tot_dt` and `e_tot_hlpr`.
-"""
-@inline function microphysics_tendencies_quadrature_0m(
-    ::GridMeanSGS, cm_params, thermo_params, ρ, T, q_tot,
-    q_liq, q_ice, Φ,
+@inline function microphysics_tendencies_0m(
+    cmp, thp, ρ, T, q_tot_nonneg, q_liq, q_ice, Φ, tst, dt,
 )
-    q_vap_sat = TD.q_vap_saturation(thermo_params, T, ρ)
+    # Computes satuartion vapor pressure, total water sink and energy helper
```
tapios:
```suggestion
    # Computes saturation vapor pressure, total water sink, and energy helper
```

src/parameterized_tendencies/microphysics/microphysics_wrappers.jl
```diff
@@ -233,84 +207,85 @@ end
...
 ```math
 \\bar{S} = \\int\\int S(T, q, \\dots) P(T, q) \\, dT \\, dq
 ```
+The option without SGS-quadratures can be used in the EDMF updrafts, or to compute
+grid-mean tendency without taking account of fluctuations.
 
-This enables microphysics tendencies to account for subgrid-scale fluctuations
-in temperature and moisture, capturing threshold effects at cloud edges and
-improving representation of phase changes.
+When running with explicit microphysics timestepping the total water sink is limited
```
tapios:
```suggestion
When running with explicit microphysics timestepping, the total water sink is limited
```

test/parameterized_tendencies/microphysics/sgs_quadrature.jl
```diff
@@ -740,17 +773,184 @@ using ClimaAtmos
...
+                    mp_0m, thp, ρ, T, q_tot, q_liq, q_ice, Φ, tst, dt,
+                )
+            end
+            t_wrapper_0m = @elapsed for _ in 1:N_bench
+                microphysics_tendencies_0m(
+                    mp_0m, thp, ρ, T, q_tot, q_liq, q_ice, Φ, tst, dt,
+                )
+            end
+
+            # 1-point, 3-point, 5-point quadrature 0M
```
tapios: This is not the number of points but the order. Points = Order^2.
```suggestion
            # 1st-order, 3rd order, and 5th order quadrature 0M
```

tapios (approved):
Looks good. 

The timing results are strange: Why is the version with the wrapper faster than the direct version? And the quadrature timings scale sublinearly (flops ~ order^2).  Do you know why?

szy21 (approved):
Thanks! I saw that you changed q_tot to q_tot_nonneg in the limiters in some places. I was wondering which one we should use. But assuming the tendency is always 0 when q_tot is negative, it probably doesn't matter?

src/parameterized_tendencies/microphysics/microphysics_wrappers.jl
trontrytel: If our implicit/explicit limiters are simple I was thinking to just add

```
if tst isa Explicit
  ...
elseif tst isa Implicit
 ...
end
```

If we end up needing more, I would then bring back the dispatching over different limiters.

PR #4416, opened 2026-04-06
Relax updraft buoy limit

sajjadazimi:
## Purpose 
Remove clipping buoyancy to limit excessive sgs mass flux through cloud top

## To-do
Fix detr

## Content

----

### DIFF
--- a/src/prognostic_equations/edmfx_entr_detr.jl
+++ b/src/prognostic_equations/edmfx_entr_detr.jl
@@ -415,6 +415,7 @@ function detrainment(
         CAP.detr_massflux_vertdiv_coeff(turbconv_params)
     max_area_limiter_scale = CAP.max_area_limiter_scale(turbconv_params)
     max_area_limiter_power = CAP.max_area_limiter_power(turbconv_params)
+    a_min = CAP.min_area(turbconv_params)
     a_max = CAP.max_area(turbconv_params)
     limit_inv_tau = CAP.entr_detr_limit_inv_tau(turbconv_params)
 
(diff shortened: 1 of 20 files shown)


toml/prognostic_edmfx_1M.toml
```diff
@@ -29,7 +29,7 @@ value = 0.0001
 value = 0
 
 [detr_buoy_coeff]
-value = 0
+value = 1
```
costachris, line 33: should the `detr_massflux_vertdiv_coeff` also be set to 0 now with the new detrainment formulation? or do we generally want the contributions from the massflux divergence term as well.

src/prognostic_equations/edmfx_entr_detr.jl
```diff
@@ -424,14 +425,18 @@ function detrainment(
         return limit_inv_tau
     end
 
+    # clip buoyancy time scale to 100s to avoid too fast detrainment when velocity is small
+    buoy_time_scale =
```
costachris, line 429: Can you just write this as an inverse timescale `inv_buoy_time_scale` or similar and multiply

toml/prognostic_edmfx_1M.toml
sajjadazimi: I think the buoyancy detrainment is only active above the boundary layer top (where buoyancy is negative). So we need another detrainment mechanism inside the boundary layer (with max close to the top). Otherwise the area fraction keeps growing in the highest layer of the boundary layer.

src/prognostic_equations/edmfx_entr_detr.jl
sajjadazimi: I'll do it in my next PR where I'm fixing detr.

szy21 (approved):

PR #4425, opened 2026-04-10
entr/detr area limiters

sajjadazimi:
## Purpose 
- Update entraiment and detrainment;
- Change area limiters functions from exp to power laws

## Content
In this PR, I introduce a function that returns a limiter factor which is close to one in the interior of the admissible range and smoothly approaches zero near the area fraction bounds (`a_min` and `a_max`). This factor is applied as
```math
detr = factor * detr + (1 - factor) * entr
```

so that the net area tendency `(entr - detr) * rho a` is effectively scaled by factor. As the area fraction approaches the bounds, the limiter gradually adjusts detrainment toward entrainment, causing `(entr - detr)` to approach zero and preventing the area fraction from leaving the allowed range.
In addition, I changed the area limiter functions (for both the minimum and maximum bounds) from exponential forms to power laws. Exponential functions decay very slowly, which causes the limiter to influence the entire range of area fractions. Power laws decay much more rapidly and therefore localize the limiter’s effect closer to the bounds. This choice is also more natural since the area fraction spans several orders of magnitude, and expressing the limiter in terms of ratios leads naturally to power-law behavior.
Finally, I made a few minor adjustments across the entrainment and detrainment parameterizations.

----

### DIFF
--- a/src/prognostic_equations/edmfx_entr_detr.jl
+++ b/src/prognostic_equations/edmfx_entr_detr.jl
@@ -62,6 +62,98 @@ function calculate_pi_groups(
 end
 
 """
+    detr_area_limiter_factor(entr::FT, detr, a, turbconv_params) where {FT}
+
+Return a multiplicative limiter for the net area tendency based on the current
+subdomain area `a`.
+
+If `entr > detr`, the net tendency would increase area, so the limiter damps
+growth as `a` approaches the upper bound `a_max`. In that case the returned
+factor is
+
+    max(0, 1 - a / a_max)^max_area_limiter_power.
+
+If `entr < detr`, the net tendency would decrease area, so the limiter damps
+shrinkage as `a` approaches the lower bound `a_min`. In that case the returned
+factor is
+
+    max(0, 1 - a_min / a)^min_area_limiter_power.
+
+Thus the limiter is close to one away from the active bound and smoothly
+approaches zero at the corresponding bound. The lower and upper bounds are
+obtained from `turbconv_params` through `CAP.min_area` and `CAP.max_area`.
+"""
+@inline function detr_area_limiter_factor(entr::FT, detr, a, turbconv_params) where {FT}
+    a_min = CAP.min_area(turbconv_params)
+    a_max = CAP.max_area(turbconv_params)
+    min_area_limiter_power = CAP.min_area_limiter_power(turbconv_params)
+    max_area_limiter_power = CAP.max_area_limiter_power(turbconv_params)
+    return ifelse(
+        entr > detr,
+        (max(0, (1 - a / a_max)))^max_area_limiter_power,
+        (max(0, (1 - a_min / a)))^min_area_limiter_power,
+    )
+end
+
+"""
+    compute_entrainment(...)
+
+Wrapper around `entrainment` that handles numerical safeguards and shared
+preprocessing before calling the model-specific entrainment formulation.
+
+If the updraft area (or associated mass) becomes numerically negligible,
+entrainment is set to a limiting value to maintain mixing with the
+environment and avoid ill-defined entrainment rates. Otherwise the
+entrainment rate is computed using the selected `entr_model`.
+"""
+function compute_entrainment(
+    thermo_params,
+    turbconv_params,
+    ᶜz,
+    z_sfc,
+    ᶜp,
+    ᶜρ,
+    ᶜaʲ,
+    ᶜwʲ,
+    ᶜRHʲ,
+    ᶜbuoyʲ,
+    ᶜw⁰,
+    ᶜRH⁰,
+    ᶜbuoy⁰,
+    ᶜtke,
+    entr_model,
+)
+
+    FT = eltype(thermo_params)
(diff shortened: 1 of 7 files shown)


src/prognostic_equations/edmfx_entr_detr.jl
```diff
@@ -415,32 +562,28 @@ function detrainment(
...
-        return limit_inv_tau
-    end
-
-    # clip buoyancy time scale to 100s to avoid too fast detrainment when velocity is small
-    buoy_time_scale =
-        max(FT(100), abs(ᶜwʲ - ᶜw⁰) / (eps(FT) + abs(min(ᶜbuoyʲ - ᶜbuoy⁰, 0))))
+    # Clip buoyancy time scale to 100s (inverse time scale to 0.01 1/s) to avoid 
+    # too fast detrainment when velocity is small
+    buoy_inv_time_scale =
+        min(FT(0.01), abs(min(ᶜbuoyʲ - ᶜbuoy⁰, 0)) / (eps(FT) + abs(ᶜwʲ - ᶜw⁰)))
```
costachris: Can you make the 0.01 a parameter (or a TODO to make a parameter). It could get hidden if the user changes the timestep

src/prognostic_equations/edmfx_entr_detr.jl
```diff
@@ -61,6 +61,98 @@ function calculate_pi_groups(
...
+shrinkage as `a` approaches the lower bound `a_min`. In that case the returned
+factor is
+
+    max(0, 1 - a_min / a)^min_area_limiter_power.
+
+Thus the limiter is close to one away from the active bound and smoothly
+approaches zero at the corresponding bound. The lower and upper bounds are
+obtained from `turbconv_params` through `CAP.min_area` and `CAP.max_area`.
+"""
+@inline function area_limiter_factor(entr::FT, detr, a, turbconv_params) where {FT}
```
costachris: Can you call it `detr_area_limiter_factor` or similar since this is being effectively used to modify the detrainment?

src/prognostic_equations/edmfx_entr_detr.jl
```diff
@@ -239,7 +319,82 @@ function entrainment(
 
     entr =
         area_limiter_factor * (entr_inv_tau + vel_diff_term + min_area_limiter)
-    return max(entr, entr_inv_tau)
+    return entr
```
costachris: Do we still want a positivity clipping for entr/detr?

costachris (approved):

src/prognostic_equations/edmfx_entr_detr.jl
```diff
@@ -378,10 +526,9 @@ function detrainment(
         entr_param_vec[12]
 
     # Detrainment proportional to negative mass flux divergence
-    detr_factor_mass_flux_div = -min(ᶜmassflux_vert_div, FT(0)) / ᶜρaʲ
-    detr = detr_factor_mass_flux_div * pi_sum_detr
-
-    return max(detr, detr_inv_tau)
+    detr_factor_mass_flux_div = -min(ᶜmassflux_vert_div, FT(0)) / (eps(FT) + ᶜρaʲ)
```
szy21:
```suggestion
    detr_factor_mass_flux_div = -min(ᶜmassflux_vert_div, FT(0)) / max(eps(FT), ᶜρaʲ)
```
(I think this is safer as ᶜρaʲ may be smaller than eps(FT))

src/prognostic_equations/edmfx_entr_detr.jl
szy21: Is there a reason why eps(FT) + x is better than max(eps(FT), x) here and in other places?

src/prognostic_equations/edmfx_entr_detr.jl
```diff
@@ -415,32 +562,28 @@ function detrainment(
...
+        max_area_limiter_scale * (max(0, ᶜaʲ - a_max) / (1 - a_max))^max_area_limiter_power
+
     detr =
         detr_inv_tau +
         detr_coeff * abs(ᶜwʲ) +
-        detr_buoy_coeff * max(0, 1 - a_min / ᶜaʲ) / buoy_time_scale -
+        detr_buoy_coeff * buoy_inv_time_scale -
         detr_vertdiv_coeff * min(ᶜvert_div, 0) -
-        detr_massflux_vertdiv_coeff * min(ᶜmassflux_vert_div, 0) / ᶜρaʲ + # Protected by ᶜaʲ check above
+        detr_massflux_vertdiv_coeff * min(ᶜmassflux_vert_div, 0) / (eps(FT) + ᶜρaʲ) +
```
szy21:
```suggestion
        detr_massflux_vertdiv_coeff * min(ᶜmassflux_vert_div, 0) / (max(eps(FT), ᶜρaʲ)) +
```
(same as before)

szy21 (approved):
Thanks!

src/prognostic_equations/edmfx_entr_detr.jl
sajjadazimi: I don't think + is better. This form is how it was before. Maybe max is better as it won't modify the result for w ~ O(eps)

PR #4429, opened 2026-04-18
Time-averaged 1m microphysics tendencies

sajjadazimi:
## Purpose 
Use time-averaged microphysics tendencies for 1m microphysics (CMv0p35). These tendencies are computed by implicitly solving a linearized microphysics model during the time step with sub-stepping (see https://github.com/CliMA/CloudMicrophysics.jl/pull/711).

This fixes the broken TRMM 1-moment test.

----


config/default_configs/default_config.yml
```diff
@@ -251,6 +251,9 @@ sgs_distribution:
 quadrature_order:
   help: "Quadrature order for SGS integration (1-5). Higher orders are more accurate but slower. Default: 2"
   value: 3
+microphysics_n_substeps:
+  help: "Number of susbteps for time averaging bulk microphysics tendencies. Default: 5"
+  value: 1
```
trontrytel: Is the default 1 or 5?

config/model_configs/prognostic_edmfx_trmm_implicit_column.yml
```diff
@@ -20,6 +20,7 @@ edmfx_vertical_diffusion: true
 edmfx_filter: true
 prognostic_tke: true
 microphysics_model: "1M"
+microphysics_n_substeps: 3
```
trontrytel: Should we make 3 the default?

src/parameterized_tendencies/microphysics/microphysics_wrappers.jl
```diff
@@ -247,22 +245,17 @@ NamedTuple with microphysics source terms:
...
-    ρ, q_tot_nonneg, q_lcl, q_icl, q_rai, q_sno, T, cmp, thp, tst, dt,
+    ρ, q_tot_nonneg, q_lcl, q_icl, q_rai, q_sno, T, cmp, thp, dt, nsubs,
 )
-    local_tendency = BMT.bulk_microphysics_tendencies(
+    local_tendency = BMT.average_bulk_microphysics_tendencies(
         BMT.Microphysics1Moment(), cmp, thp, ρ, T,
-        q_tot_nonneg, q_lcl, q_icl, q_rai, q_sno,
+        q_tot_nonneg, q_lcl, q_icl, q_rai, q_sno, dt, nsubs,
     )
-    if tst isa Explicit
```
trontrytel, line 256: Are the limiters now applied inside the sub-timestepping?

src/parameterized_tendencies/microphysics/microphysics_wrappers.jl
```diff
@@ -273,18 +266,12 @@ end
...
     evaluator = Microphysics1MEvaluator(
         scheme, cmp, thp, ρ, T, q_lcl_nonneg, q_icl_nonneg,
-        q_rai_nonneg, q_sno_nonneg, args,
+        q_rai_nonneg, q_sno_nonneg, dt, args,
     )
     # Integrate over quadrature points using functor (GPU-safe, no closure)
     local_tendency = integrate_over_sgs(
         evaluator, sgs_quad, q_tot_nonneg, T, q′q′, T′T′, corr_Tq,
     )
-    if tst isa Explicit
```
trontrytel, line 282: Same question as above

trontrytel (approved):
So much code deletion! This is great. 

Couple more comments:
- Claude thinks there are a couple more places in the code that reference `microphysics_tendency_timestepping` (`implicit_tendency`, `remaining_tendency`, `precomputed_quantities`, `type_getters`, and `types`). Do we want to fully remove it and the implicit microphysics pathway? Right now we don't have anything in the Jacobian for it anyway?
- Claude also thinks that `NonEquilibriumMicrophysics1M` parametric type is over-parameterized because the `n_substeps` will always be an `Int`.

src/parameterized_tendencies/microphysics/microphysics_wrappers.jl
sajjadazimi: We won't need the limiters for 1m anymore as we use the averaged tendencies with linearized microphysics model in each substep that keeps the solution positive.

sajjadazimi:
> So much code deletion! This is great.
> 
> Couple more comments:
> 
> * Claude thinks there are a couple more places in the code that reference `microphysics_tendency_timestepping` (`implicit_tendency`, `remaining_tendency`, `precomputed_quantities`, `type_getters`, and `types`). Do we want to fully remove it and the implicit microphysics pathway? Right now we don't have anything in the Jacobian for it anyway?
> * Claude also thinks that `NonEquilibriumMicrophysics1M` parametric type is over-parameterized because the `n_substeps` will always be an `Int`.

The option to treat microphysics tendencies implicitly vs explicitly is still there, even though the sources are precomputed and assumed constant.
I’m inclined to keep it in the implicit step for now—it’s a fast process and this aligns better with the IMEX-ARK assumptions.
Also, since the averaged tendency is currently only implemented for 1M, I’d prefer not to remove the explicit pathway in this PR. Once other pieces are in place, we can revisit and likely clean it up.

PR #4432, opened 2026-04-22
Add Presets module

nefrathenrici:
This PR contains a few UI improvements:
- Add Presets module with convenience constructors for the model/simulation, such as `bomex`, `aquaplanet`, `diagnostic_edmf`
- Always use itime for script-based interface
- Set Float32 by default for script-based 

This PR also removes repl_scripts from the docs, this page seems unused.


src/presets.jl
```diff
@@ -0,0 +1,239 @@
...
+microphysics tendency timestepping.
+Keyword arguments are forwarded to [`AtmosModel`](@ref).
+"""
+function nonequil_moist(; kwargs...)
+    defaults = (;
+        microphysics_model = NonEquilibriumMicrophysics1M(),
+        microphysics_tendency_timestepping = Explicit(),
+    )
+    return AtmosModel(; defaults..., kwargs...)
+end
```
szy21, line 72: What determines what are added to the defaults here?

docs/src/repl_scripts.jl
szy21, line 1: Some of these may still be useful for new developers. Could we move what is useful to wiki? (We can do it after this PR).

src/simulation/AtmosSimulations.jl
```diff
@@ -200,7 +198,7 @@ function AtmosSimulation{FT}(;
     params::Parameters.ClimaAtmosParameters = ClimaAtmosParameters(FT),
     context::ClimaComms.AbstractCommsContext = ClimaComms.context(),
     grid::Grids.AbstractGrid = SphereGrid(FT; radius = CAP.planet_radius(params), context),
-    initial_condition = Setups.DecayingProfile(; perturb = true, params),
+    setup = Setups.DecayingProfile(; perturb = true, params),
```
szy21, line 210: I find it a bit confusing to call e.g. `DecayingProfile` a setup, as it is really just an initial condition. But I don't know if there is a better way so it's fine.

haakon-e:
What kind of feedback are you looking for?
Is this, nominally, ready to be merged, or still work in progress?

nefrathenrici:
> What kind of feedback are you looking for? Is this, nominally, ready to be merged, or still work in progress?

This is ready to be merged. I’m mainly looking for feedback on whether it improves the new user experience. The goal is to make it easier to construct complex simulations without requiring users to understand or specify all of the underlying options.

src/simulation/AtmosSimulations.jl
```diff
@@ -115,13 +115,12 @@ function setup_diagnostics_and_writers(
...
 
 """
-    AtmosSimulation(config::AtmosConfig)
+    AtmosSimulation(; kwargs...)
 
-Construct a simulation from a YAML-based configuration.
+Construct an atmospheric simulation with the default floating-point type `Float32`.
+Equivalent to `AtmosSimulation{Float32}(; kwargs...)`.
 """
-function AtmosSimulation(config::AtmosConfig)
```
haakon-e, line 122: This constructor seems to be removed entirely. If anyone uses it downstream, their code will break.

This would be a breaking release to atmos, but I don't see any updates to `Project.toml`. Could you clarify the intent?

src/simulation/AtmosSimulations.jl
```diff
@@ -213,7 +211,6 @@ function AtmosSimulation{FT}(;
         ),
     ),
     surface_setup = SurfaceConditions.DefaultExchangeCoefficients(),
-    itime = false,
```
haakon-e, line 216: removing this kwarg may similarly break downstream code. I'm fine with it personally, but I think it should be communicated to users somehow (e.g. through `NEWS.md` file)

src/simulation/AtmosSimulations.jl
```diff
@@ -115,13 +115,12 @@ function setup_diagnostics_and_writers(
...
+    AtmosSimulation(; kwargs...)
 
-Construct a simulation from a YAML-based configuration.
+Construct an atmospheric simulation with the default floating-point type `Float32`.
+Equivalent to `AtmosSimulation{Float32}(; kwargs...)`.
 """
-function AtmosSimulation(config::AtmosConfig)
-    return get_simulation(config)
-end
+AtmosSimulation(; kwargs...) = AtmosSimulation{Float32}(; kwargs...)
```
haakon-e, line 132: Could this (`Float32` being default) be flagged in the docs and/or `NEWS.md`

src/presets.jl
```diff
@@ -0,0 +1,221 @@
...
+(60 uniform levels, z_max = 3 km), [`Setups.Bomex`](@ref) setup,
+[`equil_moist`](@ref) physics, `dt = 10 s`, `t_end = 6 h`.
+
+No EDMF turbulence-convection scheme is enabled by default; pass
+`model = Presets.diagnostic_edmf(FT)` or `Presets.prognostic_edmf(FT)` to add one.
+Keyword arguments are forwarded to [`AtmosSimulation`](@ref).
+"""
+function bomex(::Type{FT} = Float32; kwargs...) where {FT}
+    # Bomex setup needs thermo_params at construction time, so resolve params up front.
+    params = get(kwargs, :params, ClimaAtmosParameters(FT))
```
haakon-e: minor nit: this call eagerly constructs `ClimaAtmosParameters` even if it provided to the method. Could consider writing a tenary or ifelse block to conditionally construct the default set.

src/presets.jl
```diff
@@ -0,0 +1,221 @@
...
+    defaults = (;
+        params,
```
haakon-e, line 217: minor nit: `params` is explicitly inserted here, but is also passed to the constructor on L218, if `params` is part of `kwargs`. In the end, the intended parameters are used to construct the model. A short comment or rewrite may be worth it.

src/presets.jl
```diff
@@ -0,0 +1,221 @@
...
+# ============================================================================
+# Simulation presets — return an AtmosSimulation
+# ============================================================================
+
+"""
+    aquaplanet([FT = Float32]; kwargs...)
+
+Aquaplanet simulation preset: global [`SphereGrid`](@ref) with
+[`equil_moist`](@ref) physics (0M microphysics, prescribed zonally-symmetric
+SST, idealized insolation) and a [`DecayingProfile`](@ref) initial condition.
```
haakon-e: `DecayingProfile` is the default arg for `AtmosSimulation`, but not mentioned here. Could be worth explicitly putting it in the `defaults` list below to avoid confusion about where it comes from, or explicitly saying ```uses the default `DecayingProfile` initial condition from AtmosSimulation```

src/presets.jl
```diff
@@ -0,0 +1,221 @@
...
+        cloud_model = GridScaleCloud(),
+        surface_model = PrescribedSST(),
+        sfc_temperature = ZonallySymmetricSST(),
+        insolation = IdealizedInsolation(),
+    )
+    return AtmosModel(; defaults..., kwargs...)
+end
+
+"""
+    nonequil_moist(; kwargs...)
```
haakon-e: we're moving away from talking about `moisture` model separately from `microphysics`/`precipitation` models. Better to call this `nonequil_moist1m` or similar.

This setup is also much less "complete" compared to `equil_moist`. Why is that?

I would suggest either making `equil` and `nonequil` overwrite the same kwargs (for symmetry), or document that this is a minimal call, meant to be combined with other kwargs.

haakon-e (approved):
Nice PR!

Providing users with friendlier ways to set up models consistently is a good step forward. Test coverage also seems useful.

Some minor comments attached, but fine moving forward as-is, after `NEWS.md` has been updated.

src/simulation/AtmosSimulations.jl
nefrathenrici: This constructor does not need to be removed. I would rather this PR not break the interface.

PR #4438, opened 2026-04-24
Amip fixers

sajjadazimi:
## Purpose 
- Enforce physical constraints as a callback: moisture partition, edmf decomposition
- Prevent area growth when updraft velocity is zero
- Add fixed terminal velocity option

----


src/cache/microphysics_cache.jl
```diff
@@ -160,12 +160,12 @@ function set_precipitation_velocities!(
     ᶜq_lcl⁰ = ᶜspecific_env_value(@name(q_lcl), Y, p)
     ᶜρa⁰χ⁰ = @. lazy(max(zero(Y.c.ρ), ᶜρa⁰) * max(zero(Y.c.ρ), ᶜq_lcl⁰))
     @. ᶜρχ = ᶜρa⁰χ⁰
-    @. ᶜwₗ = ᶜρa⁰χ⁰ * CMNe.terminal_velocity(
-        cmc.liquid,
-        cmc.stokes,
-        ᶜρ⁰,
-        ᶜq_lcl⁰,
-    )
+    @. ᶜwₗ = ᶜρa⁰χ⁰ * 0 #CMNe.terminal_velocity(
```
trontrytel: Could you make all four velocities be part of CloudMicrophysics or ClimaAtmos parameters? I think we will want to calibrate them. Also, it would be good to set cloud velocities to something small but non-zero. Cloud ice sedimentation is said to be important

src/cache/prognostic_edmf_precomputed_quantities.jl
```diff
@@ -255,7 +257,7 @@ NVTX.@annotate function set_prognostic_edmf_precomputed_quantities_explicit_clos
                 ᶜq_iceʲs.:($$j),
             ),
             vertical_buoyancy_acceleration(Y.c.ρ, ᶜρʲs.:($$j), ᶜgradᵥ_ᶠΦ, ᶜlg),
-            get_physical_w(ᶜu, ᶜlg),
```
trontrytel, line 258: Why do we always pass zero here? From our discussion I thought it was only supposed to be for insignificant updrafts?

src/prognostic_equations/mass_flux_closures.jl
```diff
@@ -214,29 +214,69 @@ function edmfx_vertical_diffusion_tendency!(
 end
 
 """
-    edmfx_filter_tendency!(Y, p, t, turbconv_model)
-
-Apply EDMF physical constraints: immediately mix the updraft with the environment if
-area fraction is non-positive
+    enforce_physical_constraints!(Y, p)
+
```
trontrytel, line 218: Do we want to add similar filtering for non-EDMF runs?

src/prognostic_equations/mass_flux_closures.jl
```diff
@@ -214,29 +214,69 @@ function edmfx_vertical_diffusion_tendency!(
...
+Enforce physical constraints on the model state `Y` in-place.
+
```
trontrytel, line 220: Could we add some tests and docs for this new filter?

src/prognostic_equations/mass_flux_closures.jl
```diff
@@ -214,29 +214,69 @@ function edmfx_vertical_diffusion_tendency!(
...
+This function is used as a callback and is not a tendency evaluation. It applies
```
trontrytel, line 221: We are not filtering on negative grid mean q_tot because of mass conservation constraints?

src/cache/prognostic_edmf_precomputed_quantities.jl
sajjadazimi: It seems that preventing area growth at the surface when buoy_flux is negative is enough to make AMIP stable. So I revert this change.

src/prognostic_equations/mass_flux_closures.jl
sajjadazimi: I didn't add it for other turn_conv options because they are already stable. But we can add it in future PRs (and maybe make it another non-negativity option) if we see that the results for EDMF are satisfactory.

src/prognostic_equations/mass_flux_closures.jl
sajjadazimi: Yes, the only option for q_tot that respects mass conservation is water borrowing. We can choose to apply water borrowing here instead of applying it after each dss. But I leave it to future PRs because AMIP was stable without fixing negative q_tot.

trontrytel (approved):

PR #4440, opened 2026-04-27
Consolidate Jacobian interface in AtmosSimulation

nefrathenrici:
This PR slightly simplifies the Jacobian interface. 

Content:
- replace four Jacobian-related AtmosSimulation kwargs with a single JacobianAlgorithm
- move the seven DerivativeFlags off ManualSparseJacobian into the cache where they're derived from AtmosModel at build time.

### DIFF
--- a/src/simulation/AtmosSimulations.jl
+++ b/src/simulation/AtmosSimulations.jl
@@ -177,10 +177,9 @@ Construct an atmospheric simulation with floating-point type `FT` (default: Floa
 
 ### Numerics
 - `ode_config`: ODE solver algorithm. Default: `IMEXAlgorithm(ARS343(), NewtonsMethod(...))`.
-- `use_dense_jacobian::Bool = false`: Use a dense Jacobian matrix.
-- `use_auto_jacobian::Bool = false`: Use automatic differentiation for the Jacobian.
-- `approximate_linear_solve_iters::Int = 1`: Number of approximate linear solve iterations.
-- `auto_jacobian_padding_bands::Int = 0`: Extra bandwidth for auto-differentiated Jacobian.
+- `jacobian::JacobianAlgorithm = ManualSparseJacobian(; approximate_solve_iters = 1)`:
+  Jacobian algorithm for the implicit solve. Use [`ManualSparseJacobian`](@ref),
+  [`AutoSparseJacobian`](@ref), or [`AutoDenseJacobian`](@ref).
 - `debug_jacobian::Bool = false`: Enable Jacobian debugging output.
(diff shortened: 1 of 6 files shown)


imreddyTeja:
@coderabbitai review

src/simulation/AtmosSimulations.jl
```diff
@@ -234,11 +233,8 @@ function AtmosSimulation{FT}(;
     default_diagnostics = true, # Enable standard ClimaAtmos diagnostics
     diagnostics = (),           # User-provided diagnostics (YAML string format or ScheduledDiagnostics )
     # Numerics
-    use_dense_jacobian = false,
-    use_auto_jacobian = false,
-    approximate_linear_solve_iters = 1,
-    auto_jacobian_padding_bands = 0,
-    debug_jacobian = false,
+    jacobian::JacobianAlgorithm = ManualSparseJacobian(approximate_solve_iters = 1),
```
imreddyTeja, line 236: Is this change to the kwargs a breaking change?

src/solver/type_getters.jl
```diff
@@ -524,46 +524,42 @@ function _config_surface_setup(parsed_args)
...
-            DerivativeFlag(atmos.sgs_entr_detr_mode),
-            DerivativeFlag(atmos.sgs_mf_mode),
-            DerivativeFlag(atmos.sgs_nh_pressure_mode),
-            DerivativeFlag(atmos.sgs_vertdiff_mode),
-            approximate_linear_solve_iters,
+    @info "Jacobian algorithm: $(summary_string(jacobian))"
+    jac = Jacobian(jacobian, Y, atmos; verbose = debug_jacobian)
+    if hasproperty(jac.cache, :derivative_flags)
+        flags_str = join(
+            ("$k = $(typeof(v).name.name)" for (k, v) in pairs(jac.cache.derivative_flags)),
```
imreddyTeja, line 557: What does `typeof(v).name.name` do?

imreddyTeja (approved):
This looks good to me. Sorry about to long review wait

src/solver/type_getters.jl
haakon-e: it gets the type of some object as a `Symbol` without the module prefix or type parameters. It's kind-of accessing internals, but reasonably common in printing contexts, like here.

haakon-e (approved):

src/simulation/AtmosSimulations.jl
nefrathenrici: Technically yes, but this interface isn't in use downstream so it should be fine.

PR #4450, opened 2026-04-29
unify prognostic edmf parameters in 0M and 1M

szy21:
## Purpose 

There are some case specific edmf tomls, which are used for testing pigroup etc. I didn't change them. I will clean them up together with the tests in the future.

## Content

----

### DIFF
--- a/reproducibility_tests/ref_counter.jl
+++ b/reproducibility_tests/ref_counter.jl
@@ -1,4 +1,4 @@
-338
+339
 
 # **README**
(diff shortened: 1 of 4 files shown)


trontrytel (approved):
Thank you!

Yes, it would be great to end up with the same toml files for single column test cases and out coupler long runs. At least for the edmf + microphysics combo.

sajjadazimi (approved):

PR #4451, opened 2026-04-29
Fix double header for default config

jefftyoder:
Quick typo fix.


imreddyTeja (approved):
Good find, thanks!

szy21:
Manually merging this as there is no conflicts.

PR #4453, opened 2026-04-30
set sgs 1m microphysics tracers when overwrite initial conditions

szy21:
## Content

----


costachris (approved):
LGTM

PR #4456, opened 2026-05-01
Add CLAUDE.md that points to AGENTS.md

imreddyTeja:
Claude does not use AGENTS.md, unlike the other
major agent players.

## To-do
- I think this works, but maybe I should test it further

## Content

----

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,1 @@
+@AGENTS.md


ph-kev (approved):
I am not too familiar with this, but this looks like it would work. Maybe someone with Claude can test this?

Also, I was skimming through the docs (https://code.claude.com/docs/en/memory#agents-md) which recommends a different solution.

PR #4457, opened 2026-05-01
use max_newton_iters_ode 1

szy21:
## Purpose 

remove max_newton_iters_ode in most configs, which is default to 1

## Content

----


szy21:
manually merging this so that weekend longrun can run with this.

PR #4459, opened 2026-05-01
Add auto-retry for "lost agent" buildkite gpu steps

imreddyTeja:
## Purpose 
I find the "agent lost (-1)" to be annoying with the gpu steps. This changes those steps to auto retry up to 2 times when that happens.

Let me know if there are other pipelines you would like this feature in.
## To-do
- verify this actually works

## Content

----

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -386,6 +386,9 @@ steps:
           - "gpu_plane_no_topography_float64_test/output_active/*.dat"
         env:
           CLIMACOMMS_DEVICE: "CUDA"
+        retry:
+          automatic:
+            - exit_status: -1
         agents:
           slurm_gpus: 1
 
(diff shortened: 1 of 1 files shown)


.buildkite/pipeline.yml
```diff
@@ -386,6 +386,10 @@ steps:
           - "gpu_plane_no_topography_float64_test/output_active/*.dat"
         env:
           CLIMACOMMS_DEVICE: "CUDA"
+        retry:
+          automatic:
+            - exit_status: -1
+            - limit: 2
```
ph-kev: The default for the limit is 2, so I don't think it is necessary to include this, unless you want to be transparent.

ph-kev (approved):

.buildkite/pipeline.yml
imreddyTeja: good point. Do you think 2 is a reasonable value to keep, or should I change this to something else?

ph-kev: I think 1 or 2 is a reasonable value for this.

imreddyTeja:
Closed as duplicate to #4458

PR #4460, opened 2026-05-02
Add fixers to diagnostic EDMF 1M

trontrytel:
As discussed - we wanted to check if adding similar clipping of negative microphysics tracers would help with diagnostic EDMF 1M stability.

### DIFF
--- a/src/prognostic_equations/mass_flux_closures.jl
+++ b/src/prognostic_equations/mass_flux_closures.jl
@@ -235,92 +235,118 @@ area fractions, negative condensate masses, or condensate mass exceeding the
 available total moisture. Ideally, the need for this correction is minimized 
 by the numerical scheme.
 """
-enforce_physical_constraints!(Y, p, t, turbconv_model) = nothing
 
-function enforce_physical_constraints!(Y, p, t, turbconv_model::PrognosticEDMFX)
+# Private helper: clips grid-mean condensate tracers to non-negative values and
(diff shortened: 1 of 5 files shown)


src/prognostic_equations/mass_flux_closures.jl
```diff
@@ -324,3 +324,44 @@ function enforce_physical_constraints!(Y, p, t, turbconv_model::PrognosticEDMFX)
...
+    enforce_physical_constraints!(Y, p, t, turbconv_model::DiagnosticEDMFX)
+
+For `DiagnosticEDMFX`, only the grid-mean microphysics constraints are enforced:
+updraft fields (`sgsʲs`, `u₃`) are not modified because they are not prognostic
+in diagnostic EDMF — they are recomputed each step from the vertical integral.
+
+Gated on `p.atmos.edmfx_model.filter isa Val{true}` (config key `edmfx_filter`),
+the same flag used for `PrognosticEDMFX`.
+"""
+function enforce_physical_constraints!(Y, p, t, turbconv_model::DiagnosticEDMFX)
```
szy21: I think you can use the same function with PrognosticEDMFX by changing `n_mass_flux_subdomains` to `n_prognostic_mass_flux_subdomains` in that function (or you can add a check for PrognosticEDMFX). It may be clearer to unify them.

szy21 (approved):
Note that edmf_filter is false for all the diagnostic edmf simulations, so you may want to change them.

PR #4464, opened 2026-05-05
remove type conversion from `0` comparison

simone-silvestri:
## Purpose 

remove a type conversion from `0` equality in orographic wave parameterization. 
The comparison is equivalent even without converting to FT. The `FT` conversion is actually omitted in llvm:

```julia
julia> test(x) = x == 0
test (generic function with 1 method)

julia> test2(x) = x == eltype(x)(0)
test2 (generic function with 1 method)

julia> @code_llvm test(0.0)
; Function Signature: test(Float64)
;  @ REPL[1]:1 within `test`
define i8 @julia_test_1357(double %"x::Float64") #0 {
top:
; ┌ @ float.jl:661 within `==` @ float.jl:621
   %0 = fcmp oeq double %"x::Float64", 0.000000e+00
; │ @ float.jl:661 within `==`
; │┌ @ bool.jl:40 within `&`
    %1 = zext i1 %0 to i8
    ret i8 %1
; └└
}

julia> @code_llvm test2(0.0)
; Function Signature: test2(Float64)
;  @ REPL[2]:1 within `test2`
define i8 @julia_test2_1390(double %"x::Float64") #0 {
top:
; ┌ @ float.jl:621 within `==`
   %0 = fcmp oeq double %"x::Float64", 0.000000e+00
   %1 = zext i1 %0 to i8
   ret i8 %1
; └
}
```

I think the best way to do it should be to use `iszero` but I am not sure how it will play with the ClimaCore machinery.

### DIFF
--- a/src/parameterized_tendencies/gravity_wave_drag/orographic_gravity_wave.jl
+++ b/src/parameterized_tendencies/gravity_wave_drag/orographic_gravity_wave.jl
@@ -493,7 +493,7 @@ function calc_nonpropagating_forcing!(
     # Exclude cells with zero weights from the mask to avoid division by zero.
     # Zero weight means p == p_ref at that cell, so it contributes nothing
     # to the pressure-weighted average.
-    @. ᶜmask = ᶜmask && (ᶜweights != FT(0))
+    @. ᶜmask = ᶜmask && (ᶜweights != 0)
 
     parent(ᶜweights) .= parent(ᶜweights .* ᶜmask)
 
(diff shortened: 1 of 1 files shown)


imreddyTeja:
I've found in more complex expressions that include climacore fields and operators, the FT in FT(0) is not always removed by the compiler. Did removing these two uses of FT fix the isbits error? Climacore should play nicely with iszero

simone-silvestri:
I haven't tried actually. I can try now. Probably then `iszero` is the best option here.

imreddyTeja (approved):

imreddyTeja:
Tested with 1.10.10. Can confirm that switching to `iszero` fixes the `isbits` error

PR #4465, opened 2026-05-05
Stabilize snow-induced fluctuations

sajjadazimi:
## Purpose 
- Use two substeps for computing grid-mean microphysics tendencies to avoid fluctuations. These fluctuations can lead to instabilities in AMIP simulations.
- Pass `w^0 = 0` to entrainment and detrainment to avoid artificial area fraction growth when `w^j = 0`

----

### DIFF
--- a/reproducibility_tests/ref_counter.jl
+++ b/reproducibility_tests/ref_counter.jl
@@ -1,4 +1,4 @@
-340
+341
 
 # **README**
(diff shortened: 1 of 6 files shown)


szy21 (approved):

trontrytel (approved):
Thank you!

Issue #4461, opened 2026-05-04
Clean up parsed_args


nefrathenrici:
Currently, parsed_args are spread throughout the codebase, entangling configuration with runtime concerns. 
Now that we have a programmatic constructor, we should unify the two systems and parsed_args should only be used to construct a simulation.

Plan to clean up parsed_args:
1. Make programmatic `AtmosSimulation{FT}(...)` capable of matching all the YAML arguments.
2. Route the parsed_args/YAML constructor into the programmatic path, building the AtmosSimulation kwargs from the parsed_args. This is mostly handled by the existing parsing: `get_grid`, `get_setup`, but these functions may be standardized further.
3. Clean up all other parsed_args usage outside of the core parsing needed in `src/solver`

PR #4469, opened 2026-05-06
Make AtmosSimulation match YAML config

nefrathenrici:
This PR updates the YAML and programmatic interfaces to have the same functionality, and unifies the way they set up callbacks and diagnostics. This is the first step outlined in #4461 

Content
- Diagnostics are now specified via a `DiagnosticsConfig` struct. The new struct accepts many input shapes for diagnostics: `"ts" => "1hours"` (shortest form), `"ua" => (; period, reduction)` (pair with options), `(; short_name = "va", period = "1hours", writer = "netcdf")` (NamedTuple) raw Dicts, or pre-built ScheduledDiagnostic objects. Mixed lists allowed.
- Set one path to set up diagnostics via `setup_diagnostics_and_writers` -> `scheduled_diagnostics_from_specs`
- Changed the YAML config to use `common_callbacks/default_model_callbacks` functions that the programmatic interface uses
- Deleted YAML keys: `use_itime` (itime is now always used), `netcdf_horizontal_method` (hardcoded to default bilinear)

Todo: make the diagnostic specification clearer - reduction is not clear now.

### DIFF
--- a/src/callbacks/get_callbacks.jl
+++ b/src/callbacks/get_callbacks.jl
@@ -1,107 +1,46 @@
-function get_diagnostics(
-    parsed_args,
-    atmos_model,
-    Y, p, dt,
-    t_start, start_date, output_dir,
+# Reduction-time keys allowed in diagnostic spec dicts.
+# Lowercased so callers can use "Max" or "max" interchangeably.
+const _DIAG_ALLOWED_REDUCTIONS = Dict(
+    "inst" => (nothing, nothing),       # just dump the variable
+    "nothing" => (nothing, nothing),    # also accepts the literal string "nothing"
+    "max" => (max, nothing),
+    "min" => (min, nothing),
+    "average" => ((+), CAD.average_pre_output_hook!),
 )
 
+"""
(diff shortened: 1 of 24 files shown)


src/callbacks/get_callbacks.jl
```diff
@@ -666,50 +466,66 @@ function default_model_callbacks(radiation::AtmosRadiation;
...
-        checkpoint_frequency;
+        checkpoint_frequency,
     )
 end
 
-# Gravity wave component callbacks
+# Gravity-wave component callbacks (both orographic and non-orographic)
 function default_model_callbacks(gravity_wave::AtmosGravityWave;
     dt_nogw = "3hours",
+    dt_ogw = "6hours",
```
szy21, line 476:
```suggestion
    dt_ogw = "3hours",
```
(it doesn't really matter much, but I think it makes more sense to make the default of dt_nogw and dt_ogw to be the same)

haakon-e (approved):

szy21 (approved):

PR #4472, opened 2026-05-07
add snow water path diagnostics

szy21:
## Purpose 

Also adds rwp and swp to the default diagnostics.

## Content

----


src/diagnostics/core_diagnostics.jl
```diff
@@ -1012,14 +1012,40 @@ end
 
 add_diagnostic_variable!(short_name = "rwp", units = "kg m^-2",
     long_name = "Rainwater Path",
-    standard_name = "atmosphere_mass_content_of_rainwater",
```
szy21, line 1015: removed because there seem to be no standard name for this

trontrytel (approved):

PR #4473, opened 2026-05-08
Add condensate SGS variability

sajjadazimi:
## Purpose 
Include SGS condensate variability in microphysics tendencies using equilibrium-based scaling

----

### DIFF
--- a/src/parameterized_tendencies/microphysics/microphysics_wrappers.jl
+++ b/src/parameterized_tendencies/microphysics/microphysics_wrappers.jl
@@ -159,12 +159,61 @@ end
 ###
 
 """
+    equilibrium_cloud_condensates_1m(
+        tps,
+        T,
+        ρ,
+        q_tot,
+        q_lcl_mean,
+        q_icl_mean,
+        q_rai,
+        q_sno,
+    )
+
+Compute the equilibrium cloud liquid and cloud ice specific humidities
+(`q_lcl_eq`, `q_icl_eq`) for a thermodynamic state `(T, ρ, q_tot)`
+in the one-moment microphysics scheme.
+
+The supersaturation excess is partitioned into liquid and ice using the
+liquid fraction diagnosed from the total liquid and ice condensates,
+including rain (`q_rai`) and snow (`q_sno`) contributions. Rain and snow
+condensates are then removed from the equilibrium partition to obtain the
+equilibrium cloud condensates.
+
+Returns `(q_lcl_eq, q_icl_eq)`.
+"""
+@inline function equilibrium_cloud_condensates_1m(
+    tps,
+    T,
+    ρ,
+    q_tot,
+    q_lcl_mean,
+    q_icl_mean,
+    q_rai,
+    q_sno,
+)
+    FT = typeof(ρ)
+
+    # Saturation excess
+    q_sat = TD.q_vap_saturation(tps, T, ρ)
+    excess = max(FT(0), q_tot - q_sat)
+
+    # Partition excess condensate
+    λ = TD.liquid_fraction(tps, T, q_lcl_mean + q_rai, q_icl_mean + q_sno)
+
+    q_lcl_eq = max(FT(0), λ * excess - q_rai)
+    q_icl_eq = max(FT(0), (FT(1) - λ) * excess - q_sno)
+
+    return q_lcl_eq, q_icl_eq
+end
+
+"""
     (eval::Microphysics1MEvaluator)(T_hat, q_tot_hat)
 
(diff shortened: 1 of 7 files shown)


src/parameterized_tendencies/microphysics/microphysics_wrappers.jl
```diff
@@ -185,33 +209,51 @@ struct Microphysics1MEvaluator{S, MP, TPS, FT, Args <: Tuple}
...
+
+    # q_mean ≥ α * q_mean_eq and q_eq_hat ≥ 0, so q_hat must be positive,
+    # and we don't need to clip
+    q_lcl_hat = eval.q_lcl_mean + eval.α_lcl * (q_lcl_eq_hat - eval.q_lcl_mean_eq)
+    q_icl_hat = eval.q_icl_mean + eval.α_icl * (q_icl_eq_hat - eval.q_icl_mean_eq)
 
     # Call CloudMicrophysics point-wise tendencies
-    # Average tendencies over dt with nsubs = 1 because with sgs quadrature the resulting
-    # tendencies are smoother and one substep for time-averaging suffices
+    # Average tendencies over dt with nsubs = 2; with sgs quadrature the resulting
```
tapios: nsubs should be a configurable parameter.

src/parameterized_tendencies/microphysics/microphysics_wrappers.jl
```diff
@@ -185,33 +209,51 @@ struct Microphysics1MEvaluator{S, MP, TPS, FT, Args <: Tuple}
...
+    # Compute saturation at quadrature point
+    q_sat_hat = TD.q_vap_saturation(eval.tps, T_hat, eval.ρ)
+
+    # Saturation excess at the quadrature point
+    excess_hat = max(FT(0), q_tot_hat - q_sat_hat)
+
+    # Partition using grid-mean liquid fraction
+    λ = TD.liquid_fraction(eval.tps, T_hat, eval.q_lcl_mean, eval.q_icl_mean)
+    q_lcl_eq_hat = max(0, λ * excess_hat - eval.q_rai)
+    q_icl_eq_hat = max(0, (FT(1) - λ) * excess_hat - eval.q_sno)
```
tapios: Since this logic is repeated, may be a good idea to extract it to helper function.

src/parameterized_tendencies/microphysics/microphysics_wrappers.jl
```diff
@@ -255,18 +297,29 @@ NamedTuple with microphysics source terms:
...
+    q_lcl_nonneg = max(0, q_lcl_mean)
+    q_icl_nonneg = max(0, q_icl_mean)
     q_rai_nonneg = max(0, q_rai)
     q_sno_nonneg = max(0, q_sno)
 
+    # Pre-compute grid-mean condensate and saturation for perturbation model
+    q_excess_mean = max(0, q_tot_nonneg - TD.q_vap_saturation(thp, T, ρ))
+    λ = TD.liquid_fraction(thp, T, q_lcl_mean, q_icl_mean)
+    q_lcl_mean_eq = max(0, λ * q_excess_mean - q_rai_nonneg)
+    q_icl_mean_eq = max(0, (1 - λ) * q_excess_mean - q_sno_nonneg)
```
tapios: As above, consider using a helper to avoid repetition.

tapios (approved):
LGTM

src/parameterized_tendencies/microphysics/microphysics_wrappers.jl
trontrytel: This will be harder to do for the P3 scheme, where we only have one category. But we can cross that bridge later

trontrytel (approved):

src/parameterized_tendencies/microphysics/microphysics_wrappers.jl
haakon-e: I think this is fine -- we'll have a separate 2M evaluator anyways.

szy21:
This PR seems to significantly affects clouds. Here is the TOA SW bias for a month, with the same parameters:
before:
Screenshot 2026-05-09 at 11 27 51 AM
after:
Screenshot 2026-05-09 at 11 27 19 AM

Would it be possible to have a ci case so we can know these changes faster? Maybe baroclinic wave with 1M? It is a bit hard for development if we only realize these changes after many hours of simulations.

tapios:
Maybe we should revert this and keep condensate at the mean. Probably the fluctuations in the saturated direction increase precipitation efficiency (autoconversion/accretion). Not having them in the first place may limit harm.

PR #4474, opened 2026-05-08
avoid dividing by zero ustar in edmf boundary conditions

szy21:
## Purpose 

We recently came across this problem here: https://github.com/CliMA/ClimaCoupler.jl/pull/1914. Even though ustar should be non zero, I think this is safer.

## Content

----

### DIFF
--- a/src/prognostic_equations/edmfx_boundary_condition.jl
+++ b/src/prognostic_equations/edmfx_boundary_condition.jl
@@ -122,7 +122,7 @@ function get_first_interior_variance(
     obukhov_length,
     local_geometry,
 ) where {FT}
-    c_star = -kinematic_scalar_flux / ustar
+    c_star = -kinematic_scalar_flux / max(ustar, eps(FT))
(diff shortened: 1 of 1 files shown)


costachris (approved):

haakon-e:
wouldn't it be more consistent to limit this value where it's set in `p.precomputed.sfc_conditions`?

szy21:
> wouldn't it be more consistent to limit this value where it's set in `p.precomputed.sfc_conditions`?

That is already non zero because the output of surface flux is non zero. However, when running from the coupler, p.precomputed.sfc_conditions is set in the coupler rather than in the atmos, where ustar is set to non zero. So technically we don't need this change, but I think it's safer in case ustar is set somewhere else.

In general though, I'm not sure if I want to set things like ustar and L_MO to non zero, except when it is used in the denominator. For example, if someone ever calculates momentum flux from ustar, we probably want momentum flux to be strictly zero when ustar is zero. But that's a different issue that we can think about later.

PR #4479, opened 2026-05-11
Make cloud formulation consistent with quadrature and resolution-adaptive

tapios:
## Purpose 

Unifies the cloud fraction and microphysics quadrature pipelines so they share the same SGS moments, eliminating physical inconsistencies between the two.

## Changes

* New `sgs_moments pre-pass` (`compute_sgs_moments` / `set_sgs_moments!`): A Gauss–Hermite quadrature pass to compute (`μ_S`, `σ_S²`, `M_l`, `M_i`): the mean and variance of the saturation variable (saturation excess in Gaussian case) plus the SGS-mean equilibrium condensates. Both cloud fraction and microphysics consume these cached moments, guaranteeing they operate on identical physics.

* Hybrid cloud fraction (`compute_cloud_fraction_hybrid`): Replaces the former Sommeria–Deardorff approximation with a smooth logistic-CDF formula evaluated on the quadrature-derived (`μ_S`, `σ_S²`). An activation factor `α` gates the cloud fraction to zero in clear states. Dispatches on `GaussianSGS`, `LogNormalSGS`, and `GridMeanSGS`.
* Shape-function partition in `Microphysics1MEvaluator` (similar to @sajjadazimi's approach): Distributes prognostic grid-mean condensate across quadrature points via `q_lcl_hat = q_lcl_mean · (γ_l · q_lcl_eq_hat + β_l)` where `γ_l · M_l + β_l = 1`, giving exact mass conservation by construction. The β-factor provides a smooth fallback when equilibrium moments vanish. The shape-function coefficients (`γ`, `β`) are precomputed once per grid cell.

## To-do (later)

* Symmetric subdomain treatment: give each EDMF subdomain its own cloud fraction and SGS quadrature, using environment-only covariances for the environment. Currently, grid-mean covariances and state are used for all subdomains, which can double-count updraft variance.
* Calibrate cf_steepness_scale and (and check sensitibity to `q_min`) on cloud fraction and precipitation in DYCOMS/TRMM.

## Content

----

### DIFF
--- a/src/parameterized_tendencies/microphysics/cloud_fraction.jl
+++ b/src/parameterized_tendencies/microphysics/cloud_fraction.jl
@@ -61,6 +61,13 @@ function set_covariance_cache_and_cloud_fraction!(Y, p)
 
     # One Picard step: use the current cloud fraction to update buoyancy
     # gradient and covariance cache, then recompute cloud fraction.
+    #
+    # The hybrid cloud fraction reads `(μ_S, σ_S²)` from a Gauss-Hermite
+    # pre-pass, but those moments are computed inline inside the CF broadcast
+    # (see `compute_cloud_fraction_hybrid`) — there is no separate
+    # `set_sgs_moments!` call during Picard. The microphysics-only moments
+    # `(M_l, M_i)` are written once after Picard converges, in
+    # `set_sgs_moments_mp!`.
     function picard_step!()
         @. ᶜlinear_buoygrad = buoyancy_gradients(
             BuoyGradMean(), # TODO: modify for NonEq + 1M tracers if needed
@@ -123,6 +130,12 @@ function set_covariance_cache_and_cloud_fraction!(Y, p)
     )
     set_covariance_cache!(Y, p, thermo_params)
 
+    # Write the microphysics moments `(M_l, M_i)` once against the post-Aitken
+    # variance closure. The downstream microphysics tendency reads
+    # `ᶜsgs_moments_mp` for the shape-function partition. No-op when that
+    # cache is not allocated (e.g. 0M / dry).
+    set_sgs_moments_mp!(Y, p)
+
     return nothing
 end
 
@@ -166,14 +179,10 @@ function compute_∂T_∂θ!(dest, Y, p, thermo_params)
     ᶜρ = Y.c.ρ
     if p.atmos.microphysics_model isa Union{DryModel, EquilibriumMicrophysics0M}
         (; ᶜq_liq, ᶜq_ice, ᶜq_tot_nonneg) = p.precomputed
-        # TODO - follow up with renaming the cached variables to liq and ice
-        ᶜq_liq = ᶜq_liq
-        ᶜq_ice = ᶜq_ice
         ᶜq_tot = ᶜq_tot_nonneg
     else
-        # TODO - change in the next PR. Keeping this non behavior changing
-        ᶜq_liq = @. lazy(specific(Y.c.ρq_lcl, Y.c.ρ)) # TODO + specific(Y.c.ρq_rai, Y.c.ρ))
-        ᶜq_ice = @. lazy(specific(Y.c.ρq_icl, Y.c.ρ)) # TODO + specific(Y.c.ρq_sno, Y.c.ρ))
+        ᶜq_liq = @. lazy(specific(Y.c.ρq_lcl, Y.c.ρ))
+        ᶜq_ice = @. lazy(specific(Y.c.ρq_icl, Y.c.ρ))
         ᶜq_tot = @. lazy(specific(Y.c.ρq_tot, Y.c.ρ))
     end
     ᶜθ_li = @. lazy(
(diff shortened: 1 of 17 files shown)


src/parameterized_tendencies/microphysics/cloud_fraction.jl
```diff
@@ -246,176 +256,265 @@ end
...
+   `α → 0` and the cloud fraction goes smoothly to zero.
+2. Activation-scaled SGS standard deviation: `σ_qc = α √(σ_S²)`.
+3. Effective excess `Q_eff` from cached `μ_S`, dispatched on `sgs_dist`:
+   - Gaussian/GridMean: `Q_eff = q_c + min(0, μ_S)` (linear coordinates).
+   - Lognormal: `Q_eff = log(1 + q_c/q_v_sat) + min(0, μ_S)` (log coordinates).
+4. Cloud fraction via variance-matched logistic CDF
+   (`erf(x/√2) ≈ tanh((π/(2√3)) x)`):
+   `f_c = ½ [1 + tanh(c_f · Q_eff / σ_qc)]` with
+   `c_f = (π/(2√3)) · cf_steepness_scale`.
+5. Branchless gate: `f_c = 0` when no prognostic condensate is present.
```
sajjadazimi: When `\mu_S` is positive, `Q_eff` and `\sigma_qc` go to zero together. This can result in nonzero cf for very small values of `qc`. Maybe we can have a small negative offset in `f_c` to ensure zero cloud fraction when `Q_eff ~ eps`. Then step 5 won't be necessary.

src/parameterized_tendencies/microphysics/cloud_fraction.jl
```diff
@@ -246,176 +256,265 @@ end
...
-   from supersaturation to obtain cloud fraction.
-5. Merge liquid and ice via maximum overlap: `cf = max(cf_l, cf_i)`.
-6. Enforce zero cloud fraction when no condensate exists.
-
-With zero variance the function returns 0 when no condensate exists and
-1 when condensate is present, recovering the grid-scale behaviour.
+1. Activation factor on linear grid-mean excess:
+   `α = clamp(q_c / √((q_t − q_sat)_+² + q_min²), 0, 1)` with
+   `q_sat = q_vap_saturation(T, ρ)`. As `q_c → 0` in a clear subdomain,
+   `α → 0` and the cloud fraction goes smoothly to zero.
```
sajjadazimi: Now that we have `\mu_S = \bar{q_sat - q_t}`, maybe we can use that instead of equilibrium of mean values.

src/parameterized_tendencies/microphysics/microphysics_wrappers.jl
```diff
@@ -159,172 +163,131 @@ end
...
+quadrature points using a **shape-function partition** of cloud condensate.
 
-    return q_lcl_eq, q_icl_eq
-end
+At each quadrature point `(T_hat, q_tot_hat)` sampled from the SGS PDF,
+the local cloud condensate is constructed as
 
-"""
-    (eval::Microphysics1MEvaluator)(T_hat, q_tot_hat)
+    q_lcl_hat = q_lcl_mean · (γ_l · q_lcl_eq_hat + β_l)
```
sajjadazimi: It might be more intuitive if we express this equation as the mean plus fluctuations:
`q_lcl_hat = q_lcl_mean \cdot (1 + \gamma_l \cdot q_lcl_eq_hat + \beta_l - 1) = q_lcl_mean + q_lcl_mean \cdot \gamma_l \cdot (q_lcl_eq_hat - M_l)`

sajjadazimi (commented):
Thanks!

src/parameterized_tendencies/microphysics/cloud_fraction.jl
tapios: Good point. But subtracting something from Q_eff is dimensionally messy because it is a specific humidity in the Gaussian case and a log ratio in the lognormal case. 

Another option may be to return `cf * α` to force the cloud fraction to zero when `α` goes to zero? It's a bit hacky but does the right thing. What do you think, @sajjadazimi?

Just make the changes you want directly.

src/parameterized_tendencies/microphysics/cloud_fraction.jl
tapios: Good idea!

src/parameterized_tendencies/microphysics/cloud_fraction.jl
tapios: I added the alpha scaling to the branch. Feel free to modify though.

src/parameterized_tendencies/microphysics/microphysics_wrappers.jl
```diff
@@ -254,24 +309,47 @@ NamedTuple with microphysics source terms:
...
+    q_min = TD.Parameters.q_min(thp)
+
+    # Precompute shape-function coefficients once per grid cell (saving
+    # divisions and multiplications inside the quadrature loop).
+    M_l = max(FT(0), moments.M_l)
+    M_i = max(FT(0), moments.M_i)
+    M_sq = q_min * q_min
+    denom_l = M_sq + M_l * M_l
+    denom_i = M_sq + M_i * M_i
+    γ_l = M_l / denom_l
```
jbphyswx: If I understand correctly, these need to be clamped to prevent multiplication from scaling perturbations beyond the real equilibrium perturbation magnitude. As written I believe when `q > q_min`, the saturated portion of the domain receives ~all the condensate, leading to potentially severe crowding and subsaturation when `q_c >> M_l `. Only once we are fully sub-saturated (or at least below q_min) do we transition to a uniform distribution. I believe the we should clamp to `1/q_lcl` (?), such that the contribution from the perturbations are no larger than they are in equilibrium. Under clamping we also need to adjust `\beta`

src/parameterized_tendencies/microphysics/cloud_fraction.jl
sajjadazimi: This could be problematic as for supersaturated cases with `q_l < M_l`, more supersaturation with the same `q_l` causes less cloud fraction.

sajjadazimi (approved):

PR #4482, opened 2026-05-12
Use central guidelines from DeveloperGuides repo

tapios:
## Purpose 

Removes local agent files (except repo-specific files) and instead uses DeveloperGuides as a subtree.

## Content

----


imreddyTeja:
We ran into issues using submodules in the past in ClimaCore.  Here is the ClimaCore [PR](https://github.com/CliMA/ClimaCore.jl/pull/2442) where it was removed. There is an open [issue](https://github.com/JuliaLang/Pkg.jl/issues/708) for this in julia. I can  check if this is still a problem and look into alternatives.

tapios:
> We ran into issues using submodules in the past in ClimaCore. Here is the ClimaCore [PR](https://github.com/CliMA/ClimaCore.jl/pull/2442) where it was removed. There is an open [issue](https://github.com/JuliaLang/Pkg.jl/issues/708) for this in julia. I can check if this is still a problem and look into alternatives.

Thanks. I made it a subtree instead, with a workflow that pulls it regularly from the DeveloperGuides repo. 

PTAL.

petebachant:
@kmdeck had an idea for an alternative approach that might be a little simpler: only document repo-specific guidelines in individual repos, and add a link to the more general guidelines that apply across the org. I believe agents can handle fetching links without approval.

tapios:
We can give that a try. It's simple at least.

.gitmodules
imreddyTeja, line 1: This file should be removed before merging. Empty .gitmodules still causes issues [pr example](https://github.com/CliMA/ClimaCore.jl/pull/2442/changes)

imreddyTeja (approved):
Will this PR still be merged, or should we try the link suggestion first.

tapios:
Here’s Claude’s take on the question:

> The subtree approach is meaningfully more reliable. The trade-offs:
> 
> URL-only approach (AGENTS.md links to GitHub URLs)
> Agents can follow links, but only if they have a working web-fetch tool and network access at that moment. Many agent setups don't, or sandbox it. Claude Code's WebFetch works, but it costs a round-trip per file and is subject to rate limits / auth.
> Each link is a separate fetch — if an agent needs to consult five guides, that's five HTTP calls instead of five local reads. In practice agents skip "optional reading" when it's remote.
> No guarantee the agent sees the version that matches the repo's current state — main of DeveloperGuides may have drifted from what the consumer repo was designed against. With a subtree, the guides are pinned to a commit alongside the code.
> Offline / air-gapped / restricted-network CI agents simply can't read them.
> Grep, Read, and Glob don't traverse URLs — anything that depends on filesystem search misses the guides entirely.
> 
> Subtree approach (current setup)
> Guides sit at docs/dev-guides/ as plain files. Every tool — Read, Grep, Glob, editor search — sees them. No network needed.
> Pinned per commit: the guides a reviewer reads are the guides the PR was written against. Drift is explicit (a sync PR), not silent.
> Slightly heavier maintenance: the weekly update_dev_guides.yml workflow handles propagation, so in practice it's set-and-forget.
> Cost: ~a few hundred KB per consumer repo, and contributors need to know edits go to CliMA/DeveloperGuides, not the vendored copy (the README already calls this out).
> 
> Recommendation: keep the subtree as the primary mechanism. Linking to the canonical URL from a top-level README is fine as a secondary hint for humans browsing on GitHub, but agents should be pointed at the local docs/dev-guides/ path. The current AGENTS.md.template already does this correctly.
> 
> One hybrid worth considering: in the consumer's root AGENTS.md, mention both ("Shared guides live at docs/dev-guides/ — canonical source: https://github.com/CliMA/DeveloperGuides"). That gives an agent a fallback if the subtree is somehow missing, without making it the default path.

My personal take is also that having the docs in a repo in which I’m working will make it more likely and easier for me to consult them (rather than having to go to a different repo). The downside of the subtree approach is PR noise. We can reduce that by reducing the update frequency (e.g., to monthly), as the updates usually won’t be critical.

petebachant (approved):

PR #4486, opened 2026-05-13
Standardize YAML config parsing

### LINKED ISSUE #4176: Change `vert_diff` config option
In `default_config.yml`, we have:
```yaml
vert_diff:
  help: "Vertical diffusion [`false` (default), `VerticalDiffusion`, `true` (defaults to `VerticalDiffusion`), `DecayWithHeightDiffusion`]"
  value: "false"
```
This is a rather confusing combination of booleans and string flags. Additionally, many (12) configs have set this flag to:
```yaml
vert_diff: true
```
which leads to the warning:
```julia
┌ Warning: Failed to convert `config_dict["vert_diff"] = true` to default type String, keeping original value
└ @ ClimaAtmos ~/ClimaAtmos.jl/src/solver/yaml_helper.jl:78
```
Ultimately, the expected vertical diffusion option is applied, but it is indeed rather confusing and not user-friendly.

----

I think we should update the option as follows:
```yaml
vert_diff:
  help: "Vertical diffusion [`~` (no diffusion, default), `VerticalDiffusion`, `DecayWithHeightDiffusion`]"
  value: ~
```
that is, removing `true/false` options.

---

If changing the valid options is a breaking change, I think we should consider changing the name to something more descriptive, e.g. `vertical_diffusion_model`. 

---

This issue may be less relevant when the new simulation interface is merged.

nefrathenrici:
This PR standardizes the YAML config parsing. Closes #4176. This is a breaking change for the coupler.

Content:
- Strict coercion in override_default_config: error on type mismatch instead of silently keeping the wrong-typed value. Added EXCEPTED_KEYS bypass for this, restricted to `hyperdiff` (String|nothing union) and `diagnostics` (structured list).
- Schema cleanup: 
  - vert_diff/hyperdiff default to ~ (was "false")
  - prognostic_surface defaults to "PrescribedSST" (was "false")
  - diagnostics now declared in the schema instead of bypassed
- Drop dead keys: subsidence, ls_adv, scm_coriolis. Setups own these via dispatch and this was never updated.
- Simplify get_vertical_diffusion_model, get_hyperdiffusion_model, get_subsidence_model, get_large_scale_advection_model, get_scm_coriolis, and check_case_consistency to match.
- Migrate 22 user configs (vert_diff: true --> "VerticalDiffusion", hyperdiff: "false" --> ~, prognostic_surface: true --> "SlabOceanSST").

### DIFF
--- a/config/default_configs/default_config.yml
+++ b/config/default_configs/default_config.yml
@@ -97,7 +97,7 @@ debug_jacobian:
   value: false
 # Radiation
 rad:
-  help: "Radiation model [`nothing` (default), `gray`, `clearsky`, `allsky`, `allskywithclear`]"
+  help: "Radiation model [`nothing` (default), `gray`, `clearsky`, `allsky`, `allskywithclear`, `held_suarez`, `DYCOMS`, `TRMM_LBA`, `ISDAC`]"
   value: ~
 idealized_h2o:
   help: "Use idealized H2O in radiation model [`false` (default), `true`]"
@@ -149,8 +149,8 @@ reference_job_id:
     Identifier of job to use as the "reference" solution in the quicklook plot; the current job's results get compared to the results of the quicklook job on the main branch
   value: ~
 vert_diff:
-  help: "Vertical diffusion [`false` (default), `VerticalDiffusion`, `true` (defaults to `VerticalDiffusion`), `DecayWithHeightDiffusion`]"
-  value: "false"
+  help: "Vertical diffusion [`~` (no diffusion, default), `VerticalDiffusion`, `DecayWithHeightDiffusion`]"
+  value: ~
 hyperdiff:
   help: |
     Fourth-order hyperdiffusion. Options:
@@ -158,7 +158,7 @@ hyperdiff:
       set these with the `vorticity_hyperdiffusion_coefficient`, `hyperdiffusion_prandtl_number`,
       `divergence_damping_factor` config options.
     - `CAM_SE` (Lauritzen et al. 2018): Use CAM_SE coefficients for hyperdiffusion
-    - `false`: Disable hyperdiffusion
+    - `~`: Disable hyperdiffusion
   value: "Hyperdiffusion"
 amd_les:
   help: "AMD LES closure [`true`, `false` (default)]"
@@ -287,9 +287,6 @@ check_conservation:
 check_steady_state:
   help: "Compare steady-state velocity to analytic solution; only available for certain choices of `topography` [`false` (default), `true`]"
   value: false
-ls_adv:
-  help: "Large-scale advection [`nothing` (default), `Bomex`, `Rico`]"
-  value: ~
 external_forcing:
   help: "External forcing for single column experiments [`nothing` (default), `GCM`, `Reanalysis`, `ReanalysisTimeVarying`, `ReanalysisMonthlyAveragedDiurnal`]"
   value: ~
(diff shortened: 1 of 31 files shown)


config/default_configs/default_config.yml
```diff
@@ -338,6 +332,9 @@ output_default_diagnostics:
 enable_diagnostics:
   help: "Set to false to fully disable the diagnostics"
   value: true
+diagnostics:
+  help: "User-specified diagnostics, as a list of dicts each containing at least `short_name` and `period`. See docs for the full schema."
```
haakon-e: can you provide a link to the relevant docs here?

src/solver/model_getters.jl
```diff
@@ -82,10 +82,12 @@ function get_hyperdiffusion_model(parsed_args, ::Type{FT}) where {FT}
             @assert isapprox(cam_coef, config_val, atol = 1e-8) "CAM_SE hyperdiffusion overwrites $config_coef, use `hyperdiff: Hyperdiffusion` to set this value manually in the config instead."
         end
         return cam_se_hyperdiff
-    elseif hyperdiff_name ∈ ("false", false, nothing)
+    elseif isnothing(hyperdiff_name)
         return nothing
     else
-        error("Uncaught hyperdiffusion model type.")
+        error(
+            "Uncaught hyperdiff `$hyperdiff_name`. Expected: ~ | \"Hyperdiffusion\" | \"CAM_SE\".",
```
haakon-e, line 89: This is fine. One alternative that may reduce future maintainability burden is to instead say e.g.
```julia
"Uncaught hyperdiff `$hyperdiff_name`. See `src/default_configs/default_config.yml` for valid options"
```
If you have `ClimaAtmos.jl` open in an IDE like VSCode, you can `Cmd+click` that path and it will open the file, making it easy for users to navigate to the relevant option.

This suggestion is also relevant for other similar error messages.

src/solver/yaml_helper.jl
```diff
@@ -60,41 +60,56 @@ function override_default_config(::Nothing)
...
+# (hyperdiff has a String default but accepts `~` to disable) and structured
+# values (`diagnostics` is a list of dicts).
+const EXCEPTED_KEYS = Set([
+    "hyperdiff",
+    "diagnostics",
+])
+
+# Coerce a user-supplied value to the schema default's type. Handles the
+# common YAML quirk of quoted Bool/Int/Float values (`"true"`, `"42"`).
+coerce_to_default(::Type{T}, v::T) where {T} = v
```
haakon-e, line 118: Could you add a bit more comprehensive information about what these sets of methods do?

parsing and types often confuse developers, so we should try to minimize future maintenance burden by explaining, with examples, what these do

haakon-e (commented):
Great work! 

I have a few minor comments intended to help users and developers down the line with this new interface.

haakon-e (approved):

src/solver/yaml_helper.jl
```diff
@@ -60,41 +60,56 @@ function override_default_config(::Nothing)
     return default_config_dict()
 end
 
+# Keys that bypass scalar coercion: `named String | nothing` unions
+# (hyperdiff has a String default but accepts `~` to disable) and structured
+# values (`diagnostics` is a list of dicts).
+const EXCEPTED_KEYS = Set([
+    "hyperdiff",
```
szy21, line 67: do we need to add vert_diff too? I think it's similar to hyperdiff?

szy21 (approved):
Thanks! Could you check some single column cases and see if the results changed, as we don't have a regresstion test now?

src/solver/yaml_helper.jl
nefrathenrici: vert_diff now has a default of nothing, so it never gets converted and does not need to be excepted.

src/solver/model_getters.jl
nefrathenrici: I prefer to keep it as-is. I think this could be useful for config keys with many options, such as `rad`, but for something like hyperdiff it adds an additional step.

src/solver/yaml_helper.jl
nefrathenrici: I added a comprehensive docstring to clarify the behavior of these methods.

haakon-e: great improvement, thanks!

PR #4488, opened 2026-05-13
Cleanup in prognostic edmf 1M CI cases

trontrytel:
- Unified the numerics setup for the 3 precipitating prognostic EDMF cases.
- Deleted the variants of them that are not useful for flagship AMIP configuration (like explicit diffusion, microphysics or edmf).
- Deleted the trmm sparse autodif case (with consent from the person who worked on it). we still have two CI cases for autodiff
- Added the three precipitating cases to regression tests


szy21 (approved):
Thank you! It's probably good enough to just add one to regression tests, but no harm adding all of them.

trontrytel:
> Thank you! It's probably good enough to just add one to regression tests, but no harm adding all of them.

Lets see how much we suffer with it.

PR #4493, opened 2026-05-14
Split downgrade tests into groups

nefrathenrici:
This PR splits the downgrade test into groups, matching the "ci" GHA.

### DIFF
--- a/.github/workflows/downgrade.yml
+++ b/.github/workflows/downgrade.yml
@@ -17,11 +17,21 @@ concurrency:
 
 jobs:
   test:
-    name: Downgrade ${{ matrix.version }} - ubuntu-latest
+    name: Downgrade ${{ matrix.version }} - ubuntu-latest - ${{ matrix.test_group }}
     runs-on: ubuntu-latest
     strategy:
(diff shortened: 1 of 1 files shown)


ph-kev:
This might causes concurrency limits (see https://docs.github.com/en/actions/reference/limits#job-concurrency-limits-for-github-hosted-runners). I think parallelizing via GHA runners shouldn't be done and it would probably be better to use Distributed.jl, ParallelTestRunner.jl, or something else for running multiple jobs in parallel if possible.

haakon-e:
> This might causes concurrency limits (see https://docs.github.com/en/actions/reference/limits#job-concurrency-limits-for-github-hosted-runners).

which plan are we on?

ph-kev:
> > This might causes concurrency limits (see https://docs.github.com/en/actions/reference/limits#job-concurrency-limits-for-github-hosted-runners).
> 
> which plan are we on?

I am not sure what plan we are, but the organization is capped at 5 MacOS runners. I am guessing the plan we are on gives us 40 or 60 concurrent runners.

PR #4504, opened 2026-05-20
Remove duplicate callback, add coupler perf pipeline

nefrathenrici:
This PR removes the duplicate `enforce_physical_constraints_callback` callback, which was duplicated  in `common_callbacks` and `default_model_callbacks` by PR #4469. 
To catch future performance regressions, this PR also adds an end-to-end coupler performance pipeline. This pipeline runs the flagship `amip_progedmf_1m_land_he16` config on clima and checks that the SYPD has not decreased by more than 1%. This pipeline is adapted from the ClimaCore end to end perf [pipeline](https://github.com/CliMA/ClimaCore.jl/tree/main/.buildkite/perf). It is not designed to block PRs, but to track performance regressions in the coupler. It will run on every merged PR and can be manually triggered by adding [perf] to the commit message. This PR increases the SYPD of the pipeline by ~10%, from 0.22 to 0.24 ([job](https://buildkite.com/clima/climaatmos-end-to-end-performance/builds/2#019e4688-1f17-4e34-986d-1fae3bf91735)).

PR #4505, opened 2026-05-20
Fix duplication of enforce physical constraints callback

imreddyTeja:
## Purpose 
I think this caused the performance regression. This callback is already added in `default_model_callbacks`

## To-do
- verify performance diff before and after fix

Looking at this, I also realized that there are further complications from `microphysics_tendencies_1m`. It is probably worth looking into that. I believe this is the main cause of the highly unstable sypd

----


szy21:
Thanks, it's actually fixed here: https://github.com/CliMA/ClimaAtmos.jl/pull/4504

imreddyTeja:
closed as duplicate of #4504

PR #4507, opened 2026-05-21
Modify dependabot versioning-strat

imreddyTeja:
## Purpose 
Dependabot adds new compat entries when upgrading packages. We do not want this for the .buildkite env. I think this might fix it, but I'm not sure how to test. [dependabot docs](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#versioning-strategy--)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -15,6 +15,7 @@ updates:
       patterns:
       - "*"
   package-ecosystem: julia
+  versioning-strategy: increase-if-necessary
   schedule:
     day: wednesday
     interval: weekly


ph-kev (commented):
It seems that this option is only supported by: `bundler`, `cargo`, `composer`, `mix`, `npm`, `pip`, `pub`, and `uv`. Is there a reason why you think this should work?

imreddyTeja:
> It seems that this option is only supported by: `bundler`, `cargo`, `composer`, `mix`, `npm`, `pip`, `pub`, and `uv`. Is there a reason why you think this should work?

Ah, good point. I don't think there is any way to get Dependabot to not add compats when they do not exist (based on [this](https://github.com/dependabot/dependabot-core/blob/4199a46f6d647512a686e8295bd7b7acf64e9e79/julia/spec/dependabot/julia/requirements_updater_spec.rb#L80). Do you think I should add ignores for all the deps without a compat, or just Allow Dependabot to add compats. (or manually edit every Dependabot pr)?

PR #4508, opened 2026-05-21
make pressure drag a function of area fraction

szy21:
## Purpose 

Following the EDMF paper. The `pressure_normalmode_drag_coeff` is reduced by a factor of 5 so that the drag coefficient is the same as before when the area fraction is 0.01, which is a typical value in AMIP. I didn't change the diagnostic EDMF as I don't want to retune the parameters.

## Content

----


src/prognostic_equations/implicit/initialize_implicit_problem.jl
```diff
@@ -160,7 +159,9 @@ function solve_sgs_u₃_implicit_stage_analytic!(Y, p, dtγ)
 
         # Implicit NH pressure drag contributes a quadratic sink in w².
         if p.atmos.edmfx_model.nh_pressure isa Val{true}
-            @. ᶠa += ᶠinterp(drag_coeff * ᶜρ_over_ρa⁰ * ᶜρ_over_ρa⁰) / ᶠdz
+            ᶜaʲ = @. lazy(draft_area(Y.c.sgsʲs.:($$j).ρa, ᶜρʲs.:($$j)))
+            ᶜdrag_coeff = @. lazy(α_d / (scale_height * sqrt(max(ᶜaʲ, a_min))))
```
tapios: This is a little different than what's in the paper draft in that it does not ensure the symmetry of the harmonic mean in the paper draft (and hence does not guarantee that the sum of the subdomain drag terms vanishes). I think this may be fine for now, but at least we should comment on it.

tapios (approved):
Just a comment on the symmetry property of the drag term.

src/prognostic_equations/implicit/initialize_implicit_problem.jl
szy21: Changed to the harmonic mean form.

PR #4512, opened 2026-05-22
Delete explicit vs implicit options (always implicit)

sajjadazimi:
## Purpose 
Remove SGS explicit/implicit timestepping options (always implicit): 
- SGS advection, 
- entrainment/detrainment, 
- nh pressure drag, 
- SGS vertical diffusion, 
- and SGS mass flux.

----


szy21 (approved):
Thank you! It's much cleaner now.

PR #4513, opened 2026-05-22
Add pre-commit config and instructions for use with `prek`

petebachant:
Addresses one low-hanging-fruit dev productivity item from the CliMA retreat brainstorm.


tapios (approved):
This is very helpful (dealing with the formatter has been frustrating). Thank you!

Is there a good way to make this part of the central DeveloperGuides, so we can easily gets this into every repo (or at least those that include the DeveloperGuides as a subtree)? I want to minimize the maintenance burden here.

petebachant:
Added a more complex subscription/sync system in https://github.com/CliMA/DeveloperGuides/pull/3 and used it here.

tapios:
> Added a more complex subscription/sync system in [CliMA/DeveloperGuides#3](https://github.com/CliMA/DeveloperGuides/pull/3) and used it here.

Thank you. It would be good, e.g., for @imreddyTeja to look at this, to see if there is a less custom sync solution.

imreddyTeja:
I didn't find any recently maintained pre-made tools that would accomplish a "subscriber opt-in sync". There are several that could work for a "push to all downstream repos" sync

PR #4516, opened 2026-05-22
Add perfect SCM calibration

### LINKED ISSUE #4506: SCM calibration pipeline
We will add a perfect-model calibration pipeline which will run at least nightly on the clima gpu server, depending on runtime.

This pipeline will initially run the [TRMM column config](https://github.com/CliMA/ClimaAtmos.jl/blob/main/config/model_configs/prognostic_edmfx_trmm_column.yml).

Once the initial pipeline is complete, it will be adapted to:
- handle multiple configurations per ensemble member
- use LES data

Covariance: diagonal, no latitude weighting needed

Observations: hourly liquid and ice water path

Open questions:

- At what times should instantaneous observations be sampled?
- Should minibatching be used to initialize from different dates?
- What priors should be used?
- What results / deliverables should we expect from a pipeline run? How will the calibration outputs be used?

ph-kev:
closes #4506

This PR adds a perfect SCM calibration using ClimaCalibrate v0.3 which should be similar to the calibration pipeline in ClimaCoupler and ClimaLand.

The entry point to the calibration is `run_calibration.jl` which will generate data from a simulation and run the calibration to fit against the data.

TODO

- [x] Add support for including average profiles, time series data, and aggregation/reductions
- [x] ~~Ability to run multiple simulations from multiple configurations for a single calibration~~ (This feature will be added in another PR)
- [x] https://github.com/CliMA/ClimaCalibrate.jl/pull/319 which is needed for postprocessing iterations
- [x] Replace buildkite pipeline at `calibration/test/pipeline.yml`

Optional

- [ ] Make a new release of ClimaCalibrate to use the new reconstruct functions


calibration/experiments/perfect_scm/model_interface/generate_observations.jl
```diff
@@ -0,0 +1,120 @@
...
+        short_names = diagnostic_dict["short_name"]
+        reduction = diagnostic_dict["reduction_time"]
+        # TODO: We can't use the period from diagnostic_dicts to determine which
+        # OutputVar to load. See the comment in process_member_data!
+        for short_name in short_names
+            periods = ClimaAnalysis.available_periods(simdir; short_name, reduction)
+            length(periods) > 1 &&
+                error(
+                    "There are more than one period ($periods) for $short_name with $reduction. Remove the conflicting diagnostic in the config or from the diagnostic dictionary",
+                )
```
ph-kev: The limitation is that we can't distinguish between diagnostics with the same short name and reduction, but different periods.

Is this limitation okay or should it be fixed in this PR?

calibration/experiments/perfect_scm/model_interface/generate_observations.jl
```diff
@@ -0,0 +1,120 @@
...
+            push!(vars, get(simdir; short_name, reduction))
+        end
+    end
+
+    # Create observations using all the data
+    # This is where we would do preprocessing and minibatching before creating the
+    # EKP.Observation
```
ph-kev, line 110: I was not sure what preprocessing is desired here.

calibration/experiments/perfect_scm/model_interface/postanalyze_iteration.jl
```diff
@@ -0,0 +1,43 @@
...
+function ClimaCalibrate.analyze_iteration(
+    ::PerfectAtmosModelInterface,
+    ekp,
+    g_ensemble,
+    prior,
+    output_dir,
+    iteration,
+)
+    plot_constrained_params_and_errors(output_dir, ekp, prior)
+    return nothing
```
ph-kev: Is there any specific plotting that need to happen here?

calibration/experiments/perfect_scm/run_calibration.jl
```diff
@@ -0,0 +1,85 @@
...
+
+model_interface_filepath = joinpath("model_interface", "model_interface.jl")
+include(model_interface_filepath)
+
+# Set up model interface
+config = "config/model_configs/baroclinic_wave.yml"
+output_dir = "perfect_model_calibration"
+diagnostic_dict =
+    Dict("short_name" => ["pr"], "period" => "1days", "reduction_time" => "average")
+interface = PerfectAtmosModelInterface(config, output_dir, diagnostic_dict)
```
ph-kev: This is not a good choice of config, diagnostic, and priors.

nefrathenrici: To start, I think  you can use 

`config = "config/model_configs/prognostic_edmfx_trmm_column.yml"`

and

"short_name" => ["lwp", "clivi"].

@trontrytel can correct me if this is wrong.

calibration/experiments/perfect_scm/Manifest.toml
nefrathenrici, line 1: I don't think we need a checked-in manifest for this.

calibration/experiments/perfect_scm/run_calibration.sh
nefrathenrici, line 1: Is this file needed? I think it would be better to run this calibration in a buildkite step.

calibration/experiments/perfect_scm/run_calibration.jl
```diff
@@ -0,0 +1,85 @@
...
+    ),
+)
+
+# Set seed for ensure reproducibility
+rng_seed = 1234
+rng = Random.MersenneTwister(rng_seed)
+
+# Set priors for parameters of interest
+calibration_priors = [
+    PD.constrained_gaussian("precipitation_timescale", 1200, 300, 300, 1800),
```
nefrathenrici: We still need to decide on the priors for this calibration.

nefrathenrici (commented):
This looks clean, we still need to decide on the priors and observation details.

calibration/experiments/perfect_scm/model_interface/generate_observations.jl
ph-kev: After using the calibration a bit more, I think this can cause problems.

calibration/experiments/perfect_scm/run_calibration.sh
ph-kev: I am still not sure if this should be in a buildkite pipeline or not. I found it easier to use the `JuliaBackend` for testing since it keeps all the precompiled code around.

nefrathenrici: If the runtime is under an hour when using the `JuliaBackend` I think it can replace the test that's current in `calibration/test/pipeline.yml`

ph-kev:
I ran the calibration again on the latest [commit](https://github.com/CliMA/ClimaAtmos.jl/commit/a1992640193db18b777a90239474f90eca7d647b) and it seems to work now. Some of the parameters are not identifiable though.

image

nefrathenrici:
> I ran the calibration again on the latest [commit](https://github.com/CliMA/ClimaAtmos.jl/commit/a1992640193db18b777a90239474f90eca7d647b) and it seems to work now. Some of the parameters are not identifiable though.
> image

The cloud ice terminal velocity parameter doesn't seem to be informed by the data, I think we can remove it from the calibration.

calibration/experiments/perfect_scm/run_calibration.sh
nefrathenrici: I still think this file can be deleted.

nefrathenrici (approved):
Thanks! I have a few comments to add finishing touches:
- Delete the now-unused test files in `calibration/test/`. I think the entire folder can be removed.
- Delete `experiments/perfect_scm/run_calibration.sh` if it is not run anywhere.
- Update the base `calibration/README.md` to reference the nice README you already wrote.
- Update the artifact path in the pipeline so that plots can be seen directly from buildkite.

calibration/experiments/perfect_scm/model_interface/generate_observations.jl
```diff
@@ -0,0 +1,129 @@
...
+function to see what kind of processing is done.
+
+The covariance matrix for each variable is the
+`1e-11 + (model_error_scale * Statistics.mean(var.data))^2` where the
+`model_error_scale = 0.05`.
+"""
+function create_ekp_observations(
+    diagnostics_filepath,
+    interface::PerfectAtmosModelInterface,
+)
```
odunbar, line 88: Do we not want the user to be able to pass in their own ObservationRecipe here? 
I would think this is something that is often subject to change.

I think a good calibration experiment hides a lot of the "mess" but leaves open the ability for users to modify key features that affect their experiments

calibration/experiments/perfect_scm/model_interface/generate_observations.jl
```diff
@@ -0,0 +1,129 @@
...
+    end
+
+    # Create observations using all the data
+    # This is where we would do preprocessing and minibatching before creating the
+    # EKP.Observation
+    obs_vec = map(vars) do var
+        var = preprocess(var, interface)
+
+        model_error_scale = 0.05
+        scalar = 1e-11 + (model_error_scale * Statistics.mean(var.data))^2
```
odunbar, line 115: It may be useful to have another script that can return the preprocessed var's in case users want to adjust the ObservationRecipes out of this function.

calibration/experiments/perfect_scm/run_calibration.jl
```diff
@@ -0,0 +1,113 @@
...
+
+import Distributed
+import JLD2
+import CairoMakie
+
+model_interface_filepath = joinpath("model_interface", "model_interface.jl")
+include(model_interface_filepath)
+
+# Set up model interface
+config = joinpath("config", "model_configs", "prognostic_edmfx_trmm_column.yml")
```
odunbar, line 13: It is not clear what the true parameters are - I would obtain these for the user - extracted from the simulation, or at least direct them to their location (e.g. ClimaParams.jl defaults?), else explicitly overwrite them locally with fixed values so they do not change

calibration/experiments/perfect_scm/run_calibration.jl
```diff
@@ -0,0 +1,113 @@
...
+    ),
+)
+
+# Set seed for ensure reproducibility
+rng_seed = 1234
+rng = Random.MersenneTwister(rng_seed)
+
+# Set priors for parameters of interest
+means = [5.0, 1.0, 0.01, 0.01]
+mean_std_dev_lub(m) = (m + 0.05 * m, m * 0.2, 0.0, m + 4 * m * 0.2)
```
odunbar: Just checking if the upper bound here is justified? (hard to tell as the user does not easily know the "true" params here

odunbar: I typically would not place hard bounds unless they are needed (e.g. the 0.0 likely is)

calibration/experiments/perfect_scm/run_calibration.jl
```diff
@@ -0,0 +1,113 @@
...
+# Run calibration
+n_iterations = 5
+eki = ClimaCalibrate.calibrate(
+    backend,
+    ekp,
+    interface,
+    n_iterations,
+    prior,
+    interface.output_dir,
+)
```
odunbar, line 120: It might be helpful for a user looking at this script, for them to see how to obtain the final `u_ens` and `g_ens` and `reconstructed_g_ens` ( via `reconstruct_g_ens`).  Just to demonstrate the functionality at the top level?

odunbar (changes requested):
Hi @ph-kev This looks nice! 

I think my points are essentially of one theme: over-wrapping. Currently we have really nicely contained a lot of the boilerplate - but at times I think this will come to the detriment to what a future user might want to use this for. (I understand by default it is fed into a test pipeline) but it should not be written as if it will not be run by a user.

With that in mind, I would urge exposing a couple of items to the `run_calibration.jl` top-level script:
- The true parameter values should be accessible
- The true processed data should be accessible (before creating the observations)
- The observation recipe should be changeable.
- Perhaps add at the end of a script a function that gets the iterations from the eki, and builds them with the outputVar
In this way we are giving users a similar degree of freedom as you do with the prior - which is easily modifiable at the top level! 


Update: Created Issue https://github.com/CliMA/ClimaCalibrate.jl/issues/328 for the user-defined covariances, that are beyond the scope of the PR

calibration/experiments/perfect_scm/model_interface/generate_observations.jl
odunbar: https://github.com/CliMA/ClimaCalibrate.jl/issues/328

calibration/experiments/perfect_scm/model_interface/generate_observations.jl
odunbar: https://github.com/CliMA/ClimaCalibrate.jl/issues/328

calibration/experiments/perfect_scm/run_calibration.jl
ph-kev: I am going to put `Inf` as the upper bound, but @trontrytel would know.

calibration/experiments/perfect_scm/run_calibration.jl
ph-kev: I don't think there is an easy way of querying for the parameters of a simulation after it is created, so I ended up letting the user specify their TOML file that can be used to overwrite the parameters of the simulation.

calibration/experiments/perfect_scm/run_calibration.jl
ph-kev: I think it would be better to have a section for this in ClimaCalibrate (see https://github.com/CliMA/ClimaCalibrate.jl/pull/330).

calibration/experiments/perfect_scm/model_interface/generate_observations.jl
```diff
@@ -0,0 +1,134 @@
...
+    obs_vec = create_ekp_observations(diagnostics_filepath, interface)
+    JLD2.save_object(obs_vec_filename, obs_vec)
+    return nothing
+end
+
+"""
+    generate_perfect_model_data(interface::PerfectAtmosModelInterface; tomls = [])
+
+Generate data by running a `ClimaAtmos` simulation with the configuration
+at `interface.config`.
```
odunbar, line 34: Quick note in the docstring about the tomls kwarg

odunbar (approved):
Just add a new `tomls` kwarg note in the docstring - otherwise lgtm! Thanks!

calibration/experiments/perfect_scm/run_calibration.jl
trontrytel: Definitely 0 as the lower bound. For upper bound I'm not sure. `Inf` seems like an overkill since I know that those sedimentation velocities should not be faster than for example your highway speed limit in California. Does the algorithm benefit much from adding an upper bound?

calibration/experiments/perfect_scm/run_calibration.jl
trontrytel: I think that specifying from toml file should work great

trontrytel:
> Hi @ph-kev This looks nice!
> 
> I think my points are essentially of one theme: over-wrapping. Currently we have really nicely contained a lot of the boilerplate - but at times I think this will come to the detriment to what a future user might want to use this for. (I understand by default it is fed into a test pipeline) but it should not be written as if it will not be run by a user.
> 
> With that in mind, I would urge exposing a couple of items to the `run_calibration.jl` top-level script:
> 
> * The true parameter values should be accessible
> * The true processed data should be accessible (before creating the observations)
> * The observation recipe should be changeable.
> * Perhaps add at the end of a script a function that gets the iterations from the eki, and builds them with the outputVar
>   In this way we are giving users a similar degree of freedom as you do with the prior - which is easily modifiable at the top level!
> 
> Update: Created Issue [CliMA/ClimaCalibrate.jl#328](https://github.com/CliMA/ClimaCalibrate.jl/issues/328) for the user-defined covariances, that are beyond the scope of the PR

Along the same lines. I think when using it I would want to be able to switch:
- between different forward model configurations (i.e. run it for different single column cases, or combinations of multiple single column cases)
- change the observations and loss function definitions
- change the calibrated parameters.

calibration/experiments/perfect_scm/model_interface/postanalyze_iteration.jl
trontrytel: Would be nice to also plot the model results (for example the water paths we are calibrating).

trontrytel (approved):

calibration/experiments/perfect_scm/model_interface/postanalyze_iteration.jl
ph-kev: This is one of the plots that are generated every iteration.

image

ph-kev:
> > Hi @ph-kev This looks nice!
> > I think my points are essentially of one theme: over-wrapping. Currently we have really nicely contained a lot of the boilerplate - but at times I think this will come to the detriment to what a future user might want to use this for. (I understand by default it is fed into a test pipeline) but it should not be written as if it will not be run by a user.
> > With that in mind, I would urge exposing a couple of items to the `run_calibration.jl` top-level script:
> > 
> > * The true parameter values should be accessible
> > * The true processed data should be accessible (before creating the observations)
> > * The observation recipe should be changeable.
> > * Perhaps add at the end of a script a function that gets the iterations from the eki, and builds them with the outputVar
> >   In this way we are giving users a similar degree of freedom as you do with the prior - which is easily modifiable at the top level!
> > 
> > Update: Created Issue [CliMA/ClimaCalibrate.jl#328](https://github.com/CliMA/ClimaCalibrate.jl/issues/328) for the user-defined covariances, that are beyond the scope of the PR
> 
> Along the same lines. I think when using it I would want to be able to switch:
> 
>     * between different forward model configurations (i.e. run it for different single column cases, or combinations of multiple single column cases)
> 
>     * change the observations and loss function definitions
> 
>     * change the calibrated parameters.

The functionality to switch which simulation is being ran is already implemented. However, it is not possible to calibrate using combinations of multiple single column cases. This can be done in another PR.

I think it should be easy to change the observations and which parameters are calibrated, but it might not be so easy to change the loss function definition. I am currently working on https://github.com/CliMA/ClimaCalibrate.jl/issues/328, so it should be easier.

PR #4518, opened 2026-05-24
fix toml filename in some longruns

szy21:
## Purpose 

Fix the longrun error here: https://buildkite.com/clima/climaatmos-gpulongruns/builds/764

## Content

----


szy21:
Manually merging this as it only changes longrun

PR #4521, opened 2026-05-26
Implicit entrainment and detrainment

sajjadazimi:
## Purpose 
Make entrainment and detrainment use updated (implicit) velocity

----

### DIFF
--- a/src/cache/precomputed_quantities.jl
+++ b/src/cache/precomputed_quantities.jl
@@ -291,9 +291,12 @@ function precomputed_quantities(Y, atmos)
         (;
             ρtke_flux = similar(Fields.level(Y.f, half), C3{FT}),
             ᶜentrʲs = similar(Y.c, NTuple{n, FT}),
+            ᶜentr_vel_scaleʲs = similar(Y.c, NTuple{n, FT}),
+            ᶜentr_nonvelʲs = similar(Y.c, NTuple{n, FT}),
             ᶜdetrʲs = similar(Y.c, NTuple{n, FT}),
+            ᶜdetr_nonvelʲs = similar(Y.c, NTuple{n, FT}),
             ᶜturb_entrʲs = similar(Y.c, NTuple{n, FT}),
-            ᶠρ_diffʲs = similar(Y.f, NTuple{n, FT}),
+            ᶜρ_diffʲs = similar(Y.c, NTuple{n, FT}),
             ᶠu₃_tendencyʲs = similar(Y.f, NTuple{n, C3{FT}}),
             precipitation_sgs_quantities...,
         ) : (;)
--- a/src/cache/prognostic_edmf_precomputed_quantities.jl
+++ b/src/cache/prognostic_edmf_precomputed_quantities.jl
@@ -87,6 +87,77 @@ NVTX.@annotate function set_prognostic_edmf_precomputed_quantities_environment!(
 end
 
 """
+    update_prognostic_edmfx_entr_detr!(Y, p)
+
+Compute and store total entrainment/detrainment rates for all updrafts using the
+current values of the velocity-dependent coefficient (`ᶜentr_vel_scaleʲs`) and
+velocity-independent rates (`ᶜentr_nonvelʲs`, `ᶜdetr_nonvelʲs`), which must
+already have been populated (by `set_prognostic_edmf_precomputed_quantities_explicit_closures!`
+or equivalent).
+
+Called from both the explicit closures (to initialise the rates before the first
+implicit solve) and the implicit draft (to refresh them with the post-solve velocity).
+"""
+function update_prognostic_edmfx_entr_detr!(Y, p)
+    turbconv_params = CAP.turbconv_params(p.params)
+    (; ᶜgradᵥ_ᶠΦ) = p.core
+    (;
+        ᶜuʲs,
+        ᶠu³ʲs,
+        ᶜρʲs,
+        ᶜρ_diffʲs,
+        ᶜentrʲs,
+        ᶜentr_vel_scaleʲs,
+        ᶜentr_nonvelʲs,
+        ᶜdetrʲs,
+        ᶜdetr_nonvelʲs,
+    ) = p.precomputed
+    ᶜlg = Fields.local_geometry_field(Y.c)
+    ᶜmassflux_vert_div = p.scratch.ᶜtemp_scalar_2
+    n = n_mass_flux_subdomains(p.atmos.turbconv_model)
+    buoyancy_flux_val =
+        Fields.field_values(p.precomputed.sfc_conditions.buoyancy_flux)
+
+    for j in 1:n
+        @. ᶜentrʲs.:($$j) = compute_entrainment(
+            ᶜentr_vel_scaleʲs.:($$j),
+            ᶜentr_nonvelʲs.:($$j),
+            get_physical_w(ᶜuʲs.:($$j), ᶜlg),
+        )
+        # ρa is not necessarily equal to its final implicit solution here; the
+        # detrainment will be recomputed in the implicit draft after ρa converges.
+        @. ᶜmassflux_vert_div =
+            ᶜdivᵥ(ᶠinterp(Y.c.sgsʲs.:($$j).ρa) * ᶠu³ʲs.:($$j))
+        @. ᶜdetrʲs.:($$j) = compute_detrainment(
+            turbconv_params,
+            draft_area(Y.c.sgsʲs.:($$j).ρa, ᶜρʲs.:($$j)),
+            Y.c.sgsʲs.:($$j).ρa,
+            get_physical_w(ᶜuʲs.:($$j), ᶜlg),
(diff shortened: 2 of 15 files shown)


src/cache/prognostic_edmf_precomputed_quantities.jl
```diff
@@ -86,6 +86,77 @@ NVTX.@annotate function set_prognostic_edmf_precomputed_quantities_environment!(
...
+            p.atmos.edmfx_model.detr_model,
+        )
+        detr_int_val = Fields.field_values(Fields.level(ᶜdetrʲs.:($j), 1))
+        detr_nonvel_int_val =
+            Fields.field_values(Fields.level(ᶜdetr_nonvelʲs.:($j), 1))
+        @. detr_int_val = ifelse(
+            buoyancy_flux_val < 0,
+            detr_int_val,
+            detr_nonvel_int_val,
+        )
```
szy21, line 149: why do we need to treat detrainment differently for the first layer too?

sajjadazimi: Because close to a = 0.1 the area limiter increases detr by almost 10%. So the boundary value of area finds a lower equilibrium if we don't change it back to how it was before applying area limiter. I know it's a bit messy but all these will be fixed once we solve area fraction equation analytically (so boundary conditions and limiters can be applied more easily.)

src/cache/precomputed_quantities.jl
```diff
@@ -291,9 +291,12 @@ function precomputed_quantities(Y, atmos)
         (;
             ρtke_flux = similar(Fields.level(Y.f, half), C3{FT}),
             ᶜentrʲs = similar(Y.c, NTuple{n, FT}),
+            ᶜentr_vel_coeffʲs = similar(Y.c, NTuple{n, FT}),
+            ᶜentr_nonvelʲs = similar(Y.c, NTuple{n, FT}),
```
costachris, line 295: can you make the names consistent or self-descriptive? Here and throughout. 
Perhaps ᶜentr_nonvel_rateʲs or just a comment for which is 1/m and which is 1/s (following some of what's in update_prognostic_edmfx_entr_detr!). The terminology "fractional rate" and  "total rates" is sometimes used to distinguish them.
calling the first a coefficient is confusing since there's also entrainment coefficients in the parameters. maybe prefactor or scale?

src/prognostic_equations/implicit/manual_sparse_jacobian.jl
```diff
@@ -823,13 +815,7 @@ function update_jacobian!(alg::ManualSparseJacobian, cache, Y, p, dtγ, t)
                 dtγ * (
                     DiagonalMatrixRow(ᶜadvdivᵥ(ᶠu³ʲs.:(1))) -
                     ᶜadvdivᵥ_matrix() ⋅
-                    ᶠset_upwind_matrix_bcs(ᶠupwind_matrix(ᶠu³ʲs.:(1))) -
-                    DiagonalMatrixRow(
```
costachris, line 827: why can we drop this term now and similarly the buoyancy term in advection.jl? is that a direct result of making entr/detr implicit or something else?

src/prognostic_equations/edmfx_entr_detr.jl
```diff
@@ -88,89 +88,77 @@ obtained from `turbconv_params` through `CAP.min_area` and `CAP.max_area`.
...
+
+where `entr` has units of [1/s], `entr_coeff` has units of [1/m],
+and `entr_nonvel` has units of [1/s]. The environment velocity is
+approximated as zero when computing the coefficient, so `abs(wʲ)` is
+used rather than the full velocity difference `abs(w⁰ - wʲ)`. This
+split allows `entr_coeff` to be reused after the implicit `u₃` solve
+(where the updraft velocity `wʲ` is updated) without recomputing the
+full coefficient.
+
+`entrainment_coefficient` computes the velocity-scaling factor [1/m] and
```
costachris: As before, I think using "scaling" or "prefactor" or "scale", rather than coefficient, is more clear.  Right now entrainment_coefficient() the function uses entr_coeff the parameter internally, and then the result is stored in ᶜentr_vel_coeffʲs.

src/prognostic_equations/edmfx_entr_detr.jl
```diff
@@ -305,320 +347,151 @@ function entrainment(
...
-    elev_above_sfc = ᶜz - z_sfc
-    # Velocity difference term divided by elev_above_sfc; set to zero if elev_above_sfc is 
-    # not positive to prevent division by zero or excessively large values.
-    vel_diff_term = if elev_above_sfc > eps(FT)
-        entr_coeff * abs(ᶜwʲ - ᶜw⁰) / elev_above_sfc
-    else
-        FT(0)
-    end
-
     area_limiter_factor =
```
costachris: perhaps a helper function for this like you did for detr `detr_area_limiter_factor()`, or generalize and use the same function. Or otherwise remove detr_area_limiter_factor() so its consistent.

src/prognostic_equations/edmfx_entr_detr.jl
```diff
@@ -789,6 +660,87 @@ function edmfx_first_interior_entr_tendency!(
...
+- `entr_coeff_int`      — entrainment coefficient [1/m] (read/write)
+- `detr_nonvel_int` — additional detrainment [1/s] (read/write)
+- `aʲ_int`              — scratch for draft area [-] (write-only; pre-allocated by caller)
+- `buoyancy_flux`       — surface buoyancy flux [m²/s³] (read-only scalar field values)
+- `dz_int`              — first-cell height [m] (read-only)
+- `surface_area`        — target updraft area fraction [-] (scalar)
+- `dt`                  — timestep [s] (scalar)
+- `FT`                  — float type
+"""
+function set_first_cell_entr_detr_bc!(
```
costachris, line 702: Treatment of the bottom boundary is handed in two places (perhaps by necessity?). Would be cleaner to either skip the area limiter at level 1 inside compute_detrainment, or consolidate all the first-cell logic into one place. if it can't be consolidated, can you just note this in the docstring or comment so we remember to update them accordingly in the right place. 
e.g.
```
# Note: the first-cell boundary condition is applied in two stages:
# 1. This function sets the input coefficients (bg_entr, entr_coeff, detr_nonvel).
# 2. `update_prognostic_edmfx_entr_detr!` overrides the total detrainment at
#    level 1 to strip the area limiter added by `compute_detrainment`.
# These two stages should be kept in sync.
```

costachris (changes requested):

src/prognostic_equations/mass_flux_closures.jl
```diff
@@ -48,6 +50,12 @@ function vertical_buoyancy_acceleration(ρ_ref, ρ, gradᵥ_Φ, local_geometry)
     # Project onto vertical axis and return signed scalar value
     return projected_vector_data(C3, buoy_vector, local_geometry)
 end
+function vertical_buoyancy_acceleration(ρ_diff, gradᵥ_Φ, local_geometry)
```
szy21, line 53: Where is this used?

szy21: Oh I see it now.

src/prognostic_equations/edmfx_entr_detr.jl
```diff
@@ -305,320 +347,151 @@ function entrainment(
...
 
-Wrapper around `detrainment` that applies numerical safeguards and common
-post-processing.
+Total detrainment rate [1/s].
 
-If the updraft area fraction `aʲ` becomes numerically negligible,
-detrainment is set to a limiting value to maintain mixing with the
-environment. Otherwise the detrainment rate is computed using the selected
-`detr_model`.
+Calls `detrainment_rate` for the model-specific rate, adds `detr_nonvel`
```
szy21, line 361: Just to make sure I understand. `detrainment rate` is not treated implicitly, and the reason we need to separate `detrainment_rate` and `detr_nonvel` is for boundary conditions?

src/prognostic_equations/implicit/manual_sparse_jacobian.jl
sajjadazimi: It's not directly a result of making entr/detr implicit. I changed the mse eq to use the precomputed buoyancy term (`rho_diff`) for consistency with updraft velocity. Since it's precomputed its Jacobian elements are all zero. So we won't need these lines anymore.
I did it in this PR because here I changed the cached variable `rho_diff` and store it on cell centers so it can be used for detr and mse computations too.

src/prognostic_equations/edmfx_entr_detr.jl
sajjadazimi: `detrainment_rate` is now called in cache_imp with the updated velocity. And yes, `detr_nonvel` is needed for boundary conditions. Hopefully after making area fraction solve analytical these will get much simpler.

costachris (approved):
Thanks!

PR #4525, opened 2026-05-27
Os/musica helloworld

obin1:
## Purpose 
This is an initial pull request for chemistry in ClimaAtmos. It currently prints the Musica.jl version every time the chemical tendencies are calculated.

## To-do
We might want to make chemistry optional and off by default, rather than always on. We also discussed Musica.jl as an optional dependency. The "hello world" version printing should also probably not happen every time a tendency is calculated.

## Content
Currently complete: printing Musica.jl version in the BOMEX single column model. I tested it as follows:

```
julia --project=.buildkite
import ClimaAtmos as CA
Base.pathof(CA) # making sure it is using the correct source code (the local changes)
simulation = CA.get_simulation(CA.AtmosConfig("config/model_configs/prognostic_edmfx_bomex_column.yml"))
sol_res = CA.solve_atmos!(simulation) # run the simulation
```

With the current code, this test succeeds if it prints `MUSICA version: 0.15.0` more times than anyone could ever possibly want

----


trontrytel:
Done via https://github.com/CliMA/ClimaAtmos.jl/pull/4533

PR #4533, opened 2026-05-28
Osaj/musica helloworld

trontrytel:
Following from: https://github.com/CliMA/ClimaAtmos.jl/pull/4525

## Purpose 
This is an initial pull request for chemistry in ClimaAtmos. It currently prints the Musica.jl version every time the chemical tendencies are calculated.

## To-do
We might want to make chemistry optional and off by default, rather than always on. We also discussed Musica.jl as an optional dependency. The "hello world" version printing should also probably not happen every time a tendency is calculated.

## Content
Currently complete: printing Musica.jl version in the BOMEX single column model. I tested it as follows:

```
julia --project=.buildkite
import ClimaAtmos as CA
Base.pathof(CA) # making sure it is using the correct source code (the local changes)
simulation = CA.get_simulation(CA.AtmosConfig("config/model_configs/prognostic_edmfx_bomex_column.yml"))
sol_res = CA.solve_atmos!(simulation) # run the simulation
```

With the current code, this test succeeds if it prints `MUSICA version: 0.15.0` more times than anyone could ever possibly want

### DIFF
--- a/.buildkite/Manifest-v1.11.toml
+++ b/.buildkite/Manifest-v1.11.toml
@@ -1,6 +1,6 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.11.9"
+julia_version = "1.11.6"
 manifest_format = "2.0"
 project_hash = "6dc55dcd2743828c52d667c059aa8515e72171ce"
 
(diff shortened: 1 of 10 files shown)


src/types.jl
```diff
@@ -853,6 +853,18 @@ with single-column model setups. Most external users will not need these compone
     scm_coriolis::SC = nothing
 end
 
+"""
+    AtmosChem
+
+Groups chemistry models and types.
+"""
```
nefrathenrici, line 864: This docstring should go right above `@kwdef struct AtmosChem{CM}`

Project.toml
```diff
@@ -76,6 +82,7 @@ PrettyTables = "3"
 RRTMGP = "0.21.9"
 Random = "1"
 SafeTestsets = "0.1"
+Musica = "0.15"
```
nefrathenrici: Could you put this in alphabetical order please?

ext/ClimaAtmosMusica/ClimaAtmosMusica.jl
```diff
@@ -0,0 +1,23 @@
...
+Loaded automatically when `Musica` is imported alongside `ClimaAtmos`.
+"""
+function ClimaAtmos.chemistry_tendency!(
+    Yₜ,
+    Y,
+    p,
+    t,
+    ::ClimaAtmos.GasPhaseChem,
+)
+    println("MUSICA version: ", Musica.get_version())
```
nefrathenrici:
```suggestion
    @info "MUSICA version: $(Musica.get_version())"
```

nefrathenrici (approved):
Looks good. I think this is neatly isolated so far. I just had a few nitpick comments.

obin1:
thanks for the feedback @nefrathenrici, I just addressed those comments and suggested changes in the latest commits

.buildkite/Manifest-v1.11.toml
```diff
@@ -1,6 +1,6 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.11.9"
+julia_version = "1.11.6"
```
szy21, line 3: It would be good not to change this in the future (unless it's needed for this PR?)

trontrytel: No, sorry thats by accident. Not sure why I thought the Manifest is on 11.6. Must have looked the wrong way.

PR #4534, opened 2026-05-28
Test climacore refactor

nefrathenrici:
Update to ClimaCore v0.14.54. 

Code changes
- Update adjoint storage types, since the recent CC release changed the adjoint representation
- `g³³` helper function now calls new `reshape` rather than rebuilding from components


.buildkite/pipeline.yml
```diff
@@ -1015,6 +1015,7 @@ steps:
           --config_file $CONFIG_PATH/baroclinic_wave_sparse_autodiff.yml
           --job_id baroclinic_wave_sparse_autodiff
         artifact_paths: "baroclinic_wave_sparse_autodiff/output_active/*"
+        soft_fail: true
```
szy21, line 1018: is this needed?

szy21 (approved):

.buildkite/pipeline.yml
nefrathenrici: Yes, the test is unstable now. I can debug it and fix it in climacore this week.

PR #4535, opened 2026-05-29
Initialize albedo as NaN for CouplerAlbedo

juliasloan25:
## Purpose 
All surface models in coupled simulations provide albedo during initialization, before the first step. This means we don't need to set albedo in the radiation cache to a temporary dummy value anymore.

The albedos are already set to NaN at initialization, so we'll be able to easily identify and debug if albedos are not being set correctly by the coupler.

There are no downstream changes needed in ClimaCoupler, unless we want to explicitly check that the atmosphere albedos are not NaN after a CoupledSimulation is created. We don't do this for any other variables though so maybe it's odd to just do it for albedo.

## To Do
- [x] update albedo initial value
- [x] test from coupler - see https://github.com/CliMA/ClimaCoupler.jl/pull/1956 passes [here](https://buildkite.com/clima/climacoupler-ci/builds/9064)

### DIFF
--- a/test/surface_albedo.jl
+++ b/test/surface_albedo.jl
@@ -46,15 +46,18 @@ Random.seed!(1234)
         simulation = ClimaAtmos.get_simulation(config)
         (; u, p) = simulation.integrator
 
-        # At t=0, should initialize to default (0.38)
-        ClimaAtmos.set_surface_albedo!(u, p, 0.0, p.atmos.surface_albedo)
-        @test all(p.radiation.rrtmgp_model.direct_sw_surface_albedo .== FT(0.38))
-        @test all(p.radiation.rrtmgp_model.diffuse_sw_surface_albedo .== FT(0.38))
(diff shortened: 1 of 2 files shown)


test/surface_albedo.jl
```diff
@@ -46,15 +46,18 @@ Random.seed!(1234)
         simulation = ClimaAtmos.get_simulation(config)
         (; u, p) = simulation.integrator
 
-        # At t=0, should initialize to default (0.38)
+        direct_albedo_init = deepcopy(p.radiation.rrtmgp_model.direct_sw_surface_albedo)
+        diffuse_albedo_init = deepcopy(p.radiation.rrtmgp_model.diffuse_sw_surface_albedo)
```
szy21: do we need these?

szy21 (approved):

test/surface_albedo.jl
juliasloan25: We don't! Thank you

PR #4536, opened 2026-05-29
Update fixed coupler commit in e2e pipeline

imreddyTeja:
This has been broken for a bit because of compat issues.

### DIFF
--- a/.buildkite/coupler_perf/pipeline.yml
+++ b/.buildkite/coupler_perf/pipeline.yml
@@ -10,7 +10,7 @@ env:
   GKSwstype: 100
   SLURM_KILL_BAD_EXIT: 1
   COUPLER_PATH: ".buildkite/perf/ClimaCoupler.jl"
-  COUPLER_COMMIT: "6beb62031cb6a8a8a18a4fcd81008d93e1ff0a8f"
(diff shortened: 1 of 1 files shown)


nefrathenrici (approved):
Thanks!

petebachant (approved):
Thanks Teja! If possible, it would be great to document the general procedure you followed to fix this in DeveloperGuides. I saw a broken API with RRTMGP and then a compat conflict with ClimaCalibrate when I tried to update the coupler in my local dev setup.

imreddyTeja:
> Thanks Teja! If possible, it would be great to document the general procedure you followed to fix this in DeveloperGuides. I saw a broken API with RRTMGP and then a compat conflict with ClimaCalibrate when I tried to update the coupler in my local dev setup.

Is there any chance you still have this local dev setup in a state that results in compat issues (or some way for me to replicate this)? I think it would be easiest for me to write a guide based on a real scenario

petebachant:
I do. I'm going to head over to CliMA in a few minutes and I can show you where things are going wrong.

petebachant:
Fixed by manually updating Julia registry, since that doesn't happen for a dev install.

PR #4503, opened 2026-05-20
`noinline` quadrature point evaluation

petebachant:
This is an [~8% speedup](https://buildkite.com/clima/climaatmos-end-to-end-performance/builds/37/list?sid=019e8a56-eb0d-4352-b532-597b8c06b2dd&tab=output) for the flagship AMIP.

Closes https://github.com/CliMA/CloudMicrophysics.jl/pull/713 as it achieves the same thing without an upstream change.

Seems to be failing reproducibility tests, which is strange since no logic was changed.

## TODO

- [x] Double check performance gains after merging in `main`

PR #4541, opened 2026-06-02
Update package registry as part of coupler perf pipeline

petebachant:
From https://github.com/CliMA/ClimaAtmos.jl/pull/4503#issuecomment-4597469857


imreddyTeja (approved):
Looks like this fixes it. Thanks!

PR #4542, opened 2026-06-02
improve downgrade tests

imreddyTeja:
## Purpose 
Make the downgrade tests less likely to fail

## Content

I think the downgrade tests have been failing because they are running out of memory. The GitHub actions runners for ubuntu have [16GB](https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories) of memory. I ran the tests locally and monitored the memory usage. It grows to >16GB, even when explicitly calling the garbage collector. I'm not sure what causes this, but a few guesses are:
- maybe some files are not closing as expected
- the generated code and cached type inference builds up to a large size, particularly with tests that create a whole simulation.  I don't think this is cleared by the garbage collector, and small changes to configs can cause more type inference.  The anonymous functions used for diagnostics can also result in repeated inference, even if the config does not change.

When the memory of the Julia process got close to or over the amount of physical memory of my laptop, I noticed that simple tests took much longer than expected.

This PR makes the dowgrade action start and end a Julia process for each test group, which should reduce peak memory usage.

Unrelated to the downgrade tests:
Also updates the call to `EveryCalendarDtSchedule`. This was throwing a lot of warnings in the tests.
Also limits the number of warnings about using low resolution artifacts

----

### DIFF
--- a/.github/workflows/downgrade.yml
+++ b/.github/workflows/downgrade.yml
@@ -33,4 +33,27 @@ jobs:
           skip: Artifacts, Dates, Downloads, LazyArtifacts, LinearAlgebra, Logging, Printf, Test, Statistics, Random, Pkg, Glib_jll, Musica
 
       - uses: julia-actions/julia-buildpkg@latest
-      - uses: julia-actions/julia-runtest@latest
+      - name: Run infrastructure tests
+        uses: julia-actions/julia-runtest@latest
+        env:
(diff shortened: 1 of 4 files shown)


szy21 (approved):

ph-kev (approved):

trontrytel:
Thank you!

PR #4545, opened 2026-06-03
 Add microphysics process tendency diagnostics

trontrytel:
This PR adds the ability to output individual 1-moment microphysics process tendencies (e.g., condensation, autoconversion, accretion, melting) as diagnostic variables. All 18 source terms from `CloudMicrophysics.BulkMicrophysicsTendencies` are supported, with three variants each:

- **Grid-mean** (`mp1m_S_*`) — available for any 1M microphysics configuration
- **Updraft** (`mp1mup_S_*`) — available with prognostic EDMF
- **Environment** (`mp1men_S_*`) — available with prognostic EDMF

The 54 diagnostics are registered via a loop over source term names, avoiding boilerplate repetition.

### Design choice between recomputing vs caching

**Caching approach (rejected):**
- Call `bulk_microphysics_tendencies(InstantaneousVerbose(), ...)` once per diagnostic period
- Store the full `MicrophysicsSourceTerms` struct in a pre-allocated cache field
- Pro: single evaluation
- Con: requires allocating and managing a new cache field with 18 arrays; broadcasting the verbose function over the column returns a struct-of-arrays that must be unpacked; invasive changes to the cache infrastructure

**Recompute approach (chosen):**
- Each diagnostic independently calls `bulk_microphysics_tendencies(InstantaneousVerbose(Val(:field_name)), ...)` which returns a single scalar per grid point
- The microphysics evaluations are fast (simple arithmetic), so recomputing is cheap
- Pro: zero cache allocations, no changes to the model's cache structure, each diagnostic is self-contained
- Con: redundant computation if many tendencies are requested simultaneously

The recompute approach was chosen because it requires no changes to the model's state or cache and avoids allocation concerns. Hopefully the microphysics tendency functions are lightweight enough that recomputing them is negligible compared to I/O costs.

> [!NOTE]
> These are *instantaneous* diagnostic tendencies, not the actual linearized tendencies seen by the ODE solver. They are useful for understanding which microphysics processes are active and where, but may differ from the solver's view.

### CI integration

The three prognostic EDMF single-column cases (DYCOMS-RF02, RICO, TRMM) now output all 54 microphysics tendency diagnostics. CI plots include:
- **Vertical profile plots** — grid-mean, updraft, and environment overlaid per source term
- **Time-height contour plots** — 3-column layout (grid-mean | updraft | environment) per row

Both are saved as PDF artifacts in Buildkite CI output.

### DIFF
--- a/post_processing/ci_plots.jl
+++ b/post_processing/ci_plots.jl
@@ -190,6 +190,7 @@ function make_plots_generic(
     summary_files = String[],
     MAX_NUM_COLS = 1,
     MAX_NUM_ROWS = min(4, length(vars)),
+    fig_size = nothing,
     save_jpeg_copy = false,
     kwargs...,
 )
@@ -223,7 +224,7 @@ function make_plots_generic(
 
     # Define fig, grid, and grid_pos, used below. (Needed for scope)
     function makefig()
-        fig = CairoMakie.Figure(; size = (900, 300 * MAX_NUM_ROWS))
+        fig = CairoMakie.Figure(; size = something(fig_size, (900, 300 * MAX_NUM_ROWS)))
         if is_comparison
             for (col, path) in enumerate(output_path)
                 # CairoMakie seems to use this Label to determine the width of the figure.
@@ -1354,6 +1355,40 @@ end
 
 
 """
+    plot_mp1m_row!(grid_loc, row_vars)
+
+Plot a row of 3 mp1m time-height contour panels (grid-mean, updraft,
+environment) with shared symmetric colorbar and divergent RdBu colormap.
+
+`row_vars` is a tuple/vector of 3 OutputVar objects.
+"""
+function plot_mp1m_row!(grid_loc, row_vars)
+    # Compute shared symmetric color range across the row
+    maxabs = maximum(
+        maximum(abs.(v.data)) for v in row_vars;
+        init = eps(Float64),
+    )
+    crange = (-maxabs, maxabs)
+
+    for (col, var) in enumerate(row_vars)
+        gl = grid_loc[1, col] = CairoMakie.GridLayout()
+        viz.plot!(
+            gl,
+            var;
+            more_kwargs = Dict(
+                :axis => ca_kwargs(
+                    title = parse_var_attributes(var),
+                ),
+                :plot => ca_kwargs(
+                    colormap = Makie.Reverse(:RdBu),
+                    colorrange = crange,
+                ),
+            ),
+        )
+    end
+end
+
+"""
     plot_parsed_attribute_title!(grid_loc, var)
 
 Helper function for `make_plots_generic`. Plots an OutputVar `var`,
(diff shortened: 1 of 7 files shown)


.buildkite/Manifest-v1.11.toml
```diff
@@ -1,6 +1,6 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.11.6"
+julia_version = "1.11.9"
```
haakon-e, line 3: probably safe to revert this.

post_processing/ci_plots.jl
```diff
@@ -223,7 +224,8 @@ function make_plots_generic(
 
     # Define fig, grid, and grid_pos, used below. (Needed for scope)
     function makefig()
-        fig = CairoMakie.Figure(; size = (900, 300 * MAX_NUM_ROWS))
+        _size = isnothing(fig_size) ? (900, 300 * MAX_NUM_ROWS) : fig_size
+        fig = CairoMakie.Figure(; size = _size)
```
haakon-e:
```suggestion
        fig = CairoMakie.Figure(; size = something(fig_size, (900, 300 * MAX_NUM_ROWS)))
```

post_processing/ci_plots.jl
```diff
@@ -1353,6 +1355,89 @@ function plot_edmf_vert_profile!(grid_loc, var_group)
...
+
+    output_file = joinpath(save_path, "mp1m_timeseries.pdf")
+    if length(page_files) == 1
+        Base.Filesystem.mv(page_files[1], output_file; force = true)
+    else
+        pdfunite_args = Cmd([page_files..., output_file])
+        run(`$(pdfunite()) $pdfunite_args`)
+        Filesystem.rm.(page_files, force = true)
+    end
+    return output_file
```
haakon-e: I'm surprised you do this manually. I thought this is already done elsewhere?

post_processing/ci_plots.jl
```diff
@@ -1477,11 +1562,89 @@ function make_plots(
...
+        # since mp1m diagnostics are new and won't exist in older runs.
+        mp1m_simdirs = simdirs[1:1]
+        mp1m_output_paths = output_paths[1:1]
+
+        # Interleave as (grid-mean, updraft, environment) per source term
+        # so each row in the contour plot shows one process across domains
+        mp1m_all_names = vcat(
+            [["mp1m_$(s)", "mp1mup_$(s)", "mp1men_$(s)"]
+             for s in mp1m_source_names]...,
+        )
```
haakon-e:
```suggestion
        mp1m_all_names = vec(["mp1m_", "mp1mup_", "mp1men_"] .* permutedims(mp1m_source_names))
```

src/diagnostics/microphysics_diagnostics.jl
```diff
@@ -0,0 +1,206 @@
...
+        units = "kg kg^-1 s^-1",
+        comments = "Instantaneous source term $(field) from " *
+                   "bulk_microphysics_tendencies(InstantaneousVerbose()). " *
+                   "Evaluated at environment state.",
+        compute = let F = Val(field)
+            (state, cache, time) ->
+                compute_mp1m_source_env(state, cache, time, F)
+        end,
+    )
+end
```
haakon-e:
```suggestion
for (prefix, label, location, compute_fn) in (
    ("mp1m",   "",             "grid-mean state",                 compute_mp1m_source),
    ("mp1mup", " Updraft",     "updraft state (first subdomain)", compute_mp1m_source_updraft),
    ("mp1men", " Environment", "environment state",               compute_mp1m_source_env),
)
    for (field, description) in MP1M_SOURCE_TERMS
        add_diagnostic_variable!(;
            short_name = "$(prefix)_$(field)",
            long_name = "1M Microphysics$(label): $(description)",
            units = "kg kg^-1 s^-1",
            comments = "Instantaneous source term $(field) from " *
                       "bulk_microphysics_tendencies(InstantaneousVerbose()). " *
                       "Evaluated at $(location).",
            compute = let F = Val(field), f = compute_fn
                (state, cache, time) -> f(state, cache, time, F)
            end,
        )
    end
end
```

haakon-e (approved):
LGTM

Left some purely stylistic suggestions. Take or leave as you like.

szy21 (approved):
This is the tendency without the quadrature, right?

trontrytel:
> This is the tendency without the quadrature, right?

Yes, thats without the quadratures. And not taking in the timestepper stages or the linearization done in the microphysics solve either. You pass in the updraft, environment or grid mean state and you get the individual microphysics process rates as is. 

It is intended to help debug and get an idea of what process is driving, but won't be super accurate to do any budget analysis on mass, or it won't spot a NaN on a quadrature point. I think it's still useful to add, but probably not as useful as an actual tendency that the model feels

.buildkite/Manifest-v1.11.toml
trontrytel: Actually the 11.6 was a mistake. We should all be back to the future of 11.9

post_processing/ci_plots.jl
trontrytel: Thanks! We had to do some parts of it manually because the tendency timeseries are plotted in 3 columns for the grid-mean, updraft and environment. But at least the file management part can be re-sued more

src/diagnostics/microphysics_diagnostics.jl
trontrytel: Thanks! Thats much cleaner

PR #4552, opened 2026-06-05
Split up 1m microphysics tendency cache

petebachant:
Should be another 7% speedup for flagship AMIP.


petebachant:
This change is now moving into CloudMicrophysics, so I'm going to close this.

PR #4556, opened 2026-06-08
add common diagnostic and modify edmf plots

szy21:
## Purpose 

Make EDMF ci plots more helpful. Now they contain:

- Profiles of grid-mean and edmf variables averaged over the last hour
- height-time contours of grid-mean and edmf variables
- height-time contours of microphysics tendencies
- Timeseries of column integrated quantities (e.g. lwp)

Also adds some common diagnostics configs for single column EDMF, so the configs are shorter:)

The ref counter is bumped because new plots are added. There is no actual behavior change.

## Content

----

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -158,6 +158,7 @@ steps:
         retry: *retry_policy
         command: >
           julia --color=yes --project=.buildkite .buildkite/ci_driver.jl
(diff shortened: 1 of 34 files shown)


costachris (approved):
Can you just add `turbentr` to the output as well?

nefrathenrici (approved):

trontrytel (approved):

PR #4557, opened 2026-06-08
ci: Add PR preview links as PR comment

haakon-e:
Add a comment step to post preview links for pull requests. This is intended to be a more visible and easier to find alternative to scrolling through the (rather long) list of checks at the bottom of the PR.


haakon-e:
@nefrathenrici intended as a small convenience for contributors. For whatever reason, the docs don't build successfully for this PR, so the link in the comment appears not to actually work.

nefrathenrici (approved):
Thank you, I think this will especially make it easier for new contributors to find doc previews. Ideally this would only run if the docs are changed in the PR, but that can be done later.

PR #4559, opened 2026-06-09
Trim all trailing whitespace, format docstrings

imreddyTeja:
This applies formatting repo wide.

First, it sets `format_docstrigns` to true in the `JuliaFormatter.yml`.
Next, it applies just the Julia formatter repo wide.
Then it applies the trailing whitespace trim repo wide. I think all of the trimmed whitespaces are ok, except for whitespaced trimmed from the agent workflow lock file. I excluded that.
Adds a git-blame-ignore-rev. Not sure where a good place to tell users they should set `git config blame.ignoreRevsFile .git-blame-ignore-revs`

----


petebachant:
Trailing whitespace does technically have a meaning in the original Markdown spec ([ref](https://github.com/pre-commit/pre-commit-hooks/issues/57)), so it may be worth checking if any of these are important for the doc build process (maybe Julia docstrings make use of trailing whitespace too). Beyond that, it is possible to define this Git rev as one that should be ignored for `git blame`: https://thinkthroo.com/blog/git-blame-ignore-revs

haakon-e:
Ah! The never-ending struggle between the camp of people that don't format whitespace and those whose editor does so automatically 😄

I think we can use JuliaFormatter to format docstrings and markdown files ([link](https://juliaeditorsupport.github.io/JuliaFormatter.jl/stable/formatting_options/#options-format-docstrings)). Yaml files seem to be edited infrequently enough that it might not matter?

haakon-e (approved):

imreddyTeja:
> Trailing whitespace does technically have a meaning in the original Markdown spec ([ref](https://github.com/pre-commit/pre-commit-hooks/issues/57)), so it may be worth checking if any of these are important for the doc build process (maybe Julia docstrings make use of trailing whitespace too). Beyond that, it is possible to define this Git rev as one that should be ignored for `git blame`: https://thinkthroo.com/blog/git-blame-ignore-revs

good point. I looked through each changed line, and ~~I think everything should be OK~~ (other than the agent lock file, which I reverted). 

I didn't know that git-blame-ignore-revs is a thing. If my understanding is correct, users would need to set their local git config to use this by default. Should I document this somewhere?

ph-kev (approved):

ph-kev:
It might be worthwhile to use a Markdown formatter to format all the markdown files as well.

haakon-e:
Additional config for markdown files ([link](https://juliaeditorsupport.github.io/JuliaFormatter.jl/stable/file_options/#options-format-markdown)):
```yaml
format_markdown: true
```
Uses [CommonMark.jl](https://github.com/MichaelHatherly/CommonMark.jl)

petebachant:
> I didn't know that git-blame-ignore-revs is a thing. If my understanding is correct, users would need to set their local git config to use this by default. Should I document this somewhere?

Maybe in the dev guides somewhere. Could also put a comment up at the top of the file explaining how to use it.

PR #4572, opened 2026-06-10
JuliaFormatter v2 + markdown formatting

imreddyTeja:
## TODO
remove required Julia-fromatter action
I should follow up this PR with an update to dev-guides, or figure out a way to modify the version in this repo.

## Content
- applies Julia-fromatter's markdown formatter to most .md files (excludes NEWS, README, ad dev-guides"
- also fixes two code blocks that cause the parser to error
- adds ignores files in .github and dev-docs
- updates JuliaFormatter version

~~The trailing whitespace remover used in the pre-commit hook is at odds with the JuliaFormatter's markdown style. I made the trailing whitespace hook ignore markdown files. They disagree on if empty lines should be empty or have four spaces~~ edit: fixed with new version of juliaformatter

----


imreddyTeja:
@petebachant @ph-kev Do you have any thoughts on the conflict between trailing-whitespaces and JuliaFormatter? I went with JuliaFormatter here because its markdown formatting is more comprehensive. There is probably a way to configure JuliaFormatter to agree with trailing-whitespaces, but I'm not sure it is worth the effort

ph-kev:
I think the simplest thing to do is to use the Markdown formatting from JuliaFormatter. I don't have a strong preference using trailing-whitespaces versus JuliaFormatter for formatting the markdown files.

petebachant:
I find the Markdown formatting a little weird, e.g., indenting list items, but what really matters is that whatever system is put in place prevents superfluous formatting changes from making it into PRs. Might want to add running `prek` to the formatter Actions workflow so we're not depending on devs to format on their own.

.pre-commit-config.yaml
```diff
@@ -12,11 +12,22 @@ repos:
...
+      - id: check-case-conflict
+      - id: check-merge-conflict
+      - id: check-symlinks
+      - id: check-toml
+      - id: check-yaml
+        exclude: .*lock\.yml
+      - id: end-of-file-fixer
+        exclude: .*lock\.yml
+      - id: mixed-line-ending
+      # JuliaFormatter does not strip all trailing whitespace (e.g. in  non-.jl files),
```
petebachant: Nit:

```suggestion
      # JuliaFormatter does not strip all trailing whitespace (e.g., in non-.jl files),
```

.git-blame-ignore-revs
```diff
@@ -4,7 +4,5 @@
 # To set your local git to use this file by default, run:
 # 'git config blame.ignoreRevsFile .git-blame-ignore-revs'
 ##########################################################
-# Formatting docstrings
-493f52f7c1367667cdd692ec41566a0f3a0da017
-# trimming whitespaces repo wide, all files expect lock files
-2b8910c166619cbe720771c5b6376de102dd0bae
+# Formatting docstrings amd trimming whitespaces repo wide, all files expect lock files
```
petebachant:
```suggestion
# Formatting docstrings and trimming whitespace repo wide, all files except lock files
```

calibration/README.md
```diff
@@ -8,15 +8,15 @@ This directory contains the model interface (`model_interface.jl`) for ClimaCali
...
+  - Configuration: equilmoist, 0-moment microphysics scheme, spherical with Held-Suarez forcing.
+  - Observational data: zonal and 60-day time average air temperature at 242m altitude. Provided by [ClimaArtifacts](https://github.com/CliMA/ClimaArtifacts/tree/main/atmos_held_suarez_obs).
+  - Parameter being calibrated: `equator_pole_temperature_gradient_wet`
 
 For more details, see the experiment directory.
 
 To reproduce the results, on the Caltech central cluster run
 
-```sbatch calibration/experiments/sphere_held_suarez_rhoe_equilmoist/pipeline.sbatch```
+`sbatch calibration/experiments/sphere_held_suarez_rhoe_equilmoist/pipeline.sbatch`
```
petebachant:
````suggestion
```sh
sbatch calibration/experiments/sphere_held_suarez_rhoe_equilmoist/pipeline.sbatch
```
````

**Review by petebachant — APPROVED** [review]
A few minor nitpicks. Note that you'll need to change the main branch ruleset requiring the `format` check to pass and replace with the new workflow.

JuliaFormatter still feels pretty weird on Markdown TBH--maybe worth trying something more conventional, e.g.,

```yaml
  - repo: https://github.com/executablebooks/mdformat
    rev: 0.7.21  # Use the latest stable version
    hooks:
      - id: mdformat
        # Optional: Add plugins if you use GitHub Flavored Markdown (GFM)
        additional_dependencies:
          - mdformat-gfm
```

and disabling JuliaFormatter Markdown handling (`format_markdown = false`?), and re-enabling trailing whitespace trimming for Markdown files. But this is getting into the weeds a bit.

petebachant:
Maybe worth updating `.dev/format/Project.toml` in this PR too. Can we update to v2, update the comments in there, etc.?

docs/src/setups.md
```diff
@@ -130,8 +130,13 @@ Add a branch in `get_setup_type` in `src/config/type_getters.jl` that maps
 the `initial_condition` config string to your setup constructor:
 
 ```julia
+if ic_name == "OtherCase1"
+    return Setups.OtherCase1()
+elseif ic_name == "OtherCase1"
+    return Setups.OtherCase1()
 elseif ic_name == "MyCase"
     return Setups.MyCase()
+end
```
ph-kev, line 139: I think the conditions for the `if` and first `elseif` statements are the same?

src/cache/cache.jl
```diff
@@ -19,52 +19,80 @@ struct AtmosCache{
     SSV,
     CONSCHECK,
 }
-    """Timestep of the simulation (in seconds). This is also used by callbacks and tendencies"""
+    """
+    Timestep of the simulation (in seconds). This is also used by callbacks and tendencies
+    """
```
szy21, line 24: I like the previous version better as the comment is shorter (one line instead of three lines). I don't have a strong preference though. Is this from the formatter version change?

haakon-e: This is what the markdown formatter does, I think

haakon-e:
Could you rebase the commits so one applies Julia formatter rules and any necessary file code changes, and one applies the markdown formatter?

README.md
```diff
@@ -1,118 +1,65 @@
...
 <p align="center">
   <img src="https://github.com/user-attachments/assets/778b0c14-a5d7-4907-82db-6d1f8a0c5b07" alt="animation (1)">
 </p>
-
 Condensed water path from a global simulation using diagnostic EDMF and 0M microphysics, initialized with ERA5 on 8-31-25 00Z. Output every 30 minutes; ran for ~4 days.
-
 ## Installation instructions
-
 Recommended Julia: Stable release v1.11.6
-
```
petebachant, line 56: Won't removing these blank lines make these all render as one paragraph? Could put `remove_extra_newlines = false` in `.JuliaFormatter.toml`, or do some other hack like backslashes, but again, I think formatting Markdown with something else would make it more in line with typical Markdown conventions, especially since this is being rendered on GitHub.

.dev/format/Project.toml
```diff
@@ -1,13 +1,7 @@
...
-# in .github/workflows/julia_formatter.yml (`version: '1'`), independent of
-# whatever JuliaFormatter version lives in your base environment (v2 is now the
-# default `Pkg.add` result and produces a different diff). Bump this compat
-# entry in lockstep with the CI `version:` input.
 [deps]
 JuliaFormatter = "98e50ef6-434e-11e9-1051-2b60c6c9e899"
 
 [compat]
-JuliaFormatter = "1"
+JuliaFormatter = "=2.8.3" # bump version assert in .pre-commit-config-yaml if changing this
```
ph-kev: It seems strange to me that we are using this Project.toml when in the past, it seems that this file was deprecated in favor of keeping JuliaFormatter in your base environment. Does the developer guides have its own Project.toml file to use and can we get rid of this?

README.md
imreddyTeja: good catch. I'm going to mark this PR as do not merge for now

.dev/format/Project.toml
petebachant: Do you know the reasoning behind keeping in the base environment? Most workflows have been moving in the opposite direction--dev environments defined at the project level rather than the system level.

I would drop the version assertion though. That seems unnecessary.

ph-kev: Maybe the terminology is confusing me, but when I said base environment, I am referring to the environment at `/home/USER/.julia/environments/v1.12`. The reasoning for why the `.dev` environment was removed in https://github.com/CliMA/ClimaAtmos.jl/pull/3546. I would guess that the reasoning behind keeping JuliaFormatter in the base environment is that it is more in line with what other Julia packages do and there are less environments to maintain.

ph-kev (commented):
This [Formatting section](https://clima.github.io/ClimaAtmos.jl/dev/contributor_guide/#Formatting) in the contributor's guide should be updated. Also, it seems that there are malformed note admonitions and TODOs that should be done.

.dev/format/Project.toml
petebachant: Yeah, that's what I mean by the base environment too. The problem is that the base environment is a system-level environment, so you end up telling users they need to maintain their system environment in a way that's compatible with all packages they work on. Maybe okay for a dev who works on one or two packages, but it's not scalable. When developing Python or Node.js packages, for example, dev dependencies are declared separately so their versions can be specific to the project being worked on.

In other words, removing the `.dev` environment doesn't remove the need to maintain an environment. It just makes it harder because then you need to coordinate all developers to get them keep a similar base environment, and that doesn't scale well beyond a few people/projects.

Now, if CliMA considers many packages to be part of the same logical "project" (or product), and that they all must be formatted identically, that's another reason to merge them together, because then you could have one dev environment for the whole thing, and not need complex infrastructure or coordination through developer rules to keep them tightly coupled yet spread apart.

TL;DR: Separate repos --> separate products --> separate dev dependencies --> loose coupling

[Relevant thread on Julia forums](https://discourse.julialang.org/t/is-there-concept-of-dev-dependencies-in-pkg/46908)

.JuliaFormatter.toml
```diff
@@ -5,3 +5,5 @@ whitespace_typedefs = true
 whitespace_ops_in_indices = true
 join_lines_based_on_source = true  # https://domluna.github.io/JuliaFormatter.jl/dev/#join_lines_based_on_source
 format_docstrings = true
+format_markdown = true
+ignore = ["docs/dev-guides", ".github", "**/README.md", "NEWS.md"]
```
petebachant: Maybe want to ignore `AGENTS.md` too?

```suggestion
ignore = ["docs/dev-guides", ".github", "**/README.md", "NEWS.md", "AGENTS.md"]
```

Should README, NEWS, and AGENTS get their own pre-commit formatter additionally?

calibration/experiments/perfect_scm/model_interface/model_interface.jl
```diff
@@ -31,13 +31,19 @@ analysis of the iteration is done by `ClimaCalibrate.analyze_iteration`.
 The calibration is saved in `output_dir`.
 """
 struct PerfectAtmosModelInterface{DIAGS <: Vector} <: ClimaCalibrate.AbstractModelInterface
-    "A filepath to a configuration file for a ClimaAtmos simulation"
+    """
+    A filepath to a configuration file for a ClimaAtmos simulation
+    """
```
ph-kev: Is the kind of formatting that we want for struct definitions?

docs/src/contributor_guide.md
```diff
@@ -284,11 +295,11 @@ documentation for more information.
 
 !!! note
 
+
 In the past, `ClimaAtmos` used to have a `.dev/up_deps.jl` script. We moved away
 from it because `PkgDevTools` provides a much simpler and more efficient way to
 accomplish the same result.
```
ph-kev: This admonition is malformed.

docs/src/contributor_guide.md
```diff
@@ -187,31 +190,40 @@ use [JuliaFormatter.jl](https://github.com/domluna/JuliaFormatter.jl) to achieve
...
+```sh
 alias julia_format_here="julia -e 'using JuliaFormatter; format(\".\")'"
 ```
 
 !!! note
 
+
 In the past, `ClimaAtmos` used to have a `.dev/climaformat.jl` script. We moved
 away from it to reduce complexity in our repository and to align with the
 general tools used by the Julia community. If you are still using
```
ph-kev: This admontion is also malformed.

calibration/experiments/perfect_scm/model_interface/model_interface.jl
imreddyTeja: No, it is not. I can disable formatting for this struct if you think this formatting is bad enough. I don't think there is a way to have single line docstrings and use the doctoring formatting feature

haakon-e: This is a reasonably common pattern when using `DocStringExtensions.jl` to auto-document struct fields ([link](https://docstringextensions.juliadocs.org/stable/showcase/#Types)). Though it's often more useful imo to manually curate the docstring than rely to the auto-generated form.

In the case where `$(DocStringExtensions.FIELDS)` is not used, as is the case here, one might as well just document the purpose of these (private)[^1] fields with regular comments, which circumvents this issue altogether.
```julia
# A filepath to a configuration file for a ClimaAtmos simulation`
config::String
```

[^1]: it's a recurring topic of discussion in the Julia community whether struct fields should be considered private or public by default. Documenting them in user-facing API would potentially indicate to users that they are considered public.

PR #4579, opened 2026-06-12
reduce ci duration for autodiff baroclinic wave

imreddyTeja:
## Purpose 
buildkite ci takes two hours because the baroclinic wave dense autodiff test times out after an hour, and then retries.

These tests are mainly used to test infrastructure, not numerics, so they do not need to be long.

----


szy21 (approved):
@trontrytel do you want to just combine this with your PR?

szy21:
I'll merge this as Anna's PR need more work.

PR #4590, opened 2026-06-16
Non-dimensionalize variance floor parameter in CF

tapios:
## Purpose 
Replace the hardcoded `σ_S_fix = 1e-6` in the truncated-Gaussian cloud-fraction closure with a scale-aware floor `σ_S_fix = sqrt((ε_rel · q_sat)² + σ_abs²)`, proportional to local saturation humidity. This prevents the absolute threshold from over-suppressing cirrus where `q_sat` is small.

## Content

----

### DIFF
--- a/src/parameterized_tendencies/microphysics/cloud_fraction.jl
+++ b/src/parameterized_tendencies/microphysics/cloud_fraction.jl
@@ -346,7 +346,7 @@ end
 # where Φ is the standard normal CDF and φ is its PDF.
 #
 # For the *cloud fraction* we use an augmented variance with a fixed
-# non-equilibrium floor `σ_S_floor` (hardcoded inside
+# non-equilibrium floor `σ_S_floor` (defined inside
 # `_compute_cloud_fraction`):
 #
 #     σ_aug = α · sqrt(σ_S² + σ_S_floor²),
--- a/src/parameterized_tendencies/microphysics/cloud_fraction.jl
+++ b/src/parameterized_tendencies/microphysics/cloud_fraction.jl
@@ -575,8 +575,6 @@ NVTX.@annotate function set_sgs_moments_and_cloud_fraction!(Y, p)
     # with a value consistent with the final SGS moments, so EDMF weighting
     # must be re-applied here even though `set_cloud_fraction!` already
     # applied it during Picard.
-    # q_sat is computed inline here (not cached) — it's cheap and avoids
-    # adding a field to the ᶜsgs_moments named tuple.
     @. p.precomputed.ᶜcloud_fraction = _compute_cloud_fraction(
         ᶜq_lcl + ᶜq_icl,
         p.precomputed.ᶜsgs_moments.sigma_S,
--- a/src/parameterized_tendencies/microphysics/cloud_fraction.jl
+++ b/src/parameterized_tendencies/microphysics/cloud_fraction.jl
@@ -287,7 +287,7 @@ end
 @inline function (eval::SGSMomentsEvaluator)(T_hat, q_tot_hat)
     q_sat_hat = TD.q_vap_saturation(eval.tps, T_hat, eval.ρ)
     s = q_tot_hat - q_sat_hat
-    return (mu_S = s, s_sq = s * s)
+    return (mu_S = s, s_sq = s * s, s_cube = s^3)
 end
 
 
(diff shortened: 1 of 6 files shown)


src/parameterized_tendencies/microphysics/cloud_fraction.jl
```diff
@@ -318,6 +321,7 @@ Returns `(; mu_S, sigma_S)`.
     return (;
         mu_S = raw.mu_S,
         sigma_S = max(sqrt(sigma_S_sq), ϵ_numerics(FT)),
+        q_sat = TD.q_vap_saturation(thp, T_mean, ρ),
```
sajjadazimi: `q_sat` doesn't need to be cached. It's a function of `(T_mean, ρ)`, that are available in the cloud fraction function. We can compute inline to avoids a new cache field.

src/parameterized_tendencies/microphysics/cloud_fraction.jl
```diff
@@ -489,19 +509,22 @@ materialized to a Field.
     moments = _sgs_saturation_moments(
         thermo_params, ρ, T, q_tot, sgs_quad, T′T′, q′q′, corr_Tq,
     )
-    return _compute_cloud_fraction(q_liq + q_ice, moments.sigma_S, α)
+    return _compute_cloud_fraction(
+        q_liq + q_ice, moments.sigma_S, moments.q_sat, α,
+    )
```
sajjadazimi, line 510: We can compute `q_sat` locally here instead of using the cached value (unless we want the mean q_sat instead of q_sat of mean).

sajjadazimi (approved):
Thanks! I left a comment about caching `q_sat`. Also, I think eps_rel = 0.15 might be too large. With `q_sat ~1e-2` in the BL, `sig_S_floor \approx 0.0015` dominates `sigma_S` and significantly decreases CF (see DYCOMS in CI). Since we already carry `(T'T', q'q')` and with explicit EDMF drafts, the floor should probably model only the residual SGS misses. I suggest starting with 0.01~0.02 and then we can calibrate. Alternatively, we can compute sigma_aug as an alpha-weighted average of sigma_S and sigma_S_Quass. So if alpha -> 0 microphysics uses uniform condensate and CF sees sigma_S_floor and when alpha -> 1 microphysics and CF both see the full equilibrium variability. Then we can probably land on alpha ~ 0.9 (and then sigma_abs should move out of sigma_Quass and be applied as a pure numerical thing).

src/parameterized_tendencies/microphysics/cloud_fraction.jl
tapios: Good point. I changed it.

src/parameterized_tendencies/microphysics/cloud_fraction.jl
tapios: Done.

tapios:
> Thanks! I left a comment about caching `q_sat`. Also, I think eps_rel = 0.15 might be too large. With `q_sat ~1e-2` in the BL, `sig_S_floor \approx 0.0015` dominates `sigma_S` and significantly decreases CF (see DYCOMS in CI). Since we already carry `(T'T', q'q')` and with explicit EDMF drafts, the floor should probably model only the residual SGS misses. I suggest starting with 0.01~0.02 and then we can calibrate. Alternatively, we can compute sigma_aug as an alpha-weighted average of sigma_S and sigma_S_Quass. So if alpha -> 0 microphysics uses uniform condensate and CF sees sigma_S_floor and when alpha -> 1 microphysics and CF both see the full equilibrium variability. Then we can probably land on alpha ~ 0.9 (and then sigma_abs should move out of sigma_Quass and be applied as a pure numerical thing).

All good points. Thank you!

I lowered eps_rel and removed the cached q_sat. While a lower eps_rel is clearly needed for the DYCOMS comparisons, I wonder if we might need a larger one at the coarser GCM resolutions. I hope calibration can help resolve this. (@szy21 ?)

The alpha-weighted average is also reasonable. We would need to bound alpha < 1 not to get the previous singular behavior in the alpha -> 1 limit. A compromise maybe something like `σ_aug = √((α · σ_S)² + σ_S_Quaas²)`. I'll leave it to you to experiment with these choices in a later PR.

szy21:
> I lowered eps_rel and removed the cached q_sat. While a lower eps_rel is clearly needed for the DYCOMS comparisons, I wonder if we might need a larger one at the coarser GCM resolutions. I hope calibration can help resolve this. (@szy21 ?)
> 

Sure, happy to calibrate more parameters!

tapios:
> > I lowered eps_rel and removed the cached q_sat. While a lower eps_rel is clearly needed for the DYCOMS comparisons, I wonder if we might need a larger one at the coarser GCM resolutions. I hope calibration can help resolve this. (@szy21 ?)
> 
> Sure, happy to calibrate more parameters!

@szy21 Could you please have a quick look at the Buildkite output and see if this looks reasonable to you? DYCOMS is back to 100% after the parameter changes, but I do not have a good mental baseline for these cases.

szy21:
> > > I lowered eps_rel and removed the cached q_sat. While a lower eps_rel is clearly needed for the DYCOMS comparisons, I wonder if we might need a larger one at the coarser GCM resolutions. I hope calibration can help resolve this. (@szy21 ?)
> > 
> > 
> > Sure, happy to calibrate more parameters!
> 
> @szy21 Could you please have a quick look at the Buildkite output and see if this looks reasonable to you? DYCOMS is back to 100% after the parameter changes, but I do not have a good mental baseline for these cases.

Dycoms rf02 looks a little worse (cloud fraction is ~80% rather than 100%) but I think we can calibrate it. Other cases look fine. I think we can merge it.

PR #4591, opened 2026-06-16
Delay WallTimeInfo until cbs are compiled

imreddyTeja:
Delay estimating sypd until either a quarter of the tspan is finished, or all non-post-simulation callbacks are called at least once.

We don't need to merge this if people prefer the old behavior.
I excluded the nan check callback because the default period is 1024 steps.

## Content

----


src/callbacks/get_callbacks.jl
```diff
@@ -519,30 +524,42 @@ function common_callbacks(
...
+    callbacks = (
+        callbacks...,
+        checkpoint_callback(checkpoint_frequency, output_dir, start_date, t_start)...,
+    )
     # Progress logging
     if log_progress
-        callbacks = (callbacks..., progress_logging_callback(dt, t_start, t_end)...)
+        # avoid estimating SYPD until the above (non-cleanup cbs) callbacks are compiled.
+        # raw SYPD between each call is still reported
+        steps_per_cycle = n_steps_per_cycle(callbacks, dt)
```
nefrathenrici, line 540: I'm not sure that this changes progress logging in the intended way. I think `steps_per_cycle` will always equal 1 since the function `n_steps_per_cycle` only sees the conservation callback, resulting in `sypd_estimate_step_delay = 10`. To fix I think we need to call `n_steps_per_cycle` on the full list of callbacks.

PR #4593, opened 2026-06-16
Add "explicit" Bomex with passive tracer A

trontrytel:
This PR adds a passive tracer `A` to the atmospheric model (both grid-mean and EDMF). All SGS tendencies (entrainment, advection, diffusion, mass flux) are auto-discovered via `sgs_tracer_names(Y)`. Grid-mean tendencies (advection, diffusion, hyperdiffusion) are auto-discovered via `gs_tracer_names(Y)` / `foreach_gs_tracer`. The Jacobian is still hard-coded. To be changed in future PRs.

To switch it on set `chemistry_model: "passive"`. As an example, the PR adds an SCM Bomex with passive tracer to the CI. It's switched off by default, so should not affect any of our typical simulations.

I think it's a good test of our sgs vertical transport. So I'm adding it to the CI.

### Chemistry model (`chemistry.jl`)
- New `GasPhaseChem <: AbstractChemistryModel` type, selected via `chemistry_model: "passive"` in the YAML config.

### Physical state (`physical_state.jl`)
- Added `A` field (default `zero(T)`) to the `physical_state` NamedTuple. Present in all setups but only added to the Atmos model state when chemistry is active.

### Prognostic variables (`prognostic_variables.jl`)
- `chemistry_variables(ρ, ps, ::AbstractChemistryModel)` → adds `ρA` to grid-scale state.
- `chemistry_sgs_variables(ps, ::AbstractChemistryModel)` → adds `A` to updraft `sgsʲs`.
- Both return `(;)` when chemistry model is `nothing`, so no impact on existing configs.

### Initial condition (`Bomex.jl`)
- Initializes `A = 1` in the Bomex physical state (a uniform background for testing transport).

### Jacobian (`manual_sparse_jacobian.jl`)
- `ρA` gets a `-I` identity block (explicit treatment).
- `sgsʲs.:(1).A` gets Jacobian entries for SGS vertical advection (TridiagonalRow) and entrainment (Diagonal), matching the implicit treatment of other SGS scalars.
- All entries rely on `MatrixFields.has_field(Y, @name(...))`.

### Diagnostics (`core_diagnostics.jl`, `edmfx_diagnostics.jl`)
- `passiveA`: grid-mean specific concentration `ρA/ρ`.
- `passiveAup`: updraft concentration `sgsʲs.:(1).A`.

### CI (`pipeline.yml`)
- New single-column job using `prognostic_edmfx_bomex_tracerA_column.yml`

### DIFF
--- a/config/model_configs/prognostic_edmfx_bomex_tracerA_column.yml
+++ b/config/model_configs/prognostic_edmfx_bomex_tracerA_column.yml
@@ -1,0 +1,73 @@
+initial_condition: "Bomex"
+turbconv: "prognostic_edmfx"
+implicit_diffusion: true
+approximate_linear_solve_iters: 2
+edmfx_entr_model: "Generalized"
+edmfx_detr_model: "Generalized"
+edmfx_sgs_mass_flux: true
+edmfx_sgs_diffusive_flux: true
+edmfx_nh_pressure: true
+edmfx_vertical_diffusion: true
+edmfx_filter: true
+prognostic_tke: true
+microphysics_model: "1M"
+cloud_model: "MLCloud"
+config: "column"
+z_max: 3e3
+z_elem: 60
+z_stretch: false
+perturb_initstate: false
+dt: "1secs"
+t_end: "6hours"
+dt_save_state_to_disk: "10mins"
+toml: [toml/prognostic_edmfx_bomex.toml]
+netcdf_interpolation_num_points: [2, 2, 60]
+ode_algo: "ARS222"
+chemistry_model: "passive"
+diagnostics:
+  - short_name: [ta, thetaa, ha, pfull, rhoa, ua, va, wa, hur, hus, cl, clw, cli, husra, hussn]
+    period: 10mins
+  - short_name: [ts, hfss, hfls, pr, lwp, clivi, rwp, swp]
+    period: 10mins
+  - short_name: [arup, waup, taup, thetaaup, haup, husup, hurup, clwup, cliup, husraup, hussnup, waen, taen, thetaaen, haen, husen, huren, clwen, clien, tke, lmix]
+    period: 10mins
+  - short_name: [entr, detr, turbentr, bgrad, strain, edt, evu]
+    period: 10mins
+  - short_name: [q_gas_A, q_gas_Aup]
+    period: 10mins
+  - short_name: [husraup, hussnup, q_gas_A, q_gas_Aup]
+    period: 1hours
+    reduction_time: average
+  - short_name: [
+      mp1m_S_phase_change_vap_lcl, mp1m_S_phase_change_vap_icl,
+      mp1m_S_acnv_lcl_rai, mp1m_S_acnv_icl_sno,
+      mp1m_S_accr_lcl_rai, mp1m_S_accr_lcl_sno_cold, mp1m_S_accr_lcl_sno_warm,
+      mp1m_S_accr_melt_lcl_sno,
+      mp1m_S_accr_icl_rai, mp1m_S_accr_freeze_icl_rai, mp1m_S_accr_icl_sno,
+      mp1m_S_accr_rai_sno_cold, mp1m_S_accr_rai_sno_warm, mp1m_S_accr_melt_rai_sno,
+      mp1m_S_phase_change_vap_rai, mp1m_S_phase_change_vap_sno,
+      mp1m_S_melt_icl_lcl, mp1m_S_melt_sno_rai,
+    ]
+    period: 10mins
+  - short_name: [
+      mp1mup_S_phase_change_vap_lcl, mp1mup_S_phase_change_vap_icl,
+      mp1mup_S_acnv_lcl_rai, mp1mup_S_acnv_icl_sno,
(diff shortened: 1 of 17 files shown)


config/model_configs/prognostic_edmfx_bomex_tracerA_column.yml
```diff
@@ -0,0 +1,73 @@
...
+  - short_name: [
+      mp1m_S_phase_change_vap_lcl, mp1m_S_phase_change_vap_icl,
+      mp1m_S_acnv_lcl_rai, mp1m_S_acnv_icl_sno,
+      mp1m_S_accr_lcl_rai, mp1m_S_accr_lcl_sno_cold, mp1m_S_accr_lcl_sno_warm,
+      mp1m_S_accr_melt_lcl_sno,
+      mp1m_S_accr_icl_rai, mp1m_S_accr_freeze_icl_rai, mp1m_S_accr_icl_sno,
+      mp1m_S_accr_rai_sno_cold, mp1m_S_accr_rai_sno_warm, mp1m_S_accr_melt_rai_sno,
+      mp1m_S_phase_change_vap_rai, mp1m_S_phase_change_vap_sno,
+      mp1m_S_melt_icl_lcl, mp1m_S_melt_sno_rai,
+    ]
```
haakon-e, line 50: I think someone added some other diagnostics config that could be included in this config, so that this (and other) long lists could be avoided?

post_processing/ci_plots.jl
```diff
@@ -1480,6 +1482,10 @@ function make_plots(
     period_inst =
         available_periods_inst[argmin(CA.time_to_seconds.(available_periods_inst))]
 
+    short_names = filter(
+        n -> n in ClimaAnalysis.available_vars(simdirs[1]),
+        short_names,
+    )
```
haakon-e: could you move this code up up where `short_names` is defined for clarity.

src/parameterized_tendencies/chemistry/chemistry.jl
```diff
@@ -11,3 +11,12 @@
 No-op: no chemistry model is active.
 """
 chemistry_tendency!(Yₜ, Y, p, t, ::Nothing) = nothing
+
+"""
+    chemistry_tendency!(Yₜ, Y, p, t, ::GasPhaseChem)
+
+No-op for passive tracers: transport is handled by the auto-discovery
+machinery (`gs_tracer_names` / `sgs_tracer_names`), and there are no
```
haakon-e: instead of "auto-discovery machinery" jargon, could you point to some relevant docs that explain this; or cut?

haakon-e (approved):
Looks like a reasonable first pass!

My main concern with this is the chosen name `A`.
Reading this PR diff, it's very clear what `A` is, but if you stumble upon it in the code, and ask the same question, I think it may be much harder to intuit about. I would prefer a clearer name, such as `passive_tracer`.

src/cache/temporary_quantities.jl
```diff
@@ -16,9 +16,11 @@ function implicit_temporary_quantities(Y, atmos)
     return (;
         ᶠtemp_scalar = Fields.Field(FT, face_space), # ᶠρaK_h
         ᶠtemp_scalar_2 = Fields.Field(FT, face_space), # ᶠρaK_u
+        ᶠtemp_scalar_3 = Fields.Field(FT, face_space), # ᶠwaχ in advection
         ᶜtemp_scalar = Fields.Field(FT, center_space), # ᶜρχₜ_diffusion, ᶜa_scalar
         ᶜtemp_scalar_2 = Fields.Field(FT, center_space), # ᶜKᵥʲ
         ᶜtemp_scalar_3 = Fields.Field(FT, center_space), # ᶜK_h_scaled
+        ᶜtemp_scalar_4 = Fields.Field(FT, center_space), # ᶜq_icl (in surface conditions)
```
nefrathenrici, line 23:
```suggestion
        ᶜtemp_scalar_4 = Fields.Field(FT, center_space), # ᶜvtt (sgs tracer advection)
```
nitpick :)

nefrathenrici (approved):
Looks good to me. Should we add a conservation test or will that come later?

trontrytel:
> My main concern with this is the chosen name `A`. 

This name is appropriate from the point of view of the example "ABC" chemistry reaction we are trying to add here. But I agree that from the non-chemistry view point it'll not be as obvious. I'll check with @obin1 to see what would be a better naming convention for his project

config/model_configs/prognostic_edmfx_bomex_tracerA_column.yml
trontrytel: I thought merging the two yaml configs would replaces the `diagnostics`. What I need here is to add the tracerA to the 1M microphysics list.

src/cache/temporary_quantities.jl
trontrytel: It came up when trying to visualize the Jacobian. And it was erroring from this line: https://github.com/CliMA/ClimaAtmos.jl/blob/c6aadf74a31e25dec498a19843f9ea2270e50146/src/cache/microphysics_cache.jl#L1210

config/model_configs/prognostic_edmfx_bomex_tracerA_column.yml
haakon-e: ah I see, no problem then!

PR #4594, opened 2026-06-17
remove diagnostic EDMF

szy21:
## Purpose 

Goodbye diagnostic EDMF. Thank you for your hard work during the past few years. Also removes small box plotting support in CI, as we have changed to true single column.

(Prognostic EDMF is faster and has better performance than diagnostic EDMF now, so we decided not to support diagnostic EDMF anymore. For a simple turbulence scheme, use EDonly EDMF.)

## Content

----


src/cache/edonly_precomputed_quantities.jl
szy21, line 1: This file can probably be combined with prognostic EDMF, but I want to only remove things in this PR and work on it later.

calibration/experiments/gcm_driven_scm/model_config_diagnostic.yml
```diff
@@ -1,6 +1,6 @@
 initial_condition: "GCM"
 external_forcing_file: "/resnick/groups/esm/zhaoyi/GCMForcedLES/forcing/corrected/HadGEM2-A_amip.2004-2008.07.nc"
-turbconv: "diagnostic_edmfx"
```
costachris, line 3: you can just delete this config since there's a prognostic one (and the rest of the config is for diagnostic)

post_processing/ci_plots.jl
```diff
@@ -1423,22 +1404,15 @@ end
...
-        EDMFBoxPlots,
-        DiagEDMFBoxPlotsWithPrecip,
         EDMFColumnPlots,
         EDMFColumnPlotsWithPrecip,
     },
     output_paths::Vector{<:AbstractString},
 )
     simdirs = SimDir.(output_paths)
 
-    # Determine if this is a box or column type
```
costachris, line 1435: perhaps add to PR description: remove small box plotting support in CI (since we're back to doing real columns instead of small boxes)

sajjadazimi (approved):
RIP DiagnosticEDMF!

costachris (approved):
Looks good to me. We may want to do a release soon since this is a big feature change

szy21:
> Looks good to me. We may want to do a release soon since this is a big feature change

Yes, we can release atmos this week.

PR #4596, opened 2026-06-17
Replace prescribed surface-area BC with a surface mass-flux BC

sajjadazimi:
## Purpose

**Item 1 from code alignment tasks**

Replace the prescribed surface-area boundary condition in PrognosticEDMFX (`turbconv_params.surface_area` force-relaxed via an artificial first-cell entrainment override) with a physical surface mass-flux boundary condition. The previous BC pinned the first-cell area fraction to a fixed target, which prevented `a` from evolving with surface forcing and required a hand-tuned relaxation rate.

### New surface BC

At the first cell, the updraft receives a Deardorff convective surface mass flux

$$
F_{\mathrm{surf}} = a_s \rho w^\*
$$

with

$$
a_s = a_{s,\max} \frac{(w^\*)^3}{(w^\*)^3 + c_u \cdot (u^\*)^3}
$$

and

$$
w^\* = \sqrt{\max\left(z_i \langle w'b' \rangle_s, 0\right)}
$$

Here, $a_{s,\max} =$ `turbconv_params.surface_area` is now interpreted as the *asymptotic* plume area fraction in the free-convection limit ($w^\* \gg u^\*$). The factor

$$
\frac{(w^\*)^3}{(w^\*)^3 + (u^\*)^3}
$$

smoothly shuts the source off in shear-dominated conditions. The boundary-layer depth $z_i$ is taken from `SFP.gustiness_zi`.

### Edge cases

- **Stable boundary layer** ($\langle w'b' \rangle_s \le 0$): $F_{\mathrm{surf}} = 0$, $a_s = 0$, no inflow, and the area fraction decays through advective sinks.
- **Shear-only or very weak convection** ($a_s \to 0$): $F_{\mathrm{surf}} \to 0$ linearly, while the `sgs_scalar_first_interior_bc` percentile coefficient grows only logarithmically. Their product therefore vanishes. `eps` guards in `percentile_bounds_mean_norm` prevent NaNs near the floating-point floor.
- **Area-fraction saturation** ($a \to a_{\max}$): the limiter tends to zero, and the $\rho a$ source, MSE and $q_{\mathrm{tot}}$ tendencies, and the Jacobian rate all vanish consistently.

---


tapios (commented):
@costachris will be in a better position to judge whether this makes sense and works. 

This isn't quite implementing what's in the paper draft, and what I though @costachris had said is the best way forward. In the paper draft, we have no Dirichlet condition on the area fraction itself. This adds a surface mass flux source. Do we want this?

reproducibility_tests/ref_counter.jl
```diff
@@ -31,6 +31,17 @@
 # 3) (optional) leave a link to the buildkite run that prompted this ref counter bump.
 
 #=
+364
+- PrognosticEDMFX: replace the prescribed surface-area BC with a
+  surface-mass-flux BC. The first-cell ρa equation now receives an
+  a-independent source F_surf · limiter / dz with
+  F_surf = a_s · ρ · w*, a_s = a_s_max · w*³/(w*³+u*³), and the
```
tapios:
```suggestion
  F_surf = a_s · ρ · w*, a_s = a_s_max · w*³/(w*³+c_u* u*³), and the
```
(We need some tuning constant multiplying u* for this to make sense, I think.)

src/prognostic_equations/implicit/initialize_implicit_problem.jl
```diff
@@ -399,6 +400,35 @@ function solve_sgs_ρa_implicit_stage_analytic!(Y, p, dtγ)
             Fields.field_values(Fields.level(ᶜmass_flux_factor_bot, 1))
         @. ᶜmass_flux_factor_bot_first = FT(0)
 
+        # Surface mass-flux boundary condition. F_surf = a_s·ρ·w* (see
+        # `surface_mass_flux`) enters as an a-independent source F_surf/dz
+        # in the first-cell numerator. a_s = a_s_max · w*³/(w*³+u*³) blends
```
tapios: As above, I think we need a tuning parameter multiplying u*.

src/prognostic_equations/mass_flux_closures.jl
```diff
@@ -112,6 +112,52 @@ function ᶠupdraft_nh_pressure_drag(params, ᶠlg, ᶠu3ʲ, ᶠu3⁰)
...
+"""
+@inline function surface_mass_flux_coefficient(
+    buoyancy_flux,
+    z_i,
+    ustar,
+    a_s_max,
+)
+    FT = typeof(ustar)
+    w3 = max(z_i * buoyancy_flux, FT(0))
+    return a_s_max * w3 / max(eps(FT), w3 + ustar^3)
```
tapios: Add O(1) tuning parameter multiplying ustar.

tapios:
Would also be good to unit tests for this.

src/prognostic_equations/implicit/initialize_implicit_problem.jl
```diff
@@ -399,6 +400,35 @@ function solve_sgs_ρa_implicit_stage_analytic!(Y, p, dtγ)
             Fields.field_values(Fields.level(ᶜmass_flux_factor_bot, 1))
         @. ᶜmass_flux_factor_bot_first = FT(0)
 
+        # Surface mass-flux boundary condition. F_surf = a_s·ρ·w* (see
```
tapios: That's not what we have in the paper draft buy may be reasonable. (@costachris?)

If we pursue the approach here, an issue is that this is not scale-aware, i.e., the mass flux at the boundary does not vanish in the limit of high resolution. This is something we need to think about improving.

tapios: There's also a risk that setting the updraft area and scalars at the boundary explicitly may yield unphysical environmental values--we should check this.

szy21:
For `surface_mass_flux_source`, if you can, I think it would be safer to construct a C3 vector and use the divergence operator, for example, apply it in edmfx_vertical_diffusion_tendency. But maybe you can't do it easily.

(Btw, I forgot to comment in previous PRs with the analytical u3 and area fraction, it might be safer to use `get_physical_w` instead of doing `u_3 / dz` explicitly.)

src/prognostic_equations/mass_flux_closures.jl
costachris: can you add a new parameter for it in ClimaParams? or reuse an existing one if it makes sense

src/prognostic_equations/implicit/initialize_implicit_problem.jl
```diff
@@ -399,6 +400,35 @@ function solve_sgs_ρa_implicit_stage_analytic!(Y, p, dtγ)
...
+        # `surface_mass_flux`) enters as an a-independent source F_surf/dz
+        # in the first-cell numerator. a_s = a_s_max · w*³/(w*³+u*³) blends
+        # free-convection and shear-driven regimes, with `a_s_max` the
+        # asymptotic plume area fraction in free convection. The source is
+        # scaled by `entr_upper_area_limiter_factor` so it shuts off
+        # smoothly as the plume area approaches `a_max`, preventing the
+        # first-cell area from being driven beyond physical bounds.
+        z_i = SFP.gustiness_zi(CAP.surface_fluxes_params(p.params))
```
costachris: This is using the same gustiness_zi parameter used in the Deardorff gustiness model for wind speed. So it may be confusing if it changes since it enters in 2 places now. I would add a new parameter (maybe `EDMF_convective_zi` or similar) or just make it very clear. 
Eventually it maybe be good to incorporate a better or non-constant boundary layer height estimate (diagnosed from profiles or simple closure). That could be used in a few places, but this is reasonable and standard.

src/prognostic_equations/edmfx_boundary_condition.jl
```diff
@@ -2,6 +2,194 @@
...
+    dz,
+    turbconv_params,
+)
+    return surface_mass_flux(buoyancy_flux, ρ, z_i, ustar, a_s_max) *
+           entr_upper_area_limiter_factor(draft_area(ρa, ρ), turbconv_params) /
+           dz
+end
+
+"""
+    edmfx_boundary_condition_tendency!(Yₜ, Y, p, t, turbconv_model)
```
costachris, line 226: Can you document that the new BC results in 2 contributions at the bottom cell center: both the surface mass-flux BC here as well as the standard entrainment that operates for the rest of the column (`edmfx_entr_detr_tendency!` and `edmfx_boundary_condition_tendency!` both affect the first cell center).  We may or may not want to consider just having `edmfx_boundary_condition_tendency!` active in the bottom center (but that gives less control). And the same assumption is in the implicit Jacobians

src/prognostic_equations/edmfx_boundary_condition.jl
```diff
@@ -2,6 +2,194 @@
...
+            ustar_val,
+            obukhov_length_val,
+            sfc_local_geometry_val,
+        )
+
+        # Apply source · (val_buoyant − val) / ρa_floor where
+        # `source = surface_mass_flux_source(...)` is F_surf·limiter/dz.
+        # The `max(ρa, ρ·a_min)` floor keeps the divisor away from zero
+        # while the updraft is just starting to grow.
+        @. sgsʲs_mseₜ_int_val +=
```
costachris: At cell 1 I think we're relaxing scalars toward two different targets: `mse_env` (from standard entrainment) and `mse_gridmean + fluctuation` (from this BC). These can compete depending on values is that the desired behavior? (before `edmfx_first_interior_entr_tendency!` handled it). this ties into the previous comment about 2 contributions at the bottom cell center

src/prognostic_equations/implicit/initialize_implicit_problem.jl
costachris: As far as scale awareness, since we want massflux to smoothly vanish at higher resolutions, that could be incorporated through area at the surface with something like 
`a_s_eff = a_s_max * (1 - exp(-(Δx / C)²))`
with C as a parameter.  Or for something more physical, perhaps a modification of the surface updraft mass flux source based on the updraft buoyancy. e.g. a ρa source that decreases as `sgsʲs_mse` approaches `ᶜmse_buoyant_air_int_val` (effectively scaling with the env variance)

src/prognostic_equations/edmfx_boundary_condition.jl
sajjadazimi: I added a note in the code to clarify we have these two both contributing in the first cell. I think the two contributions could represent distinct physical processes and both should act at the first cell. `edmfx_entr_detr_tendency!` is lateral entrainment from the environment at every cell while `edmfx_boundary_condition_tendency!` is the surface mass flux from the high-tail (buoyant) sub-cell air below the first cell, brought in through the bottom face. Together they could reproduce the physically intuitive picture: the plume at level 1 is a mix of environment lateral entrainment and surface mass-flux inflow.

costachris (approved):

PR #4601, opened 2026-06-22
Update .buildkite/coupler_perf/pipeline.yml

nefrathenrici:
This PR updates the end to end coupler pipeline to use the most recent coupler commit and automatically compare SYPD to the most recent CA main build.

### DIFF
--- a/.buildkite/coupler_perf/pipeline.yml
+++ b/.buildkite/coupler_perf/pipeline.yml
@@ -10,7 +10,7 @@ env:
   GKSwstype: 100
   SLURM_KILL_BAD_EXIT: 1
   COUPLER_PATH: ".buildkite/perf/ClimaCoupler.jl"
-  COUPLER_COMMIT: "44fcb38b98ed31d753c464b54a2b9a910d1aaab9"
+  COUPLER_COMMIT: "8f827b9118a6a0edfa2199c6b7340ae509168ac4"
   CONFIG_PATH: "$COUPLER_PATH/config/nightly_configs"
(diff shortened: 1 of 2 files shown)


.buildkite/coupler_perf/pipeline.yml
```diff
@@ -48,7 +48,7 @@ steps:
     artifact_paths: "output/gpu_amip_progedmf_1M_land_he16/artifacts/*"
     env:
       CLIMACOMMS_DEVICE: "CUDA"
-      BASELINE_SYPD: "0.24"
+      BASELINE_SYPD: "0.175"
```
petebachant, line 52: It would be great if we could get this from the last time this pipeline ran on `main` so it stays up to date on its own. Is that artifact something we could easily get in Buildkite?

nefrathenrici: I updated the PR to use [build metadata](https://buildkite.com/docs/pipelines/configure/build-meta-data) to track the baseline SYPD automatically.

`check-sypd.sh` now does the following:
1. Read SYPD from sypd.txt (as before)
2. Fetch the most recent passing main build baseline, which is stored as metadata variable `baseline_sypd`. If the fetch fails, use the baseline SYPD hardcoded by the pipeline
3. If current branch is main, set the metadata `baseline_sypd` for this build, so future builds can use it.

nefrathenrici: On second thought, it makes the most sense to always update the baseline SYPD from passing main builds. I updated my comment above with the correct behavior.

petebachant (approved):
LGTM!

PR #4609, opened 2026-06-25
COSP subcolumn

cflsepyt:
## Purpose 

The goal of this PR is to convert grid-mean cloud, precipitation, and hydrometeor quantities into subcolumn quantities that can later be used to calculate radar reflectivity for the CloudSat simulator. 

## Content

- Added cloud subcolumn mask generation from grid-mean cloud fraction.
- Added precipitation subcolumn mask generation.
- Added hydrometeor slicing into subcolumns for cloud liquid, cloud ice, rain, and snow based on the 1M microphysics hydrometeor categories.
- Added precomputed fields needed for the online subcolumn workflow.
- Added tests for the implemented subcolumn and hydrometeor logic.

----


src/callbacks/callbacks.jl
```diff
@@ -123,6 +125,150 @@ NVTX.@annotate function rrtmgp_model_callback!(integrator)
...
+function set_cosp_large_scale_precipitation_flux!(
+    Y,
+    p,
+    ::NonEquilibriumMicrophysics1M,
+)
+    (; ᶜlarge_scale_precipitation_flux, ᶜwᵣ, ᶜwₛ) = p.precomputed
+    FT = eltype(Y)
+
+    @. ᶜlarge_scale_precipitation_flux =
+        max(FT(0), -Y.c.ρq_rai * ᶜwᵣ - Y.c.ρq_sno * ᶜwₛ)
```
nefrathenrici:
```suggestion
        max(FT(0), Y.c.ρq_rai * ᶜwᵣ + Y.c.ρq_sno * ᶜwₛ)
```

The signs need to be flipped as the terminal velocities are positive. To catch this, we need a 1M test that executes this path.

config/default_configs/default_config.yml
```diff
@@ -105,6 +105,12 @@ idealized_h2o:
 dt_rad:
   help: "Time between calling radiation callback for sphere configurations. Only applies to RRTMGP modes, ignored for `held_suarez`, which is applied every stage."
   value: "6hours"
+enable_subcol:
+  help: "Enable the subcolumn cloud overlap callback [`false` (default), `true`]"
+  value: false
+dt_subcol:
+  help: "Time between calling the subcolumn cloud overlap"
+  value: "6hours"
```
nefrathenrici: The `enable_subcol` option shouldn't be needed, since it is always used with `dt_subcol`. Instead, the default value for `dt_subcol` should be `Inf`, which will disable the cloud overlap callback.

src/ClimaAtmos.jl
```diff
@@ -170,4 +170,16 @@ include(joinpath("config", "yaml_helper.jl"))
...
+# Added for COSP testing
+include(joinpath("cosp", "subcol.jl"))
+import .COSPSubcolumns
+include(joinpath("cosp", "prec_subcol.jl"))
+import .COSPPrecipSubcolumns
+include(joinpath("cosp", "hydrometeor_subcol.jl"))
+import .COSPHydrometeorSubcolumns
+include(joinpath("cosp", "reff_np_1m.jl"))
+import .COSP1MReffNpDiagnostics
+
```
nefrathenrici: The files in `src/cosp/*` should be grouped in a module call `COSP`, which can be imported here 


```suggestion
# Added for COSP testing
include(joinpath("cosp", "cosp.jl")
import .COSP
```

src/cache/precomputed_quantities.jl
```diff
@@ -148,6 +148,35 @@ function precomputed_quantities(Y, atmos)
     ᶜcloud_fraction = similar(Y.c, FT)
     @. ᶜcloud_fraction = FT(0)
 
+    n_cosp_subcolumns = 256
```
nefrathenrici: This block of code currently allocates unconditionally. We should only allocate fields if they are needed. To fix this, add a type. This type will be added as a property of the main `AtmosModel` type, and will centralize other options that are currently somewhat scattered.
```
struct COSPModel{N}
    n_subcolumns::Val{N} = Val(256)
    overlap::Symbol = :maximum_random        # :maximum | :random | :maximum_random
    random_seed::UInt64 = UInt64(1)          # replaces SUBCOL_RANDOM_SEED
end
```

This will allow us to allocate fields as needed, depending on the number of subcolumns:

```
if !isnothing(atmos.cosp)
    n_cosp_subcolumns = atmos.cosp.n_subcolumns
    ᶜsubcolumn_cloud = ntuple(_ -> similar(Y.c, FT), n_cosp_subcolumns)
...
end
```

config/model_configs/baroclinic_wave_equil.yml
```diff
@@ -3,7 +3,7 @@ dt_save_state_to_disk: "2days"
 reproducibility_test: true
 initial_condition: "MoistBaroclinicWave"
 dt: "450secs"
-t_end: "6days"
+t_end: "1days"
```
nefrathenrici: A bunch of configs have had their `t_end` values changed. I assume this was for testing and should be reverted.

src/callbacks/get_callbacks.jl
```diff
@@ -421,19 +438,29 @@ Creates the tuple of model callbacks for any AtmosModel by calling
...
+    dt_subcol = "6hours",
+    kwargs...,
+)
     callbacks = ()
-    # Physical constraints callback is registered here rather than at the component
-    # level because the decision depends on both the microphysics model AND the
-    # turbconv model simultaneously — registering from each component independently
-    # would cause double-registration for EDMF + 1M/2M configurations.
-    # Placed before the component callbacks so it fires before any other component
-    # callback that reads cache fields (e.g. radiation)
```
nefrathenrici: This comment should not be deleted.

src/cosp/subcol.jl
```diff
@@ -0,0 +1,273 @@
...
+                    random_seed,
+                    column_id,
+                    nlev - ilev + 1,
+                    isubcolumn,
+                    FT,
+                )
+        end
+    end
+
+    return nothing
```
nefrathenrici:
```suggestion
    coords = Fields.coordinate_field(axes(rand_field))
    @. rand_field = _rand_for_point(random_seed, coords, isubcolumn)
    return nothing
```

I think this should be GPU-compatible. You may need to adjust the `_rand_for_point` function to work with this.

src/cosp/prec_subcol.jl
```diff
@@ -0,0 +1,198 @@
...
+        ntuple(
+            isubcolumn -> _has_large_scale_cloud(
+                frac_out[isubcolumn],
+                large_scale_cloud,
+                nlev,
+            ),
+            nsubcolumns,
+        )
+
+    @inbounds for ilev in nlev:-1:1
```
nefrathenrici: I think this loop can be reformulated as a `column_accumulate!` but it would require all 256 subcolumns to be in memory at once and likely won't be performant. We will need to rework this to work on GPU.

nefrathenrici (commented):
This looks good considering the amount of new code. I marked a few big-picture things to fix. In particular, the core loop will need work to become GPU compatible.

jefftyoder:
@cflsepyt could you write up a brief description for this PR describing the goal and scope?

cflsepyt:
> @cflsepyt could you write up a brief description for this PR describing the goal and scope?

@jefftyoder I've updated the PR description. Please let me know if you have any questions or if anything needs to be clarified.

src/cosp/cosp.jl
```diff
@@ -0,0 +1,18 @@
...
+include("prec_subcol.jl")
+import .COSPPrecipSubcolumns
+
+include("hydrometeor_subcol.jl")
+import .COSPHydrometeorSubcolumns
+
+include("reff_np_1m.jl")
+import .COSP1MReffNpDiagnostics
+
+include("cloudsat_optics.jl")
```
nefrathenrici: This file needs to be committed for the module to load.

test/cosp/reff_np_1m_test.jl
```diff
@@ -0,0 +1,208 @@
+using Test
+using ClimaAtmos
+using ClimaCore: Domains, Meshes, Spaces, Fields, Geometry
+
+const CA = ClimaAtmos
+const CNR = CA.COSP1MReffNpDiagnostics
```
nefrathenrici:
```suggestion
import ClimaAtmos.COSP.COSP1MReffNpDiagnostics as CNR
```

test/cosp/subcol_test.jl
```diff
@@ -0,0 +1,674 @@
+using Test
+using ClimaAtmos
+using ClimaCore: Domains, Meshes, Spaces, Fields, Geometry
+
+const CA = ClimaAtmos
```
nefrathenrici:
```suggestion
import ClimaAtmos as CA
```

test/cosp/subcol_test.jl
```diff
@@ -0,0 +1,674 @@
...
+        return box_position
+    elseif overlap === :random
+        threshold_min = convective_cloud
+        random_number = CA.COSPSubcolumns._rand_for_point(
+            seed,
+            ipoint,
+            ilev,
+            isubcolumn,
+            typeof(threshold_min),
+        )
```
nefrathenrici: `_rand_for_point` now only has 3 arguments. I recommend testing locally to catch bugs like this.

nefrathenrici (commented):
I think the main remaining work is the GPU compatibility and testing.

src/cosp/prec_subcol.jl
nefrathenrici: Ok, I've planned a bit more- we do not need all 256 subcolumns in memory. The only thing coupling subcolumns is the per-level reduction over subcolumns, which can collapse to two per-level scalar fields:
  - "any subcolumn precipitates at k"  ==  flux[k] > 0
  - "any subcolumn cloudy at k"        ==  cloud_fraction[k] > 0
With those, each subcolumn can be an independent top->bottom column_accumulate! with a 2-scalar state (precip, flux_above).

We  split the work into:

1. shared per-level selectors: From the single grid-mean cloud
  fraction field, build selectors `has_cloud`, `has_cloud_below`, `has_cloud_anywhere`.
  These are identical for every subcolumn and encode "what the other subcolumns
  are doing," so a subcolumn never needs to look at its neighbors.
2. one subcolumn at a time: For each subcolumn, a single top-to-bottom
  `column_accumulate!` applies the four rules over each subcolumn independently.


These are set in the cache:
```
sel = (;
    has_cloud          = similar(cloud_fraction),  # "any subcol cloudy at height k?"
    has_cloud_below    = similar(cloud_fraction),  # value from one level below
    has_cloud_anywhere = similar(cloud_fraction),  # column-OR of has_cloud
    scratch            = similar(cloud_fraction),  # temp used by _column_any!
)
```

Populate these values with `set_scops_selectors!`. This will require two new functions:
-  `shift_up!`: shift all values from sel.has_cloud up one into sel.has_cloud_below. This can be implemented using column_accumulate!
- `column_any!`: `any` applied over the column. If one value is 1, they all get set to 1. This can be implemented using column_accumulate!
```julia
function set_scops_selectors!(sel, cloud_fraction)
    FT = eltype(cloud_fraction)
    # Set has_cloud
    @. sel.has_cloud = ifelse(cloud_fraction > zero(FT), one(FT), zero(FT))
    # Set has_cloud_below
    shift_up!(sel.has_cloud_below, sel.has_cloud)
    # Set has_cloud_anywhere
    column_any!(sel.has_cloud_anywhere, sel.has_cloud, sel.scratch) 
    return nothing
end
```


Fill ONE precipitation subcolumn with `scops_subcolumn_precip!`
```julia
function scops_subcolumn_precip!(precip_subcol, cloud_s, flux, sel, scratch)
    FT = eltype(precip_subcol)
    large_scale_cloud = one(eltype(cloud_s))

    # Create cloud mask: 1 if cloud else 0
    @. scratch.cloud = ifelse(cloud_s == large_scale_cloud, one(FT), zero(FT))

    # Shift cloud mask up one
    shift_up!(scratch.cloud_below, scratch.cloud)

    # cloud anywhere: 1 at every level if this column is cloudy at any level, else 0
    column_any!(scratch.any_cloud, scratch.cloud, scratch.or)

    # Build the broadcasted object with seven per-level inputs the rule reads. The order here must match
    # the do-block argument list below.
    input = Base.broadcasted(
        tuple,
        scratch.cloud,           # cloudy in this subcol?
        scratch.cloud_below,     # cloudy one level below?
        sel.has_cloud,           # any subcol cloudy here?
        sel.has_cloud_below,     # any subcol cloudy below?
        sel.has_cloud_anywhere,  # any cloud in the column
        flux,                    # precip flux here
        scratch.any_cloud,       # this subcol cloudy anywhere?
    )

    # Accumulate top to bottom
    Operators.column_accumulate!(
        precip_subcol,
        input;
        init = (; precip = zero(FT), flux_above = zero(FT)),  # nothing above model top
        transform = state -> state.precip,                    # write precip into precip_subcol
        reverse = true,                                       # top -> bottom
    ) do state, (c, c_below, hc, hc_below, hc_any, fx, any_s)
        # No precip, exit early
        fx <= zero(FT) && return (; precip = zero(FT), flux_above = fx)

        # Precip above or cloudy subcolumn
        step1 = (hc > zero(FT)) | (state.flux_above > zero(FT))

        # fallback A: No precip above but some subcol is cloudy one level below
        fallback_A = (!step1) & (hc_below > zero(FT))

        precip =
            step1 ? max(c, state.precip) :          # cloudy here OR precip from above
            fallback_A ? c_below :                  # this subcol cloudy one level below
            (hc_any > zero(FT)) ? any_s :           # fallback B: this subcol cloudy anywhere
            one(FT)                                 # fallback C: no cloud in column -> fill all

        # Carry this level's precip and flux down
        return (; precip, flux_above = fx)
    end

    return nothing
end
```

Putting it together:
```julia
# Phase 1: build the shared selectors once from the grid-mean cloud fraction
set_scops_selectors!(sel, cloud_fraction)

# reset accumulators before the sweep
@. acc.precip_fraction = zero(eltype(cloud_fraction))

# Phase 2: stream one subcolumn at a time
for s in 1:N
    scops_subcolumn!(cloud_s, cloud_fraction, s, seed, overlap) # Get the current subcolumn's overlap mask ( to be implemented)
    scops_subcolumn_precip!(prec_s, cloud_s, flux, sel, scratch)
    # Use the set precip subcolumn for the simulator (placeholder)
    @. acc.precip_fraction += prec_s / N
end
```

src/cosp/reff_np_1m.jl
szy21, line 1: We should double check if we can use CloudMicrophysics functions here.

src/cosp/subcol.jl
```diff
@@ -0,0 +1,224 @@
...
+end
+
+@inline _box_position(::Type{FT}, isubcolumn, nsubcolumns) where {FT} =
+    (FT(isubcolumn) - FT(0.5)) / FT(nsubcolumns)
+
+function _fill_random_field!(rand_field, random_seed, isubcolumn)
+    coords = Fields.coordinate_field(axes(rand_field))
+    @. rand_field = _rand_for_point(random_seed, coords, isubcolumn)
```
szy21: Could we just use the `rand` here? We have something similar in RRTMGP: https://github.com/CliMA/RRTMGP.jl/blob/ed4377233521429bff232ce794d7c70cf2559d2b/src/optics/cloud_optics.jl#L275

nefrathenrici: Is the goal a bitwise reproduction of COSP? I think this is needed for that, otherwise I think we can just use `rand`

cflsepyt: Yeah, the current version is for reproducibility of subcolumn masks

szy21: We won’t have bitwise reproduction of the Fortran code I think. And I think it’s fine to drop the reproducibility of cosp if that makes the code simpler. It’s only for diagnostics.

src/cosp/prec_subcol.jl
szy21: This prec_subcol.jl file still loops over columns manually. Could it follow the same pattern as in subcol.jl?

cflsepyt: Yes, I just modified it following Nat's comments. I will commit it after the tests passed.

config/model_configs/baroclinic_wave_equil.yml
```diff
@@ -13,4 +13,4 @@ diagnostics:
     period: 1days
     reduction_time: average
     pressure_coordinates: true
-    compute_every: 1hours
+    compute_every: 1hours
```
nefrathenrici: Needs newline

src/callbacks/callbacks.jl
```diff
@@ -123,6 +123,176 @@ NVTX.@annotate function rrtmgp_model_callback!(integrator)
...
+
+    @debug "subcol callback" t = integrator.t
+
+    return nothing
+end
+
+function set_cosp_large_scale_precipitation_flux!(
+    Y,
+    p,
+    ::NonEquilibriumMicrophysics1M,
```
nefrathenrici: This method and another method of `set_cosp_hydrometeor_subcolumns!` exclude microphysics 2M/2MP3, they should be added to the dispatch.

src/cosp/subcol.jl
```diff
@@ -0,0 +1,279 @@
...
+    horizontal_key = xor(
+        horizontal_key,
+        _mix_uint64(_coord_bits(_coord_component(coords, Val(:long), FT))),
+    )
+    vertical_key = _coord_bits(coords.z)
+    x = random_seed
+    x = xor(x, horizontal_key * 0x9e3779b97f4a7c15)
+    x = xor(x, vertical_key * 0xbf58476d1ce4e5b9)
+    x = xor(x, UInt64(isubcolumn) * 0x94d049bb133111eb)
+    return FT(_mix_uint64(x) >> 11) * FT(0x1.0p-53)
```
nefrathenrici: We should double check this function to ensure it is working as intended with both floating point types.

test/runtests.jl
```diff
@@ -90,6 +90,11 @@ if TEST_GROUP in ("parameterizations", "all")
     # Chemistry tests
     @safetestset "Chemistry tendency tests" begin @time include("parameterized_tendencies/chemistry/chemistry_tendency.jl") end
 
+    # COSP subcolumn tests
```
nefrathenrici: This should be put into a test group in this file, right now it will always run regardless of test group.

src/callbacks/callbacks.jl
```diff
@@ -123,6 +123,176 @@ NVTX.@annotate function rrtmgp_model_callback!(integrator)
...
+function set_cosp_large_scale_precipitation_flux!(Y, p, _)
+    (; ᶜlarge_scale_precipitation_flux) = p.precomputed
+    FT = eltype(ᶜlarge_scale_precipitation_flux)
+
+    @. ᶜlarge_scale_precipitation_flux = FT(0)
+
+    return nothing
+end
+
+function set_cosp_hydrometeor_subcolumns!(
```
nefrathenrici: I think we can avoid the cache allocation entirely, since nothing reads these fields back yet. Instead of storing the subcolumn fields persistently, we can expose one streaming entry point that regenerates each subcolumn's masks and hands the consumer lazy sliced hydrometeors:

```

function foreach_cosp_subcolumn(consume!::F, Y, p) where {F}
    (;
        ᶜcloud_fraction,
        ᶜsubcolumn_cloud,
        ᶜsubcolumn_threshold,
        ᶜsubcolumn_precip,
        ᶜscops_selectors,
        ᶜprecip_subcolumn_scratch,
        ᶜlarge_scale_precipitation_flux,
        ᶜsampled_cloud_fraction,
        ᶜsampled_precip_fraction,
    ) = p.precomputed
    cosp = p.atmos.cosp
    nsubcolumns = _cosp_nsubcolumns(cosp.n_subcolumns)

    # Materialize the four grid means once (not per subcolumn), reuse
    # p.scratch like microphysics_cache.jl does, instead of lazy specific().
    ᶜq_lcl = p.scratch.ᶜtemp_scalar
    ᶜq_icl = p.scratch.ᶜtemp_scalar_2
    ᶜq_rai = p.scratch.ᶜtemp_scalar_3
    ᶜq_sno = p.scratch.ᶜtemp_scalar_4
    @. ᶜq_lcl = specific(Y.c.ρq_lcl, Y.c.ρ)
    @. ᶜq_icl = specific(Y.c.ρq_icl, Y.c.ρ)
    @. ᶜq_rai = specific(Y.c.ρq_rai, Y.c.ρ)
    @. ᶜq_sno = specific(Y.c.ρq_sno, Y.c.ρ)

    for isubcolumn in 1:nsubcolumns
        COSP.COSPSubcolumns.scops_subcolumn!(
            ᶜsubcolumn_cloud, ᶜsubcolumn_threshold, ᶜcloud_fraction,
            isubcolumn, nsubcolumns, cosp.random_seed;
            overlap = cosp.overlap,
        )
        COSP.COSPPrecipSubcolumns.scops_subcolumn_precip!(
            ᶜsubcolumn_precip, ᶜsubcolumn_cloud,
            ᶜlarge_scale_precipitation_flux,
            ᶜscops_selectors, ᶜprecip_subcolumn_scratch,
        )
        hydrometeors = (;
            q_lcl = @. lazy(COSP.COSPHydrometeorSubcolumns._sliced_cloud_value(
                ᶜq_lcl, ᶜsampled_cloud_fraction, ᶜsubcolumn_cloud)),
            q_icl = @. lazy(COSP.COSPHydrometeorSubcolumns._sliced_cloud_value(
                ᶜq_icl, ᶜsampled_cloud_fraction, ᶜsubcolumn_cloud)),
            q_rai = @. lazy(COSP.COSPHydrometeorSubcolumns._sliced_precip_value(
                ᶜq_rai, ᶜsampled_precip_fraction, ᶜsubcolumn_precip)),
            q_sno = @. lazy(COSP.COSPHydrometeorSubcolumns._sliced_precip_value(
                ᶜq_sno, ᶜsampled_precip_fraction, ᶜsubcolumn_precip)),
        )
        consume!(isubcolumn, hydrometeors)
    end
    return nothing
end
```
A consumer then only keeps accumulators, whose count is independent of N:
```
foreach_cosp_subcolumn(Y, p) do i, h
    # h.q_rai is a lazy broadcast, the += materializes it, fused, right here
    @. ᶜq_rai_mean += h.q_rai / nsub
end
```

nefrathenrici (commented):
This looks good to me! I think the main roadblock is the large cache allocation, I left a suggestion for it.

src/callbacks/callbacks.jl
cflsepyt: Thanks! I updated it to stream each subcolumn following the comment. The current consumer is a placeholder and can be replaced in future CloudSat simulator PR

src/callbacks/callbacks.jl
```diff
@@ -125,6 +125,209 @@ NVTX.@annotate function rrtmgp_solver_callback!(integrator)
...
+        cosp.overlap,
+        ᶜprecip_subcolumn_scratch.column_any,
+    )
+
+    set_cosp_large_scale_precipitation_flux!(Y, p, p.atmos.microphysics_model)
+
+    FT = eltype(ᶜcloud_fraction)
+    @. ᶜsampled_cloud_fraction = zero(FT)
+    @. ᶜsampled_precip_fraction = zero(FT)
+    for isubcolumn in 1:nsubcolumns
```
nefrathenrici, line 172: This loop is repeated a few times in the code, we could move it into a helper function.

src/ClimaAtmos.jl
```diff
@@ -171,4 +171,9 @@ include(joinpath("config", "yaml_helper.jl"))
 
 include(joinpath("utils", "show.jl"))
 
+
+
```
nefrathenrici:
```suggestion
```
These added lines should be deleted.

nefrathenrici (approved):
Thank you, this looks great! I didn't have any blocking comments.

config/default_configs/default_config.yml
```diff
@@ -111,6 +111,18 @@ idealized_h2o:
...
+dt_subcol:
+  help: "Time between calling the subcolumn cloud overlap. Set to `Inf` to disable."
+  value: "Inf"
+cosp_n_subcolumns:
+  help: "Number of COSP subcolumns."
+  value: 256
+cosp_overlap:
+  help: "COSP cloud overlap assumption [`maximum`, `random`, `maximum_random` (default)]."
+  value: "maximum_random"
+cosp_random_seed:
```
szy21: why do we need the random_seed to be a config?

src/callbacks/callbacks.jl
```diff
@@ -125,6 +125,209 @@ NVTX.@annotate function rrtmgp_solver_callback!(integrator)
     return nothing
 end
 
+NVTX.@annotate function subcol_model_callback!(integrator)
+    Y = integrator.u
+    p = integrator.p
+    foreach_cosp_subcolumn(consume_cosp_subcolumn!, Y, p)
+
+    @debug "subcol callback" t = integrator.t
+
```
szy21:
```suggestion
```

src/callbacks/callbacks.jl
```diff
@@ -125,6 +125,209 @@ NVTX.@annotate function rrtmgp_solver_callback!(integrator)
...
+    @. ᶜq_rai = specific(Y.c.ρq_rai, Y.c.ρ)
+    @. ᶜq_sno = specific(Y.c.ρq_sno, Y.c.ρ)
+
+    grid_mean_hydrometeors =
+        (; q_lcl = ᶜq_lcl, q_icl = ᶜq_icl, q_rai = ᶜq_rai, q_sno = ᶜq_sno)
+
+    return foreach_prepared_cosp_subcolumn!(consume!, grid_mean_hydrometeors, p)
+end
+
+function foreach_cosp_subcolumn(consume!::F, _, p, _) where {F}
```
szy21: We do have q_lcl, q_icl for 0M too, so it's not accurate. But to make things simpler, we can just support cosp with 1M / 2M, and throw an error if someone uses cosp with 0M or dry model.

src/cosp/subcol.jl
```diff
@@ -0,0 +1,302 @@
...
+        init = (; current = zero(FT), shifted = zero(FT)),
+        transform = state -> state.shifted,
+    ) do state, value
+        (; current = value, shifted = state.current)
+    end
+    return nothing
+end
+
+"""
+Set every level to one iff any input level in that column is nonzero.
```
szy21:
```suggestion
Set every level to one if any input level in that column is nonzero.
```

szy21 (approved):

PR #4614, opened 2026-06-29
Os/abba chem

obin1:
## Purpose 
This PR contributes a simple gas-phase chemical mechanism tested on the Bomex single-column model and a global aquaplanet model.

## To-do

What remains to be done is adding tests in the CI pipeline, for example comparing certain gridcells of the aquaplanet run that are minimally influenced by transport perturbations against a box model, both of which should fairly quickly reach equilibrium.
ABBA_singlecell

While the mechanism is read in automatically, the species themselves are hard coded into the CliMA code. This could be changed in the future to be mechanism-agnostic, dynamically creating CliMA tracers from a mechanism specified in a config file.

## Content

This uses a Musica/MICM backend to solve a gas phase mechanism with 3 species and one reversible reaction represented as two forward reactions.

```
AB -> A + B 
A + B -> AB
```

The two configurations tested are in `./config/model_configs/prognostic_edmfx_bomex_ABBA_column.yml` for the Bomex case and `./config/model_configs/prognostic_edmfx_aquaplanet_abba.yml` for the aquaplanet case.

----


ext/ClimaAtmosMusica/ClimaAtmosMusica.jl
```diff
@@ -16,8 +16,90 @@ function ClimaAtmos.chemistry_tendency!(
...
+function ClimaAtmos.update_chemistry!(
+    Y,
+    p,
+    t,
+    chemistry_model::ClimaAtmos.GasPhaseChem,
+)
+    isnothing(chemistry_model.config_path) && return nothing
+
+    micm = Musica.MICM(; config_path = chemistry_model.config_path) # chemistry model should have a micm
+    state = Musica.create_state(micm)                               # also make the state outside of the solve set
```
K20shores: Just a reminder, and I'm sure you know, that it would be better to read the configuration once and create a state once, and then reuse them for each solve

obin1: Thanks @K20shores! The most recent commit puts these in a chemistry cache rather than remaking the mechanism every update chemistry step

PR #4615, opened 2026-06-29
Larcform1 microphysics tendency plots

jefftyoder:
Implement plotting pattern from prognostic edmfx 1M microphysics tendency plots into the Larcform1 ci plotting block. 

This PR now also changes the Larcform1 plots to use dispatch rather than string pattern matching for the plotting logic.

### DIFF
--- a/post_processing/ci_plots.jl
+++ b/post_processing/ci_plots.jl
@@ -1859,6 +1859,60 @@ function make_plots(::Larcform1Plots, output_paths::Vector{<:AbstractString})
         )
     end
 
+    # Microphysics 1M process tendency plots (grid-mean, updraft, environment)
+    mp1m_source_names = [
+        "S_phase_change_vap_lcl",
+        "S_phase_change_vap_icl",
+        "S_acnv_lcl_rai",
+        "S_acnv_icl_sno",
+        "S_accr_lcl_rai",
+        "S_accr_lcl_sno_cold",
+        "S_accr_lcl_sno_warm",
+        "S_accr_melt_lcl_sno",
+        "S_accr_icl_rai",
+        "S_accr_freeze_icl_rai",
+        "S_accr_icl_sno",
+        "S_accr_rai_sno_cold",
+        "S_accr_rai_sno_warm",
+        "S_accr_melt_rai_sno",
+        "S_phase_change_vap_rai",
+        "S_phase_change_vap_sno",
+        "S_melt_icl_lcl",
+        "S_melt_sno_rai",
+    ]
+    mp1m_all_names =
+        vec(["mp1m_", "mp1mup_", "mp1men_"] .* permutedims(mp1m_source_names))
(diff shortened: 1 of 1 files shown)


post_processing/ci_plots.jl
```diff
@@ -1740,15 +1740,22 @@ end
...
+    s = String(S)
+    if startswith(s, "larcform1")
+        if occursin("_1M", s)
+            return make_plots(Val(:larcform1_1M), output_paths)
+        else
+            return make_plots(Val(:larcform1), output_paths)
+        end
     end
     @warn "No plot found for $val"
 end
```
haakon-e: What are you trying to achieve here? What are some examples of what `val` would be that you're hoping this function will catch?

post_processing/ci_plots.jl
```diff
@@ -1881,6 +1888,80 @@ function make_plots(::Larcform1Plots, output_paths::Vector{<:AbstractString})
         )
     end
 
+    # Microphysics 1M process tendency plots — only for larcform1_1M
+    if sim_type isa Val{:larcform1_1M}
```
haakon-e: you don't _really_ need to do this condition since you filter based on what's actually available in `available_short_names` anyways, but if it's clearer to you that's fine to keep.

haakon-e (approved):
Looks fine, see some minor comments.

post_processing/ci_plots.jl
jefftyoder: My goal was call this plotting code on any runs with larcform1 in the job_id. For example, `larcform1` , `larcform1_1M`, `larcform1_seaicetest`. This has been useful when iterating during model implementation.

However, I could see this string-matching method having unforeseen side-effects; I would prefer to use type dispatch, but I have not figured out a way to cleanly implement that.

haakon-e: You can specify `reference_job_id` in the launch config to point to a stable `"job_id"` different from the one used to name the output folder, see the `default_config.yml`:

https://github.com/CliMA/ClimaAtmos.jl/blob/17a5478ccdf572f239a6abf4803e55ed62428dc0/config/default_configs/default_config.yml#L147-L150

so if your `job_id=larcform1_seaicetest`, you can still set `reference_job_id=larcform1` for the purposes of plotting.

jefftyoder: I see; the quicklook plots are identical to the ci_plots in that case. Noted.

haakon-e: I don't know what "quicklook plots" are; I presume it's just old wording that hasn't been updated. I only know of ci_plots.

jefftyoder: Changed to use dispatch rather than string matching for plotting logic. Thanks for the feedback @haakon-e. Marking as resolved.

jefftyoder:
@haakon-e I've cleaned up this PR and separated it from the Larcform1 sea ice PR. Would you take a final look over this before I merge?

PR #4616, opened 2026-06-29
Add Larcform1 Arctic SCM case with sea-ice surface models

jefftyoder:
## Add Larcform1 Arctic SCM case with sea-ice surface models

Implements the single-column model configuration for the Larcform1 Arctic winter boundary layer intercomparison (Pithan et al. 2016), including two sea-ice surface temperature types and a prognostic Eisenman-Zhang thermodynamic ice model.

### New SCM case: `Larcform1`

- `Larcform1` setup struct initialises temperature, specific humidity (from RH profile), and geostrophic wind from `AtmosphericProfilesLibrary` profiles
- `Larcform1Insolation`: zero TOA shortwave flux with `cos(60°)` reference zenith, representing Arctic January at 80°N
- Coriolis parameter derived at 80°N (f = 1.432×10⁻⁴ s⁻¹)
- Hydrostatic pressure profile computed from liquid-water-relative RH
- Config: `larcform1_1M_prognostic_edmfx.yml` + `larcform1_1M_prognostic_edmfx.toml` (1-moment microphysics, prognostic EDMFx)
- Pithan et al. (2016) added to `docs/bibliography.bib`

### New surface temperature types

**`ConstantSeaIceTemperature`** (`src/surface_conditions/surface_temperature.jl`)
- Fixed-temperature sea-ice boundary (default 250 K); no prognostic state
- Dispatches `surface_saturation_phase → TD.Ice()` so surface humidity is saturated over ice rather than liquid, giving physically correct sublimation LHF
- Used by `larcform1_1M_prognostic_edmfx` config; replaces the prior hand-computed liquid `q_vap` override in `Larcform1.surface_condition`

**`EisenmanIceTemperature`** (`src/surface_conditions/surface_temperature.jl`)
- Prognostic 0-layer thermodynamic sea ice (Semtner 1976; Eisenman & Wettlaufer 2009; Zhang et al. 2021)
- Carries three prognostic surface fields in `Y.sfc`: `T` (surface temperature), `h_ice` (ice thickness), `T_ml` (ocean mixed-layer temperature)
- Also dispatches `surface_saturation_phase → TD.Ice()`

### Eisenman-Zhang callback

`eisenman_seaice_step!` (`src/prognostic_equations/eisenman_seaice.jl`) advances the ice/ocean state operator-split (once per timestep via callback rather than as an IMEX tendency), because freezing/frazil/melt transitions are discontinuous:

1. Basal flux `F_base = C₀(T_ml − T_base)` advances ice thickness and mixed-layer temperature
2. Frazil formation: prevents `T_ml` below `T_freeze` by converting energy deficit to ice growth
3. Ice-free transition: surplus melt energy warms the mixed layer
4. Surface temperature: one Newton iteration balancing conductive flux through ice vs. atmospheric flux, capped at `T_freeze`; ice-free → surface equals `T_ml`

`∂F_atm/∂T_sfc ≈ 4σT_sfc³` (radiative linearisation; shortwave term omitted — zero insolation case).

### New config: `larcform1_eisenman_1M_prognostic_edmfx`

Full prognostic path: `prognostic_surface: EisenmanSeaIce`, 1M microphysics, prognostic EDMFx, `allskywithclear` radiation, 90 levels, `dz_bottom = 10 m`, `dt = 30s`.

### Tests

- `test/larcform1.jl`: exercises `ConstantSeaIceTemperature` surface init at 250 K and verifies ice saturation phase dispatch

### TODO
- [ ] add Eisenman & Wettlauferb (2009) to references.bib


docs/bibliography.bib
```diff
@@ -286,3 +286,15 @@ @article{ShipwayHill2012
...
+  title = {Select Strengths and Biases of Models in Representing the {Arctic} Winter Boundary Layer over Sea Ice: The {Larcform} 1 Single Column Model Intercomparison},
+  author = {Pithan, Felix and Ackerman, Andrew and Angevine, Wayne M. and Hartung, Kerstin and Ickes, Luisa and Kelley, Maxwell and Medeiros, Brian and Sandu, Irina and Steeneveld, Gert-Jan and Sterk, H. a. M. and Svensson, Gunilla and Vaillancourt, Paul A. and Zadra, Ayrton},
+  journal = {Journal of Advances in Modeling Earth Systems},
+  volume = {8},
+  number = {3},
+  pages = {1345--1357},
+  year = {2016},
+  issn = {1942-2466},
+  doi = {10.1002/2016MS000630}
+}
```
haakon-e:
```suggestion
}

```
nit: add newline at the end of files.

post_processing/ci_plots.jl
```diff
@@ -1859,6 +1846,60 @@ function make_plots(::Larcform1Plots, output_paths::Vector{<:AbstractString})
         )
     end
 
+    # Microphysics 1M process tendency plots (grid-mean, updraft, environment)
+    mp1m_source_names = [
```
haakon-e: is this duplicated from your other PR?

src/prognostic_equations/eisenman_seaice.jl
```diff
@@ -0,0 +1,124 @@
...
+ 4. Surface temperature: if ice remains, solve the conductive/atmospheric
+    balance (one Newton iteration) capped at `T_freeze`; otherwise the surface
+    equals the mixed-layer temperature.
+
+The turbulent flux is always included (the model is driven by SH + LH + radiation);
+`disable_surface_flux_tendency` only gates the slab-ocean tendency path and has no
+effect here. The atmospheric-flux derivative is approximated by the radiative term
+`∂F_atm/∂T_sfc ≈ 4 σ T_sfc³` (shortwave albedo term omitted: zero insolation).
+"""
+function eisenman_seaice_step!(integrator)
```
haakon-e, line 43: many calls in this function appear to allocate, e.g.,
```Julia
ice_covered = @. h_ice > 0
```
see e.g., `docs/dev-guides/performance/gpu_performance.md`

src/surface_conditions/surface_temperature.jl
```diff
@@ -59,6 +58,49 @@ diagnostics.
...
+"""
+@kwdef struct EisenmanIceTemperature{FT} <: SurfaceTemperature
+    # --- sea ice ---
+    C0_base::FT = 120       # ice-base heat transfer coefficient [W / m² / K]
+    T_base::FT = 273.16     # ice-base (ocean/ice interface) temperature [K]
+    L_ice::FT = 3e8         # volumetric latent heat of fusion [J / m³]
+    T_freeze::FT = 273.16   # freezing temperature [K]
+    k_ice::FT = 2           # thermal conductivity of ice [W / m / K]
+    α_ice::FT = 0.70        # ice albedo (unused at zero insolation)
+    σ::FT = 5.67e-8         # Stefan-Boltzmann constant [W / m² / K⁴]
```
haakon-e, line 95: fetch constants like this from ClimaParams instead:
https://github.com/CliMA/ClimaParams.jl/blob/fafee9d147e6b6e31bfdae906756afbbccae058b/src/parameters.toml#L2499-L2502

src/surface_conditions/surface_conditions.jl
```diff
@@ -173,7 +173,10 @@ function surface_state_to_conditions(
     else
         # Assume that the surface is water with saturated air directly
         # above it.
-        q_vap_sat = TD.q_vap_saturation(thermo_params, T_sfc, ρ_sfc, TD.Liquid())
+        q_vap_sat = TD.q_vap_saturation(
+            thermo_params, T_sfc, ρ_sfc,
+            surface_saturation_phase(atmos.surface.temperature),
```
haakon-e, line 178: Can you add a test that this returns `TD.Liquid()` for the base `SurfaceTemperature` case?

haakon-e (approved):
As far as I can tell, this looks good.

I can't speak to the surface parameterization itself; it would be better if someone else has a look at that.

Left a few comments for you to consider.

szy21:
Do we want a sea ice model in ClimaAtmos? I think it would be better if we run this in ClimaCoupler.

post_processing/ci_plots.jl
jefftyoder: Yes, these changes ended up on the wrong branch. I'll clean this up now.

jefftyoder:
@szy21 The Eisenman sea ice model in this PR is inspired by the now removed [ClimaCoupler implementation](https://github.com/CliMA/ClimaCoupler.jl/pull/1284). If we instead added the sea ice model(s) into the coupler, I would be happy with that solution.

trontrytel:
Looks good. I agree with @haakon-e comments on allocations and parameters. @akshaysridhar - what do you think about the sea ice model? Here or the Coupler?

szy21:
> @szy21 The Eisenman sea ice model in this PR is inspired by the now removed [ClimaCoupler implementation](https://github.com/CliMA/ClimaCoupler.jl/pull/1284). If we instead added the sea ice model(s) into the coupler, I would be happy with that solution.

The Eisenman sea ice model was remove at that time because no one was using it. The sea ice models are still in the coupler, but the only option is Prescribed sea ice: https://github.com/CliMA/ClimaCoupler.jl/blob/main/src/Models/prescr_seaice.jl

I think it is cleaner to have the sea ice models in the coupler. But if you don't want to deal with the coupler for single column simulations, adding a simple sea ice model in the atmos is probably ok. I think the coupler has been refactored significantly, so it's probably not too complicated to do a coupled single column simulation. @akshaysridhar would know more. Also tagging @nefrathenrici, who may have some thoughts on what the scope of ClimaAtmos should be.

haakon-e:
ideally ClimaAtmos and ClimaCoupler would have reasonably similar API for setting up components and running (coupled) simulations[^1] so that gluing together the atmosphere and sea ice doesn't also mean learning an entirely new way of working with the atmosphere model.

[^1]: I have no idea if this is true today, as I don't really run the coupler myself.

From  a conceptual point of view I think sea ice models, as with land models, should be run from the coupler; though it's convenient for very simple instances (e.g. constant temperature) to be a part of ClimaAtmos.
It might be a bit of a gray area exactly when a boundary condition is sufficiently complicated for it to conceptually live with the coupler; perhaps when the surface model has its own prognostic state, as is the case with this PR, if I understand correctly?

szy21:
> From a conceptual point of view I think sea ice models, as with land models, should be run from the coupler; though it's convenient for very simple instances (e.g. constant temperature) to be a part of ClimaAtmos. It might be a bit of a gray area exactly when a boundary condition is sufficiently complicated for it to conceptually live with the coupler; perhaps when the surface model has its own prognostic state, as is the case with this PR, if I understand correctly?
> 

Yes, I think so. Atmos always needs a boundary condition, and to me, a constant temperature is a boundary condition, not a surface model. And anything that has its own prognostic state is a surface model, and should ideally stay in the coupler.

The only exception right now is that we do have a "slab ocean" in the atmosphere. The major reason it is there is to accumulate diagnostics for conservation check, which is only accurate if you add a prognostic variable and timestepping it. Another reason is that some people may want to run e.g. RCE with slab ocean without worrying about the coupler. But I'm still debating whether I want to remove the slab ocean from ClimaAtmos.

But in this case, the sea ice model is more complicated than a slab ocean, so I am leaning towards moving this, and the future surface models to the coupler.

akshaysridhar:
I agree that this is better handled in ClimaCoupler.jl - @jefftyoder and I spoke offline earlier today and we'll move the implementation to ClimaCoupler soon.

jefftyoder:
> I agree that this is better handled in ClimaCoupler.jl - @jefftyoder and I spoke offline earlier today and we'll move the implementation to ClimaCoupler soon.

PR #4618, opened 2026-06-29
Add verbose logging option for AtmosSimulation

nefrathenrici:
This PR adds a verbose logging option for the AtmosSimulation constructor, false by default. Config-based simulations always run with this option on, with no logging removed. This PR also replaced the `@timed_str` macro with a `@timed_log` macro that takes the `verbose` boolean. No existing log messages have been removed, a few have been added.

### DIFF
--- a/src/utils/utilities.jl
+++ b/src/utils/utilities.jl
@@ -521,15 +521,32 @@ function prettymemory(b)
 end
 
 """
-    @timed_str expr
-    @timed_str "description" expr
+    @timed_log verbose "message" expr
 
-Returns a string containing `@timed` information.
+Evaluate `expr` and return its value. If `verbose` is true, also `@info` the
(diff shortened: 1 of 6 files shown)


src/utils/utilities.jl
```diff
@@ -521,15 +521,32 @@ function prettymemory(b)
...
+                " (",
+                prettytime(stats.time * 1e9),
+                ", ",
+                prettymemory(stats.gcstats.allocd),
+                ")",
+            )
+            stats.value
+        else
+            $(esc(ex))
+        end
```
haakon-e, line 549: I'm not macro expert, so I will just say I hope and presume `local stats` is good macro hygiene without risk of contaminating code space

haakon-e (approved):
LGTM!

szy21 (approved):

PR #4626, opened 2026-07-02
ci: refresh dev-guides sync workflow from template

imreddyTeja:
Replaces the hand-maintained `update_dev_guides.yml` with the current template from CliMA/DeveloperGuides.

This fixes a permissions bug in the workflow that prevented the monthly sync job from opening PRs correctly.


imreddyTeja:
Also repaired the broken subtree metadata: `git subtree pull` was hitting a merge conflict caused by broken lineage from the original squash-merged add PR, compounded by later commits (e.g. #4594) editing the vendored `docs/dev-guides` content directly. Removed and re-added the subtree from upstream DeveloperGuides so monthly syncs work going forward.

Note: this discards the local edit in #4594 that removed a `DiagnosticEDMFX` reference from `documentation_policy.md` (that feature was removed from ClimaAtmos but the vendored doc mirrors upstream, which still references it). If that reference should stay gone, it needs to be fixed upstream in CliMA/DeveloperGuides so it doesn't get reintroduced on the next sync.

PR #4629, opened 2026-07-02
Process-based jacobian refactor

trontrytel:
Refactor `manual_sparse_jacobian.jl` to organize tracer lists and jacobian blocks by physical process (advection, sedimentation, diffusion, SGS transport), with centralized process-based tracer-name utilities and automated block assembly, while keeping the exact same sparsity pattern and numerical values.

## What changed

New file `src/utils/tracer_processes.jl` — the central, process-based tracer classification:
- `sedimenting_tracer_names(Y)`, `sedimenting_mass_names(Y)`, `diffused_gs_scalar_names(Y)`, `advected_gs_scalar_names(Y)`, `microphysics_tracer_names(Y)` for the grid scale, and `sedimenting_sgs_tracer_names(Y)`, `advected_sgs_scalar_names(Y)`, `passive_sgs_tracer_names(Y)` for the updrafts. All lists filter candidate names by what's actually in Y, so a new tracer opts into its processes automatically.

- Per-tracer process properties: `sedimentation_velocity_name (e.g. ρq_rai → ᶜwᵣ)`, `condensate_phase` (returns `TD.Liquid()` / `TD.Ice()`), and phase-dispatched internal_energy_function, `condensate_e_int_offset`, `condensate_cv_difference`. They replace the hardcoded and duplicated `(name, e_int, ∂cv∂q)`.

`manual_sparse_jacobian.jl` rewritten around processes:

- `jacobian_cache` now assembles the matrix from per-process block builders (`advection_jacobian_blocks`, `diffusion_jacobian_blocks`, `sedimentation_jacobian_blocks`, `sgs_advection_jacobian_blocks`, `sgs_massflux_jacobian_blocks`). Blocks requested by multiple processes are de-duplicated by `merge_jacobian_blocks`, and a final `fallback_identity_blocks` pass gives every state variable without a diagonal a `-I` block. That fallback eliminated all the `ρq_gas_A_if_available`-style special cases. A passive tracer now needs no mention in this file at all (!)

- The 900-line `update_jacobian!` is split into per-process functions mirroring `implicit_tendency!`: `update_advection_jacobian!`, `update_sedimentation_jacobian!`, `update_diffusion_jacobian!`, and `update_sgs_{advection,diffusion,entr_detr,boundary_condition,massflux}_jacobian!`. The scratch-space and block-initialization dependencies between them are documented in an ordering contract in `update_jacobian!`; eddy diffusivities are now computed once and passed to both diffusion updates.

- Removed unused `ᶜkappa_mʲ` computation and a few unused constants. Left TODO comments where the Jacobian intentionally diverges from the tendencies: SGS entrainment skips number concentrations; SGS diffusion skips passive tracers.


src/prognostic_equations/implicit/manual_sparse_jacobian.jl
```diff
@@ -150,244 +116,417 @@ function jacobian_cache(alg::ManualSparseJacobian, Y, atmos)
...
+merge_jacobian_blocks(block_pairs) =
+    unrolled_reduce(block_pairs, ()) do merged, block_pair
+        matching = unrolled_filter(pair -> pair.first == block_pair.first, merged)
+        if isempty(matching)
+            (merged..., block_pair)
         else
-            ()
+            typeof(matching[1].second) == typeof(block_pair.second) ||
+                error("Jacobian block requested with two different types")
+            merged
```
sajjadazimi, line 295: In principle two different types are possible (for example DiagonalRow from thermodynamics and TridiagonalRow from diffusion). Maybe in this case we should use the larger type.

sajjadazimi: This is not crucial for this PR because we don't have such a case now. But maybe add a note in the function doc.

src/prognostic_equations/implicit/manual_sparse_jacobian.jl
```diff
@@ -150,244 +116,417 @@ function jacobian_cache(alg::ManualSparseJacobian, Y, atmos)
...
+            tracer_names,
+        )...,
+        map(
+            name -> (name, @name(f.u₃)) => similar(Y.c, BidiagonalRow_ACT3),
+            sedimenting_names,
+        )...,
+        (@name(c.ρe_tot), @name(c.sgsʲs.:(1).mse)) =>
+            similar(Y.c, TridiagonalRow),
+        # (ρe_tot, ρ) and (ρq_tot, ρ) are also needed by implicit diffusion;
+        # duplicates are dropped by merge_jacobian_blocks.
```
sajjadazimi: This comment is no longer necessary as they are more examples like this -- for example between sedimentation and diffusion blocks. So either remove or add a similar comment there.

src/prognostic_equations/implicit/manual_sparse_jacobian.jl
```diff
@@ -517,700 +645,845 @@ function update_jacobian!(alg::ManualSparseJacobian, cache, Y, p, dtγ, t)
...
+    }
+        MatrixFields.unrolled_foreach(
+            sedimenting_sgs_mass_names(Y),
+        ) do qʲ_name
+            q_state_name = sgs_state_name(qʲ_name)
+            ∂ᶜqʲ_err_∂ᶜqʲ = matrix[q_state_name, q_state_name]
+            @. ∂ᶜqʲ_err_∂ᶜqʲ -=
+                dtγ * DiagonalMatrixRow(ᶜentrʲ + ᶜturb_entrʲs.:(1))
+        end
+    end
```
sajjadazimi: Can we use the same loop for all sgs tracers (including q_tot, mse, and numbers)?

src/prognostic_equations/implicit/manual_sparse_jacobian.jl
```diff
@@ -517,700 +645,845 @@ function update_jacobian!(alg::ManualSparseJacobian, cache, Y, p, dtγ, t)
...
-                        (@name(c.ρn_lcl), @name(c.sgsʲs.:(1).n_lcl), @name(n_lcl)),
-                        (@name(c.ρn_rai), @name(c.sgsʲs.:(1).n_rai), @name(n_rai)),
-                    )
+    # passive tracers, e.g. chemistry tracers
+    MatrixFields.unrolled_foreach(passive_sgs_tracer_names(Y)) do χ_name
+        χ_state_name = sgs_state_name(χ_name)
+        ∂ᶜχʲ_err_∂ᶜχʲ = matrix[χ_state_name, χ_state_name]
+        @. ∂ᶜχʲ_err_∂ᶜχʲ -=
+            dtγ * DiagonalMatrixRow(ᶜentrʲ + ᶜturb_entrʲs.:(1))
+    end
```
sajjadazimi: This can probably be merged with the previous loop!

src/prognostic_equations/implicit/manual_sparse_jacobian.jl
```diff
@@ -517,700 +645,845 @@ function update_jacobian!(alg::ManualSparseJacobian, cache, Y, p, dtγ, t)
...
+    )
+    update_sgs_advection_jacobian!(matrix, Y, p, dtγ)
+    update_sgs_diffusion_jacobian!(
+        matrix,
+        Y,
+        p,
+        dtγ,
+        diffusion_flag,
+        eddy_diffusivities,
+    )
```
sajjadazimi, line 1402: Maybe we can call `update_sgs_diffusion_jacobian!` right after `update_diffusion_jacobian!` to avoid bugs due to overwriting K_h in between.

sajjadazimi: But then we must be careful with changing the order as some functions initialize blocks and some other add to the blocks. It would be good to think if it's possible to initialize all blocks with zero in the beginning of this function and then add to them in all update functions (but maybe it's slower).

sajjadazimi (approved):
Thank you! It's much cleaner now. I left a few comments. I also suggest to separate all the helper functions from Jacobian cache and updates in a new file.
And on the Jacobian sanity checks: 
 - we need all the processes that initialize a block before the processes that update those blocks. For example we need sgs advection active before we can add to the blocks in subsequent sgs processes;
 - we need gm processes active if their corresponding sgs processes are included. For example we need gm diffusion active to compute K_h for sgs diffusion;
- also physically some processes go together: for example sgs mass flux makes sense if you have both gm and sgs advection.

src/prognostic_equations/implicit/manual_sparse_jacobian.jl
trontrytel: Done. Left a comment for now.

src/prognostic_equations/implicit/manual_sparse_jacobian.jl
trontrytel: Agreed. I think I want to do that in the next PR. This one is already unwieldy when rebasing

trontrytel:
> Thank you! It's much cleaner now. I left a few comments. I also suggest to separate all the helper functions from Jacobian cache and updates in a new file. And on the Jacobian sanity checks:
> 
> * we need all the processes that initialize a block before the processes that update those blocks. For example we need sgs advection active before we can add to the blocks in subsequent sgs processes;
> * we need gm processes active if their corresponding sgs processes are included. For example we need gm diffusion active to compute K_h for sgs diffusion;
> * also physically some processes go together: for example sgs mass flux makes sense if you have both gm and sgs advection.

Agreed on moving to a separate file. And thank you for the test ideas. I'll do this in the next PR, since this one is already difficult to manage

PR #4631, opened 2026-07-02
Update CTS

sajjadazimi:
## Purpose 
Update CTS:
Route enforce_physical_constraints! through ClimaTimeSteppers' new `constrain_state!` hook (per-step by default) and drop the separate callback. Frequencies are configurable via `update_cache_every` / `update_constrain_state_every`. This avoids redundant cache calls.

----


akshaysridhar:
The downstream coupler compats will have to be changed (should be minor changes in coupler since it will inherit the new config args related to `update_constrain_state_every` and `update_cache_every` from ClimaAtmos.jl). Thanks @sajjadazimi ; if the small regression in performance caught by the julia1.10 ci-check is acceptable cc @trontrytel we can merge this.

akshaysridhar (approved):

PR #4632, opened 2026-07-02
Speed up cloud-fraction quadrature kernel on GPU

imreddyTeja:
## Disclosure

This PR was written by **Claude** (Anthropic's Claude Code), working from a prompt to profile `ClimaTimeSteppers.step!(integrator)` on the GPU (`longrun_aquaplanet_allsky_progedmf_1M` config) and find a speedup, explicitly excluding `set_microphysics_tendency_cache!` from consideration. Everything below — the profiling, the two dead ends, and the fix — was done by Claude in that session; I'm flagging this explicitly per the user's request so reviewers know the provenance and can scrutinize accordingly.

## Summary

- Materializes the lazy `air_density` broadcast (`ᶜρ_env`) into a scratch field before the fused cloud-fraction quadrature broadcast in `set_cloud_fraction!` (`QuadratureCloud` dispatch), so its `Field` arguments no longer get fused into that kernel's argument/type footprint.
- Marks the SGS-quadrature scalar math (`_sgs_saturation_moments`, the `SGSVarianceEvaluator` functor, `_compute_z`, `_compute_cloud_fraction`, and `normal_cdf`/`normal_cdf_inv` in `utilities.jl`) `@fastmath`.

## How I got here

1. Set up a GPU REPL in `.buildkite`, built the aquaplanet EDMF simulation via `.buildkite/prof.jl`, and profiled `step!` with `CUDA.@profile` (with `NAME_KERNELS_FROM_STACK_TRACE` enabled for readable kernel names).
2. Excluding `set_microphysics_tendency_cache!`, the next largest kernel was the fused cloud-fraction quadrature broadcast inside `set_cloud_fraction!` — **~1.44–1.50 ms/call**, called ~8×/step (~5.5% of total step time). That's 20–70× more expensive than comparably-sized kernels elsewhere in the step, which pointed at a broadcast-fusion/kernel-footprint problem rather than raw arithmetic cost.
3. **First attempt — `@fastmath` alone**: no measurable effect on the dominant kernel (1.44 ms → 1.50 ms, within noise). It did shave ~10% off two much smaller sub-kernels, but those are negligible slices of the step budget. Kept anyway since it's free and harmless.
4. **Second attempt — `@noinline` function barrier** on `_sgs_saturation_moments` (mirroring a register-pressure fix that worked for an analogous kernel in `CloudMicrophysics.jl` on a prior GPU-optimization pass): also no effect. Reverted.
5. Dug into the actual mangled CUDA kernel name and found the fused broadcast pulls in a full nested `air_density` computation (over `ᶜT⁰`, `ᶜp`, `ᶜq_tot_nonneg⁰`, `ᶜq_liq⁰`, `ᶜq_ice⁰`) as part of its argument tree — the same pattern already flagged elsewhere in this file (`set_ml_cloud_fraction!`'s comment about the 4 KiB GPU kernel parameter limit). **Third attempt** — materializing that one lazy field before the broadcast — is the actual fix.

## Measured results (GPU, `longrun_aquaplanet_allsky_progedmf_1M`)

All numbers are from clean, warmed-up `CUDA.@profile` captures over 5 steps (re-verified after catching that an initial comparison was contaminated by first-call JIT compilation):

| | Before | After |
|---|---|---|
| `set_cloud_fraction!` dominant kernel | ~1.44–1.50 ms/call | ~459 µs/call |
| Total GPU busy time (5 steps) | ~1.04–1.05 s | ~1.02 s |
| Steady-state wall time per step | ~0.204–0.206 s | ~0.201 s (~2% faster) |

Verified no NaNs / errors introduced in the state after each rebuild.

## Notes for reviewers

- `set_microphysics_tendency_cache!` was intentionally left untouched (out of scope per the original request; it's the subject of separate, ongoing GPU-optimization work).
- The materialization change reuses `p.scratch.ᶜtemp_scalar_2`. I checked that this scratch slot is free at that point in the call sequence (used by `set_covariance_cache!`'s internal `compute_∂T_∂θ!`, which always completes and returns before `set_cloud_fraction!` is invoked within `picard_step!`), and that nothing downstream in `set_cloud_fraction!` (`_apply_edmf_cloud_weighting!`) reads `p.scratch` at all — but please double check this reasoning, since scratch-field lifetime bugs can be silent (wrong physics, no crash).
- Formatted with the pinned JuliaFormatter version per this repo's contributing guide.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/parameterized_tendencies/microphysics/cloud_fraction.jl
+++ b/src/parameterized_tendencies/microphysics/cloud_fraction.jl
@@ -287,7 +287,7 @@ struct SGSVarianceEvaluator{TPS, FT}
     mu_S::FT
 end
 
-@inline function (eval::SGSVarianceEvaluator)(T_hat, q_tot_hat)
+@inline @fastmath function (eval::SGSVarianceEvaluator)(T_hat, q_tot_hat)
     q_sat_hat = TD.q_vap_saturation(eval.tps, T_hat, eval.ρ)
     s = q_tot_hat - q_sat_hat
     return (s - eval.mu_S)^2
@@ -316,7 +316,7 @@ Gaussian closure. Using this analytic μ_S lets us accumulate σ_S² as
 
 Returns `(; mu_S, sigma_S)`.
 """
-@inline function _sgs_saturation_moments(
+@inline @fastmath function _sgs_saturation_moments(
     thp, ρ, T_mean, q_tot_mean,
     sgs_quad, T′T′, q′q′, corr_Tq,
 )
@@ -429,7 +429,7 @@ for computing it (typically with `σ_eff = α · σ_S` or `σ_eff = σ_aug` for
 the smooth-floored CF formula) so this helper stays free of parameter-
 dependent logic.
 """
-@inline function _compute_z(C)
+@inline @fastmath function _compute_z(C)
     FT = typeof(C)
 
     # 1. Initial guess: Φ(z₀) = tanh(1.35 · C).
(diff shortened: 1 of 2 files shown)


imreddyTeja:
I think the `@fastmath` speed-up is pretty negligible, so it can be removed. I'm on the fence about keeping it.

TODO: Run the performance check pipeline after #4633 is merged

imreddyTeja:
Closed over fastmath concerns

PR #4637, opened 2026-07-04
Fixes problems near upper boundary and tracer-mass continuity

tapios:
Fixes two problems at the upper boundary and with tracer-mass consistency
* Ensures zero mass flux through top boundary even in the presence of warped coordinate surfaces (before the flux would only be zero if top layers are level)
* Ensures tracer-mass consistency in EDMF by treating all tracers (e.g., condensates) like `mse` and `q_tot`, in difference form, with corresponding Jacobian updates. Reduces SGS mass-flux tendency with an artificially uniform `q_lcl` from 1.1e-7 to 6.7e-14 (Float32 roundoff).

### DIFF
--- a/src/prognostic_equations/edmfx_sgs_flux.jl
+++ b/src/prognostic_equations/edmfx_sgs_flux.jl
@@ -16,6 +16,10 @@ deviation of a conserved variable `ϕ` (such as total enthalpy or specific humid
 from its grid-mean value. These terms represent the redistribution of energy and tracers
 by the resolved SGS circulations relative to the grid mean flow.
 
+The SGS flux of `q_tot` redistributes water mass, so `Yₜ.c.ρ` receives the
+same tendency as `Yₜ.c.ρq_tot` (mirroring the diffusive-flux treatment of
+moist air mass).
+
 The specific implementation depends on the `turbconv_model` (e.g., `PrognosticEDMFX`).
 A generic fallback doing nothing is also provided.
 The function modifies `Yₜ.c` (grid-mean tendencies) in place.
@@ -110,6 +114,7 @@ function edmfx_sgs_mass_flux_tendency!(
                     edmfx_sgsflux_upwinding,
                 )
                 @. Yₜ.c.ρq_tot += vtt
+                @. Yₜ.c.ρ += vtt  # Effect of SGS water flux on (moist) air mass
             end
             # Add the environment fluxes
             ᶜq_tot⁰ = ᶜspecific_env_value(@name(q_tot), Y, p)
@@ -124,21 +129,28 @@ function edmfx_sgs_mass_flux_tendency!(
                 edmfx_sgsflux_upwinding,
             )
             @. Yₜ.c.ρq_tot += vtt
+            @. Yₜ.c.ρ += vtt  # Effect of SGS water flux on (moist) air mass
         end
 
         # Auto-discovered SGS tracer fluxes (microphysics species and any
-        # user-defined passive tracers)
+        # user-defined passive tracers). Like the mse and q_tot fluxes above,
+        # these are difference-form fluxes ρᵏaᵏ(u³ᵏ - u³)(χᵏ - χ), which
+        # vanish identically for uniform χ and are consistent with the SGS
+        # transport of q_tot. The grid-mean advection -∇·(ρ u³ χ) of each
+        # tracer is applied in explicit_vertical_advection_tendency!.
         # Draft fluxes
         for χ_name in sgs_tracer_names(Y)
             ρχ_name = get_ρχ_name(χ_name)
+            ᶜρχ = MatrixFields.get_field(Y.c, ρχ_name)
             for j in 1:n
                 ᶜχʲ = MatrixFields.get_field(Y.c.sgsʲs.:($j), χ_name)
+                @. ᶠu³_diff = ᶠu³ʲs.:($$j) - ᶠu³
                 @. ᶜa_scalar =
-                    ᶜχʲ *
+                    (ᶜχʲ - specific(ᶜρχ, Y.c.ρ)) *
                     draft_area(Y.c.sgsʲs.:($$j).ρa, ᶜρʲs.:($$j))
                 vtt = vertical_transport(
                     ᶜρʲs.:($j),
-                    ᶠu³ʲs.:($j),
+                    ᶠu³_diff,
                     ᶜa_scalar,
                     dt,
                     edmfx_tracer_upwinding,
@@ -148,13 +160,16 @@ function edmfx_sgs_mass_flux_tendency!(
             end
         end
         # Environment fluxes
+        @. ᶠu³_diff = ᶠu³⁰ - ᶠu³
         for χ_name in sgs_tracer_names(Y)
             ρχ_name = get_ρχ_name(χ_name)
+            ᶜρχ = MatrixFields.get_field(Y.c, ρχ_name)
             ᶜχ⁰ = ᶜspecific_env_value(χ_name, Y, p)
-            @. ᶜa_scalar = ᶜχ⁰ * draft_area(ᶜρa⁰, ᶜρ⁰)
+            @. ᶜa_scalar =
+                (ᶜχ⁰ - specific(ᶜρχ, Y.c.ρ)) * draft_area(ᶜρa⁰, ᶜρ⁰)
             vtt = vertical_transport(
                 ᶜρ⁰,
-                ᶠu³⁰,
+                ᶠu³_diff,
                 ᶜa_scalar,
                 dt,
                 edmfx_tracer_upwinding,
--- a/src/prognostic_equations/edmfx_sgs_flux.jl
+++ b/src/prognostic_equations/edmfx_sgs_flux.jl
@@ -44,7 +44,7 @@ function edmfx_sgs_mass_flux_tendency!(
 )
 
     n = n_mass_flux_subdomains(turbconv_model)
-    (; edmfx_sgsflux_upwinding, edmfx_tracer_upwinding) = p.atmos.numerics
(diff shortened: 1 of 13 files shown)


tapios:
@trontrytel Please have a look at the chemistry/Musica changes (mostly test changes). The tests were very brittle, with passing dependent on Julia version/machine used. The intent of the changes here was to make them more robust.

trontrytel:
> @trontrytel Please have a look at the chemistry/Musica changes (mostly test changes). The tests were very brittle, with passing dependent on Julia version/machine used. The intent of the changes here was to make them more robust.

We also have this PR from @ph-kev with fixes for Musica: https://github.com/CliMA/ClimaAtmos.jl/pull/4641

tapios:
> > @trontrytel Please have a look at the chemistry/Musica changes (mostly test changes). The tests were very brittle, with passing dependent on Julia version/machine used. The intent of the changes here was to make them more robust.
> 
> We also have this PR from @ph-kev with fixes for Musica: #4641

#4641 looks simpler. If it covers all failure modes, it's preferable.

tapios:
> > > @trontrytel Please have a look at the chemistry/Musica changes (mostly test changes). The tests were very brittle, with passing dependent on Julia version/machine used. The intent of the changes here was to make them more robust.
> > 
> > 
> > We also have this PR from @ph-kev with fixes for Musica: #4641
> 
> #4641 looks simpler. If it covers all failure modes, it's preferable.

I removed the chemistry/Musica changes from this branch.

src/prognostic_equations/edmfx_sgs_flux.jl
```diff
@@ -124,21 +129,28 @@ function edmfx_sgs_mass_flux_tendency!(
...
-                    ᶜχʲ *
+                    (ᶜχʲ - specific(ᶜρχ, Y.c.ρ)) *
                     draft_area(Y.c.sgsʲs.:($$j).ρa, ᶜρʲs.:($$j))
                 vtt = vertical_transport(
                     ᶜρʲs.:($j),
-                    ᶠu³ʲs.:($j),
+                    ᶠu³_diff,
                     ᶜa_scalar,
                     dt,
                     edmfx_tracer_upwinding,
```
sajjadazimi: Should we use the same upwinding as q_tot and mse here?

src/prognostic_equations/edmfx_sgs_flux.jl
```diff
@@ -148,13 +160,16 @@ function edmfx_sgs_mass_flux_tendency!(
...
-            @. ᶜa_scalar = ᶜχ⁰ * draft_area(ᶜρa⁰, ᶜρ⁰)
+            @. ᶜa_scalar =
+                (ᶜχ⁰ - specific(ᶜρχ, Y.c.ρ)) * draft_area(ᶜρa⁰, ᶜρ⁰)
             vtt = vertical_transport(
                 ᶜρ⁰,
-                ᶠu³⁰,
+                ᶠu³_diff,
                 ᶜa_scalar,
                 dt,
                 edmfx_tracer_upwinding,
```
sajjadazimi: and here?
Of course then the gm and sgs equations will be inconsistent and that can result in negative env values (which will be fixed in fixers). But I think q_tot/tracer consistency is preferred.

src/prognostic_equations/remaining_tendency.jl
```diff
@@ -164,6 +164,13 @@ NVTX.@annotate function additional_tendency!(Yₜ, Y, p, t)
         rst_ρtke = rayleigh_sponge_tendency_tracer(Y.c.ρtke, rayleigh_sponge)
         @. Yₜ.c.ρtke += rst_ρtke
     end
+    # TODO: This damps ρq_lcl/ρq_icl/ρq_rai/ρq_sno toward zero with no
+    # counterpart in ρq_tot, ρ, or ρe_tot, so it implicitly converts
+    # condensate to vapor without accounting for the associated latent
+    # heat, which can generate spurious heating/cooling and supersaturation
+    # in the sponge layer. Either remove this once it's confirmed unneeded
+    # (e.g. after the top-boundary flux fixes), or make it mass- and
+    # energy-consistent by applying the same sink to ρq_tot, ρ, and ρe_tot.
```
sajjadazimi, line 173: I think we cannot modify e_tot, rho, q_tot for mass/energy  conservation. Hopefully this is small and we will not need it at the end.

src/prognostic_equations/implicit/manual_sparse_jacobian.jl
```diff
@@ -1169,12 +1204,13 @@ function update_jacobian!(alg::ManualSparseJacobian, cache, Y, p, dtγ, t)
...
                     @. ᶜtridiagonal_matrix =
                         ᶜtracer_advection_matrix ⋅ ᶠset_tracer_upwind_matrix_bcs(
-                            ᶠtracer_upwind_matrix(ᶠu³⁰),
+                            ᶠtracer_upwind_matrix(ᶠu³⁰ - ᶠu³),
                         )
                     MatrixFields.unrolled_foreach(
                         microphysics_tracers,
                     ) do (ρχ_name, χʲ_name, χ_name)
                         MatrixFields.has_field(Y, ρχ_name) || return
+                        ᶜρχ = MatrixFields.get_field(Y, ρχ_name)
```
sajjadazimi: I think it's safe to remove the whole env block as its contribution is small in difference mode. This is also consistent with q_tot and mse equations where env blocks are absent.

src/prognostic_equations/implicit/manual_sparse_jacobian.jl
```diff
@@ -1143,6 +1151,33 @@ function update_jacobian!(alg::ManualSparseJacobian, cache, Y, p, dtγ, t)
...
+
+                        ∂ᶜρχ_err_∂ᶠu₃ =
+                            matrix[ρχ_name, @name(f.u₃)]
+                        @. ∂ᶜρχ_err_∂ᶠu₃ =
+                            -dtγ * ᶜtracer_advection_matrix ⋅
+                            DiagonalMatrixRow(
+                                ᶠset_tracer_upwind_bcs(
+                                    ᶠtracer_upwind(
+                                        CT3(sign(ᶠu³ʲ_data - ᶠu³_data)),
+                                        (ᶜχʲ - specific(ᶜρχ, Y.c.ρ)) *
```
sajjadazimi: For q_tot and mse the Jacobian is written only for central differencing. Maybe we should treat tracers similarly. Then the their code blocks can be merged.

sajjadazimi (approved):
Thank you! The top velocity fix looks good. The sgs tracer flux changes are also reasonable. However, if we want full consistency probably we should treat them numerically similar to sgs q_tot and mse fluxes. We should use the same upwinding scheme and similar Jacobian treatments (no env blocks and only central differences in Jacobian). The current changes with the default settings are fully-implicit first-order upwinding for tracer fluxes and central-differencing implicit q_tot and mse fluxes.
Also, the gm advection is treated differently. For q_tot and energy the advection happens in implicit tendency while tracer advection happens in explicit tendency.

tapios:
> Thank you! The top velocity fix looks good. The sgs tracer flux changes are also reasonable. However, if we want full consistency probably we should treat them numerically similar to sgs q_tot and mse fluxes. We should use the same upwinding scheme and similar Jacobian treatments (no env blocks and only central differences in Jacobian). The current changes with the default settings are fully-implicit first-order upwinding for tracer fluxes and central-differencing implicit q_tot and mse fluxes. Also, the gm advection is treated differently. For q_tot and energy the advection happens in implicit tendency while tracer advection happens in explicit tendency.

Thank you, @sajjadazimi. And, yes, these are good points about the consistency with mse and q_tot. The origin for the difference is the implicit treatment necessary for wave motions for energy and q_tot, and this was inherited downstream here. I did think about changing it, but maybe that's better in a separate PR.

src/prognostic_equations/remaining_tendency.jl
tapios: Yes, you'd break energy and mass conservation in the domain. If it's necessary, the physical idea would be outflow of condensate through the top. I agree that I hope we won't need it.

src/prognostic_equations/edmfx_sgs_flux.jl
tapios: Yes, good point.

src/prognostic_equations/edmfx_sgs_flux.jl
tapios: Yes

src/prognostic_equations/implicit/manual_sparse_jacobian.jl
tapios: Yes, good point. It also helps collapse the different Jacobian blocks.

src/prognostic_equations/remaining_tendency.jl
tapios: Updated comment

tapios:
> > Thank you! The top velocity fix looks good. The sgs tracer flux changes are also reasonable. However, if we want full consistency probably we should treat them numerically similar to sgs q_tot and mse fluxes. We should use the same upwinding scheme and similar Jacobian treatments (no env blocks and only central differences in Jacobian). The current changes with the default settings are fully-implicit first-order upwinding for tracer fluxes and central-differencing implicit q_tot and mse fluxes. Also, the gm advection is treated differently. For q_tot and energy the advection happens in implicit tendency while tracer advection happens in explicit tendency.
> 
> Thank you, @sajjadazimi. And, yes, these are good points about the consistency with mse and q_tot. The origin for the difference is the implicit treatment necessary for wave motions for energy and q_tot, and this was inherited downstream here. I did think about changing it, but maybe that's better in a separate PR.

Now an issue, #4645. We should think about whether we want or need the unification--not clear to me it's worth the extra cost. Let's talk about it.

PR #4638, opened 2026-07-05
Fix subsidence top BC and remove top-cell tendency-zeroing workaround

tapios:
## Fix subsidence top BC in cases with prescribed forcing (GCM- or ERA-driven)

* Replace `ᶜsubdivᵥ` (top extrapolatation, which copied the cell below into the
top cell) with `ᶜadvdivᵥ` in the advective-form subsidence operator. With
zero boundary fluxes on both divergences, the advective form implements
a zero-gradient inflow condition at the lid: for `w < 0` the upwinded
top-face flux `w·χ_top` cancels the `χ_top·w` term exactly, equivalent to
prescribing `χ = χ_top` above the lid. For uniform `χ` the tendency vanishes
identically at every level, so tracer-mass (`χ ≡ 1`) consistency holds. 
This also consolidates all advective divergences into a
single operator (`ᶜadvdivᵥ`).

* Delete both `external_forcing.jl` blocks that zeroed the accumulated
`ρe_tot`/`ρq_tot` tendencies in the top cell; they existed to mask the
extrapolation problem and also discarded radiation/nudging/sponge
tendencies there. GCM/ERA5-driven cases will change as a result.

* Extend the tracer-mass consistency test with subsidence assertions:
top-cell tendency vanishes for descending flow with a non-constant
profile (fails at 4.2e-8 on the pre-fix operator), and the tendency
vanishes at every level for uniform `χ` under all reconstruction schemes.

This will affect calibration of SCM cases; they should be recalibrated. 

This goes on top of #4637.


costachris (approved):
The math checks out and the previous issues at the top do not reappear. LGTM
The reproducibility tests should be verified after 4637 goes in.

tapios:
> The math checks out and the previous issues at the top do not reappear. LGTM The reproducibility tests should be verified after 4637 goes in.

Thank you!

PR #4642, opened 2026-07-06
Update CloudMicrophysics compat to v0.37.0

imreddyTeja:
## Summary

This PR was written by Claude Code, working through the CloudMicrophysics v0.37.0 compatibility gap end to end. Both commits on this branch are Claude-authored (see commit trailers).

- Bumps the `CloudMicrophysics` compat bound from `0.36` to `0.37.0` in `Project.toml`, and fixes the code paths that broke as a result.
- A separate, unrelated doc fix (stale `Pkg.test` single-test-group invocation) is kept in its own commit.

## Process

1. **Diffed CloudMicrophysics itself.** With a local checkout of `CloudMicrophysics.jl` (tags `v0.36.0`/`v0.37.0` present), ran `git diff v0.36.0..v0.37.0` over `src/` to see what actually changed, plus `git log` for commit context. This showed the P3 ice scheme was restructured around a new `P3State` struct (`state_from_prognostic`, field renames `L_ice`→`ρq_ice`/`N_ice`→`ρn_ice`), a new `Quadrature` module (pulling in a new `FastGaussQuadrature` dependency), and `BulkMicrophysicsTendencies.bulk_microphysics_tendencies` signature/return changes for the 2-moment (+P3) path.
2. **Found every ClimaAtmos call site into the changed modules** (`P3`, `Microphysics2M`, `BulkMicrophysicsTendencies`, parameter construction) via targeted greps/reads across `src/` and `test/`, to know exactly what needed updating vs. what was untouched by the version bump (e.g. 0M/1M/`LinearizedAverage` paths were unaffected).
3. **Cross-checked each call site's actual arguments against the new function signatures** read directly from the CloudMicrophysics source (not just the diff or docstrings, which turned out to be stale in one place) to catch mismatches precisely:
   - `microphysics_cache.jl`'s 2M+P3 tendency path already called the new `P3State`/`get_distribution_logλ`/terminal-velocity API correctly, but was missing `import CloudMicrophysics.P3Scheme as CMP3` entirely — this looks like it was written in anticipation of v0.37 and never exercised (no test covers `NonEquilibriumMicrophysics2MP3` end-to-end).
   - The same call site was also missing the `q_tot` argument to `bulk_microphysics_tendencies`, which v0.37 made a required positional argument (previously defaulted).
4. **Verified interactively in a Julia REPL** (`.buildkite` project env, CloudMicrophysics dev'd to a local v0.37.0 checkout) before running the full test suite: resolved the manifest, loaded `ClimaAtmos`, and manually exercised `P3State`/`get_distribution_logλ`/terminal-velocity/`bulk_microphysics_tendencies` calls with the exact argument shapes ClimaAtmos uses, to confirm the signatures matched and catch the `q_tot` bug before running anything expensive.
5. **Ran `Pkg.test("ClimaAtmos")` scoped to the `parameterizations` group.** This surfaced a second, more consequential bug: v0.37 added new fields (`dn_lcl_activation_dt`, `dn_ice_dt`) to `bulk_microphysics_tendencies`'s return `NamedTuple`, and ClimaCore's field broadcast assignment does not auto-narrow a superset `NamedTuple` to the destination's fixed-field type — so the *standard* 2-moment warm-rain path (not just the untested 2M+P3 path) failed with a `MethodError` at runtime. Fixed by adding a small `restrict_to_2m_tendency_fields` helper used at all 3 affected call sites.
6. **Re-ran tests**, which then caught a third issue: `test/parameterized_tendencies/microphysics/bmt_integration.jl` asserted the old 7-field return shape for the warm-rain-only BMT call; updated it to expect the new 8-field shape (with `dn_lcl_activation_dt`).
7. **Iterated until the `parameterizations` test group passed cleanly** (all microphysics/BMT/allocation/SGS-quadrature tests green; pre-existing `Broken` tests unaffected).
8. While setting up that final verification run, discovered `docs/clima_atmos_specific.md`'s documented `Pkg.test(...; test_args=["parameterizations"])` invocation doesn't actually scope the test group — `TEST_GROUP` is read from an environment variable, and passing it via `test_args` instead leaks into `ARGS` for the whole test process, breaking `test/restart_utils.jl`'s own argument parsing. Fixed the doc in a separate commit since it's unrelated to the CloudMicrophysics compatibility work itself.

## Test plan

- [x] `Pkg.test("ClimaAtmos")` with `TEST_GROUP=parameterizations` passes (all tests green, no new failures)
- [x] `using ClimaAtmos` precompiles cleanly against CloudMicrophysics v0.37.0
- [x] Manually verified `P3State`, `get_distribution_logλ`, `ice_terminal_velocity_number_weighted`/`_mass_weighted`, and `bulk_microphysics_tendencies` against the exact call shapes used in `microphysics_cache.jl`
- [x] `JuliaFormatter` (pinned version) run with no diff


imreddyTeja:
I don't think the update is this straight-forwards...

haakon-e:
I already have branches that cover this and other needed changes. If you need to update the compat sooner, I suggest we just comment out 2M/P3 tests in ClimaAtmos to get ci to pass, and I'll put them back in the PRs to come. It's fine letting 2M be broken in ClimaAtmos in the meantime.

trontrytel:
Can we do what @haakon-e is suggesting and comment out the 2M/P3 stuff in Atmos and update it to the new CloudMicrohysics? I think there were some non-P3 performance improvements that we want to check

imreddyTeja:
@trontrytel will do. I'll tag you both in the PR when I open it (hopefully within the hour)

trontrytel:
Thanks! And apologies for the mess

PR #4644, opened 2026-07-06
EDMFX corrections: TKE buoyancy production, enthalpy-flux decomposition, cloud-fraction floor

tapios:
Three physics corrections for prognostic EDMFX, stacked on #4638 (which stacks on #4637):

1. **Remove coherent (mass-flux) buoyancy production from the TKE budget.** The coherent updraft/environment buoyancy flux belongs to the resolved SGS circulation, not to the isotropic turbulence that TKE represents; retaining it double-counts buoyancy production already accounted for by the mass-flux transport. This affects TKE budgets (e.g., Sc: TKE peaks at cloud tops, drops sharply above, and is significantly lower below too--as it should be).

2. **Decompose the diffusive enthalpy flux into dry static energy + water enthalpy** (`F_h = -K_h ∇s_d + Σ_μ h_tot,μ (-K_h ∇q_μ)`, unit turbulent Lewis number), at both diffusion sites (`edmfx_sgs_diffusive_flux_tendency!` and `vertical_diffusion_boundary_layer_tendency!`). Diffusing `h_tot` directly implies a spurious enthalpy flux carried by dry-air diffusion, which systematically warms entrained air across capping inversions where `h_tot` jumps up while `q_t` jumps down, contributing to Sc bias. The implicit-Jacobian diffusion factors are updated to match. In DYCOMS RF01, air just above cloud top is now ~1.6 K cooler. May affect mid-troposphere biases too.

3. **Damp the relative part of the cloud-fraction floor as the subdomain mean saturates** (`D = x/√(1+x²)`, `x = max(-μ_S, 0)/(ε_rel·q_sat)`), so an equilibrated overcast deck is no longer capped below CF = 1 by the patchiness floor. No new parameter; the floor still enters only the CF diagnosis, so quadrature mass conservation is untouched. DYCOMS RF01 deck cloud fraction 96–99% (was 74–89%), hour-6 LWP 145 g/m² (was 51).

### DIFF
--- a/src/prognostic_equations/edmfx_sgs_flux.jl
+++ b/src/prognostic_equations/edmfx_sgs_flux.jl
@@ -252,13 +252,49 @@ function edmfx_sgs_diffusive_flux_tendency!(
         ᶠρaK_u = p.scratch.ᶠtemp_scalar_2
         @. ᶠρaK_u = ᶠinterp(Y.c.ρ) * ᶠinterp(ᶜK_u)
 
-        # Total enthalpy diffusion
+        # Total enthalpy diffusion, using the dry-static-energy + water-
+        # enthalpy decomposition
+        #   F_h = -K_h ∇s_d + Σ_μ h_tot,μ F_qμ,   F_qμ = -K_h ∇q_μ,
+        # with s_d = h_d + Φ, h_tot,μ = h_μ(T) + Φ for μ ∈ {vap, liq, ice},
+        # and unit turbulent Lewis number (K_qμ = K_h). Diffusing h_tot
+        # directly would imply a spurious enthalpy flux carried by dry-air
+        # diffusion (h_tot depends on 1 - q_t through the dry-air mass
+        # fraction), systematically warming entrained air at inversions where
+        # h_tot jumps up while q_t jumps down. The thermal piece diffuses dry
+        # static energy (constant under dry-adiabatic displacement), and each
+        # water constituent carries its own enthalpy with its diffusive mass
+        # flux, consistent with the ρq_tot and ρ updates below.
+        #
+        # Note: F_qμ for liquid and ice uses unscaled K_h (omitting the tracer
+        # vertical diffusion factor α_vert_diff_tracer). While microphysics
+        # tracers are diffused with α * K_h to prevent unphysical upward
+        # transport of precipitation, omitting α here maintains exact energetic
+        # consistency with the unscaled ρq_tot diffusion equation, preserves
+        # total water invariance under moist-adiabatic processes, and aligns
+        # with the implicit solver's Jacobian formulation.
         ᶜdivᵥ_ρe_tot = Operators.DivergenceF2C(
             top = Operators.SetValue(C3(FT(0))),
             bottom = Operators.SetValue(C3(FT(0))),
         )
-        (; ᶜh_tot) = p.precomputed
-        @. Yₜ.c.ρe_tot -= ᶜdivᵥ_ρe_tot(-(ᶠρaK_h * ᶠgradᵥ(ᶜh_tot)))
+        thermo_params = CAP.thermodynamics_params(params)
+        (; ᶜΦ) = p.core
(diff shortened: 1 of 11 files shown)


src/prognostic_equations/edmfx_sgs_flux.jl
```diff
@@ -252,13 +252,41 @@ function edmfx_sgs_diffusive_flux_tendency!(
...
+            -(
+                ᶠρaK_h * (
+                    ᶠgradᵥ(TD.dry_static_energy(thermo_params, ᶜT, ᶜΦ)) +
+                    ᶠinterp(TD.enthalpy_vapor(thermo_params, ᶜT) + ᶜΦ) *
+                    ᶠgradᵥ(ᶜq_vap) +
+                    ᶠinterp(TD.enthalpy_liquid(thermo_params, ᶜT) + ᶜΦ) *
+                    ᶠgradᵥ(ᶜq_liq) +
+                    ᶠinterp(TD.enthalpy_ice(thermo_params, ᶜT) + ᶜΦ) *
+                    ᶠgradᵥ(ᶜq_ice)
+                )
```
sajjadazimi, line 295: liq and ice contain precipitation species. Should we involve alpha parameter (that controls precipitation diffusion) here too?

src/prognostic_equations/implicit/manual_sparse_jacobian.jl
```diff
@@ -628,29 +628,53 @@ function update_jacobian!(alg::ManualSparseJacobian, cache, Y, p, dtγ, t)
             @. ᶜdiffusion_u_matrix = ᶜadvdivᵥ_matrix() ⋅ ∂ᶠρχ_dif_flux_∂ᶜχ
         end
 
+        # Jacobian of the decomposed enthalpy diffusion flux
+        #   F_h = -K_h ∇s_d + Σ_μ h_tot,μ (-K_h ∇q_μ)
+        # (see edmfx_sgs_diffusive_flux_tendency!). The derivatives below hold
+        # the h_tot,μ prefactors and the equilibrium condensate partition
```
sajjadazimi: Should it be `non-equilibrium`?

sajjadazimi (approved):
Looks good to me! Thank you! For the derivatives I didn't double check the equations. I just asked ChatGPT and it approved the results. Let me know if I should double check the derivatives.

tapios:
> ᶠinterp

Maybe just have a quick plausibility check? Gemini seems to like everything I do, no matter how wrong...

sajjadazimi:
> Maybe just have a quick plausibility check? Gemini seems to like everything I do, no matter how wrong...

I also gave them a quick skim myself, and they still look consistent.

src/prognostic_equations/implicit/manual_sparse_jacobian.jl
tapios: Thanks. Condensate should just be without qualification.

src/prognostic_equations/edmfx_sgs_flux.jl
tapios: Good point. Yes, definitely.

tapios:
> > Maybe just have a quick plausibility check? Gemini seems to like everything I do, no matter how wrong...
> 
> I also gave them a quick skim myself, and they still look consistent.

Thank you!!

src/prognostic_equations/edmfx_sgs_flux.jl
tapios: Upon further thought, it's complicated...

We really should carry the alpha factors through consistently everywhere, if we have them in the first place. But then we cannot diffuse q_tot but most diffuse water species separately, creating substantial added complexity, e.g., in the Jacobian structure for the implicit solver.

Maybe for now I leave this unscaled with a comment that we should revisit the alpha scaling in the future. The energetic differences here are negligible anyway.

trontrytel (approved):

PR #4647, opened 2026-07-07
SGS inversion reconstruction: make EDMF (PROPHET!) vertically scale-aware (and stabilize clouds)

tapios:
## Purpose

Makes the EDMFX (PROPHET) boundary-layer scheme less sensitive to vertical resolution by reconstructing subgrid inversion structure that a coarse grid cannot resolve directly, and stabilizes low clouds against the coarse-grid "flip to overcast" failure mode.

## Changes

**Stability side (recovers inversion sharpness):**
- **Harmonic-mean face interpolation** of eddy diffusivities (`1/interp(1/max(K,ϵ))`) at the sites that interpolate a center diffusivity to faces — the boundary-layer `VerticalDiffusion`/`DecayWithHeightDiffusion` path and EDMFX updraft-internal diffusion. At a capping inversion, arithmetic averaging assigns ~half the boundary-layer diffusivity to the face with the largest gradient, causing spurious entrainment; the harmonic mean is controlled by the smaller (quiescent-side) value.
- **Stability-biased buoyancy gradient** `N²_stab = max(N²₋, N²₊)` (the more stable of the two one-sided gradients) for the mixing length and turbulent Prandtl number, so a resolved inversion suppresses mixing from both sides.
- **Interface-aware effective stability** `N²_eff = N² + [(Δb)₊]²/(c_b·κ_iso)` senses unresolved buoyancy jumps across a face.

**Entrainment side (recovers coarse-grid cloud cover):**
- **Interfacial entrainment diffusivity** `K_e = γ·w_e·Δz` (`w_e = A·√TKE/max(Ri_b,1)`), gated so it acts only where a stable jump is under-resolved (`γ → 1` at a sheet interface, `→ 0` as the gradient is resolved), restoring finite entrainment across the reconstructed inversion. New calibratable parameter `EDMF_interface_entr_efficiency` (`A`); **ClimaParams default `A = 0.4`, i.e. the closure is on by default** (`A = 0` disables it).

**Numerics and structure:**
- **Face-native diffusivities:** the EDMFX grid-mean stability closure and eddy diffusivities are evaluated directly at cell faces (`ᶠK_h`, `ᶠK_u`, `ᶠK_entr`) rather than interpolated from centers; one shared pipeline serves grid-mean and updraft-internal diffusion, in both the tendencies and the implicit Jacobian. Buoyancy-gradient thermodynamics are factored into cloud-fraction-independent coefficients reused across the centered, one-sided, and face stencils. The TKE budget uses the same face diffusivities and face buoyancy gradient as the fluxes.
- **Resolvability filter-scale cap** on the mixing length: `l ≤ Δ_f = max(Δx_h, Δz)` (horizontal node scale; `Inf` for single columns) instead of the vertical-only `l ≤ Δz` — an eddy is subgrid only if unresolvable in every direction, so the cap is set by the coarsest grid direction (inert for columns/GCM, binding at `Δx_h` in the gray zone, Deardorff `l ≤ Δ` in the LES limit).
- **Jacobian consistency:** the sedimenting-tracer diffusion diagonal carries the full `ρ(α·K_h + K_e)` matching its tendency (previously zeroed at inversions when `α = 0`); the 0M density-weighted microphysics sources are allocated in the implicit cache so the autodiff Jacobians work (fixes a pre-existing `AutoDenseJacobian` crash on EDMFX+0M).
- **Diagnostics:** `edt`/`evu` now report the effective coefficients the model actually applies (face-native `K_h`/`K_u` + `K_entr`, interpolated to centers) instead of recomputing a center-based value; the interfacial contribution is output separately as the new default EDMFX diagnostic `kentr` (so the turbulent part is recoverable as `edt − kentr`).
- Under `EDOnlyEDMFX`, the diagnostic-covariance and ML-cloud closures now use the EDMFX mixing length (previously fell back to Smagorinsky–Lilly).
- Unit tests (136 assertions, Float32/Float64) lock the analytic structure of every new pointwise closure.
- Modest (16%) GPU speed-up of timestep by
  -  factoring the buoyancy-gradient coefficients 
  -  materializing `ᶜmixing_length(Y, p)` (`mixing_length_lopez_gomez_2020`) in `set_covariance_cache!` across each Picard iteration to cut down repeated evaluations and eliminating redundant mixing length evaluation out of `set_explicit_precomputed_quantities!` after Aitken extrapolation 
  - materializing the potential temperature (`TD.liquid_ice_pottemp`) into scratch storage inside `set_buoyancy_gradient_inputs!` early within the step, to remove repeated evaluations across center and face interpolations

## Validation

Column resolution sweeps (DYCOMS RF01, BOMEX, RICO; Δz = 25–200 m; A = 0 vs 0.4), plus aquaplanet and EDOnly-aquaplanet smoke tests. Two findings worth stating:

- **The stability half alone stops inversion erosion.** DYCOMS `z_i` is resolution-independent (788–862 m, one cell's width, across Δz = 25–200 m) **with or without** the entrainment closure — harmonic interpolation + `N²_stab` + `N²_eff` fix the erosion on their own. So the entrainment closure's job is not "hold the inversion" but "recover the right cloud amount on a coarse grid."
- **BOMEX is affected.** At Δz = 150 m with the closure off, the classic coarse-EDMF failure appears — spuriously overcast (cloud cover 100%, LWP 110 g/m²); `A = 0.4` recovers the fine-grid state (20% vs 22%, LWP 20 vs 27), cutting cloud-cover resolution dependence from ~5× to ~10%. A BOMEX coarse-grid `A`-sweep shows a sharp recovery threshold between `A = 0.2` (still 94% cover) and `A = 0.4` (20%), and `A = 1.2` is harmful — so **`A = 0.4` sits just above the minimum effective value and is the current default.** Calibrate `A` against equilibrium (≳ 24 h) targets, not spin-up snapshots.

## Known limitations

- **Validity domain — weak-inversion cumulus at coarse resolution.** The closure is built for strong, mixed-layer-capping inversions (stratocumulus). At a weak-`Δb`, moisture-dominated trade inversion on a coarse grid, the down-gradient form moistens the inversion-base cell faster than it dries it: a 24 h RICO run at Δz = 160 m relapses to overcast for *any* `A` (by moisture trapping at `A = 0`, faster by diffusive moistening at larger `A`), while the fine grid holds trade cumulus. Real cumulus-top entrainment is plume-localized and belongs to the entrainment/detrainment closures, not to a grid-mean `K_e`. Documented in the `set_face_diffusivities!` docstring; a regime-aware gating of `K_e` may help.
- **Residual LWP scatter is discretization, not the closure.** With `z_i` pinned, DYCOMS LWP still scatters non-monotonically (~±25%) with Δz. A 0M (precipitation-free) run shows a comparable gap of opposite sign, so this is vertical cloud-layer representation (how many cells span the deck, where cloud top falls relative to a face), not an entrainment-velocity bias. 
- **Follow-ups evaluated and set aside.** Three refinements aimed at the residual Δz-dependence were implemented/analyzed and tested against the sweeps; none measurably improved convergence, so they are not included: (i) reconstructing SGS TKE at the inversion face; the coarse/fine near-inversion TKE profiles differ rather than being the same profile sampled at different depths, and the `γ` gate already compensates so `z_i` converges; (ii) bounding `ℓ_e` by the inversion thickness: `ℓ_e = l_W` is height-based and already resolution-independent; (iii) a moisture-conserving flux limiter on `K_e`: no single-step over-evaporation, oscillation, or solver failure occurs even at Δz = 200 m, and `K_e` is already treated implicitly. A flux-BC reformulation of the interfacial TKE sink was likewise tested and shelved (it shrank harmless negative-TKE excursions but did not move the coupled solution).


src/prognostic_equations/edmfx_sgs_flux.jl
```diff
@@ -311,7 +303,7 @@ function edmfx_sgs_diffusive_flux_tendency!(
                         turbconv_params,
                         Y.c.ρtke,
                         ᶜtke,
-                        ᶜmixing_length_field,
+                        p.precomputed.ᶜl_mix,
```
costachris: can you extract above with the others, eg
`(; ᶠK_h, ᶠK_u, ᶠK_entr, ᶜl_mix) = p.precomputed`

src/cache/temporary_quantities.jl
```diff
@@ -17,6 +17,7 @@ function implicit_temporary_quantities(Y, atmos)
         ᶠtemp_scalar = Fields.Field(FT, face_space), # ᶠρaK_h
         ᶠtemp_scalar_2 = Fields.Field(FT, face_space), # ᶠρaK_u
         ᶠtemp_scalar_3 = Fields.Field(FT, face_space), # ᶠwaχ in advection
+        ᶠtemp_scalar_4 = Fields.Field(FT, face_space), # ᶠstrain in set_face_diffusivities!
```
costachris: I think `ᶠtemp_scalar_4` here in `implicit_temporary_quantities` is never used (just the one below in `temporary_quantities`), so you can kill it.

src/cache/temporary_quantities.jl
```diff
@@ -44,6 +45,7 @@ function temporary_quantities(Y, atmos)
         ᶠtemp_scalar = Fields.Field(FT, face_space), # ᶠp, ᶠρK_h
         ᶠtemp_scalar_2 = Fields.Field(FT, face_space), # ᶠρK_u
         ᶠtemp_scalar_3 = Fields.Field(FT, face_space),
+        ᶠtemp_scalar_4 = Fields.Field(FT, face_space), # ᶠstrain in set_face_diffusivities!
```
costachris:
```suggestion
        ᶠtemp_scalar_4 = Fields.Field(FT, face_space), # ᶠPr in set_face_diffusivities!
```

src/cache/temporary_quantities.jl
```diff
@@ -44,6 +45,7 @@ function temporary_quantities(Y, atmos)
         ᶠtemp_scalar = Fields.Field(FT, face_space), # ᶠp, ᶠρK_h
         ᶠtemp_scalar_2 = Fields.Field(FT, face_space), # ᶠρK_u
         ᶠtemp_scalar_3 = Fields.Field(FT, face_space),
```
costachris:
```suggestion
        ᶠtemp_scalar_3 = Fields.Field(FT, face_space), # ᶠstrain
```

costachris (changes requested):

costachris:
`compute_edt`/`compute_evu`, used in the output diagnostics, don't match the diffusivities used in the EDMF now, so there's a mismatch between what's done internally and what's output by the diagnostics. Specifically, they omit `K_entr` entirely and still use the center-based diffusivity. Can we either add new diagnostics or just update the ᶜK_h/ᶜK_u diagnostics to match what's done in the code? we can add `K_entr` as it's own diagnostic if that makes more sense

src/prognostic_equations/eddy_diffusion_closures.jl
costachris: I think we can kill this previous definition now unless we want backwards compatibility, it's no longer explicitly used

src/parameterized_tendencies/microphysics/cloud_fraction.jl
szy21:
```suggestion
        turbconv_model isa AbstractEDMF ?
```
I think this was a bug before. For EDOnlyEDFMX we were using smagorinsky-lilly mixing length for cloud fraction calculation.

szy21:
```suggestion
        turbconv_model isa AbstractEDMF ?
```

src/prognostic_equations/implicit/manual_sparse_jacobian.jl
szy21:
```suggestion
ᶜK_u = p.scratch.ᶜtemp_scalar_3
ᶜK_h = p.scratch.ᶜtemp_scalar_4
```
It doesn't matter too much, but will be less confusing if someone adds more scratch fields in this function.

szy21: (If you change this, remember to change the comment before)

src/prognostic_equations/implicit/manual_sparse_jacobian.jl
```diff
@@ -813,14 +813,16 @@ function eddy_diffusivity_coefficients!(Y, p)
...
-        (; ᶜlinear_buoygrad, ᶜstrain_rate_norm) = p.precomputed
+        (; ᶜbuoygrad_stab, ᶜstrain_rate_norm, ᶜl_mix) = p.precomputed
         ᶜtke = @. lazy(specific(Y.c.ρtke, Y.c.ρ))
-        ᶜmixing_length_field = p.scratch.ᶜtemp_scalar_3
-        ᶜmixing_length_field .= ᶜmixing_length(Y, p)
+        # Center K (from the precomputed master mixing length) is needed only
+        # for the updraft internal-diffusion block
+        # (update_sgs_diffusion_jacobian!); the grid-mean diffusion matrices
+        # use the face-native ᶠK_h/ᶠK_u.
         ᶜK_u = p.scratch.ᶜtemp_scalar_4
```
szy21:
```suggestion
```
(redundant)

src/prognostic_equations/implicit/manual_sparse_jacobian.jl
```diff
@@ -813,14 +813,16 @@ function eddy_diffusivity_coefficients!(Y, p)
...
-        @. ᶜK_u = eddy_viscosity(turbconv_params, ᶜtke, ᶜmixing_length_field)
+        @. ᶜK_u = eddy_viscosity(turbconv_params, ᶜtke, ᶜl_mix)
         ᶜprandtl_nvec = @. lazy(
-            turbulent_prandtl_number(params, ᶜlinear_buoygrad, ᶜstrain_rate_norm),
+            turbulent_prandtl_number(params, ᶜbuoygrad_stab, ᶜstrain_rate_norm),
         )
         ᶜK_h = p.scratch.ᶜtemp_scalar_6
```
szy21:
```suggestion
```

szy21:
Why do we use face diffusivities for grid-mean but center diffusivities for updraft?

src/prognostic_equations/edmfx_sgs_flux.jl
tapios: Done.

src/cache/temporary_quantities.jl
tapios: Autodiff uses `p.scratch.ᶠtemp_scalar_4` for `ᶠPr`, so we need it, otherwise AD breaks.

src/cache/temporary_quantities.jl
tapios: Done

src/prognostic_equations/eddy_diffusion_closures.jl
tapios: Done (also removed dead helpers).

src/parameterized_tendencies/microphysics/cloud_fraction.jl
tapios: Done.

tapios: Done

src/prognostic_equations/implicit/manual_sparse_jacobian.jl
tapios: Thank you! This simplified things (in eddy_diffusivity_coefficients!)

src/prognostic_equations/implicit/manual_sparse_jacobian.jl
tapios: Done

tapios:
> Why do we use face diffusivities for grid-mean but center diffusivities for updraft?

No longer a good reason. I unified it.

src/cache/temporary_quantities.jl
tapios: There was a wrong comment on the field; it's not ᶠstrain. I fixed the comments and added a short note above the block saying that the autodiff Jacobians reach set_face_diffusivities! with this scratch.

src/cache/temporary_quantities.jl
tapios: Done

src/cache/temporary_quantities.jl
tapios: Scratch the above. @costachris you were right. I removed it.

src/cache/temporary_quantities.jl
tapios: Done

tapios:
> `compute_edt`/`compute_evu`, used in the output diagnostics, don't match the diffusivities used in the EDMF now, so there's a mismatch between what's done internally and what's output by the diagnostics. Specifically, they omit `K_entr` entirely and still use the center-based diffusivity. Can we either add new diagnostics or just update the ᶜK_h/ᶜK_u diagnostics to match what's done in the code? we can add `K_entr` as it's own diagnostic if that makes more sense

Fixed, with `kentr` added to diagnostics (interpolated to centers; because interpolation is linear, you get turbulent diffusivities as `edt − kentr`).

src/cache/temporary_quantities.jl
```diff
@@ -14,6 +14,11 @@ function implicit_temporary_quantities(Y, atmos)
     FT = Spaces.undertype(center_space)
     uvw_vec = UVW(FT(0), FT(0), FT(0))
     return (;
+        # set_face_diffusivities! runs only in the explicit precomputed
+        # update, with the (larger) scratch set of temporary_quantities;
+        # the implicit tendency consumes its outputs (ᶠK_h, ᶠK_u, ᶠK_entr)
+        # from p.precomputed as frozen coefficients, so its face scratch is
+        # not needed here.
```
szy21: I don't think we need this comment

src/diagnostics/default_diagnostics.jl
```diff
@@ -451,6 +451,7 @@ function default_diagnostics(
         "lmix",
         "evu",
         "edt",
+        "kentr",
```
szy21: I would probably not add kentr to the default diagnostics, but I don't have a strong preference.

.buildkite/Manifest-v1.11.toml
```diff
@@ -443,9 +443,9 @@ weakdeps = ["CUDA"]
 
 [[deps.ClimaParams]]
 deps = ["Dates", "TOML"]
-git-tree-sha1 = "9741f6b6d55df618cd0d893620202e108f136980"
+git-tree-sha1 = "c8bc84467973afdbc070d353a81b2d6439024b84"
 uuid = "5c42b081-d73a-476f-9059-fd94b934656c"
-version = "1.1.1"
+version = "1.1.2"
```
szy21, line 448: We should bump the ClimaParams compat in Project.toml to 1.1.2

src/cache/precomputed_quantities.jl
```diff
@@ -149,6 +156,35 @@ function precomputed_quantities(Y, atmos)
...
+        # viscosity, and interfacial entrainment diffusivity K_e = γ w_e Δz;
+        # filled by `set_face_diffusivities!` for EDMF runs with prognostic
+        # TKE, zero otherwise. Evaluating the stability closure at the faces,
+        # where the fluxes live, keeps the collapse of K at an unresolved
+        # inversion from leaking to the adjacent interior face (which a
+        # center-based evaluation with interpolation cannot avoid).
+        ᶠbuoygrad = zeros(axes(Y.f)),
+        ᶠK_h = zeros(axes(Y.f)),
+        ᶠK_u = zeros(axes(Y.f)),
+        ᶠK_entr = zeros(axes(Y.f)),
```
szy21: Why do they need to be `zero` not `similar`? Using similar is usually more robust, as we will be sure everything is initialized correctly.

src/parameterized_tendencies/microphysics/cloud_fraction.jl
```diff
@@ -190,6 +174,55 @@ function compute_∂T_∂θ!(dest, Y, p, thermo_params)
...
+function materialized_mixing_length!(Y, p)
+    turbconv_model = p.atmos.turbconv_model
+    if turbconv_model isa AbstractEDMF && MatrixFields.has_field(Y, @name(c.ρtke))
+        ᶜl_mix = p.precomputed.ᶜl_mix
+        ᶜl_mix .= ᶜmixing_length(Y, p)
+        return ᶜl_mix
+    elseif turbconv_model isa AbstractEDMF
+        ᶜscratch = p.scratch.ᶜtemp_scalar_6
+        ᶜscratch .= ᶜmixing_length(Y, p)
+        return ᶜscratch
```
szy21:
```suggestion
    if turbconv_model isa AbstractEDMF
        ᶜl_mix = p.precomputed.ᶜl_mix
        ᶜl_mix .= ᶜmixing_length(Y, p)
        return ᶜl_mix
```
The comment is not accurate here because `has_field(Y, @name(c.ρtke))` doesn't determine whether TKE is prognostic or not (use_prognostic_tke does, we actually always have Y.c.ρtke). But to make things simpler I think we can always have `ᶜl_mix` in the cache when `turbconv_model isa AbstractEDMF`.

src/cache/precomputed_quantities.jl
szy21: Also, do we only need to allocate these fields for `AbstractEDMF?`

src/parameterized_tendencies/microphysics/cloud_fraction.jl
szy21: And not related to this PR, given that we are materializing mixing length now, maybe we can change ᶜmixing_length to return a materialized field instead of a lazy broadcasting. But we don't need to change it now.

src/prognostic_equations/edmfx_tke.jl
```diff
@@ -51,37 +37,52 @@ function edmfx_tke_tendency!(
...
-        # kinetic energy through the buoyancy term of the subdomain momentum
-        # equations, which the prognostic subdomain velocities already carry;
-        # adding it here double-counts buoyancy production and spuriously
-        # inflates K near cloud tops with active drafts.
-        @. Yₜ.c.ρtke -= Y.c.ρ * ᶜK_h * ᶜlinear_buoygrad
-    end
+    use_prognostic_tke(turbconv_model) || return nothing
+    edmfx_tke_sources!(Yₜ, Y, p)
+    return nothing
+end
```
szy21: These two functions can be combined (use_prognostic_tke(::EDOnlyEDMFX) is always true).

(Maybe I will remove the use_prognostic_tke option... let me think)

src/diagnostics/core_diagnostics.jl
```diff
@@ -203,15 +203,15 @@ function compute_lmix(state, cache, _)
         return ᶜmixing_length(state, cache)
```
szy21: This can return `ᶜlmix` in the cache now.

src/diagnostics/default_diagnostics.jl
tapios: @costachris? Do we want this separately?

src/cache/temporary_quantities.jl
tapios: REmoved.

.buildkite/Manifest-v1.11.toml
tapios: Done

src/cache/precomputed_quantities.jl
tapios: I think we may need the zeros here: with the interface entrainment closure off (`EDMF_interface_entr_efficiency == 0`), `set_face_diffusivities!` does not touch `ᶠK_entr`, but it's still read when assembling the face diffusivity. Added a comment. Let me know if I'm missing something.

tapios: And, yes, allocating only for AbstractEDMF is done.

src/prognostic_equations/edmfx_tke.jl
tapios: Done

src/diagnostics/core_diagnostics.jl
tapios: Done.

src/cache/precomputed_quantities.jl
szy21: Could you set ᶠK_entr to 0 when EDMF_interface_entr_efficiency == 0? It's generally safer not to set a value when building the cache (and can break restarts).

src/prognostic_equations/eddy_diffusion_closures.jl
```diff
@@ -414,27 +420,402 @@ function mixing_length_lopez_gomez_2020(
...
+        val_energy_containing = Val{:energy_containing}()
+        @. ᶠK_entr = interface_entrainment_diffusivity(
+            ᶠbuoygrad,
+            ᶠΔz,
+            ᶠκ,
+            get_mixing_length_field(ᶠml, val_energy_containing),
+            c_b,
+            A_entr,
+        )
+    end
```
szy21, line 747:
```suggestion
    else
        @. ᶠK_entr = 0
    end
```
like this

src/parameterized_tendencies/microphysics/cloud_fraction.jl
tapios: Good point. Changed it.

src/prognostic_equations/eddy_diffusion_closures.jl
tapios: Ok. Thank you. Will do.

szy21:
Could you rename ᶜlinear_buoygrad to ᶜbuoygrad, and ᶜbuoygrad_stab to ᶜN²_eff (I don't like that many superscript in this variable name, but that's up to you), or the other way around, so it is clear they have the same physical meaning as ᶠbuoygrad and ᶠN²_eff, just different construction?

src/cache/precomputed_quantities.jl
tapios: Done

tapios:
> Could you rename ᶜlinear_buoygrad to ᶜbuoygrad, and ᶜbuoygrad_stab to ᶜN²_eff (I don't like that many superscript in this variable name, but that's up to you), or the other way around, so it is clear they have the same physical meaning as ᶠbuoygrad and ᶠN²_eff, just different construction?

Done

src/diagnostics/default_diagnostics.jl
costachris: Taking it out of default diagnostics is fine I think, since we can still output it if needed when  debugging.

costachris (approved):

szy21 (approved):

PR #4648, opened 2026-07-07
Fixes energy  bookkeeping in EDMF sedimentation fluxes

tapios:
The prognostic-EDMF subdomain corrections to the grid-mean sedimentation
energy flux in `vertical_advection_of_water_tendency!` had two
inconsistencies:

1. **Sign error on the environment correction.** The environment term
   entered with an extra `-1` relative to the updraft term, even though
   both fluxes use the same `WVector(-(w·q))` convention. The environment
   enthalpy correction therefore entered the energy budget with the wrong
   sign.

2. **Kinetic-energy double counting.** The grid-mean reference energy was
   defined as `e_int - Kin`, while the grid-mean flux being corrected
   carries `e_int + Φ + Kin`. The correction thus re-added the grid-mean
   kinetic energy instead of replacing it with the subdomain value.

Both corrections are now defined per subdomain as

    (e_intᵏ + Kinᵏ) - (e_int + Kin),

with `Kinʲ = Kin(wʲ, uʲ)` for the updraft and `Kin(w, u⁰)` for the
environment (no separate environment terminal velocity is stored, since
the environment mass flux is the residual `ρqw - ρaʲqʲwʲ`). Both terms now enter 
with the same sign convention as the grid-mean flux.

## To do:
- Bump ref_counter when rebased and ready to merge


src/prognostic_equations/water_advection.jl
```diff
@@ -110,26 +123,35 @@ function vertical_advection_of_water_tendency!(Yₜ, Y, p, t)
...
+                        p.scratch.ᶜtemp_scalar_3,
                     ),
                 )
-            @. p.scratch.ᶜtemp_scalar_3 = e_int_func(thp, ᶜT⁰) - p.scratch.ᶜtemp_scalar_2
+            # Environment correction: (e_int⁰ + Kin⁰) - (e_int + Kin). The
+            # environment sedimentation velocity is not stored separately
+            # (the environment mass flux is the residual ρqw - ρaʲqʲwʲ), so
+            # Kin⁰ is approximated with the grid-mean terminal velocity ᶜw.
+            @. p.scratch.ᶜtemp_scalar_3 =
+                e_int_func(thp, ᶜT⁰) + $(Kin(ᶜw, ᶜu⁰)) -
```
trontrytel, line 164: Just a note, that maybe if we are thinking of running it at different resolutions with updraft area not always small, we should rethink using grid mean vertical velocity here?

trontrytel (approved):
Thank you!

src/prognostic_equations/water_advection.jl
tapios: Agreed. I left a TODO for later.

PR #4649, opened 2026-07-07
Use scratch fields for microphysics quantities

petebachant:
Was an 11% bump in performance yesterday, but TBD against latest `main`.


petebachant:
Yay, compat hell:

image

PR #4652, opened 2026-07-07
Precompute quadrature-invariant microphysics inputs to cut kernel cost

petebachant:
In the words of Mr. Claude:

>The SGS-quadrature 1-moment microphysics broadcast (set_microphysics_tendency_cache!, PrognosticEDMFX path) is register-bound, and much of its cost is fixed per grid point rather than in the quadrature loop. It recomputed several quantities inline at every point: the lazy environment inputs (air_density and the specific_env_value reconstructions for q_lcl/q_icl/q_rai/q_sno), and the quadrature-invariant liquid fraction and linearized saturation-excess mean.
>
>Materialize those into scratch fields once (in cheap, high-occupancy kernels) and pass the precomputed λ and mu_S into microphysics_tendencies_1m, so the heavy kernel reads them instead of recomputing them. Numerically identical; large speedup on the AMIP 1M prognostic-EDMF config (env microphysics kernel roughly halved).

This is an 11% speedup for the flagship AMIP (prog. EDMF 1M), though it's kind of hard to test because of dependency conflicts.

image


sajjadazimi (approved):
Looks good, thank!

trontrytel (approved):
Thank you!

PR #4653, opened 2026-07-07
Jacobian changes: Exner-form PGF linearization, entrainment relaxation, and sedimentation in the continuity row

tapios:
Three changes to the manual sparse Jacobian to make it discretely consistent with the implicit tendency:

- Exner-form PGF rows. The `u₃` rows previously linearized the old `∇p/ρ + ∇Φ` form while the tendency uses `cp_d (θ_v − θ_vr) ∇Π + ∇(Φ − Φ_r)`. The rows now carry the exact product-rule linearization (turned out to be essentially identical to what
we had).

- Entrainment relaxation diagonals. The updraft-scalar diagonals now include the feedback through the relaxation target, scaling `−(ε + ε_turb)` by `(1 + w ρaʲ/ρa⁰)` (the derivative used to diagnose the environment value) and include the 2M number concentrations `(n_lcl, n_rai)`.

- Sedimentation in the continuity row. vertical_advection_of_water_tendency! adds the same sedimentation flux divergence to `ρ` as to `ρq_tot`, but only the latter had Jacobian entries. Added `(ρ, ρq_x)` blocks; since the scalar solve ordered `ρ` first, the scalar `BlockLowerTriangularSolve` is reordered to `[condensate masses] → [ρ] → [ρq_tot] → [rest]` (the always-zero `(ρq_x, ρ)` keys are removed, keeping the solve exactly block-triangular). (It's cost-neutral computationally.)

## To-Do

- [x] Update `ref_counter` after rebase


src/prognostic_equations/implicit/manual_sparse_jacobian.jl
```diff
@@ -656,9 +792,21 @@ function update_jacobian!(alg::ManualSparseJacobian, cache, Y, p, dtγ, t)
...
+            if !(
+                ρχ_name in (
+                    @name(c.ρq_lcl),
+                    @name(c.ρq_icl),
+                    @name(c.ρq_rai),
+                    @name(c.ρq_sno),
+                )
+            )
+                ∂ᶜρχ_err_∂ᶜρ = matrix[ρχ_name, @name(c.ρ)]
+                @. ∂ᶜρχ_err_∂ᶜρ = zero(typeof(∂ᶜρχ_err_∂ᶜρ))
```
sajjadazimi: Why do we have these blocks. It seems they are set to zero.

sajjadazimi (approved):
Looks good, thank! I also checked derivatives and they look correct.

src/prognostic_equations/implicit/manual_sparse_jacobian.jl
tapios: They were there originally because they were allocated. However, they can be removed by extending the allocation filter, now done.

PR #4654, opened 2026-07-08
Publish reproducibility references on merge instead of via a main build

nefrathenrici:
Previously, publishing a new reproducibility reference required a full main build that re-ran the entire simulation suite just to move its output into the reference store. This eliminates that: the reference is published from the bundle the PR's CI already computed shortly after the PR merges.

### How it works

1. Move reproducibility bundle to the staging folder at the end of a build after all tests have passed.
2. When CI runs on the main branch (on merge), run one job to publish the bundle in the staging folder.

See the README for full info.

Notes:
- If in a merge queue, the final merge queue build will publish the reproducibility bundle immediately. 
- This needs branch protection requiring PRs to be up to date with main, so the PR-build state matches the merged state.
- Every buildkite job group now needs an `if: 'build.branch != "main"'` guard. We can simplify this by splitting the "publish bundle on main" step into a different file.


tapios (approved):
Looks good. Thanks for doing this. It'll help. 

Rather than wrapping each build group in `if: 'build.branch != "main"'` to escape it on main, can't we just exit on main after moving the reproducibility files? Or wrap the entire rest of the build group in one conditional, rather than repeating it?

nefrathenrici:
> Rather than wrapping each build group in `if: 'build.branch != "main"'` to escape it on main, can't we just exit on main after moving the reproducibility files? Or wrap the entire rest of the build group in one conditional, rather than repeating it?

I don't think it's possible to exit early / cancel a build from the results of a step since each step is independent. The only way to apply the conditional to the whole build group is by wrapping them in one `group`, flattening all of the existing groups because groups can't be nested. I think the cleanest way to split the behavior is to make a new pipeline file and change the [pipeline upload step](https://buildkite.com/clima/climaatmos-ci/settings/steps) to upload that one when running on the main branch. I'd rather do that in a followup PR in case we need to revert this.

PR #4655, opened 2026-07-08
Apply upwind corrections post-Newton

sajjadazimi:
## Purpose 
Implicit vertical advection reverts to central-only. The upwind correction is applied through ClimaTimeSteppers' new T_post_imp! hook, so the upwind direction is taken with respect to the Newton-solved velocity rather than the initial guess.

----


src/prognostic_equations/implicit/implicit_tendency.jl
```diff
@@ -220,3 +220,41 @@ function implicit_vertical_advection_tendency!(Yₜ, Y, p, t)
...
+    correct_implicit_advection_tendency!(Yₜ, Y, p, t)
+
+Post-Newton upwind correction to the central-differenced implicit vertical
+advection of `ρe_tot` and `ρq_tot` in
+[`implicit_vertical_advection_tendency!`](@ref). Called by ClimaTimeSteppers
+as the `T_post_imp!` hook on `ClimaODEFunction`: evaluated at the
+Newton-solved stage state `U*` and applied as `U ← U* + dtγ · Yₜ`.
+
+Writes `vtt_upwind - vtt_central` for `ρe_tot` (and `ρq_tot` when a
+non-dry microphysics model is active). All other fields of `Yₜ` are zero.
```
tapios:
```suggestion
Writes `vtt_upwind - vtt_central` for `ρe_tot` (and `ρq_tot` when available). All other fields of `Yₜ` are zero.
```

src/prognostic_equations/implicit/implicit_tendency.jl
```diff
@@ -220,3 +220,41 @@ function implicit_vertical_advection_tendency!(Yₜ, Y, p, t)
...
+Evaluating the correction *after* Newton — rather than folding it into
+the implicit tendency — means the upwind direction is taken with respect
+to the Newton-solved velocity, avoiding the "wrong-cell" upwinding that
+occurs when the sign of `ᶠu³` flips between the initial guess and the
+Newton solution (a real concern with `max_iters = 1`).
+"""
+NVTX.@annotate function correct_implicit_advection_tendency!(Yₜ, Y, p, t)
+    Yₜ .= zero(eltype(Yₜ))
+    (; energy_q_tot_upwinding) = p.atmos.numerics
+    energy_q_tot_upwinding == Val(:none) && return nothing
```
tapios: This is ok. But with `energy_q_tot_upwinding == Val(:none)`, we still enter the post_imp path in every stage, don't we, just doing nothing? Maybe good to elide at the hook with something like `T_post_imp! = energy_q_tot_upwinding == Val(:none) ? nothing : correct_implicit_advection_tendency!`.

src/prognostic_equations/implicit/implicit_tendency.jl
tapios, line 74: This problem is not new here, but it is perpetuated below for van Leer. Is there some way to fix it? I guess it's yet another CTS change, and perhaps we should take it on in a separate PR.

tapios (approved):
Looks good. But no real test. Consider adding a simple test such as corrected + central == full upwind on a small column.

src/prognostic_equations/implicit/implicit_tendency.jl
sajjadazimi: We can cache dtgamma = float(dt) * gamma in p at ODE-function construction (similar to p.dt). We can then read from p on the implicit path. So probably no need to update CTS. Let's do it in a separate CA PR.

PR #4657, opened 2026-07-09
feat(edmfx): add horizontal component to the EDMFX SGS diffusive flux

haakon-e:
Add the horizontal component of the previously vertical-only EDMFX environment SGS diffusive flux. 

Adds `edmfx_sgs_horizontal_diffusive_flux` option (default `false`) for high-res configs (e.g., the gray-zone to LES regime) where the horizontal grid spacing is comparable to or smaller than the environment mixing length.

Water and enthalpy follow the approach in #4753. Water diffuses as a single substance `q_tot_eff = q_tot - q_rai - q_sno`. Enthalpy flux on the form `ρ K_{h,h} (∇ₕs_d + (h_eff + Φ) ∇ₕq_tot_eff)`  (`h_eff` the mass-weighted specific enthalpy of the suspended water). 

The momentum tendency is the horizontal divergence of `τ = -2 K_{u,h} S` calculated using the full strain rate
- applied to u,v,w and with TKE shear production from horizontal gradients. 

Note that the stress and the production are not energetically consistent with each other since the work done by the stress includes cross terms such as `S₁₃ ∂ₓw` that the production `2 ρ K_{u,h} |S_h|²` does not.

Anisotropy enters only through the mixing-length grid-scale limiter.
- added `grid_scale` kwarg to `ᶜmixing_length`, which defaults to the resolvability filter scale 
- the horizontal path passes the horizontal node scale, so that `l_h = min(l_phys, Δx_h)`. 
- diffusivities are cached once per explicit cache update (`ᶜK_u_h`/`ᶜK_h_h`)
- new diagnostics: `lmixh`, `edth`, and `evuh`.

For prognostic updrafts, the `edmfx_horizontal_diffusion` option (default `false`) applies the grid-mean horizontal specific tendencies uniformly to the updraft scalars, matching the uniform vertical treatment of `edmfx_vertical_diffusion` (#4720). 
- the `ρaⱼ` dry-mass counterpart of the water tendency is omitted, matching `edmfx_sgs_diffusive_flux_tendency!`, where it was removed in #4551.

Added docs describing the math in `docs/src/edmf_horizontal_diffusion.md`. 
Added a few unit tests too.


haakon-e:

- #4657 ◀
    - #4702
        - #4704

src/prognostic_equations/eddy_diffusion_closures.jl
```diff
@@ -463,6 +468,85 @@ function ᶜmixing_length(Y, p, property::Val{P} = Val{:master}()) where {P}
...
+  - `buoyancy_gradient`: stability input of the mixing length and the turbulent
+    Prandtl number. By default, the centered `ᶜlinear_buoygrad`.
+  - `ᶜK_u_field`, `ᶜK_h_field`: output fields for `K_u` and `K_h`. By default
+    `nothing`, in which case the corresponding value is lazy.
+
+# Returns
+
+  - `(; ᶜK_u, ᶜK_h)`.
+"""
+function ᶜeddy_diffusivities!(
```
costachris: The  center-K consolidation looks like it collides with #4647, which moved everything to face-native `set_face_diffusivities!` (`ᶠK_h`/`ᶠK_u`/`ᶠK_entr`). (https://github.com/CliMA/ClimaAtmos.jl/blob/dba91cab240cb9606b7cdbdaced14f6b24ec285e/src/prognostic_equations/eddy_diffusion_closures.jl#L655 on main)

src/prognostic_equations/eddy_diffusion_closures.jl
```diff
@@ -421,20 +423,23 @@ end
 @inline get_mixing_length_field(ml::MixingLength, ::Val{:buoy}) = ml.buoy
 @inline get_mixing_length_field(ml::MixingLength, ::Val{:l_grid}) = ml.l_grid
 
-function ᶜmixing_length(Y, p, property::Val{P} = Val{:master}()) where {P}
+function ᶜmixing_length(
```
costachris, line 866: Both this PR and #4647 edit `ᶜmixing_length`, but differently. Here you add `grid_scale` / `buoyancy_gradient` so the horizontal path can pass `Δx` and a centered buoyancy gradient. On main, it's written around `resolvability_filter_scale` and a split `ᶜN²_eff` / `ᶜbuoygrad` without keywards kwargs. So this can be consolidated

src/prognostic_equations/edmfx_sgs_flux.jl
```diff
@@ -361,3 +329,103 @@ function edmfx_sgs_diffusive_flux_tendency!(
...
+    nothing
+
+function edmfx_sgs_horizontal_diffusive_flux_tendency!(
+    Yₜ, Y, p, t, turbconv_model::Union{EDOnlyEDMFX, PrognosticEDMFX},
+)
+    p.atmos.edmfx_model.sgs_diffusive_flux_horizontal isa Val{true} ||
+        return nothing
+    iscolumn(axes(Y.c)) && return nothing
+    (; params) = p
+    (; ᶜlinear_buoygrad) = p.precomputed
```
costachris: `ᶜlinear_buoygrad` got renamed to `ᶜbuoygrad`

src/prognostic_equations/edmfx_sgs_flux.jl
```diff
@@ -361,3 +329,103 @@ function edmfx_sgs_diffusive_flux_tendency!(
...
+    ᶜρ = Y.c.ρ
+    ᶜtke = @. lazy(specific(Y.c.ρtke, ᶜρ))
+
+    # Mixing length limited by the horizontal node spacing, with the centered
+    # buoyancy gradient as the stability input
+    Δx = Spaces.node_horizontal_length_scale(Spaces.horizontal_space(axes(Y.c)))
+    ᶜK = ᶜeddy_diffusivities!(
```
costachris: The vertical path adds `ᶠK_entr` into all its diffusivities now, the horizontal terms should likely not include it after rebasing .

src/prognostic_equations/edmfx_sgs_flux.jl
```diff
@@ -361,3 +329,103 @@ function edmfx_sgs_diffusive_flux_tendency!(
...
+        Y, p;
```
costachris: We recompute the horizontal diffusivities here + in the updraft tendency + a few times in these diagnostics. Since main precomputes face K's anyway, could we precompute `ᶜK_u_h`/`ᶜK_h_h` once?

haakon-e:
This is implemented. The grid-mean horizontal specific tendencies are applied uniformly to the updraft scalars, behind the `edmfx_horizontal_diffusion` option, which requires `edmfx_sgs_horizontal_diffusive_flux` and is rejected at construction if set alone.

src/prognostic_equations/eddy_diffusion_closures.jl
haakon-e: I think this is resolved now. I kept a single `grid_scale` kwarg, but it now defaults to `resolvability_filter_scale(axes(Y.c))`, so the default path is exactly the formulation on main (there's a unit test asserting the identity). The horizontal path just passes the horizontal grid spacing in `set_horizontal_diffusivities!(Y, p)`:
```julia
ᶜl_h = ᶜmixing_length(Y, p; grid_scale = Δx_h)
```
The `buoyancy_gradient` kwarg is dropped, so the horizontal path reads the same `ᶜN²_eff` and strain rate from the cache as the vertical closure.

src/prognostic_equations/eddy_diffusion_closures.jl
haakon-e: The face-native vertical pipeline from #4647 (`set_face_diffusivities!`, `ᶠK_h`/`ᶠK_u`/`ᶠK_entr`) is untouched, and the horizontal path gets its own `set_horizontal_diffusivities!(Y, p)` that caches center-based `ᶜK_u_h`/`ᶜK_h_h` once per explicit cache update. 

These coefficients are center-fields because the horizontal operators themselves (`gradₕ`, `wdivₕ`) act at cell centers. the vertical fluxes stay on faces. Both are built from the same closure functions and stability inputs, so there's no duplicated physics.

src/prognostic_equations/edmfx_sgs_flux.jl
```diff
@@ -403,3 +403,124 @@ function edmfx_sgs_diffusive_flux_tendency!(
...
+    # Momentum: horizontal weak divergence of the SGS stress `τ = -2 K_u S`
+    # with the full strain rate. The vertical stress divergence is handled by
+    # the vertical diffusion pathway.
+    ᶠρ = @. p.scratch.ᶠtemp_scalar = ᶠinterp(ᶜρ)
+    ᶜτ_h = compute_strain_rate_center_full!(p.scratch.ᶜtemp_UVWxUVW, ᶜu, ᶠu)
+    @. ᶜτ_h = -2 * ᶜK_u_h * ᶜτ_h
+    @. Yₜ.c.uₕ -= C12(wdivₕ(ᶜρ * ᶜτ_h) / ᶜρ)
+    ᶠτ_h = compute_strain_rate_face_full!(p.scratch.ᶠtemp_UVWxUVW, ᶜu, ᶠu)
+    @. ᶠτ_h = -2 * ᶠinterp(ᶜK_u_h) * ᶠτ_h
+    @. Yₜ.f.u₃ -= C3(wdivₕ(ᶠρ * ᶠτ_h) / ᶠρ)
```
haakon-e, line 623: I include `u₃` here because the horizontal flux of vertical momentum `u'w'` is a covariance. The vertical counterpart `w'w'` is a variance and is already in TKE / mass flux / pressure, which is why the vertical pathway doesn't diffuse `u₃`. Some additional explanation in the new docs page.

src/prognostic_equations/eddy_diffusion_closures.jl
```diff
@@ -872,6 +908,29 @@ Summands are combined in a fixed order: dry static energy, then vapor, liquid, i
...
+
+Write the horizontal gradient of total enthalpy in dry-static-energy +
+water-enthalpy form, `∇ₕs_d + Σ_μ (h_μ + Φ) ∇ₕq_μ` for `μ ∈ {vap, liq, ice}`,
+into the center field `ᶜ∇h` and return it.
+
+Summands are combined in a fixed order: dry static energy, then vapor, liquid,
+ice. The gradient is accumulated in two broadcasts; see the horizontal EDMF
+diffusion documentation page.
+"""
+function ᶜtotal_enthalpy_gradientₕ!(
```
haakon-e: This mirrors `ᶠtotal_enthalpy_gradientᵥ`, but I materialize into scratch in two broadcasts. A single fused broadcast with four spectral gradients exceeds GPU kernel parameter limits on extruded spaces with warped topography (the lazy form also needs CliMA/ClimaCore.jl#2551). There's a regression test that the `ρe_tot` tendency equals the sum of the four constituent fluxes.

src/prognostic_equations/eddy_diffusion_closures.jl
```diff
@@ -846,6 +861,27 @@ function ᶜmixing_length(Y, p, property::Val{P} = Val{:master}()) where {P}
...
 end
 
+"""
+    set_horizontal_diffusivities!(Y, p)
+
+Compute and cache the horizontal eddy viscosity `ᶜK_u_h` and eddy diffusivity
+`ᶜK_h_h` of the TKE-based closure, with the mixing length limited by the
+horizontal node spacing, `l_h = min(l_phys, Δx_h)`.
+"""
+function set_horizontal_diffusivities!(Y, p)
```
haakon-e, line 925: I don't include `ᶠK_entr` here, since my understanding is that it parameterizes vertical entrainment across an unresolved inversion face (`K_e = γ w_e Δz`) and has no horizontal analogue.

haakon-e (commented):
a few comments to assist review

src/utils/utilities.jl
```diff
@@ -206,6 +206,21 @@ function compute_strain_rate_face_vertical(ᶜu)
...
+"""
+    compute_strain_rate_center_horizontal(ᶜu)
+
+Compute the strain rate at cell centers from velocity at cell centers, with horizontal gradients only.
+"""
+function compute_strain_rate_center_horizontal(ᶜu)
+    axis_uvw = Geometry.UVWAxis()
+    return @. lazy(
+        (
+            Geometry.project((axis_uvw,), gradₕ(UVW(ᶜu))) +
```
costachris, line 273: minor, but `gradₕ(UVW(ᶜu)))` is computed twice in the lazy broadcast. Maybe it can be materialized like in `compute_strain_rate_center_full!`?  Or otherwise maybe `compute_strain_rate_center_full!` can be converted to lazy for consistency. Not too critical right now

src/prognostic_equations/edmfx_sgs_flux.jl
```diff
@@ -403,3 +403,124 @@ function edmfx_sgs_diffusive_flux_tendency!(
...
+        @. ᶜρχₜ_diffusion = wdivₕ(ᶜρ * ᶜK_h_h * gradₕ(ᶜq_tot))
+        @. Yₜ.c.ρq_tot += ᶜρχₜ_diffusion
+        @. Yₜ.c.ρ += ᶜρχₜ_diffusion
+        if apply_sgs_updraft
+            # The updraft dry-air mass is unchanged by the water flux.
+            for j in 1:n
+                ᶜρaʲ = Y.c.sgsʲs.:($j).ρa
+                ᶜq_totʲ = Y.c.sgsʲs.:($j).q_tot
+                @. Yₜ.c.sgsʲs.:($$j).q_tot += ᶜρχₜ_diffusion / ᶜρ
+                @. Yₜ.c.sgsʲs.:($$j).ρa +=
```
costachris: Note we're including the `ρa` correction for the horizontal tendencies but not the vertical one (https://github.com/CliMA/ClimaAtmos.jl/blob/023a8e9024945d660a4146e8a609fa8fa364c370/src/prognostic_equations/edmfx_sgs_flux.jl#L347). This should be a small term so maybe it should be neglected?  At the least the difference should be noted.

docs/src/edmf_horizontal_diffusion.md
```diff
@@ -0,0 +1,133 @@
...
+
+## Horizontal eddy diffusivity
+
+The horizontal eddy viscosity and diffusivity use the same TKE-based closure as the vertical flux,
+
+```math
+K_u = c_m \, l \, \sqrt{e}, \qquad K_h = K_u \,/\, \mathrm{Pr}_t ,
+```
+
+where ``l`` is the horizontal mixing length (see [Mixing length](@ref)), ``e`` is the environment turbulent kinetic energy (TKE), ``c_m`` is the eddy-diffusivity coefficient, and ``\mathrm{Pr}_t`` the turbulent Prandtl number.
```
costachris: The earlier equations uses \(K_u = c_m\, l\, \sqrt{e}\) and says `l` is the horizontal mixing length, but later `l` is the master/vertical length and (`l_h`) is horizontal. Same for \(K_u\)/\(K_h\) vs cached `ᶜK_u_h`/`ᶜK_h_h`.

Can you line up the symbols in the later sections with the earlier ones?

src/prognostic_equations/edmfx_sgs_flux.jl
```diff
@@ -403,3 +403,124 @@ function edmfx_sgs_diffusive_flux_tendency!(
...
+
+    # Turbulent TKE transport, and shear production from horizontal gradients;
+    # the production from vertical gradients is applied in the TKE tendency.
+    if use_prognostic_tke(turbconv_model)
+        @. Yₜ.c.ρtke += wdivₕ(ᶜρ * ᶜK_u_h * gradₕ(ᶜtke))
+        ᶜS_h = compute_strain_rate_center_horizontal(ᶜu)
+        @. Yₜ.c.ρtke += 2 * ᶜρ * ᶜK_u_h * norm_sqr(ᶜS_h)
+    end
+
+    # Momentum: horizontal weak divergence of the SGS stress `τ = -2 K_u S`
```
costachris, line 614: Above the TKE shear production uses horizontal strain `S_h`.
Whereas here momentum stress uses the full 3D strain rate.

Those don’t match energy-wise. Is there an approximation or assumption being made? If intentional can you note it in the docs; if not, should `τ` be built from `S_h` as well?

costachris (approved):

src/utils/utilities.jl
haakon-e: Copied the style of `compute_strain_rate_face_vertical` just above, which does the same double `ᶠgradᵥ(UVW(ᶜu))`. So the lazy helpers are consistent with each other and only `compute_strain_rate_center_full!` materializes.

I'd be happy to change both in a separate PR.

src/prognostic_equations/edmfx_sgs_flux.jl
haakon-e: Removed, matching the vertical path.
Note that hyperdiffusion still applies the correction.

Would probably be good to separately test turning on/off for both spatial directions in AMIP or similar.

docs/src/edmf_horizontal_diffusion.md
haakon-e: I made `l`, `K_u`, `K_h` the master/vertical quantities throughout, with `l_h`, `K_{u,h}`, `K_{h,h}` in the horizontal.

That keeps `l` consistent with the rest of the docs, and matches the cached `ᶜK_u_h`/`ᶜK_h_h`.

src/prognostic_equations/edmfx_sgs_flux.jl
haakon-e: That's intentional and now noted in the docs.

`τ` uses the full strain because `S₁₃` and `S₂₃` are the horizontal flux of vertical momentum `<u_h'w'>`, which is the reason `u₃` is diffused horizontally at all. Building `τ` from `S_h` would remove exactly that term.

Shear production uses `S_h` because `2 ρ K_u |S_h|²` is positive definite. The exact source, `τ : ∇_h u`, is sign-indefinite and can drive TKE negative.

So right, they don't balance (the cross-terms is the difference). I'll have to follow up with some positivity constraint.

PR #4658, opened 2026-07-09
feat(edmfx): add the momentum component of the horizontal SGS diffusive flux

haakon-e:
Applies the horizontal weak divergence of the SGS stress `τ = -2 K_u S`, with the full three-dimensional strain rate of the grid-mean velocity, to the grid-mean momentum: on cell centers for the horizontal wind and on cell faces for the vertical wind, mirroring the Smagorinsky–Lilly stress split by divergence direction. The eddy viscosity is the horizontal `K_u` from the TKE-based closure with the mixing length limited by the horizontal node spacing. Enabled under the existing `edmfx_sgs_horizontal_diffusive_flux` option; the vertical stress divergence remains with the vertical diffusion pathway.

Adds the corresponding TKE shear production from horizontal gradients, `2 ρ K_u S_h : S_h` with `S_h` the strain rate built from horizontal gradients only, so the resolved kinetic energy sink from the horizontal stress lands with its TKE source in the same PR. The term is positive definite; the production from vertical gradients and its stencil are unchanged, mirroring the decoupled Smagorinsky–Lilly split in which each directional norm is built from that direction's gradients.

Extends the tests: horizontally uniform velocity produces no momentum tendency, perturbed velocity produces one, the production term is positive definite under uniform TKE, and the single-column tendency is identically zero.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,11 @@ ClimaAtmos.jl Release Notes
 main
 ----
 
+- [#4658](https://github.com/CliMA/ClimaAtmos.jl/pull/4658) ![][badge-✨feature/enhancement]
+  Add the momentum component of the horizontal EDMFX SGS diffusive flux — the
+  horizontal weak divergence of the stress `τ = -2 K_u S` with the full strain
+  rate — and the corresponding TKE shear production from horizontal gradients,
+  applied under the same `edmfx_sgs_horizontal_diffusive_flux` option.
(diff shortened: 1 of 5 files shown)


haakon-e:

- #4657
    - #4658 ◀
        - #4659

haakon-e:
Folded into #4657: the momentum stress and its TKE production counterpart land with the scalar fluxes so each PR is energetically self-contained.

PR #4660, opened 2026-07-09
feat(timestepping): add horizontal acoustic substepping

haakon-e:
## What

Add an optional horizontal acoustic-substepping mode (`AcousticMultirate`): the fast horizontal acoustic terms and the grid-mean momentum advection sub-cycle at a small sub-step while the slow dynamics and physics advance with the outer step, so the expensive physics is evaluated once per outer step. The generic step-exchange multirate (frozen forcing, inner IMEX sub-cycle, `LieSplitOuter`/`TrapezoidalSplitOuter` combination) comes from CliMA/ClimaTimeSteppers.jl#442; this PR supplies the acoustic fast tendency, the forcing freeze, the sub-step count and divergence damping (both defaulting from the horizontal acoustic CFL), the implicit operator, and the config wiring. The vertically implicit solve is unchanged, and `acoustic_substeps: 0` (default) leaves existing behavior unchanged.

The sub-cycle advances the vector-invariant momentum advection as a pair, the kinetic-energy gradient and the rotational (vorticity and Coriolis) momentum flux. Sub-cycling only the gradient leaves a non-conservative residual that grows a shear-driven vertical mode, negligible on weak-wind boxes but destabilizing for the sphere baroclinic jet.

## Dependencies

CliMA/ClimaTimeSteppers.jl#442 and ClimaUtilities >= 0.1.31 (CliMA/ClimaUtilities.jl#232, exact `ITime` division). CI resolves once both are released; the branch was validated locally against develop pins (session artifacts, not part of this branch).

## Verification

- Unit suite (`test/acoustic_substepping.jl`): the tendency-split identity (fast tendency plus frozen forcing reconstructs the full explicit tendency at the freeze state, for both vertical treatments), the divergence-damping forms, config resolution, a state-update check, a timestep sweep, and the exact sub-step-count rounding.
- Conservation: the dry box conserves mass and energy at the reference budget level (`box_acoustic_substep_conservation`).
- Sphere: the dry baroclinic wave (h_elem 6) at dt 800 s with the auto sub-step count completes one day and tracks the baseline extrema, in shallow and deep geometries.
- Quasi-equilibrium: the RCEMIP-II gray-zone 2 km box at dt 10 s (outer Courant 1.74) runs 50 days to a stationary radiative-convective moisture state; the implicit-split and no-split arms agree to 0.2% in column water at day 20, and the ARS343 dt 2 s baseline dries along the same trajectory, so the pre-equilibrium drift is spin-up rather than the integrator. Throughput at the matched 20-day horizon is 3.3x the baseline (sypd 0.60 versus 0.184). The 533 m box holds 5 days; the Bomex prognostic-EDMF box reaches steady shallow convection.

The soak evidence was generated on the earlier ClimaAtmos-hosted driver; the split identity and conservation checks tie it to this CTS-consumer driver (same fast tendency, same frozen forcing), and the throughput is expected to carry over rather than re-measured here.

## Scope

- Active hyperdiffusion limits the outer step to about three times the baseline; a coefficient clamp is proposed as follow-up work.
- Configurations with implicit physics re-solve those terms every sub-step; restricting the sub-cycle implicit solve to the acoustic block is follow-up work.
- Terrain at large outer steps is untested.
- An `ITime` outer step must divide exactly into sub-steps; `auto` rounds the resolved count up to the next such divisor.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -525,6 +525,14 @@ steps:
           --job_id box_density_current_test
         artifact_paths: "box_density_current_test/output_active/*"
 
+      - label: ":package: Non-spherical: 3D box, dry density current, acoustic substepping, conservation"
+        retry: *retry_policy
+        command: >
+          julia --color=yes --project=.buildkite .buildkite/ci_driver.jl
+          --config_file $CONFIG_PATH/box_acoustic_substep_conservation.yml
+          --job_id box_acoustic_substep_conservation
+        artifact_paths: "box_acoustic_substep_conservation/output_active/*"
+
       - label: ":package: Non-spherical: Reproducibility test for radiative convective equilibrium in a 3D box"
         retry: *retry_policy
         command: >
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,10 @@ ClimaAtmos.jl Release Notes
 main
 ----
 
+- [#4660](https://github.com/CliMA/ClimaAtmos.jl/pull/4660) ![][badge-✨feature/enhancement] Add an optional horizontal acoustic-substepping timestepping mode (`AcousticMultirate`) that sub-cycles the fast horizontal acoustic terms and the kinetic-energy gradients at a small sub-step while the slow dynamics and physics advance with a larger outer step.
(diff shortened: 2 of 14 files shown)


haakon-e:

- #4660 ◀
    - #4661
        - #4677

PR #4661, opened 2026-07-09
feat(timestepping): restrict the acoustic sub-cycle implicit solve to the acoustic block

haakon-e:
## What

When `acoustic_substep_implicit_split` is enabled, the implicit solve inside the acoustic sub-cycle is restricted to the vertical grid-mean acoustic block: vertical mass, enthalpy, and total-water transport, and the vertical pressure gradient, gravity, and Rayleigh sponge on `u₃`.
The restricted system has a dedicated sparse `AcousticJacobian` with a diagonal scalar block, factorized by the direct block-arrowhead solver.
The remaining implicit terms (EDMF SGS, vertical diffusion, implicit microphysics, sedimentation) are solved once per outer step by an outer implicit complement that ClimaTimeSteppers sequences around the sub-cycle.
The full implicit tendency, its Jacobian, and the mode-off and unsplit paths are unchanged.

## Why

With EDMF or implicit diffusion active, the unsplit sub-cycle re-solves the full implicit operator every sub-step, which removes the throughput benefit in the configurations the mode targets.
On a Bomex EDMF box the restricted Jacobian has 25 blocks instead of 59, and the sub-step cost drops from 56 ms to 8 ms.

## Verification

Dry box (`test/acoustic_substepping_split.jl`): the split reproduces the unsplit sub-cycle; `ρ`, `ρe_tot`, and `u₃` are bitwise identical over the checkpointed steps and `uₕ` agrees to machine precision, for both outer orders.
Bomex EDMF box: the outer solve matches the full-cache solve at Float32 machine precision, with the subdomain state remaining physical.
Conservation: the split conservation job (`box_acoustic_substep_split_conservation`) conserves mass and energy to the reference budget level.

Solution neutrality under the quasi-equilibrium soak: the split arm and the baseline agree to 0.2% of the domain statistics at 20 simulated days on the 2 km RCEMIP-II box, so the split does not change the equilibrium the mode reaches.
As in #4660, the soak was generated on the pre-repartition ClimaAtmos-hosted driver and transfers through the split-equals-unsplit identity and the conservation checks.

## Dependencies

Stacked on #4660.
Same ClimaTimeSteppers #442 and ClimaUtilities >= 0.1.30 dependencies; CI cannot resolve them until they are released, and the branch was validated locally against uncommitted `.buildkite` develop pins (session artifacts).

## Scope

Conditioning of the outer implicit solve at large outer steps and GPU behavior are untested.
The second-order outer combination performs two outer solves per step.


haakon-e:

- #4660
    - #4661 ◀
        - #4677

PR #4662, opened 2026-07-09
docs: add a recipe for localizing a deferred GPU KernelException

haakon-e:
## What

Adds a recipe to `docs/dev-guides/workflow/debugging.md` for localizing a device-side `throw` (a `DomainError` from an out-of-domain `sqrt`/`log`/`x^y`, a bounds error, a `SingularException`) that surfaces on the host as a generic `CUDA.KernelException` with no kernel name and no device line. A row in the "silently-wrong values" lookup table points to the new section.

## Why

The exception is deferred: CUDA detects it at the next synchronization, so the host stacktrace names that synchronization point, not the throw, and the named line often contains no transcendental operation at all. Compiling with device line info (`-g2`) is not an option here because ptxas rejects the non-ASCII identifiers common in CliMA kernels. The section documents the practical alternative: force a synchronous throw with `CUDA_LAUNCH_BLOCKING=1`, read the reported line correctly, bisect with `ClimaComms.@sync` barriers paired with `@info` markers, restart from a crash-proximate checkpoint to keep iterations short, and confirm the candidate site with a floor or clamp.

## Scope

Documentation only; no source changes. The page lives under the shared `docs/dev-guides/`, so the recipe is kept repo-agnostic. No docs build is required (CI covers it).


haakon-e:
Dev-guide content belongs in CliMA/DeveloperGuides; replaced by CliMA/DeveloperGuides#18.

PR #4664, opened 2026-07-09
fix(setups): GPU-compatible single-column setups and finer hydrostatic-pressure integration

haakon-e:
Initializes the AtmosphericProfilesLibrary single-column setups (Bomex, DYCOMS, GABLS, GATE_III, ISDAC, Larcform1, PrecipitatingColumn, Rico, Soares, ShipwayHill2012, TRMM_LBA) on GPU spaces, and integrates their hydrostatic pressure on a finer grid.

These setups build host-resident profile interpolants, so broadcasting the initial-condition closure on a device fails. `initial_condition_field` now broadcasts the closure on the device when it is isbits after adapting to the device array type, and otherwise evaluates it on the host and copies the result to the device. Interpolant setups take the host path, so no interpolant enters a device kernel, while analytic setups keep their device broadcast. The setup API is unchanged, with profiles remaining plain callables of height and no GPU-specific code.

`hydrostatic_pressure_profile` integrates the initial value problem on a 1000-element column instead of 100, returning a `ClimaInterpolations.Interpolation1D.Interpolate1D`. The finer grid removes the interpolation error of the coarser 300 m spacing, changing initial center pressures by about `2e-4` to `4e-4` relative, within about `4e-6` of a reference on a 16 times finer grid. This changes the reproducibility reference, so the counter is bumped; the difference is a chaotic decorrelation of the convective fields with the thermodynamic climate preserved (see the comment below).

Also: `ρ_from_profile` is hoisted to a top-level function; the ShipwayHill2012 and TRMM_LBA `θ` and `q_tot` profiles use `ClimaInterpolations` interpolants; and the `ShipwayHill2012` constructor returns the setup type (its interface methods were previously unreachable). A new buildkite GPU test evaluates the pressure profiles on a column and a stretched box for Float64 and Float32, with exact values checked in `test/larcform1.jl`.

### DIFF
--- a/.buildkite/full_pipeline.yml
+++ b/.buildkite/full_pipeline.yml
@@ -483,6 +483,17 @@ steps:
         agents:
           slurm_gpus: 1
 
+  - group: "Setups"
+    steps:
+      - label: ":crown: Setup hydrostatic pressure profiles are GPU-compatible"
+        retry: *retry_policy
+        command: >
+          julia --color=yes --project=.buildkite test/gpu_setups.jl
+        env:
+          CLIMACOMMS_DEVICE: "CUDA"
+        agents:
+          slurm_gpus: 1
+
   - group: "Conservation Tests"
     steps:
       - label: ":computer: Conservation tests: Dry baroclinic wave, no sources"
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,7 @@ ClimaAtmos.jl Release Notes
 
 main
 ----
+- [#4664](https://github.com/CliMA/ClimaAtmos.jl/pull/4664) ![][badge-🔥behavioralΔ] Initialize the AtmosphericProfilesLibrary single-column setups (Bomex, DYCOMS, GABLS, GATE_III, ISDAC, Larcform1, PrecipitatingColumn, Rico, Soares, ShipwayHill2012, TRMM_LBA) on GPU spaces. `hydrostatic_pressure_profile` integrates the hydrostatic initial value problem on a dedicated 1000-element column (previously 100) and returns a `ClimaInterpolations` interpolant over host arrays. Because the setup profiles are host-resident interpolants, the initial condition of these setups is evaluated on the host and copied to the device rather than broadcast on the device. Initial center pressures change by about `2e-4` to `4e-4` relative, dominated by the removed interpolation error of the coarser grid; the new profiles are within about `4e-6` relative of a reference solution on a 16 times finer grid. The `ShipwayHill2012` constructor now returns the setup type (its interface methods were previously unreachable).
 
 0.42.2
 -------
--- a/reproducibility_tests/ref_counter.jl
+++ b/reproducibility_tests/ref_counter.jl
@@ -1,4 +1,4 @@
-391
+392
 
 # **README**
 #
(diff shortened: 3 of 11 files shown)


haakon-e:

- #4664 ◀
    - #4712

haakon-e:
@akshaysridhar this should be ready now, whenever you have a change to look at it.

akshaysridhar:
Can we add a GPU test for this ? I still see 

```
ERROR: Scalar indexing is disallowed.
...
 [18] column_indefinite_integral(f::Function, ϕ₀::Float32, zspan::Tuple{Float32, Float32}; nelems::Int64)
    @ ClimaAtmos.Setups ~/Codes/ClimaAtmos.jl/src/setups/common/physical_state.jl:184
 [19] column_indefinite_integral
    @ ~/Codes/ClimaAtmos.jl/src/setups/common/physical_state.jl:163 [inlined]
```

with `Julia Version 1.11.8` and deps as

 ```
    [[deps.AtmosphericProfilesLibrary]]
deps = ["ClimaInterpolations", "Interpolations", "LinearAlgebra", "StaticArrays"]
git-tree-sha1 = "3e053ef899873642ac6a87b4a43be4a26333d376"
uuid = "86bc3604-9858-485a-bdbe-831ec50de11d"
version = "0.1.9"
``` 

on 

``` 
[b2c96348] ClimaAtmos v0.41.3 `..
```

akshaysridhar (approved):
As of commit `4fdd24` this compiles and runs on GPU (Julia 1.11.8, A100 device) - LGTM once it's rebased (testing locally against the DYCOMS/BOMEX cases)

haakon-e:
Quick justification for the reproducibility counter bump, since that's the main thing to check before merging.

What changed in the initial state: `hydrostatic_pressure_profile` now integrates the initial hydrostatic balance on a 1000-element column instead of the old 100-element one. That shifts the initial center pressure of the APL single-column setups by about 1e-4 to 4e-4 relative within each model domain. It removes the interpolation error of the coarser grid, and the new profiles are within ~4e-6 relative of a 16x-finer reference solution. So the initial condition really does move, which is why the reference changes and the counter has to go up.

pr_counter_bump_pressure_delta

The more important question is whether that small change to the initial pressure does anything to the simulated climate, or just breaks bit-for-bit reproducibility. These SCM cases are convecting and chaotic, so a tiny perturbation to the initial state grows and the convection decorrelates in phase. The figure below overlays main (N=100) and this PR (N=1000) for the four APL SCM cases in the repro suite, with the time mean as solid lines and a final instantaneous snapshot faintly colored. The thermodynamic profiles `(θ_l, q_t)` overlap in the time mean (`θ_l` to well under 0.1%). The differences are confined to the intermittent convective and cloud fields (cloud liquid, updraft velocity, cloud fraction), and they are larger in the instantaneous snapshots than in the time means, which is what decorrelation looks like rather than a shifted mean state.

pr_counter_bump_overlay

To rule out that the convective-field differences are a systematic response to the pressure change, I blended the two initial pressure profiles, `p_α = (1-α) p_100 + α p_1000`, and ran the four cases for `α in {0, 0.25, 0.5, 0.75, 1}`. If the final statistics scaled with the size of the pressure perturbation, the distance from main (`α=0`) would grow monotonically with `α`, which doesn't happen. The distance is non-monotone in `α` for every field and every case, the intermediate blends are usually the largest outliers, and the full-branch profile (`α=1`) often sits closest to main. `θ_l` and `q_t` overlap for all `α`.  So I think this is chaotic decorrelation from initial condition variations, not biases introduced in this PR.

pr_counter_bump_alpha_sweep

So the counter bump is the reference updating to a more accurate initial pressure. The thermodynamic climate of the cases is preserved and only the phase of the chaotic convection differs, so I think this is safe to merge.

PR #4666, opened 2026-07-09
fix(setups): make ColumnInterpolatableField GPU-compatible

haakon-e:
`ColumnInterpolatableField` was backed by an Interpolations.jl object holding host arrays, so evaluating a hydrostatic pressure profile inside a broadcast over a GPU space failed to compile with a non-isbits kernel argument. This replaces that backing with a linear interpolation over knot and value vectors stored as `SVector`s, making the interpolant an isbits value that embeds directly into a device broadcast kernel; evaluation uses `searchsortedfirst` with flat extrapolation. CPU results are bitwise-identical to the previous Interpolations.jl implementation, and the profile now evaluates correctly inside a GPU broadcast, verified end-to-end for a Bomex-style pressure profile on an A100. Stacked on #4664, which removed the closure-boxing kernel-compilation failure in `hydrostatic_pressure_profile`; this was the remaining blocker for building these initial conditions on GPU.

### DIFF
--- a/src/setups/Setups.jl
+++ b/src/setups/Setups.jl
@@ -1,5 +1,6 @@
 module Setups
 
+import StaticArrays as SA
 import ClimaCore.Geometry as Geometry
 import ClimaCore: Fields
 import Thermodynamics as TD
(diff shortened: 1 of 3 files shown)


haakon-e:
Folded into #4664, which now carries both the closure-boxing fix and the ColumnInterpolatableField interpolation fix in a single PR. Closing as superseded; the branch is retained.

PR #4667, opened 2026-07-09
feat(timestepping): remove the acoustic-substepping resonance source and generalize the divergence damping

haakon-e:
## What and why

Acoustic substepping freezes the slow forcing over an outer step and sub-cycles the acoustic system. The
kinetic-energy gradient sits in that frozen forcing, and `K = |u|²/2` oscillates at the acoustic
frequency, so freezing it injects an acoustic-frequency signal into the sub-cycle once per outer step —
the drive of a parametric resonance. The resonance grows fastest at the wavenumber where the acoustic
phase advances by π per outer step, and that wavenumber migrates to larger scales as the outer step grows,
so divergence damping tuned to the grid scale delivers only ∝ 1/n_sub there and cannot close the
stability window as the outer step increases.

This PR removes the source instead of damping it: the kinetic-energy and reference-relative geopotential
gradients are re-evaluated at every sub-step (on both `uₕ` and `u₃`), so they follow the sub-cycled state
rather than acting as a fixed periodic source. It also generalizes the divergence damping to a
three-dimensional form (the full-velocity divergence, exact over terrain) and a perturbation form
(deviation from the sub-cycle-start divergence), and resolves the damping coefficient automatically from
the kinetic-energy treatment.

`acoustic_substeps: 0` is unchanged; all changes live in the substepping driver, the config resolution,
and one new tendency function.

## Configuration

| Key | Default | Meaning |
|---|---|---|
| `acoustic_substep_kinetic_energy` | `fast` | `fast` re-evaluates the kinetic-energy and geopotential gradients every sub-step; `slow` holds them frozen. |
| `acoustic_substep_damping_form` | `3d` | `3d` \| `3d_perturbation` \| `horizontal` — the divergence damped. |
| `acoustic_substep_damping` | `auto` | `auto` resolves to 0.4 (`fast`) or 1.5 (`slow`); a number keeps its direct value. |

The PR2-validated configuration is recovered exactly with `slow` / `horizontal` / `1.5`.

## Verification

Dry Float32 340 m box (node spacing ≈ 113 m), the resonance benchmark from the falsification experiment,
now with the final code (the `u₃` kinetic-energy term active and the 3-D damping form). Baseline stable
outer step dt_base = 0.3 s.

**Stability window** (order 2, 80 steps). Stable β_d ranges:

| variant | 2× | 4× | 6× |
|---|---|---|---|
| `fast/3d` (default) | [0, 1.5] | [0, 1.5] | [0, 1.5] |
[...]

PR #4668, opened 2026-07-09
feat(timestepping): scale the hyperdiffusion coefficient with the acoustic-substepping outer step

haakon-e:
## What

Add the config key `acoustic_substep_hyperdiffusion_scaling` (`auto` | `Real`, default `auto`), which scales the vorticity hyperdiffusion coefficient under acoustic substepping.

- With `acoustic_substeps: 0` the key is ignored and the coefficient is unchanged, so the default configuration and hyperdiffusion-off runs are byte-identical to `main`.
- Under substepping, `auto` multiplies the coefficient by `s = min(1, Δt_hd_limit / (safety·Δt_outer))` with `safety = 2` and `Δt_hd_limit = 2·Δx_node / (F·ν₄_vorticity_coeff·β⁴)`, where `F = max(divergence_damping_factor, 1/prandtl_number)` (read from the parsed config) and `β = 4` is a maximum-wavenumber prefactor calibrated for degree-3 spectral elements (`Δt_hd_limit ≈ 0.95 s` at `Δx_node = 113 m`).
- A real value `r` scales the coefficient by exactly `r` (unclamped); `r = 1` reproduces today's behavior even under substepping.

The factor is folded into the stored coefficient once at simulation construction (`get_simulation`), where `Δx_node`, `dt`, and the acoustic configuration are available. The hyperdiffusion tendency code (`hyperdiffusion.jl`) is untouched, and every channel that derives from the vorticity coefficient — the divergent channel (`×divergence_damping_factor`) and the scalar channel (`/prandtl_number`) — is scaled with it.

## Why

Under substepping the hyperdiffusion tendency lives in the frozen slow forcing, so it is integrated once per outer step as a forward-Euler filter. Its explicit stability limit therefore caps the outer-step multiple. On the dry box the resonance is de-sourced to a 6× ceiling (PR #4667), but the frozen hyperdiffusion binds first, at 3×.

Because the hyperdiffusion is a numerical filter rather than a physical mixing term, its coefficient can be scaled with the outer step: reducing `ν₄` lowers the frozen forward-Euler limit's binding and lifts the ceiling in exact proportion to `1/ν₄`, while the per-step filter strength `ν₄ k⁴ Δt` is held roughly fixed. The falsification experiment measured the ladder on the dry box (`ν₄` scale → ceiling): 1.0 → 3×, 0.5 → 6×, 0.33 → ≥8× (the hyperdiffusion-off ceiling), with the accuracy guard clean (ρ/ρe relative error 2–4×10⁻⁵ vs a matched-filter fine reference, conservation at the budget floor, grid-scale roughness bounded and monotone) and the no-substepping baseline unharmed at 1×. The principled split is that numerical filters are scaled with the timestep; the physical Smagorinsky and EDMF mixing coefficients are never scaled (both were measured not to bind: Smagorinsky ≥2.86×, EDMF horizontal diffusion ~115×).

## Verification

- Unit tests (dynamics group, `test/acoustic_substepping.jl`): auto-resolution arithmetic (clamps to 1 at small `Δt_outer`, matches the hand-computed factor at a large `Δt_outer`), real pass-through, and the recovery identities (the scaling key with `acoustic_substeps: 0` and `r = 1` under substepping both leave the stored coefficient unchanged). 15/15 pass; the pre-existing 20/20 implicit-split tests are unaffected.
- Dry-box confirmation of the shipped `auto` path (not a coefficient override): _pending — table to follow._
- Moist gray-zone spot-check with hyperdiffusion on + `auto` scaling: _pending (the un-draft gate)._

## Scope and caveats

- Draft. The evidence is single-box (dry density-current box, one resolution, Float32); resolution independence of the ceiling is inferred from `Δt_hd_limit ∝ Δx_node`, not re-measured at a second resolution.
- The filter's job under vigorous moist convection at reduced strength is the un-draft validation: the moist gray-zone spot-check confirms no grid-scale noise buildup and conservation at the operating point.
- The `β⁴` prefactor is calibrated from the degree-3 dry box; other spectral-element degrees would need recalibration or `λ_max` computed from the operator.
- Auto-scaling applies to config-driven construction (`get_simulation`); the programmatic `AtmosSimulation` constructor path, where coefficients are set explicitly, is unaffected.
- Physical mixing coefficients (Smagorinsky, EDMF) are never scaled.


haakon-e:
Superseded by #4673.

PR #4671, opened 2026-07-09
fix(setups): make ShipwayHill2012 and TRMM_LBA profiles GPU-compatible

haakon-e:
Replace the ad hoc Interpolations.jl interpolants in the ShipwayHill2012 initial condition (θ and q_tot) and the TRMM_LBA q_tot profile with ClimaInterpolations.Interpolation1D.Interpolate1D over SVector knots, following the pattern of #4664; the interpolants become isbits and evaluate inside a device broadcast, so these setups' center initial conditions can be built on a GPU space.

The CI environment manifest moves to AtmosphericProfilesLibrary 0.1.9, which provides isbits data-backed profiles (tke and sounding data) for the setups audited in #4664. The existing compat entry already admits it, and the pinned calibration experiment environments are unchanged.

Stacked on #4664. Validated on an A100: the audited setups construct isbits and the affected initial conditions evaluate in a GPU broadcast with CPU values unchanged.

### DIFF
--- a/.buildkite/Manifest-v1.11.toml
+++ b/.buildkite/Manifest-v1.11.toml
@@ -146,10 +146,10 @@ uuid = "56f22d72-fd6d-98f1-02f0-08ddc0907c33"
 version = "1.11.0"
 
 [[deps.AtmosphericProfilesLibrary]]
-deps = ["Interpolations", "LinearAlgebra"]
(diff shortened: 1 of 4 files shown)


haakon-e:

- #4664
    - #4671 ◀

PR #4675, opened 2026-07-09
Add read-only depot that gets rebuilt periodically

### LINKED ISSUE #4612: Race conditions using the shared depot cause job failures
I have recently seen more jobs flake due to missing folders/files. This seems to happen when one job overwrites/rebuilds a file in the depot that another job is reading or compiling against. I have tracked down a specific example of this caused by GPUCompiler, I will dig up previous errors to see if they match this pattern.

Sample jobs, I will extend this:
- https://buildkite.com/clima/climaatmos-ci/builds/30415#019f01f6-df97-489e-ad3b-3b938cb82791
- https://buildkite.com/clima/climacoupler-ci/builds/9185#019f05b2-836e-4c3c-b67d-79107d7a78d6

nefrathenrici:
This PR adds a read-only depot that gets rebuilt every night to stay up-to-date with the `.buildkite` Manifest. This is meant to enable the current quick precompilation while preventing failures caused by concurrent writes to the shared depot. Closes #4612 

## Corea idea:
Each build reads from the shared depot, and write any changes it needs to its own depot. This is controlled via the environment variable:
```
  JULIA_DEPOT_PATH: "${BUILDKITE_BUILD_PATH}/${BUILDKITE_PIPELINE_SLUG}/${BUILDKITE_BUILD_NUMBER}/depot:${BUILDKITE_BUILD_PATH}/${BUILDKITE_PIPELINE_SLUG}/depot/shared"
```
See [Julia docs](https://pkgdocs.julialang.org/dev/depots/#Setting-up-shared-depots-for-multi-user-systems) for more.

The depot is managed by the new `depot_pipeline.yml`, which runs nightly at https://buildkite.com/clima/atmos-depot. The three most recent depots are stored. When a new depot is built, it is swapped out atomically so this should not cause errors in CI.


szy21 (approved):

.buildkite/pipeline.yml
```diff
@@ -26,20 +26,19 @@ steps:
...
     key: "init_cpu_env"
-    concurrency: 1
-    concurrency_group: "depot/climaatmos-ci"
     command:
-      - "echo $$JULIA_DEPOT_PATH"
-
+      - echo $$JULIA_DEPOT_PATH
       - echo "--- Instantiate .buildkite"
-      - "julia --project=.buildkite -e 'using Pkg; Pkg.instantiate(;verbose=true); Pkg.precompile(;strict=true); Pkg.status()'"
+      - julia --project=.buildkite -e 'using Pkg; Pkg.instantiate(;verbose=true, update_registry=false); Pkg.precompile(;strict=true); Pkg.status()'
```
ph-kev, line 41: Why do we need `update_registry = false`? Is it because we can't write to it anymore?

.buildkite/build_depot.sh
```diff
@@ -0,0 +1,30 @@
...
+set -euo pipefail
+
+STAGING="$JULIA_DEPOT_PATH"                 # Julia builds here
+mkdir -p "$STAGING"
+
+# Start from the last published depot so we only build the delta and stay
+# self-contained. First run has no `shared` symlink -> build from scratch.
+if [ -e "$SHARED_DEPOT_ROOT/shared" ]; then
+  cp -a "$(readlink -f "$SHARED_DEPOT_ROOT/shared")/." "$STAGING/"
+  chmod -R u+w "$STAGING"                   # published copy is read-only
```
ph-kev, line 14: I don't understand too well how Julia package manager work, but does this work if you are using a specific branch of a package's repo? (e.g. `add PkgName@blah`)

.buildkite/pipeline.yml
nefrathenrici: I think this is needed to prevent the instantiate step from attempting to write to the read-only depot, which would cause the job to fail.

.buildkite/build_depot.sh
nefrathenrici: This will compile the `.buildkite` env on main branch HEAD. A branch will be able to use the precompiled packages in the depot that match the .buildkite env Manifest, other packages will be precompiled separately in the per-build depot.

PR #4676, opened 2026-07-10
refactor(timestepping): host the acoustic substepping driver in ClimaTimeSteppers

haakon-e:
This reworks the acoustic substepping driver to use the step-exchange `Multirate` family in ClimaTimeSteppers (CliMA/ClimaTimeSteppers.jl#442) instead of a hand-written timestepping loop. The frozen-forcing pass, the inner IMEX sub-cycle, the outer combination, and the outer implicit complement sequencing are now in ClimaTimeSteppers as the `LieSplitOuter` and `TrapezoidalSplitOuter` outer family. This file supplies the acoustic fast tendency, the forcing-freeze operation, the sub-step count, the restricted implicit operator, the outer complement, and the config wiring.

The public surface is unchanged. `AcousticMultirate`, `AcousticSubstepTendency`, `acoustic_slow_forcing!`, the divergence-damping types and forms, `reference_sound_speed`, and `auto_n_sub` keep their names and signatures. The driver's `init_cache` now builds a `SplitODEProblem` and a `Multirate(inner_alg, LieSplitOuter | TrapezoidalSplitOuter)` and returns the ClimaTimeSteppers cache wrapped in a thin `AcousticMultirateCache` that exposes the full model function for the tests and diagnostics.

The driver tracks the ClimaTimeSteppers step-exchange API: the outer-complement time refinement uses `CTS.refine_time` (renamed from `refine_ns`), and `init_cache` rounds the resolved sub-step count (auto or requested) up to the next count that divides `dt` exactly, since `CTS.sub_timestep` now throws on an `ITime` `dt` that the count does not divide at nanosecond resolution. `test/acoustic_substepping.jl` gains a unit test for this exact-division behavior; `test/acoustic_substepping_split.jl` is unchanged.

This is a preview PR that depends on the ClimaTimeSteppers PR (CliMA/ClimaTimeSteppers.jl#442). The `.buildkite` develop of the local ClimaTimeSteppers is an uncommitted artifact and is not part of this branch; it becomes a compat bump once a ClimaTimeSteppers release includes the step-exchange family.

Both acoustic suites pass. The implicit-split test confirms the split sub-cycle reproduces the unsplit one bitwise in `ρ`, `ρe_tot`, and `u₃` at both outer orders. One behavior changes relative to the pre-exactness driver: because `CTS.sub_timestep` now requires the sub-step count to divide `dt` at nanosecond resolution, a resolved count that does not divide `dt` is rounded up to the nearest count that does (the dry box's three sub-steps at `dt = 0.5 s` become four) instead of being silently truncated, so the result matches the earlier driver bitwise only where the count already divides `dt`.


haakon-e:

- #4660
    - #4661
        - #4676 ◀
            - #4677

haakon-e:
Closing: this rehosting refactor is absorbed into the re-partitioned stack. #4660 now implements the substepping mode as a ClimaTimeSteppers step-exchange consumer from the start (the LieSplitOuter/TrapezoidalSplitOuter driver, plus the exact_n_sub helper and its unit test), and #4661 adds the implicit split on that CTS-hosted base. The CTS.refine_time / CTS.sub_timestep hooks this PR adopted are dropped: CliMA/ClimaTimeSteppers.jl#442's final form removes those hooks, so the consumer uses plain ITime arithmetic (t + dt self-refines) and ClimaUtilities >= 0.1.30 exact ITime division (dt / n_sub). #4677 is rebased onto #4661.

PR #4677, opened 2026-07-10
feat(timestepping): forward-backward acoustic inner sub-step scheme

haakon-e:
This adds a forward-backward inner sub-step scheme for acoustic substepping, selected by `acoustic_substep_scheme: forward_backward`. It is a cost feature at matched stability: it replaces the IMEX-ARK inner sub-cycle with a cheaper split-explicit advance, so each acoustic sub-step is roughly three times cheaper while the outer combination, the frozen slow forcing, the restricted implicit operator, and the divergence-damping coefficient are unchanged. The default `acoustic_substep_scheme: imex` is byte-identical to the current driver.

The scheme is `AcousticForwardBackward`, a ClimaAtmos-defined `ClimaTimeSteppers.TimeSteppingAlgorithm` used as the inner algorithm of the acoustic `Multirate`. It is the second consumer of the step-exchange inner-stepper slot added in CliMA/ClimaTimeSteppers.jl#442; the first is the IMEX-ARK inner of the substepping driver in #4660. This PR is stacked on #4661. The dependency chain is #442 then #4660 then #4661 then this, and it also requires ClimaUtilities >= 0.1.30.

Each sub-step of size Δτ advances the horizontal momentum forward with the horizontal pressure gradient, the horizontal kinetic-energy gradient, and the frozen slow momentum forcing; advances the scalars and the vertical kinetic-energy gradient against the updated momentum (backward), with the frozen slow forcing; performs one off-centered vertically-implicit Newton iteration on the vertical grid-mean acoustic block, reusing the restricted block-arrowhead Jacobian at Δt_γ = θ Δτ; and applies a time-adjusted divergence filter on the horizontal velocity at the post-solve state. The off-centering weight θ = (1 + β_off) / 2 is set by `acoustic_substep_off_centering` (default β_off = 0.1). Off-centering is required rather than a margin: the divergence filter damps only the horizontal velocity, so the vertical density-vertical-velocity acoustic oscillator is damped only by θ > 1/2.

The forward-backward inner calls `constrain_state!` explicitly and nulls `initialize_imp!`, so the analytic EDMF SGS stage solve does not fire inside the acoustic sub-cycle; with prognostic EDMF the SGS implicit terms are carried by the outer half-step, which is why the scheme requires `acoustic_substep_implicit_split: true` there. Three `check_config` errors enforce the supported surface: `forward_backward` requires `acoustic_substep_vertical: implicit`, requires `acoustic_substep_implicit_split: true` with prognostic EDMF, and rejects the horizontal SEM limiter.

Two fast tendency functions are split into per-block parts (`horizontal_acoustic_tendency!` into scalar and momentum parts, `kinetic_energy_gradient_tendency!` into horizontal-velocity and vertical-velocity parts), with the originals kept as their composition, so the IMEX path is unchanged and the existing exact-split identity test covers the refactor. The frozen forcing reaches the horizontal velocity exactly once per sub-step (in the forward momentum step); the backward scalar step advances every other prognostic block, leaving the horizontal-velocity increment as the implicit-diffusion predictor alone (zero for the restricted inner).

The `constrain_state!` threading through the step-exchange inner assembler lives in ClimaTimeSteppers (`init_inner`), alongside `dss!` and `lim!`; it is a no-op for the IMEX inner (whose inner function keeps the default) and load-bearing for the forward-backward inner. As with #4660 and #4661, the `.buildkite` develop pins of the local ClimaTimeSteppers and ClimaUtilities are uncommitted artifacts and are not part of this branch.

Verification covers dry and moist boxes (CPU, Float32 unless noted) and a GPU smoke. On the dry 340 m box the forward-backward scheme matches the IMEX inner's stability window at the shipped defaults across outer steps of 2x to 6x the baseline, with the horizontal-velocity peak within a few percent and mass and energy drift at the Float32 floor. The probe horizon is 120 steps, twice the 60-step campaign standard; the nearest measured crash at 4x is the beta_d=0.8 forward-Euler stability bound at step 2, so every stable point clears the two-times-crash-count probe rule by a wide margin. Mass and energy are conserved to Float64 machine precision on the accuracy box, and the density and total-energy error is the same 1e-5 class as the IMEX inner, with the horizontal-velocity perturbation within the IMEX worst-case bound. The extended split test reproduces the unsplit sub-cycle bitwise for the forward-backward scheme at both outer orders. The per-sub-step cost ratio, measured by differencing ms/step across n_sub at fixed outer step so the fixed per-step cost cancels, is 3.06x on the dry box and 3.44x on a moist 1M box, clearing the 2.5x threshold and the 3x target. On the coarse-Bomex prognostic-EDMF box the forward-backward scheme is stable through 60 steps with physical updraft area fraction at dt=4s. The GPU smoke runs the sub-step loop on an A100 and stays finite.

Two limitations are worth a reviewer's attention.

The forward-backward filter is a post-solve forward-Euler increment rather than an ERK-integrated term, so it is stable only for beta_d CFL_h^2 <= 2; at the auto sub-step the upper beta_d edge sits near 0.8, where the IMEX inner still runs. The shipped default 0.4 is stable with margin, so this narrower upper edge does not affect the default.

The forward-backward inner is a weak-shear-only scheme. Under vertical wind shear it degrades scalar acoustic refraction, and the onset is at realistic boundary-layer shear, not only extreme shear. On the shear-refraction plane benchmark (a log-law wind with a near-surface acoustic pulse, measured by the surface density variance J and the downwind trapping centroid at a fixed horizon, referenced to a fine no-substepping run), the forward-backward scheme matches the IMEX inner with no shear and loses the refraction as soon as shear is present:

| max wind | IMEX J rel. dev. | IMEX centroid | FB J rel. dev. | FB centroid | reference centroid |
| --- | --- | --- | --- | --- | --- |
| 0 (no shear) | 0.36 | +0.0 m | 0.36 | +0.0 m | -0.1 m |
| ~23 m/s | 0.36 | +13.7 m | 9.1 | +3.8 m | +14.4 m |
| ~92 m/s | 0.36 | +54.7 m | 1540 | +1.7 m | +55.6 m |

At about 23 m/s, a realistic boundary-layer maximum wind, the forward-backward surface density variance is already an order of magnitude too large and the trapping centroid collapses to roughly a quarter of the reference; at about 92 m/s it is catastrophic. The velocity field tracks the IMEX inner in every case (the covariant horizontal-velocity peak differs by a few percent), so this is a scalar refraction-accuracy degradation, not a velocity instability.

The degradation is structural, a property of the inner stage composition rather than a term-placement or solver-tuning defect. With the sub-step size, the restricted Jacobian, the frozen forcing, and every other ingredient held identical to the IMEX inner, one converged single-theta acoustic solve per sub-step cannot represent the horizontal-velocity, vertical-velocity, and scalar phase coupling that shear-mediated refraction requires at per-sub-step acoustic CFL 0.5 to 0.8, while the three-stage IMEX inner at the same sub-step size can. Every term-placement and solver hypothesis was falsified: the horizontal and vertical kinetic-energy gradients are already explicit in the forward and predictor stages; raising the inner Newton count to two or three leaves J unchanged while eroding the cost win to 2.3x and 1.8x; freezing the vertical kinetic-energy gradient is indistinguishable from the base; a uniform wind at the same peak speed sits in the IMEX band, so the failure requires shear specifically; and off-centering only modulates the error, since a centered (Crank-Nicolson) solve still fails by roughly twenty times.

The consequence is scope, not a pending fix. The scheme is a documented weak-shear-only inner for quiescent and weakly-sheared regimes, and #4677 is kept in draft on that basis rather than pending a shear repair. A design-level rework to a two-solve predictor-corrector forward-backward inner would land near a 1.5x per-sub-step cost ratio, marginal against the IMEX inner, and is not pursued here. The cost win of the shipped single-solve inner is real and retained: 3.1x to 3.4x per sub-step at one Newton iteration.

Off-centering was measured non-binding on the anisotropic box tested; the 0.1 default stands as the vertical-mode margin from the theta-scheme analysis. Horizontal SEM limiters and the vertical-explicit path are rejected by check_config.

### DIFF
--- a/config/default_configs/default_config.yml
+++ b/config/default_configs/default_config.yml
@@ -89,6 +89,12 @@ acoustic_substep_damping:
 acoustic_substep_implicit_split:
   help: "Solve the non-acoustic implicit terms once per outer step instead of every acoustic sub-step (inner/outer implicit split). Requires `acoustic_substep_vertical: implicit`."
   value: false
+acoustic_substep_scheme:
+  help: "Inner sub-step scheme for acoustic substepping [`imex` (default): the IMEX-ARK inner sub-step; `forward_backward`: a forward-backward inner sub-step with an off-centered vertically-implicit solve and an end-of-sub-step divergence filter, roughly three times cheaper per sub-step]. `forward_backward` requires `acoustic_substep_vertical: implicit`."
+  value: "imex"
+acoustic_substep_off_centering:
+  help: "Off-centering weight beta_off in [0, 1] for the forward-backward vertically-implicit solve, theta = (1 + beta_off) / 2. Only used by `acoustic_substep_scheme: forward_backward` [`0.1` (default)]."
+  value: 0.1
 krylov_rtol:
   help: "Relative tolerance of the Krylov method (only for ClimaTimeSteppers.jl; only used if `use_krylov_method` is `true`)"
   value: 0.1
--- a/docs/src/acoustic_substepping.md
+++ b/docs/src/acoustic_substepping.md
@@ -108,6 +108,27 @@ The first-order outer combination performs one outer solve per step; the second-
 The split is additive: the full implicit tendency and its Jacobian are unchanged, and the restricted operator duplicates the acoustic subset rather than extracting it, so the mode-off and unsplit paths are unchanged.
 The split requires `acoustic_substep_vertical: implicit`.
 
+## Forward-backward inner sub-step
+
+By default each sub-step is an IMEX-ARK inner sub-cycle (`acoustic_substep_scheme: imex`).
+The forward-backward scheme (`acoustic_substep_scheme: forward_backward`) replaces it with a cheaper split-explicit advance, following the practice of split-explicit dynamical cores (Klemp, Skamarock, and Dudhia 2007).
+Each sub-step of size ``\Delta\tau``
+
+1. advances the horizontal momentum forward with the horizontal pressure gradient, the horizontal kinetic-energy gradient, and the frozen slow momentum forcing;
+2. advances the scalars and the vertical kinetic-energy gradient against the updated horizontal momentum (backward), together with the frozen slow forcing;
+3. performs one off-centered vertically-implicit Newton iteration for the vertical grid-mean acoustic block, reusing the restricted block-arrowhead Jacobian at ``\Delta t_\gamma = \theta \Delta\tau``;
+4. applies a time-adjusted divergence filter on ``u_h`` at the post-solve state.
+
+The off-centering weight is ``\theta = (1 + \beta_\mathrm{off}) / 2``, set by `acoustic_substep_off_centering` (default ``\beta_\mathrm{off} = 0.1``, ``\theta = 0.55``); ``\theta = 1`` is backward Euler.
+Off-centering is required, not a margin: the divergence filter damps only ``u_h``, so the vertical ``\rho'``-``u_3`` acoustic oscillator is damped only by ``\theta > 1/2``.
+
+The scheme is a cost feature at matched stability: it reuses the same fast tendency, off-centered solve, and divergence-damping coefficient as the IMEX inner, replacing three implicit solves, three Jacobian factorizations, and roughly ten direct-stiffness summations per sub-step with one, one, and two.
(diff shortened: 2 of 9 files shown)


haakon-e:

- #4660
    - #4661
        - #4677 ◀

PR #4678, opened 2026-07-10
Reparameterize TKE dissipation via a critical Richardson number (c_d = c_m c_b / Ri_c)

tapios:
## Summary

In the buoyancy-limited branch of the mixing-length closure (`l = l_N`), local TKE balance reduces to a threshold Richardson number for TKE production:`Ri_c = c_m·c_b/c_d`. Only this combination of the three mixing-length coefficients controls the stable-regime cutoff for turbulence maintenance. Calibrating `(c_m, c_b, c_d)` independently leads to correlated parameters and possibly unphysical `Ri_c`. The current default values (`c_m = 0.56`, `c_b = 0.4`, `c_d = 0.22`) imply `Ri_c ≈ 1.0`, above the literature range of 0.2–0.3, without that being visible.

This PR reparameterizes to `(c_m, c_b, Ri_c)` and **derives** `c_d = c_m·c_b/Ri_c`.

## Why this basis is better conditioned

Since `c_m/c_d = Ri_c/c_b`, neutral-equilibrium TKE (`∝ c_m/c_d`) becomes independent of `c_m`. Each parameter then is more directly tied to an observable:
- `Ri_c`: stability cutoff for turbulence maintenance (direct literature priors);
- `c_b`: partitions TKE amplitude against the buoyancy length;
- `c_m`: flux-magnitude scaling.

## Implementation

- `Ri_c` is read from the **existing** ClimaParams parameter `mixing_length_Ri_crit` (default 0.25); the parameter otherwise is unused across CliMA FWIK.
- `c_d` is derived in `TurbulenceConvectionParameters` construction. Explicit `tke_diss_coeff` overrides still take precedence; **TOML overrides of `mixing_length_diss_coeff` no longer have any effect**; set `mixing_length_Ri_crit` instead.
- All TOMLs overriding the coefficients are converted: the `c_m = 0.56` EDMF family pins `Ri_c = 1.0181818181818183` (preserving `c_d = 0.22` and making the implied high `Ri_c` of the current calibration explicit); the calibrated experiment TOMLs replace their `c_d` values with the equivalent `Ri_c`; the `gcm_driven_scm` calibration priors now target `mixing_length_Ri_crit` (mean 0.25, same relative width as the old `c_d` prior; a prior range 0.2-0.5 is reasonable).
- Unit tests cover the derivation, the TOML override path.

## Behavior

- Configurations using the converted TOMLs: preserved.
- Configurations at ClimaParams defaults: derived `c_d = 0.14·0.4/0.25 = 0.224` instead of 0.22 (+1.8 % dissipation): tiny behavioral change, ref counter bumped.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,10 @@ ClimaAtmos.jl Release Notes
 
 main
 ----
+- [#4678](https://github.com/CliMA/ClimaAtmos.jl/pull/4678) ![][badge-🔥behavioralΔ] Reparameterize the TKE dissipation coefficient in terms of a critical Richardson number: `c_d = c_m·c_b/Ri_c`, with `Ri_c` read from the ClimaParams parameter `mixing_length_Ri_crit` (default 0.25). In the buoyancy-limited branch of the mixing-length closure, local TKE balance reduces to the threshold `Ri_c = c_m·c_b/c_d`, so `(c_m, c_b, Ri_c)` form a nearly orthogonal calibration basis (flux magnitude, buoyancy-length/TKE partition, stability cutoff) and `c_d` is derived. TOML overrides of `mixing_length_diss_coeff` no longer have any effect — set `mixing_length_Ri_crit` instead. At ClimaParams defaults the derived `c_d` is 0.224 instead of 0.22.
+
+0.41.3
+-------
 
 0.41.2
 -------
--- a/src/parameters/Parameters.jl
+++ b/src/parameters/Parameters.jl
@@ -27,7 +27,7 @@ Base.@kwdef struct TurbulenceConvectionParameters{FT, VFT1, VFT2, VTF3} <: ATCP
     max_area::FT
     min_area::FT
     tke_ed_coeff::FT
-    tke_diss_coeff::FT
+    Ri_crit::FT
     tke_surf_scale::FT
     tke_surf_flux_coeff::FT
     diagnostic_covariance_coeff::FT
--- a/calibration/experiments/gcm_driven_scm/prior_diagnostic.toml
+++ b/calibration/experiments/gcm_driven_scm/prior_diagnostic.toml
@@ -23,8 +23,8 @@ type = "float"
 prior = "constrained_gaussian(mixing_length_eddy_viscosity_coefficient, 0.14, 0.07, 0, Inf)"
 type = "float"
 
-[mixing_length_diss_coeff]
-prior = "constrained_gaussian(mixing_length_diss_coeff, 0.22, 0.15, 0, Inf)"
+[mixing_length_Ri_crit]
+prior = "constrained_gaussian(mixing_length_Ri_crit, 0.25, 0.17, 0, Inf)"
 type = "float"
 
 [mixing_length_tke_surf_scale]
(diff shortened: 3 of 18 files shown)


src/parameters/Parameters.jl
```diff
@@ -203,6 +203,51 @@ for fn in fieldnames(TCPS)
...
+This combination of the three coefficients controls the stable-regime
+cutoff, so `(c_m, c_b, Ri_c)` are calibrated and `c_d` follows. The basis is
+nearly orthogonal: `c_m/c_d = Ri_c/c_b`, so the neutral-limit equilibrium
+TKE (`e = 2 (c_m/c_d) l² S²`) is independent of `c_m`, which acts as
+a flux-magnitude scaling, while `c_b` partitions TKE amplitude against the
+buoyancy length and `Ri_c` sets the stability cutoff.
+"""
+tke_diss_coeff(ps::ATCP) =
+    tke_ed_coeff(ps) * static_stab_coeff(ps) / Ri_crit(ps)
+tke_diss_coeff(ps::ACAP) = tke_diss_coeff(turbconv_params(ps))
```
szy21: How about instead of doing this, just calculating tke_diss_coeff from other parameters when it is used? There are only three places. Also, I think it will be less confusing, as if it's here, people may change this parameter and think it will do something. I can remove this parameter in ClimaParams in the next breaking release.

I also think this long comment doesn't belong in this parameter file, but that's just my personal opinion.

szy21: Maybe it's better to centralize it. We can keep this but move the physical explanation to edmfx_tke.jl?

szy21:
@costachris instead of matching previous behavior, I removed the mixing_length_Ri_crit and mixing_length_eddy_viscosity_coefficient in all the tomls so they are just using the ClimaParams. I think it's ok because we need to recalibrate anyway.

szy21 (approved):

costachris (approved):
I see no issues, thanks!

NEWS.md
```diff
@@ -4,6 +4,7 @@ ClimaAtmos.jl Release Notes
 main
 ----
 
+- [#4678](https://github.com/CliMA/ClimaAtmos.jl/pull/4678) ![][badge-🔥behavioralΔ] Reparameterize the TKE dissipation coefficient in terms of a critical Richardson number: `c_d = c_m·c_b/Ri_c`, with `Ri_c` read from the ClimaParams parameter `mixing_length_Ri_crit` (default 0.25). In the buoyancy-limited branch of the mixing-length closure, local TKE balance reduces to the threshold `Ri_c = c_m·c_b/c_d`, so `(c_m, c_b, Ri_c)` form a nearly orthogonal calibration basis (flux magnitude, buoyancy-length/TKE partition, stability cutoff) and `c_d` is derived. TOML overrides of `mixing_length_diss_coeff` no longer have any effect — set `mixing_length_Ri_crit` instead; all repository TOMLs and the `gcm_driven_scm` calibration priors are converted, preserving each configuration's effective `c_d` to within 1 ulp. At ClimaParams defaults the derived `c_d` is 0.224 instead of 0.22.
```
costachris: "preserving each configuration's effective `c_d` to within 1 ulp. ."
Note this is no longer true since overrides in the toml are now deleted. The `mixing_length_eddy_viscosity_coefficient` is quite different between the overrides and the default in ClimaParams, which affects `c_d`

szy21: thanks. I'll update it.

PR #4680, opened 2026-07-10
 Skip merge queue if only one up-to-date PR is being tested 

nefrathenrici:
This PR enables us to skip redundant buildkite builds when the only PR in merge queue has already had CI run on its commit. Merge queue still requires a successful build for status checks, so we still run a minimal build. 

## Content

The large diff is misleading due to a rename, if you look at the commits separately you can see the changes more clearly.

- Rename `pipeline.yml` to `full_pipeline.yml`. This is not necessary but makes it easier to revert.
- Add script `.buildkite/should_skip_merge_queue.sh` which determine whether to skip CI or not. This skips the normal pipeline upload if it is in merge queue, the base commit is on main, the merge group tree is identical to the PR tree, and that the PR has a reproducibility bundle staged.
- Add script `.buildkite/decide_pipeline.sh` which runs the `.buildkite/should_skip_merge_queue.sh` script to decide whether to upload the full pipeline. For now, to ensure it works properly, it just reports when it would have skipped the build, then uploads the full pipeline as before. 
- Add new `pipeline.yml` to run `.buildkite/decide_pipeline.sh`.

See dry run build https://buildkite.com/clima/climaatmos-ci/builds/30795#annotation-skip-ci

### DIFF
--- a/.buildkite/decide_pipeline.sh
+++ b/.buildkite/decide_pipeline.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+# Decider step for non-`main` climaatmos-ci builds (see .buildkite/pipeline.yml).
+#
+# Merge-queue fast path: when the merge group's only change over `main` is a
+# single PR whose reproducibility bundle was already staged by that PR's build
+# (see should_skip_merge_queue.sh), re-running the full pipeline would only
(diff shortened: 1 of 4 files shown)


imreddyTeja (approved):
thanks!

PR #4687, opened 2026-07-13
Skip merge queue for GHA CI

nefrathenrici:
Extends the merge-queue fast path from #4680 (Buildkite) to the docbuild and CI GitHub Actions checks. When a merge group is provably one PR stacked on the current main and byte-identical to a PR head whose checks already passed, the GHA unit-test matrix and doc build are reused instead of re-run. These checks use the same reasoning as the fast path for Buildkite.

To accomplish this, I expanded the ci.yml and docs.yml into 3 jobs each:
- decide: runs the script, outputs boolean skip
- real job (test matrix / docbuild): gated on skip != 'true'.
- ci-required / docs-required: always() gate job that becomes the single required status check. Fails if decide errored or the real job failed, passes on real success or a legitimate reuse.

If this is approved, we will need to update the required checks to `ci-required` and `docs-required`.


imreddyTeja (approved):
Why do the ci and docs workflows run on the merge queue, but not other actions (like downgrade)?

nefrathenrici:
> Why do the ci and docs workflows run on the merge queue, but not other actions (like downgrade)?

~~I think that the merge queue only runs the required status checks, so only prek / ci / docbuild / buildkite run.~~

That was totally wrong, merge queue runs actions that are triggered during merge queue, which is basically all of the normal actions. However I think it only waits to merge until the required checks have passed.

PR #4693, opened 2026-07-15
Different version of precipitation sedimentation

trontrytel:
Include entrainment of environment microphysics species when falling through a narrowing updraft.

### DIFF
--- a/src/prognostic_equations/implicit/manual_sparse_jacobian.jl
+++ b/src/prognostic_equations/implicit/manual_sparse_jacobian.jl
@@ -1199,17 +1199,23 @@ function update_sgs_advection_jacobian!(matrix, Y, p, dtγ)
                 ) - (I,)
 
             # sedimentation
-            # (pull out common subexpression for performance)
+            # Base: a·∂_z(ρwχ) — always the same regardless of ∂a/∂z sign
+            # Correction: min(∂a/∂z, 0)·(ρ¹w¹χ¹ − ρ⁰w⁰χ⁰)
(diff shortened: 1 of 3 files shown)


src/prognostic_equations/implicit/manual_sparse_jacobian.jl
```diff
@@ -1199,17 +1199,20 @@ function update_sgs_advection_jacobian!(matrix, Y, p, dtγ)
...
-                dtγ * ifelse(ᶜ∂a∂z < 0,
-                    -(ᶜprecipdivᵥ_matrix()) ⋅ ᶠsed_tracer_advection *
-                    DiagonalMatrixRow(ᶜa),
+                dtγ * (
                     -DiagonalMatrixRow(ᶜa) ⋅ ᶜprecipdivᵥ_matrix() ⋅
-                    ᶠsed_tracer_advection,
+                    ᶠsed_tracer_advection +
+                    DiagonalMatrixRow(
+                        min(ᶜ∂a∂z, zero(ᶜ∂a∂z)) * ᶜρʲs.:(1) * ᶜwʲ,
+                    )
```
sajjadazimi, line 1218: I think we should decompose `ρ⁰w⁰χ⁰` into grid-mean and updraft variables and take the updraft contribution into account. It will be `O(a)` but consistent with the rest of entrainment block (where this contribution is added via `env_relaxation_feedback`, but here is simpler as the function `specific` is not used).

sajjadazimi (approved):
Thanks! It looks good to me. I just left one minor comment.

src/prognostic_equations/implicit/manual_sparse_jacobian.jl
trontrytel: Done. Thank you!

PR #4702, opened 2026-07-16
refactor(diffusion): route EDMFX vertical diffusive fluxes through shared helpers

haakon-e:
Completes the consolidation from #4699 for the EDMFX vertical path: the grid-mean SGS diffusive fluxes in `edmfx_sgs_diffusive_flux_tendency!` (total enthalpy via the dry static energy + water enthalpy decomposition, specific humidity, and grid-scale tracers with the α and interfacial-entrainment weighting folded into the flux coefficient) now use the shared `ᶠtotal_enthalpy_gradientᵥ` and `ᶜdiffusive_flux_divergenceᵥ`/`ᶜdiffdivᵥ` helpers. 

The TKE flux keeps its dedicated divergence operator, since its bottom boundary has a nonzero surface TKE flux, and the momentum stress keeps its strain-rate form.


haakon-e:
@sajjadazimi This PR probably overlaps with what you're changing with EDMF now, so let me know which you'd prefer to merge first.

haakon-e:

- #4657
    - #4702 ◀
        - #4704

src/prognostic_equations/edmfx_sgs_flux.jl
```diff
@@ -225,7 +225,6 @@ function edmfx_sgs_diffusive_flux_tendency!(
     (; dt, params) = p
     turbconv_params = CAP.turbconv_params(params)
     (; ᶜu) = p.precomputed
-    ᶠgradᵥ = Operators.GradientC2F()
```
costachris, line 226: Removing this local override means the TKE path below (unchanged in this diff, ~L306) now uses the shared ᶠgradᵥ (GradientC2F with SetGradient(0)) instead of a bare GradientC2F(). I think that should be equivalent, but just noting

costachris (approved):

src/prognostic_equations/edmfx_sgs_flux.jl
costachris: Could you update the PR comment and/or commit to say something like "The TKE flux keeps its dedicated divergence operator, but it now also uses the shared `ᶠgradᵥ ` operators" for clarity

haakon-e: Yes, they're equivalent.

`ᶜdivᵥ_ρtke` sets `SetValue(ρtke_flux)` at the bottom and `SetValue(0)` at the top, so `ᶠgradᵥ` (with any BCs) is only evaluated on the interior faces.

The PR body is also updated with a note.

PR #4703, opened 2026-07-16
Unify sgs hyperdiffusion and fix energy hyperdiff

sajjadazimi:
## Purpose 
Unify sgs hyperdiffusion and fix energy hyperdiff

----

### DIFF
--- a/reproducibility_tests/ref_counter.jl
+++ b/reproducibility_tests/ref_counter.jl
@@ -1,4 +1,4 @@
-383
+384
 
 # **README**
 #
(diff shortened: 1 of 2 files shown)


szy21:
Could we add a function as ᶠtotal_enthalpy_gradientᵥ so we can use it in other horizontal diffusion too?

szy21 (approved):

sajjadazimi:
> Could we add a function as ᶠtotal_enthalpy_gradientᵥ so we can use it in other horizontal diffusion too?

@haakon-e is working on it.

PR #4704, opened 2026-07-17
fix(les): diffuse total enthalpy with the dry-static-energy decomposition

haakon-e:
Diffuses total enthalpy in the `Smagorinsky-Lilly` and `AMD` tendencies, vertical and horizontal, with the dry-static-energy + water-enthalpy decomposition already used by the boundary-layer and EDMFX diffusion, instead of diffusing `h_tot` directly. Diffusing `h_tot` contributes a spurious enthalpy flux with the dry-air mass gradient and additionally diffuses kinetic energy through the energy flux. The decomposition removes both, making the energy-moisture coupling of the LES closures consistent with the other diffusion models. 

- The vertical fluxes use `ᶠtotal_enthalpy_gradientᵥ`
- the horizontal fluxes use `ᶜtotal_enthalpy_gradientₕ!` from #4657
- AMD keeps evaluating its energy diffusivities from the total-enthalpy gradient. Only the flux assembly changes

Behavioral for every Smagorinsky/AMD configuration (reference counter bumped).


haakon-e:

- #4657
    - #4702
        - #4704 ◀

haakon-e:
This PR bumps the ref counter. The only affected reproducibility test is a single, short, test that runs RCEMIP with Smagorinsky. Some additional Claude-assisted figures of the ref counter bump is provided below

<details> <summary> click here to see explanation and figures </summary

Justification for the reference counter bump, since that is the main thing to check before merging.

What changed: the Smagorinsky-Lilly and AMD energy tendencies now diffuse total enthalpy through the dry-static-energy + water-enthalpy decomposition instead of diffusing `h_tot` directly. One job in the reproducibility suite is affected, `rcemipii_box_CRM_1M`, which is the only reproducibility configuration that enables `smagorinsky_lilly`. `amd_les` defaults to false and no configuration sets it, so AMD contributes no reference change.

The difference between the two formulations is exact and can be written down. At uniform temperature and uniform kinetic energy,

```
∇h_tot - (∇s_d + Σ_μ (h_μ + Φ) ∇q_μ) = -s_d ∇q_tot,   μ ∈ {vap, liq, ice}
```

so the old formulation includes an additional flux proportional to the dry-air mass gradient. I checked this numerically in the small box the unit tests use: the two formulations agree to 2e-14 relative when `q_tot` is uniform, and differ by exactly the term above, to 3e-13 relative, when it is not. In that state the spurious term is 1.6 percent of the horizontal energy tendency and 0.6 percent of the vertical one. The unit test in this PR asserts both cases, in both directions.

The first figure is the reproducibility metric itself, the relative RMS of the final prognostic state after 4 simulated hours. The thermodynamic fields are preserved: density to 3e-6, total energy to 9e-5, total water to 1e-4. The velocities are fully decorrelated, at a relative RMS near √2, which is what two uncorrelated fields of the same magnitude give. That is the expected response of a convecting box to any perturbation of its state.

fig1_prog_state_rms


The second figure shows the horizontal-mean profiles, which tell the same story. Temperature, specific humidity, potential temperature, total enthalpy and relative humidity overlay to between 3e-6 and 3e-5 relative. The visible difference is in the domain RMS of vertical velocity, which is the turbulent field, and it differs in phase rather than in magnitude.

fig2_profiles


One caveat: this case forms no condensate in 4 hours. Cloud liquid, cloud ice, rain and snow are identically zero in both runs, so the reproducibility comparison exercises only the dry and vapour parts of the decomposition. The liquid and ice constituents are covered by the unit test rather than by this job.

So the counter bump is the reference tracking a small and exactly characterised change to the energy flux, and the simulated climate of the case is preserved.

</details>

akshaysridhar (approved):

PR #4705, opened 2026-07-17
SCM setup consolidation: ColumnDatasets interface and per-term forcing

### LINKED ISSUE #4617: SCM Setup Consolidation
Over time, additional single column setups have been added to ClimaAtmos, with potentially more in the future. Thus far we have a pattern of adding single column "cases" defined by forcings, boundary conditions, initial conditions, and ground truth (obs or LES). While some consolidation has been done, there is room for more. Currently, adding a case typically requires new `src/` code: a branch in the config dispatch (`get_external_forcing_model` / `get_setup_type`), a new forcing type, and its own `external_forcing_cache` and `external_forcing_tendency!` methods. 

A current limitation of consolidating the pipeline is differences in forcing file formats (LES field campaigns vs GCM-driven vs ERA5 vs VARANAL). This requires processing logic in either ClimaAtmos or ClimaArtifacts to process each appropriately. Every path also assumes the forcing lives in a file on disk, read with a source-specific naming and preprocessing convention. This is despite the fact that the underlying forcing is largely common across cases: column tendencies and profiles, surface terms, and `(z, t)` coordinates. 

One solution is to agree on a standard internal format (or follow a standardized file format like DEPHY) for the forcing data, and route the standardized files through a single generic/consolidated forcing type. Source-specific conversion logic would likely live upstream (e.g. in ClimaArtifacts) rather than in the model, so that adding a case becomes mostly a matter of data + config. Case-specific parts [upper boundary condition / domain top, surface treatment (prescribed SST vs. surface fluxes vs. land/ocean), and which terms are prescribed vs. nudged] would remain expressible through the config and a thin setup layer. Ideally, whether a term is present in the file would itself indicate whether it's prescribed, rather than introducing a growing set of flags.

Open questions: whether to adopt an existing community standard like DEPHY or define a CliMA-internal one; how a config can select or inject a forcing source without a new hardcoded branch.

nefrathenrici:
This PR introduces a generic framework for driving single-column simulations from netCDF forcing datasets.

Key changes:
- Added ColumnDatasets, a shared interface for reading column forcing files:
	- One native, self-describing format using (z, time) atmospheric variables and (time,) surface variables. It is currently named ClimaColumn, though we can bikeshed on the name.
	- New formats plug in as small dataset modules, with no changes to the forcing, setup, or config code.
- Changed generated ERA5 files to the native column schema. Stale or non-conforming cached files are regenerated from the source rather than read.
- Composed external forcing from per-process terms (HorizontalAdvection, VerticalFluctuation, Nudging, Subsidence) assembled into ExternalDrivenTVForcing, letting runscripts select terms and set per-variable nudging timescales and height masks.
- Added the ColumnFromFile setup, which can source initial conditions, external forcing, surface data, and insolation from one file.
- Extracted shared forcing-tendency kernels, de-duplicating the energy-conversion, nudging, and subsidence math across the GCM, ISDAC, and file-driven paths.
- Renamed the old large-scale-subsidence forcing to LargeScaleSubsidence, freeing the name Subsidence for the new per-process forcing term.
- Ported the ARM VARANAL case onto the interface: a converter (VaranalFiles.to_climacolumn) rewrites pressure-level profile into the native schema, and the case runs through ColumnFromFile. Removed the custom setup and forcing.

Future work (TBD):
- Migrate the cfsite (GCM-driven) case onto the new interface
- Add MOSAiC case?
- Generic in-memory reader, so a runscript can drive forcing from arrays or a non-native format without first writing a native file to disk via to_climacolumn
- Formalize derived variables (quantities computed from the file, e.g. density or vertical velocity) as a first-class part of the interface
- Give the file-driven terms an analytic-target data source, so idealized profiles flow through the same AbstractForcingTerm composition

Documentation to understand the new interface:
- docs/src/single_column_prospect.md section "Column forcing datasets"
- docs/src/setups.md section "Defining a Case in a Runscript"
- Docstrings: `ColumnDatasets`, `AbstractForcingTerm`, `ExternalDrivenTVForcing`, `ColumnFromFile`, `FileHeatFluxes`


src/config/era5_observations_to_forcing_file.jl
```diff
@@ -491,10 +423,8 @@ function generate_external_forcing_file(
         )
         return (μ, F)  # return (coszen, TOA_flux)
     end
```
Julians42: Nat I know this is my code but the `first.` and `last.` below (currently 511/2) could be a bit confusing since \mu and F are actually returned as entries 3 and 1 from `Insolation.insolation`. Could you maybe find a cleaner way to do this - e.g., I bet the list could just be generated here as \mus and Fs and then those arrays passed below.

src/config/model_getters.jl
```diff
@@ -628,16 +628,23 @@ function get_external_forcing_model(
...
-        ExternalDrivenTVForcing{FT}(external_forcing_file)
+        # The ColumnFromFile setup already carries the probed dataset handle;
+        # reuse its forcing rather than re-probing the file.
+        model =
+            isnothing(setup_type) ? nothing :
+            Setups.external_forcing(setup_type, FT)
+        isnothing(model) && error(
+            """external_forcing "ReanalysisTimeVarying" requires initial_condition "ReanalysisTimeVarying" (which supplies the same forcing automatically, so the key can simply be omitted).""",
+        )
+        model
```
Julians42: Nothing is written 3 times here and is took me a few seconds to parse. Suggest simplifying logic. I'd write this as an assert on the setup type and then just assign. Feel free to choose something else

src/config/model_getters.jl
```diff
@@ -653,15 +660,25 @@ function get_external_forcing_model(
                 ],
             )
         end
-        ExternalDrivenTVForcing{FT}(external_forcing_file)
-
+        # The monthly-averaged-diurnal file stores one day; repeat it in time.
+        ExternalDrivenTVForcing(
+            external_forcing_file;
+            time_interpolation_method = ColumnDatasets.periodic_calendar_method(),
```
Julians42, line 665: I'm assuming periodic calendar just wraps the data? It would be nice to throw a warning if the requested simulation time is longer than the file length. For monthly setup we want to wrap every day but for the daily one I'd be a bit more cautious as the wrap around would likely be quite discontinuous for this setup

src/config/model_getters.jl
```diff
@@ -653,15 +660,25 @@ function get_external_forcing_model(
...
-        ARMVARANALForcing{FT}(parsed_args["external_forcing_file"])
+    elseif external_forcing == "ColumnFromFile"
+        # Reuse the ColumnFromFile setup's already-probed forcing when it
+        # supplies one (initial_condition == "ColumnFromFile"), otherwise build
+        # from the file directly (forcing-only, no ColumnFromFile initial state).
+        model =
+            isnothing(setup_type) ? nothing :
+            Setups.external_forcing(setup_type, FT)
+        isnothing(model) ?
+        ExternalDrivenTVForcing(parsed_args["external_forcing_file"]) : model
```
Julians42: Same logic as before - just a bit hard to parse

src/prognostic_equations/forcing/external_forcing.jl
```diff
@@ -403,144 +413,282 @@ function external_forcing_tendency!(
...
+    ᶜdqtdt .= 0
+    unrolled_foreach(forcing_terms, term_caches) do term, cache
+        accumulate_Tq_tendency!(ᶜdTdt, ᶜdqtdt, term, cache, Y, p)
+    end
+    apply_Tq_forcing!(Yₜ, Y, p, ᶜdTdt, ᶜdqtdt)
+
+    unrolled_foreach(forcing_terms, term_caches) do term, cache
+        apply_direct_forcing!(Yₜ, Y, p, term, cache)
+    end
+    return nothing
```
Julians42, line 691: One thing I liked about the old code was that it was very easy to comment out any given tendency and see the effect on the simulation. For example, for a good number of cases the horizontal advective tendencies were resulting in negative moisture above the boundary layer because ERA5 and our model had different boundary layer heights. As a result, the simulations often were better without horizontal advection. Is there a simple way to do it here?

src/setups/ColumnFromFile.jl
```diff
@@ -0,0 +1,99 @@
...
+    return ColumnFromFile(
+        dataset,
+        start_date_dt,
+        external_forcing,
+        flux_scheme,
+        surface_temperature,
+        insolation,
+        profiles,
+    )
+end
```
Julians42: I really like this!

src/types.jl
```diff
@@ -550,39 +550,53 @@ end
...
+Runscripts can construct this model as
+`ExternalDrivenTVForcing(path; forcing = (...,))`. Surface-temperature and
+insolation requirements are derived from the resolved `AtmosModel` during
+cache construction rather than from the forcing terms.
+"""
+struct ExternalDrivenTVForcing{CD <: ColumnDatasets.ColumnDataset, F <: Tuple, M}
+    dataset::CD
+    forcing::F
+    time_interpolation_method::M
+end
```
Julians42, line 574: Just for my understanding `ColumnFromFile` is for constant forcing while this setup now contains arrays for each time? It's not possible to use the same one easily right even though the difference should just be arrays versus matrices essentially?

test/era5_tests.jl
```diff
@@ -76,19 +78,48 @@ end
...
 
-    # data is stored from top of atmosphere to surface
-    @test monotonic_decreasing(processed_data["zg"], 3)
-    @test monotonic_increasing(processed_data["ta"], 3)
-    @test monotonic_increasing(processed_data["hus"], 3)
+    # data is stored with a strictly ascending height coordinate, so
+    # temperature and humidity decrease along the z dimension
+    @test issorted(processed_data["z"][:]; lt = <=)
+    @test monotonic_decreasing(processed_data["ta"], 1)
+    @test monotonic_decreasing(processed_data["hus"], 1)
```
Julians42, line 122: Why does the monotonicity change here?

test/era5_tests.jl
```diff
@@ -179,7 +215,7 @@ end
     @test length(processed_data["time"][:]) == expected_time_steps
 
     # Test that data is consistent across time
-    @test monotonic_increasing(processed_data["ta"], 3)
+    @test monotonic_decreasing(processed_data["ta"], 1)
```
Julians42, line 218: Same here

test/era5_tests.jl
```diff
@@ -457,3 +494,71 @@ end
     close(processed_data_0K)
     close(processed_data_4K)
 end
+
+@testset "Multiday regeneration of non-conforming daily files" begin
```
Julians42, line 498: I'm assuming the addition of the test means that all of my 100s of old cached files will no longer work with this new setup? Could you explain what changed in the files and why the change is needed?

nefrathenrici: The previous files will no longer work, but should be regenerated automatically as before from the `era5_hourly_atmos_raw` artifact. The old files had a 3D box shape, the new files just have a 1D purely vertical column. This happened  because we used minimal boxes for the column spatial config before, since https://github.com/CliMA/ClimaAtmos.jl/pull/4568 we have been using actual column geometry.

src/prognostic_equations/forcing/external_forcing.jl
nefrathenrici: It should be simple to switch off, since you can remove it from the `forcing` argument:
```
data = CA.ColumnDatasets.ColumnDataset(forcing_file)
setup = CA.Setups.ColumnFromFile(
    data, "20070701";
    forcing = (
        CA.VerticalFluctuation(),
        CA.Nudging(:ta, :hus),
        CA.Nudging(:ua, :va),
        CA.Subsidence(),
    ),   # NO HorizontalAdvection()
)
# alternative call
forcing = CA.ExternalDrivenTVForcing(forcing_file; forcing = (CA.VerticalFluctuation(), ...))
```
One thing is that `HorizontalAdvection` currently bundles temperature and moisture, it isn't possible to keep one and disable the other.

src/types.jl
nefrathenrici: `ColumnFromFile` is not a constant forcing, it provides an IC and a timevarying forcing. I'm a bit confused by the array vs matrix distinction, but the IC is a 1D (z,) array and the forcing is a 2D (z, time) matrix. The constant forcing is the cases where the time dimension has a length of 1.

test/era5_tests.jl
nefrathenrici: The ordering in the new file format is flipped, so walking up the index goes up in height, so ta and hus decrease with index.

src/config/model_getters.jl
nefrathenrici: Yes, PeriodicCalendar wraps the data, so when the sim time goes past, it restarts from the first time. The dates in the file also need to be evenly spaced. I can add a small check at construction that compares the file's timespan to the sim's t_end.

test/era5_tests.jl
Julians42: Got it - figured was something like this - thanks for clarifying!

test/era5_tests.jl
Julians42: Yup makes sense - that was in the works for a while

src/types.jl
Julians42: Yes by matrix I meant that there was a 2D (z, time) matrix as opposed to just a static forcing that was 1D (time). Looking back at the struct for `ColumnFromFile` answered my question:
```julia
struct ColumnFromFile{
    CD <: ColumnDatasets.ColumnDataset,
    F <: ExternalDrivenTVForcing, # time varying struct used here
    FS,
    ST,
    I,
    P <: ColumnProfiles,
}
```
Before they were all separate structs but now I see the hierarchy :)

src/config/model_getters.jl
Julians42: Discussed offline - just add `TODO need a cleaner way to dispatch to each of the SCM models` or something

src/config/era5_observations_to_forcing_file.jl
Julians42: Looks clean now thanks

src/config/model_getters.jl
Julians42: Yup check would be perfect - thanks

Julians42 (approved):
This looks good to me from the TimeVarying forcing models perspective

src/column_datasets/ClimaColumnFiles.jl
```diff
@@ -0,0 +1,247 @@
...
+    "tntha" => "K s-1",
+    "tnhusha" => "kg kg-1 s-1",
+    "tntva" => "K s-1",
+    "tnhusva" => "kg kg-1 s-1",
+    "clw" => "kg kg-1",
+    "cli" => "kg kg-1",
+    "ts" => "K",
+    "hfls" => "W m-2",
+    "hfss" => "W m-2",
+    "coszen" => "1",
```
szy21, line 46: It makes more sense for this to be unitless, but if 1 is what we currently have in the file it's fine.

szy21:
We don't have to do it now, but I wonder if it is better for files like `src/column_datasets/VaranalFiles.jl` or `src/config/era5_observations_to_forcing_file.jl` to stay outside ClimaAtmos. That is, we always require users to follow a specific format when preparing forcing files.

src/setups/ForcingFromFile.jl
szy21, line 1: Maybe better to name it something like ForcingFromFile.jl (other files in setup, e.g. Bomex.jl, DYCOMS.jl, are all columns, and the name indicates what the forcing is like.)

config/default_configs/default_config.yml
```diff
@@ -228,7 +228,7 @@ surface_setup:
   help: "Surface flux scheme [`DefaultExchangeCoefficients` (default), `DefaultMoninObukhov`]"
   value: "DefaultExchangeCoefficients"
 initial_condition:
-  help: "Initial condition [`DryBaroclinicWave`, `MoistBaroclinicWave`, `ConstantBuoyancyFrequencyProfile`, `DecayingProfile` (default), `IsothermalProfile`, `Bomex`, `DryDensityCurrentProfile`, `RisingThermalBubbleProfile`, `ISDAC`, `GCM`, `ARMVARANAL`, `Reanalysis`, `ReanalysisTimeVarying`], or a file path for a NetCDF file (read documentation about requirements)."
+  help: "Selects the simulation case/setup. Global/idealized options: `DryBaroclinicWave`, `MoistBaroclinicWave`, `ConstantBuoyancyFrequencyProfile`, `DecayingProfile` (default), `IsothermalProfile`, `Bomex`, `DryDensityCurrentProfile`, `RisingThermalBubbleProfile`, `ISDAC`, or a NetCDF file path (see docs). Externally-driven single-column cases: `GCM`, `ARMVARANAL`, `ColumnFromFile`, `ReanalysisTimeVarying`)."
```
costachris:
```suggestion
  help: "Selects the simulation case/setup. Global/idealized options: `DryBaroclinicWave`, `MoistBaroclinicWave`, `ConstantBuoyancyFrequencyProfile`, `DecayingProfile` (default), `IsothermalProfile`, `Bomex`, `DryDensityCurrentProfile`, `RisingThermalBubbleProfile`, `ISDAC`, or a NetCDF file path (see docs). Externally-driven single-column cases: `GCM`, `ARMVARANAL`, `ColumnFromFile`, `ReanalysisTimeVarying`."
```

src/column_datasets/VaranalFiles.jl
```diff
@@ -0,0 +1,192 @@
...
+    return [base + Dates.Millisecond(round(Int, 1000 * Float64(t))) for t in time_raw]
+end
+
+# A (lev, time) field in the file's level order, fill values replaced with NaN.
+function read_lev_time(ds, name, nlev, ntime)
+    haskey(ds, name) ||
+        error("ARM VARANAL file is missing required column variable `$name`")
+    raw = Array(ds[name])
+    ndims(raw) == 2 || error("`$name` must be a 2D (lev, time) field")
+    data = fill_to_nan.(raw)
```
costachris, line 148: Perhaps throw a warning if there are fill values in the forcing data/initial conditions so we know where they come from if the simulation crashes due to the nan fill.

src/types.jl
```diff
@@ -533,7 +533,7 @@ end
 
 abstract type AbstractForcing end
 struct HeldSuarezForcing end
-struct Subsidence{T} <: AbstractForcing
+struct LargeScaleSubsidence{T} <: AbstractForcing
```
costachris, line 536: The analytic SCM subsidence is now LargeScaleSubsidence and Subsidence() is the file-driven AbstractForcingTerm that reads wa. Can you mention this in the docs or News?

src/config/type_getters.jl
```diff
@@ -142,17 +142,58 @@ function get_setup_type(parsed_args, thermo_params)
...
+                fluxes = SurfaceConditions.FileHeatFluxes(data, start_date),
+            )
+        else
+            SurfaceConditions.MoninObukhov(; z0 = FT(0.05), ustar = FT(0.28))
+        end
+        return Setups.ColumnFromFile(
+            data,
+            start_date;
+            forcing = (
+                HorizontalAdvection(),
```
costachris, line 178: Having the types in the code to specify terms is nice. We may not want to typically expose these to users at the config-level, but there are use cases where you may want to turn off terms. Could these also be exposed in the configs? scm_forcing_terms: [HorizontalAdvection, Nudging, Subsidence] . Maybe not needed now

costachris (approved):

costachris:
Can we add a small table to the docs of the default large-scale forcing and surface settings per SCM setup?

eg:

| Setup | Large-scale forcing (default) | Surface / insolation (default) |
| --- | --- | --- |
| ColumnFromFile / ERA5 TV | `default_forcing_terms()`: HAdv + VertFluc + Nudge(T,q) + Nudge(u,v) + Subsidence | MO (`z0=1e-4`); `ExternalTemperature`; `ExternalTVInsolation` |
| ARMVARANAL | HAdv + Nudge(T,q) + Nudge(u,v) + Subsidence | MO + `FileHeatFluxes` when LH/SH present; `ExternalTemperature`; `TimeVaryingInsolation` |

even just the file-driven cases from this PR would help.

src/setups/ForcingFromFile.jl
nefrathenrici: I think it's slightly ambiguous since `ForcingFromFile` provides more than just the forcing but it is still better than `ColumnFromFile`. I will rename.

src/config/type_getters.jl
nefrathenrici: I think it would be a quick addition on top of this interface. I can add it when we migrate the GCMDriven case next.

PR #4712, opened 2026-07-17
fix(radiation): make the TRMM_LBA radiation GPU-compatible

haakon-e:
Follow-up to #4664: makes the `TRMM_LBA` radiation GPU-compatible, so a full `TRMM_LBA` simulation runs on a GPU space.

`APL.TRMM_LBA_radiation` is a 2D (time × height) `Interpolations.jl` interpolant that is not isbits. It was held by `RadiationTRMM_LBA`, which lives in the `AtmosModel`, and the `AtmosModel` is captured whole into device kernels (for example `update_surface_conditions!`), so a `TRMM_LBA` run crashed at initialization on a GPU space with a non-isbits kernel argument.

`RadiationTRMM_LBA` becomes an isbits marker and the profile moves into the radiation cache. The radiative-cooling tendency resolves the time coordinate on the host into the height profile at `t`, then interpolates in height on the device with a `ClimaInterpolations.Interpolation1D.Interpolate1D` over `SVector` knots. Only the small height-profile `SVector`, not the full 36 × 33 table, enters the kernel, so the kernel-parameter footprint stays small and portable across GPUs. 

On CPUs, we profile is broadcast directly, so there's no changes to reproducibility tests.

Similar cases not addressed here: `ARMVARANAL` has the same `TimeZProfile`, plus a separate host scalar loop over `parent(z)` in `update_varanal_forcing_fields!`.

### DIFF
--- a/src/parameterized_tendencies/radiation/radiation.jl
+++ b/src/parameterized_tendencies/radiation/radiation.jl
@@ -18,6 +18,9 @@ import ClimaUtilities.TimeVaryingInputs:
     LinearInterpolation
 
 import Interpolations as Intp
+import StaticArrays as SA
+import ClimaInterpolations.Interpolation1D as CI1D
+import AtmosphericProfilesLibrary as APL
 using Statistics: mean
 
 radiation_model_cache(Y, atmos::AtmosModel, args...) =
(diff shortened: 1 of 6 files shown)


haakon-e:

- #4664
    - #4712 ◀

src/parameterized_tendencies/radiation/radiation.jl
```diff
@@ -521,26 +524,58 @@ end
 ##### TRMM_LBA radiation
 #####
 
+# Height grid of the AtmosphericProfilesLibrary TRMM_LBA radiative-cooling
+# table, read from the underlying interpolant.
+function trmm_lba_radiation_z_knots(rad_profile)
+    z_grid = getfield(getfield(getfield(rad_profile, :prof), :itp), :knots)[2]
```
nefrathenrici, line 530: This line depends on the structure of the APL profile. I think it should be fine given that APL is not likely to have any internal changes, but this is fragile.

src/parameterized_tendencies/radiation/radiation.jl
```diff
@@ -521,26 +524,58 @@ end
...
+    return SA.SVector{length(z_grid)}(z_grid)
```
nefrathenrici, line 531: This may need an FT type annotation.

nefrathenrici (approved):
Thank you! Do we need to run TRMM on GPU?

haakon-e:
> Thank you! Do we need to run TRMM on GPU?

In a column, TRMM is about 2x faster on 1 CPU compared to 1 GPU (though the difference is as little as 10% if you save hdf5/nc files). 
in a 3D box however, GPU runs quickly become much faster. I’ve greatly enjoyed running TRMM in both columns and boxes to debug microphysics, and it’s useful to run it end-to-end on the GPU.

Thanks for the comments, I’ll look at them carefully and see what’s a good set of fixes before merging.

PR #4713, opened 2026-07-19
Add 5-bin sea salt aerosol tracer

ziglaser:
## PR Dependency
 - Ocean fraction wiring requires [ClimaCoupler.jl #2049](https://github.com/CliMA/ClimaCoupler.jl/pull/2049), *AMIP runs or outdated Coupler runs currently default to all 100% ocean coverage*

## New feature
 - Sea salt aerosol emitted via Gong 2003 wind-speed dependent scheme (with toggled sea surface temperature dependence as per Jaegle 2011) in proportion to `ocean_fraction` of cells
 - Emission fluxes are precomputed per bin in `set_sea_salt_emission_flux!` (`aerosols/sea_salt.jl`) during `set_explicit_precomputed_quantities!`, using a 10 m wind extrapolated from the lowest model level via MOST, and applied as a bottom boundary flux by `sea_salt_emission_tendency!` in `surface_flux_tendency!`
 - Aerosols are treated as $\rho \chi$ standard passive trated into horizontal advection via`horizontal_tracer_advection_tendency!` and vertical transport via `vertical_transport` in `explicit_vertical_advection_tendency!` (both in `prognostic_equations/advection.jl`)
 - Sea salt aerosol is removed through a tendency enacting a 0.55 day half life for all bins and vertical levels (`sea_salt_deposition_tendency!`,
called from `additional_tendency!`), a placeholder until weional
 - Created diagnostics in `tracer_diagnostics.jl` for individual aerosol size bins (`mmrsslt01`–`mmrsslt05`, and analogous `mmrdst0X` for dust); new diagnostics for SSA emissions: total flux `emiss` and per-bin `emisslt01`–`emisslt05`

## Config changes
 - New optional `interactive_aerosols` config option, takes list of (currently sea salt aerosol) i.e. ["SSLT01", ...], defaults empty. Stored on `AtmosModel` as a `Val` of names so tracer sets are inferrable at compile time.

## Wiring changes
 - Ocean fraction is passed from `ClimaCoupler.jl` via `Intcluded in `AtmosSimulation` struct
 - Prescribed and interactive aerosols stored in `tracer_cache` in `tracer_cache.jl`
 - Interactive aerosol bins are added to the PrognosticEDMFX updraft state as specific tracers (`interactive_aerosol_sgs_variables` in `prognostic_variables.jl`), zero-initialized following wiring of [ClimaAtmos #4629](https://github.com/CliMA/ClimaAtmos.jl/pull/4629)
 - `radiation/update_inputs.jl`: `update_aerosol_concentrations!` split into `update_prescribed_aerosol_concentrations!` (evaluates MERRA-2 time-varying inputs) and `update_radiation_aerosol_concentrations!` (packages per-layer column mass densities for RRTMGP). For each species, the interactive (prognostic) mass is used when active, falling back to the prescribed climatology otherwise, so interactive sea salt feeds aerosol radiation directly. The RRTMGP↔MERRA-2 name mapping is hoisted into a shared `RRTMGP_AEROSOL_NAMES_PAIR` constant.
 - `radiation/radiation.jl`: sea-salt bin radii passed to RRTMGP now come from ClimaParams (`MERRA2_seasalt_aerosol_bin0X_radius`). Aerosol check in `radiation_model_cache` is derived from `RRTMGP_AEROSOL_NAMES_PAIR` rather than a duplicated name list, and now accepts interactive as well as prescribed aerosol names.

## Renames
 - Uses of "aerosol_[VAR]" renamed to the following scheme: "prescribed_aerosol_[VAR]" when only applicable to MERRA-2 prescribed aerosols, "interactive_aerosol_[VAR]" only applicable to interactive aerosols, "aerosol_[VAR]" when applicable to both. This includes the `AtmosSimulation` kwarg `aerosol_names` -> `prescribed_aerosol_names`.

## Future Work
 - More thoughtful treatment of 10m wind estimate with MOST — current formulation assumes bottom level center wind value is point value, not layer averaged.
 - Replace Gong 2003 parameterization with more physically supported parameterization (likely Deike 2022)
 - Replace half-life decay with wet and dry deposition tendencies
 - Implementation of hygroscopic growth: wet/dry deposition and optics are dependent on RH-influenced particle size
 - Interactions with cloud microphysics

----


src/parameterized_tendencies/radiation/radiation.jl
```diff
@@ -229,21 +228,26 @@ function rrtmgp_solver_kwargs(
...
-            center_ss1_radius = 0.55,
-            center_ss2_radius = 1.4,
-            center_ss3_radius = 2.4,
-            center_ss4_radius = 4.5,
-            center_ss5_radius = 8,
+            center_ss1_radius = aero_params.SSLT01_radius * FT(1e6),
+            center_ss2_radius = aero_params.SSLT02_radius * FT(1e6),
+            center_ss3_radius = aero_params.SSLT03_radius * FT(1e6),
+            center_ss4_radius = aero_params.SSLT04_radius * FT(1e6),
+            center_ss5_radius = aero_params.SSLT05_radius * FT(1e6),
```
szy21: Let's not change the radius here. It depends on how RRTMGP creates the lookup table for aerosol optics, and I need to double check it. But I think what we have before is correct.

src/cache/cache.jl
```diff
@@ -168,6 +172,14 @@ function build_cache(
     scratch = temporary_quantities(Y, atmos)
 
     precomputed = precomputed_quantities(Y, atmos)
+    tracers = tracer_cache(
+        Y,
+        prescribed_aerosol_names,
```
trontrytel: Do we want to support prescribed and prognostic aerosol at the same time? To allow having prognostic sea salt but prescribed everything else? This is appealing for testing but may result in complicated code. The alternative would be to allow for either fully prescribed or fully interactive.

src/cache/cache.jl
```diff
@@ -168,6 +172,14 @@ function build_cache(
...
+        time_varying_trace_gas_names,
+        start_date,
+        _interactive_aerosol_names(atmos.interactive_aerosols),
```
trontrytel: I would add it right after prescribed aerosol names, drop the underscore and name in prognostic aerosol names. "Interactive" is a new way to say prognostic and will be confusing if we start to use it interchangeably in the model.

config/default_configs/default_config.yml
```diff
@@ -417,6 +408,9 @@ reproducible_restart:
 prescribed_aerosols:
   help: "Which aerosols to add. List of keys from the data file (e.g., CB1, CB2)."
   value: []
+interactive_aerosols:
```
trontrytel: prognostic_aerosols

src/cache/cache.jl
```diff
@@ -14,6 +14,7 @@ struct AtmosCache{
     ORGW,
     RAD,
     TRAC,
+    OFRAC,
```
trontrytel, line 17: Not for this PR, but we should re-think this pattern of adding new fields to this struct whenever we add new functionality to the model

src/cache/cache.jl
```diff
@@ -168,6 +172,14 @@ function build_cache(
...
 
     precomputed = precomputed_quantities(Y, atmos)
+    tracers = tracer_cache(
+        Y,
+        prescribed_aerosol_names,
+        time_varying_trace_gas_names,
+        start_date,
+        _interactive_aerosol_names(atmos.interactive_aerosols),
+    )
+    ocean_fraction = ones(axes(Fields.level(Y.f, Fields.half))) # Populated via Interfacer.update_field! before first timestep
```
trontrytel: What to do in purely Atmos runs?

trontrytel: Also, do we really have to cache it?

src/cache/cache.jl
```diff
@@ -187,20 +201,23 @@ function build_cache(
 
     set_precomputed_quantities!(Y, precomputing_arguments, FT(0))
 
+    radiation_aerosol_names = (
```
trontrytel: Is this used anywhere else? Maybe we can add both categories directly to radiation_args without specifying another entity

src/cache/cache.jl
```diff
@@ -187,20 +201,23 @@ function build_cache(
...
-            aerosol_names,
+            radiation_aerosol_names,
             time_varying_trace_gas_names,
             atmos.insolation,
         ) : ()
     hyperdiff = hyperdiffusion_cache(Y, atmos)
     non_orographic_gravity_wave = non_orographic_gravity_wave_cache(Y, atmos)
     orographic_gravity_wave = orographic_gravity_wave_cache(Y, atmos)
     radiation = radiation_model_cache(Y, atmos, radiation_args...)
-    tracers = tracer_cache(Y, aerosol_names, time_varying_trace_gas_names, start_date)
```
trontrytel, line 203: We don't need the trace_gas_names here anyomre?

src/cache/precomputed_quantities.jl
```diff
@@ -707,6 +707,8 @@ NVTX.@annotate function set_explicit_precomputed_quantities!(Y, p, t)
         SurfaceConditions.update_surface_conditions!(Y, p, FT(t))
     end
 
+    set_sea_salt_emission_flux!(Y, p)
```
trontrytel: Do we want to always apply it, or only under some user defined option?

src/parameterized_tendencies/radiation/radiation.jl
```diff
@@ -229,21 +228,26 @@ function rrtmgp_solver_kwargs(
...
+        # TODO: dust bin radii have no ClimaParams entries yet — move these
+        # literals there alongside the sea-salt ones.
+        aero_params = CAP.prescribed_aerosol_params(params)
         kwargs = (;
             kwargs...,
             aod_sw_extinction = NaN,
             aod_sw_scattering = NaN,
             # assuming fixed aerosol radius
-            center_dust1_radius = 0.55,
+            center_dust1_radius = 0.73,
```
trontrytel: Why this needs to be changed?

src/parameterized_tendencies/radiation/radiation.jl
trontrytel: We should not change those numbers by accident. They are consistent with RRTMGP assumptions, I think

src/setups/common/prognostic_variables.jl
```diff
@@ -137,6 +151,7 @@ function turbconv_center_variables(
     turbconv_model::PrognosticEDMFX,
     microphysics_model,
     chemistry_model,
+    interactive_aerosols,
```
trontrytel: aerosol_model ?

test/diagnostics/unit_diagnostics.jl
```diff
@@ -477,6 +446,9 @@ SKIP_CASES = Set([
     # tracer_diagnostics.jl
     "loadss", "mmrbcpi", "mmrbcpo", "mmrdust", "mmrocpi", "mmrocpo",
     "mmrso4", "mmrss", "o3",
+    "mmrdst01", "mmrdst02", "mmrdst03", "mmrdst04", "mmrdst05",
```
trontrytel: Do we document anywhere what thise acronyms stand for?

src/prognostic_equations/remaining_tendency.jl
```diff
@@ -340,6 +340,8 @@ NVTX.@annotate function additional_tendency!(Yₜ, Y, p, t)
     # at the cost of water vapor.
     tracer_nonnegativity_vapor_tendency!(Yₜ, Y, p, t, microphysics_model)
 
+    sea_salt_deposition_tendency!(Yₜ, Y, p, t)
```
trontrytel: Should we dispatch on it based on the aerosol_model type?

src/prognostic_equations/surface_flux.jl
```diff
@@ -144,8 +144,13 @@ function surface_flux_tendency!(Yₜ, Y, p, t)
...
-            @. ᶜχʲₜ -= specific(btt, p.precomputed.ᶜρʲs.:(1))
+            ᶜχʲ_name = get_χʲ_name_from_ρχ_name(ρχ_name)
+            if MatrixFields.has_field(Yₜ.c, ᶜχʲ_name)
+                ᶜχʲₜ = MatrixFields.get_field(Yₜ.c, ᶜχʲ_name)
+                @. ᶜχʲₜ -= specific(btt, p.precomputed.ᶜρʲs.:(1))
+            end
         end
     end
+
+    sea_salt_emission_tendency!(Yₜ, Y, p, t)
```
trontrytel: Consider dispatching based on aerosol model type?

src/prognostic_equations/surface_flux.jl
```diff
@@ -144,8 +144,13 @@ function surface_flux_tendency!(Yₜ, Y, p, t)
 
         if turbconv_model isa PrognosticEDMFX
             # assuming one updraft
-            ᶜχʲₜ = MatrixFields.get_field(Yₜ.c, get_χʲ_name_from_ρχ_name(ρχ_name))
-            @. ᶜχʲₜ -= specific(btt, p.precomputed.ᶜρʲs.:(1))
+            ᶜχʲ_name = get_χʲ_name_from_ρχ_name(ρχ_name)
```
trontrytel: Could you explain why this needs the additional if now?

src/config/model_getters.jl
```diff
@@ -718,9 +638,14 @@ function get_detrainment_model(parsed_args)
 end
 
 function get_tracers(parsed_args)
-    aerosol_names = Tuple(parsed_args["prescribed_aerosols"])
+    prescribed_aerosol_names = Tuple(parsed_args["prescribed_aerosols"])
+    interactive_aerosol_names = Tuple(parsed_args["interactive_aerosols"])
```
trontrytel: prognostic_aerosol_names ?

src/parameterized_tendencies/radiation/update_inputs.jl
```diff
@@ -155,67 +157,79 @@ function update_volume_mixing_ratios!((; u, p, t)::I) where {I}
...
-            ("sea_salt4", :SSLT04),
-            ("sea_salt5", :SSLT05),
-        )
+Packages aerosol mass into the per-layer column mass densities (kg m⁻²)
+that RRTMGP expects, when `aerosol_radiation` is enabled. For each supported
+species, the interactive (prognostic, `Y.c.ρ<name>`) mass is used if the
+species is interactive; otherwise the prescribed (climatology) concentration
+is used.
+"""
+function update_radiation_aerosol_concentrations!((; u, p, t)::I) where {I}
```
trontrytel: I think the naming is a bit confusing here. We have the `update_prescribed_aerosol_concentrations!` followed by `update_radiation_aerosol_concentrations!` To me this reads as prescribed aerosol do not affect radiation. There is a separate group of aerosol that do affect radiation that gets updated later.

src/parameterized_tendencies/aerosols/sea_salt.jl
```diff
@@ -0,0 +1,186 @@
+import SurfaceFluxes as SF
+import SurfaceFluxes.Parameters as SFP
+import SurfaceFluxes.UniversalFunctions as UF
+
+# BIN DEFNITIONS
+const SEA_SALT_BIN_BOUNDS = (
+    (0.03, 0.1),   # μm, rep size 0.079 μm, SSLT01,
```
trontrytel: Are we ok with hard coding those, or could this be a parameter?

src/parameterized_tendencies/aerosols/sea_salt.jl
```diff
@@ -0,0 +1,186 @@
...
+"""
+function monin_obukhov_wind_extrapolated(z_target, z_anchor, u_anchor, L, uf_params, κ, z₀)
+    FT = typeof(u_anchor)
+    ζ_target = ifelse(iszero(L), FT(0), z_target / L)
+    ζ_anchor = ifelse(iszero(L), FT(0), z_anchor / L)
+    F_target =
+        UF.dimensionless_profile(uf_params, z_target, ζ_target, z₀, UF.MomentumTransport())
+    F_anchor =
+        UF.dimensionless_profile(uf_params, z_anchor, ζ_anchor, z₀, UF.MomentumTransport())
+    return max(u_anchor * F_target / F_anchor, FT(0))
```
trontrytel: Is it possible for `F_anchor` to be zero?

src/parameterized_tendencies/aerosols/sea_salt.jl
```diff
@@ -0,0 +1,186 @@
...
+end
+
+
+"""
+    _gong2003_r_integrand(r, theta)
+
+Precomputable radius-dependent part of the Gong (2003) integrand
+"""
+function _gong2003_r_integrand(r, theta)
+    A = 4.7 * (1 + theta * r)^(-0.017 * r^(-1.44))
```
trontrytel: Those should not be hardcoded. We should put them in ClimaParams

src/parameterized_tendencies/aerosols/sea_salt.jl
```diff
@@ -0,0 +1,186 @@
...
+    B = (0.433 - log10(r)) / 0.433
+    return 1.373 * r^(-A) * (1 + 0.057 * r^3.45) * 10^(1.607 * exp(-B^2))
+end
+
+function _gong_bin_integral(r_lo, r_hi, theta = 30.0, N = 512)
```
trontrytel: What is theta and N?

src/parameterized_tendencies/aerosols/sea_salt.jl
```diff
@@ -0,0 +1,186 @@
...
+"""
+    sea_salt_emission_flux(u_10, T_sfc, bin_index)
+
+Compute the upward sea salt number flux (particles m⁻² s⁻¹) per bin
+using [Gong2003](@cite). SST correction from [Jaegle2011](@cite).
+"""
+function sea_salt_emission_flux(u_10, T_sfc, bin_index; SST_adj = false)
+    FT = typeof(u_10)
+    bin_integral = FT(SEA_SALT_BIN_R_INTEGRALS[bin_index])
+    number_flux = bin_integral * abs(u_10)^FT(3.41)
```
trontrytel: Those parameters should be part of Clima Params

src/parameterized_tendencies/aerosols/sea_salt.jl
```diff
@@ -0,0 +1,186 @@
...
+    return s * dr
+end
+
+const SEA_SALT_BIN_R_INTEGRALS = ntuple(
+    i -> _gong_bin_integral(SEA_SALT_BIN_BOUNDS[i]...),
+    Val(5),
+)
+
+"""
+    sea_salt_emission_flux(u_10, T_sfc, bin_index)
```
trontrytel: Could you add a short entry to ClimaAtmos documentation?

src/parameterized_tendencies/aerosols/sea_salt.jl
```diff
@@ -0,0 +1,186 @@
...
+    s = (_gong2003_r_integrand(r_lo, theta) + _gong2003_r_integrand(r_hi, theta)) / 2
+    for i in 1:(N - 1)
+        s += _gong2003_r_integrand(r_lo + i * dr, theta)
+    end
+    return s * dr
+end
+
+const SEA_SALT_BIN_R_INTEGRALS = ntuple(
+    i -> _gong_bin_integral(SEA_SALT_BIN_BOUNDS[i]...),
+    Val(5),
```
trontrytel: Do we ever want to support more than 5 bins?

src/parameterized_tendencies/aerosols/sea_salt.jl
```diff
@@ -0,0 +1,186 @@
...
+"""
+    set_sea_salt_emission_flux!(Y, p)
+
+Compute per-bin and total sea salt surface emission fluxes (kg m⁻² s⁻¹) and
+store them in `p.tracers`. u_10 dependence extrapolated from u* and
+lowest model level wind using MOST and `surface_fluxes_params`.
+
+Called during `set_explicit_precomputed_quantities!'. `sea_salt_emission_tendency!` reads the cached values rather than recomputing.
+"""
+function set_sea_salt_emission_flux!(Y, p)
```
trontrytel: You could dispatch based on the aerosol_model and avoid the isempty. I think its a cleaner pattern. Thats what we do in most other tendencies

src/parameterized_tendencies/aerosols/sea_salt.jl
```diff
@@ -0,0 +1,186 @@
...
+    interactive_aerosol_names = _interactive_aerosol_names(p.atmos.interactive_aerosols)
+    isempty(interactive_aerosol_names) && return
+
+    FT = eltype(Y)
+    (; sfc_conditions) = p.precomputed
+    surface_fluxes_params = CAP.surface_fluxes_params(p.params)
+    uf_params = SFP.uf_params(surface_fluxes_params)
+    κ = SFP.von_karman_const(surface_fluxes_params)
+    ustar_p = parent(sfc_conditions.ustar)
```
trontrytel: Why do we need the `parent` here?

src/parameterized_tendencies/aerosols/sea_salt.jl
```diff
@@ -0,0 +1,186 @@
...
+    L_p = parent(sfc_conditions.obukhov_length)
+    roughness_spec = SF.COARE3RoughnessParams{FT}()
+
+    z0_eff = p.scratch.temp_field_level
+    z₀_p = parent(z0_eff)
```
trontrytel: I don't think the `parent` is needed here?

src/parameterized_tendencies/aerosols/sea_salt.jl
```diff
@@ -0,0 +1,186 @@
...
+    z₀_p .= SF.momentum_roughness.(roughness_spec, ustar_p, surface_fluxes_params, nothing)
```
trontrytel: Maybe
```suggestion
    @. z₀_p = SF.momentum_roughness(roughness_spec, ustar_p, surface_fluxes_params, nothing)
```

src/parameterized_tendencies/aerosols/sea_salt.jl
```diff
@@ -0,0 +1,186 @@
...
+
+    z_c1_p = parent(Fields.level(Fields.coordinate_field(axes(Y.c)).z, 1))
```
trontrytel: Same as above

src/parameterized_tendencies/aerosols/sea_salt.jl
```diff
@@ -0,0 +1,186 @@
...
+    bin_radii = (
+        aero_params.SSLT01_radius,
+        aero_params.SSLT02_radius,
+        aero_params.SSLT03_radius,
+        aero_params.SSLT04_radius,
+        aero_params.SSLT05_radius,
+    )
+    mass_per_particle = ntuple(
+        i -> FT(4 / 3 * π * bin_radii[i]^3 * aero_params.seasalt_density),
+        Val(5),
```
trontrytel: I think its safer to make the number of bins a parameter. If we ever change it it'll be hard to trace all the places it is being hardcoded

src/parameterized_tendencies/aerosols/sea_salt.jl
```diff
@@ -0,0 +1,186 @@
...
+    aero_params = p.params.prescribed_aerosol_params
+    bin_radii = (
+        aero_params.SSLT01_radius,
+        aero_params.SSLT02_radius,
+        aero_params.SSLT03_radius,
+        aero_params.SSLT04_radius,
+        aero_params.SSLT05_radius,
+    )
+    mass_per_particle = ntuple(
+        i -> FT(4 / 3 * π * bin_radii[i]^3 * aero_params.seasalt_density),
```
trontrytel: I suspect we don't need `FT` here

src/parameterized_tendencies/aerosols/sea_salt.jl
```diff
@@ -0,0 +1,186 @@
...
+
+
+"""
+    monin_obukhov_wind_extrapolated(z_target, z_anchor, u_anchor, L, uf_params, κ, z₀)
+
+Extrapolate wind speed at `z_target` (m) by anchoring to a known model-level wind
+`u_anchor` at height `z_anchor` (m) via the MOST profile ratio.
+TODO: This MOST computation assumes center-face layer 1 wind is is point value, not layer average
+"""
+function monin_obukhov_wind_extrapolated(z_target, z_anchor, u_anchor, L, uf_params, κ, z₀)
```
tapios: Why do you need this? Can't you just use 'compute_profile_value` to get the wind wherever you need it?

src/parameterized_tendencies/aerosols/sea_salt.jl
tapios: Yes, and it should be written in a way that is invariant under changes in units. That is, `r -> r/r_gong_ref`, with a reference scale `r_gong_ref`, with the exponents correspondingly adjusted. We do not want to take fractional powers of dimensional quantities.

src/parameterized_tendencies/aerosols/sea_salt.jl
```diff
@@ -0,0 +1,186 @@
...
+
+
+"""
+    _gong2003_r_integrand(r, theta)
+
+Precomputable radius-dependent part of the Gong (2003) integrand
+"""
+function _gong2003_r_integrand(r, theta)
+    A = 4.7 * (1 + theta * r)^(-0.017 * r^(-1.44))
+    B = (0.433 - log10(r)) / 0.433
```
tapios: Same here: use `r/r_gong_ref` and adjust constants to match result. Make type-stable.

src/parameterized_tendencies/aerosols/sea_salt.jl
```diff
@@ -0,0 +1,186 @@
...
+    return 1.373 * r^(-A) * (1 + 0.057 * r^3.45) * 10^(1.607 * exp(-B^2))
```
tapios: Same here.

src/parameterized_tendencies/aerosols/sea_salt.jl
tapios: Also, here and throughout, the expressions promote to f64. Make type stable.

src/parameterized_tendencies/aerosols/sea_salt.jl
tapios: Make parameter

src/parameterized_tendencies/aerosols/sea_salt.jl
tapios: And, as before, never take a fractional power of a dimensional quantity. Introduce a reference scale and adjust constants to match results.

src/parameterized_tendencies/aerosols/sea_salt.jl
```diff
@@ -0,0 +1,186 @@
...
+Compute the upward sea salt number flux (particles m⁻² s⁻¹) per bin
+using [Gong2003](@cite). SST correction from [Jaegle2011](@cite).
+"""
+function sea_salt_emission_flux(u_10, T_sfc, bin_index; SST_adj = false)
+    FT = typeof(u_10)
+    bin_integral = FT(SEA_SALT_BIN_R_INTEGRALS[bin_index])
+    number_flux = bin_integral * abs(u_10)^FT(3.41)
+    if SST_adj
+        SST_factor =
+            FT(0.3) + FT(0.1) * T_sfc - FT(0.0076) * T_sfc^2 + FT(0.00021) * T_sfc^3
```
tapios: Same here: Introduce reference scales and make sure end results match by adjusting coefficients. Also double check: here temperatures are in K. Are they in the original reference?

src/parameterized_tendencies/aerosols/sea_salt.jl
```diff
@@ -0,0 +1,186 @@
...
+    z0_eff = p.scratch.temp_field_level
+    z₀_p = parent(z0_eff)
+    z₀_p .= SF.momentum_roughness.(roughness_spec, ustar_p, surface_fluxes_params, nothing)
+
+    z_c1_p = parent(Fields.level(Fields.coordinate_field(axes(Y.c)).z, 1))
+    u_z1_p = parent(norm.(Fields.level(Y.c.uₕ, 1)))
+
+    u_10 = p.scratch.ᶠtemp_field_level
+    parent(u_10) .=
+        monin_obukhov_wind_extrapolated.(FT(10), z_c1_p, u_z1_p, L_p, uf_params, κ, z₀_p)
```
tapios: You can just use 'compute_profile_value` from SurfaceFluxes.

src/parameterized_tendencies/aerosols/sea_salt.jl
tapios: Agree

src/parameterized_tendencies/aerosols/sea_salt.jl
```diff
@@ -0,0 +1,186 @@
...
+Currently implemented as a simple exponential decay. Separate wet and dry deposition
+tendencies are forthcoming
+"""
+function sea_salt_deposition_tendency!(Yₜ, Y, p, t)
+    interactive_aerosol_names = _interactive_aerosol_names(p.atmos.interactive_aerosols)
+    isempty(interactive_aerosol_names) && return
+
+    FT = eltype(Y)
+
+    half_life = FT(0.55 * 86400) # AeroCOM SS residence is ~0.55 days
```
tapios: Use parameter for length of day

tapios (changes requested):
Avoid magic numbers; make sure we only have nondimensional parameters (i.e., nondimensionalize the dimensional quantities appearing by introducing reference scales and and making them parameters).

src/parameterized_tendencies/radiation/radiation.jl
ziglaser: Changed dust radius based on this MERRA document (https://gmao.gsfc.nasa.gov/media/gmao_products/IqPeIBhIWLGrweILfVPj9Vmdoc/Dust_z72405udf3g1la8jwxmyspohi.pdf), given the rest of the values matched, but don't know how the RRTMGP lookup tables work, so can revert. Should these also be placed in ClimaParams?

src/cache/cache.jl
ziglaser: Can refactor, originally had as "prognostic" but chickened out after reading the Wikipedia page for "prognostic variable" since we're not integrating anything to get these

src/cache/cache.jl
ziglaser: Currently in purely Atmos runs I just treat everything as ocean. I presume somewhere there's an Artifact or something with an ocean mask strictly in Atmos but don't know where, will discuss.

Don't know if we need to cache since it only gets consumed by `set_sea_salt_emission_flux!`. That currently takes state and cache, is there another good way to pass it?

src/cache/cache.jl
ziglaser: Was not deleted, moved upwards so it can be placed in precomputing_arguments which I think was necessary for a previous EDMF wiring. I assume it can be moved back down with the other caches since passive tracer EDMF now runs through automatic inclusion, will experiment.

src/parameterized_tendencies/aerosols/sea_salt.jl
ziglaser: Can convert to ClimaParam! Just wasn't sure what belongs in parameters.toml, and what data types to use (i.e. for a bin with a low_bound, high_bound, and representative size, should those be three separate entries or a tuple)

src/parameterized_tendencies/aerosols/sea_salt.jl
ziglaser: Theta a the shape parameter set in Gong, which controls the shape of the  particle size distribution in the <0.2
\mu m size scale. The paper doesn't make an effort to fit it, given a lack of measurements at that size, and all implementations I've seen default to theta=30. 

N is the number of quadrature points used in computing the integral.

src/parameterized_tendencies/aerosols/sea_salt.jl
ziglaser: As in having different AerosolModel structs, one for prescribed and one for prognostic, and dispatch based on the input struct?

src/cache/cache.jl
ziglaser: It's not used outside creating a radiation cache, we could pass both--I think the right approach will depend on how we handle the prescribed vs prognostic aerosol model

src/prognostic_equations/surface_flux.jl
ziglaser: Yeah, I don't think it does need the if given we're sourcing $\rho \chi$ names from the state already. I think that wound up there during debugging with Claude but don't recall why, will remove and test.

src/parameterized_tendencies/radiation/update_inputs.jl
ziglaser: When it was just prescribed aerosols, this script both updated the prescribed aerosol fields forward in time, and then fed them to RRTMGP. To add prognostic, I split it into two functions, so we have separate functions that step froward in time, and then one unified function to compute the aerosol column density from both prescribed and prognostic sources and feed it to RRTMGP. I agree I should change the naming

src/parameterized_tendencies/radiation/radiation.jl
szy21: I think we can keep these values hard-coded and move them to parameters after finalizing aerosol optics. Where is e.g. SSLT01_radius defined now?

src/cache/cache.jl
szy21: I would prefer only supporting prescribed or prognostic aerosols, if supporting both makes the code more complicated.

PR #4715, opened 2026-07-20
Add precipitation entrainment for updrafts widening downwards

trontrytel:
This PR adds a term ~ `da/dz * (rho * w * psi)_env` to sgs precipitation sedimentation. The rest is the same as it was last week, before we merged the version that is breaking nightly amip. The term is limited by my interpretation of CFL condition for mixing laterally. 

Lets merge this and see if nightly amip works. If not, I can limit the term more, and think some more

### DIFF
--- a/src/prognostic_equations/advection.jl
+++ b/src/prognostic_equations/advection.jl
@@ -339,6 +339,7 @@ function edmfx_sgs_vertical_advection_tendency!(
     (; ᶜgradᵥ_ᶠΦ) = p.core
 
     FT = eltype(p.params)
+    (; dt) = p
     ᶠJ = Fields.local_geometry_field(axes(Y.f)).J
 
     for j in 1:n
@@ -433,6 +434,7 @@ function edmfx_sgs_vertical_advection_tendency!(
                     ᶜqʲ,
                     ᶠJ,
                     ᶜρ⁰w⁰χ⁰,
+                    dt,
                 )
                 @. ᶜqʲₜ += ᶜinv_ρ̂ * vtt
                 @. Yₜ.c.sgsʲs.:($$j).q_tot += ᶜinv_ρ̂ * vtt
@@ -485,6 +487,7 @@ function edmfx_sgs_vertical_advection_tendency!(
                     ᶜχʲ,
                     ᶠJ,
                     ᶜρ⁰w⁰χ⁰,
+                    dt,
                 )
                 @. ᶜχʲₜ += ᶜinv_ρ̂ * vtt
             end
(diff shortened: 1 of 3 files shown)


src/prognostic_equations/advection.jl
```diff
@@ -493,39 +496,39 @@ function edmfx_sgs_vertical_advection_tendency!(
...
+Sedimenting mass exits laterally (detrainment) and prevents artifical incease
+in precipitation specific humidity due to updraft area changes.
 
-When `∂a/∂z < 0` (updraft widens downward), environment condensate enters through the
-tilted boundary (entrainment). The correction term uses the upwind (donor-cell) principle:
-the entrained sedimentation flux has **environment** properties, not updraft properties.
+When `∂a/∂z < 0` (updraft widens downward), two effects occur:
 
-# Equation
+ 1. **Detrainment + dilution**: captured by `∂_z(ρ w a χ)`, which is fully implicit.
```
sajjadazimi, line 514: No detrainment in this case. Only dilution.

src/prognostic_equations/advection.jl
```diff
@@ -493,39 +496,39 @@ function edmfx_sgs_vertical_advection_tendency!(
...
+ 2. **Entrainment of environment condensate**: environment rain enters through the
```
sajjadazimi, line 515: rain -> sedimenting species

src/prognostic_equations/advection.jl
```diff
@@ -493,39 +496,39 @@ function edmfx_sgs_vertical_advection_tendency!(
...
+    tilted boundary. This is an external source, CFL-limited:
+    `−α_CFL · min(∂a/∂z, 0) · ρ⁰w⁰χ⁰`
```
sajjadazimi, line 517: min is not necessary as da/dz is assumed to be negative in this branch

src/prognostic_equations/advection.jl
```diff
@@ -548,17 +552,21 @@ function updraft_sedimentation!(
...
     ᶠwχ = @. p.scratch.ᶠtemp_scalar_2 = ᶠright_bias(-(ᶜw) * ᶜχ)
-    # Base: within-updraft flux convergence a · ∂_z(ρ w χ)
-    # Entrainment correction: min(∂a/∂z, 0) · (ρ¹w¹χ¹ − ρ⁰w⁰χ⁰)
-    @. vtt =
-        -(ᶜa * ᶜprecipdivᵥ(ᶠρ * Geometry.WVector(ᶠwχ))) +
-        min(∂a∂z, 0) * (ᶜρ * ᶜw * ᶜχ - ᶜρ⁰w⁰χ⁰)
+    @. vtt = ifelse(
+        ∂a∂z < 0,
+        -(ᶜprecipdivᵥ(ᶠρ * Geometry.WVector(ᶠwaχ))) -
+        min(1, ᶜa / max(-min(∂a∂z, 0) * ᶜw * dt, eps(dt))) *
```
sajjadazimi, line 567: The CFL limit must be computed based on `w^0`. Instead of passing w^0 to the function, maybe it's easier to limit `rho w x^0` to `rho x^0 * dz / dt` before passing it to the function (which implies limiting w^0 to dz / dt)

src/prognostic_equations/implicit/manual_sparse_jacobian.jl
```diff
@@ -1199,23 +1200,36 @@ function update_sgs_advection_jacobian!(matrix, Y, p, dtγ)
                 ) - (I,)
 
             # sedimentation
-            # Base: a·∂_z(ρwχ) — always the same regardless of ∂a/∂z sign
-            # Correction: min(∂a/∂z, 0)·(ρ¹w¹χ¹ − ρ⁰w⁰χ⁰)
-            #   ρ⁰w⁰χ⁰ = (w_GS·ρχ_GS − ρa¹·w¹·χ¹)/(1−a), so
-            #   ∂(ρ⁰w⁰χ⁰)/∂χʲ = −ρa¹·w¹/(1−a) and
-            #   ∂/∂χʲ of correction = min(∂a/∂z, 0)·ρ¹w¹/(1−a)
+            # ∂a/∂z < 0 (widening down): ∂_z(ρwaχ) for detrainment/dilution
```
sajjadazimi, line 1203: no detrainment in this case. Only dilution

src/prognostic_equations/implicit/manual_sparse_jacobian.jl
```diff
@@ -1199,23 +1200,36 @@ function update_sgs_advection_jacobian!(matrix, Y, p, dtγ)
...
+                    -(ᶜprecipdivᵥ_matrix()) ⋅ ᶠsed_tracer_advection *
+                    DiagonalMatrixRow(ᶜa) +
                     DiagonalMatrixRow(
-                        min(ᶜ∂a∂z, zero(ᶜ∂a∂z)) * ᶜρʲs.:(1) * ᶜwʲ /
-                        max(1 - ᶜa, eps(eltype(ᶜa))),
-                    )
+                        min(
+                            FT(1),
+                            ᶜa / max(
+                                -min(ᶜ∂a∂z, zero(ᶜ∂a∂z)) * ᶜwʲ * dt,
```
sajjadazimi, line 1225: min is not necessary

sajjadazimi: And w must be env w

src/prognostic_equations/implicit/manual_sparse_jacobian.jl
```diff
@@ -1199,23 +1200,36 @@ function update_sgs_advection_jacobian!(matrix, Y, p, dtγ)
...
+                                eps(FT),
+                            ),
+                        ) * min(ᶜ∂a∂z, zero(ᶜ∂a∂z)) * ᶜρʲs.:(1) * ᶜa *
```
sajjadazimi, line 1228: same as above: min is redundant

src/prognostic_equations/advection.jl
sajjadazimi: Also I think we should remove the contribution of `alpha rho w x ^0` in Jacobian because the limiting factor alpha is a non-trivial function of `w^0` (and as a result x^j). I think it's clearer if we treat the whole sedimentation entrainment term as a constant (no Jacobian contributions).

sajjadazimi (commented):
Thanks! I left a few comments. My main comment is on using env terminal velocity in the computation of the alpha. And also what it means for the Jacobian.

src/prognostic_equations/advection.jl
trontrytel: Thanks for the catch! Will fix it

trontrytel:
Closing. The model seems stable now without the CFL limit

PR #4716, opened 2026-07-21
SGS hyperdiff

sajjadazimi:
## Purpose 
Per-subdomain SGS hyperdiffusion with the dry-static-energy + water-enthalpy split for mseʲ. Reverses the previous unification with grid-mean tendencies.

----

### DIFF
--- a/src/prognostic_equations/hyperdiffusion.jl
+++ b/src/prognostic_equations/hyperdiffusion.jl
@@ -36,11 +36,11 @@ function hyperdiffusion_cache(
     FT = eltype(Y)
     n = n_mass_flux_subdomains(turbconv_model)
 
-    # Grid scale quantities
+    # Grid scale quantities. Energy hyperdiffusion is applied via the
+    # dry-static-energy + water-enthalpy decomposition (mirrors
+    # `vertical_diffusion_boundary_layer_tendency!`), so we DSS ∇²s_d and
+    # ∇²q_{v,l,i} separately instead of a single ∇²h_tot.
     ᶜ∇²u = similar(Y.c, C123{FT})
-    # Energy hyperdiffusion is applied via the dry-static-energy +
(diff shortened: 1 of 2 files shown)


src/prognostic_equations/hyperdiffusion.jl
```diff
@@ -99,8 +123,29 @@ NVTX.@annotate function prep_hyperdiffusion_tendency!(Yₜ, Y, p, t)
...
-    # hyperdiff tendency in `apply_hyperdiffusion_tendency!`.
+    # Sub-grid scale hyperdiffusion. SGS mse uses the same dry-static-energy +
+    # water-enthalpy split as the grid mean (reassembled with subdomain
+    # thermodynamic quantities in `apply_hyperdiffusion_tendency!`).
+    if turbconv_model isa PrognosticEDMFX
+        for j in 1:n
+            @. ᶜ∇²uʲs.:($$j) =
+                C123(wgradₕ(divₕ(ᶜuʲs.:($$j)))) - C123(wcurlₕ(C123(curlₕ(ᶜuʲs.:($$j)))))
+            @. ᶜ∇²uₕʲs.:($$j) = C12(ᶜ∇²uʲs.:($$j))
+            @. ᶜ∇²uᵥʲs.:($$j) = C3(ᶜ∇²uʲs.:($$j))
```
szy21, line 134: Do we need to calculate the horizontal and vertical components here, then recompute ᶜ∇²uʲs later from the two components, instead of just using ᶜ∇²uʲs in line 131? I know this is what we had before, but just wondering.

szy21 (approved):
Thanks!

src/prognostic_equations/hyperdiffusion.jl
sajjadazimi: That's a good question. It seems like we don't need the decomposition. I'd say we merge this PR and then test this and fix another couple of cleanup issues in a followup PR.

PR #4717, opened 2026-07-21
clean up edmf tomls

szy21:
## Purpose 

- Change to min_area_limiter_scale: 0.0001, turb_conv_param_vec: [0.0001, 10000] except for calibrated ones
- Unify some tomls, remove unused ones

## Content

----


sajjadazimi (approved):
Thank you!

config/model_configs/prognostic_edmfx_gabls_column_sparse_autodiff.yml
```diff
@@ -26,5 +26,5 @@ dt: "10secs"
 t_end: "5hours"
 dt_save_state_to_disk: "30mins"
 perturb_initstate: false
-toml: [toml/prognostic_edmfx.toml]
+toml: [toml/prognostic_edmfx_1m.toml]
```
costachris: I think you want `prognostic_edmfx_1M.toml` here and elsewhere

config/model_configs/prognostic_edmfx_bomex_tracerA_column.yml
```diff
@@ -20,7 +20,7 @@ perturb_initstate: false
 dt: "1secs"
 t_end: "6hours"
 dt_save_state_to_disk: "10mins"
-toml: [toml/prognostic_edmfx_bomex.toml]
+toml: [toml/prognostic_edmfx_1M.toml]
```
costachris, line 22: This needs a `cloud_fraction_param_vec` because the config uses `cloud_model: "MLCloud"`, which isn't in `toml/prognostic_edmfx_1M.toml`. this should probably use `toml/prognostic_edmfx_1m_mlcloud.toml `

toml/prognostic_edmfx_1M_mlcloud.toml
```diff
@@ -1,3 +1,12 @@
+[zd_rayleigh]
```
costachris, line 1: maybe rename this file to `toml/prognostic_edmfx_1M_mlcloud.toml` for name consistency?

costachris (approved):

config/model_configs/prognostic_edmfx_bomex_tracerA_column.yml
szy21: Oh, right, I'll change the cloud_model in this config then. MLCloud is tested in another job.

toml/prognostic_edmfx_simpleplume.toml
```diff
@@ -7,20 +7,17 @@ value = 1.0e-5
 [EDMF_max_area]
 value = 0.7
 
-[entr_inv_tau]
-value = 0.0001
+[turb_entr_param_vec]
+value = [0.0001, 10000.0]
 
 [entr_coeff]
-value = 0
```
costachris, line 14: I think `entr_coeff = 0` (no horizontal mixing) and the area entr limiters off (`min_area_limiter_scale = 0`), like before,  are probably better assumptions for simple plume setup? Turb entrainment might need to be 0 too.

szy21: let me see what the simple plume is trying to do. If entrainment is 0 everywhere then a plume can't start (although now we change the boundary condition, so it might be different). Before we start a plume by having a positive entr_inv_tau, and I change it to a positive entr_coeff to be more similar to other cases.

PR #4718, opened 2026-07-21
Wu/cloudsat simulator

cflsepyt:
# CloudSat simulator

Simulator for the 94 GHz CloudSat Cloud Profiling Radar (CPR) following COSPv2. At each callback, the model samples grid-mean cloud and precipitation into subcolumns using the configured overlap assumption. For each subcolumn, the simulator:

1. distributes cloud liquid, cloud ice, rain, and snow from the 1M microphysics scheme
2. diagnoses particle sizes using the 1M particle-size distributions
3. computes hydrometeor backscatter and extinction, together with atmospheric gas attenuation
4. integrates two-way attenuation downward from the model top
5. calculate total cloud cover and CFAD

## Configuration

The simulator is enabled by assigning a finite value to `dt_subcol`. It currently produces CloudSat diagnostics only with the 1M microphysics scheme.

```yaml
microphysics_model: "1M"

dt_subcol: "10mins"
cosp_n_subcolumns: 100
cosp_overlap: "maximum_random"

diagnostics:
  - short_name: [cloudsat_tcc, cloudsat_tcc2]
    period: "10mins"

### CODE REVIEW

File: src/cache/precomputed_quantities.jl [hunk 1]
```diff
@@ -226,6 +226,43 @@ function precomputed_quantities(Y, atmos)
...
+        @. hydro_path_attenuation_cloudsat_work = zero(FT)
+        @. gas_path_attenuation_cloudsat = zero(FT)
+        for size in values(cloudsat_grid_mean_sizes)
+            @. size = zero(FT)
+        end
+        for DBZe_subcolumn in DBZe_cloudsat
+            @. DBZe_subcolumn = FT(-1e30)
+        end
+        detected_column_cloudsat .= false
+        cloudsat_tcc .= zero(FT)
```
szy21: These should be allocated only when cosp is on. Also do we have to cache DBZe_cloudsat?

szy21:
In general, I want to separate the cosp code from ClimaAtmos more cleanly, so it would be easier if we want to refactor it, and users who don't care about cosp don't need to worry about it. For example, could you move cosp precomputed quantities and callbacks into separate files in the cosp folder, and include those when building the cache and callbacks?

src/cosp/cloudsat_const_tables.jl
nefrathenrici, line 1: We should verify the data in this file to ensure it doesn't introduce silent bugs.

src/cache/precomputed_quantities.jl
```diff
@@ -226,6 +226,43 @@ function precomputed_quantities(Y, atmos)
...
+        Ze_non_cloudsat_work = similar(Y.c, FT)
+        hydro_path_attenuation_cloudsat_work = similar(Y.c, FT)
+        gas_path_attenuation_cloudsat = similar(Y.c, FT)
+        height_km_cloudsat = Fields.coordinate_field(axes(Y.c)).z ./ FT(1000)
+        top_height_km_cloudsat =
+            Fields.level(
+                Fields.coordinate_field(Spaces.face_space(axes(Y.c))).z,
+                Spaces.nlevels(axes(Y.c)) + Fields.half,
+            ) ./ FT(1000)
+        DBZe_cloudsat = ntuple(_ -> similar(Y.c, FT), nsubcolumns)
```
nefrathenrici: This is a large allocation we will have to optimize away.

src/callbacks/callbacks.jl
```diff
@@ -97,15 +96,150 @@ end
...
+        consumer.kr_vol_work,
+        consumer.hydro_path_attenuation_work,
+        consumer.gas_path_attenuation,
+        consumer.height_km,
+        consumer.top_height_km,
+    )
+    return nothing
+end
+
+function run_cosp_cloudsat!(Y, p, ::NonEquilibriumMicrophysics1M)
```
nefrathenrici: The types above and this code are not callbacks and should be moved out of this file. I suggest moving this to src/cosp/cloudsat.jl

src/cosp/cloudsat_optics.jl
```diff
@@ -0,0 +1,583 @@
...
+    return CloudSatRadarConfig{FT}(
+        FT(freq),
+        FT(k2),
+        use_gas_abs != 0,
+        FT(min_mixing_ratio),
+    )
+end
+
+struct Clima1MPSDParameters{P}
+    phase::Symbol
```
nefrathenrici: I don't think this will be GPU-compatible.

src/cache/precomputed_quantities.jl
```diff
@@ -226,6 +226,43 @@ function precomputed_quantities(Y, atmos)
...
+            r_lcl = similar(Y.c, FT),
+            lambda_inv_icl = similar(Y.c, FT),
+            lambda_inv_rai = similar(Y.c, FT),
+            lambda_inv_sno = similar(Y.c, FT),
+        )
+        nsubcolumns = _cosp_nsubcolumns(atmos.cosp.n_subcolumns)
+        z_vol_cloudsat_work = similar(Y.c, FT)
+        kr_vol_cloudsat_work = similar(Y.c, FT)
+        g_vol_cloudsat = similar(Y.c, FT)
+        Ze_non_cloudsat_work = similar(Y.c, FT)
```
nefrathenrici: These scratch/temporary fields should be moved to `implicit_temporary_quantities` or `temporary_quantities`

nefrathenrici (dismissed):
This looks good so far. There are a few issues to fix to make this code performand better organized.

src/cosp/cloudsat.jl
```diff
@@ -0,0 +1,360 @@
...
+    (;
+        cloudsat_hydrometeor_optics_work,
+        g_vol_cloudsat,
+        DBZe_cloudsat_work,
+        gas_path_attenuation_cloudsat,
+        cloudsat_grid_mean_sizes,
+        detected_column_cloudsat,
+    ) = p.scratch
+
+    ᶜq_vap = @. lazy(ᶜq_tot_nonneg - ᶜq_liq - ᶜq_ice)
```
szy21: You can use `TD.vapor_specific_humidity` here

src/cosp/cloudsat_optics.jl
```diff
@@ -0,0 +1,575 @@
...
+    radar_cfg.use_gas_abs || return zero(FT)
+    return _quickbeam_gases(_pressure_hpa(p), T, max(qv, zero(FT)), radar_cfg.freq)
+end
+
+@inline function _quickbeam_gases(pres_hpa, T, qv, freq)
+    FT = typeof(pres_hpa + T + qv + freq)
+    T_safe = max(T, FT(1))
+    th = FT(300) / T_safe
+    e = qv * pres_hpa / (qv + FT(0.622)) / FT(1000)
+    p = pres_hpa / FT(1000) - e
```
nefrathenrici:
```suggestion
    e = qv * pres_hpa / (qv + FT(0.622)) / FT(10)
    p = pres_hpa / FT(10) - e
```

Wrong conversion factor

src/cosp/cloudsat_const_tables.jl
```diff
@@ -0,0 +1,609 @@
...
+    52.5423930, 53.0669060, 53.5957480, 54.1299999, 54.6711570, 55.2213650,
+    55.7838000, 56.2647770, 56.3378700, 56.9681000, 57.6124810, 58.3238740,
+    58.4465890, 59.1642040, 59.5909820, 60.3060570, 60.4347750, 61.1505580,
+    61.8001520, 62.4112120, 62.4862530, 62.9979740, 63.5685150, 64.1277640,
+    64.6789000, 65.2240670, 65.7647690, 66.3020880, 66.8368270, 67.3695950,
+    67.9008620, 68.4310010, 68.9603060, 69.4890210, 70.0173420, 118.7503410,
+    368.4983500, 424.7631200, 487.2493700, 715.3931500, 773.8387300,
+    834.1453300,
+)
+const O2_A1 = (
```
nefrathenrici, line 22: I think this table is 100x too small. We really need to validate these tables before trusting this code.

szy21: where is this table from?

cflsepyt: The gas absorption table is from: Liebe, H. J. (1985), Radio Science, 20, 1069–1089. doi:10.1029/RS020i005p01069

The ice refractive index is from Warren, S. G. (1984), Applied Optics, 23, 1206–1225. doi:10.1364/AO.23.001206

cflsepyt: There were some transcription errors in COSP, and those errors were carried over when I ported the table from COSP, including the error that @nefrathenrici  mentioned. I’ve fixed them now, so the table should be correct.

szy21: Is this a bug in COSP then? It would be good to open an issue.

test/cosp/subcol_test.jl
```diff
@@ -369,191 +258,93 @@ end
         end
     end
 
-    @testset "COSPv2 golden references" begin
```
nefrathenrici, line 372: Why were these tests removed?

src/cosp/cosp_quantities.jl
```diff
@@ -0,0 +1,80 @@
...
+    top_height_km_cloudsat =
+        Fields.level(
+            Fields.coordinate_field(Spaces.face_space(axes(Y.c))).z,
+            Spaces.nlevels(axes(Y.c)) + Fields.half,
+        ) ./ FT(1000)
+    cloudsat_dbze_bin_edges =
+        COSP.COSPCloudSatCFAD.cloudsat_cfad_bin_edges(FT)
+    cloudsat_dbze_bin_centers =
+        COSP.COSPCloudSatCFAD.cloudsat_cfad_bin_centers(FT)
+    cfadDbze94 = similar(Y.c, typeof(cloudsat_dbze_bin_centers))
```
nefrathenrici, line 22: Will diagnostics for cfadDbze94 be added in a future PR?

src/cosp/cloudsat.jl
```diff
@@ -0,0 +1,360 @@
...
+    work::W
+    atmosphere::A
+    optics::O
+    statistics::S
+end
+
+function (consumer::CloudSatSubcolumnConsumer)(_, hydrometeors)
+    (; work, atmosphere, optics, statistics) = consumer
+    hydrometeor_optics = work.hydrometeor_optics
+    COSP.COSPCloudSatOptics.cloudsat_optics_subcolumn!(
```
nefrathenrici, line 16: This is called redundantly for every subcolumn and could be replaced with a mask, since `hydrometeors` only contains two values per cell. This could be done in a followup PR, but it could be called in between `prepare_cosp_subcolumns!` and the loop over the subcolumns `foreach_prepared_cosp_subcolumn!`.

src/cosp/cloudsat.jl
```diff
@@ -0,0 +1,360 @@
...
+)
+    # The subcolumn generator supports 2M, but CloudSat hydrometeor optics do
+    # not. Preserve the sampled-fraction workflow while returning explicit
+    # unsupported simulator outputs.
+    foreach_cosp_subcolumn(NoOpCOSPSubcolumnConsumer(), Y, p)
+    reset_cloudsat_statistics!(p.precomputed)
+    return nothing
+end
+
+function run_cosp_cloudsat!(Y, p, _)
```
nefrathenrici: This should error instead of silently doing nothing for unsupported microphysics models.

src/cosp/cosp_quantities.jl
cflsepyt: Yes, but we would have to modify ClimaDiagnostics to support an auxiliary dimension for the reflectivity bin.

szy21: @ph-kev how much work would it be to add this functionality?

src/cosp/cloudsat.jl
cflsepyt: Ok, I'll address this in a follow-up PR.

src/cosp/cloudsat_optics.jl
```diff
@@ -0,0 +1,575 @@
...
+        hydrometeors,
+        grid_mean_sizes,
+        temperature,
+        rho_air,
+        microphysics_params,
+        radar_cfg,
+    )
+
+Compute hydrometeor optical quantities for one streamed subcolumn using particle
+sizes diagnosed from the grid-mean state. The composite output field is working
```
szy21, line 170: I think it might be more physically correct to diagnose the particle size from the in-cloud state. Could you double check?

szy21 (approved):
The PR looks good. I'm fine with merging it after the comments are addressed, and we can fix any remaining issue in following PRs. It would be good to add some docs, and run a global simulation with it and check if the cloud cover diagnostics make sense.

src/cosp/cosp_quantities.jl
ph-kev: I think it would be a decent amount of work to add it since we need to let the user pass the dimension values and its metadata to be saved in the NetCDF file and it is not obvious to me what the input is going to be when writing the data into the NetCDF file.

It would help if an issue is opened for this since I am not sure what is required beyond being able to store another dimension in the NetCDF files.

src/cosp/cloudsat_optics.jl
cflsepyt: That was the original method. However, it produced an overestimation of radar reflectivity because reflectivity is highly sensitive to particle size.

cflsepyt: There are two possible ways to construct the subcolumn particle distribution from the model's grid-mean hydrometeor amount:

1. Keep the particle number concentration fixed and diagnose particle size from the in-cloud amount
2. Keep the particle size diagnosed from the grid-mean state and increase the particle number concentration within cloudy subcolumns to conserve mass

The current implementation uses the second approach

szy21: It seems a bit strange to me as the prescribed particle number concentration in 1M microphysics should be for the cloudy air. But if option 2 gives better results that's fine, just make sure to document it clearly.

szy21:
ci is finicky, I'll manually merge this in.

PR #4719, opened 2026-07-21
hyperdiff cleanup

sajjadazimi:
## Purpose 
Clean up hyperdiffusion

----

### DIFF
--- a/src/prognostic_equations/hyperdiffusion.jl
+++ b/src/prognostic_equations/hyperdiffusion.jl
@@ -25,14 +25,12 @@ function ν₄(hyperdiff, Y)
 end
 
 function hyperdiffusion_cache(Y, atmos)
-    (; hyperdiff, turbconv_model, microphysics_model) = atmos
+    (; hyperdiff, turbconv_model) = atmos
     isnothing(hyperdiff) && return (;)  # No hyperdiffiusion
-    hyperdiffusion_cache(Y, hyperdiff, turbconv_model, microphysics_model)
+    hyperdiffusion_cache(Y, hyperdiff, turbconv_model)
 end
 
-function hyperdiffusion_cache(
-    Y, ::Hyperdiffusion, turbconv_model, microphysics_model,
-)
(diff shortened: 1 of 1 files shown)


src/prognostic_equations/hyperdiffusion.jl
```diff
@@ -128,10 +124,13 @@ NVTX.@annotate function prep_hyperdiffusion_tendency!(Yₜ, Y, p, t)
...
-                C123(wgradₕ(divₕ(ᶜuʲs.:($$j)))) - C123(wcurlₕ(C123(curlₕ(ᶜuʲs.:($$j)))))
-            @. ᶜ∇²uₕʲs.:($$j) = C12(ᶜ∇²uʲs.:($$j))
-            @. ᶜ∇²uᵥʲs.:($$j) = C3(ᶜ∇²uʲs.:($$j))
+            # Only the C3 (vertical) component of ∇²uⱼ is needed for the
+            # u₃ⱼ hyperdiff (no prognostic uₕⱼ). By the vector Laplacian
+            # identity, C3(∇²_h u) depends only on u_C3, and the
+            # `wgradₕ(divₕ(⋅))` piece contributes zero to C3, so:
+            #   C3(∇²_h uⱼ) = -C3(wcurlₕ(C123(curlₕ(uⱼ)))).
+            @. ᶜ∇²uᵥʲs.:($$j) =
+                -C3(wcurlₕ(C123(curlₕ(ᶜuʲs.:($$j)))))
```
szy21: Is this correct when there is topography? It probably doesn't matter much though.

szy21 (approved):

src/prognostic_equations/hyperdiffusion.jl
sajjadazimi: According to Claude, `C3(wgradₕ(...))` is structurally zero (since `wgradₕ` returns a `Covariant12Vector`). So only the curl curl part contribute here.

PR #4722, opened 2026-07-23
not include precipitation in cloud fraction and radiation

szy21:
## Purpose 

Currently, precipitation (rain and snow) is included in the radiation scheme. However, RRTMGP assumes a single effective radius for all liquid water and a single effective radius for all ice. Applying the cloud droplet effective radius to rain is unphysical, as raindrops are significantly larger than cloud droplets. Because rain cannot be represented accurately, we have excluded it, and for consistency, we have excluded snow from the radiative calculations as well. We can come back to it if we can treat cloud condensate and precipitation differently in RRTMGP in the future.

Note: `clwvi` and `clivi` diagnostics are not changed. This may not be consistent with the CMIP convention "only including precipitating hydrometeors when they are included in radiation". I didn't change it because otherwise `clivi` is just a duplicate of `iwp`. We can change it in the future if it's important.

## Content

----

### DIFF
--- a/config/model_configs/prognostic_edmfx_trmm_column.yml
+++ b/config/model_configs/prognostic_edmfx_trmm_column.yml
@@ -20,7 +20,7 @@ z_elem: 82
 z_stretch: false
 perturb_initstate: false
 dt: 120secs
-t_end: 12hours
+t_end: 6hours
 dt_save_state_to_disk: 60mins
 toml: [toml/prognostic_edmfx_1M.toml]
 netcdf_interpolation_num_points: [2, 2, 82]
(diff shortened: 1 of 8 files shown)


config/model_configs/prognostic_edmfx_trmm_column.yml
```diff
@@ -20,7 +20,7 @@ z_elem: 82
 z_stretch: false
 perturb_initstate: false
 dt: 120secs
-t_end: 12hours
+t_end: 6hours
```
szy21, line 23: Changed t_end so the ci plots are more meaningful

costachris (approved):

PR #4726, opened 2026-07-24
Port GCMDriven SCM case to the generic column forcing interface

nefrathenrici:
Changes
- Added in-memory data reader `InMemoryColumnData`
- Added new supertype `AbstractColumnData` to group `InMemoryColumnData` and `ColumnDataset`
- Added `GCMColumnData.read` which reads the grouped cfsite netCDF, time-averages it into steady profiles, and returns an `InMemoryColumnData`
- Added `era5_dataset` to consolidate `ReanalysisTimeVarying` and `ReanalysisMonthlyAveragedDiurnal` cases
- Removed the `GCMDriven` Setup. To use this case, set `initial_condition: GCM` as well as `cfsite_number` and `external_forcing_file`
- Trimmed redundant `external_forcing` options to just `nothing` and `ReanalysisMonthlyAveragedDiurnal`


Julians42:
I have to submit a couple things to AGU before I can spend more time reviewing but wanted to ask about ReanalysisTimeVarying and ReanalysisMonthlyAveragedDiurnal consolidation. ReanalysisTimeVarying was set up to read from ERA5 for hourly data (potentially many days of forcing for a simulation, but ended up being tricky to calibrate). While ReanalysisMonthlyAveragedDiurnal was set up for hourly data but with data averaged over a longer period of time. I'm assuming the changes you've made just take advantage of the fact that this is essentially the same thing just with different amounts of ERA5 data (e.g. the monthly setup only has one day of data - which is an average in itself). 

I'm just wondering why we shouldn't name the setup simply `ERA5` or `Reanalysis` since you've been able to combine them? Might be cleaner and would pair nicely with `initial_condition: GCM`. Maybe its the insolation that's different?

src/column_datasets/ColumnDatasets.jl
```diff
@@ -488,6 +531,125 @@ function surface_timevaryinginputs(
...
 end
 
+# ============================================================================
+# In-memory column data
+# ============================================================================
+
+"""
+    InMemoryColumnData(; z, column, surface, site_location = nothing, source)
+
+A column data source backed by in-memory arrays instead of a file. `z` is the
```
Julians42, line 541: Is there a memory limit? E.g., for how many days could we have a simulation. Is this even a concern? Fine if you don't know

src/column_datasets/ColumnDatasets.jl
```diff
@@ -488,6 +531,125 @@ function surface_timevaryinginputs(
...
+    field = similar(ᶜz)
+    itp = Intp.extrapolate(
+        Intp.interpolate(
+            (Float64.(z_src),),
+            Float64.(vals_src),
+            Intp.Gridded(Intp.Linear()),
+        ),
+        Intp.Flat(),
+    )
+    parent(field) .= itp.(parent(ᶜz))
```
Julians42, line 613: Request not using parent. Maybe  `Fields.field2array`? Why can't you `.=` here to fill `field`?

src/column_datasets/GCMColumnData.jl
```diff
@@ -0,0 +1,105 @@
...
+temperature `ts` (time mean), and a constant `coszen`/`rsdt` chosen so that
+`coszen = coszen[1]` and `rsdt / coszen = mean(rsdt / coszen)`.
+"""
+function read(path, cfsite_number; thermo_params)
+    g = TDP.grav(thermo_params)
+    NC.NCDataset(path, "r") do ds
+        site = ds.group[cfsite_number]
+        tmean(v) = vec(mean(site[v][:, :], dims = 2))
+
+        z = tmean("zg")
```
Julians42, line 45: Is there a check for these fields existing in the dataset somewhere earlier? Otherwise should add a check

src/column_datasets/GCMColumnData.jl
```diff
@@ -0,0 +1,105 @@
...
+`wa = w̄ₐₚ · (−ᾱ) / g` (mean pressure velocity and specific volume), density is
+`mean(1/α)`, and the vertical-fluctuation tendencies `tntva`/`tnhusva` are the
+eddy part of the GCM's total vertical advective tendency, `tntva + w̄ · ∂χ̄/∂z`,
+with the vertical gradient differenced on the GCM grid (Shen et al. 2022).
+
+Surface variables reproduce the previous `GCMDrivenInsolation`: the skin
+temperature `ts` (time mean), and a constant `coszen`/`rsdt` chosen so that
+`coszen = coszen[1]` and `rsdt / coszen = mean(rsdt / coszen)`.
+"""
+function read(path, cfsite_number; thermo_params)
```
Julians42, line 39: Read is generic naming - suggest more informative name unless you're confident the function won't conflict with other reads / it's not used elsewhere

nefrathenrici: Agreed, renamed to `read_cfsite`.

src/column_datasets/GCMColumnData.jl
nefrathenrici: Added

nefrathenrici:
> I'm just wondering why we shouldn't name the setup simply `ERA5` or `Reanalysis` since you've been able to combine them? Might be cleaner and would pair nicely with `initial_condition: GCM`. Maybe its the insolation that's different?

I agree we could rename it to be simpler, but that would be a breaking change and increase the PR's scope. I can make an issue and rename in a followup PR.

src/column_datasets/ColumnDatasets.jl
nefrathenrici: Done:
```
field = similar(ᶜz)
field .= itp.(ᶜz)
```

src/column_datasets/ColumnDatasets.jl
nefrathenrici: This should only be used to populate column fields. Right now the `InMemoryColumnData` struct just has one vertical dimension, no time axis, so the memory requirement does not grow with run length here. I added a line to clarify this in the docstring.

config/default_configs/default_config.yml
```diff
@@ -362,7 +362,7 @@ check_steady_state:
   value: false
 # Single-column model (SCM) forcing
 external_forcing:
-  help: "External forcing for single-column experiments. Leave `~` for almost every case: the forcing is then taken from the setup chosen by `initial_condition` (the preferred route). Set it only to apply forcing that differs from the initial condition — the intended use is `ReanalysisMonthlyAveragedDiurnal` (monthly-averaged diurnal ERA5 forcing, which has no `initial_condition` counterpart), paired with `initial_condition: ReanalysisTimeVarying`. The values `GCM`/`ISDAC`/`ForcingFromFile`/`ReanalysisTimeVarying` merely mirror `initial_condition` and are redundant with it. [`~` (default), `ReanalysisMonthlyAveragedDiurnal`, `GCM`, `ISDAC`, `ForcingFromFile`, `ReanalysisTimeVarying`]"
```
costachris, line 365: Can you also update `docs/column_datasets_reference.md` to include details on how to configure a GCM driven run, given the refactor.

costachris (approved):
Can you note the behavior change for eddy vertical fluctuation, which is now computed on the GCM grid vs the interpolated SCM grid before. The new way seems more accurate but might change the forcing relative to before, esp at sharp gradients

PR #4727, opened 2026-07-28
feat(perf): add draft timing harness

### LINKED ISSUE #4728: Measurement infrastructure
To enable performance optimisation we need to set up the infrastructure to collect performance metrics from smaller, isolated parts of ClimaAtmos. 

The metrics we want to collect, will be used later to guide performance optimisations. 

Metrics we wish to  be able to collect are: 
- Total runtime (GPU)
- GPU time resolved into kernels
- Kernel run information (shmem and register usage) 
- Kernel usage (number of launches), total enqueue time

The 'isolated parts' we want to consider in the first instance, are the ones used on the [`tr/perf-scripts` branch](https://github.com/CliMA/ClimaAtmos.jl/tree/tr/perf-scripts/run-and-time).

Mikolaj-A-Kowalski:
## Purpose 

Propose a draft of architecture for modular way to collect performance metrics for isolated parts of the code.

## TODO

- [x] Move other components (cases) from the [tr/perf-scripts](https://github.com/CliMA/ClimaAtmos.jl/tree/tr/perf-scripts/run-and-time) branch over. At the moment only `sgs` is present.
- [x] Add documentation 
- [x] Add a smoke test in CI to verify it runs

## Content

@dennisYatunin  @imreddyTeja 
This in preparation for the following task of trying to isolate pieces of the ClimaAtmos functionality and iterate on them to find performance opportunities.  

The problem that we want to address is that we expect to have many cases representing different pieces of ClimaAtmos as well as different metrics we would wish to collect. Furthermore, we wish to avoid duplication (i.e. copy and paste) of the code to collect metrics of interest into the scripts. 

To solve this we can decouple the code that does the collection from the code that contains instructions on how to setup the case and run a piece of code we wish to measure.  

In the `models` directory we collect julia source files that need to define two functions `case_setup` and `case_run`. First one produces some "state" representing the case. `case_run` executes the function we wish to measure. 

The `*_harness.jl` files are envisioned to 'dynamically load' the case via `include` and then apply the measurements.   

At the moment the `timer_harness.jl` is the bare-bones example of just collecting runtime information. If we are happy with the setup we can add extra harnesses to do the profiling with `nsys` or `ncu`.

----

### DIFF
--- a/perf/run-and-time/Project.toml
+++ b/perf/run-and-time/Project.toml
@@ -3,6 +3,7 @@ BenchmarkTools = "6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf"
 CUDA = "052768ef-5323-5732-b1bb-66c8b64840ba"
 ClimaAtmos = "b2c96348-7fb7-4fe0-8da9-78d88439e717"
 ClimaComms = "3a4d1b5c-c61d-41fd-a00a-5873ba7a1b0d"
+DataFrames = "a93c6f00-e57d-5684-b7b6-d8193f3e46c0"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 
 [sources]
--- a/perf/run-and-time/Project.toml
+++ b/perf/run-and-time/Project.toml
@@ -1,0 +1,9 @@
+[deps]
+BenchmarkTools = "6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf"
+CUDA = "052768ef-5323-5732-b1bb-66c8b64840ba"
+ClimaAtmos = "b2c96348-7fb7-4fe0-8da9-78d88439e717"
+ClimaComms = "3a4d1b5c-c61d-41fd-a00a-5873ba7a1b0d"
+Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
+
+[sources]
+ClimaAtmos = {path = "../.."}
--- a/perf/run-and-time/ci_test_models.jl
+++ b/perf/run-and-time/ci_test_models.jl
@@ -1,0 +1,39 @@
+#=
+This file is intended to be used as a smoke test in CI to verify that all
+reduced component models run correctly.
+=#
+using Logging
+
+# List of files to skipe
+case_files = Set([
+    "models/hyperdiff_tendency.jl",
+    "models/implicit_tendency.jl",
+    "models/set_cloud_frac.jl",
+    "models/set_CM_cache.jl",
+    "models/sgs.jl",
+    ])
+
+has_faliures = false
+
+
+for model in case_files
+    # It overrides defined `case_setup` and `case_run` functions
+    include(model)
+    print("$model : ")
+
+    # Do not print the log messages
+    with_logger(NullLogger()) do
+        try
+            s = case_setup()
+            case_run(s)
+            printstyled("OK - it runs\n"; color = :green, bold = true)
+        catch e
+            printstyled("$model : ERROR\n"; color = :red, bold = true)
+            print(" - $e\n")
+            has_faliures = true
+        end
+    end
+end
+
+exit_code = has_faliures ? 1 : 0
+exit(exit_code)
--- a/perf/run-and-time/ci_test_models.jl
+++ b/perf/run-and-time/ci_test_models.jl
@@ -11,7 +11,7 @@ case_files = Set([
     "models/set_cloud_frac.jl",
     "models/set_CM_cache.jl",
     "models/sgs.jl",
-    ])
+])
 
 has_faliures = false
 
--- a/perf/run-and-time/ci_test_models.jl
+++ b/perf/run-and-time/ci_test_models.jl
@@ -4,7 +4,6 @@ reduced component models run correctly.
 =#
 using Logging
 
-# List of files to skipe
 case_files = Set([
     "models/hyperdiff_tendency.jl",
     "models/implicit_tendency.jl",
@@ -30,7 +29,7 @@ for model in case_files
         catch e
             printstyled("$model : ERROR\n"; color = :red, bold = true)
             print(" - $e\n")
-            has_faliures = true
+            global has_faliures = true
         end
     end
 end
--- a/perf/run-and-time/ci_test_models.jl
+++ b/perf/run-and-time/ci_test_models.jl
@@ -1,0 +1,39 @@
+#=
+This file is intended to be used as a smoke test in CI to verify that all
+reduced component models run correctly.
+=#
+using Logging
+
+# List of files to skipe
+case_files = Set([
+    "models/hyperdiff_tendency.jl",
+    "models/implicit_tendency.jl",
+    "models/set_cloud_frac.jl",
+    "models/set_CM_cache.jl",
+    "models/sgs.jl",
+])
+
+has_faliures = false
+
+
+for model in case_files
+    # It overrides defined `case_setup` and `case_run` functions
+    include(model)
+    print("$model : ")
+
+    # Do not print the log messages
+    with_logger(NullLogger()) do
+        try
+            s = case_setup()
+            case_run(s)
+            printstyled("OK - it runs\n"; color = :green, bold = true)
+        catch e
+            printstyled("$model : ERROR\n"; color = :red, bold = true)
+            print(" - $e\n")
+            has_faliures = true
+        end
+    end
+end
+
+exit_code = has_faliures ? 1 : 0
+exit(exit_code)
--- a/perf/run-and-time/models/hyperdiff_tendency.jl
+++ b/perf/run-and-time/models/hyperdiff_tendency.jl
@@ -1,0 +1,119 @@
+import ClimaComms
+ClimaComms.@import_required_backends
+import ClimaAtmos as CA
+import Random
+Random.seed!(1234)
+
+# ---------------------------------------------------------------------------
+# Minimal setup to call `hyperdiffusion_tendency!`, exposed as the
+# `case_setup`/`case_run` functions expected by the timing harness.
+#
+# `hyperdiffusion_tendency!` only needs `Y` (state), `p` (cache), and `t`, so
+# we skip `CA.get_simulation` (integrator, Jacobian, callbacks, diagnostics,
+# output writers) entirely and build just `Y`/`p`/`t` — this mirrors the
+# `Y`/`p`/`t`-producing subset of `AtmosSimulation{FT}(...)` (see
+# src/simulation/AtmosSimulations.jl). Restart handling is skipped (this
+# config starts from an initial condition, not a restart file).
+#
+# `hyperdiffusion_tendency!` reads only `p.atmos.{hyperdiff, turbconv_model}`,
+# `p.core.ᶜΦ`, `p.precomputed.ᶜT`/`ᶜTʲs`, `p.hyperdiff.*`, and `p.scratch`. It
+# never reads the radiation, aerosol, trace-gas, gravity-wave, tracer, or
+# external-forcing parts of `p`. Those are also the *expensive* part of model
+# setup (RRTMGP compilation + lookup tables, ETOPO topography load + spectral
+# smoothing, aerosol/trace-gas data ingestion, Beres gravity-wave tables), so
+# we strip them from the config below *before* building anything.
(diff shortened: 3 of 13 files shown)


imreddyTeja:
If you find it simpler, each component case can use a shared setup, which creates the whole simulation. This does increase compilation time for some cases, but it is also less brittle to potential unrelated changes to ClimaAtmos. 
For my workflow, I find the simple component case setups make restarting the REPL less painful.

Mikolaj-A-Kowalski:
> If you find it simpler, each component case can use a shared setup, which creates the whole simulation. This does increase compilation time for some cases, but it is also less brittle to potential unrelated changes to ClimaAtmos.
For my workflow, I find the simple component case setups make restarting the REPL less painful.

I guess what we want is to share some basic setup between the local components, but each component can accept it and modify it? My impression is that you are doing something along these lines, right? You read the full configuration, but then in each script you disable not necessary settings? 

I would say that in the first instance we should try to keep the compilation times as short as possible and if we start running into stability problems as ClimaAtmos evolves we can revisit that. I don't know what would be your feeling.

imreddyTeja:
> > If you find it simpler, each component case can use a shared setup, which creates the whole simulation. This does increase compilation time for some cases, but it is also less brittle to potential unrelated changes to ClimaAtmos.
> > For my workflow, I find the simple component case setups make restarting the REPL less painful.
> 
> I guess what we want is to share some basic setup between the local components, but each component can accept it and modify it? My impression is that you are doing something along these lines, right? You read the full configuration, but then in each script you disable not necessary settings?
> 
> I would say that in the first instance we should try to keep the compilation times as short as possible and if we start running into stability problems as ClimaAtmos evolves we can revisit that. I don't know what would be your feeling.

Yes, that is what those scripts are doing. Lets go with the minimal component setup scripts then

Mikolaj-A-Kowalski:
I have added the rest of the component models and created two additional harnesses. 
- One for quick iterative profiling using the build-in profiler. It gives basic GPU time information for kernels and the compile  parameters(?) like local, shmem memory usage and number of registers. 
- One to run with an external profiler like ncu/nsys

At the moment the output is produced to screen. If we need we can potentially change the output format, but at the moment is is not clear to me what would be the best option.

imreddyTeja:
Is this ready for review? Is it something that should be merged into main?

Mikolaj-A-Kowalski:
> Is this ready for review? Is it something that should be merged into main?

I would say... not yet? 

I guess ultimately having a toolset in the repository to help with profiling an optimisation of the sections of the code may be beneficial, it still needs to be used in practice a lot more before it gets merged. 

At the moment the interface is based on what I though might be convenient/useful. We still need to use it to verify it is really the case.

Mikolaj-A-Kowalski:
Rebased on main to avoid problems with breaking changes in CloudMicrophysics.jl `0.37.2`.

Mikolaj-A-Kowalski:
In b4ca8e1581e897adae8d7dfff0e516bd49ee7837 and acb783e3ad3a22844eff86e2cc57d19e88c92465 I have added a test that would run and verify that the 'component' models (I need to figure out better word for this) would run correctly. 

I wanted to do it in a single julia session to avoid redundant recompilation. I think ATM the tests would run on a CPU, but I suspect it is fine since all we want to do is to verify that they build and run correctly. 

The entry in the `full_pipeline.yaml` has been my best guess on how one should add a smoke test like this correctly. I am suspecting it is not quite correct. Any feedback would be appreciated.

I am also think that we will test (and run) CI after this PR is ready to avoid triggering things while still in the draft stage.

imreddyTeja:
> In [b4ca8e1](https://github.com/CliMA/ClimaAtmos.jl/commit/b4ca8e1581e897adae8d7dfff0e516bd49ee7837) and [acb783e](https://github.com/CliMA/ClimaAtmos.jl/commit/acb783e3ad3a22844eff86e2cc57d19e88c92465) I have added a test that would run and verify that the 'component' models (I need to figure out better word for this) would run correctly.
> 
> I wanted to do it in a single julia session to avoid redundant recompilation. I think ATM the tests would run on a CPU, but I suspect it is fine since all we want to do is to verify that they build and run correctly.
> 
> The entry in the `full_pipeline.yaml` has been my best guess on how one should add a smoke test like this correctly. I am suspecting it is not quite correct. Any feedback would be appreciated.
> 
> I am also think that we will test (and run) CI after this PR is ready to avoid triggering things while still in the draft stage.

I think the cpu tests added in the `full_pipeline.yml` are reasonable.

sjavis:
This is probably ready for a review now. Also, should it now have the buildkite label attached to run the tests?

imreddyTeja:
> This is probably ready for a review now. Also, should it now have the buildkite label attached to run the tests?

Yes, we should add the "Launch Buildkite" label now. I'll add it now

perf/run-and-time/Project.toml
```diff
@@ -0,0 +1,10 @@
+[deps]
+BenchmarkTools = "6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf"
+CUDA = "052768ef-5323-5732-b1bb-66c8b64840ba"
+ClimaAtmos = "b2c96348-7fb7-4fe0-8da9-78d88439e717"
+ClimaComms = "3a4d1b5c-c61d-41fd-a00a-5873ba7a1b0d"
+DataFrames = "a93c6f00-e57d-5684-b7b6-d8193f3e46c0"
+Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
+
+[sources]
+ClimaAtmos = {path = "../.."}
```
imreddyTeja, line 10: This extra julia project env isn't strictly needed. We coulde also use the `.buildkite` project env. I personally like having a more minimal project like this one to make precomplation less painful. I'm OK with keeping or removing this project.
If we do keep it, should a `Manifest.toml` be checked in for reproducibility?

perf/run-and-time/ci_test_models.jl
```diff
@@ -0,0 +1,39 @@
...
+            s = case_setup()
+            case_run(s)
+            printstyled("OK - it runs\n"; color = :green, bold = true)
+        catch e
+            printstyled("$model : ERROR\n"; color = :red, bold = true)
+            print(" - $e\n")
+            has_faliures = true
+        end
+    end
+end
```
imreddyTeja, line 35: I think there might be a variable scoping bug here.

```
julia> has_faliures = false
false

julia> for i in 1:3
           try
               error("b")
           catch e
               has_failures = true
           end
       end

julia> has_faliures
false

```

Not sure if this is good julia style, but this should fix it:
```
julia> has_faliures = false
false

julia> for i in 1:3
           try
               error("b")
           catch e
               global has_failures = true
           end
       end

julia> has_failures
true
```

perf/run-and-time/timer_harness.jl
```diff
@@ -0,0 +1,62 @@
...
+
+    benchmark = BenchmarkTools.@benchmark CUDA.@sync begin
+        case_run($s)
+    end
+
+    display(benchmark)
+    @info """Basic run statistics:
+        • Setup time: $setup_time [s]
+        • Warmup (compilation) time: $compile_time [s]
+        • Mean GPU time: $(mean(benchmark.times * 1e-6)) [μs]
```
imreddyTeja: Should this be milli instead of micro?

perf/run-and-time/models/hyperdiff_tendency.jl
```diff
@@ -0,0 +1,119 @@
...
+    # branches the tendency takes.
+    pa = config.parsed_args
+    pa["rad"] = nothing                        # no RRTMGP model / lookup tables
+    pa["aerosol_radiation"] = false
+    pa["prescribed_aerosols"] = String[]       # no aerosol data ingestion
+    pa["time_varying_trace_gases"] = String[]  # no trace-gas data ingestion
+    pa["insolation"] = "idealized"             # unused once `rad` is off
+    pa["non_orographic_gravity_wave"] = false  # no Beres source tables
+    pa["orographic_gravity_wave"] = nothing
+    pa["topography"] = "NoWarp"                # skip ETOPO load + spectral smoothing
```
imreddyTeja, line 61: When I madethe topography simplification, I was assuming it would only affect the the initial local geometry computation during set-up. It does make set-up faster. @dennisYatunin can you confirm that this should not impact runtime performance?

perf/run-and-time/ci_test_models.jl
```diff
@@ -0,0 +1,39 @@
+#=
+This file is intended to be used as a smoke test in CI to verify that all
+reduced component models run correctly.
+=#
+using Logging
+
+# List of files to skipe
```
imreddyTeja: Is this a leftover comment?

perf/run-and-time/Project.toml
imreddyTeja: If not, we shoud probably add an entry to the .gitignore for the manifest

imreddyTeja (commented):
Left some minor comments. I'm not sure why unrelated buildkite tests are failing now...

perf/run-and-time/Project.toml
Mikolaj-A-Kowalski: The only reason this Project exists is that I wanted to have `DataFrames` for the result analysis. As far as I'm aware `.buildkite` does not contain it. I am not sure what policy is there about modifying the `.buildkite` environment. I imagine that it needs to be kept lean with the minimum number of packages and re-resolving it might be quite tricky. 

As for the Manifest it is not committed since I am not sure what the policy/best practice is here :sweat_smile:

perf/run-and-time/ci_test_models.jl
Mikolaj-A-Kowalski: :grimacing: Sorry. Good catch.

perf/run-and-time/Project.toml
imreddyTeja: We aren't adding any "real" uses of this to .buildkite, so it is probably fine to not check in the manifest. Adding to the .gitignore is up to you

perf/run-and-time/ci_test_models.jl
Mikolaj-A-Kowalski: Should be fixed in bb236bb0e355f15d5c47214dc2500cff524b0610

perf/run-and-time/ci_test_models.jl
Mikolaj-A-Kowalski: Indeed. Removed in bb236bb0e355f15d5c47214dc2500cff524b0610

perf/run-and-time/timer_harness.jl
Mikolaj-A-Kowalski: I don't really know. The kernel I was testing with was I think in a range of couple hundred micro seconds.  But it was an arbitrary choice without much though into that really.

Mikolaj-A-Kowalski: I changed it in 04f158864d463a561f4bc296b32bf111095f8448

We can drop it if we decide to stick with micro seconds in the end.

imreddyTeja (approved):
Thank you!

sjavis:
I've rebased onto main and cleaned up the history. As far as I can tell the failing buildkite jobs are unrelated so I will merge it when the checks are complete.

imreddyTeja:
> I've rebased onto main and cleaned up the history. As far as I can tell the failing buildkite jobs are unrelated so I will merge it when the checks are complete.

Yeah, those look unrelated. ci has been flaky lately... I'll bring it up in the software standup later today

Mikolaj-A-Kowalski:
@imreddyTeja I am not quite sure what has happened here with the merge queue.

imreddyTeja:
> @imreddyTeja I am not quite sure what has happened here with the merge queue.

I had to manually trigger the buildkite-ci on the merge pipeline. I'm not sure why. It then failed the smoke tests because the new Julia project this PR adds cannot instantiate. [here](https://buildkite.com/clima/climaatmos-ci/builds/31203#019ff1b6-4d78-4803-8fb8-6a97a690754f/L177) is the failure.

I believe this is because the compats updated last night, and the package manager is not seeing the new ClimaDiagnostics release. You might need to add a manual `Pkg.Registry.update` to the buildkite step.

Mikolaj-A-Kowalski:
I'm afraid I cannot figure out why the `climaatmos-ci` in the merge queue does not seem to trigger... 

I have tried to remove all `[skip-ci]` in the commit messages just in case (but I doubt it was the reason)

PR #4731, opened 2026-07-28
no sponge on grid-mean micro tracers

szy21:
## Purpose 

I kept the sponge for updraft because it may be useful for passive tracers. It shouldn't have an effect on microphysics tracers.

## Content

----


sajjadazimi (approved):
Thanks!

PR #4737, opened 2026-07-30
Move grid, params, setup fields into AtmosModel

### LINKED ISSUE #4696: Programmatic AtmosSimulation(; model, setup) ignores the setup's model traits
When you build a simulation directly with `AtmosSimulation{FT}(; model, setup)`, the setup only supplies the initial state. All the model-physics traits a setup defines are silently ignored. 

Those traits are only read on the config path. The direct `AtmosSimulation{FT}` constructor uses setup only for `Setups.initial_state` and `overwrite_initial_state!`. 

I think that this is indicative of a larger issue created by the `AtmosModel`/`AtmosSimulation` partitioning: the setup effectively needs to be passed into both but is only available to `AtmosSimulation`. On the config path `get_simulation` builds the setup once and threads the same object into `get_atmos` (for `AtmosModel`) and `AtmosSimulation`.

nefrathenrici:
This interface-changing PR moves the grid, parameters, and setup arguments from`AtmosSimulation` to `AtmosModel`. This allows the Setup traits to always be applied. Closes #4696.

Interface changes:
```julia
# Previous -- RCEMIP script must add Setups traits manually to the model
insolation = CA.RCEMIPIIInsolation()
setup = CA.Setups.RCEMIPIIProfile_300()
temperature = CA.Setups.surface_temperature_model(setup)
model = CA.AtmosModel(; insolation, temperature)
simulation = CA.AtmosSimulation{FT}(; model, params, context, grid, setup)

# New -- the setup applies everything properly
model = CA.AtmosModel(grid; params, setup = CA.Setups.RCEMIPIIProfile_300())
simulation = CA.AtmosSimulation(model; context)
```

Changes
- Move grid, params, setup into `AtmosModel`. These non-physics fields are dropped when Adapting for the GPU.
- Added `physics_hash(::AtmosModel)` which hashes the model without grid, params, or setup, so a restart checks the physics only. The grid is separately checked: a restart now errors if the file's grid does not match the model's grid, instead of quietly using the file's.
- For the `AtmosModel` constructor, sources now set each field with clear priority: kwargs > setup traits > defaults
- New Setups.setup_model_traits asks the setup for all its values at once. The config code reads that instead of calling each setup hook itself
- Removed baroclinic_wave and bomex Presets since they are now defined just by the corresponding Setup case


docs/src/first_simulation.md
```diff
@@ -55,30 +62,42 @@ propertynames(Y.f)  # e.g., (:u₃,)
...
-Presets matter beyond brevity: the `setup` argument sets the initial state,
-while the physics comes from the model, so the two have to be chosen together.
-BOMEX with the default dry model would have no moisture to convect. Each preset
-pairs a setup with a matching grid and model. The pairing is minimal rather
-than complete: `Presets.bomex` enables moist physics but no
-turbulence-convection scheme or case forcings; pass
-`model = CA.Presets.prognostic_edmf(Float32)` to add convective transport, or
-run the corresponding YAML case config for the full published setup. See the
-[Presets](api.md#Presets) section of the API for the full list.
+The setup and the physics are not independent: the setup sets the initial state
```
haakon-e, line 91: would it be accurate to say that selecting the physics is a prerequisite for choosing an appropriate setup, in the sense that some setups may be incompatible with some physics (for example, because it lacks certain prognostic variables)?

I'm not requesting any particular changes here, just wondering if clarifying how I/we think about this would help clarify the overall design and dependencies.

test/restart_utils.jl
```diff
@@ -106,7 +106,13 @@ function compare(
     return _compare(pass, v1, v2; name, ignore)
 end
 
-function _compare(pass, v1::T, v2::T; name, ignore) where {T}
+# Don't specialized on the argument types: the cache holds thousands of distinct
+# nested struct and NamedTuple types, and a specialized method would have to be
+# compiled once for each of them. The leaf methods below (numbers, arrays,
+# Fields) still dispatch on type.
+function _compare(pass, @nospecialize(v1), @nospecialize(v2); name, ignore)
+    typeof(v1) === typeof(v2) ||
```
ph-kev: Would it be easier to do what they do in the docs for `@nospecialize`?

```julia
@nospecialize
 f(y) = [x for x in y]
 @specialize
```

and no specialize for all of the _compare functions?

docs/src/glossary.md
```diff
@@ -23,8 +23,10 @@ These are accessed through the integrator after a run, e.g.
 
   - **Setup**: the initial conditions, boundary conditions, and optional forcing
     that define a simulation *case* (BOMEX, DYCOMS, RICO, Held–Suarez, …). See [Setups](@ref).
-  - **Preset**: a one-line shortcut (`CA.Presets`) that bundles a grid, setup, and
-    model choices for a common configuration.
+  - **Preset**: a shortcut (`CA.Presets`) bundling the model choices for a
```
ph-kev: I don't like the word "shortcut" here.

docs/src/glossary.md
```diff
@@ -23,8 +23,10 @@ These are accessed through the integrator after a run, e.g.
...
+    common configuration, as a NamedTuple of `AtmosModel` keyword arguments for
+    its `defaults` slot. Distinct from a setup: a preset carries physics, a
+    setup carries the case.
```
ph-kev: I don't think the last sentence is not clear. It mentions that a preset carries physics but no examples of the `ClimaAtmos` keyword arguments is shown here.

docs/src/interfaces.md
```diff
@@ -4,9 +4,11 @@ ClimaAtmos provides two ways to set up and run simulations. Both produce an
 `AtmosSimulation` that is run with `solve_atmos!`; they differ in where the
 configuration is written.
```
ph-kev: Could you explicitly write that the two interfaces are the script and config interface in the intro?

docs/src/interfaces.md
```diff
@@ -4,9 +4,11 @@ ClimaAtmos provides two ways to set up and run simulations. Both produce an
...
 
-The **script API** builds a simulation from Julia keyword arguments:
-`AtmosSimulation{FT}(; grid, model, setup, dt, t_end, ...)`. It is best for
-interactive exploration, notebooks, and programmatic parameter sweeps.
+The **script API** builds a model on a grid and then wraps it in a simulation:
+`AtmosSimulation(AtmosModel(grid; params, setup, ...); dt, t_end, ...)`. The
+model owns the grid, the parameters, and the setup; the simulation owns run
+control. It is best for interactive exploration, notebooks, and programmatic
```
ph-kev: I think this part is unclear to me. Is "run control" just the time stepping or does it include more stuff than that?

perf/run-and-time/models/hyperdiff_tendency.jl
```diff
@@ -65,8 +65,8 @@ function case_setup()
 
     params = CA.ClimaAtmosParameters(config)
     setup_type = CA.get_setup_type(pa, CA.CAP.thermodynamics_params(params))
-    model = CA.get_atmos(config, params; setup_type = setup_type)
     grid = CA.get_grid(pa, params, config.comms_ctx)
+    model = CA.get_atmos(config, params, grid; setup_type = setup_type)
```
ph-kev, line 69: You can do

```suggestion
    model = CA.get_atmos(config, params, grid; setup_type)
```

but I don't know if that is clearer or not.

src/setups/Setups.jl
```diff
@@ -248,6 +249,30 @@ explicitly, so a configuration can always override the setup's radiation.
...
+
+Evaluate every model-facing trait hook of `setup` once and return the raw
+results as a `NamedTuple`: `subsidence`, `ls_adv`, `scm_coriolis` (raw SCM
+profiles; consumers wrap them into model components), `external_forcing`,
+`insolation`, `radiation_mode`, `prescribed_flow` (model objects),
+`surface` (from [`surface_condition`](@ref)), and `surface_temperature`
+(from [`surface_temperature_model`](@ref)).
+"""
+function setup_model_traits(setup, params, ::Type{FT}) where {FT}
+    return (;
```
ph-kev: I don't like the name of this function. This function is parsing  `setup` to create model components that will be used by `AtmosModel`. I don't understand how traits are involved here.

src/types.jl
```diff
@@ -2398,10 +2413,14 @@ names.
   - `cosp`: `nothing`, or a `COSPModel` for the satellite simulator.
   - `disable_surface_flux_tendency`: Whether to skip applying the surface flux
     tendency, independently of whether surface conditions are computed.
+  - `grid`: The `Grids.AbstractGrid` the model is built on.
+  - `params`: The `ClimaAtmosParameters` for the run.
+  - `setup`: The case setup (a `Setups` case) the model was built from.
```
ph-kev, line 2431: Maybe my mental model is wrong, but if the model is built from a setup, why do we need to keep it around?

test/diagnostics/unit_diagnostics.jl
```diff
@@ -88,33 +88,37 @@ import ClimaCore: Fields, Spaces
 # ---------------------------------------------------------------------------
 
 """
-    build_state_cache(FT, model; grid, kwargs...) -> (Y, p)
+    build_state_cache(FT, model_kwargs; grid, kwargs...) -> (Y, p)
```
ph-kev: It is strange that it is called `model_kwargs` even though it is a positional argument.

src/simulation/restart.jl
```diff
@@ -69,8 +69,10 @@ function handle_restart(
         @warn "Non zero `t_start` passed with a restarting simulation. The provided `t_start` will be ignored."
     end
 
+    # Both sides of the model-hash comparison use hash_physics, which
+    # ignores `grid`, `params`, and `setup`.
     (Y, t_start) = get_state_restart(
-        restart_file, start_date, hash(model), context,
+        restart_file, start_date, hash_physics(model), context,
     )
```
ph-kev: It is not clear to me why it is okay to ignore the `grid`, `params`, and `setup` when checking for restart. Also, I don't ever see `get_state_restart` being used without `hash_physics`. Would this function be `get_state_restart(restart_file, start_date, model, context)` and `hash_physics` is called internally?

src/types.jl
```diff
@@ -2499,19 +2527,114 @@ end
...
+        get(atmos_model_kwargs, :disable_surface_flux_tendency, false)
+
+    prescribed_flow = get(atmos_model_kwargs, :prescribed_flow, nothing)
+
+    # `nothing` for a model that describes physics only
+    grid = get(atmos_model_kwargs, :grid, nothing)
+    params = get(atmos_model_kwargs, :params, nothing)
+    setup = get(atmos_model_kwargs, :setup, nothing)
+
+    return AtmosModel{
```
haakon-e: nit: could do something like
```Julia
fields = (
        water, scm_setup, radiation, turbconv, prescribed_flow, gravity_wave,
        vertical_diffusion, sponge, surface, numerics, chemistry, cosp,
        disable_surface_flux_tendency, grid, params, setup,
)
types = map(typeof, fields)
return AtmosModel{types...}(fields...)
```
to save a few lines and make it slightly easier to edit later (no need to remember order of both types and fields) :)

haakon-e (approved):

src/cache/cache.jl
```diff
@@ -150,13 +148,12 @@ gravity waves, radiation, tracers).
...
-  - `time_varying_trace_gas_names`: Names of time-varying trace gases (e.g. `"CO2"`,
-    `"O3"`).
   - `steady_state_velocity`: Predicted steady-state velocity for the
     `check_steady_state` diagnostic, or `nothing`.
-  - `vwb_species`: Species tuple for the vertical-water-borrowing limiter, or
-    `nothing`.
+
+The prescribed aerosol names, the time-varying trace gas names, and the
+vertical-water-borrowing species tuple come from `atmos` (`atmos.radiation` and
+`atmos.numerics`).
```
ph-kev: Do we need this? This feel like documentation for someone used to the old interface and it doesn't seem relevant to new users.

docs/src/first_simulation.md
```diff
@@ -18,13 +18,20 @@ using Logging # hide
...
+Construction has two steps: `AtmosModel(grid; ...)` defines the physical
+system -- it holds the grid, the parameters, and the case setup -- and
+`AtmosSimulation(model; ...)` defines how to run that system: the timestep, the
+duration, the callbacks, and the output.
+
 `t_end` accepts a number of seconds or a duration string (`secs`, `mins`,
 `hours`, `days`, `weeks`), as does the timestep `dt`. Every other aspect of the
-simulation has a keyword argument too; omitted ones take their defaults.
+model and the simulation has a keyword argument too; omitted ones take their
+defaults.
```
ph-kev: This sentence doesn't seem too informative.

docs/src/first_simulation.md
```diff
@@ -55,30 +62,42 @@ propertynames(Y.f)  # e.g., (:u₃,)
...
-turbulence-convection scheme or case forcings; pass
-`model = CA.Presets.prognostic_edmf(Float32)` to add convective transport, or
-run the corresponding YAML case config for the full published setup. See the
-[Presets](api.md#Presets) section of the API for the full list.
+The setup and the physics are not independent: the setup sets the initial state
+and the case forcings, while the parameterizations come from the rest of the
+model, so the two have to be chosen together. BOMEX with the default dry model
+would have no moisture to convect, which is why `equil_moist_0m` appears above.
+Model presets are keyword bundles for exactly this purpose; each is a weakest
+tier of defaults that the setup and your own keyword arguments override. The
```
ph-kev: This doesn't seem too helpful for the readers. If you wan to add a sentence about the model presets, you can just say that the model presets are reasonable default for common simulations that scientists want to run or something similar to that.

src/ClimaAtmos.jl
```diff
@@ -10,17 +10,20 @@ the sphere, for global weather and climate simulation. Energy, air mass, and
...
-everything up; `solve_atmos!` advances it to `t_end`. Common configurations are
-available as one-line presets in the `Presets` submodule. See the Your First
-Simulation, Governing Equations, and API pages of the documentation.
+A simulation is described by an `AtmosSimulation`, which wraps an `AtmosModel` (the
+physics, together with the grid, the parameters, and the case setup) and adds run
+control and an integrator. Constructing one sets everything up; `solve_atmos!`
+advances it to `t_end`. A case (its initial state, boundary conditions, and
+forcing) comes from the `Setups` submodule; common bundles of physics options
+come from the `Presets` submodule. See the Your First Simulation, Governing
+Equations, and API pages of the documentation.
```
ph-kev: I think it is possible to add a `@ref` here.

src/presets.jl
```diff
@@ -17,105 +18,97 @@ import ..InvZEntrainment, ..BuoyancyVelocityDetrainment
...
 ```
 """
-function dry(; kwargs...)
-    defaults = (; microphysics_model = DryModel())
-    return AtmosModel(; defaults..., kwargs...)
-end
+dry(; kwargs...) = (; microphysics_model = DryModel(), kwargs...)
 
 """
     equil_moist_0m(; kwargs...)
```
ph-kev, line 48: How are users supposed to know what `kwargs` are allowed? Does this stop the user from passing bad inputs?

docs/src/first_simulation.md
nefrathenrici: Partly. Some pairs are incompatible, since a setup can require prognostic variables the physics does not carry. I think the dependency usually runs the other way, as the setup constrains the possible physics: a user chooses BOMEX and then adds physics (microphysics, EDMF) on top of it.

test/restart_utils.jl
nefrathenrici: Thanks, that should be easier.

docs/src/glossary.md
nefrathenrici: Agreed, changed it to be more concrete.

docs/src/glossary.md
nefrathenrici: Rewrote: A preset chooses the parameterizations, while a setup chooses the case.

docs/src/interfaces.md
nefrathenrici: rewrote to be more thorough: "run control" is the timestepping algorithm and Jacobian, callbacks, checkpointing, diagnostics

src/setups/Setups.jl
nefrathenrici: That's fair - this was intended to centralize what a Setup is expected to provide, I can rename it to `model_components`.

src/types.jl
nefrathenrici: Ideally the setup would be discarded, but it supplies runtime functions: `initial_state` and possibly `overwrite_initial_state!`

test/diagnostics/unit_diagnostics.jl
nefrathenrici: This is a NamedTuple that gets splatted into AtmosModel keyword arguments, but they are not used as kwargs here. I renamed it to `model_options`.

src/simulation/restart.jl
nefrathenrici: The grid is checked separately in the `AtmosSimulation` function. The setup does not make sense to check as it provides the initial state, which gets overridden on restart. Parameters are allowed to change between restarted simulations, this is needed for calibration.

src/config/model_getters.jl
```diff
@@ -1150,16 +1126,18 @@ function AtmosWater(config::AtmosConfig, params, ::Type{FT}) where {FT}
...
-function AtmosRadiation(config::AtmosConfig, ::Type{FT}; setup_type = nothing) where {FT}
+function AtmosRadiation(config::AtmosConfig, ::Type{FT}; setup_components) where {FT}
     pa = config.parsed_args
     return AtmosRadiation(;
-        radiation_mode = get_radiation_mode(pa, FT; setup_type),
-        insolation = get_insolation_form(pa; setup_type),
+        radiation_mode = get_radiation_mode(pa, FT; setup_components),
+        insolation = get_insolation_form(pa; setup_components),
+        aerosol_names = Tuple(pa["prescribed_aerosols"]),
+        time_varying_trace_gases = Tuple(pa["time_varying_trace_gases"]),
```
imreddyTeja, line 1140: Is `AtmosRadiation` used within broadcast expressions? If so, I think this could lead to `isbits` issues

NEWS.md
```diff
@@ -4,6 +4,15 @@ ClimaAtmos.jl Release Notes
 main
 ----
 
+- [#4737](https://github.com/CliMA/ClimaAtmos.jl/pull/4737) ![][badge-💥breaking] `AtmosModel` is now built on a grid and owns the
+  parameters and case setup. `AtmosSimulation` wraps a model and carries only
+  run control:
+
+  ```julia
+  model = AtmosModel(grid; params, setup, microphysics_model = ..., ...)
+  sim   = AtmosSimulation(model; dt, t_end, ...)
```
imreddyTeja, line 13: README needs to be updated as well

test/restart_utils.jl
```diff
@@ -106,6 +106,13 @@ function compare(
     return _compare(pass, v1, v2; name, ignore)
 end
 
+# Don't specialize `_compare` on the argument types: the cache holds thousands
+# of distinct nested struct and NamedTuple types, and a specialized method would
+# have to be compiled once for each of them. Dispatch on the declared argument
+# types (numbers, arrays, Fields) still works, only the compiled bodies are
+# shared.
+@nospecialize
```
imreddyTeja, line 119: The `where {T <: ` s below should be removed as well. I'm not sure if the `@nospecialize` macro or `where` wins

PR #4743, opened 2026-08-03
Update coupler hash --perf

imreddyTeja:
## Purpose 
this currently fails on main
## Content

----


ph-kev (approved):
Do you know why the change in this PR doesn't work? https://github.com/CliMA/ClimaAtmos.jl/pull/4601

imreddyTeja:
> #4601

I don't think that PR makes the pipeline.yaml use the latest coupler commit.

nefrathenrici:
The pipeline regularly breaks due to compat restrictions when trying to use ClimaAtmos main with an old coupler hash. We should remove clima-specific compat checks to reduce these errors, otherwise we'll need to keep updating the hash.

PR #4744, opened 2026-08-03
Use non-constant rain sedimentation velocity

trontrytel:
This PR makes the model use non-constant rain sedimentation velocity by default. There is an assumed minimum velocity of about 0.8 m/s to allow for stable computation with 1 Newton iteration. 

In the same PR I also switch on the lateral mixing of precipitation in the updrafts where area fraction decreases with height.

### DIFF
--- a/.buildkite/full_pipeline.yml
+++ b/.buildkite/full_pipeline.yml
@@ -343,13 +343,20 @@ steps:
       - label: ":chart_with_upwards_trend: SCM intercomparison plots"
         retry: *retry_policy
         depends_on:
-          - prognostic_edmfx_soares_column
-          - prognostic_edmfx_gabls_column
-          - prognostic_edmfx_bomex_column
-          - prognostic_edmfx_dycoms_rf01_column
-          - prognostic_edmfx_dycoms_rf02_column
-          - prognostic_edmfx_rico_column
-          - prognostic_edmfx_trmm_column
+          - step: prognostic_edmfx_soares_column
+            allow_failure: true
+          - step: prognostic_edmfx_gabls_column
+            allow_failure: true
+          - step: prognostic_edmfx_bomex_column
+            allow_failure: true
+          - step: prognostic_edmfx_dycoms_rf01_column
+            allow_failure: true
+          - step: prognostic_edmfx_dycoms_rf02_column
+            allow_failure: true
+          - step: prognostic_edmfx_rico_column
+            allow_failure: true
+          - step: prognostic_edmfx_trmm_column
+            allow_failure: true
         command: |
           for JOB in \
             prognostic_edmfx_soares_column \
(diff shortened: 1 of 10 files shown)


.buildkite/full_pipeline.yml
```diff
@@ -344,12 +344,19 @@ steps:
         retry: *retry_policy
         depends_on:
           - prognostic_edmfx_soares_column
+            allow_failure: true
```
trontrytel, line 347: Hopefully this will allow to (finally) generate LES SCM comparison even after SCM case is flagged as breaking reproducibility.

src/parameters/create_parameters.jl
```diff
@@ -384,7 +384,7 @@ function TurbulenceConvectionParameters(
         cloud_fraction_floor_residual = FT(0),
         # Lateral correction scaling for updraft sedimentation
         # (see `updraft_sedimentation!`). 1.0 = full correction, 0.0 = disabled.
-        sedimentation_lateral_coeff = FT(0),
+        sedimentation_lateral_coeff = FT(1), # Testing if stable now. To be removed, if yes.
```
trontrytel, line 540: Testing with lateral mixing of precipitation through updraft top if area becomes smaller with height.

config/default_configs/default_config.yml
```diff
@@ -263,8 +263,17 @@ microphysics_n_substeps:
...
-fixed_terminal_velocity:
-  help: "Use fixed terminal velocities for microphysics tracers [`true` (default), `false`]"
+fixed_terminal_velocity_liquid:
+  help: "Use fixed terminal velocity for cloud liquid sedimentation [`true` (default), `false`]"
+  value: true
+fixed_terminal_velocity_ice:
+  help: "Use fixed terminal velocity for cloud ice sedimentation [`true` (default), `false`]"
+  value: true
+fixed_terminal_velocity_rain:
+  help: "Use fixed terminal velocity for rain sedimentation [`true` (default), `false`]"
```
sajjadazimi: `true`, `false` (default)

src/parameterized_tendencies/microphysics/terminal_velocity_utils.jl
```diff
@@ -48,33 +66,46 @@ terminal_velocity(
...
     cmp,
     ρ,
-    q,
-) = CM1.terminal_velocity(cmp.precip.rain, cmp.terminal_velocity.rain, ρ, q)
+    q::FT,
+) where {FT} = CM1.terminal_velocity(
+    cmp.precip.rain,
+    cmp.terminal_velocity.rain,
+    ρ,
+    max(q, FT(1e-8)),
```
sajjadazimi, line 146: We can define a function in CM that computes q_min based on the assumed distribution and minimum of raindrop diameter. For Marshall-Palmer dist I think it is D = 4/lambda = 4 (q/pi rho N_0)^1/4. So q = 1e-8 gives D = 84 um.

sajjadazimi (approved):
Thanks, I just left two minor comments. One is just a suggestion.

szy21 (approved):
It would be good to add to NEWS

src/parameterized_tendencies/microphysics/terminal_velocity_utils.jl
trontrytel: Thats neat! I don't want to delay this PR more. But I'll add it to CM.jl next time I do a release there

PR #4751, opened 2026-08-07
Move 1M options into AtmosModel

### LINKED ISSUE #4709: 1-moment microphysics options only available via config
The new 1-moment microphysics knobs in the default config (rain_autoconversion, cloud_ice_formation, cloud_liquid_rain_accretion, etc) can only be set through the YAML pipeline. A caller building a model programmatically is not able to select any of them.

nefrathenrici:
This PR moves the CloudMicrophysics 1M process options into NonEquilibriumMicrophysics1M within AtmosModel. This enables direct use of these options in ClimaAtmos run scripts:

```julia
import ClimaAtmos as CA

microphysics_model = CA.NonEquilibriumMicrophysics1M(;
    n_substeps = 3,
    n_substeps_quad = 2,
    rain_autoconversion = CA.CMP.PrescribedNd(),        # instead of Kessler1M()
    cloud_ice_formation = CA.CMP.TemperatureDependent(), # instead of ConstantTimescale()
    snow_autoconversion = CA.CMP.WithSupersaturation(),
    rain_snow_accretion = nothing,                    # process off
)

simulation = CA.Presets.aquaplanet(
    Float64;
    model = CA.Presets.nonequil_moist_1m(; microphysics_model),
    t_end = 86400,
)
CA.solve_atmos!(simulation)
```

### DIFF
--- a/src/config/model_getters.jl
+++ b/src/config/model_getters.jl
@@ -11,7 +11,11 @@ function get_microphysics_model(parsed_args, params = nothing)
     elseif model_name == "1M"
         n_substeps = parsed_args["microphysics_n_substeps"]
         n_substeps_quad = parsed_args["microphysics_n_substeps_quadrature"]
-        NonEquilibriumMicrophysics1M(; n_substeps, n_substeps_quad)
+        NonEquilibriumMicrophysics1M(;
+            n_substeps,
+            n_substeps_quad,
+            get_microphysics_1m_options(parsed_args)...,
+        )
     elseif model_name == "2M"
         NonEquilibriumMicrophysics2M()
     elseif model_name == "2MP3"
@@ -27,11 +31,12 @@ end
     get_microphysics_1m_options(parsed_args)
 
 Parse the YAML config keys for 1-moment microphysics process options and
-return a `NamedTuple` of keyword arguments for `CMP.Microphysics1MParams`.
+return a `NamedTuple` of keyword arguments for
+[`NonEquilibriumMicrophysics1M`](@ref).
 
-Each YAML key maps to one field of `get_microphysics_1m_options`, selecting the
-process option type that controls dispatch inside `bulk_microphysics_tendencies`.
-Setting a YAML value to `~` (null) disables the process (`nothing`).
+Each YAML key maps to one process, selecting the option type that controls
+dispatch inside `bulk_microphysics_tendencies`. Setting a YAML value to `~`
+(null) disables the process (`nothing`).
 """
 function get_microphysics_1m_options(parsed_args)
     CMP = CM.Parameters
--- a/src/config/model_getters.jl
+++ b/src/config/model_getters.jl
@@ -12,7 +12,7 @@ Accepted values:
   - `"dry"`: `DryModel`, no water in the model.
   - `"0M"`: `EquilibriumMicrophysics0M`, instantaneous removal of supersaturation.
   - `"1M"`: `NonEquilibriumMicrophysics1M`, built from the config keys parsed by
-    [`get_microphysics_1m_options`](@ref).
+    `get_microphysics_1m_options`.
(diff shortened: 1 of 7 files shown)


src/config/model_getters.jl
```diff
@@ -29,7 +29,11 @@ function get_microphysics_model(parsed_args, params = nothing)
     elseif model_name == "1M"
         n_substeps = parsed_args["microphysics_n_substeps"]
         n_substeps_quad = parsed_args["microphysics_n_substeps_quadrature"]
-        NonEquilibriumMicrophysics1M(; n_substeps, n_substeps_quad)
+        NonEquilibriumMicrophysics1M(;
+            n_substeps,
+            n_substeps_quad,
+            get_microphysics_1m_options(parsed_args)...,
```
ph-kev: Why are `n_substeps` and `n_substeps_quad` special enough that they are not parsed by `get_microphysics_1m_options`?

src/types.jl
```diff
@@ -59,24 +64,45 @@ The substep counts are handed to CloudMicrophysics' averaged tendency
 evaluation, which subdivides the dynamics step into substeps and returns the
 time-averaged rates; this keeps the explicit sources stable at large `dt`.
 
+`process_options` selects the variant used for each individual process; they are
+collected into a `CMP.Microphysics1MOptions`, which documents the full list of
+processes and their available variants. Passing `nothing` for a process disables
+it.
+
```
ph-kev, line 71: What is the default is someone doesn't pass anything for `process_options`?

src/simulation/AtmosSimulations.jl
```diff
@@ -306,7 +309,10 @@ simulation = CA.AtmosSimulation{Float64}(;
 """
 function AtmosSimulation{FT}(;
     model = AtmosModel(),
-    params::Parameters.ClimaAtmosParameters = ClimaAtmosParameters(FT),
+    params::Parameters.ClimaAtmosParameters = ClimaAtmosParameters(
+        FT;
+        microphysics_model = model.microphysics_model,
+    ),
```
ph-kev, line 314:
```suggestion
    params::Parameters.ClimaAtmosParameters = ClimaAtmosParameters(
        FT;
        model.microphysics_model,
    ),
```

You can write it like this, but I don't have a preference for either.

src/config/model_getters.jl
trontrytel: Good point. I think they could be done in the same way.

trontrytel:
I don't have any more comments than what @ph-kev wrote. Thank you for reviewing it!

trontrytel (approved):
It would be great to clarify the default behavior before merging

src/config/model_getters.jl
oalcabes: +1

src/types.jl
oalcabes: It would be great for the defaults to be the same as those set when running via a config file

nefrathenrici: Config-based runs use the defaults in the config file. I added documentation for cloudmicrophysics defaults that the script-based runs use.

src/config/model_getters.jl
nefrathenrici: They're not, I will fold them in.

oalcabes (approved):

oalcabes:
Thank you so much for this! And +1 to ironing out defaults, otherwise looks great to me and I am so excited to use it

src/types.jl
nefrathenrici: @oalcabes I made the defaults for the programmatic constructor match the config defaults.

There are three differences between the CM and default config values:
- n_substeps is 1 in CM, 3 in default_config.yml
- n_substeps_quad is 1 in CM, 2 in default_config.yml
- cloud_ice_formation is ConstantTimescale in CM, TemperatureDependent in default_config.yml

PR #4714, opened 2026-07-20
Update tests after ClimaCore changes

dennisYatunin:
This PR updates our unit tests following CliMA/ClimaCore.jl#2522. Every `DataLayout` in ClimaAtmos is now a `VIJFH`, so several parent array manipulations needed to be changed.

----
- [x] I have read and checked the items on the review checklist.

PR #4752, opened 2026-08-07
Patch release v0.24.4

imreddyTeja:
## Purpose 
 Patch release. Uses latest CM

This also preemptively bumps the upper ClimaCore compat. The release is expected today and will be breaking, but all of ClimaAtmos source code will still function. Only some of the tests will fail with the new release. 

PR #4714 will update the Manifest after the ClimaCore release


ph-kev (approved):

nefrathenrici (approved):
Thanks! I think you need to bump the ref counter.

PR #4754, opened 2026-08-08
Gravity wave bug fixes

ray-chew:
## Purpose 

- Remove the `linear` case in orographic gravity wave.
- Fix the non-orographic gravity wave break criteria, which started one level below the launch level and dropped the flux of any wave that broke there.
- Fix the redistribution of flux that escapes the model top, which redistributed it by level count instead of by level mass, so the momentum redistributed did not match the momentum that escaped.
- Add unit test for AD99's momentum budget.

## Content

A column should deposit exactly the momentum the scheme launches. It did not. Two independent leaks were responsible.

**1. Off-by-one at the launch level.** The break/reflection test looped from `source_level_eff - 1` upward, but both the deposition (`fm += B0[n]`) and the flux-to-acceleration conversion were triggered by `level >= source_level_eff`.
So in that one extra level, a wave could be removed from the spectrum while its momentum flux went nowhere. The loop now starts at `source_level_eff` everywhere, so a wave that is already unstable at its own launch level
deposits into the first layer rather than being discarded.

**2. Escaped flux was redistributed by count, not by mass.** Flux that reached the model top without breaking was force-broken there, and `gw_deposit` redistributed it as `wave_forcing_top / (height + 2 - damp_level)`, i.e. the top
acceleration divided by one more than the number of receiving levels. However, the average receiving level's mass differs substantially from the ghost top, so the momentum deposited did not equal the momentum that escaped.

**Verification.** New unit test `test_ad99_momentum_budget.jl` checks the column budget against a closed form, `F_launch = source_ampl · Σ B0 / Σ |B0|`, which is independent of the density profile, the wind profile, the grid, and the number of phase-speed bins. Three cases isolate the three places the equality can fail: the interior conversion and half-level-to-cell remap, the escaped-flux redistribution, and deletion at the source level.

Note that the one remaining place momentum is not conserved is the ±3e-3 m s⁻² clip applied per component when the forcing is added to the momentum tendency. Where that clip is active the deposited momentum is smaller than the launched momentum by construction.

----

### DIFF
--- a/src/parameterized_tendencies/gravity_wave_drag/non_orographic_gravity_wave.jl
+++ b/src/parameterized_tendencies/gravity_wave_drag/non_orographic_gravity_wave.jl
@@ -1057,6 +1057,21 @@ function non_orographic_gravity_wave_forcing(
     # for up to 256 gravity wave break data.
     level_end = Spaces.nlevels(axes(ᶜρ))
 
+    # Per-column rescaling for the escaped-flux redistribution.
+    gw_deposit_scale = p.scratch.temp_field_level_2
+    ᶜmasked_ρ = @. lazy(ifelse(ᶜlevel >= damp_level, ᶜρ, FT(0)))
+    Operators.column_integral_definite!(gw_deposit_scale, ᶜmasked_ρ)
+    Fields.field_values(gw_deposit_scale) .=
+        Fields.field_values(ρ_endlevel) .*
+        sqrt.(
+            Fields.field_values(ρ_endlevel) ./
+            Fields.field_values(ρ_endlevel_m1),
+        ) .*
+        (
+            Fields.field_values(z_endlevel) .-
+            Fields.field_values(z_endlevel_m1)
+        ) ./ Fields.field_values(gw_deposit_scale)
+
     # Collect the per-column fields each source mode needs into a broadcasted
     # tuple. Both modes share the same 7 fields in slots 1–7; slots 8–16 hold the
     # 9 fields specific to that mode (AD99 source vs. Beres convective source).
@@ -1189,7 +1204,7 @@ function non_orographic_gravity_wave_forcing(
             u_waveforcing, v_waveforcing,
             u_waveforcing_top, v_waveforcing_top,
             uforcing, vforcing,
-            damp_level, ᶜlevel, level_end, gw_avg_scratch,
+            damp_level, ᶜlevel, gw_deposit_scale, gw_avg_scratch,
         )
 
         # --- Beres convective source (when configured) ---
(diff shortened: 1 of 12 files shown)


src/parameterized_tendencies/gravity_wave_drag/non_orographic_gravity_wave.jl
```diff
@@ -1057,6 +1057,21 @@ function non_orographic_gravity_wave_forcing(
     # for up to 256 gravity wave break data.
     level_end = Spaces.nlevels(axes(ᶜρ))
```
akshaysridhar, line 1058: How are `ρ_endlevel` and `level_end` (which seems to look at `nlevels` in the `ᶜρ` field) different?  Based on L1192 (old)/ L1207 (new) diff below it seems that we could remove references to `level_end`

src/parameterized_tendencies/gravity_wave_drag/orographic_gravity_wave.jl
akshaysridhar, line 50:
```suggestion
```

test/parameterized_tendencies/gravity_wave/non_orographic_gravity_wave/test_ad99_momentum_budget.jl
```diff
@@ -0,0 +1,598 @@
...
+    B0 = spectrum(u_source, Bw)
+    s_signed = sum(B0)
+    s_abs = sum(abs, B0)
+    return (
+        F_launch = GW.Bt_0 * s_signed / s_abs,
+        asymmetry = abs(s_signed) / s_abs,
+    )
+end
+
+# Signed momentum flux (Pa) the OLD kernel removed at `source_level - 1` without
```
akshaysridhar, line 270: References to `OLD` kernels may become confusing here without pointers to specific commits - could you please rephrase this comment ?

src/parameterized_tendencies/gravity_wave_drag/non_orographic_gravity_wave.jl
akshaysridhar, line 1397: This needs to be `source_level` following the changes in this PR right ?

test/parameterized_tendencies/gravity_wave/non_orographic_gravity_wave/test_ad99_momentum_budget.jl
ray-chew: Opps, that is fixed now, thanks.

src/parameterized_tendencies/gravity_wave_drag/non_orographic_gravity_wave.jl
ray-chew: `level_end` contains the index of the topmost level, and not the values like in `ρ_endlevel`.

akshaysridhar (approved):

PR #4762, opened 2026-08-12
add more terms to entrainment

szy21:
## Purpose 

Change the InvZEntrainment to be c1 * w + c2 * w/z + c3 * b/w + c4. The first two terms are considered w-dependent in the implicit solver. The InvZEntrainment is not an accurate name anymore; we can rename it after finalizing the formula. Also remove some entrainment and detrainment options that are no longer correct.

## Content

----


sajjadazimi (approved):
Thanks! Looks good to me.

PR #4763, opened 2026-08-13
Add downstream test hook to buildkite

imreddyTeja:
## Purpose 

When upstream packages have a gha downstream test for ClimaAtmos, they tend to be very flaky. Additionally, it is not easy to run downstream tests on a gpu. This allows upstream packages to run climaatmos buildkite ci using their branch of the upstream package.

## Content

Modify the pipeline to use a `UPSTREAM_PACKAGES` ENV var. If that var is set, it will drop the compats for the upstream pkg being tested, and then add the requested version/commit/branch. This should let upstream package use atmos buildkite ci as a downstream test.

This was mostly written by claude.

----

### DIFF
--- a/.buildkite/full_pipeline.yml
+++ b/.buildkite/full_pipeline.yml
@@ -9,7 +9,8 @@ env:
   SLURM_KILL_BAD_EXIT: 1
   JULIA_NVTX_CALLBACKS: gc
   JULIA_MAX_NUM_PRECOMPILE_FILES: 100
-  JULIA_DEPOT_PATH: "${BUILDKITE_BUILD_PATH}/${BUILDKITE_PIPELINE_SLUG}/${BUILDKITE_BUILD_NUMBER}/depot:${BUILDKITE_BUILD_PATH}/${BUILDKITE_PIPELINE_SLUG}/depot/shared"
+  # Shared layer hard-coded to climaatmos-ci: see the note in pipeline.yml.
+  JULIA_DEPOT_PATH: "${BUILDKITE_BUILD_PATH}/${BUILDKITE_PIPELINE_SLUG}/${BUILDKITE_BUILD_NUMBER}/depot:${BUILDKITE_BUILD_PATH}/climaatmos-ci/depot/shared"
   COMMON_CONFIG_PATH: "config/common_configs"
   CONFIG_PATH: "config/model_configs"
   GPU_CONFIG_PATH: "config/gpu_configs"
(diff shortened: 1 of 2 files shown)


.buildkite/full_pipeline.yml
```diff
@@ -29,6 +29,50 @@ steps:
       - echo $$JULIA_DEPOT_PATH
       - echo "--- Instantiate .buildkite"
       - julia --project=.buildkite -e 'using Pkg; Pkg.instantiate(;verbose=true, update_registry=false); Pkg.precompile(;strict=true); Pkg.status()'
+      # Downstream test hook: upstream packages (e.g. ClimaCore) trigger this
+      # pipeline with UPSTREAM_PACKAGES set to test themselves against
+      # ClimaAtmos CI. Space-separated entries, one per package
+      # (same convention as ClimaCoupler's .buildkite/nightly/pipeline.yml):
```
ph-kev: Can you include a permalink to ClimaCoupler's .buildkite/nightly/pipeline.yml if this ever change?

.buildkite/full_pipeline.yml
```diff
@@ -29,6 +29,50 @@ steps:
...
+      #     Package        -> use the main branch
+      #     Package@1.2.3  -> use a registered version
+      #     Package#rev    -> use a branch or commit
+      # The tested packages are dropped from [compat] in both the base and the
+      # .buildkite projects, so unreleased upstream version bumps still
+      # resolve. (Compat bounds baked into registered dependencies still
```
ph-kev: Can you give a concrete example of this? The only one I can think of is for a breaking release of ClimaCore.

.buildkite/pipeline.yml
```diff
@@ -8,6 +8,13 @@
 #         (merge-queue fast path, see .buildkite/should_skip_merge_queue.sh), or
 #       * uploads the full CI pipeline (.buildkite/full_pipeline.yml).
 #
+# Downstream builds: upstream packages (e.g. ClimaCore) trigger this pipeline
+# with UPSTREAM_PACKAGES set to test their branch against ClimaAtmos CI (see
+# the "Downstream test hook" in full_pipeline.yml for the entry format). Such
+# builds run on `main` but must behave like PR builds (run the full pipeline)
+# and must not touch the reproducibility reference store; hence the
+# UPSTREAM_PACKAGES guards on the conditions below and in full_pipeline.yml.
```
ph-kev: Is it possible to set the environment variable in the buildkite web API and start a pipeline like that?

ph-kev (approved):

.buildkite/full_pipeline.yml
imreddyTeja: Yeah that is the main one, but I think it still applies for stuff like ClimaParams and Thermodynamics jl. I'll add an example

.buildkite/pipeline.yml
imreddyTeja: Yes. I will test this functionality and add instructions

ph-kev (approved):

PR #4765, opened 2026-08-13
Next round of docs updates: Add and edit atmosphere core model docs

tapios:
Documents the dycore, following Yatunin et al. (2026).

New Explanation pages:
- **Thermodynamics** — working fluid, equation of state,
  the calorically perfect assumption, reference-temperature invariance.
- **Discretization and operators** — one operator glossary for the whole suite,
  plus why the scheme is hybrid and what each strong/weak/split form conserves.
- **Conservation properties** — what is conserved, the four ingredients that make
  it work, what is deliberately given up, and how to measure it in a run.
- **Hyperdiffusion** — suspended-species water partitioning, enthalpy-consistent flux.
- **Model top and sponge layer** — both sponges, with verified defaults.

The operator glossary was duplicated between `equations.md` and `edmf_equations.md`
in two different notations, and neither copy was complete. Both now point at the
single definition. Also added the coordinate-independent equation set to `equations.md`, 
changed `ρe` → `ρe_tot`, and added an equation-to-source map.

### DIFF
--- a/docs/src/equations.md
+++ b/docs/src/equations.md
@@ -1,364 +1,380 @@
-# Equations
+# Governing Equations
 
-!!! note
+ClimaAtmos solves the fully compressible, nonhydrostatic equations of motion for
+a deep atmosphere [Yatunin2026](@cite). This page states those equations in
+continuous form. Their discretization is treated separately, in
+[Discretization and Operators](discretization.md), and the thermodynamic
+quantities they contain are defined in
+[Thermodynamics and the Working Fluid](thermodynamics.md).
 
-    This follows what is currently implemented in `src/prognostic_equations/`: it should be kept up-to-date as code is modified. If you think something _should_ be changed (but hasn't been), please open an issue.
+The equations appear first in a form independent of any coordinate system, which
+is what lets the same equation set serve Cartesian geometries for large-eddy and
+cloud-resolving simulation and the sphere for global weather and climate
+simulation.
 
-This describes the ClimaAtmos model equations and its discretizations. Where possible, we use a coordinate invariant form: the ClimaCore operators generally handle the conversions between bases internally.
+That coordinate-independent form is also the one ClimaAtmos implements. The
+tendencies in `src/prognostic_equations/` are written as the equations are
+written here — divergences, gradients, and curls of vectors — with no metric
+terms spelled out. Vectors carry their basis in their type, so ClimaCore
+converts to the generalized-coordinate components and applies the metric terms
+when it evaluates an operator. A later section gives that generalized-coordinate
+form, which is what ClimaCore works in, and which a reader comparing the code
+against the metric expressions will need.
 
 ## Prognostic variables
 
-  - ``\rho``: _density_ in kg/m³. This is discretized at cell centers.
-  - ``\boldsymbol{u}`` _velocity_, a vector in m/s. This is discretized via ``\boldsymbol{u} = \boldsymbol{u}_h + \boldsymbol{u}_v`` where
-      + ``\boldsymbol{u}_h = u_1 \boldsymbol{e}^1 + u_2 \boldsymbol{e}^2`` is the projection onto horizontal covariant components (covariance here means with respect to the reference element), stored at cell centers.
-      + ``\boldsymbol{u}_v = u_3 \boldsymbol{e}^3`` is the projection onto the vertical covariant components, stored at cell faces.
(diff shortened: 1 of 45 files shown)


docs/src/sponge.md
```diff
@@ -0,0 +1,121 @@
...
+    [Yatunin2026](@cite) specifies the default as an offset below the top
+    instead.
+
+    The practical consequence is that ``z_d`` does not follow the domain, so it
+    must be set for any domain of a different depth:
+
+      - With a top **below** ``z_d`` — a 3 km large-eddy-simulation box, say —
+        nothing exceeds the onset height, and the sponge is silently inactive.
+      - With a much **deeper** domain, a 60 km top for instance, the layer spans
+        45 km rather than the intended 15 km.
```
szy21: I don't think this is necessary. I think it's sufficient to note that users should specify z_d accordingly when using different domain heights.

docs/src/equations.md
```diff
@@ -362,3 +462,27 @@ is treated implicitly (as for total energy), with the upwind-central difference
...
+|:---------------------------------------- |:-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| Mass continuity                          | [advection.jl](https://github.com/CliMA/ClimaAtmos.jl/blob/main/src/prognostic_equations/advection.jl)                                                                                                                                                   |
+| Momentum, horizontal                     | [advection.jl](https://github.com/CliMA/ClimaAtmos.jl/blob/main/src/prognostic_equations/advection.jl)                                                                                                                                                   |
+| Momentum, vertical and pressure gradient | [pressure_work.jl](https://github.com/CliMA/ClimaAtmos.jl/blob/main/src/prognostic_equations/pressure_work.jl), [implicit/implicit_tendency.jl](https://github.com/CliMA/ClimaAtmos.jl/blob/main/src/prognostic_equations/implicit/implicit_tendency.jl) |
```
szy21:
```suggestion
| Momentum, vertical and pressure gradient | [advection.jl](https://github.com/CliMA/ClimaAtmos.jl/blob/main/src/prognostic_equations/advection.jl), [implicit/implicit_tendency.jl](https://github.com/CliMA/ClimaAtmos.jl/blob/main/src/prognostic_equations/implicit/implicit_tendency.jl) |
```
(pressure_work.jl is the EDMF pressure work term, which does nothing now)

docs/src/equations.md
```diff
@@ -1,62 +1,162 @@
...
+
+Total energy:
+
+```math
+\frac{\partial}{\partial t} (\rho e_{tot})
+  + \nabla \cdot \left[ \rho (h_{tot} \boldsymbol{u} - W_h \hat{\boldsymbol{k}}) \right]
+  = -\nabla \cdot \left[ \rho \left( \boldsymbol{\mathcal{F}}_R
+    + \boldsymbol{\mathcal{F}}_h + \boldsymbol{\mathcal{H}}_h
+    + \boldsymbol{u} \cdot (\boldsymbol{\mathcal{T}}
+      + \boldsymbol{\mathcal{H}}_u) \right) \right]
```
szy21, line 103: we still don't have these two terms in the code I think

docs/src/hyperdiffusion.md
```diff
@@ -0,0 +1,171 @@
...
+```math
+\boldsymbol{\mathcal{H}}_\psi = \nu_\psi \nabla_h
+  \left[ \nabla_h^2 (\psi - \psi_r) \right],
+```
+
+with a hyperdiffusion coefficient ``\nu_\psi`` and the horizontal del operator
+``\nabla_h`` taken along terrain-following coordinate surfaces.
+
+A vertically varying reference state ``\psi_r`` is subtracted from dry static
+energy and total specific humidity. This reduces the spurious vertical mixing
```
szy21, line 27: I don't think we have it in the code now.

szy21 (approved):

docs/src/sponge.md
tapios: Removed

docs/src/equations.md
tapios: You're right. APplied suggestion.

docs/src/hyperdiffusion.md
tapios: kept with a TODO

docs/src/equations.md
tapios: kept with a TODO note

PR #4769, opened 2026-08-14
feat: use foreach_point in `set_cloud_fraction!`

Mikolaj-A-Kowalski:
## Purpose 

Apply the `foreach_point` construct to evaluate the `set_cloud_fraction!` kernel. 

Change the way it is evaluated to hold more  data in the registers

## To-do

- [ ] Investigate evaluating `ᶜT′T′ and  ᶜq′q′` inside the kernel
- [ ] Pass the `Field`s to the contruct (following https://github.com/CliMA/ClimaCore.jl/pull/2571)

## Content

The `set_cloud_fraction__NVTX` kernel

| State | Register Usage  |   Mean Time           |
|-------|-----------------|-----------------------|
| New   |        121      |      628.586 μs       |
| Old   |        48       |      937.293 μs       |

Just by applying the construct we see good ~30% speedup in the execution time of the kernel. As expected, the register usage increases.  This was measured on the [current `main`](https://github.com/CliMA/ClimaCore.jl/commit/a84e74d5529ca9c0d297cc8309c113257b96ec91) of ClimaCore. 

Currently still DataLayouts are passed in since the lazy broadcasted expression  (ᶜq_lcl and ᶜq_icl) fails (in dispatch chain, the `each_slice_index` wants to take `size` of it which does not work correctly)  

----
- [x]  I have read and checked the items on the review checklist.


Mikolaj-A-Kowalski:
Hera are the NSIght compute reports [reports.tar.gz](https://github.com/user-attachments/files/31077223/reports.tar.gz). 

There is one odd thing going on which is that the NSight is reporting local memory usage (basically 1st argument gets copied to the stack in the kernel) but the quick profiling utility we have reports no local_memory usage... 

It is possible that we do something wrong in the post-processing of the in-build profiler data.

Mikolaj-A-Kowalski:
I have rebased on the current main. 

In order pass the Fields directly I also had to remove some lazy annotations and evaluate arguments more eagerly. 

It does go against the goal of kernel fusion a bit but the net effect is positive overall with the total runtime of the split kernels of around ~540 μs.

Ideally we would like to put these new kernels inside the `foreach_point` as well, but I don't think it is possible without a major refactoring since we cannot pass the `FieldVector`s (i.e. `Y` and `p.precomupted`) into the construct.

Mikolaj-A-Kowalski:
There is a failure in the CI https://github.com/CliMA/ClimaAtmos.jl/actions/runs/32856706702/job/97830619518?pr=4769#step:7:702

It seems to be caused by some (cached?) older version of ClimaCore (0.15) in the environment [see here](https://github.com/CliMA/ClimaAtmos.jl/actions/runs/32856706702/job/97830619518?pr=4769#step:7:80). Is there something I need to update on my side and forgot?

Mikolaj-A-Kowalski:
[split.tar.gz](https://github.com/user-attachments/files/31424831/split.tar.gz)
The NSight Compute report for the version with `Fields` and no `lazy` annotations.

Mikolaj-A-Kowalski:
@dennisYatunin @imreddyTeja 
With the compatibility fix the downgrade now works. What is failing is the documentation PR, but I have noticed that other PRs seem to be also affected so I think it is a wider problem? 

Since this one is a small change and id does lead to some small performance improvement, I was hoping to drive towards getting it merged. Hence I have tagged you for review.

Mikolaj-A-Kowalski:
Hi. I do have a trouble to figure out what is the problem withe the `climaatmos-ci` job. When I read the log it feels like a CI problem, but I am not sure if I haven't accidentality done something to break it.

imreddyTeja:
> Hi. I do have a trouble to figure out what is the problem withe the `climaatmos-ci` job. When I read the log it feels like a CI problem, but I am not sure if I haven't accidentality done something to break it.

This does look like a CI issue, because other PRs ci is failing with the same error. I will clear the shared depot/

src/parameterized_tendencies/microphysics/cloud_fraction.jl
```diff
@@ -959,8 +959,8 @@ microphysics, uses the prognostic cloud condensate only; the precomputed
 not count as cloud.
 """
 _grid_mean_cloud_condensate(Y, p, ::NonEquilibriumMicrophysics) = (
-    (@. lazy(max(0, specific(Y.c.ρq_lcl, Y.c.ρ)))),
-    (@. lazy(max(0, specific(Y.c.ρq_icl, Y.c.ρ)))),
+    (@. (max(0, specific(Y.c.ρq_lcl, Y.c.ρ)))),
+    (@. (max(0, specific(Y.c.ρq_icl, Y.c.ρ)))),
```
dennisYatunin, line 963: To avoid heap allocations here, use a similar `foreach_point` implementation in the functions that call it. Inline `_grid_mean_cloud_condensate` (and also `_env_cloud_condensate`) into the calling functions, passing a closure into `foreach_point`, rather than a lazy `Broadcasted` expression that consumes registers or a `Field` that requires heap allocations.

Mikolaj-A-Kowalski: I have to admit that I am not sure I know how to do it. The problem is that under some circumstances these functions require access to `p.precomputed` ( the generic fallback for `_grid_mean_could_condensate`) and I fear that we are not able to pass the `p.precomputed` inside the `foreach_point`.

The only option is basically to change the interface to pass all the field that may need in their argument list. But I was reluctant do do it sine while it works for `_get_condensate_means`, for e.g. `_get_env_ρ_T_q` the list of variables that may be need is quite long and I fear it might hurt readability and maintenance.

Mikolaj-A-Kowalski: I have tried inlining on a separate branch https://github.com/CliMA/ClimaAtmos.jl/tree/iccs/cloud-fraction-perf-inlined. 

I have compared the performance in the reports below
[reports.tar.gz](https://github.com/user-attachments/files/32340671/reports.tar.gz)

The version with inlining has increased in register usage (67 -> 117) and overall got a bit slower (total from ~500 us to ~600 us), probably as the result of decrease in occupancy (~40% -> ~25%)

dennisYatunin: Right, I see what you mean about inlining hurting readability and maintenance. I suspect we'll eventually need to change parts of the ClimaAtmos interface to avoid this register pressure issue, but here's a potential way for you to avoid the verbosity of your inlining branch:
```julia
        DataLayouts.foreach_point(
            Y.c,
            p.precomputed.ᶜq_liq,
            p.precomputed.ᶜq_ice,
            p.precomputed.ᶜq_liq⁰,
            p.precomputed.ᶜq_ice⁰,
            ᶜcloud_fraction,
            ᶜT_mean,
            ᶜρ_env,
            ᶜq_mean,
            ᶜT′T′,
            ᶜq′q′,
        ) do Yc, ᶜq_liq, ᶜq_ice, ᶜq_liq⁰, ᶜq_ice⁰, ᶜcloud_fraction, ᶜT_mean, ᶜρ_env, ᶜq_mean, ᶜT′T′, ᶜq′q′
            Y_fake = (; c = Yc)
            p_fake = (; atmos = (; turbconv_model), precomputed = (; ᶜq_liq, ᶜq_ice, ᶜq_liq⁰, ᶜq_ice⁰))
            (ᶜq_lcl, ᶜq_icl) = _get_condensate_means(Y_fake, p_fake, turbconv_model, microphysics_model)
            @. ᶜcloud_fraction = _compute_cloud_fraction(...)
        end
```
As long as we're not specializing on the types of `Y` and `p` inside `_get_condensate_means`, you can construct nested `NamedTuple` containers with equivalent field names, and populate them with only the point views required for `_get_condensate_means`.

Still, this version will also use more registers, since it passes 5 extra fields to `foreach_point` (`ᶜY`, `ᶜq_liq`, `ᶜq_ice`, `ᶜq_liq⁰`, and `ᶜq_ice⁰`). I'm surprised by the register counts you saw in the other branch, since 50 registers for 5 fields seems like a bit much, but maybe the implied 10 registers/field is something we can only fix from ClimaCore. If so, my recommendation is to keep the implementation you have here, with `ᶜq_lcl` and `ᶜq_icl` computed outside the `foreach_point` loop.

If you need `ᶜq_lcl` and `ᶜq_icl` outside the loop, you can also avoid extra heap allocations as follows:
```julia
ᶜq_lcl_lazy, ᶜq_icl_lazy = _get_condensate_means(Y, p, turbconv_model, microphysics_model)
ᶜq_lcl = (p.scratch.ᶜtemp_scalar .= ᶜq_lcl_lazy)
ᶜq_icl = (p.scratch.ᶜtemp_scalar_2 .= ᶜq_icl_lazy)
```
This way, `_get_condensate_means` can still return a pair of lazy expressions, but they'll be materialized into pre-allocated scratch fields before being passed to `foreach_point`. This will probably perform even better than the current version, since heap allocating `CuArray`s on every step is relatively slow.

Mikolaj-A-Kowalski: Thank you for the trick with the NamedTuples! It does makes things much more cleaner. 

I tried it out and as you suspected it basically was equivalent to the inline version in terms of runtime (~600 us). 

I have used the scratch fields to avoid the allocation now. In NCU it shaved of extra 50us from the total runtime. I have pushed the change to this branch.

src/parameterized_tendencies/microphysics/cloud_fraction.jl
```diff
@@ -987,24 +990,38 @@ NVTX.@annotate function set_cloud_fraction!(
...
-        ᶜq′q′,
-        corr_Tq,
-        FT(α),
-        $(floor),
-    )
+    ᶜcloud_fraction = p.precomputed.ᶜcloud_fraction
+    α = FT(α)
 
+    let α = α, thermo_params = thermo_params, corr_Tq = corr_Tq, floor = floor,
+        sgs_quad = sgs_quad
```
dennisYatunin, line 997: Is this `let`-block necessary? There's a similar pattern that confused me [here](https://github.com/CliMA/ClimaAtmos.jl/pull/4788#discussion_r4042223974).

src/parameterized_tendencies/microphysics/cloud_fraction.jl
```diff
@@ -1050,15 +1067,13 @@ function _get_env_ρ_T_q(Y, p, thermo_params, turbconv_model)
...
-                ᶜq_ice⁰,
-            ),
+        ᶜρ_env = @. TD.air_density(
+            thermo_params,
+            ᶜT⁰,
+            ᶜp,
+            ᶜq_tot_nonneg⁰,
+            ᶜq_liq⁰,
+            ᶜq_ice⁰,
         )
```
dennisYatunin, line 1077: This should also be either a lazy field or a scratch field, rather than a heap-allocated field.

dennisYatunin (approved):

src/parameterized_tendencies/microphysics/cloud_fraction.jl
Mikolaj-A-Kowalski: It is necessary for α, not strictly for the other variables but I put it there for safety. 

The problem is that the Julia will box a variable (under some circumstances e.g. here) if it is captured by a closure and reassigned even outside the closure scope.  In this case it is the `α = FT(α)` assignment. I have wrapped the `foreach_point` basically to force a capture by value (as far as I am away this is the only way to do it in julia).

PR #4774, opened 2026-08-15
Viscous sponge: energy split

sajjadazimi:
## Purpose 
Diffuse ρe_tot as a dry-static-energy + h_eff-weighted q_tot_eff split. Diffuse ρq_tot on q_tot_eff (rain/snow excluded) and distribute the aggregate tendency to suspended cloud mass species. Bundle the sponge accumulation into viscous_sponge_tendency!(Yₜ, Y, p) to keep additional_tendency! readable.

----

### DIFF
--- a/reproducibility_tests/ref_counter.jl
+++ b/reproducibility_tests/ref_counter.jl
@@ -1,4 +1,4 @@
-402
+403
 
 # **README**
 #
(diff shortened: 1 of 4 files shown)


szy21:
I think Tapio wants it to be the same as hyperdiffusion and vertical diffusion: no viscous sponge on precipitation, tendencies for q_lcl and q_icl proportional to qt tendency. But it's not very important so we don't have to change it now if it's too complicated.

szy21 (approved):

PR #4777, opened 2026-08-20
Fixes microphysics/cloud fraction bug: erroneous subtraction of precip i

tapios:
Precipitation (rain/snow) was erroneously subtracted in reconstruction of SGS *cloud* condensate, violating condensate conservation in the quadratures. 

This PR fixes this and adds tests to make sure mid point equals grid mean.


szy21:
But q_tot_hat has precipitation in it?

tapios:
> But q_tot_hat has precipitation in it?

Yes, precip is subtracted from it later, and the evaluator only uses the centered fluctuation S′_hat = (q_tot_hat − q_sat_hat) − μ_S. I added a code comment to clarify.

szy21 (approved):
I think this makes sense, but I don't fully understand the water filling yet, so it would be good if @sajjadazimi takes a look too.

sajjadazimi (approved):
Great catch, thanks!

szy21:
let me release atmos in this PR.

PR #4778, opened 2026-08-20
RRTMGP v0.23 compat updates

tapios:
Updates atmos to be compatible with new names in RRTMGP v0.23 (which will become required in v1).


szy21 (approved):
RRTMGP in the Manifest needs to be updated too.

PR #4779, opened 2026-08-20
Make the SGS quadrature closure more self-consistent:

tapios:
- Fit `λ_lagrange` to the discrete quadrature constraint `Σᵢ wᵢ·max(0, λ + α·S′ᵢ) = q_c`  instead of the continuous Gaussian; the condensate mass constraint was previously violated by −15% to +160% (worst for thin clouds, q_c ≲ 0.1·σ_S); now it holds to machine precision.
- Add a second Newton step to `_compute_z`, removing a one-sided thin-cloud CF overestimate.
- 0M: return the flux-weighted energy helper E[dq_tot·e]/E[dq], so the precipitation energy sink is the true SGS average that takes covariances between `dq_tot/dt` and `e_tot` (was off by up to +56% in mixed regions, leading to warm bias near and above the melting layer).
- Adds regression tests across regimes, quadrature orders, and precisions.

### DIFF
--- a/src/parameterized_tendencies/microphysics/microphysics_wrappers.jl
+++ b/src/parameterized_tendencies/microphysics/microphysics_wrappers.jl
@@ -91,9 +91,10 @@ Diagnoses the local condensate by saturation adjustment, then calls
 
 # Returns
 
-NamedTuple with `dq_tot_dt` [kg/kg/s] and `e_tot_hlpr` [J/kg]. Both are returned
-from a single call so that `integrate_over_sgs` SGS-averages the energy helper
-consistently with the water sink that carries it.
+NamedTuple with `dq_tot_dt` [kg/kg/s] and the energy-flux product
+`dq_e = dq_tot_dt · e_tot_hlpr` [W/kg]. The product is formed per point so
+that its SGS average is the true energy sink `E[dq·e]` (averaging `dq` and
+`e` separately and multiplying the means would drop their covariance).
 """
 @inline function (eval::Microphysics0MEvaluator)(T_hat, q_hat)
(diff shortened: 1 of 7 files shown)


sajjadazimi (approved):
Thanks! Looks like a necessary fix.

src/parameterized_tendencies/microphysics/microphysics_wrappers.jl
```diff
@@ -150,20 +151,36 @@ via `apply_0m_tendency_limit`.
...
     )
+    # Flux-weighted energy helper: E[dq·e] / E[dq]. The ratio is stable for
+    # any strictly negative mean sink because numerator and denominator share
+    # the dq scale. Where no quadrature point precipitates (E[dq] = 0), fall
+    # back to the mean-state helper.
+    sa = evaluator.sat_eval(T, q_tot_nonneg)
+    e_hlpr_mean = e_tot_0M_precipitation_sources_helper(
+        thp, T, sa.q_liq, sa.q_ice, Φ,
+    )
+    e_tot_hlpr = ifelse(dq_tot_dt < zero(FT), dq_e / dq_tot_dt, e_hlpr_mean)
```
szy21, line 183:
```suggestion
    e_tot_hlpr = ifelse(dq_tot_dt < zero(FT), dq_e / dq_tot_dt, 0)
```
This should work as dq_tot_dt can only be 0 or negative, and we don't need an additional sat_eval. But it doesn't matter much.

tapios: Ys, that's right. Sorry I got this too late. I'll have a separate PR for this change.

PR #4783, opened 2026-08-22
add quadratures for precipitation

szy21:
## Purpose 

wip. do not review it yet.

## Content

----

### DIFF
--- a/src/parameterized_tendencies/microphysics/cloud_fraction.jl
+++ b/src/parameterized_tendencies/microphysics/cloud_fraction.jl
@@ -733,9 +733,97 @@ uses the same quadrature points.
 end
 
 """
+    discrete_cloudy_weight_width_coeff(FT)
+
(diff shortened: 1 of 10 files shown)


src/parameterized_tendencies/microphysics/cloud_fraction.jl
```diff
@@ -826,6 +925,149 @@ NVTX.@annotate function set_sgs_moments_and_cloud_fraction!(Y, p)
...
+    #         input;
+    #         init = zero(FT),
+    #         reverse = true,
+    #     ) do a_p_above, (CF_d_level, q_precip_level)
+    #         ifelse(
+    #             q_precip_level > q_precip_min,
+    #             max(CF_d_level, f_decay * a_p_above),
+    #             zero(FT),
+    #         )
+    #     end
```
szy21, line 1029: This results in large allocation. The for loop below doesn't allocate, and should give the same result. @nefrathenrici do you have an idea why column_accumulate allocates here?

PR #4784, opened 2026-08-22
Fix random GPU CI failures: `LLVM error: No such file or directory`

nefrathenrici:
Every step of a build shares one depot, and one directory in it holds the compiled GPU runtime library that every GPU kernel compilation reads. Precompiling CUDA.jl deletes that directory. So when one step precompiles CUDA.jl while GPU jobs are running, a GPU job reading the file at that moment crashes. Rare and unreproducible because the window is tiny, but each job reads the file frequently. Two steps precompiled CUDA.jl in the middle of the build: checkbounds and the new `perf/run-and-time` environment. 

Two commits, the first one contains the minimal fix:
- Precompile the checkbounds and `perf/run-and-time` environments in the init step to prevent competing recompilation during the job steps
- Fold `perf/run-and-time` into `.buildkite`, since its only additional dep is `DataFrames`. This is not necessary to fix the issue, but simplifies the testing environments slightly.


imreddyTeja (approved):
Hopefully this fixes the issues. Thanks! 

The perf/run-and-time project was originally meant to be a minimal env for ICCS to use, but I'm fine with folding it into .buildkite.

.buildkite/full_pipeline.yml
```diff
@@ -77,6 +77,9 @@ steps:
             Pkg.status()'
         fi
 
+      - echo "--- Precompile --check-bounds=yes caches (Checkbounds step)"
+      - julia --check-bounds=yes --project=.buildkite -e 'using Pkg; Pkg.precompile(; strict=true)'
```
imreddyTeja, line 81: Does this also need to be added to `build_depot.sh`?

ph-kev (approved):

.buildkite/full_pipeline.yml
nefrathenrici: This is already in [build_depot.sh](https://github.com/CliMA/ClimaAtmos.jl/blob/main/.buildkite/build_depot.sh#L27)

Issue #4745, opened 2026-08-04
Use `foreach_point` premitive to fuse pointwise kernels


Mikolaj-A-Kowalski:
With the recent changes in ClimaCore we have access to the `foreach_point` primitive that we can use to fuse pointwise operations into larger kernels. 

We need to apply it throughout the repository.  

We should use sub-issues to track different parts of the codebase we will review for the `foreach_point` usage.

PR #4785, opened 2026-08-24
feat: use foreach_point in `set_sgs_moments_and_cloud_fraction!`

sjavis:
## Purpose 

This refactors the `set_sgs_moments_and_cloud_fraction!` function to use the `foreach_point` construct (part of #4745).

## To-do

- [x] Pass the fields to `foreach_point` directly. (Leaving this as DataLayouts because there is an issue when using the Fields)

## Content

- This uses `foreach_point` to fuse the broadcasted evaluations of sgs moments and cloud fraction into a single kernel. The runtime of the function is reduced by 20%.
- Also uses local caching for `sgs_moments` inside `foreach_point` although this has negligable impact. It's possible the compiler is doing this optimisation anyway.
- Removes repeated computation of `ᶜq_lcl + ᶜq_icl`. This reduces register usage.

Using the `sgs.jl` test case in `perf/run-and-time/`:

| Case | Mean runtime (ms) | Registers |
| --- | --- | --- |
| Baseline | 1.398 | 201 & 173 |
| `foreach_point` | 1.133 | 225 |
| + local caching | 1.129 | 225 |
| + `ᶜq_c` | 1.134 | 219 |

----


src/parameterized_tendencies/microphysics/cloud_fraction.jl
```diff
@@ -803,28 +803,61 @@ NVTX.@annotate function set_sgs_moments_and_cloud_fraction!(Y, p)
...
+    # Obtain the DataLayouts for the foreach_point loop
+    ᶜρ_env_dl = Fields.field_values(ᶜρ_env)
+    ᶜT_mean_dl = Fields.field_values(ᶜT_mean)
+    ᶜq_mean_dl = Fields.field_values(ᶜq_mean)
+    ᶜq_lcl_dl = Fields.field_values(ᶜq_lcl)
+    ᶜq_icl_dl = Fields.field_values(ᶜq_icl)
+    ᶜT′T′_dl = Fields.field_values(ᶜT′T′)
+    ᶜq′q′_dl = Fields.field_values(ᶜq′q′)
+    ᶜsgs_moments_dl = Fields.field_values(p.precomputed.ᶜsgs_moments)
+    ᶜcloud_fraction_dl = Fields.field_values(p.precomputed.ᶜcloud_fraction)
```
imreddyTeja, line 815: We should be able to use `foreach_point` on the fields now, so pulling out the datalayouts is un-neeeded

src/parameterized_tendencies/microphysics/cloud_fraction.jl
```diff
@@ -803,28 +803,61 @@ NVTX.@annotate function set_sgs_moments_and_cloud_fraction!(Y, p)
...
+
+    α = FT(α)
+    let α = α, thermo_params = thermo_params, floor = floor, sgs_quad = sgs_quad,
+        corr_Tq = corr_Tq
```
imreddyTeja, line 819: nit:
```suggestion
    let α = FT(\alpha), thermo_params = thermo_params, floor = floor, sgs_quad = sgs_quad,
        corr_Tq = corr_Tq
```

imreddyTeja (commented):
Looks good to me. Only thing is the comments about not needing to pull out the datalayouts anymore

src/parameterized_tendencies/microphysics/cloud_fraction.jl
sjavis: This should be addressed by e61ecdf. The lazy fields get materialized as it doesn't accept these.

src/parameterized_tendencies/microphysics/cloud_fraction.jl
sjavis: Also addressed by e61ecdf.

imreddyTeja (approved):
Looks good to me. Thanks!

sjavis:
This should be working now. The issue I mentioned didn't seem to be fixed by rebasing onto main, but removing the local sgs moments variable did.

Having rebased, the timings are now from 1.40 ms before to 0.737 ms after.

PR #4786, opened 2026-08-24
Tr/bools

imreddyTeja:
## Purpose 
Clean up orographic gravity wave code

## Content
- rm uses of `parent`
- ensure output of operator applied to bools is a `bool`.  This is needed for changes in climacore because they way boolean arithmetic works in Julia.

```
julia> true * true
true

julia> true + true * true
2

```

----

### DIFF
--- a/src/parameterized_tendencies/gravity_wave_drag/orographic_gravity_wave.jl
+++ b/src/parameterized_tendencies/gravity_wave_drag/orographic_gravity_wave.jl
@@ -179,7 +179,7 @@ function orographic_gravity_wave_compute_tendency!(Y, p, ::FullOrographicGravity
         end
 
         # DEBUG: Check if Y has NaNs at entry
-        if any(isnan, parent(Y.c.ρ))
+        if any(isnan, Y.c.ρ)
             @error "OGWD: Input Y.c.ρ already has NaNs at function entry!"
             error("Cannot compute OGWD tendency with NaN inputs")
         end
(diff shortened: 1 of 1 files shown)


src/parameterized_tendencies/gravity_wave_drag/orographic_gravity_wave.jl
```diff
@@ -383,8 +383,8 @@ function orographic_gravity_wave_forcing!(
     # the z-values don't change, but this is necessary for
     # calc_nonpropagating_forcing! to work on the GPU
     get_pbl_z!(topo_ᶜz_pbl, ᶜp, ᶜT, ᶜz, grav, cp_d)
-    parent(topo_ᶠz_pbl) .= parent(topo_ᶜz_pbl) .- FT(1 / 2) .* parent(Δz_bot)
     topo_ᶠz_pbl = topo_ᶠz_pbl.components.data.:1
+    topo_ᶠz_pbl .= topo_ᶜz_pbl .- FT(1 / 2) .* Δz_bot
```
ph-kev, line 387: Is there any particular reason we need to add this after `topo_ᶠz_pbl = topo_ᶠz_pbl.components.data.:1`?

ph-kev (approved):

src/parameterized_tendencies/gravity_wave_drag/orographic_gravity_wave.jl
imreddyTeja: Yes. `topo_ᶠz_pbl = topo_ᶠz_pbl.components.data.:1` makes  `topo_ᶠz_pbl` point to a scalar field instead of a vector field. If I didn't swap them, I would need to do:
```
topo_ᶠz_pbl.components.data.:1 .= topo_ᶜz_pbl .- FT(1 / 2) .* Δz_bot
```

PR #4788, opened 2026-08-25
apply foreach_point in CM cache

Mikolaj-A-Kowalski:
## Purpose 

Apply the foreach_point to the longest running kernel in the `set_microphysics_tendency_cache!` function

## To-do
 - [x] Obtain some speed-up
 

## Content

At the moment we don't really observe any notable runtime improvement: 

| State | Register Usage  |   Mean Time           |
|-------|-----------------|-----------------------|
| New   |        255     |     31.96 ms (-1.5%)      |
| Old   |        255       |      32.45ms      |

Taken from the NCU reports below

[reports.tar.gz](https://github.com/user-attachments/files/31427193/reports.tar.gz)

Looking at what the profiler is saying it feels like the main bottleneck for this kernel is the traffic to the local memory (~ 100 million memory requests to local memory vs around 800k issued to the global memory). 

In my opinion it is caused by the usage of the stack by the kernel. The compiler does not seem to be able assert that the structs we pass as arguments are not mutated, hence it feels compelled to create stack, which uses local memory and creates the heavy memory traffic. 

I don't think it is the case of simple register spillover. In that case I would expect to see similar number of stores and loads (at least it feels like it should be the case). Here we have a heavy imbalance   (72 million loads vs 31 million stores) .

----


Mikolaj-A-Kowalski:
@imreddyTeja @dennisYatunin @sjavis 

I have spend some time trying to eliminate the stack in the heaviest CM cache kernel and in the end what did the trick was to parametrise the `Microphysics1MEvaluator` on the values of `thp` and `cmp`. 

This basically allows the values to be passed to the function not via variables, but via instructions. This, in turn, allows the compiler to avoid copying the `thp` and `cmp` to stack and removes the majority of the local memory traffic. Overall when tested in the minimal reproducer it speeds up the kernel by good 30%. 

I do assume that the values of `thp` and `cmp` are constant over the course of the calculation so we should not run into recompilation penalty? 

Also to shrink the size of the closure I have used `Val` to pass the parameters via the type.

Below is the summary of the kernel runtimes from the NCU reports (and original NCU reports as well)

State | Kernel Time  (ncu) |
--------|-------------------------------|
Old   |               32.45  ms                      |
foreach_point |       31.96 ms (-1.51%)               |
foreach_point + value |   21.91 (-32.48%)      | 

[new-reports.tar.gz](https://github.com/user-attachments/files/31841052/new-reports.tar.gz)


I would like now to clean up this PR and move it out of the draft status. For that I have two questions: 

- I have the utility function `unwrap_value` to get the object from the Value type. What is the best place to put it? 
- I have went with the path of least resistance and I wrap the parameters into  Value when constructing the evaluator. Would it be better to do  the wrapping inside its outer constructor? (EDIT: I have wrapped them in outer constructor to keep backwards compatibility with the tests in the end)

src/cache/microphysics_cache.jl
```diff
@@ -987,37 +989,79 @@ function set_microphysics_tendency_cache!(
...
+    (; ᶜT′T′, ᶜq′q′, ᶜsgs_moments) = p.precomputed
+
+    # Wrap all arguments in data layouts
+
+    ᶜsgs_moments_λ_lagrange = ᶜsgs_moments.λ_lagrange
+
+    corr_Tq = correlation_Tq(p.params)
+    α = sgs_variance_fidelity(CAP.cloud_fraction_steepness_scale(p.params))
+    let thp_val=Val(thp), cmp_val=Val(cmp), dt=dt, nsubs=nsubs, nsubs_quad=nsubs_quad,
+        sgs_quad=sgs_quad, corr_Tq=corr_Tq, α=α
```
dennisYatunin, line 1008: What's the purpose of this `let`-block? I've only seen something like `let dt = dt` used to avoid boxing inside a closure where `dt` is reassigned, but this isn't inside a closure. The `do`-block after `foreach_point` is a closure, but the `let`-block is outside of that.

src/parameterized_tendencies/microphysics/microphysics_wrappers.jl
```diff
@@ -253,6 +257,21 @@ struct Microphysics1MEvaluator{S, MP, TPS, FT, Args <: Tuple}
...
+    function Microphysics1MEvaluator(
+        scheme::S, mp, tps, ρ::FT,
+        q_rai::FT, q_sno::FT,
+        λ::FT, λ_lagrange::FT, mu_S::FT, α::FT,
+        dt::FT, nsubs::Int, args::Args,
+    ) where {S, FT, Args <: Tuple}
+        return new{S, unwrap_value(mp), unwrap_value(tps), FT, Args}(
+            scheme, ρ, q_rai, q_sno, λ, λ_lagrange, mu_S, α, dt, nsubs, args,
+        )
+    end
```
dennisYatunin, line 274: If you're always going to store `mp` and `tps` as type parameters, there's no benefit to passing them in without `Val` wrappers. Doing so would probably trigger dynamic dispatch, which can't be compiled on GPUs. So, instead of adding this inner constructor, you can keep the original default constructor and use `unwrap_val(eval.mp)`/`unwrap_val(eval.tps)` below.

dennisYatunin (approved):
Moving the parameters into instruction memory seems like the right call here, since they are guaranteed to be constant for every simulation. I suspect we'll need to generalize this pattern to keep fusing more kernels.

However, this implementation is still limited by inlining---if the `CloudMicrophysics.jl` functions that take in the parameters aren't inlined, LLVM might not be able to perform SROA (scalar reduction of aggregates) and infer that only a few parameters need to be stored in registers. To guarantee minimal register usage, the `Val` wrapper (or any singleton equivalent) would need to be propagated all the way through the `CloudMicrophysics.jl` source code. For sufficiently complex thermodynamics functions, the same would have to be done in `Thermodynamics.jl`. I'm happy to see that such extensive changes aren't required for this example, though; maybe we need a lot more complexity before they become necessary.

trontrytel:
@haakon-e - The debate is here

src/cache/microphysics_cache.jl
Mikolaj-A-Kowalski: Here it is not necessary per see, but I am wrapping the variables in the let block to ensure they are captured by value. In the [other case](https://github.com/CliMA/ClimaAtmos.jl/pull/4769#discussion_r4052665295) I had a problem that a variable would get boxed and captured by reference if it was reassigned in the parent scope of the `foreach_block` (i.e. outside of closure)

src/parameterized_tendencies/microphysics/microphysics_wrappers.jl
Mikolaj-A-Kowalski: I'm afraid I don't quite follow I am not 100% sure what you mean by 
> So, instead of adding this inner constructor, you can keep the original default constructor and use unwrap_val(eval.mp)/unwrap_val(eval.tps) below.

I think you are right that I didn't do thinks idiomatically, but my impression is that we are kind of protected from dynamic dispatch depending how we call this constructor.

If we wrap the parameters in Val like:
```
Microphysics1MEvaluator(scheme, Val(mp), Val(thp) ...) 
```
or `mp isa Value && thp is a Value` the constructor should not have type instability and we would avoid dynamic dispatch. 

It will only happen if we call the constructor with the values of the arguments. 

I guess the more full-proof approach would be to just replace `mp` -> `::Val{mp}` in the argument list Is this what you meant?

PR #4789, opened 2026-08-25
Pass ClimaComms context when loading local OGW drag files.

costachris:
Pass run ClimaComms context when loading local OGW drag files. Needed for multi-rank runs with local gw files.

### DIFF
--- a/src/parameterized_tendencies/gravity_wave_drag/orographic_gravity_wave_helper.jl
+++ b/src/parameterized_tendencies/gravity_wave_drag/orographic_gravity_wave_helper.jl
@@ -1061,14 +1061,17 @@ end
 """
     load_preprocessed_topography(; topography, topo_smoothing,
                                  topography_damping_factor, h_elem)
-    load_preprocessed_topography(filename::String)
+    load_preprocessed_topography(filename::String, comms_ctx = ClimaComms.SingletonCommsContext())
 
 Read a preprocessed orographic drag field from an HDF5 file in the package root.
+The optional `comms_ctx` must be the run's MPI context on distributed runs so the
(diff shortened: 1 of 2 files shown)


src/parameterized_tendencies/gravity_wave_drag/orographic_gravity_wave_helper.jl
ray-chew: "Both read on a single CPU context, so a GPU run has to move the result with `move_topo_info_to_gpu`." is no longer valid. Please update the docstring.

src/parameterized_tendencies/gravity_wave_drag/orographic_gravity_wave_helper.jl
```diff
@@ -1061,9 +1061,11 @@ end
 """
     load_preprocessed_topography(; topography, topo_smoothing,
                                  topography_damping_factor, h_elem)
-    load_preprocessed_topography(filename::String)
+    load_preprocessed_topography(filename::String, comms_ctx = ClimaComms.SingletonCommsContext())
 
 Read a preprocessed orographic drag field from an HDF5 file in the package root.
+The optional `comms_ctx` must be the run's MPI context on distributed runs so the
+field is read as correctly.
```
ray-chew: "read correctly"

ray-chew (approved):
Tested this on 2 MPI ranks (CPU) with a local `computed_drag_Earth_false_1_6.hdf5`. The run completes, and it is bit-identical to the same 2-rank run with computed drag loaded via ClimaArtifacts. However, two minor docstring require updating. Otherwise, approving this PR.

PR #4793, opened 2026-08-28
Pressure-drag tke source

sajjadazimi:
## Purpose 
Add updraft pressure drag tke tendency

----

### DIFF
--- a/reproducibility_tests/ref_counter.jl
+++ b/reproducibility_tests/ref_counter.jl
@@ -1,4 +1,4 @@
-405
+406
 
 # **README**
 #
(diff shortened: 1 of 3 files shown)


szy21 (approved):
Previously when we had this, it includes both the drag and virtual mass terms. Is the drag the only term we need here?

sajjadazimi:
> Previously when we had this, it includes both the drag and virtual mass terms. Is the drag the only term we need here?

Based on the paper draft, the virtual mass term is a reactive term that doesn't contribute to coh -> iso. But I realized my implementation of C_d is inconsistent with the advection eq. Fixing now ...

sajjadazimi:
> > Previously when we had this, it includes both the drag and virtual mass terms. Is the drag the only term we need here?
> 
> Based on the paper draft, the virtual mass term is a reactive term that doesn't contribute to coh -> iso. But I realized my implementation of C_d is inconsistent with the advection eq. Fixing now ...

Done!

PR #4794, opened 2026-08-28
(0.42.8) tag patch release + up deps

nefrathenrici:


szy21 (approved):
Thanks! Could we wait for https://github.com/CliMA/ClimaAtmos.jl/pull/4793? If it didn't get merged later this afternoon, let's get this in without that PR.

PR #4796, opened 2026-09-01
Use ParallelTestRunner for GHA tests

ph-kev:
This PR experiments with using [ParallelTestRunner](https://github.com/JuliaTesting/ParallelTestRunner.jl) to speed up the GHA CI.


ph-kev:
The GHA CI did not speed up. When a memory threshold is reached, the worker is restarted which mean that any JIT compiled code is gone. Even though there are more workers, a lot of the time spend is recompiling code.

PR #4797, opened 2026-09-02
Guards against unphysical extrapolated ρa in solving u_3.

sajjadazimi:
## Purpose 
Guards against unphysical extrapolated ρa (u₃ is solved before ρa).

----

### DIFF
--- a/src/prognostic_equations/implicit/initialize_implicit_problem.jl
+++ b/src/prognostic_equations/implicit/initialize_implicit_problem.jl
@@ -155,12 +155,15 @@ function solve_sgs_u₃_implicit_stage_analytic!(Y, p, dtγ)
     α_b = CAP.pressure_normalmode_buoy_coeff1(turbconv_params)
     α_d = CAP.pressure_normalmode_drag_coeff(turbconv_params)
     a_min = CAP.min_area(turbconv_params)
+    a_max = CAP.max_area(turbconv_params)
     scale_height = CAP.R_d(params) * CAP.T_surf_ref(params) / CAP.grav(params)
 
-    # Approximation factor used in w₀ - w ≈ -(ρ / ρa⁰) w (single-updraft case).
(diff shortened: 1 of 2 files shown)


src/prognostic_equations/implicit/initialize_implicit_problem.jl
```diff
@@ -155,12 +155,15 @@ function solve_sgs_u₃_implicit_stage_analytic!(Y, p, dtγ)
...
     α_d = CAP.pressure_normalmode_drag_coeff(turbconv_params)
     a_min = CAP.min_area(turbconv_params)
+    a_max = CAP.max_area(turbconv_params)
     scale_height = CAP.R_d(params) * CAP.T_surf_ref(params) / CAP.grav(params)
 
-    # Approximation factor used in w₀ - w ≈ -(ρ / ρa⁰) w (single-updraft case).
-    # For multiple updrafts we approximate ρ / ρa⁰ ≈ 1, which implies w₀ ≈ 0.
-    ᶜρ_over_ρa⁰ = p.scratch.ᶜtemp_scalar
-    @. ᶜρ_over_ρa⁰ = Y.c.ρ / ρa⁰(Y.c.ρ, Y.c.sgsʲs, turbconv_model)
+    # Approximation factor used in w₀ - w ≈ -(1 / a⁰) w (single-updraft case).
```
szy21, line 161: is this assumption better than the previous one?

szy21 (approved):

src/prognostic_equations/implicit/initialize_implicit_problem.jl
sajjadazimi: I assume `rho \approx rho^0` and simplify the ratio.

sajjadazimi: We need `a^0` for drag coeff anyways, so using it for this approximation instead of defining a new variable helps readability I think.

szy21: ok!

PR #4798, opened 2026-09-02
Fix command-line job ID derivation

### LINKED ISSUE #3029: Running jobs from the command line without setting `--job_id` results in `job_id = default_config`
Example with `flame.jl` (works also with `driver.jl`):
```sh
julia --color=yes --project=perf perf/flame.jl --config_file config/perf_configs/flame_perf_target.yml
```
We can see that:
```
┌ Info: Simulation info
│   job_id = "default_config"
└   output_dir = "output/default_config/output_active"
```
It would be nice if the behavior mirrored the one of `AtmosConfig` (where the job_id is constructed from the filename when `job_id` is not explicitely passed)

barislabs:
## Purpose

Fixes #3029.

Passing `--config_file` without `--job_id` currently leaves the run named `default_config`, even when another configuration is selected. This lets `AtmosConfig` derive the implicit job ID from the selected config files, while preserving explicit CLI and YAML job IDs.

## To-do

- [x] Add a `NEWS.md` entry.
- [ ] Confirm CI.

## Content

- Changed the parser default for `--job_id` to `nothing`, allowing the existing `AtmosConfig` fallback order to apply.
- Added a regression test covering command-line selection of a config without an explicit job ID.
- Added the required bug-fix/behavioral-change release note.
- Verified the default config, explicit `--job_id`, YAML `job_id`, and filename-derived paths locally.
- Ran the pinned JuliaFormatter 2.10.1 formatter and `git diff --check`.
- Ran `julia --startup-file=no --project=. test/config.jl` successfully.
- Confirmed with a sabotage run that the regression test fails with the old parser default and passes with the fix restored.

I used Hermes Agent to help implement and verify this change.

----

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,8 @@ ClimaAtmos.jl Release Notes
 main
 ----
 
+- [#4798](https://github.com/CliMA/ClimaAtmos.jl/pull/4798) ![][badge-🐛bugfix] ![][badge-🔥behavioralΔ] Made command-line runs without `--job_id` use the `job_id` from the selected configuration, or derive it from the selected config file names, instead of always using `default_config`.
+
 0.42.9
 -------
 - Update to ClimaCore.jl v0.16 (support for v0.15 is dropped). The biased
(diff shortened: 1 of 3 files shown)


barislabs:
Rebased onto the current `main` and resolved the `NEWS.md` release-note conflict. The branch is mergeable again; the six first-time-contributor workflows are awaiting maintainer authorization.

PR #4799, opened 2026-09-03
Update to CloudMicrophysics v0.39

trontrytel:
This PR updates ClimaAtmos to CloudMicrophysics v0.39. It also changes the behavior by making:
- vapor to ice formation timescale dpenend on prescribed number concentration of ice crystals
- add cloud liquid to ice freezing by homogeneous and heterogeneous pathways


.buildkite/Project.toml
```diff
@@ -13,6 +13,7 @@ ClimaDiagnostics = "1ecacbb8-0713-4841-9a07-eb5aa8a2d53f"
 ClimaParams = "5c42b081-d73a-476f-9059-fd94b934656c"
 ClimaTimeSteppers = "595c0a79-7f3d-439a-bc5a-b232dc3bde79"
 ClimaUtilities = "b3f4f4ca-9299-4f7f-bd9b-81e1242a7513"
+CloudMicrophysics = "6a9e3e04-43cd-43ba-94b9-e8782df3c71b"
```
szy21, line 16: We shouldn't need this

.buildkite/Manifest-v1.11.toml
szy21, line 1: Do you know why some packages are removed here?

.buildkite/Project.toml
szy21: @nefrathenrici Do you know why some packages are in this file, but some others are not?

trontrytel: I struggled to make package manager update CM to v 0.39 in buildkite, even though it was already updated for the Atmos project. This is what finally made it work. Maybe its ok to keep it, since we have already so many other packages from Clima mixed here?

.buildkite/Manifest-v1.11.toml
trontrytel: In my fight to update CM to v0.39 at some point I updated my Julia registers. I think this is the effect of this.

.buildkite/Project.toml
szy21: Do we need a compat for it then, like the other packages?

trontrytel:
Regarding the Beres unit tests, per Claude:

`The Beres test compares two independent formulations of the same latent heating calculation. Both use SpecialFunctions internally (via gamma functions in the Marshall-Palmer size distributions), and the bump from 2.7.2 → 2.9.0 in the Pkg.update() changed intermediate rounding enough to push the relative error from ~1e-11 to ~1e-7 — still excellent agreement, but past the old rtol = 1e-10 gate.`

trontrytel:
The documentation is changed because

```
Cannot find `ClimaCore.Operators.WeakDivergence` in any InterLinks inventory
Cannot find `ClimaCore.Operators.WeakGradient` in any InterLinks inventory  
Cannot find `ClimaCore.Operators.WeakCurl` in any InterLinks inventory
```

those docstrings were deleted from ClimaCore

.buildkite/Project.toml
trontrytel: Added

trontrytel:
Here are effects it has on deep clouds - no more unreasonable liquid water high up, still maintaining mixed phase region, less snow accumulating high up.

New TRMM

Screenshot 2026-09-04 at 8 40 00 AM

Old TRMM

Screenshot 2026-09-04 at 8 41 10 AM

szy21 (approved):

szy21:
> Regarding the Beres unit tests, per Claude:
> 
> `The Beres test compares two independent formulations of the same latent heating calculation. Both use SpecialFunctions internally (via gamma functions in the Marshall-Palmer size distributions), and the bump from 2.7.2 → 2.9.0 in the Pkg.update() changed intermediate rounding enough to push the relative error from ~1e-11 to ~1e-7 — still excellent agreement, but past the old rtol = 1e-10 gate.`

@ray-chew please double check this is ok

PR #4800, opened 2026-09-03
TKE fixes

sajjadazimi:
## Purpose 
- Add tke source due to entr/detr mixing
- Delete stability-biased buoygrad at cell centers and use unbiased buoygrad instead. The stability-biased buoygrad causes too small mixing length and too large dissipation close to the BL top.

----

### DIFF
--- a/src/prognostic_equations/edmfx_tke.jl
+++ b/src/prognostic_equations/edmfx_tke.jl
@@ -80,6 +80,8 @@ function edmfx_tke_sources!(Yₜ, Y, p)
 
     # Pressure-drag return-to-isotropy
     edmfx_pressure_drag_tke_source!(Yₜ, Y, p, p.atmos.turbconv_model)
+    # Entr/detr shear-mixing source: 0.5 δ · ρa · |Δw|² per updraft.
+    edmfx_entr_detr_tke_source!(Yₜ, Y, p, p.atmos.turbconv_model)
     return nothing
 end
 
@@ -95,10 +97,11 @@ function edmfx_pressure_drag_tke_source!(
     α_d = CAP.pressure_normalmode_drag_coeff(turbconv_params)
     a_min = CAP.min_area(turbconv_params)
     scale_height = CAP.R_d(p.params) * CAP.T_surf_ref(p.params) / CAP.grav(p.params)
-    (; ᶜρʲs, ᶠu³ʲs, ᶠu³⁰) = p.precomputed
+    (; ᶜρʲs, ᶜuʲs, ᶜu⁰) = p.precomputed
     # Environment area, shared across all updrafts.
     ᶜa⁰ = @. lazy(a⁰(Y.c.sgsʲs, ᶜρʲs, turbconv_model))
     ᶜdrag_coeff = p.scratch.ᶜtemp_scalar
+    ᶜlg = Fields.local_geometry_field(Y.c)
     for j in 1:n
         ᶜaʲ = @. lazy(draft_area(Y.c.sgsʲs.:($$j).ρa, ᶜρʲs.:($$j)))
         # Same `ᶜdrag_coeff` form as the momentum equation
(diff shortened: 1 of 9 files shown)


src/prognostic_equations/edmfx_tke.jl
```diff
@@ -107,13 +110,83 @@ function edmfx_pressure_drag_tke_source!(
...
+                Y.c.sgsʲs.:($$j).ρa,
+                ᶜbuoy_inv_time_scale,
+                ᶜdivᵥ(ᶠleft_bias_zero_bot(Y.c.sgsʲs.:($$j).ρa) * ᶠu³ʲs.:($$j)),
+                ᶜarea_bounding_entr_detrʲs.:($$j),
+                detr_model,
+            ),
+        )
+
+        # (ε + δ) · ρa_j · |Δw|² at faces, brought to centers.
+        @. Yₜ.c.ρtke += (ᶜentr + ᶜdetr) * Y.c.sgsʲs.:($$j).ρa *
```
szy21: why are both entrainment and detrainment TKE sources?

sajjadazimi: It was a mistake. I thought tke is produced due to shear between updraft and env which scales with total mass exchange -- entr + detr. I guess we don't model that kind of shear. So we only need detr cow->iso term and don't need entr here.

src/prognostic_equations/edmfx_tke.jl
```diff
@@ -107,13 +110,71 @@ function edmfx_pressure_drag_tke_source!(
...
+    (; ᶠgradᵥ_ᶜΦ) = p.core
+
+    turbconv_params = CAP.turbconv_params(p.params)
+    entr_detr_buoy_inv_tau_max =
+        CAP.entr_detr_buoy_inv_tau_max(turbconv_params)
+    detr_model = p.atmos.edmfx_model.detr_model
+    ᶜlg = Fields.local_geometry_field(Y.c)
+    ᶠlg = Fields.local_geometry_field(Y.f)
+    # Zero-bottom-BC upwind for the mass-flux divergence (matches `compute_detr`
+    # diagnostic to avoid NaN at the first cell).
```
szy21: I did that just to avoid NaNs in the diagnostics, and the detrainment actually used in the implicit solver is not exactly the same as the one in diagnostics. Is it ok to use a slightly different detrainment here?

src/prognostic_equations/edmfx_tke.jl
```diff
@@ -107,13 +110,71 @@ function edmfx_pressure_drag_tke_source!(
...
+                ᶜbuoy_inv_time_scale,
+                ᶜdivᵥ(ᶠleft_bias_zero_bot(Y.c.sgsʲs.:($$j).ρa) * ᶠu³ʲs.:($$j)),
+                ᶜarea_bounding_entr_detrʲs.:($$j),
+                detr_model,
+            ),
+        )
+
+        @. Yₜ.c.ρtke +=
+            FT(0.5) * ᶜdetr * Y.c.sgsʲs.:($$j).ρa *
+            get_physical_w(ᶜuʲs.:($$j) - ᶜu⁰, ᶜlg)^2
```
szy21, line 175: FYI when I had this before I used `norm_sqr(ᶜinterp(ᶠu³⁰) - ᶜinterp(ᶠu³ʲs.:($$j)))`. Not sure which one is better, it's probably similar.

src/prognostic_equations/edmfx_tke.jl
```diff
@@ -80,6 +80,8 @@ function edmfx_tke_sources!(Yₜ, Y, p)
 
     # Pressure-drag return-to-isotropy
     edmfx_pressure_drag_tke_source!(Yₜ, Y, p, p.atmos.turbconv_model)
+    # Entr/detr shear-mixing source: (ε + δ) · ρa · |Δw|² per updraft.
```
szy21:
```suggestion
    # Entr/detr shear-mixing source: 0.5 δ · ρa · |Δw|² per updraft.
```

szy21 (approved):

src/prognostic_equations/edmfx_tke.jl
sajjadazimi: I think it's fine. The small difference is compensated with dissipation.

PR #4802, opened 2026-09-04
add sgs variance

szy21:
## Purpose 

Add horizontal resolved-gradient (geometric) SGS variance term. A step towards improving RH in the extratropics.

## Content

----

### DIFF
--- a/src/parameterized_tendencies/microphysics/cloud_fraction.jl
+++ b/src/parameterized_tendencies/microphysics/cloud_fraction.jl
@@ -220,6 +220,26 @@ function materialized_mixing_length!(Y, p)
 end
 
 """
+    hgrad_invariant!(ᶜinv, ᶜψ, p)
+
+Write the horizontal-gradient invariant `|∇_h ψ|²` of the center field `ᶜψ` into `ᶜinv`.
+The gradient vector is first made continuous across element boundaries by a weighted DSS.
+On a space that needs no DSS (single column) the nodal gradient is used.
+"""
+function hgrad_invariant!(ᶜinv, ᶜψ, p)
+    buf = p.scratch.ᶜC12_dss_buffer
+    if buf !== nothing
+        ᶜg = p.scratch.ᶜtemp_C12
+        @. ᶜg = gradₕ(ᶜψ)
+        Spaces.weighted_dss!(ᶜg, buf)
+        @. ᶜinv = norm_sqr(ᶜg)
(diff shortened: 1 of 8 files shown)


src/parameterized_tendencies/microphysics/cloud_fraction.jl
```diff
@@ -219,6 +219,95 @@ function materialized_mixing_length!(Y, p)
...
+it conserves the WJ-weighted element integral exactly (Σ_ab B_ia B_jb = 1), reproduces a
+constant exactly, and retains about one third of a linear slope inside the element (it is
+a smoothing filter, not an L2 projection). It removes the element-edge enhancement of
+the nodal `|∇_h ψ|²` while conserving each element's WJ-weighted mean.
+Implemented as reductions over the two horizontal node dimensions of the
+`(Nv, Ni, Nj, Nf, Nh)` data layout (CPU and GPU arrays); allocates temporaries of a few
+times the field size per call. A no-op on spaces without a horizontal spectral element
+(single columns).
+"""
+function element_linear!(ᶜf)
```
szy21, line 263: This is claude generated, and I don't really understand it. And it's probably not ideal as it uses `parent`. But this is the only way I tried that fixes the grid imprinting caused by horizontal gradients. By default this is turned off, but I want to keep this code as an option. I'm fine with removing it though.

src/parameterized_tendencies/microphysics/cloud_fraction.jl
```diff
@@ -262,10 +387,88 @@ function set_covariance_cache!(Y, p, thermo_params)
...
+                ifelse(
+                    TD.has_condensate(thermo_params, ᶜq_lcl + ᶜq_icl),
+                    one(FT_w),
+                    zero(FT_w),
+                ),
+                projected_vector_data(C3, ᶜgradᵥ_θ_liq_ice, ᶜlg_w),
+                projected_vector_data(C3, ᶜgradᵥ_q_tot, ᶜlg_w),
+            ),
+        )
+        ᶜw = @. lazy(sgs_geometric_stability_weight(ᶜN²_w, ᶜstrain_rate_norm, Ri₀))
```
szy21, line 441: This weight is not very physical, but that's the only thing I tried that can differentiate boundary layer and free troposphere. We can revisit it. It's off by default too.

sajjadazimi (approved):

PR #4804, opened 2026-09-04
Update to ClimaParams v1.1.7; read the reference-state exponent from ClimaParams

ray-chew:
## Purpose

Update to ClimaParams v1.1.7, which changes the default reference temperatures to match the dycore manuscript (CliMA/ClimaParams.jl#253), and read the reference-profile exponent `s_ref` from ClimaParams instead of a hard-coded constant.

## Content

- Remove `const s_ref = 7` from `refstate_thermodynamics.jl`. The exponent of  `T_r(p) = T_min + (T_sfc - T_min) Π^s_ref` now comes from the ClimaParams parameter `reference_temperature_exponent`, exposed as `CAP.s_ref(params)`.
- `air_temperature_reference`, `theta_vr`, `phi_r`, `q_tot_r`, and `sd_r` take `s_ref` as an argument; callers in `advection.jl`, `implicit_tendency.jl`, and `hyperdiffusion.jl` pass it in.
- The field is stored as an `Int`, so `Π^s_ref` is computed exactly as the old `Π^7` was. Verified bitwise identical for Float32 and Float64.
- `temperature_surface_reference` 290 K → 288 K and `temperature_min_reference` 220 K → 215 K.
-  `ref_counter` 409 → 411 (an earlier commit didn't bump to 410).
- Docstrings, `docs/src/equations.md`, and NEWS updated with the behavioral-change badge.

----


src/parameters/Parameters.jl
```diff
@@ -559,6 +561,8 @@ Base.@kwdef struct ClimaAtmosParameters{
     zd_viscous::FT
     zd_rayleigh::FT
     kappa_2_sponge::FT
+    # Hydrostatic reference state
+    s_ref::Int
```
szy21, line 565: Is this Int or FT? it's float in ClimaParams

PR #4806, opened 2026-09-07
Initialize microphysics condensates from file-based ICs

costachris:
Initialize cloud liquid/ice and rain/snow from clwc/ciwc/crwc/cswc when present for 1M/2M,
and assemble `ρq_tot` as vapor plus those condensates so diagnosed vapor
matches the ERA5 file.

ERA5 `q` is vapor; Atmos `ρq_tot` is total water. IC path now uses `ᶜq_vap`, seeds condensates from clwc/ciwc/crwc/cswc, and builds ρq_tot = vapor + condensates. Density/energy Thermo still vapor only (should be a very small error).

Shorten RCEMIP test, remove from reproducibility

## Content

----

### DIFF
--- a/src/setups/common/overwrite_from_file.jl
+++ b/src/setups/common/overwrite_from_file.jl
@@ -10,7 +10,7 @@
 
 """
     correct_surface_pressure_for_topography!(
-        p_sfc, file_path, face_space, Y, ᶜT, ᶜq_tot,
+        p_sfc, file_path, face_space, Y, ᶜT, ᶜq_tot, ᶜq_liq, ᶜq_ice,
         thermo_params, regridder_kwargs;
         surface_altitude_var = "z_sfc",
     )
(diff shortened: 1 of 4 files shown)


szy21:
Why not using q_tot in the total energy too?

costachris:
> Why not using q_tot in the total energy too?

Originally because we started with 0M and the marginal benefit in terms of bias is small. 
But now updated to use the full `q_tot` throughout, instead of `q_vap` for Density/energy Thermo

szy21:
If you want to make the notation consistent with the rest of the code, it would be: ᶜq_liq -> ᶜq_lcl, ᶜq_ice -> ᶜq_icl, ᶜq_liq_thermo -> ᶜq_liq, ᶜq_ice_thermo -> ᶜq_ice. I think that may be less confusing, but up to you.

szy21 (approved):

src/setups/common/overwrite_from_file.jl
```diff
@@ -116,23 +129,142 @@ function assign_velocity_energy!(
...
+partitions). To keep the diagnosed vapor equal to the file value, each condensate
+species is added to `ᶜq_tot` only when it is both present in the file and carried
+as a prognostic variable in `Y` — exactly the species that
+`assign_moisture_edmf!` stores into `ρq_tot`. Following the runtime
+convention in the precomputed thermodynamic state, rain is grouped with the
+liquid partition and snow with the ice partition.
+
+For the equilibrium (0M) scheme (no prognostic condensate fields) and for files
+without condensate, this reduces to `ᶜq_tot = ᶜq_vap` and `ᶜq_liq = ᶜq_ice = 0`,
+i.e. the previous vapor-only behavior.
```
szy21: If you want to make 0M a bit more correct, you can add ᶜq_liq and ᶜq_ice to ᶜq_tot when the files have condensate. Up to you.

PR #4811, opened 2026-09-08
Add RCEMIPI setup

oalcabes:
This PR adapts the RCEMIPIIProflle.jl setup to be a more general RCEMIPProfile.jl setup, that allows for both RCEMIPII and RCEMIPI experiments. The only difference between the two experiments are how surface SSTs are treated: RCEMIPII has a constant SST with a surface wave, whereas RCEMIPI has a constant SST, and noise is added to the lowest 5 atmospheric layers with a decreasing amplitude to induce convection. It also adds new documentation for box setups in ClimaAtmos!


src/setups/RCEMIPProfile.jl
```diff
@@ -0,0 +1,191 @@
...
+const NOISE_SEED = 2018
+
+function rcemipi_thermal_noise(coords, FT)
+    z = coords.z
+    k = findfirst(zc -> abs(z - zc) < NOISE_TOL, NOISE_LAYER_HEIGHTS)
+    k === nothing && return zero(FT)
+    x = hasproperty(coords, :x) ? coords.x : zero(FT)
+    y = hasproperty(coords, :y) ? coords.y : zero(FT)
+    rng = Random.Xoshiro(hash((NOISE_SEED, x, y, z)))
+    return FT(NOISE_AMPLITUDES[k]) * (2 * rand(rng) - 1)  # uniform in [-A, A]
```
haakon-e: what is `A` here?

src/setups/RCEMIPProfile.jl
```diff
@@ -0,0 +1,191 @@
...
+const NOISE_TOL = 1.0    # m
+const NOISE_SEED = 2018
+
+function rcemipi_thermal_noise(coords, FT)
+    z = coords.z
+    k = findfirst(zc -> abs(z - zc) < NOISE_TOL, NOISE_LAYER_HEIGHTS)
+    k === nothing && return zero(FT)
+    x = hasproperty(coords, :x) ? coords.x : zero(FT)
+    y = hasproperty(coords, :y) ? coords.y : zero(FT)
+    rng = Random.Xoshiro(hash((NOISE_SEED, x, y, z)))
```
haakon-e: what's the intention of `hash((NOISE_SEED, x, y, z))`?

src/setups/RCEMIPProfile.jl
```diff
@@ -0,0 +1,191 @@
...
+#
+# noise layers are set by the prescribed RCEMIP zmesh, the 5 lowest layers
+const NOISE_LAYER_HEIGHTS = (18.5, 74.5, 153.0, 241.0, 341.5)  # m, lowest 5 centers
+const NOISE_AMPLITUDES = (0.1, 0.08, 0.06, 0.04, 0.02)      # K, layers 1..5
+const NOISE_TOL = 1.0    # m
+const NOISE_SEED = 2018
+
+function rcemipi_thermal_noise(coords, FT)
+    z = coords.z
+    k = findfirst(zc -> abs(z - zc) < NOISE_TOL, NOISE_LAYER_HEIGHTS)
```
haakon-e, line 105: what does this do?

src/setups/RCEMIPProfile.jl
```diff
@@ -0,0 +1,191 @@
...
+
+# ── RCEMIP-I symmetry-breaking noise (Wing et al. 2018) ───────────────────────
+# One-off thermal perturbation added to the *initial* temperature in the five
+# lowest layers: 0.1 K in the lowest, decreasing linearly to 0.02 K in the
+# fifth, drawn independently per column and per layer so convective symmetry is
+# broken. Applied for RCEMIPI only. Seeded from coordinates → reproducible and
+# decomposition-independent.
+#
+# noise layers are set by the prescribed RCEMIP zmesh, the 5 lowest layers
+const NOISE_LAYER_HEIGHTS = (18.5, 74.5, 153.0, 241.0, 341.5)  # m, lowest 5 centers
```
haakon-e, line 98: seems a bit fragile if someone runs rcemip with different vertical levels than the default.
What would happen then?

src/setups/RCEMIPProfile.jl
```diff
@@ -0,0 +1,191 @@
...
+const NOISE_AMPLITUDES = (0.1, 0.08, 0.06, 0.04, 0.02)      # K, layers 1..5
+const NOISE_TOL = 1.0    # m
+const NOISE_SEED = 2018
+
+function rcemipi_thermal_noise(coords, FT)
```
haakon-e: nit: you can probably get `FT` from `eltype(coords)`. but this signature is fine, too.

haakon-e (approved):
Just some questions from me.
Overall looks good.

src/setups/RCEMIPProfile.jl
oalcabes: amplitude! i added a note

src/setups/RCEMIPProfile.jl
oalcabes: checks that the height is one of the first 5

src/setups/RCEMIPProfile.jl
oalcabes: then no noise should be implemented in those layers unless if one of their layers is very close to these values. i do think because RCEMIP is given with a prescribed mesh it's probably okay to assume the layers are fixed

haakon-e: Could we document this somewhere? I know I'll forget about this if I or someone else is interested in running high-resolution RCEMIP where we need to modify the vertical levels.

Perhaps we can add a new page for "Running Box Simulations" that can be mostly empty/TODO for now but contain this note in a subsection about RCEMIP?
See the [SCM page](https://clima.github.io/ClimaAtmos.jl/dev/single_column/) and [global simulations page](https://clima.github.io/ClimaAtmos.jl/dev/global_simulations/) for reference.

oalcabes: Sure! Good idea. I will draft something up momentarily.

oalcabes: I added docs, check it out!

trontrytel:
My Claude is worried that it's almost impossible to hit all prescribed noise layers heights with the way Atmos defines its stretched vertical grid.  For `config/model_configs/rcemipii_box_CRM_1M.yml` in our CI we have ` (dz_bottom 30, 86 levels, 30 km)` which ends up meaning our 5 lowest levels are at:  `15.0, 45.8, 78.4, 112.7, 148.9 m`. So none of them would be affected. If I specify `dz_bottom 37, 74 to 100 levels, 33 km`, I get `18.5, 56.5, 96.4, 138.4, 182.5 m` and only one gets affected.

trontrytel:
Maybe a more robust alternative would be to define the perturbation profile as a decaying with height function and interpolate into it?

src/setups/RCEMIPProfile.jl
haakon-e: There's some minor formatting issues with backslashes in e.g., `RCE\_small` that can be fixed. 
Otherwise, I think these docs are a great start we can expand upon later! :)

src/setups/RCEMIPProfile.jl
```diff
@@ -0,0 +1,186 @@
...
+RCEMIPIProfile_295() = RCEMIPProfile(RCEMIPI(), 295.0, 12e-3)
+RCEMIPIProfile_300() = RCEMIPProfile(RCEMIPI(), 300.0, 18.65e-3)
+RCEMIPIProfile_305() = RCEMIPProfile(RCEMIPI(), 305.0, 24e-3)
+
+# ── RCEMIP-I symmetry-breaking noise (Wing et al. 2018) ───────────────────────
+# One-off thermal perturbation added to the *initial* temperature in the five
+# lowest layers: 0.1 K in the lowest, decreasing linearly to 0.02 K in the
+# fifth, drawn independently per column and per layer so convective symmetry is
+# broken. Applied for RCEMIPI only. Seeded from coordinates → reproducible and
+# decomposition-independent.
```
trontrytel, line 95: The code does: `FT(NOISE_AMPLITUDES[k]) * (2 * rand(FT) - 1)` so I'm not seeing the seeding from the coordinates. I think this will actually mean different rand values per run and can affect the reproducibility tests.

src/setups/RCEMIPProfile.jl
```diff
@@ -0,0 +1,186 @@
...
+function rcemipii_temperature(
+    coordinates::Union{Geometry.XZPoint, Geometry.XYZPoint},
+    surface_temp_params, _,
+)
+    (; x) = coordinates
+    (; SST_mean, SST_delta, SST_wavelength) = surface_temp_params
+    return SST_mean - SST_delta / 2 * cospi(2 * x / SST_wavelength)
+end
+
+import Random
```
trontrytel, line 171: Is ^ this needed?

src/config/type_getters.jl
```diff
@@ -158,6 +158,7 @@ values map to same-named types in `Setups`:
   - LES/SCM cases: `"Bomex"`, `"Rico"`, `"Soares"`, `"GATE_III"`, `"DYCOMS_RF01"`,
     `"DYCOMS_RF02"`, `"TRMM_LBA"`, `"Larcform1"`, `"GABLS"`, `"ISDAC"`, which read
     `prognostic_tke` (and, for ISDAC, `perturb_initstate`).
+  - RCEMIP I: `"RCEMIPIProfile_295"`, `"RCEMIPIProfile_300"`, `"RCEMIPIProfile_305"`.
```
trontrytel, line 161: This is a bad pattern we had before: Se set the SST in the toml but the profiles through the config. That means we can be running simulations with buggy temperature jump between SST and profile and never notice. It might be good to add a sanity check somewhere and throw an error. Or rewrite it and set SST at the same time as we chose the profile

docs/src/box_simulations.md
```diff
@@ -0,0 +1,35 @@
...
+### RCEMIPII
+
+Differing from the original RCEMIP, the RCEMIPII simulation is initialized with a sinusoidal wave in sea surface temperature as described by [Wing et. al. (2024)](https://doi.org/10.5194/gmd-17-6195-2024):
+
+```math
+    SST(x) = \braket{SST} - \frac{\Delta SST}{2} \cos\left(\frac{2 \pi x}{L_x} \right)
+```
+
+The `SST_mean` and `SST_delta` are set by parameters in the toml and through the choice of RCEMIPII profile: `RCEMIPIIProfile_295`, `RCEMIPIIProfile_300`, or `RCEMIPIIProfile_305`.
+$L_x$ is the zonal size of the domain, which is automatically calculated from the simulation and is 100km for `RCE\_small`.
```
trontrytel, line 33: This is still set by the toml file `[SST_wavelength]`? So it's possible to run the small domain simulation with large wavelength and vice versa.

trontrytel (changes requested):
I left some Claude augmented comments. I'm mostly worried about:
- how to ensure that our grid hits the noise levels 
- how to ensure the wavelength, SST, and profile consistency with the current simulation setup

oalcabes:
> My Claude is worried that it's almost impossible to hit all prescribed noise layers heights with the way Atmos defines its stretched vertical grid. For `config/model_configs/rcemipii_box_CRM_1M.yml` in our CI we have ` (dz_bottom 30, 86 levels, 30 km)` which ends up meaning our 5 lowest levels are at: `15.0, 45.8, 78.4, 112.7, 148.9 m`. So none of them would be affected. If I specify `dz_bottom 37, 74 to 100 levels, 33 km`, I get `18.5, 56.5, 96.4, 138.4, 182.5 m` and only one gets affected.

When Haakon and I run this using the julia runfiles, we specify a specific vertical grid that has the correct values based on the Wing et. al. paper. I see two options here: either (1) add one of those runfiles and remove the .yml or (2) add the z-mesh into this profile somehow. Lmk what you think

src/config/type_getters.jl
oalcabes: It is unfortunately annoying to just set it with the profile. Instead I have attempted to add an error if there is a toml/initial condition profile mismatch.

trontrytel:
> > My Claude is worried that it's almost impossible to hit all prescribed noise layers heights with the way Atmos defines its stretched vertical grid. For `config/model_configs/rcemipii_box_CRM_1M.yml` in our CI we have ` (dz_bottom 30, 86 levels, 30 km)` which ends up meaning our 5 lowest levels are at: `15.0, 45.8, 78.4, 112.7, 148.9 m`. So none of them would be affected. If I specify `dz_bottom 37, 74 to 100 levels, 33 km`, I get `18.5, 56.5, 96.4, 138.4, 182.5 m` and only one gets affected.
> 
> When Haakon and I run this using the julia runfiles, we specify a specific vertical grid that has the correct values based on the Wing et. al. paper. I see two options here: either (1) add one of those runfiles and remove the .yml or (2) add the z-mesh into this profile somehow. Lmk what you think

I think that having a CI example of the setup you both are actually running would be great. That would also allow us to actually test the noise application. And make sure we don't break things for you. I'm not sure how your runfiles are compatible with the CI? @nefrathenrici  - what do you think?

src/config/type_getters.jl
trontrytel: Thank you!

nefrathenrici:
> > > My Claude is worried that it's almost impossible to hit all prescribed noise layers heights with the way Atmos defines its stretched vertical grid. For `config/model_configs/rcemipii_box_CRM_1M.yml` in our CI we have ` (dz_bottom 30, 86 levels, 30 km)` which ends up meaning our 5 lowest levels are at: `15.0, 45.8, 78.4, 112.7, 148.9 m`. So none of them would be affected. If I specify `dz_bottom 37, 74 to 100 levels, 33 km`, I get `18.5, 56.5, 96.4, 138.4, 182.5 m` and only one gets affected.
> > 
> > 
> > When Haakon and I run this using the julia runfiles, we specify a specific vertical grid that has the correct values based on the Wing et. al. paper. I see two options here: either (1) add one of those runfiles and remove the .yml or (2) add the z-mesh into this profile somehow. Lmk what you think
> 
> I think that having a CI example of the setup you both are actually running would be great. That would also allow us to actually test the noise application. And make sure we don't break things for you. I'm not sure how your runfiles are compatible with the CI? @nefrathenrici - what do you think?

I am not against a new runfile, but it depends what it looks like and must be tested in CI. We already have an example for RCEMIPII in `runscripts`, we could move its grid into a shared script for both. I'd also like to move them both into `examples` but that's a separate issue.

src/setups/RCEMIPProfile.jl
oalcabes: So in the runscript Random.seed(1234) is set globally, which should dictate what rand does. I'm not sure what the yaml that is in our current ci is using, but if we switched to use the julia runscript instead, this should be fine for reproducibility

src/setups/RCEMIPProfile.jl
oalcabes: I changed this to just be a linear function with height and added info to the docs.

oalcabes:
> > > > My Claude is worried that it's almost impossible to hit all prescribed noise layers heights with the way Atmos defines its stretched vertical grid. For `config/model_configs/rcemipii_box_CRM_1M.yml` in our CI we have ` (dz_bottom 30, 86 levels, 30 km)` which ends up meaning our 5 lowest levels are at: `15.0, 45.8, 78.4, 112.7, 148.9 m`. So none of them would be affected. If I specify `dz_bottom 37, 74 to 100 levels, 33 km`, I get `18.5, 56.5, 96.4, 138.4, 182.5 m` and only one gets affected.
> > > 
> > > 
> > > When Haakon and I run this using the julia runfiles, we specify a specific vertical grid that has the correct values based on the Wing et. al. paper. I see two options here: either (1) add one of those runfiles and remove the .yml or (2) add the z-mesh into this profile somehow. Lmk what you think
> > 
> > 
> > I think that having a CI example of the setup you both are actually running would be great. That would also allow us to actually test the noise application. And make sure we don't break things for you. I'm not sure how your runfiles are compatible with the CI? @nefrathenrici - what do you think?
> 
> I am not against a new runfile, but it depends what it looks like and must be tested in CI. We already have an example for RCEMIPII in `runscripts`, we could move its grid into a shared script for both. I'd also like to move them both into `examples` but that's a separate issue.

I just fixed a lot of these problems, but the runscript that currently exists does have the right height levels, so if it makes sense to add that to ci as opposed to the yaml in the future that would be great!

haakon-e (approved):

trontrytel (approved):

PR #4814, opened 2026-09-10
Replace deprecated ClimaCore names, aliases and constructors

imreddyTeja:
ClimaCore keeps backwards-compatibility shims for the DataLayouts, Geometry and Operators rewrites. This removes every remaining use of them here, so those shims can be deleted upstream. All results should be unchanged.

This mainly changed a few things:

Usage of AxisTensor after the Geometry refactor

DataLayouts after the DataLayouts refactor
MatMul /cdot to * ; This was changed a long time ago
RecursiveApply -> autobroadcaasters

Disclosure: This commit was heavily assisted by claude

The rest of this PR description was made by claude. Feel free to ignore if it is not useful.

## Deprecated usage changes

| Deprecated | Replacement | Where |
|---|---|---|
| `Geometry.AxisTensor{T, N, B, S}` | `Geometry.Tensor{N, T, B, S}` (T and N swapped) | 4 code sites + docstrings |
| `Geometry.AxisTensor(bases, comps)` | `Geometry.Tensor(comps, bases)` | `g³ʰ` |
| `Geometry.components(x)` | `parent(x)` | `g³ʰ` |
| `Geometry.ContravariantAxis{(3,)}` / `{(1, 2)}` | `Contravariant3Axis` / `Contravariant12Axis` | `temporary_quantities.jl` |
| `DataLayouts.AbstractData` | `DataLayouts.DataLayout` | 6 |
| `Operators.Weak{Divergence,Gradient,Curl}` | `{Divergence,Gradient,Curl}{WeakForm}` | `abbreviations.jl` |
| `MatrixFields.⋅` | `*` | 75, `manual_sparse_jacobian.jl` |
| `FaceFiniteDifferenceSpace(mesh)` | `(device, mesh)` | 6 tests (one dead, deleted) |
| `ClimaCore.RecursiveApply` | `Utilities.add`/`drop_auto_broadcasters` | 4 files |

For `RecursiveApply`, `sum_over_quadrature_points` now wraps its accumulators once and unwraps once rather than at every step, and `unrolled_dotproduct`'s private recursion moved to `_unrolled_dotproduct` so the wrap/unwrap boundary sits at the public two-argument entry point. Plain `zero(T)` would not do: the reduction builds zeros of composite types.

Three dead imports are dropped along the way (`Geometry: AxisTensor` in both restart test files, the unused `RecursiveApply` import in `cloud_fraction.jl`), and `docs/src/discretization.md` no longer claims the code uses the legacy weak-operator names.

## Verification

- **Bit-for-bit identical results.** Ten implicit steps of a prognostic-EDMFX Bomex column with 1M microphysics, run once with these changes and once with `src/` reverted: `max abs diff: 0.0` across the full state vector. Repeated against a shim-free ClimaCore with the same result.
- **The `⋅` → `*` rewrite is AST-equivalent**, checked by parsing before/after and normalising both operators to left-folded binary `*`. `⋅` parsed as a binary operator with `*`'s precedence and left associativity, and `MatrixFields` reduces an n-ary `*` broadcast left-associatively (`MatrixFields.jl:107`), so the two agree.
- **Each alias substitution is type-identical**, confirmed by reflection against ClimaCore: `AxisTensor{FT, 2, …} === Tensor{2, FT, …}`, `AbstractData{<:Real} == DataLayout{<:Real}`, `WeakDivergence() === Divergence{WeakForm}()`, and `Geometry.components(v) === parent(v)`.


src/utils/variable_manipulations.jl
```diff
@@ -615,43 +616,46 @@ the result of `f` on the first elements, which keeps the reduction type-stable.
...
-    unrolled_dotproduct(s, Base.tail(a), Base.tail(b))
-@inline unrolled_dotproduct(s, a::Tuple{<:Any}, b::Tuple{<:Any}) =
-    s ⊞ (first(a) ⊠ first(b))
+@inline _unrolled_dotproduct(s, ::Tuple{}, ::Tuple{}) = s
+@inline _unrolled_dotproduct(s, a::Tuple, b::Tuple) =
+    s +
+    add_auto_broadcasters(first(a)) * add_auto_broadcasters(first(b)) +
+    _unrolled_dotproduct(s, Base.tail(a), Base.tail(b))
+@inline _unrolled_dotproduct(s, a::Tuple{<:Any}, b::Tuple{<:Any}) =
+    s + add_auto_broadcasters(first(a)) * add_auto_broadcasters(first(b))
```
imreddyTeja, line 661: I don't think the `unrolled_dotproduct` -> `_unrolled_dotproduct` change is needed, but Claude thought this was a good idea. I'm impartial about this, so I'm leaving it to reviewers.

src/utils/utilities.jl
```diff
@@ -464,14 +464,14 @@ Extract the `g³ʰ` sub-tensor of the metric tensor `gⁱʲ`, the coupling betwe
...
-The result is always a `Contravariant3Axis × Contravariant12Axis` `AxisTensor`;
+The result is always a `Contravariant3Axis × Contravariant12Axis` `Tensor`;
 in 2D spaces the missing horizontal component is filled with zero. Throws if
 `gⁱʲ` has no vertical or no horizontal sub-axis.
 """
 function g³ʰ(gⁱʲ)
     full_CT_axis = axes(gⁱʲ)[1]
     N = length(full_CT_axis)
-    gⁱʲ_components = Geometry.components(gⁱʲ)
+    gⁱʲ_components = parent(gⁱʲ)
```
szy21, line 474: I thought we want to avoid `parent`?

imreddyTeja: I think we only want to avoid `parent` on `Fields` or `DataLayouts`. I believe this is the new preferred way of accessing the components, and `components` is [deprecated](https://github.com/CliMA/ClimaCore.jl/blob/877647c5a5898cab2aa3f54645959695e75a66f4/src/Geometry/deprecated.jl#L21)

szy21 (approved):
I skimmed through the source code change and it seems fine. Thanks!

src/utils/variable_manipulations.jl
nefrathenrici: I'm confused why this is needed -- can't we just have `unrolled_dotproduct(a, b) = unrolled_mapreduce(*, +, a, b)`?

nefrathenrici (approved):

src/utils/variable_manipulations.jl
imreddyTeja: Good point. I misunderstood what this function was doing. I think it may have been written before unrolled_utilities existed

PR #4817, opened 2026-09-11
Add support for multi col simulation

ph-kev:
This PR adds support for running multiple independent column simulation on a single space. I used a LLM to write a majority of the code.

## TODO

- [ ] Merge https://github.com/CliMA/ClimaCore.jl/pull/2635 in ClimaCore
- [ ] Merge https://github.com/CliMA/ClimaDiagnostics.jl/pull/188 in ClimaDiagnostics
- [ ] Deal with forcing data in ClimaAtmos, ClimaUtilities, and ClimaInterpolations

## Questions

It is unclear to me what work need to be done for the forcing data. The simplest case is to duplicate the same forcing data for all the columns (see https://github.com/CliMA/ClimaUtilities.jl/pull/251 which implements this). Another case is to use different forcing data for each column, but the times shared between the forcing data is the same. This is already implemented in https://github.com/CliMA/ClimaUtilities.jl/pull/248 with linear interpolation and all the data must fit in memory. The hardest case is if the forcing data is different across times and cannot all fit in memory.

Are there examples of use cases where a multiple column simulation would be used and what the forcing data would look like? In particular, it would help to know what the NetCDF files look like for the forcing data so that it can properly implemented in ClimaUtilities.jl and ClimaInterpolations.jl.


src/column_datasets/ColumnDatasets.jl
```diff
@@ -501,6 +501,9 @@ function column_timevaryinginputs(
 )
     names = Tuple(names)
     d = cd.format
+    # One site's profiles apply to every column of a horizontally extended space.
+    horizontally_uniform =
+        !(target_space isa ClimaCore.Spaces.FiniteDifferenceSpace)
```
nefrathenrici, line 506: This is a bit unsafe, since this includes any non-column space, such as a sphere. Could you just check that the space isa MultiColumnFiniteDifferenceSpace?

test/restart_AtmosSimulation.jl
```diff
@@ -373,7 +381,8 @@ if MANYTESTS
                     # Create job_id string from configuration
                     config_name =
                         mesh isa Meshes.EquiangularCubedSphere ? "sphere" :
-                        mesh isa Meshes.RectilinearMesh ? "box" : "column"
+                        mesh isa Meshes.RectilinearMesh ? "box" :
+                        hasproperty(grid, :horizontal_grid) ? "multicolumn" : "column"
```
nefrathenrici, line 385:
```suggestion
                        grid.horizontal_grid isa Grids.MultiPointGrid
```
Does this work instead? It's a bit cleaner.

nefrathenrici (commented):
Thanks, I didn't see any major issues. A few things:

- It looks like the true SCM columns differ from the multicolumn in the latitude-dependent physics (coriolis, insolation, etc). I'm not sure if that needs to be fixed before merge, but it should be documented at the least. @costachris or @szy21 may have an opinion. 
- Could you add a `multicolumn_hydrostatic_balance_ft64.yml` case to Buildkite to complement the existing `single_column_hydrostatic_balance_ft64.yml` case?
- Detecting the grid or space type is clunky, not a blocker, but an observation

src/config/model_getters.jl
```diff
@@ -1036,11 +1038,20 @@ function check_case_consistency(parsed_args)
 
     # Geometry consistency (always checked, independent of the case-specific
     # checks below)
-    valid_configs = ("sphere", "column", "box", "plane")
+    valid_configs = ("sphere", "column", "multicolumn", "box", "plane")
     @assert(
         config in valid_configs,
         "Unknown `config = $(repr(config))`. Valid options are: $(join(valid_configs, ", "))."
     )
+    if config == "multicolumn"
```
Julians42, line 1046: Ideally we'd have a check that the forcing data (if needed) lines up with the lat-lons. The test should go in `test/column_datasets_tests.jl`and use multiple copies of `https://github.com/CliMA/ClimaAtmos.jl/blob/ccca2f70d2942b3959091126f9c7adf66380e05f/test/column_datasets_tests.jl#L270`. It's also possible that we could combine these smaller files into a larger netcdf that has another dimension of simulation or lat / lon. I'm not sure what is best?

Julians42 (commented):
This is a valiant effort and hopefully will result in a significant speed up in SCM calibrations. Does it run efficiently on GPU?

In terms of the main concern, we need to make sure that it can handle different sites both with static forcing (for @costachris 's LES setup) and with time varying forcing (my ERA5 setup). I left a comment pointing to where we test the current functionality / what the files look like as they are generated, but you can also look at sample files on clima here: `/net/sampo/data1/ClimaArtifacts/artifacts/era5_hourly_atmos_processed/monthly` (possibly these are no longer compatible with a recent update from @nefrathenrici ?). I'm not working with this setup for another month or so. 

I'd be most happy if you are able to run a subset of these simulations (or slightly reformatted forcing data for Nat's change) together on a GPU. Currently they are at different locations and times of year and all repeat a day diurnal cycle. if it's easier to say run only the same times together that's fine but multi-location with the different forcing would be fantastic. Happy to talk if this is unclear or if something is tricky to implement.

costachris:
> Thanks, I didn't see any major issues. A few things:
> 
> * It looks like the true SCM columns differ from the multicolumn in the latitude-dependent physics (coriolis, insolation, etc). I'm not sure if that needs to be fixed before merge, but it should be documented at the least. @costachris or @szy21 may have an opinion.
> * Could you add a `multicolumn_hydrostatic_balance_ft64.yml` case to Buildkite to complement the existing `single_column_hydrostatic_balance_ft64.yml` case?
> * Detecting the grid or space type is clunky, not a blocker, but an observation

Compared to the existing single-column SCM (config: column): insolation / default SST can differ in multicolumn because those columns have a real lat. Coriolis doesn’t — it still uses the same f-plane value as single column (iscolumn is true for both). The new docs currently say Coriolis differs though, so that should get fixed.

costachris:
As far as forcing data, for the GCM-driven cases at least, same time/different locations is useful and would speed things up. Forcings are split into 4 months (per season) and cfsite locations. 
Duplicating the same forcing across all columns may still be useful for a single-site cal (or a fixed small set of sites), where GPU parallelization is mostly over parameters.
Single-column GCM-driven forcing data is here (multiple sites per file). e.g. for July: /resnick/groups/esm/zhaoyi/GCMForcedLES/forcing/corrected/HadGEM2-A_amip.2004-2008.07.nc
As Julian noted, right now one site gets broadcast to every column, so you just get N copies of the same site. Without per-column forcing (and surface/ts/insolation), this doesn’t quite fully replace the existing “N independent column sims” pattern (but may have some intermediate advantages at least for gcm driven).

PR #4819, opened 2026-09-11
Release v0.42.10

ph-kev:
This PR makes a new release of ClimaAtmos which will be used in the long runs of ClimaCoupler.

### DIFF
--- a/reproducibility_tests/ref_counter.jl
+++ b/reproducibility_tests/ref_counter.jl
@@ -1,4 +1,4 @@
-410
+411
 
 # **README**
 #
@@ -32,6 +32,9 @@
 # 3) (optional) leave a link to the buildkite run that prompted this ref counter bump.
 
 #=
+411
+- Update to ClimaParams 1.1.9 from ClimaParams 1.1.6
+
 410
 - Add entr/detr shear-mixing source to the TKE budget and drop the stability-biased
   ᶜN²_eff in favor of the unbiased ᶜbuoygrad for center l_N, Pr_t, and the Smag-Lilly
--- a/.buildkite/Manifest-v1.11.toml
+++ b/.buildkite/Manifest-v1.11.toml
@@ -2,7 +2,7 @@
 
 julia_version = "1.11.9"
 manifest_format = "2.0"
-project_hash = "eef602b151ef4127374ab25f5cd9fc279c3306eb"
+project_hash = "388c9c74c1e718a8c96f80cc8a564161860be89d"
 
 [[deps.ADTypes]]
 deps = ["PrecompileTools"]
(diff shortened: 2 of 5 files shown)


ph-kev:
The reproducibility tests are failing. I think this is due to using CUDA v6 instead of CUDA v5.

szy21:
> The reproducibility tests are failing. I think this is due to using CUDA v6 instead of CUDA v5.

There was some discussion about how RRTMGP is reproducible with CUDA v5, but not CUDA v6. We can bump the ref counter and see what happens. On the other hand, coupler has a compat CUDA = "5" (I don't know why). Should we do the same thing for atmos?

ph-kev:
> > The reproducibility tests are failing. I think this is due to using CUDA v6 instead of CUDA v5.
> 
> There was some discussion about how RRTMGP is reproducible with CUDA v5, but not CUDA v6. We can bump the ref counter and see what happens. On the other hand, coupler has a compat CUDA = "5" (I don't know why). Should we do the same thing for atmos?

I can bump the ref counter and use CUDA v6. I think the compat for Coupler has `CUDA = "5"` because @imreddyTeja needed to make changes to allow CUDA v6 to be used for the coupled simulation. I can add a compat, but it is going to be `CUDA = "5,  6"`.

imreddyTeja:
> > > The reproducibility tests are failing. I think this is due to using CUDA v6 instead of CUDA v5.
> 
> > 
> 
> > There was some discussion about how RRTMGP is reproducible with CUDA v5, but not CUDA v6. We can bump the ref counter and see what happens. On the other hand, coupler has a compat CUDA = "5" (I don't know why). Should we do the same thing for atmos?
> 
> 
> 
> I can bump the ref counter and use CUDA v6. I think the compat for Coupler has `CUDA = "5"` because @imreddyTeja needed to make changes to allow CUDA v6 to be used for the coupled simulation. I can add a compat, but it is going to be `CUDA = "5,  6"`.

If there are reproducibility issues with CUDA v6, we should be able to limit the compat in the .build kite env to v5. There was no "required" reason to switch to v6, but I thought it would be good to at least enable support for it, because it has cached gpu compilation

reproducibility_tests/ref_counter.jl
```diff
@@ -1,4 +1,4 @@
-410
+411
```
nefrathenrici: Is this still needed if we're sticking with CUDA.jl v5?

nefrathenrici (approved):

reproducibility_tests/ref_counter.jl
ph-kev: Thanks for catching that! We don't.

PR #4820, opened 2026-09-11
Add on-the-fly orographic gravity wave drag preprocessing 

costachris:
Right now Earth OGW drag input comes from precomputed HDF5 artifacts that bake in resolution and parameters. This adds `raw_topo_online` (reusing the same machinery) so it's computed on the fly at init from ETOPO + ClimaParams (γ, h_frac, α_smoothing). Keeps raw_topo as-is for the fast/reproducible path, with a warning when params don't match the artifact. Replaces hard-coded alpha smoothing parameter with `ogw_smoothing_scale_fraction`, which is now in ClimaParams.


src/parameterized_tendencies/gravity_wave_drag/orographic_gravity_wave.jl
```diff
@@ -57,6 +61,46 @@ A `NamedTuple` of surface-level `Field`s: the effective maximum and minimum subg
 obstacle heights `hmax`, `hmin` [m], and the four components `t11`, `t12`, `t21`, `t22` of
 the orographic tensor `T = −∇χ (∇h)ᵀ`, stored with `tᵢⱼ = −∂χ/∂xⱼ · ∂h/∂xᵢ`.
 """
+# Shape parameters the shipped `raw_topo` (`ogw_computed_drag_h*`) artifacts were
+# generated with, i.e. the ClimaParams defaults for `ogw_mountain_height_width_exponent`,
+# `ogw_critical_height_threshold`, and `ogw_smoothing_scale_fraction`.
+const RAW_TOPO_ARTIFACT_γ = 0.4
```
nefrathenrici, line 67: Should these be sourced from ClimaParams?

nefrathenrici (commented):
I mostly have comments for code outside this PR diff:
- [elev = FT.(Array(ds["z"]))[1:skip_pt:end, 1:skip_pt:end]](https://github.com/CliMA/ClimaAtmos.jl/blob/main/src/parameterized_tendencies/gravity_wave_drag/orographic_gravity_wave_helper.jl#L671C1-L671C65) this line should be a strided read without the `Array` call: `FT.(ds["z"][1:skip_pt:end, 1:skip_pt:end])`
- `compute_OGW_info` does computation on every MPI rank. . Gate the preprocessing behind `ClimaComms.iamroot(ctx)` and `ClimaComms.bcast` the resulting `(; lon, lat, hmax, hmin, t11, t12, t21, t22)`, and keep the per-rank temp-NetCDF write + `regrid_OGW_info` as is

src/parameterized_tendencies/gravity_wave_drag/orographic_gravity_wave.jl
costachris: These were the parameters used to make the existing `ogw` artifact. The idea is that if the user changes these parameters (or they change in ClimaParams in the future) and try to use the existing artifact a warning will highlight this.

ray-chew: Yes, thanks for inserting this warning. I created the artifacts initially to not add any more compute time for the standard CI tests (these settings are what's used by the CIs).

ray-chew (approved):
I verified the unit tests passed. LGTM.

PR #4823, opened 2026-09-13
Further docs additions/revisions

tapios:
New pages: LES subgrid-scale closures (Smagorinsky-Lilly, anisotropic minimum dissipation, constant horizontal diffusion) and forcings and idealized cases (subsidence, large-scale advection, column Coriolis, the composed external-forcing terms and nudging, Held-Suarez, prescribed flow). 

Rewritten: topography (elevation sources, smoothing rule, linear and SLEVE warps as implemented, what sloped coordinates imply for the discretization) and the global-simulations guide (resolution and timestep choice, common numerical configurations, the tested production configurations, output, scale). 

Symbol and source tables added to the microphysics and gravity-wave pages. 

No changes to source code. Best reviewed by reading the built docs.

Produced with Claude 5.1


docs/src/forcings.md
```diff
@@ -0,0 +1,272 @@
...
+it this way makes the scheme consistent for a uniform tracer: both terms cancel
+level by level, so ``\chi \equiv 1`` produces no tendency.
+
+The vertical operator carries a zero-flux condition at the top and bottom faces.
+For the usual case of inflow through the lid, the two terms cancel there as
+well, which is equivalent to holding ``\chi`` at its top-level value in the air
+descending into the domain.
+
+Subsidence acts on the total enthalpy, total water, and, with nonequilibrium
+microphysics, cloud liquid and cloud ice. Rain and snow are not subsided; they
```
szy21: Should we apply subsidence to rain and snow too?

tapios: I think we should because q_tot is subsiding (and so if we do not include precip, we are implicitly moving vapor). Condensate is also not handled consistently in the analytical vs driven cases. We should make this consistent sometime, but it'll probably make little practical difference. 

I updated the docs here to be more clear and consistent with what we actually do, and added a TODO.

docs/src/forcings.md
```diff
@@ -0,0 +1,287 @@
...
+| Parameter              | ClimaParams name                          | Default |
+|:---------------------- |:----------------------------------------- |:------- |
+| ``\tau_r`` for scalars | `gcmdriven_scalar_relaxation_timescale`   | 86400 s |
+| ``\tau`` for momentum  | `gcmdriven_momentum_relaxation_timescale` | 21600 s |
+| ``z_i``                | `gcmdriven_relaxation_minimum_height`     | 3000 m  |
+| ``z_r``                | `gcmdriven_relaxation_maximum_height`     | 3500 m  |
+
+A term may instead be given a constant timescale, a profile, or a height mask.
+The ARM VARANAL case raises ``z_i`` and ``z_r`` to 5500 m and 8000 m, since its
+convective boundary layer is far deeper than the GCM-driven case's.
```
szy21:
```suggestion
```
(I think we can leave this detail)

docs/src/global_simulations.md
```diff
@@ -24,38 +22,197 @@ CA.solve_atmos!(simulation)
...
+
+  - `netcdf_output_at_levels`, `true` by default, interpolates horizontally onto
+    a longitude–latitude grid but leaves the vertical on model levels. Set it to
+    `false` to interpolate onto height levels as well; those levels are spaced
+    exponentially to approximate constant-pressure surfaces, but the coordinate
+    stays height. For true pressure coordinates, set `pressure_coordinates: true`
+    on the diagnostic.
+  - `netcdf_interpolation_num_points` overrides the size of the target grid, as
+    a list, for example `[360, 180, 2]` for the high-resolution baroclinic-wave
+    runs.
```
szy21:
```suggestion
    runs. Note that the target z grid (the third index) is ignored when `netcdf_output_at_levels` is `true`.
```
(Something like this)

docs/src/microphysics.md
```diff
@@ -147,32 +151,37 @@ hyperdiffusion.
 
 ### Diffusion
 
-ClimaAtmos provides different horizontal and vertical diffusion schemes that
+ClimaAtmos provides horizontal and vertical diffusion schemes that
```
szy21: I feel the diffusion section doesn't belong here, but I'm not sure where it would be a good place. Also, I wouldn't call vertical diffusion "schemes that improve model stability". It's more physical. But I'm fine with keeping it here before we find a better place for this section.

docs/src/forcings.md
tapios: Removed

docs/src/global_simulations.md
tapios: Done.

docs/src/microphysics.md
tapios: Changed the language, and moved diffusion to its own page.

szy21 (approved):
I skimmed through most of the files and they look fine. I'm not too familiar with les_sgs and topography, so I didn't read those files carefully, and it might be good if someone else takes a look.

trontrytel (approved):
I skimmed through a random selection of files here and they looked good.

PR #4824, opened 2026-09-13
Apply the vertical Smagorinsky diffusion implicitly; drop unused AMD cache

tapios:
The implicit diffusion Jacobian already had a Smagorinsky branch, but the vertical Smagorinsky–Lilly tendency was always applied explicitly, so with `implicit_diffusion: true` the solver linearized a term that was not in the implicit right-hand side. The tendency now follows `diff_mode`, the eddy viscosity is refreshed on every Newton iterate, and the fields it writes live in the autodiff Jacobians. A vertically-acting Smagorinsky closure now satisfies the implicit-diffusion configuration check, and the three checks in `check_case_consistency`, previously chained with `elseif`, are independent.

Running under `use_auto_jacobian`/`use_dense_jacobian` requires ClimaCore 0.16.3 (CliMA/ClimaCore.jl `ts/promote`), whose `Number * Tensor` return types keep `Dual` storage; the compat bound enforces this. So this will only work after ClimaCore [PR 2638](https://github.com/CliMA/ClimaCore.jl/pull/2638) merges.

Also: the number-density redistribution in vertical diffusion, hyperdiffusion, the viscous sponge, and the PROPHET diffusive flux looked up `ρn_icl`, a field no configuration carries; it now looks up `ρn_ice`. Four AMD precomputed fields that no tendency read are no longer allocated. 

`vapor_constraint_qtot` is documented as leaving `ρq_tot` unchanged.

Co-produced with Fable 5.1


trontrytel (approved):
Thank you! 

Could we add a CI job with `smagorinsky_lilly: UVW` and `implicit_diffusion: true`? Right now none of this is tested.

trontrytel:
Also, my Claude thinks that we now have one unneeded call to `set_smagorinsky_lilly_precomputed_quantities!` (in `eddy_diffusivity_coefficients!` line 993). The PR moves computing those to `set_implicit_precomputed_quantities!`, so it should be updated at every Newton iteration. Worth double checking.

config/default_configs/default_config.yml
```diff
@@ -195,7 +195,7 @@ tracer_nonnegativity_method:
     - `vapor_tendency`: Weakly enforce nonnegativity by applying a tendency to each tracer to redistribute tracer mass between vapor (q_tot) and each tracer
     - `vertical_water_borrowing`: Enforce nonnegativity using VerticalMassBorrowingLimiter, which redistributes tracer mass vertically. Uses a single threshold (0.0) for all tracers to which the limiter is applied.
     - `~` (default): Do not enforce nonnegativity
-    Note: append `_qtot` to the method name to enforce nonnegativity of q_tot as well (not applicable to `vertical_water_borrowing`).
+    Note: append `_qtot` to `elementwise_constraint` to include q_tot among the constrained tracers. `vapor_constraint_qtot` is accepted but leaves q_tot unchanged; `vapor_tendency` and `vertical_water_borrowing` ignore the suffix with a warning.
```
trontrytel, line 205: We should also update above the docstring for `smagorinsky_lilly`. The supported options are now `"UVW", "W", "UV_W"`

tapios: Yes, thank you. Done.

tapios:
> Thank you!
> 
> Could we add a CI job with `smagorinsky_lilly: UVW` and `implicit_diffusion: true`? Right now none of this is tested.

Thank you. Yes, done.

tapios:
> Also, my Claude thinks that we now have one unneeded call to `set_smagorinsky_lilly_precomputed_quantities!` (in `eddy_diffusivity_coefficients!` line 993). The PR moves computing those to `set_implicit_precomputed_quantities!`, so it should be updated at every Newton iteration. Worth double checking.

Good catch. Fixed.

PR #4826, opened 2026-09-14
Speed up Larcform1 case

jefftyoder:
## Purpose 
Improve SYPD for Larcform1 case in CI. Current implementation is unnecessarily slow.

## Content
- Increased dt: 30s --> 120s
- Increased dt_rad: 30s --> 30mins
- Trimmed diagnostics (worst offenders were period: 10mins)
- Added clarifying comments on key config fields


jefftyoder:
The "Prek" check is failing on a hook to fix missing EOF newlines. However, I've added a *second* newline to the config, and the check is still failing.

trontrytel (approved):

trontrytel:
> The "Prek" check is failing on a hook to fix missing EOF newlines. However, I've added a _second_ newline to the config, and the check is still failing.

Does it work if there are no new empty lines at the end?

PR #4828, opened 2026-09-15
Update to ClimaTimeSteppers v1

ph-kev:
This PR updates the `Project.toml` and manifest to use ClimaTimeSteppers v1. The benchmark tests are updated to use ClimaTimeSteppers v1. I used a LLM to update the benchmark tests.


imreddyTeja (approved):
thanks!

.buildkite/Project.toml
```diff
@@ -53,6 +53,7 @@ ClimaComms = "0.6.9"
 ClimaCore = "0.16"
 ClimaCoreSpectra = "0.1"
 ClimaCoreTempestRemap = "0.3"
+ClimaTimeSteppers = "1"
```
szy21, line 56: I still don't know what's the rule of adding a package in .buildkite/Project.toml or the main Project.toml. But I think it's fine.

szy21 (approved):

.buildkite/Project.toml
ph-kev: It was added here since the tests use features of ClimaTimeSteppers v1 but nothing in src depends on ClimaTimeSteppers v1.

PR #4835, opened 2026-09-17
Exclude Aqua 0.8.17 and update Julia version to 1.10

ph-kev:
This PR make the GHA CI not use Aqua 0.8.17 which fail on CI (see https://github.com/CliMA/ClimaAtmos.jl/actions/runs/35251130582/job/105303949353?pr=4811#step:6:559). Also, this PR bumps the minimum version of Julia to v1.10 since ClimaAtmos is not installable on Julia 1.9.


Project.toml
```diff
@@ -47,7 +47,7 @@ ClimaAtmosMusica = "Musica"
 
 [compat]
 Adapt = "4.1"
-Aqua = "0.8.9"
+Aqua = "0.8.9 - 0.8.16, 0.8.18"
```
ph-kev, line 50: This could break if the bug is still in Aqua v0.8.18 (note that this is not released yet).

nefrathenrici (approved):
Thanks!

nefrathenrici:
I see the Aqua tests in the infrastructure action have passed so I'm going to merge now and unblock CI.

PR #4838, opened 2026-09-17
Error when a sponge damping height equals the domain top

yashica-patodia:
## Purpose

The Rayleigh and viscous sponge profiles are

```julia
ζ_rayleigh(s, z, zmax) = sinpi((z - s.zd) / (zmax - s.zd) / 2)^2
```

so a damping height `zd` equal to the domain top divides by zero. Nothing checks for this, and the failure is hard to trace back to the config:

- at the top face, `0 / 0` gives `NaN`, and `αₘ * ζ = 0 * NaN = NaN`;
- at every level below, `(z - zd) / 0 = -Inf`, and on the CPU `sinpi(-Inf)` throws `DomainError with Inf` from inside the first tendency evaluation (`ifelse` evaluates both branches, so the `z > zd` mask does not protect it).

`config/model_configs/rcemipii_box_CRM_1M.yml` is in this state on `main`: it sets `z_max: 30000.0`, while `toml/rcemipii_box.toml` sets `zd_rayleigh = 30000`. From the history, that value was chosen in #4286 for `runscripts/rcemipii_box_CRM_1M.jl`, which uses `z_max = 33_000`; the YAML was re-added later with a 30 km top. The YAML is not in the Buildkite pipeline, which is presumably why this went unnoticed.

## Reproducer

ClimaParams defaults `zd_rayleigh` to 15 km, so on `main`:

```julia
import ClimaAtmos as CA
config = CA.AtmosConfig(Dict(
    "config" => "column", "initial_condition" => "DecayingProfile",
    "rayleigh_sponge" => true, "z_max" => 15000.0, "z_elem" => 10,
    "dt" => "10secs", "t_end" => "20secs", "output_default_diagnostics" => false,
))
CA.solve_atmos!(CA.get_simulation(config))
# main:    ERROR: DomainError with Inf  (from `sinpi`, under a very long broadcast stack trace)
# this PR: ERROR: `rayleigh_sponge` has its damping height `zd` = 15000.0 m equal to the
#          domain top `z_max`, so the sponge layer has zero depth and its damping profile
#          divides by zero. Lower `zd` (the `zd_rayleigh` parameter) or raise `z_max`.
```

Tested locally on CPU only (Julia 1.12, macOS): `test/config/atmos_model.jl` passes and the reproducer behaves as above. I have not run the RCEMIP box config end to end, and have not tested on GPU.


src/types.jl
```diff
@@ -2579,6 +2609,9 @@ function _atmos_model(; kwargs...)
     params = get(atmos_model_kwargs, :params, nothing)
     setup = get(atmos_model_kwargs, :setup, nothing)
 
+    isnothing(grid) ||
+        check_sponge_heights(sponge, Spaces.z_max(get_spaces(grid).face_space))
```
nefrathenrici, line 2613: I don't think this will be called from the copy constructor `AtmosModel(model::AtmosModel; changes...)` in `src/types.jl`. Could you ensure that both constructors call `check_sponge_heights`?

nefrathenrici (approved):
Thanks!

PR #4842, opened 2026-09-21
Fixes various bugs/consistency issues that surfaced while writing docs

tapios:
### Physics
- **Pressure drag coefficient.** The momentum equation floored the environment area at `1 − a_max` and the TKE return-to-isotropy source at `a_min`, four orders of magnitude apart, so the energy removed from the draft and the energy given to TKE used different coefficients. Both now call one function, `pressure_drag_coefficient`. (@sajjadazimi please check if this is consistent with what you want).
- **LES energy flux.** Smagorinsky–Lilly, AMD, and constant horizontal diffusion diffused a lumped `h_tot`, unlike the other diffusive terms, which use the split form `-ρ D [∇s_d + (h_eff + Φ) ∇q_tot_eff]`. The LES closures now use the same split. The AMD scalar diffusivity comes from one helper and uses the physical gradient norm in both directions (the horizontal closure used covariant components).
- **Hyperdiffusion exclusion set.** The cloud-species loop scaled `ρn_ice`, but the exclusion tuple listed `ρn_lcl` and `ρn_rai` only, so `ρn_ice` also received the full-strength `∇⁴` tendency, and the P3 rime species `ρq_rim` and `ρb_rim` received the full-strength tendency alone. The rime species now scale with cloud ice, and the exclusion set is the shared `gs_sedimenting_tracer_candidates`.
- **SGS tracer indexing.** Inside `for j in 1:n` loops, the entrainment tendency, the updraft constraints, the Rayleigh sponge, and the sedimentation cache hardcoded updraft 1, so with several updrafts the first would have received all tendencies. They now index updraft `j`. Latent at the default of one updraft.

### Configuration and setups
- **ERA5 forcing-file checks.** `return false` sat inside the `NCDataset` `do` block and exited only the closure, so both checks always returned `true` and a cached file that did not cover the run was kept. The block's value is now returned.
- **Config base names.** `is_unique_basename` compared an extension-stripped name against names with the extension, so it always returned `true`; the disambiguating branch also referenced an undefined `path_sep`. Both fixed.
- **ISDAC and Larcform1 TKE.** ISDAC had `prognostic_tke ? profile : 0`, the reverse of the other setups; Larcform1 accepted `prognostic_tke` and ignored it. ISDAC now follows the shared convention and Larcform1 drops the keyword.
- **Standalone `DecayingProfile`.** A setup built without `params` stored `nothing` for the thermodynamics parameters and failed in `center_initial_condition`. It now falls back to the model's parameters.
- **Surface overrides.** `SurfaceBoundaryOverrides.p` and `.beta` were stored and never read; ten setups set `p` to no effect. The fields are removed. (@nefrathenrici please check this)
- **Keyword constructors.** `AtmosNumerics` and the `EDMFXModel` convenience constructor ended in `kwargs...` that the body never read, so a misspelled option was accepted silently. The catch-all is removed.
- **Double vertical mixing.** Nothing rejected `vert_diff` together with PROPHET or a vertical Smagorinsky closure. Model construction now rejects the combination. (@szy21 is that ok?)
- **AMD configurations** inherited the default hyperdiffusion while the Smagorinsky configurations switch it off. Both AMD configurations now set `hyperdiff: ~`.
- **`orographic_gravity_wave: "linear"`** constructed a type with no tendency method and called a function defined nowhere. The option and `LinearOrographicGravityWave` are removed.
- **PROPHET velocity solve.** Corrected docs that misdescribed the area-weight factors in pressure drag closures.

Most changes are cosmetic. But please check the area-fraction and LES energy flux changes above.

Co-produced with Fable 5.1


src/types.jl
nefrathenrici:
```suggestion
    carrying per-cell overrides for surface humidity / winds / gustiness
```

nefrathenrici (commented):
I only had one nit.

src/parameterized_tendencies/les_sgs_models/constant_horizontal_diffusion.jl
szy21, line 1: No one is using constant horizontal diffusion, so we can consider removing it now or later.

src/prognostic_equations/implicit/initialize_implicit_problem.jl
```diff
@@ -181,14 +185,16 @@ function solve_sgs_u₃_implicit_stage_analytic!(Y, p, dtγ)
...
+        # Implicit entrainment E (w₀ - w) ≈ -E w / a⁰: the velocity-dependent
+        # rate `entr_vel_scale · |w|` contributes a quadratic sink (∝ w²) and the
+        # velocity-independent rates a linear sink (∝ w), each with one factor
+        # of 1/a⁰ from the eliminated velocity difference, matching the rate
+        # `compute_entrainment` uses in the ρa solve. The positive part of
+        # `area_bounding_entr_detr` is the entrainment branch of the signed
+        # area-bounding rate (see `area_bounding_entr_detr`). `entr_nonvel_rate`
+        # is included in the linear sink, with the assumption that the change
+        # in b/w is slow.
+        @. ᶠa += ᶠinterp(ᶜentr_vel_scaleʲs.:($$j) * ᶜa⁰_inv) / ᶠdz
```
szy21: I think the two a⁰ is from we are still assuming entrainment is proportional to w - w⁰ (more physical), even though we are assuming w⁰ to be 0 when calculating entrainment. @sajjadazimi knows more.

src/prognostic_equations/implicit/initialize_implicit_problem.jl
```diff
@@ -198,22 +205,17 @@ function solve_sgs_u₃_implicit_stage_analytic!(Y, p, dtγ)
...
+            ᶜaʲ = @. lazy(draft_area(Y.c.sgsʲs.:($$j).ρa, ᶜρʲs.:($$j)))
             # Use a scratch scalar here as @lazy results in a large fused kernel
             # that doesn't work on P100 GPUs.
             ᶜdrag_coeff = p.scratch.ᶜtemp_scalar_2
             @. ᶜdrag_coeff =
-                α_d / (2 * scale_height) *
-                (1 / sqrt(max(ᶜaʲ, a_min)) + sqrt(ᶜa⁰_inv))
-            @. ᶠa += ᶠinterp(ᶜdrag_coeff * ᶜa⁰_inv) / ᶠdz
+                pressure_drag_coefficient(α_d, scale_height, ᶜaʲ, ᶜa⁰, a_min, a_max)
+            @. ᶠa += ᶠinterp(ᶜdrag_coeff * ᶜa⁰_inv * ᶜa⁰_inv) / ᶠdz
```
szy21: I think there should be only one ᶜa⁰_inv here because the pressure drag is multiplied by ᶜa⁰ (eq. 24 on overleaf)

tapios: You're right, the ρ̂⁰/ρ weight cancels one factor. Restored the single 1/a⁰ and fixed the docs page.

src/parameterized_tendencies/les_sgs_models/constant_horizontal_diffusion.jl
tapios: Ok. Maybe better in a separate PR.

src/types.jl
tapios: Done, thanks.

tapios:
> I only had one nit.

Thank you!

src/prognostic_equations/implicit/initialize_implicit_problem.jl
tapios: Restored the two factors and added a comment to document the assumption. Happy to make the ρa solve use the relative velocity too if you and @sajjadazimi think this is what we should do.

docs/src/orographic_gravity_wave.md
```diff
@@ -386,7 +386,7 @@ For analytical topographies (DCMIP200, Hughes2023, Agnesi, Schar, Cosine2d, Cosi
...
 | Callback scheduling at `dt_ogw`                                        | [src/callbacks/get_callbacks.jl](https://github.com/CliMA/ClimaAtmos.jl/blob/main/src/callbacks/get_callbacks.jl), [src/callbacks/callbacks.jl](https://github.com/CliMA/ClimaAtmos.jl/blob/main/src/callbacks/callbacks.jl) |
-| Model types                                                            | [`ClimaAtmos.OrographicGravityWave`](@ref), [`ClimaAtmos.FullOrographicGravityWave`](@ref), [`ClimaAtmos.LinearOrographicGravityWave`](@ref)                                                                                 |
+| Model types                                                            | [`ClimaAtmos.OrographicGravityWave`](@ref), [`ClimaAtmos.FullOrographicGravityWave`](@ref)                                                                                                                                   |
 
 The scheme is selected with the `orographic_gravity_wave` key
 (`nothing`, `gfdl_restart`, `raw_topo`, or `linear`) and its update interval
```
szy21:
```suggestion
(`nothing`, `gfdl_restart`, or `raw_topo`) and its update interval
```

src/parameterized_tendencies/les_sgs_models/anisotropic_minimum_dissipation.jl
szy21, line 104: ᶜtemp_UVWxUVW is overwritten here, so the ∇ᶜu_uvw used below has a wrong value

src/config/model_getters.jl
```diff
@@ -1039,6 +1036,21 @@ function check_case_consistency(parsed_args)
...
     end
 
+    # The prescribed vertical diffusion, PROPHET, and a vertically acting
+    # Smagorinsky-Lilly closure each diffuse the same grid-mean fields in the
+    # vertical, so only one of them may be active.
+    smagorinsky_vertical = is_smagorinsky_vertical(
+        isnothing(smagorinsky_lilly) ? nothing :
+        SmagorinskyLilly(; axes = Symbol(smagorinsky_lilly)),
+    )
+    if !isnothing(vert_diff) && (!isnothing(turbconv) || smagorinsky_vertical)
```
szy21: Yes, this is good, but should we add amd_les too?

szy21 (approved):
Thanks! I looked at the EDMF changes and skimmed over the other changes.

src/prognostic_equations/implicit/initialize_implicit_problem.jl
sajjadazimi: We removed `w_0` from entrainment a long time ago because at that time we had updraft area growth via entrainment even when `w^j` was zero (from `w_0`) which would make simulations unstable. After solving area equation analytically, we don't have large area issues anymore (area is a constant in Newton solve). We can now safely pass `w_0` to entrainment I guess.

sajjadazimi (approved):
Thanks! looks good to me. The reason the floors were different is that in momentum equation solve area comes from ODE extrapolations, so `a_0` was floored at the physical value of `1-a_max`. But in tke return-to-isotropy area is already solved and guaranteed to be physical; the `a_min` floor was for extra caution.

docs/src/orographic_gravity_wave.md
tapios: Done

src/parameterized_tendencies/les_sgs_models/anisotropic_minimum_dissipation.jl
tapios: Wow, good catch. This was a serious (pre-existing) bug, which effectively suppressed the diffusion. Now fixed (gradient norm has its own scratch field).

tapios:
> Thanks! looks good to me. The reason the floors were different is that in momentum equation solve area comes from ODE extrapolations, so `a_0` was floored at the physical value of `1-a_max`. But in tke return-to-isotropy area is already solved and guaranteed to be physical; the `a_min` floor was for extra caution.

Thanks for the explanation and checking this!

src/config/model_getters.jl
tapios: Added

src/parameterized_tendencies/les_sgs_models/anisotropic_minimum_dissipation.jl
szy21: it's claude this time:)

PR #4845, opened 2026-09-21
Fetch ERA5 initial conditions with InitialConditions.jl

nefrathenrici:
If the relevant IC files were missing, ClimaAtmos previously stopped at an error telling you to go run WeatherQuest's Python script by hand. It now fetches and converts the data itself.

Changes
- Depend on InitialConditions.jl. `weather_model_data_path` now calls `fetch_initial_conditions` when no raw file is present
- Add to_z_levels_3d_model in src/utils/era5_model_levels.jl, ported from WeatherQuest. ERA5 stores the atmosphere on hybrid levels whose height varies per column with surface pressure and temperature, so it reconstructs geopotential height by hydrostatic integration over the IFS coefficients, then interpolates each column onto the target altitudes. Pressure comes from the same coefficients and is written as p_3d, interpolated in log pressure. The L137 coefficients are compiled in, so there is no data file to ship.


costachris (approved):
InitialConditions.ERA5.fetch_initial_conditions always writes era5_raw_YYYYMMDD_0000.nc and truncates to the calendar day (filenames.jl/fetch.jl), but we look for era5_raw_YYYYMMDD_HHMM.nc (recent PR). Until all the paths handle HHMM, can we either only auto-fetch for 00Z when using IC.jl and error clearly, or after fetching, open the file under the name IC.jl actually writes (e.g. via ERA5.raw_filename), and document that auto-fetch only supports 00Z starts for now

PR #4847, opened 2026-09-22
Update to CM 0.41

trontrytel:
This PR updated to CM.jl v 0.41. We now have to pass vertical velocity into 1M bulk microphysics tendencies


trontrytel:
Claude added some additional tests for beres gravity waves. Lmk if you want them or not.

src/cache/microphysics_cache.jl
```diff
@@ -930,17 +930,18 @@ function set_microphysics_tendency_cache!(
     sgs_quad = p.atmos.sgs_quadrature
     nsubs = mp1m.n_substeps
     nsubs_quad = mp1m.n_substeps_quad
+    ᶜw_air = @. lazy(w_component(Geometry.WVector(ᶜu)))
```
szy21:
```suggestion
    ᶜw_air = @. lazy(w_component(WVec(ᶜu)))
```
(There are other places in the code using Geometry.WVector instead of WVec too, we can change it now or later)

toml/longrun_aquaplanet_1M.toml
```diff
@@ -7,9 +7,15 @@ value = 150
...
 
+# Kessler1M (CloudMicrophysics >= 0.41) blends the rain autoconversion timescale between
+# a stratiform and a convective value with the vertical velocity; override both to keep
+# the classic velocity-independent scheme with a 150 s timescale.
 [rain_autoconversion_timescale]
 value = 150
 
+[rain_autoconversion_timescale_stratiform]
+value = 150
+
```
szy21, line 18: we can also remove all of these - I don't know why they are here.

src/parameterized_tendencies/gravity_wave_drag/non_orographic_gravity_wave.jl
```diff
@@ -644,22 +646,27 @@ function compute_beres_convective_heating!(Y, p, ᶜN)
...
             ᶜρʲ = ᶜρʲs.:($j)
             ᶜTʲ = ᶜTʲs.:($j)
+            ᶜuʲ = ᶜuʲs.:($j)
             ᶜq_totʲ = ᶜq_tot_nonnegʲs.:($j)
             ᶜq_lclʲ = Y.c.sgsʲs.:($j).q_lcl
             ᶜq_iclʲ = Y.c.sgsʲs.:($j).q_icl
             ᶜq_raiʲ = Y.c.sgsʲs.:($j).q_rai
             ᶜq_snoʲ = Y.c.sgsʲs.:($j).q_sno
+            # physical draft vertical velocity [m/s] for the velocity-dependent autoconversion
+            ᶜwʲ_air = @. lazy(get_physical_w(ᶜuʲ, ᶜlg))
```
szy21: why do we use `get_physical_w` here instead of the `w_component(Wvec(ᶜuʲ))`, like elsewhere? Maybe they are the same

szy21 (approved):
I didn't read the gravity wave change, but the rest looks good to me.

toml/longrun_aquaplanet_1M.toml
trontrytel: I'll keep them for now. Next thing I'll test with varying them anyway.

szy21: sounds good. Just note that these are not the values we use in amip.

src/cache/microphysics_cache.jl
trontrytel: Changed it everywhere in this file.

src/parameterized_tendencies/gravity_wave_drag/non_orographic_gravity_wave.jl
trontrytel: They are the same here. I think we have both, because sometimes we have to pass `lg` separately, because the broadcast magic doesn't work. But thats not the case here. I changed it.

trontrytel:
I need to bump the ref counter because we changed the default prescribed N_ice to 1e5.

sajjadazimi (approved):
Thanks!

